PR #56, opened 2016-04-26
Added event notifying when a page is rendered

anishashetty-okta:


src/util/BaseLoginController.js
```diff
@@ -92,8 +92,14 @@ define(['okta', 'vendor/lib/q'], function (Okta, Q) {
     toggleButtonState: function (state) {
       var button = this.$el.find('.button');
       button.toggleClass('link-button-disabled', state);
-    }
+    },
 
+    // Event triggered at after a page is rendered along with the classname to identify the page
+    postRender: function() {
```
rchild-okta: We shouldn't do this in postRender - we generally don't modify the hook functions in base classes. I.e. if we do this, then we have to modify every controller that has a postRender hook to makes sure to call the parent class postRender method (i.e. RecoveryChallengeController.js).

src/util/BaseLoginController.js
```diff
@@ -92,8 +92,14 @@ define(['okta', 'vendor/lib/q'], function (Okta, Q) {
...
+      _.defer(_.bind(function () {
+        this.trigger('pageRenderedEvent',[this.className]);
```
rchild-okta: Double checking - does every controller set it's className correctly?

Also, spacing.

rchild-okta: Probably nicer to just call it "pageRendered" (the "event" is implied)

src/util/BaseLoginController.js
```diff
@@ -92,8 +92,14 @@ define(['okta', 'vendor/lib/q'], function (Okta, Q) {
     toggleButtonState: function (state) {
       var button = this.$el.find('.button');
       button.toggleClass('link-button-disabled', state);
-    }
+    },
 
+    // Event triggered at after a page is rendered along with the classname to identify the page
```
rchild-okta: Nit: at after

Better comment would be 'Trigger a "pageRendered" event after the page is rendered'

src/util/BaseLoginController.js
anishashetty-okta: I added the postRender to handle cases where there are animations on the page .
I had taken care of controllers calling their parent class in another commit. I'll rectify that now

src/util/BaseLoginRouter.js
```diff
@@ -113,6 +114,11 @@ function (Okta, Backbone, xdomain, RefreshAuthStateController, Settings, Header,
       this.listenTo(this.appState, 'navigate', function (url) {
         this.navigate(url, { trigger: true });
       });
+
+      // triggering event up the chain
+      this.listenTo(Okta.Controller.prototype, 'pageRenderedEvent', function (className) {
```
rchild-okta: Looking at this now, I think we should go the "use bubble property route", i.e. it's not as nice to listen to the prototype here. Feels better to listen to an instance when it's created.

src/util/BaseLoginController.js
rchild-okta: Oh, I also think we should pass the argument (className) back as an object rather than as a string. We're going to have more events, so it'd be nice to say "each event is triggered with some data; in the case of the pageRendered event, it triggers with data { page: 'primary-auth' }".

src/widget/OktaSignIn.js
```diff
@@ -117,9 +116,17 @@ var OktaSignIn = (function () {
     require.config(requireConfig);
     OktaAuth = require('vendor/OktaAuth');
     Util = require('util/Util');
+    LoginRouter = require('LoginRouter');
+
 
     authClient = new OktaAuth({ uri: options.baseUrl, transformErrorXHR: Util.transformErrorXHR });
-    _.extend(this, getProperties(authClient, options));
+    _.extend(this, getProperties(authClient, options, LoginRouter));
```
rchild-okta: nit: personal preference, but I think it'd be nicer to have LoginRouter as the first or second arg (i.e. options usually go last).

src/widget/OktaSignIn.js
```diff
@@ -117,9 +116,17 @@ var OktaSignIn = (function () {
...
+    _.extend(this, LoginRouter.prototype.Events);
```
rchild-okta: _.extend can take multiple args, so we can use:
```javascript
// Note: I think it's nicer to extend events first, just in case later on we actually want to listen to it in the getProperties method.
_.extend(this, LoginRouter.prototype.Events, getProperties(stuff));
```

src/widget/OktaSignIn.js
```diff
@@ -117,9 +116,17 @@ var OktaSignIn = (function () {
...
+
+    // Triggers the event up the chain so it is available to the consumers of the widget.
+    this.listenTo(LoginRouter.prototype, 'pageRenderedEvent', function(className) {
```
rchild-okta: I think in this case we want to listen to everything (like what bubble will give us) and trigger everything. This'll make it so that we don't have to update this code when we add more events.

rchild-okta:
Add tests!! In general, everything we write should have tests.

src/util/BaseLoginController.js
rchild-okta: Hmm, also, are we sure this is the right time to do this? I think this won't actually be consistent:
- we have animations, so at the time we're rendered/postRendered it's possible to not have added the controller to the dom yet (it won't be a problem on primaryAuth, but could be a problem on the other pages that go through the animation).
- Check out BaseLoginRouter.render

src/util/BaseLoginController.js
```diff
@@ -21,6 +21,7 @@ define(['okta', 'vendor/lib/q'], function (Okta, Q) {
   }
 
   return Okta.Controller.extend({
+    className: 'base-login-controller',
```
rchild-okta: Hmm, I'm not sure this will work as we want it - i.e. it will be overwritten by any controller that extends it. What's the purpose of this?

src/util/BaseLoginRouter.js
```diff
@@ -183,6 +185,10 @@ function (Okta, Backbone, xdomain, RefreshAuthStateController, Settings, Header,
       var oldController = this.controller;
       this.controller = new Controller(controllerOptions);
 
+      this.listenTo(this.controller, 'all', function () {
```
rchild-okta: nit: Looking at it now, we can do this (also note that we should add a comment describing why we're doing this):
```javascript
// Bubble up all controller events
this.listenTo(this.controller, 'all', this.trigger);
```

src/util/BaseLoginRouter.js
```diff
@@ -199,9 +205,10 @@ function (Okta, Backbone, xdomain, RefreshAuthStateController, Settings, Header,
 
         if (!oldController) {
           this.el.append(this.controller.el);
+          this.controller.triggerEvent();
```
rchild-okta: This only gets called when the first page is rendered. I'm guessing you won't get these events when you navigate to the next page (i.e. if you had mfa enabled, you'd see an event on primary auth, but not on the mfa challenge page).

rchild-okta: Also, I think we should call this function something more generic (again, in line with the idea that this is just the first of many events we are going to be publishing).

How about:
```javascript
this.controller.postAnimation();
```

src/util/BaseLoginController.js
```diff
@@ -92,8 +94,14 @@ define(['okta', 'vendor/lib/q'], function (Okta, Q) {
     toggleButtonState: function (state) {
       var button = this.$el.find('.button');
       button.toggleClass('link-button-disabled', state);
-    }
+    },
 
+     // Event triggered after a page is rendered along with the classname to identify the page
+    triggerEvent: function() {
+      _.defer(_.bind(function () {
```
rchild-okta: It turns out that the way we're now invoking things no longer needs the defer, so we can remove it. The animation sequence attaches both controllers to the dom, and does some transition stuff. When we call it after render/success, we are guaranteed that it's attached to the dom and there are no other controllers.

src/widget/OktaSignIn.js
```diff
@@ -117,9 +116,21 @@ var OktaSignIn = (function () {
...
     Util = require('util/Util');
+    LoginRouter = require('LoginRouter');
+
 
     authClient = new OktaAuth({ uri: options.baseUrl, transformErrorXHR: Util.transformErrorXHR });
-    _.extend(this, getProperties(authClient, options));
+    _.extend(this, LoginRouter.prototype.Events, getProperties(authClient, options, LoginRouter));
+
+    // Triggers the event up the chain so it is available to the consumers of the widget.
+    this.listenTo(LoginRouter.prototype, 'all', function() {
```
rchild-okta: Same nit here about replacing with:
```javascript
this.listenTo(LoginRouter.prototype, 'all', this.trigger);
```
I realize that it's the way it was in framework/view - not sure why they did that.

src/widget/OktaSignIn.js
```diff
@@ -117,9 +116,21 @@ var OktaSignIn = (function () {
...
+      this.trigger.apply(this, arguments);
+    });
+
+    this.on('pageRendered', function(classname) {
```
rchild-okta: Remove this. For testing, it's probably better to just test from the consumer side anyway (i.e. if you're working in okta-signin-widget, you can modify https://github.com/okta/okta-signin-widget/blob/master/buildtools/templates/index.tpl to make sure the actual widget is publishing those events.

test/spec/LoginRouter_spec.js
```diff
@@ -533,6 +533,20 @@ function (Okta, Q, Backbone, xdomain, SharedUtil, OktaAuth, Util, Expect, Router
       });
     });
 
+    describe('Trigger event is called', function () {
+      itp('Event is triggerrd on router start', function () {
+        return setup()
+            .then(function (test) {
+              test.router.primaryAuth();
+              spyOn(test.router.controller, 'triggerEvent');
```
rchild-okta: Some comments:

1) Better names for describe/test:
```javascript
// Describe is a general subheader type name
describe('Events', function () {
  // it statements should be read as if you start with "it", i.e. "it triggers a ..."
  // these two are necessary because we go down two different code paths
  it('triggers a pageRendered event when first controller is loaded');
  it('triggers a pageRendered event when navigating to a new controller');
});
```

2) Better to test "functionality", vs. testing that "function was called", i.e.:
```javascript

// In setup, create a spy that we can check against
    function setup(settings) {
      // stuff
      var eventSpy = new jasmine.createSpy('eventSpy');
      var router = new Router(_.extend({
        // stuff
      }, settings));
      router.on('all', eventSpy);
      return tick().then(function () {
        return {
           // stuff
           eventSpy: eventSpy
        };
      });
    }

// And then in your test function you can check that the spy was called
```

3) You should actually test what event was triggered
```javascript
// Make sure we're not sending too many events
expect(test.eventSpy.calls.count()).toBe(1);

// Expect that we got the event that we're expecting
expect(test.eventSpy).toHaveBeenCalledWith('pageRendered', { page: 'primary-auth' );
```

src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -12,8 +12,7 @@
 
 define([
   'okta',
-  'views/shared/PasswordJammer'
```
rchild-okta: We should delete the PasswordJammer file too, right?

src/util/BaseLoginController.js
anishashetty-okta: I had this in case the property was not overwritten in the subclass.

src/util/BaseLoginRouter.js
```diff
@@ -211,6 +217,7 @@ function (Okta, Backbone, xdomain, RefreshAuthStateController, Settings, Header,
           success: function () {
             var flashError = this.appState.get('flashError'),
                 model = this.controller.model;
+            this.controller.postRenderAnimation();
```
rchild-okta: we should add this below line 222, i.e. after the oldController has been removed from the dom

src/util/BaseLoginController.js
```diff
@@ -92,8 +94,12 @@ define(['okta', 'vendor/lib/q'], function (Okta, Q) {
     toggleButtonState: function (state) {
       var button = this.$el.find('.button');
       button.toggleClass('link-button-disabled', state);
-    }
+    },
 
+     // Event triggered after a page is rendered along with the classname to identify the page
+    postRenderAnimation: function() {
+        this.trigger('pageRendered', {page: this.className});
```
rchild-okta: nits: Comment should probably be inside the function since it's talking about the trigger

Also, it looks like the spacing here is not correct (should not be indented as much)

src/widget/OktaSignIn.js
```diff
@@ -117,9 +116,15 @@ var OktaSignIn = (function () {
     require.config(requireConfig);
     OktaAuth = require('vendor/OktaAuth');
     Util = require('util/Util');
+    LoginRouter = require('LoginRouter');
+
 
     authClient = new OktaAuth({ uri: options.baseUrl, transformErrorXHR: Util.transformErrorXHR });
-    _.extend(this, getProperties(authClient, options));
+    _.extend(this, LoginRouter.prototype.Events, getProperties(authClient, options, LoginRouter));
```
rchild-okta: nit: Switch order of passing in options and LoginRouter (options is usually the last argument)

test/spec/LoginRouter_spec.js
```diff
@@ -55,17 +55,20 @@ function (Okta, Q, Backbone, xdomain, SharedUtil, OktaAuth, Util, Expect, Router
       var setNextResponse = Util.mockAjax();
       var baseUrl = 'https://foo.com';
       var authClient = new OktaAuth({uri: baseUrl});
+      var eventSpy = new jasmine.createSpy('eventSpy');
```
rchild-okta: no new before jasmine.createSpy

test/spec/LoginRouter_spec.js
```diff
@@ -533,6 +536,38 @@ function (Okta, Q, Backbone, xdomain, SharedUtil, OktaAuth, Util, Expect, Router
       });
     });
 
+    describe('Events', function () {
+      itp('triggers a pageRendered event when first controller is loaded', function() {
+        return setup()
+        .then(function (test) {
+          test.router.primaryAuth();
+          return tick(test)
```
rchild-okta: Since we're working with Promises, it's nicer to just return the promise and not have the indent, i.e. [here](https://github.com/anishashetty-okta/okta-signin-widget/blob/as-OKTA-87152/test/spec/LoginRouter_spec.js#L226).
```javascript
// good
return someP(function (test) {
  return tick(test);
})
.then(function (test) {
  // continue on here
});

// not good
return someP(function (test) {
  return tick(test)
  .then(function () {
    // not so bad with one promise level, but gets much worse later
  });
})


```

test/spec/LoginRouter_spec.js
```diff
@@ -533,6 +536,38 @@ function (Okta, Q, Backbone, xdomain, SharedUtil, OktaAuth, Util, Expect, Router
...
+              .then(function(test){
+                expect(test.eventSpy.calls.count()).toBe(1);
+                expect(test.eventSpy).toHaveBeenCalledWith('pageRendered', { page: 'primary-auth'});
+              });
+        });
+      });
+      itp('triggers a pageRendered event when navigating to a new controller', function() {
+        return setup()
```
rchild-okta: indenting seems inconsistent; it's similar to plugin in that we do:
```javascript
return setup()
.then(function () { // indent is at same level as "r"
  
});
```

test/spec/LoginRouter_spec.js
```diff
@@ -533,6 +536,38 @@ function (Okta, Q, Backbone, xdomain, SharedUtil, OktaAuth, Util, Expect, Router
...
+            .then(function (test) {
+              // Test navigation from primary Auth to Forgot password page
+              test.router.primaryAuth();
+              Util.mockRouterNavigate(test.router);
+              test.router.navigate('signin/forgot-password');
```
rchild-okta: Since we're leaving a page, it's nice to do something like (before test.router.navigate):
```javascript
// reset call count
test.eventSpy.calls.reset();
```
This might help you hone in on why it's triggering more events than you expect (and makes for a cleaner test, where the next assertion should be that the call count is "1" again.

src/util/BaseLoginController.js
rchild-okta: Do we have any cases of this happening currently?

anishashetty-okta: No. there aren't any current cases , so removed the class-name prop from the base class

rchild-okta:
lgtm 🚀

PR #57, opened 2016-04-26
Okta Verify Auto Push

yumingcao-okta:
Adding auto push checkbox to Okta Verify factor

[Requirement Doc] (https://oktawiki.atlassian.net/wiki/display/PM/Theme:+Core+MFA+Invest#Theme:CoreMFAInvest-AutoPushforNewSignOnUI)
[Design Doc] (https://oktawiki.atlassian.net/wiki/display/eng/Okta+Verify+Auto+Push+Design)

push

OKTA-86615

Not included in this PR: consolidate trustDevice, rememberMe, and autoPush into a single cookie, wanted to collect more feedback on benefit of doing this before proceeding

Backend feature flag PR: https://github.com/okta/okta-core/pull/14085

@rchild-okta @mauriciocastillosilva-okta @srinivasanagandla-okta @Ujjwalreddy-okta

### DIFF
--- a/src/MfaVerifyController.js
+++ b/src/MfaVerifyController.js
@@ -10,11 +10,12 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-/*jshint maxcomplexity:11, maxparams:10 */
+/*jshint maxcomplexity:12, maxparams:11 */
 define([
   'okta',
   'shared/views/forms/inputs/CheckBox',
   'util/BaseLoginController',
+  'util/CookieUtil',
   'views/mfa-verify/TOTPForm',
   'views/mfa-verify/YubikeyForm',
   'views/mfa-verify/SecurityQuestionForm',
@@ -23,7 +24,7 @@ define([
   'views/mfa-verify/InlineTOTPForm',
   'views/shared/FooterSignout'
 ],
-function (Okta, Checkbox, BaseLoginController, TOTPForm, YubikeyForm, SecurityQuestionForm, CallAndSMSForm,
+function (Okta, Checkbox, BaseLoginController, CookieUtil, TOTPForm, YubikeyForm, SecurityQuestionForm, CallAndSMSForm,
           PushForm, InlineTOTPForm, FooterSignout) {
 
   return BaseLoginController.extend({
@@ -73,6 +74,20 @@ function (Okta, Checkbox, BaseLoginController, TOTPForm, YubikeyForm, SecurityQu
         this.add(InlineTOTPForm, {
           options: { model: this.model.get('backupFactor') }
         });
+
+        if (this.settings.get('features.autoPush')) {
+          this.add(Checkbox, {
+            options: {
+              model: this.model,
+              name: 'autoPush',
+              placeholder: Okta.loc('autoPush', 'login'),
+              label: false,
+              'label-top': false,
+              className: 'margin-btm-0'
+            }
+          });
+        }
+
         // Remember Device checkbox resides outside of the Push and TOTP forms.
         if (this.settings.get('features.rememberDevice')) {
           this.add(Checkbox, {
@@ -100,6 +115,15 @@ function (Okta, Checkbox, BaseLoginController, TOTPForm, YubikeyForm, SecurityQu
           this.options.appState.get('isMfaRequired')) {
         return true;
       }
+      // update auto push cookie after user accepts Okta Verify MFA
+      if (this.options.factorType == 'push') {
+        if (this.settings.get('features.autoPush') && this.model.get('autoPush')) {
+          CookieUtil.setAutoPushCookie(this.options.appState.get('userId'));
+        } else {
+          CookieUtil.removeAutoPushCookie(this.options.appState.get('userId'));
+        }
+      }
+      return false;
     }
   });
 
--- a/src/models/AppState.js
+++ b/src/models/AppState.js
@@ -175,6 +175,15 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
           return factors && factors.length > 1;
         }
       },
+      'userId': {
+        deps: ['lastAuthResponse'],
+        fn: function (res) {
+          if (!res._embedded || !res._embedded.user) {
+            return null;
+          }
+          return res._embedded.user.id;
+        }
+      },
       'isPwdExpiringSoon': {
         deps: ['lastAuthResponse'],
         fn: function (res) {
--- a/src/models/Settings.js
+++ b/src/models/Settings.js
@@ -52,6 +52,7 @@ function (Okta, Errors, BrowserFeatures) {
       'features.rememberMe': ['boolean', true, true],
       'features.rememberDevice': ['boolean', true, true],
       'features.rememberDeviceAlways': ['boolean', true, false],
+      'features.autoPush': ['boolean', true, false],
       'features.smsRecovery': ['boolean', true, false],
       'features.windowsVerify': ['boolean', true, false],
       'features.selfServiceUnlock': ['boolean', true, false],
--- a/src/util/CookieUtil.js
+++ b/src/util/CookieUtil.js
@@ -10,11 +10,12 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-define(['okta', 'vendor/plugins/jquery.cookie'], function (Okta) {
+define(['okta', 'util/CryptoUtil', 'vendor/plugins/jquery.cookie'], function (Okta, CryptoUtil) {
 
   var $ = Okta.$;
   var LAST_USERNAME_COOKIE_NAME = 'ln';
   var REMEMBER_DEVICE_COOKIE_NAME = 'rdln';
+  var AUTO_PUSH_COOKIE_PREFIX  = 'auto_push_';
   var DAYS_SAVE_REMEMBER = 365;
 
   function removeCookie (name) {
@@ -26,6 +27,10 @@ define(['okta', 'vendor/plugins/jquery.cookie'], function (Okta) {
       expires: DAYS_SAVE_REMEMBER,
       path: '/'
     });
+  }
+
+  function getAutoPushKey(userId) {
+    return AUTO_PUSH_COOKIE_PREFIX + CryptoUtil.getStringHash(userId);
   }
 
   var fn = {};
@@ -52,6 +57,27 @@ define(['okta', 'vendor/plugins/jquery.cookie'], function (Okta) {
 
   fn.removeDeviceCookie = function () {
     removeCookie(REMEMBER_DEVICE_COOKIE_NAME);
+  };
+
+  fn.isAutoPushEnabled = function (userId) {
+    if (userId === undefined) {
+      return false;
+    }
+    return $.cookie(getAutoPushKey(userId)) === 'true';
+  };
+
+  fn.setAutoPushCookie = function (userId) {
+    if (userId === undefined) {
+      return;
+    }
+    setCookie(getAutoPushKey(userId), true);
+  };
+
+  fn.removeAutoPushCookie = function (userId) {
+    if (userId === undefined) {
+      return;
+    }
+    removeCookie(getAutoPushKey(userId));
   };
 
   return fn;
--- a/src/views/mfa-verify/PushForm.js
+++ b/src/views/mfa-verify/PushForm.js
@@ -10,8 +10,11 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-define(['okta'], function (Okta) {
+define(['okta',
+        'util/CookieUtil'
+], function (Okta, CookieUtil) {
 
+  var _ = Okta._;
   // deviceName is escaped on BaseForm (see BaseForm's template)
   var titleTpl = Okta.Handlebars.compile('{{factorName}} ({{{deviceName}}})');
 
@@ -49,6 +52,12 @@ define(['okta'], function (Okta) {
         factorName: this.model.get('factorLabel'),
         deviceName: this.model.get('deviceName')
       });
+
+      if (this.settings.get('features.autoPush') && CookieUtil.isAutoPushEnabled(this.options.appState.get('userId'))) {
+        this.model.set('autoPush', true);
+        // trigger push once DOM is fully loaded
+        _.defer(_.bind(this.submit, this));
+      }
     },
     setSubmitState: function (ableToSubmit) {
       var button = this.$el.find('.button');
@@ -62,7 +71,9 @@ define(['okta'], function (Okta) {
       }
     },
     submit: function (e) {
-      e.preventDefault();
+      if (e !== undefined) {
+        e.preventDefault();
+      }
       if (this.enabled) {
         this.setSubmitState(false);
         this.doSave();
--- a/test/helpers/dom/MfaVerifyForm.js
+++ b/test/helpers/dom/MfaVerifyForm.js
@@ -3,6 +3,7 @@ define(['./Form'], function (Form) {
   var ANSWER_FIELD = 'answer';
   var SHOW_ANSWER_FIELD = 'showAnswer';
   var REMEMBER_DEVICE = 'rememberDevice';
+  var AUTO_PUSH = 'autoPush';
 
   return Form.extend({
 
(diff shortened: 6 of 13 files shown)


src/views/mfa-verify/PushForm.js
```diff
@@ -49,6 +63,19 @@ define(['okta'], function (Okta) {
         factorName: this.model.get('factorLabel'),
         deviceName: this.model.get('deviceName')
       });
+
+      var that = this;
+      if (that.enabled) {
+        window.setTimeout(function () {
```
yumingcao-okta: Maybe there's a better way handling this?

src/util/CookieUtil.js
```diff
@@ -54,5 +55,26 @@ define(['okta', 'vendor/plugins/jquery.cookie'], function (Okta) {
     removeCookie(REMEMBER_DEVICE_COOKIE_NAME);
   };
 
+  fn.isAutoPushEnabled = function (username) {
```
srinivasanagandla-okta: Instead of username, can you use the "userId" ? 
The path to that attribute in the response JSON is: "_embedded->user->id".

src/util/CookieUtil.js
```diff
@@ -54,5 +55,26 @@ define(['okta', 'vendor/plugins/jquery.cookie'], function (Okta) {
...
+      return false;
+    }
+    return $.cookie(AUTO_PUSH_COOKIE_PREFIX + CryptoUtil.getStringHash(username)) === "true";
+  };
+
+  fn.setAutoPushCookie = function (username) {
+    if (username === undefined) {
+      return;
+    }
+    setCookie(AUTO_PUSH_COOKIE_PREFIX + CryptoUtil.getStringHash(username), true);
```
rchild-okta: Let's create a private helper function to generate the cookie name (i.e. don't have the same logic in two places, which can break if someone just updates one in the future).

src/views/mfa-verify/PushForm.js
```diff
@@ -29,6 +31,18 @@ define(['okta'], function (Okta) {
 
     initialize: function () {
       this.enabled = true;
+      this.listenTo(this.options.appState, 'change:isSuccessResponse',
```
rchild-okta: Couple comments:
- This won't work in the general case, i.e. we are hardcoding the idea that we get a SUCCESS response after the verify state (check out this diagram for state transitions: http://developer.okta.com/assets/img/auth-state-model.png)
- In general if we wanted to do something on success, we'd handle it elsewhere (i.e. this view should only have knowledge about its particular authStatus)

Why not do it the way that rememberDevice does it, i.e. when the user clicks the checkbox?

src/views/mfa-verify/PushForm.js
```diff
@@ -29,6 +31,18 @@ define(['okta'], function (Okta) {
...
+        function (state, isSuccessResponse) {
+          if (isSuccessResponse && this.settings.get('features.autoPush')) {
+            var autoPush = $('.custom-checkbox [name=autoPush]');
```
rchild-okta: In general with backbone code, you'll never see an unscoped jquery selector (i.e. just $(selector)). Usually you'll use something like this.$(selector), which is scoped to just the current view.

src/views/mfa-verify/PushForm.js
```diff
@@ -29,6 +31,18 @@ define(['okta'], function (Okta) {
...
+            if (autoPush[0].checked) {
```
rchild-okta: Continuing on that thread, we're generally not working with the dom elements themselves - those views are usually tied to a model, and we can access the model to get the current value. Check out Ujjwal's last pr for rememberDevice:
https://github.com/yumingcao-okta/okta-signin-widget/commit/7ad708db07845d89a8ad2cb84fac1d424b3a57e0

src/views/mfa-verify/PushForm.js
```diff
@@ -49,6 +63,19 @@ define(['okta'], function (Okta) {
         factorName: this.model.get('factorLabel'),
         deviceName: this.model.get('deviceName')
       });
+
+      var that = this;
```
rchild-okta: nit: convention at okta for backbone stuff is to use "self", not "that"

src/views/mfa-verify/PushForm.js
rchild-okta: Yep:
1. Put this logic in a postRender
2. Just call the function that submit calls when it's clicked, i.e. there's no reason to simulate a DOM event when we ourselves control the handler that is triggered when the event happens

i.e.
```javascript
postRender: function () {
  if (featureEnabledAndCookiebla) {
    this.submit();
  }
}
```

src/views/mfa-verify/PushForm.js
yumingcao-okta: ack

src/views/mfa-verify/PushForm.js
yumingcao-okta: @rchild-okta We don't want to save any changes before MFA `SUCCESS`, thus we are not modifying cookie on user clicks. Seems `rememberDevice` [approach] (https://github.com/okta/okta-signin-widget/commit/7ad708db07845d89a8ad2cb84fac1d424b3a57e0#diff-324bfcc07b8afb141c2419766be33d92R182) is updating cookie on push button click, that means if a user want to disable auto push, simply uncheck the box wouldn't work, b/c he has to `DENY` the notification first. I'm not sure this is the best user experience. So now wondering what might be the best place to handle successful MFA, any suggestion?

src/views/mfa-verify/PushForm.js
yumingcao-okta: turns out `submit` method could not take effect during postRender phase, staying with `setTimeout` for dom manipulation.

src/views/mfa-verify/PushForm.js
rchild-okta: The problem is that this only works if you have no other states that happen after the mfa challenge - it's not "MFA SUCCESS", it's "Auth SUCCESS". So for example, if you're password has expired, this code won't work because this controller will already be removed from the DOM.

You can do this:
Create a trapAuthResponse function. This gets called before we navigate to the next controller. Set the cookie there, and return false.

More info here: https://github.com/okta/okta-signin-widget/blob/master/src/util/RouterUtil.js#L97

src/views/mfa-verify/PushForm.js
rchild-okta: I mentioned this yesterday, but if that was the case, then wrap the submit function in a defer. We should not be "clicking on the button to trigger a submit".

yumingcao-okta:
@rchild-okta @srinivasanagandla-okta comments addressed

src/views/mfa-verify/PushForm.js
```diff
@@ -49,6 +51,25 @@ define(['okta'], function (Okta) {
...
+      }
+    },
+    preRender: function () {
+      if (this.isAutoPushEnabled()) {
+        this.model.set('autoPush', true);
+      };
+    },
+    isAutoPushEnabled: function () {
+      return this.settings.get('features.autoPush') &&
+          CookieUtil.isAutoPushEnabled(this.options.appState.get("lastAuthResponse")._embedded.user.id);
```
rchild-okta: nit: would be better to put _embedded.user.id as a derived property in AppState.js (there are a bunch of examples in there, let me know if you have questions about it)

src/views/mfa-verify/PushForm.js
```diff
@@ -49,6 +51,25 @@ define(['okta'], function (Okta) {
         factorName: this.model.get('factorLabel'),
         deviceName: this.model.get('deviceName')
       });
+
+      var self = this;
+      if (self.enabled) {
```
rchild-okta: Hmmm, judging by the code, we're always enabled at this point

src/views/mfa-verify/PushForm.js
```diff
@@ -49,6 +51,25 @@ define(['okta'], function (Okta) {
         factorName: this.model.get('factorLabel'),
         deviceName: this.model.get('deviceName')
       });
+
+      var self = this;
```
rchild-okta:
```javascript
if (this.isAutoPushEnabled()) {
  _.defer(_.bind(this.submit, this));
}
```
Where we usually bring in _ by doing the following at the top of the file:
```javascript
var _ = Okta._;
```

src/MfaVerifyController.js
```diff
@@ -100,6 +115,14 @@ function (Okta, Checkbox, BaseLoginController, TOTPForm, YubikeyForm, SecurityQu
           this.options.appState.get('isMfaRequired')) {
         return true;
       }
+      if (this.options.appState.get('isSuccessResponse')) {
```
rchild-okta: This is still not going to work - check out this diagram:
http://developer.okta.com/docs/api/resources/authn.html#transaction-state

After MFA_CHALLENGE, we can't assume that we're going to hit SUCCESS (even if the user has accepted the challenge). We can hit PASSWORD_WARN or PASSWORD_EXPIRED, which I'd imagine should still save their cookie value. I think the other times I've brought this up I've added some suggestions to work around this - ask me if you have any questions.

yumingcao-okta: Gotcha, missed the MFA -> `PASSWORD_EXPIRED` part (still wondering why do we need this path when we already have `PRIMARY_Authentication` -> `PASSWORD_EXPIRED`...)

rchild-okta: Yeah, the general idea is that the backend controls the auth flow - in the case where mfa is required, we don't go from primary auth -> password_expired, we go from primaryauth -> mfa -> password_expired.

That's one of the reasons why the controllers are set up as they are - they're in charge of getting to the next state, but they never know what the next state is (and why we'll never see logic like here where we say something like "if the state is success, do this").

src/MfaVerifyController.js
```diff
@@ -100,6 +115,15 @@ function (Okta, Checkbox, BaseLoginController, TOTPForm, YubikeyForm, SecurityQu
           this.options.appState.get('isMfaRequired')) {
         return true;
       }
+      // update auto push cookie after user accepts Okta Verify MFA
+      if (this.options.factorType == 'push') {
+        if (this.settings.get('features.autoPush') && !!this.model.get('autoPush')) {
```
rchild-okta: Curious - what is the type of the value that is returned by this.model.get('autoPush')? Is it not a boolean?

src/MfaVerifyController.js
```diff
@@ -100,6 +115,15 @@ function (Okta, Checkbox, BaseLoginController, TOTPForm, YubikeyForm, SecurityQu
...
+          CookieUtil.setAutoPushCookie(this.options.appState.get("userId"));
```
rchild-okta: I'm guessing lint will break, but you shouldn't use double quotes here and in line 123

src/models/AppState.js
```diff
@@ -175,6 +175,15 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
           return factors && factors.length > 1;
         }
       },
+      'userId': {
+        deps: ['lastAuthResponse'],
```
rchild-okta: I was wondering if we get a userId when we hit PASSWORD_WARN or PASSWORD_EXPIRED, and it looks like we do. So we're good to use userId (in case that came up).

src/util/CookieUtil.js
```diff
@@ -54,5 +59,26 @@ define(['okta', 'vendor/plugins/jquery.cookie'], function (Okta) {
     removeCookie(REMEMBER_DEVICE_COOKIE_NAME);
   };
 
+  fn.isAutoPushEnabled = function (userId) {
+    if (userId === undefined) {
+      return false;
+    }
+    return $.cookie(getAutoPushKey(userId)) === "true";
```
rchild-okta: single quotes - fyi we did notice yesterday that we weren't reporting lint failures in travis, but once you rebase you'll be able to see this fail in travis.

src/views/mfa-verify/PushForm.js
```diff
@@ -49,6 +52,20 @@ define(['okta'], function (Okta) {
...
+      };
+    },
+    preRender: function () {
+      if (this.isAutoPushEnabled()) {
+        this.model.set('autoPush', true);
+      };
+    },
+    isAutoPushEnabled: function () {
+      return this.settings.get('features.autoPush') &&
+          CookieUtil.isAutoPushEnabled(this.options.appState.get("userId"));
```
rchild-okta: single quotes

src/views/mfa-verify/PushForm.js
```diff
@@ -49,6 +52,20 @@ define(['okta'], function (Okta) {
...
         deviceName: this.model.get('deviceName')
       });
+
+      // trigger push once DOM is fully loaded
+      if (this.isAutoPushEnabled()) {
+        _.defer(_.bind(this.submit, this));
+      };
+    },
+    preRender: function () {
+      if (this.isAutoPushEnabled()) {
```
rchild-okta: Let's just move this to initialize - i.e. where we're already doing the "if(this.isAutoPushEnabled)" check (L57). And since we're only doing the check in one place, we can probably also get rid of using an "isAutoPushEnabled" helper function.

yumingcao-okta: ack

src/views/mfa-verify/PushForm.js
```diff
@@ -62,7 +79,9 @@ define(['okta'], function (Okta) {
       }
     },
     submit: function (e) {
-      e.preventDefault();
```
yiyangli-okta: `(e) && (e.preventDefault())` will do the trick

or maybe directly call `_.defer(_.bind(this.doSave, this));` ?

test/helpers/dom/MfaVerifyForm.js
```diff
@@ -81,6 +82,28 @@ define(['./Form'], function (Form) {
...
+      return this.checkbox(AUTO_PUSH).prop('checked');
+    },
+
+    setAutoPush: function (val) {
+      var autoPush = this.autoPushCheckbox();
+      autoPush.prop('checked', val);
+      autoPush.trigger('change');
+    },
+
+    isPushSent: function () {
```
yiyangli-okta: is this just for testing only?

src/MfaVerifyController.js
yumingcao-okta: it's usually a boolean except when autoPush feature is disabled, this.model.get('autoPush') would return a `undefined`

src/models/Settings.js
```diff
@@ -52,6 +52,7 @@ function (Okta, Errors, BrowserFeatures) {
       'features.rememberMe': ['boolean', true, true],
       'features.rememberDevice': ['boolean', true, true],
       'features.rememberDeviceAlways': ['boolean', true, false],
+      'features.autoPush': ['boolean', true, false],
```
yiyangli-okta: if `features.autoPush` is `undefined` or a non-boolean value when `save`, the model.validate will fail. Since there is a `validate` to strict that, I think `!!this.model.get('autoPush')` could be `this.model.get('autoPush') === true`

yumingcao-okta: Sorry for the confusion, I meant `this.model.get(`autoPush`)` would be undefined when `features.autoPush` is `false` because the `autoPush` checkbox won't be [initialized] (https://github.com/okta/okta-signin-widget/pull/57/files#diff-8ccd3af6dc30ed000779477ac78c8de4R78).

test/helpers/dom/MfaVerifyForm.js
yumingcao-okta: yup

src/views/mfa-verify/PushForm.js
yumingcao-okta: We probably don't want to skip the `setSubmitState(false)` portion in `submit()` b/c we also want to update push button label?

src/models/Settings.js
yiyangli-okta: the default value of autoPush in model is false, it shouldn't be `undefined`, because if it's `undefined`, you can't submit the form. Unless there is a code, say `parse` after `sync` set it to `undefined`.

test/helpers/dom/MfaVerifyForm.js
yiyangli-okta: so, there's no benefits to expose the method, just use it in tests.

src/models/Settings.js
yumingcao-okta: this.model.get('autoPush') is accessing value in model `Factor` instead of model `Settings`, For now there's no default value specified for Factor.autoPush, so it would remain undefined if autoPush checkbox is not initialized. I can add one though.

test/helpers/dom/MfaVerifyForm.js
yumingcao-okta: I'm not sure I follow, isn't this a test helper file?

yiyangli-okta: ah, my bad, yes, you are right, it's good.

rchild-okta:
lgtm - squash and I'll merge it 🚢

PR #58, opened 2016-04-27
Add 'Account successfully unlocked!' page

gregorydandrea-okta:
Resolves:OKTA-84374
account_unlocked

### DIFF
--- a/src/AccountUnlockedController.js
+++ b/src/AccountUnlockedController.js
@@ -1,0 +1,51 @@
+/*!
+ * Copyright (c) 2015-2016, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define([
+  'okta',
+  'util/Enums',
+  'util/FormController',
+  'util/FormType'
+],
+function (Okta, Enums, FormController, FormType) {
+
+  return FormController.extend({
+    className: 'account-unlocked',
+    Model: function () {
+      return {
+        local: {
+          userFullName: ['string', false, this.options.appState.get('userFullName')]
+        }
+      };
+    },
+
+    Form: {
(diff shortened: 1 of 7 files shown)


gregorydandrea-okta:
added a test to verify that the page is shown when the recoveryType is set on SUCCESS and that it navigates back to login when the "back to sign in" button pressed

ravikanodia-okta:
LGTM, though I'm not super familiar with this area.

🚢

src/AccountUnlockedController.js
```diff
@@ -0,0 +1,54 @@
...
+      return {
+        local: {
+          userFullName: ['string', false, this.options.appState.get('userFullName')]
+        }
+      };
+    },
+
+    Form: {
+      title: Okta.loc('account.unlock.unlocked.title', 'login'),
+      subtitle: function () {
```
rchild-okta: this doesn't need to be a function (only need a function when there are instance properties you need access to)

src/AccountUnlockedController.js
```diff
@@ -0,0 +1,54 @@
...
+      },
+      noButtonBar: true,
+      attributes: { 'data-se': 'account-unlocked' },
+      formChildren: [
+        FormType.Button({
+          title: Okta.loc('goback', 'login'),
+          className: 'button button-primary button-wide',
+          attributes: {'data-se': 'back-button'},
+          click: function () {
+            var self = this;
```
rchild-okta: We don't need "var self" here - we have access to this in the two following lines

test/helpers/dom/RecoveryQuestionForm.js
```diff
@@ -3,6 +3,7 @@ define(['./Form'], function (Form) {
   var ANSWER_FIELD = 'answer';
   var SHOW_ANSWER_FIELD = 'showAnswer';
   var CLASS_SELECTOR = '.recovery-question';
+  var BACK_TO_LOGIN_BTN = 'back-button';
```
rchild-okta: We shouldn't just add this to the RecoveryQuestionForm (it's not the RecoveryQuestion page, is it?) - these dom helpers are 1 per page

rchild-okta: Hmm, actually never mind; I see we are already doing this [here](https://github.com/okta/okta-signin-widget/blob/master/test/spec/UnlockAccount_spec.js#L196)

src/AccountUnlockedController.js
gregorydandrea-okta: will remove the superfluous function wrapping

src/AccountUnlockedController.js
gregorydandrea-okta: will do

rchild-okta:
lgtm - squash and i'll merge it 🚀

PR #62, opened 2016-05-10
Override focus() for text input in IE (to prevent placeholder text from disappearing)

Ujjwalreddy-okta:
OKTA-84944

### DIFF
--- a/src/util/BrowserFeatures.js
+++ b/src/util/BrowserFeatures.js
@@ -54,6 +54,10 @@ define(function () {
     return !!(win.history && win.history.pushState);
   };
 
+  fn.isIE = function () {
+    return /msie/i.test(navigator.userAgent);
+  };
+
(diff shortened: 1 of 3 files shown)


src/util/BrowserFeatures.js
```diff
@@ -54,6 +54,10 @@ define(function () {
     return !!(win.history && win.history.pushState);
   };
 
+  fn.isIE = function () {
```
rchild-okta: Do we need to worry about edge?

rchild-okta:
Overall looks good - would verify if edge has the same behavior, and if so add some detection for that as well.

Ujjwalreddy-okta:
Edge works as expected. The placeholder text is shown when the field is focused.

rchild-okta:
🚀

PR #63, opened 2016-05-11
Uses i18n module in okta-signin-widget.

ravikanodia-okta:


.gitignore
```diff
@@ -1,13 +1,20 @@
 # If a change is made in here, add it to .npmignore as well unless you
 # explicitly want to include it in the published artifact
-node_modules
+/node_modules/*
+!/node_modules/@okta
```
rchild-okta: Hmm, do we really have to do all of this? I think it's fine to include everything under node_modules/@okta/i18n (as long as there are no sub node_modules underneath it). We don't need to explicitly call out "everything except for the properties files" (if that's what you're doing here).

.gitignore
```diff
@@ -1,13 +1,20 @@
...
 npm-debug.log
 test/SpecRunner.html
 target
 build2
 .grunt
 .sass-cache
 .widgetrc
-
-# These paths should be included in the published artifact, but not the repo
-dist
```
rchild-okta: Why are we removing this? We should not be checking in the dist folder (which gets generated when you run a build).

Gruntfile.js
```diff
@@ -83,7 +83,10 @@ module.exports = function (grunt) {
 
     propertiesToJSON: {
       main: {
-        src: ['resources/nls/*.properties'],
+        src: [
```
rchild-okta: In the case of the sign-in-widget, we can actually get rid of most of the i18n logic (differs from the plugin). Instead of using the .properties files, we should use the exported requireJs folder.

When we switch to using that, we can get rid of these targets in the Gruntfile:
- propertiesToJson
- JSONToJs

As well as the corresponding dependency (grunt-properties-to-json) and buildtools/JSONToJs

.gitignore
ravikanodia-okta: Hmm. I did not see this folder. I can make this more specific - the problem is that "dist" matches node_modules/@okta/i18n/dist/ as well.

rchild-okta:
One thing that I just ran across - bundledDependencies:
https://docs.npmjs.com/files/package.json#bundleddependencies
http://stackoverflow.com/questions/11207638/advantages-of-bundleddependencies-over-normal-dependencies-in-npm

Not sure if it applies to this situation - can you check on it?

rchild-okta:
Can you separate out the commit into:
1) Files that have been deleted/added
2) The changes that you made to get this to work

Right now it's actually kind of hard to see what you've changed - for example, I am expecting to see some changes to package.json, but it's not visible in either commit.

rchild-okta:
Also, can you delete the buildtools that you're removing? i.e. jsonToJs. I'm also expecting package.json to remove the grunt-properties-to-json dependency

Gruntfile.js
```diff
@@ -303,17 +301,6 @@ module.exports = function (grunt) {
           }
         }
       },
```
yiyangli-okta: extra `,`

yiyangli-okta:
https://oktawiki.atlassian.net/wiki/display/eng/Setup+and+use+of+Phabricator+for+code+review

This would help to list all the changes to review.

Gruntfile.js
ravikanodia-okta: Removed, thanks.

ravikanodia-okta:
I checked out bundledDependencies and I don't think it's relevant to this situation.

rchild-okta:
Almost there - the shrinkwrap file is wrong (it contains references to artifacts). There's an "npm run shrinkwrap" command you can run to remove those (but I want to verify that we generated the whole thing correctly). I'm assuming this did not pass travis since it does not have access to our internal registry urls.

ravikanodia-okta:
Switching @okta/i18n to an optional dependency fixed the travis CI issue, but I'm not 100% positive that's the right thing to do. It feels like shrinkwrap was supposed to solve this for me.

ravikanodia-okta:
Yeah, I was able to juggle it around and get it fixed. Huzzah!

rchild-okta:
lgtm 🚀

PR #64, opened 2016-05-12
Add sms hint text and send via email link in Reset/Unlock via SMS flow

Ujjwalreddy-okta:
OKTA-81320
https://okta.box.com/s/rsrl3whn6acy150m8exer7qghsz9qj39

### DIFF
--- a/resources/nls/login.properties
+++ b/resources/nls/login.properties
@@ -116,6 +116,7 @@ password.newPassword.tooltip = New password
 password.confirmPassword.placeholder = Repeat password
 password.confirmPassword.tooltip = Repeat password
 password.error.match = New passwords must match
+recovery.sms.hint = SMS can only be used if a mobile phone number has been configured.
 
 # Enroll Choices
 enroll.choices.title = Set up multifactor authentication
@@ -228,6 +229,7 @@ password.forgot.emailSent.title = Email sent!
 password.forgot.emailSent.desc = Email has been sent to {0} with instructions on resetting your password.
 password.forgot.question.title = Answer Forgotten Password Challenge
 password.forgot.question.submit = Reset Password
+password.forgot.sms.notReceived = Didn't receive an SMS? Reset via email
 
 # Reset Password
 password.reset.title = Reset your Okta password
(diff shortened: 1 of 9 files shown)


resources/nls/login.properties
```diff
@@ -116,6 +116,7 @@ password.newPassword.tooltip = New password
 password.confirmPassword.placeholder = Repeat password
 password.confirmPassword.tooltip = Repeat password
 password.error.match = New passwords must match
+recovery.sms.hint = SMS can only be used if a mobile phone number has been configured.
```
rchild-okta: @ravikanodia-okta fyi before checking in your code for the i18n sign-in-widget change, we'll need to copy over this file again since we are modifying it in PR's like this one. 

**Important** When you copy it over, you should do this:
1. Copy file
2. Replace the last translated sha line with the one that's currently in i18n (https://github.com/okta/npm/blob/master/i18n/properties/login.properties#L2)

src/RecoveryChallengeController.js
```diff
@@ -91,7 +92,56 @@ function (Okta, FormController, FormType, FooterSignout) {
...
+        case Enums.RECOVERY_TYPE_UNLOCK:
+          sendEmailLink = Okta.loc('account.unlock.sms.notReceived', 'login');
+          break;
+        default:
+          break;
+      }
+
+      if (sendEmailLink) {
+        this.add(
+          '<a href="#" class="link send-email-link" data-se="send-email-link">' +
```
rchild-okta: Let's use a template for this (general rule of thumb to never concatenate html strings with variables)

test/helpers/dom/Form.js
```diff
@@ -90,7 +91,13 @@ define(['jquery', 'underscore', './Dom'], function ($, _, Dom) {
       }
 
       api = $(element).qtip('api');
-      return api ? api.show().tooltip.text() : undefined;
+      tooltipText = api ? api.show().tooltip.text() : undefined;
+
+      // Remove the tooltip from the DOM to
+      // prevent elements from lingering around.
+      api.tooltip.remove();
```
rchild-okta: nice

rchild-okta:
Overall, looks good. Just the change for string concat -> using a template, then 🚀

src/RecoveryChallengeController.js
Ujjwalreddy-okta: Yep, will do.

PR #69, opened 2016-05-27
Change X-Okta-SDK header to match widget version

lboyette-okta:
bacon: test
Resolves: OKTA-89060

Tests will be added when we move to webpack (OKTA-90896)

@rchild-okta 
@austinstanhope-okta

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,6 +4,7 @@
 // grunt test (to run test task)
 
 /*global module, process, JSON */
+var packageJson = require('./package.json');
 
 module.exports = function (grunt) {
   /* jshint maxstatements: false */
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -78,6 +79,17 @@ module.exports = function (grunt) {
...
   }
 
+  var WIDGET_VERSION_TEXT = '<%= widgetversion %>';
+
+  function escapeRegexText(text) {
+    return text.replace(/([.*+?^=!:${}()|\[\]\/\\])/g, '\\$1');
+  }
+
+  function substituteWidgetVersion(content) {
+    var widgetVersionRegex = new RegExp(escapeRegexText(WIDGET_VERSION_TEXT));
```
rchild-okta: I think we might have talked about this when we were doing it in the authSdk, but why do we need a regex here? Why not just a simple text replace?

lboyette-okta: Without a regex, `replace` only replaces the first instance. In this case, I guess a simple replace would work, because we only expect a single instance of the widget version.

Will change

austinstanhope-okta:
Looks ok to me

rchild-okta:
🚢

PR #71, opened 2016-06-02
Update xdomain to fix jquery 2.* compatibility

lboyette-okta:
resolves: OKTA-87632
bacon: test

@rchild-okta @Ujjwalreddy-okta


Ujjwalreddy-okta:
Can you also update the xdomain version in Third party notices file?

src/vendor/xdomain-0.7.5.js
```diff
@@ -546,7 +555,6 @@ XHookHttpRequest = window[XMLHTTP] = function() {
   return facade;
 };
 
-// OKTA - changed from this.define (which didn't work)
```
rchild-okta: Nice - it looks like they fixed an issue we had to patch

rchild-okta:
Agree with @Ujjwalreddy-okta , but after that lgtm 🚀

PR #73, opened 2016-06-06
Include the install-with-shrinkwrap module and the latest translation…

Ujjwalreddy-okta:
… strings

OKTA-91782


npm-shrinkwrap.json
```diff
@@ -1,1434 +1,1209 @@
...
   "dependencies": {
     "@okta/i18n": {
-      "version": "1.3.0-beta.3",
-      "from": "@okta/i18n@1.3.0-beta.3"
+      "version": "1.11.0"
+    },
+    "@okta/install-with-shrinkwrap": {
+      "version": "1.2.0",
+      "dependencies": {
+        "ansi-regex": {
```
rchild-okta: Can you verify what the behavior is here? Main worry is that we're going to include subdependencies when we check in our node_modules/@okta folder (i.e. it looks like these dependencies would be in the node_modules folder of install-with-shrinkwrap).

rchild-okta:
For the properties, are we adding in a bunch of the dist files that we were not including before (seems like the diff has too many additions)? It's possible that @ravikanodia-okta had set it up in a way (or did it manually when he was adding the original set) that only the requireJs files were checked in. Or perhaps only the login.properties/their variants.

rchild-okta:
Seems like my comment got lost, reposting outside of the shrinkwrap commit. It's regarding the dependencies within the @okta/install-with-shrinkwrap:

"Can you verify what the behavior is here? Main worry is that we're going to include subdependencies when we check in our node_modules/@okta folder (i.e. it looks like these dependencies would be in the node_modules folder of install-with-shrinkwrap)."

PR #74, opened 2016-06-06
Update okta-auth-js to v1.0.2

lboyette-okta:
bacon: test
resolves: OKTA-91850

@rchild-okta 
@mauriciocastillosilva-okta 
@Ujjwalreddy-okta

### DIFF
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -12,7 +12,7 @@ var OktaSignIn = (function () {
      * @param callback - callback function invoked with 'true'/'false' as the argument.
      */
     function checkSession(callback) {
-      authClient.existingSession().then(callback);
+      authClient.session.exists().then(callback);
     }
 
     /**
@@ -21,7 +21,7 @@ var OktaSignIn = (function () {
      * @param callback - function to invoke after closing the session.
      */
     function closeSession(callback) {
-      authClient.closeSession().then(callback)
+      authClient.session.close().then(callback)
       .fail(function () {
         callback('There was a problem closing the session');
       });
@@ -33,7 +33,10 @@ var OktaSignIn = (function () {
      * @param callback - function to invoke after refreshing the session.
      */
     function refreshSession(callback) {
-      authClient.refreshSession().then(callback);
+      authClient.session.refresh().then(callback)
+      .fail(function() {
+        callback({status: 'INACTIVE'});
+      });
     }
 
     /**
@@ -45,7 +48,7 @@ var OktaSignIn = (function () {
      * @param opts - OAUTH options to refresh the idToken
      */
     function refreshIdToken(idToken, callback, opts) {
-      authClient.refreshIdToken(idToken, opts).then(callback)
+      authClient.idToken.refresh(opts).then(callback)
       .fail(function () {
         callback('There was a problem refreshing the id_token');
       });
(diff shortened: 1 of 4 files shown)


src/widget/OktaSignIn.js
```diff
@@ -45,7 +45,7 @@ var OktaSignIn = (function () {
      * @param opts - OAUTH options to refresh the idToken
      */
     function refreshIdToken(idToken, callback, opts) {
-      authClient.refreshIdToken(idToken, opts).then(callback)
+      authClient.idToken.refresh(opts).then(callback)
```
lboyette-okta: Now that we don't take an idToken param, we have a few options:

1. Change the signature of `refreshIdToken` not to take an idToken. This would be a breaking change and thus, a major version bump.
2. Keep the current signature, but shuffle the arguments if the first argument is a function. We could also warn consumers that passing an idToken is deprecated when one is passed. This wouldn't be a breaking change, but it does make the refresh method a little odd.
3. Keep the current signature, but customers pass `null` if they don't want to provide an idToken.

src/widget/OktaSignIn.js
```diff
@@ -12,7 +12,7 @@ var OktaSignIn = (function () {
      * @param callback - callback function invoked with 'true'/'false' as the argument.
      */
     function checkSession(callback) {
-      authClient.existingSession().then(callback);
+      authClient.session.exists().then(callback);
```
mauriciocastillosilva-okta: nit: the name of this method is a little bit confusing ```session.exists()``` sounds like a method to know if the session exists or not, that will return a boolean. Maybe a better one will be ```session.current()``` or something similar?

mauriciocastillosilva-okta:
I am assming ```src/vendor/OktaAuth.js``` was already reviewed, am I right?

src/widget/OktaSignIn.js
lboyette-okta: This resolves a promise with a boolean of whether the session exists. It has to be a promise because we make an api call for the check

lboyette-okta:
@mauriciocastillosilva-okta Yep! OktaAuth.js has been reviewed here (everything between 1.0.1 and 1.0.2): https://github.com/okta/okta-auth-js/pulls?q=is%3Apr+is%3Aclosed

src/widget/OktaSignIn.js
rchild-okta: I like 3 for now. Let's fix the widget, and file a followup jira to address this in the next major release.

rchild-okta: And jiras for adding tests if they don't exist yet.

rchild-okta:
@Ujjwalreddy-okta can you file a followup jira for @timlim-okta or someone else to verify all these OIDC flows after it's checked in?

rchild-okta:
Overall looks good, 🚀

mauriciocastillosilva-okta:
🐚

src/widget/OktaSignIn.js
lboyette-okta: OKTA-92186 - signatures for session and oauth flow methods
OKTA-92018 - tests

Ujjwalreddy-okta:
@timlim-okta Can you please test the OIDC/Social Auth flows with this version?
JIRA - https://oktainc.atlassian.net/browse/OKTA-92201

PR #80, opened 2016-06-13
Modify the css to show the app-logo banner always for SP initiated flow

Ujjwalreddy-okta:
OKTA-91641 - Before the fix, we were hiding the app logo banner if the device height was < 700px. The fix (1) removes this hiding logic and (2) reduces the padding in the app-logo banner to save 20px in the page's height. 

Before:
screen shot 2016-06-13 at 3 07 26 pm

After:
screen shot 2016-06-13 at 3 11 43 pm


rchild-okta:
lgtm 🚢

PR #85, opened 2016-07-01
Modify the IE regex to include IE11.

Ujjwalreddy-okta:
OKTA-94262

### DIFF
--- a/src/util/BrowserFeatures.js
+++ b/src/util/BrowserFeatures.js
@@ -55,7 +55,7 @@ define(function () {
   };
 
   fn.isIE = function () {
-    return /msie/i.test(navigator.userAgent);
+    return /(msie|trident)/i.test(navigator.userAgent);
   };
 
   return fn;


alexma-okta:
:rocket:

src/util/BrowserFeatures.js
```diff
@@ -55,7 +55,7 @@ define(function () {
   };
 
   fn.isIE = function () {
-    return /msie/i.test(navigator.userAgent);
+    return /msie/i.test(navigator.userAgent) || /trident/i.test(navigator.userAgent);
```
haishengwu-okta: super picky: 
does this work `/(msie|trident)/.test(...)`?

lboyette-okta: Is this how we want to check for IE at all? Why not see if document.documentMode exists?

Ujjwalreddy-okta: @haishengwu-okta done.
@lboyette-okta There are a couple of ways to do this, but checking for the useragent string seems to be the most commonly used, so I went with it.

PR #88, opened 2016-07-07
Make @okta/i18n and @okta/courage optional

lboyette-okta:
Resolves: OKTA-94736

Npm removes the entire node_modules folder when we publish, even if the files are included in .npmignore. BundledDependencies fixes this problem, but it doesn't currently work on scoped dependencies. This is a temporary workaround.

@rchild-okta 
@alexeystadnikov-okta

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,8 +60,6 @@
     "webpack": "1.13.1"
   },
   "dependencies": {
-    "@okta/courage": "1.3.0-beta.30",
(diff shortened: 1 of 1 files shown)


rchild-okta:
I noticed earlier that the shrinkwrap had changed (and caused a failing build), but you removed it. I'm guessing that it was not necessary because shrinkwrap doesn't care if it's optional deps or not?

If yes, lgtm 🚀

PR #90, opened 2016-07-21
Bump okta-auth-js to 1.3.0 and add X-Okta-User-Agent-Extended header

lboyette-okta:
@rchild-okta 
@alexma-okta

### DIFF
--- a/webpack.common.config.js
+++ b/webpack.common.config.js
@@ -65,20 +65,21 @@ module.exports = {
     new webpack.ProvidePlugin({
       process: empty,
       setImmediate: empty
-    }),
-
-    new webpack.DefinePlugin({
-      // This is a temporary fix for okta-auth-js not populating
-      // this when consumed via npm
-      SDK_VERSION: JSON.stringify(0),
-      STATE_TOKEN_COOKIE_NAME: JSON.stringify('oktaStateToken'),
-      DEFAULT_POLLING_DELAY: 500,
-      FRAME_ID: JSON.stringify('okta-oauth-helper-frame')
     })
   ],
   resolveLoader: {
     'alias': {
       'i18n': 'json'
     }
+  },
+  module: {
+    loaders: [
+      // Use json-loader only for config.json, because
+      // we alias i18n to the json loader, which
+      // causes webpack to attempt using json-loader
+      // twice. This behavior causes the build to fail 
+      // on i18n json files.
+      { test: /config\.json$/, loader: 'json' }
+    ]
   }
 };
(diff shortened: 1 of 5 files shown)


webpack.common.config.js
```diff
@@ -68,17 +69,22 @@ module.exports = {
...
     }
+  },
+  module: {
+    loaders: [
+      // Use json-loader only for config.json, because
+      // we alias i18n to the json loader, which
+      // causes webpack to attempt using json-loader
+      // twice. This behavior causes the build to fail 
+      // on i18n json files.
+      { test: /config\.json$/, loader: 'json' }
```
lboyette-okta: The issues with i18n are specific to our build. However, any webpack consumer of okta-auth-js would still need to include `{ test: /\.json$/, loader: 'json' }`. The maintainers of webpack have explicitly said [they don't want to include json-loader by default](https://github.com/webpack/webpack/issues/184#issuecomment-36607967).

webpack.common.config.js
```diff
@@ -68,17 +69,22 @@ module.exports = {
     }),
 
     new webpack.DefinePlugin({
-      // This is a temporary fix for okta-auth-js not populating
-      // this when consumed via npm
-      SDK_VERSION: JSON.stringify(0),
-      STATE_TOKEN_COOKIE_NAME: JSON.stringify('oktaStateToken'),
-      DEFAULT_POLLING_DELAY: 500,
-      FRAME_ID: JSON.stringify('okta-oauth-helper-frame')
+      WIDGET_VERSION: JSON.stringify(packageJson.version)
```
alexma-okta: Is it necessary to apply JSON.stringify to the packageJson.vesion?

alexma-okta:
other than one comment, it looks good to me.

webpack.common.config.js
rchild-okta: @alexma-okta yeah, I had the same question when I first saw this. Turns out it's a webpack thing - https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin.

webpack.common.config.js
rchild-okta: Hmm, I see. Can you file a jira for us to look at this later and see if there's anything we can do about this?

rchild-okta:
lgtm 🚀

webpack.common.config.js
lboyette-okta: unfortunately, yes. DefinePlugin is basically a find/replace, so when webpack tries to replace `WIDGET_VERSION` in the code, it'll do it without quotes. `JSON.stringify` adds quotes.

webpack.common.config.js
lboyette-okta: Sure. An alternative is to generate a `config.js` instead of a `config.json` when we build okta-auth-js. I'll file a jira.

PR #91, opened 2016-07-21
OKTA-96127: update i18n to get Czech

mauriciocastillosilva-okta:
- Added files for Chech language
- Also, fixing small text changes in 3 tests due to the i18n update

Reviewer: please take a closer look at the 2 files below, it seems some paths change when I ran shrinkwrap, so just want to be sure its good
- node_modules/@okta/courage/package.json
- node_modules/@okta/i18n/package.json


rchild-okta:
nit - can you reword the commit message to drop the jira number into the body of the commit message? Also, you misspelled Czech in the commit message ("chech")

^^ Re the jira message - the reason is that jiras aren't exposed externally, and at some point we're going to move over to using github issues

rchild-okta:
Overall looks good to me - I'm hoping that this process will become better once npm supports bundled dependencies for scoped packages (i.e. won't have issues with the diffs in the package.json files).

lboyette-okta:
lgtm, ⛵

PR #94, opened 2016-08-11
Removes unnecessary dependency 'fsevents'

jmaldonado-okta:
...which is already consumed from 'chokidar' as optional.

Related: OKTA-94080


jmaldonado-okta:
@okta/travisci

rchild-okta:
I think this is okay temporarily, but we'll need to update our shrinkwrap generator module to do this for us in the future (i.e. next time we generate a shrinkwrap, it'll re-add fsEvents).

🚀

PR #95, opened 2016-08-15
Added password complexity requirements text to the reset password form.

alexeystadnikov-okta:
Resolved OKTA-92338

screen shot 2016-08-18 at 1 43 02 pm

@mauriciocastillosilva-okta 
@rchild-okta 

Please, review.

### DIFF
--- a/src/PasswordResetController.js
+++ b/src/PasswordResetController.js
@@ -45,6 +45,38 @@ function (Okta, FormController, FormType, ValidationUtil, FooterSignout, TextBox
     Form: {
       save: Okta.loc('password.reset', 'login'),
       title: Okta.loc('password.reset.title', 'login'),
+      subtitle: function () {
+        var policy = this.options.appState.get('policy');
+        if (!policy || !policy.complexity) {
+          return;
+        }
+
+        var fields = {
+          minLength: {i18n: 'password.complexity.length', args: true},
+          minLowerCase: {i18n: 'password.complexity.lowercase'},
+          minUpperCase: {i18n: 'password.complexity.uppercase'},
+          minNumber: {i18n: 'password.complexity.number'},
+          minSymbol: {i18n: 'password.complexity.symbol'},
+          excludeUsername: {i18n: 'password.complexity.no_username'}
+        };
+
+        var requirements = _.map(policy.complexity, function (complexityValue, complexityType) {
+          var params = fields[complexityType];
+
+          return params.args ?
+            Okta.loc(params.i18n, 'login', [complexityValue]) : Okta.loc(params.i18n, 'login');
+        });
+
+        if (requirements.length) {
+          requirements = _.reduce(requirements, function (result, requirement) {
+            return result ?
+              (result + Okta.loc('password.complexity.list.element', 'login', [requirement])) :
+              requirement;
+          });
+
+          return Okta.loc('password.complexity.description', 'login', [requirements]);
+        }
+      },
       formChildren: [
         FormType.Input({
           placeholder: Okta.loc('password.newPassword.placeholder', 'login'),
--- a/test/helpers/xhr/PASSWORD_RESET_withMaxComplexity.js
+++ b/test/helpers/xhr/PASSWORD_RESET_withMaxComplexity.js
@@ -1,0 +1,50 @@
+define({
+  "status": 200,
+  "responseType": "json",
+  "response": {
+    "stateToken": "testStateToken",
+    "expiresAt": "2015-07-17T02:33:34.874Z",
+    "status": "PASSWORD_RESET",
+    "_embedded": {
+      "policy": {
+        "complexity": {
+          "minLength": 8,
+          "minLowerCase": 1,
+          "minUpperCase": 1,
+          "minNumber": 1,
+          "minSymbol": 1,
+          "excludeUsername": true
+        }
+      },
+      "user": {
+        "id": "00uhveDu5xF26YA0j0g3",
+        "profile": {
+          "login": "administrator1@clouditude.net",
+          "firstName": "Add-Min",
+          "lastName": "O'Cloudy Tud",
+          "locale": "en_US",
+          "timeZone": "America\/Los_Angeles"
+        }
+      }
+    },
+    "_links": {
+      "next": {
+        "name": "resetPassword",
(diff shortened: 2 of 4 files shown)


rchild-okta:
As with other changes that affect the UI, can you attach a screenshot of what this looks like?

rchild-okta:
General: @mauriciocastillosilva-okta just checked in the latest i18n, so I'm guessing you can remove that from this PR.

src/PasswordResetController.js
```diff
@@ -45,6 +45,32 @@ function (Okta, FormController, FormType, ValidationUtil, FooterSignout, TextBox
     Form: {
       save: Okta.loc('password.reset', 'login'),
       title: Okta.loc('password.reset.title', 'login'),
+      subtitle: function () {
+        var policy = this.options.appState.get('policy');
+        if (policy.complexity) {
```
rchild-okta: I generally like to exit early with guard statements, i.e.:

```javascript
// Good:
if (!policy.complexity) {
  return;
}
// bunch of code

// Bad:
if (password.complexity) {
  // bunch of code
}
```

rchild-okta: Also, are we always guaranteed to have policy? I.e. if you don't have some feature flag/etc, is policy always defined or will this line throw an error?

test/spec/PasswordReset_spec.js
```diff
@@ -79,6 +79,12 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordResetForm, Beacon, Expect,
           .toEqual(['', { trigger: true }]);
       });
     });
+    itp('has a valid subtitle', function () {
```
rchild-okta: I'd expect there to be several more tests - if you add logic, test it. For example:
- What happens when they only have one?
- What happens if there are none?

src/PasswordResetController.js
alexeystadnikov-okta: Thanks. Done.

You are right, the `policy` is not guaranteed. Added additional condition.

test/spec/PasswordReset_spec.js
alexeystadnikov-okta: Added additional tests.

alexeystadnikov-okta:
@rchild-okta Any other comments?

src/PasswordResetController.js
```diff
@@ -45,6 +45,34 @@ function (Okta, FormController, FormType, ValidationUtil, FooterSignout, TextBox
...
+        complexity.minLength &&
+          requirements.push(Okta.loc('password.complexity.length', 'login', [complexity.minLength]));
+
+        complexity.minLowerCase && requirements.push(Okta.loc('password.complexity.lowercase', 'login'));
+        complexity.minUpperCase && requirements.push(Okta.loc('password.complexity.uppercase', 'login'));
+        complexity.minNumber && requirements.push(Okta.loc('password.complexity.number', 'login'));
+        complexity.minSymbol && requirements.push(Okta.loc('password.complexity.symbol', 'login'));
+        complexity.excludeUsername && requirements.push(Okta.loc('password.complexity.no_username', 'login'));
+
+        if(requirements.length) {
```
rchild-okta: nit: space between if and (

test/helpers/xhr/PASSWORD_RESET_withMaxComplexity.js
```diff
@@ -0,0 +1,50 @@
...
+  "responseType": "json",
+  "response": {
+    "stateToken": "testStateToken",
+    "expiresAt": "2015-07-17T02:33:34.874Z",
+    "status": "PASSWORD_RESET",
+    "_embedded": {
+      "policy": {
+        "complexity": {
+          "minLength": 8,
+          "minLowerCase": 1,
```
rchild-okta: Just checking - is it true that this will always be one? Currently, it looks like the code only expects one for minLowerCase, upperCase, etc.

\+ @srinivasanagandla-okta

src/PasswordResetController.js
```diff
@@ -45,6 +45,34 @@ function (Okta, FormController, FormType, ValidationUtil, FooterSignout, TextBox
...
+          return;
+        }
+
+        var complexity = policy.complexity;
+        var requirements = [];
+
+        complexity.minLength &&
+          requirements.push(Okta.loc('password.complexity.length', 'login', [complexity.minLength]));
+
+        complexity.minLowerCase && requirements.push(Okta.loc('password.complexity.lowercase', 'login'));
```
rchild-okta: There's a lot of duplicated logic here (plus in general I'm not a fan of using && when we really should be using an if statement).

Can we set up a map instead and iterate over the complexity requirements? Something like:
```javascript
var foo = {
  'minLength': 'password.complexity.length',
  // etc
};

var requirements = _.map(policy.complexity, function (val, type) {
  return Okta.loc(foo[type], 'login', [val]);
});

```

test/helpers/xhr/PASSWORD_RESET_withMaxComplexity.js
alexeystadnikov-okta: I had checked with the java code and there is the same logic.

src/PasswordResetController.js
alexeystadnikov-okta: Done.

src/PasswordResetController.js
```diff
@@ -45,6 +45,41 @@ function (Okta, FormController, FormType, ValidationUtil, FooterSignout, TextBox
...
+        var requirements = [];
+        var complexity = policy.complexity;
+
+        _.each(fields, function (params, complexityType) {
+          if (complexity[complexityType]) {
+            params.args ?
+              requirements.push(Okta.loc(params.i18n, 'login', [complexity[complexityType]])) :
+              requirements.push(Okta.loc(params.i18n, 'login'));
+          }
+        });
```
rchild-okta: We can use a map on the complexity object instead of iterating over fields:
```javascript
var requirements = _.map(complexity, function (val, complexityType) {
  var field = fields[complexityType];
  return field.args ? Okta.loc(field.i18n, 'login', [val]) : Okta.loc(field.i18n, 'login');
}
```

rchild-okta:
Overall looks good to me - just make the quick change to use the map, and then let's 🚢

alexeystadnikov-okta:
Done.

PR #96, opened 2016-08-17
U2F Support

mauriciocastillosilva-okta:
- Support for U2F enrollment and challenge flows
- Added the correct Windows Hello icons
- Added tests for enroll and verify

Resolves: OKTA-97018

u2f-signinwidget


rchild-okta:
Looks like the shrinkwrap is wrong - it should never have the resolved fields (which point to internal urls) - this is at least one of the reasons travis is failing.

package.json
```diff
@@ -60,20 +60,21 @@
...
     "handlebars": "4.0.5",
     "jquery": "1.12.1",
     "jquery.cookie": "1.4.1",
-    "backbone": "1.2.1",
+    "q": "1.4.1",
     "request-promise": "2.0.0",
-    "underscore": "1.8.3",
-    "@okta/okta-auth-js": "1.3.0",
-    "q": "1.4.1"
+    "u2f-api-polyfill": "0.4.1",
```
rchild-okta: Just verifying - did you double check with Kim about this license (BSD-3)?

src/EnrollU2FController.js
```diff
@@ -0,0 +1,137 @@
...
+],
+function (Okta, FormType, FormController, Footer, Q) {
+
+  var _ = Okta._;
+
+  return FormController.extend({
+    className: 'enroll-u2f',
+    Model: {
+      local: {
+        '__enrolled__': 'boolean'
```
rchild-okta: Do we actually need this flag? Why not just use the existing isMFAEnrollActivate?

src/EnrollU2FController.js
```diff
@@ -0,0 +1,137 @@
...
+      },
+
+      activate: function () {
+        this.set('__enrolled__', true);
+        this.trigger('errors:clear');
+
+        return this.doTransaction(function (transaction) {
+          var activation = transaction.factor.activation;
+          var appId = activation.appId;
+          var request = {
```
rchild-okta: What is this variable for? Seems like we're not using it (except to construct registerRequests, but at that point why not just do that here?

src/EnrollU2FController.js
```diff
@@ -0,0 +1,137 @@
...
+        FormType.View({
+          View: '<div class="u2f-instructions"><ol>\
+            <li>' + Okta.loc('enroll.u2f.general1', 'login') + '</li>\
+            <li>' + Okta.loc('enroll.u2f.general2', 'login') + '</li>\
+            <li>' + Okta.loc('enroll.u2f.general3', 'login') + '</li>\
+            </ol></div>'
+        })
+      ],
+      _startEnrollment: function () {
+        this.$('.u2f-instructions').html('<p>' + Okta.loc('enroll.u2f.instructions', 'login') + '</p>\
```
rchild-okta: Why are we doing it this way as opposed to using this.add(template)?

src/EnrollU2FController.js
```diff
@@ -0,0 +1,137 @@
...
+            <div class="u2f-usb"></div>\
+            <div class="u2f-bluetooth"></div>\
+          </div>\
+          <div data-se="u2f-waiting" class="okta-waiting-spinner"></div>'
+        );
+        this.$('.o-form-button-bar').hide();
+      },
+
+      _stopEnrollment: function () {
+        this.$('.u2f-instructions').html('<ol>\
```
rchild-okta: Especially in this case - we're duplicating this logic.

src/EnrollU2FController.js
```diff
@@ -0,0 +1,137 @@
...
+    },
+
+    Form: {
+      title: Okta.loc('enroll.u2f.title', 'login'),
+      save: Okta.loc('enroll.u2f.save', 'login'),
+      noCancelButton: true,
+      hasSavingState: false,
+      autoSave: true,
+      className: 'enroll-u2f-form',
+      modelEvents: function () {
```
rchild-okta: nit: doesn't need to be a function

src/VerifyU2FController.js
```diff
@@ -0,0 +1,109 @@
...
+    },
+
+    Form: {
+      autoSave: true,
+      hasSavingState: false,
+      title: Okta.loc('factor.u2f', 'login'),
+      className: 'verify-u2f-form',
+      noCancelButton: true,
+      save: Okta.loc('verify.u2f.retry', 'login'),
+      modelEvents: function () {
```
rchild-okta: Doesn't need to be a function

src/VerifyU2FController.js
```diff
@@ -0,0 +1,109 @@
...
+        return {
+          'error': '_showRetry',
+          'request': '_hideRetry'
+        };
+      },
+
+      formChildren: function () {
+        var result = [];
```
rchild-okta: Hmm, why are we doing it this way? Couple points:
- Only use function if what you're constructing is instance dependent (same reason why we don't need a function for modelEvents in the cases I mentioned before)
- No need to have result = [], result.push if we're just adding one, static item

Can probably reduce this down to:
```javascript
formChildren: [
  FormType.View({
    View: 'template'
  })
]
```

src/VerifyU2FController.js
```diff
@@ -0,0 +1,109 @@
...
+        result.push(
+          FormType.View({
+            View: '<p>' + Okta.loc('verify.u2f.instructions', 'login') + '</p>\
```
rchild-okta: I also noticed that other views use the i18n handlebars tag - are they broken or do we have access to that here (didn't see the wrapper)? I.e. https://github.com/okta/okta-signin-widget/blob/03b22ef5c2058255707bffb1dff5fdd6d260bb46/src/views/mfa-verify/inputs/ShowPasswordBox.js#L24

src/VerifyU2FController.js
```diff
@@ -0,0 +1,109 @@
...
+
+        return this.doTransaction(function (transaction) {
+          var factor = _.findWhere(transaction.factors, {
+            factorType: 'u2f',
+            provider: 'FIDO'
+          });
+          var self = this;
+          return factor.verify()
+          .then(function (transaction) {
+            var factorData = transaction.factor;
```
rchild-okta: nit: factor instead of factorData (I think we use that elsewhere)

src/VerifyU2FController.js
```diff
@@ -0,0 +1,109 @@
...
+        );
+        return result;
+      },
+
+      postRender: function () {
+        this.model.save();
+      },
+
+      _showRetry: function () {
+        this.$('.okta-waiting-spinner').hide();
```
rchild-okta: Can we control this through a class intead? I.e. add the class here and remove it in the other one, and then have the class deal with what is showing.

src/util/FactorUtil.js
```diff
@@ -86,6 +86,12 @@ define(['okta'], function (Okta) {
       description: Okta.loc('factor.windowsHello.description', 'login'),
       iconClassName: 'mfa-windows-hello',
       sortOrder: 10
+    },
+    'U2F': {
```
rchild-okta: Should probably check with @karlmcguinness-okta or @hassenkaraa-okta to see what the new sort order should be - for example, I don't think Windows Hello or U2F should be below Security Question.

test/spec/EnrollU2F_spec.js
```diff
@@ -0,0 +1,156 @@
...
+            data: {
+              stateToken: 'testStateToken',
+              factorType: 'u2f',
+              provider: 'FIDO'
+            }
+          });
+        });
+      });
+
+      itp('calls u2f.register and activates the factor', function () {
```
rchild-okta: Add tests for error states

test/spec/MfaVerify_spec.js
```diff
@@ -1808,6 +1815,59 @@ function (Q,
...
+          });
+        });
+
+        itp('shows the right title', function () {
+          return setupU2F().then(function (test) {
+            expectTitleToBe(test, 'Security Key (U2F)');
+          });
+        });
+
+        itp('shows a spinner while waiting for u2f challenge', function () {
```
rchild-okta: Do we have a test that verifies the spinner goes away when we get the response? Same goes for the enroll flow.

test/spec/EnrollU2F_spec.js
```diff
@@ -0,0 +1,156 @@
...
+          beacon: new Beacon($sandbox),
+          form: new Form($sandbox),
+          ac: authClient,
+          setNextResponse: setNextResponse
+        };
+      });
+    }
+
+    function mocku2fEnroll() {
+      window.u2f = {
```
rchild-okta: Is there a reason we're not doing this the usual way?
```
spyOn(window.u2f, 'register').andCallFake(function () {
});
```

test/spec/MfaVerify_spec.js
```diff
@@ -1808,6 +1815,59 @@ function (Q,
...
+          return setupU2F().then(function (test) {
+            expect(test.form.el('u2f-waiting').length).toBe(1);
+          });
+        });
+
+        itp('calls u2f.sign and verifies factor', function () {
+          window.u2f = {
```
rchild-okta: Same here - let's use jasmine to spy on the methods vs. overwriting them.

test/spec/MfaVerify_spec.js
```diff
@@ -1808,6 +1815,59 @@ function (Q,
...
+          .then(function () {
+            expect($.ajax.calls.count()).toBe(3);
+            Expect.isJsonPost($.ajax.calls.argsFor(2), {
+              url: 'https://foo.com/api/v1/authn/factors/u2fFactorId/verify',
+              data: {
+                clientData: 'someClientData',
+                signatureData: 'someSignature',
+                stateToken: 'testStateToken'
+              }
+            });
```
rchild-okta: Same comment here - what happens in error cases?

package.json
mauriciocastillosilva-okta: Just did. All is good 👍

src/EnrollU2FController.js
mauriciocastillosilva-okta: we need it in case of retries, so we skip straight to reading the u2f key

test/spec/EnrollU2F_spec.js
mauriciocastillosilva-okta: I tried that, but window.u2f is undefined. I tried including the 'u2f-api-polyfill' in this test, but that still didn't work. So I stayed with the overriding approach.

test/spec/EnrollU2F_spec.js
mauriciocastillosilva-okta: added

rchild-okta:
lgtm 💥 !

PR #97, opened 2016-08-22
Publish i18n labels with npm package

rchild-okta:
i18n .properties and .json files are now published to dist/labels

Extra:
- Removed grunt package task, which is no longer being used
- Removed i18n-to-target, which is no longer being used after moving to webpack
- Fixed small casing issue with one of the test xhr files

RESOLVES OKTA-99255

### DIFF
--- a/test/spec/EnrollU2F_spec.js
+++ b/test/spec/EnrollU2F_spec.js
@@ -10,7 +10,7 @@ define([
   'sandbox',
   'LoginRouter',
   'helpers/xhr/MFA_ENROLL_allFactors',
-  'helpers/xhr/MFA_ENROLL_ACTIVATE_U2F'
+  'helpers/xhr/MFA_ENROLL_ACTIVATE_u2f'
 ],
 function (Okta,
           $,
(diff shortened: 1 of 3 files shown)


test/spec/EnrollU2F_spec.js
```diff
@@ -10,7 +10,7 @@ define([
   'sandbox',
   'LoginRouter',
   'helpers/xhr/MFA_ENROLL_allFactors',
-  'helpers/xhr/MFA_ENROLL_ACTIVATE_U2F'
+  'helpers/xhr/MFA_ENROLL_ACTIVATE_u2f'
```
ukilon-okta: how did it ever work?

ukilon-okta:
one nit, otherwise 🎉

test/spec/EnrollU2F_spec.js
rchild-okta: Case sensitivity in mac vs linux and it not actually failing the test suite when files can't be loaded.

PR #101, opened 2016-08-25
Fixes issue with iOS Safari and autofill

rchild-okta:
No longer re-renders primary auth form on changes to the rememberMe checkbox,
which caused iOS Safari autofill to repopulate the username and prevent the
user from changing their username.

RESOLVES OKTA-98946


lboyette-okta:
lgtm 🚢

PR #102, opened 2016-08-26
Upgrade okta-auth-js to 1.4.0 and add tokenManager

lboyette-okta:
Resolves: OKTA-99491

@rchild-okta

### DIFF
--- a/node_modules/@okta/courage/package.json
+++ b/node_modules/@okta/courage/package.json
@@ -10,7 +10,7 @@
         "spec": "1.3.0-beta.30",
         "type": "version"
       },
-      "/Users/mauricio.castillosilva/okta/signin-widget"
+      "/Users/lboyette/okta/okta-signin-widget"
     ]
   ],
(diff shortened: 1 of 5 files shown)


rchild-okta:
Couple questions/observations:
- Can you validate the flow where we ask for the access token as well? On first pass it looks like this won't work for the "use okta as the idp in the social auth flow" because of [this check](https://github.com/okta/okta-signin-widget/blob/master/src/models/Settings.js#L87) 
- Can we replace the deprecated [idToken.authorize](https://github.com/okta/okta-signin-widget/blob/master/src/util/OAuth2Util.js#L40) with some of our new token methods?
- Looks like our shrinkwrap file is wrong (includes internal urls). I walked through the process with Mau a couple days ago, let's go over this tomorrow.

rchild-okta:
Note: Earlier comment will be addressed in the next PR.

rchild-okta:
lgtm 🚀

PR #103, opened 2016-08-26
OKTA-99168 New Signin Flow does not provide adequate visual indicatio…

eugenebakaiev-okta:
OKTA-99168 New Signin Flow does not provide adequate visual indication - violation of Section 508 and WCAG 2

Added css class for focused text input
and function to toggle that class


eugenebakaiev-okta:
@rchild-okta, please take a look

PR #106, opened 2016-09-01
Add basic protractor test

lboyette-okta:
Resolves: OKTA-99393

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "lint:report": "grunt lint --checkstyle",
     "lint": "grunt lint",
     "test": "grunt test",
-    "test:protractor": "grunt test-protractor",
+    "test:protractor": "webdriver-manager update && grunt test-protractor",
     "protractor": "protractor target/protractor/conf.js",
     "build:dev": "grunt build",
     "build:prod": "grunt build:minified",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -30,7 +32,7 @@
     "build:test": "webpack --config webpack.test.config.js",
     "prestart": "npm run build:dev",
     "start": "grunt start-server-open",
-    "prepublish": "grunt prep-release",
+    "prepublish": "grunt prep-release && webdriver-manager update",
```
rchild-okta: Why is webdriver-manager update in the prepublish step?

lboyette-okta: It's how I guarantee we have the selenium server jar iff we `npm install` from the repo.

https://docs.npmjs.com/misc/scripts
> prepublish: Run BEFORE the package is published. (Also run on local npm install without any arguments.)

test/protractor/specs/spec.js
```diff
@@ -0,0 +1,20 @@
+describe('Okta Sign-In Widget', function() {
+  var username = element(by.name('username'));
```
rchild-okta: Hmmm, does this work, and will it work across multiple tests? If it's a protractor convention, it might be okay, but it feels confusing to write this like this - i.e. where we're initializing our username object before we've navigated to the page.

It'd be clearer to do something like:
```javascript
// Use a function where it's clear that we're invoking it at the right time
function getUserName() {
  return element(by.name('username'));
}
```

rchild-okta: Ah, I see (reading the protractor docs) - looks like that's there suggestion.

package.json
rchild-okta: How about in postinstall? Or before we actually run our tests (since that's the only time we care)?

lboyette-okta: I guess it takes less than a second if the file is already downloaded, so I'll put it right before we run the tests

rchild-okta:
lgtm 🚀 !

PR #107, opened 2016-09-01
Added support for scrolling in embedded browsers flow

ukilon-okta:
- Set overflow to auto on the document body
- Reduced the padding above of the footer in order to avoid scroll bar on the primary auth flow

Before:
screen shot 2016-09-01 at 11 51 21 am

After:
screen shot 2016-09-01 at 12 46 28 pm

screen shot 2016-09-01 at 12 24 34 pm


rchild-okta:
Looks good to me - I'll merge it in once it's passed through Travis 🚀

PR #110, opened 2016-09-12
Added wait functions for easier unit testing, and fixed tests

rchild-okta:
- Added support for wait functions in the unit tests (waitForController, etc).
  This makes it easier to validate transitions between the different async
  states.
- Now removes controllers after the test has finished (better test isolation)
- Better exposure to errors. Most errors are no longer swallowed in the
  various promise chains.
- Changed global error handling - only throws when there is no global error
  callback defined

Fixes to unit tests that were either causing flakiness or errors in the
console:

- Fixed issues with images that were not loading from the file system
- Better cleanup after the test is done - check for errors after we've waited
  a tick for the test to finish.
- Remove qtip hovers after each test
- Removed unnecessary Q.stopUnhandledRejectionTracking calls
- General cleanup of tests to remove unnecessary ticks/thens/etc, and fixed
  several problems in the recently added tests

RESOLVES OKTA-101038


austinstanhope-okta:
Looks ok to me

test/unit/helpers/xhr/MFA_ENROLL_totp_success.js
```diff
@@ -32,7 +32,7 @@ define({
             "factorResult": "WAITING",
             "_links": {
               "qrcode": {
-                "href": "../../../test/assets/1x1.gif",
+                "href": "../../../test/unit/assets/1x1.gif",
```
lboyette-okta: nit: We use `1x1.gif` here, but the file is actually `1x1.GIF`. The new file (`1x1v2.gif`) uses lowercase, so maybe we should change the name of `1x1.GIF` to lowercase while we're at it.

test/unit/spec/EnrollOnPrem_spec.js
```diff
@@ -123,7 +121,6 @@ function (Q, _, $, OktaAuth, Util, Form, Beacon, Expect, $sandbox,
         itp('shows error in case of an error response', function () {
           return setup()
           .then(function (test) {
-            Q.stopUnhandledRejectionTracking();
```
lboyette-okta: So glad we can rid of these.

test/unit/spec/EnrollOnPrem_spec.js
```diff
@@ -166,11 +163,9 @@ function (Q, _, $, OktaAuth, Util, Form, Beacon, Expect, $sandbox,
 
       Expect.describe('Header & Footer', function () {
         itp('displays the correct factorBeacon', function () {
-          $.fx.off = true;
```
lboyette-okta: And these!

test/unit/spec/EnrollSymantecVip_spec.js
```diff
@@ -32,35 +32,32 @@ function (Q, _, $, OktaAuth, Util, Form, Beacon, Expect, $sandbox,
...
-      .then(function () {
-        return {
+        return Expect.waitForEnrollSymantecVip({
           router: router,
           beacon: new Beacon($sandbox),
           form: new Form($sandbox),
           ac: authClient,
           setNextResponse: setNextResponse
-        };
+
```
lboyette-okta: nit: extra space

test/unit/spec/EnrollTotpController_spec.js
```diff
@@ -96,32 +93,45 @@ function (_, $, Q, OktaAuth, LoginUtil, StringUtil, Util, DeviceTypeForm, Barcod
       test.setNextResponse(res || resTotpEnrollSuccess);
       test.form.selectDeviceType('APPLE');
       test.form.submit();
-      return tick(test);
```
lboyette-okta: hmm. We don't need this tick anymore?

test/unit/helpers/xhr/MFA_ENROLL_totp_success.js
rchild-okta: As discussed, might do this later depending on what else I need to change (casing issues on osx, etc)

test/unit/spec/EnrollSymantecVip_spec.js
rchild-okta: Will fix

lboyette-okta:
lgtm 🚢

PR #112, opened 2016-09-15
Support for Voice Call in forgot password flow

mauriciocastillosilva-okta:
As a side note, the layout for the forgot password and for the unlock account forms changed from 2 buttons in the same row, to one button per row. This also helps fixing long sentences for such buttons in other languages

Forgot Password - 2 factors:
forgotpassword - 2 factors

Forgot Password  - 3 factors:
forgotpassword - all factors

Forgot Password - voice call
forgotpassword - voice call


test/unit/spec/ForgotPassword_spec.js
```diff
@@ -350,6 +386,85 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
...
+          expect(test.form.usernameErrorField().length).toBe(1);
+        });
+      });
+      itp('makes a Voice Call', function () {
+        return setupWithCall().then(function (test) {
+          $.ajax.calls.reset();
+          test.setNextResponse(resChallengeCall);
+          test.form.setUsername('foo');
+          test.form.makeCall();
+          return tick();
```
rchild-okta: I added support for wait functions so we ~mostly~ no longer need to rely on ticks (there are a couple edge cases). Aside from being more stable, it allows us to be explicit about which pages we're waiting for. In this case, we should replace "tick()" with "Expect.waitForRecoveryChallenge()".

There's also examples of this in this file, i.e:
https://github.com/mauriciocastillosilva-okta/okta-signin-widget/blob/884472c3748235490377b1d8b7ba51ed28f0e693/test/unit/spec/ForgotPassword_spec.js#L577

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -350,6 +386,85 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
...
+          });
+        });
+      });
+      itp('updates appState username after making a Voice Call', function () {
+        return setupWithCall()
+        .then(function (test) {
+          test.setNextResponse(resChallengeCall);
+          test.form.setUsername('foo');
+          test.form.makeCall();
+          return tick(test);
```
rchild-okta: Same here

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -350,6 +386,85 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
...
+        })
+        .then(function (test) {
+          expect(test.router.appState.get('username')).toBe('foo');
+        });
+      });
+      itp('shows an error if making a Voice Call results in an error', function () {
+        return setupWithCall()
+        .then(function (test) {
+          Q.stopUnhandledRejectionTracking();
```
rchild-okta: If you remove this line and re-run the test, do you see an error in the test results/console? 

I noticed when I was going through the tests that we no longer need most of these Q.stopUnhandledRejectionTracking lines - at some point we started handling the errors better. One extra benefit is that the tests do run faster when you remove unnecessary calls to this function.

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -350,6 +386,85 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
...
+        })
+        .then(function (test) {
+          expect(test.form.hasErrors()).toBe(true);
+          expect(test.form.errorMessage()).toBe('You do not have permission to perform the requested action');
+        });
+      });
+      itp('does not have a problem with sending email after making a Voice Call', function () {
+        return setupWithCall()
+        .then(function (test) {
+          Q.stopUnhandledRejectionTracking();
```
rchild-okta: Same here - remove and check if there are errors in the test results or the console. FYI, there should be no errors/logs in the console when you do a full run (as of my last commit I cleaned up all the issues).

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -350,6 +386,85 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
...
+        });
+      });
+      itp('shows an error if making a Voice Call results in an error', function () {
+        return setupWithCall()
+        .then(function (test) {
+          Q.stopUnhandledRejectionTracking();
+          test.setNextResponse(resError);
+          test.form.setUsername('foo');
+          test.form.makeCall();
+          return tick(test);
```
rchild-okta: Same here, replace with Expect.waitForFormError

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -350,6 +386,85 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
...
+        });
+      });
+      itp('does not have a problem with sending email after making a Voice Call', function () {
+        return setupWithCall()
+        .then(function (test) {
+          Q.stopUnhandledRejectionTracking();
+          test.setNextResponse(resError);
+          test.form.setUsername('foo');
+          test.form.makeCall();
+          return tick(test);
```
rchild-okta: Expect.waitForFormError

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -350,6 +386,85 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
...
+        })
+        .then(function (test) {
+          $.ajax.calls.reset();
+          test.setNextResponse(resSuccess);
+          test.form.submit();
+          return tick();
```
rchild-okta: Expect.waitForSpyCall, i.e:
https://github.com/okta/okta-signin-widget/blob/8052fd12484cacb2a0a166f89d0be6e560b90ae7/test/unit/spec/EnrollWindowsHello_spec.js#L151

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -528,6 +643,80 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
           expect(test.form.errorMessage()).toBe('You do not have permission to perform the requested action');
         });
       });
+      itp('shows the "Reset via email" link after making a Voice Call', function () {
+        return setupWithCall().then(function (test) {
+          test.setNextResponse(resChallengeCall);
+          test.form.setUsername('foo');
+          test.form.makeCall();
+          return tick(test);
```
rchild-okta: Expect.waitForRecoveryChallenge

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -528,6 +643,80 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
...
+          expect(test.form.hasSendEmailLink()).toBe(true);
+          expect(test.form.sendEmailLink().trimmedText()).toEqual('Didn\'t receive a code? Reset via email');
+        });
+      });
+      itp('sends an email when user clicks the "Reset via email" link, after making a Voice Call', function () {
+        return setupWithCall().then(function (test) {
+          test.setNextResponse(resChallengeCall);
+          test.form.setUsername('foo@bar');
+          test.form.makeCall();
+          return tick(test);
```
rchild-okta: Expect.waitForRecoveryChallenge

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -528,6 +643,80 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
...
+        })
+        .then(function (test) {
+          $.ajax.calls.reset();
+          test.setNextResponse(resChallengeEmail);
+          test.form.clickSendEmailLink();
+          return tick(test);
```
rchild-okta: Expect.waitForPwdResetEmailSent

(and for the ones below)

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -528,6 +643,80 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
...
+        .then(function (test) {
+          expect(test.form.titleText()).toBe('Email sent!');
+          expect(test.form.getEmailSentConfirmationText().indexOf('foo@bar') >= 0).toBe(true);
+          expect(test.form.backToLoginButton().length).toBe(1);
+        });
+      });
+      itp('shows an error if sending email via "Reset via email" link results in an error, after making a Voice Call',
+        function () {
+        return setupWithCall().then(function (test) {
+          Q.stopUnhandledRejectionTracking();
```
rchild-okta: Same here, verify if we actually need this

rchild-okta: Also, Travis is showing a lint error

mauriciocastillosilva-okta: This one is actually needed (it was the only one), so will leave it there

test/unit/helpers/util/Expect.js
```diff
@@ -62,7 +62,8 @@ define([
         // back on.
         Q.resetUnhandledRejections();
         window.removeEventListener('error', errListener);
-        done();
+        // Wait a tick to make sure the tests clean up
+        fn.tick().then(done);
```
rchild-okta: I have a feeling this is a symptom of tests that aren't waiting enough - I had added the "waitForFormError" late in my last PR, and didn't go back to convert all error test ticks.

Can you remove this edit, and In primaryAuth_spec can you convert the error ticks to waitForFormError? I.e. like here:
https://github.com/okta/okta-signin-widget/blob/master/test/unit/spec/PrimaryAuth_spec.js#L1182

I'm hoping that will make these tests no longer flaky. If not, let's talk and we can think about adding another tick. However, we shouldn't add it here - it should go to line 54 (my previous PR did not remove the tick, it moved it before this function). The reason to do the wait before this function is because we were previously ignoring errors (i.e. wait, and then check to see if we have any exceptions, etc).

mauriciocastillosilva-okta: Sure, let me give that a try.

PR #114, opened 2016-09-19
changed paddings and some elements hidden on small screen height

eugenebakaiev-okta:
changed paddings and some elements hidden on small screen height
RESOLVES OKTA-98600

### DIFF
--- a/src/util/BaseLoginRouter.js
+++ b/src/util/BaseLoginRouter.js
@@ -23,6 +23,7 @@ define([
   'models/Settings',
   'views/shared/Header',
   'views/shared/SecurityBeacon',
+  'views/shared/AuthContainer',
   'models/AppState',
   './RouterUtil',
   './Animations',
@@ -30,7 +31,7 @@ define([
   'util/Bundles'
 ],
 function (Okta, Backbone, BrowserFeatures, XDomain, RefreshAuthStateController, Settings, Header,
-          SecurityBeacon, AppState, RouterUtil, Animations, Errors, Bundles) {
+          SecurityBeacon, AuthContainer, AppState, RouterUtil, Animations, Errors, Bundles) {
 
   var _ = Okta._,
       $ = Okta.$;
@@ -109,13 +110,15 @@ function (Okta, Backbone, BrowserFeatures, XDomain, RefreshAuthStateController, 
         // and then the open tooltip will lose focus and close.
       });
 
-      Okta.$(options.el).append('<div id="okta-sign-in" class="auth-container main-container"></div>');
-      this.el = '#okta-sign-in';
-
       this.appState = new AppState({
         baseUrl: this.settings.get('baseUrl'),
         settings: this.settings
       }, { parse: true });
+
+      var wrapper = new AuthContainer({appState: this.appState});
+      Okta.$(options.el).append(wrapper.render().$el);
+      this.el = '#okta-sign-in';
+
       this.header = new Header({
         el: this.el,
         appState: this.appState,
--- a/src/views/shared/AuthContainer.js
+++ b/src/views/shared/AuthContainer.js
@@ -1,0 +1,28 @@
+/*!
+ * Copyright (c) 2015-2016, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define([
+  'okta'
+],
+function (Okta) {
+  var CAN_REMOVE_BEACON_CLS = 'can-remove-beacon';
+  return Okta.View.extend({
+    className: 'auth-container main-container',
+    id: 'okta-sign-in',
+    attributes: { 'data-se': 'auth-container' },
+    initialize: function (options) {
+      this.listenTo(options.appState, 'change:beaconType', function (model, type) {
+        this.$el.toggleClass(CAN_REMOVE_BEACON_CLS, type === 'security');
+      });
+    }
+  });
+});
--- a/src/views/shared/SecurityBeacon.js
+++ b/src/views/shared/SecurityBeacon.js
@@ -133,6 +133,7 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
       this.listenTo(options.appState, 'loading', function (isLoading) {
         this.$el.toggleClass('beacon-loading', isLoading);
       });
+      this.options.appState.set('beaconType', 'security');
     },
 
     postRender: function () {
--- a/test/unit/helpers/dom/AuthContainer.js
+++ b/test/unit/helpers/dom/AuthContainer.js
@@ -1,0 +1,15 @@
+define(['jquery', './Dom'], function ($, Dom) {
+
+  return Dom.extend({
+
+    authContainer: function () {
+      return this.el('auth-container');
+    },
+
+    canBeMinimized: function () {
+      return this.authContainer().hasClass('can-remove-beacon');
+    }
+
+  });
+
+});
--- a/assets/sass/_layout.scss
+++ b/assets/sass/_layout.scss
@@ -19,7 +19,7 @@
   border-style: solid;
   border-width: 1px;
   height: auto;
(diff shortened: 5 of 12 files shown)


eugenebakaiev-okta:
@rchild-okta, @mauriciocastillosilva-okta could you look please?

rchild-okta:
General:
- If you're making a change to any repo that has a visual component (like this one), add a screenshot/screencast
- Commit message etiquette: https://oktawiki.atlassian.net/wiki/display/eng/Commit+Messages
- The travis build didn't pass. This looks similar to an issue that @mauriciocastillosilva-okta had, I think he can help you with that.

eugenebakaiev-okta:
screen_breaks mov

eugenebakaiev-okta:
screen shot 2016-09-20 at 4 26 58 pm

eugenebakaiev-okta:
@nbories-okta please take a look

rchild-okta:
Can you attach a screencast that includes MFA (i.e. okta verify, etc)? My main concern - does this only hide the security beacon, or does it hide the MFA beacon as well?

eugenebakaiev-okta:
mfascreen mov

nbories-okta:
@eugenebakaiev-okta this should only happen when user is on the login screen. The beacon should not disappear when user is on the MFA section.

eugenebakaiev-okta:
factorbeacon mov

eugenebakaiev-okta:
@rchild-okta, @nbories-okta look again, please

src/util/BaseLoginRouter.js
```diff
@@ -229,7 +233,7 @@ function (Okta, Backbone, BrowserFeatures, xdomain, RefreshAuthStateController,
           Beacon = null;
         }
         this.header.setBeacon(Beacon, controllerOptions);
-
+        this.header.setIsSecurityBeaconClass(isSecurityBeacon(Beacon));
```
rchild-okta: We shouldn't add page specific logic to the BaseLoginRouter (or generic classes in general).

Related to the other comment I left below:
- The view that knows its a security beacon is SecurityBeacon.
- It should not update dom elements outside of itself, so it will need to a) pass a message b) set appState to communicate this information to a higher level view.
- The higher level view listens for this message/change in state and then sets the class.

eugenebakaiev-okta:
@nbories-okta is it ok?

eugenebakaiev-okta:
screen shot 2016-09-23 at 10 59 10 pm
screen shot 2016-09-23 at 10 59 23 pm
screen shot 2016-09-23 at 10 59 34 pm
screen shot 2016-09-23 at 10 59 46 pm
screen shot 2016-09-23 at 10 59 55 pm

nbories-okta:
Looks good visually! Thanks @eugenebakaiev-okta

eugenebakaiev-okta:
@rchild-okta, please look on the implementation, if everything will be ok I'll start to write tests

src/util/BaseLoginRouter.js
```diff
@@ -23,14 +23,15 @@ define([
   'models/Settings',
   'views/shared/Header',
   'views/shared/SecurityBeacon',
+  'views/AuthContainer',
```
rchild-okta: nit: let's add this to views/shared

src/util/BaseLoginRouter.js
```diff
@@ -109,13 +110,15 @@ function (Okta, Backbone, BrowserFeatures, xdomain, RefreshAuthStateController,
...
 
-      Okta.$(options.el).append('<div id="okta-sign-in" class="auth-container main-container"></div>');
-      this.el = '#okta-sign-in';
-
       this.appState = new AppState({
         baseUrl: this.settings.get('baseUrl'),
         settings: this.settings
       }, { parse: true });
+
+      this.wrapper = new AuthContainer({appState: this.appState});
```
rchild-okta: Do we need to save a reference to this? For now, let's just use "var wrapper = ..."

src/views/AuthContainer.js
```diff
@@ -0,0 +1,23 @@
+define([
```
rchild-okta: nit: In new files you need to add the license header to the top (just copy it from another file)

src/views/AuthContainer.js
```diff
@@ -0,0 +1,23 @@
...
+  'okta'
+],
+function (Okta) {
+  var CAN_REMOVE_BEACON_CLS = 'can-remove-beacon';
+  return Okta.View.extend({
+    // template: '<div id="okta-sign-in" class="auth-container main-container"></div>',
```
rchild-okta: nit: remove commented code

src/views/AuthContainer.js
```diff
@@ -0,0 +1,23 @@
...
+    // el: '#okta-sign-in',
+    tagName: 'div',
```
rchild-okta: tagName is 'div' by default in backbone, so we can remove this line

eugenebakaiev-okta:
@stsai-okta could you please look? Tests are on the way

eugenebakaiev-okta:
Tests added.
@rchild-okta, @stsai-okta please look

src/views/shared/AuthContainer.js
```diff
@@ -0,0 +1,33 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define([
+  'okta'
+],
+function (Okta) {
+  var CAN_REMOVE_BEACON_CLS = 'can-remove-beacon';
```
stsai-okta: why is the variable name all caps?

src/views/shared/AuthContainer.js
```diff
@@ -0,0 +1,33 @@
...
+  return Okta.View.extend({
+    className: 'auth-container main-container',
+    id: 'okta-sign-in',
+    attributes: { 'data-se': 'auth-container' },
```
stsai-okta: i don't think we need a data-se here right now since this isn't being used in a selenium test

src/views/shared/AuthContainer.js
```diff
@@ -0,0 +1,33 @@
...
+    initialize: function (options) {
+      this.listenTo(options.appState, 'change:beaconType', function (model, type) {
+        if(type === 'security') {
```
stsai-okta: could we just use toggle here instead?

src/views/shared/AuthContainer.js
eugenebakaiev-okta: because it's something like a constant string

src/views/shared/AuthContainer.js
eugenebakaiev-okta: but our unit tests also use the data-se attr to find elements

src/views/shared/AuthContainer.js
eugenebakaiev-okta: no, some views doesn't have a beacon at all, neither security nor factor

rchild-okta: I'm guessing @stsai-okta meant toggle with the [state option](http://api.jquery.com/toggleclass/#toggleClass-state), i.e:
```javascript
this.$el.toggleClass(CAN_REMOVE_BEACON_CLS, type === 'security');
```

I'm okay with either.

src/views/shared/Header.js
```diff
@@ -170,7 +170,7 @@ function (Okta, Animations, LoadingBeacon) {
           throw new Error('the "' + transition + '" is not recognized');
       }
     },
-
+
```
rchild-okta: nit: remove this

src/views/shared/SecurityBeacon.js
```diff
@@ -133,6 +133,7 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
       this.listenTo(options.appState, 'loading', function (isLoading) {
         this.$el.toggleClass('beacon-loading', isLoading);
       });
+      this.options.appState.set('beaconType', 'security');
```
rchild-okta: More as a note to myself: I just noticed that the Beacons don't share a common view class - we should fix that in the future.

test/unit/helpers/dom/AuthContainer.js
```diff
@@ -0,0 +1,15 @@
+define(['jquery', './Dom'], function ($, Dom) {
+
+  return Dom.extend({
+
+    authContainer: function () {
+      return this.el('auth-container');
+    },
+
+    hasClass: function (className) {
```
rchild-okta: nit: I think it would be nicer to have this test class contain the actual logic, i.e. instead of "hasClass", provide "canBeMinimized" which would return a boolean. This way the test itself doesn't need to know about the details of how this is set (i.e. some class being added).

eugenebakaiev-okta:
@rchild-okta, @stsai-okta is it good now?

stsai-okta:
works for me 🖌

PR #115, opened 2016-09-19
Added better i18n support

rchild-okta:
- Localization is now exposed inside the widget itself
- "language" is now exposed as a config option
  - Can be a languageCode, i.e. "en", "ja", etc
  - Can be a function that returns a languageCode. The list of supported
    languages and user languages are passed to this hook.
- If no "language" option is passed, defaults to the user's browser settings
- Will pull languages by default from the Okta CDN, but can be overridden
  by a new "assetBaseUrl" option
- Deprecated "labels" and "country" and replaced with a new "text" option
  which also handles overriding on a per language level
- Support for developers adding their own languages through the "text" option
- Generates a new jsonp label dist type for easy CORS consumption

Other:
- Generates config file with version & supported languages
- Some updates to make it easier to run test specs on their own
- New deprecate util function

Changes to get i18n working in various flows

- Because the locale can now be updated dynamically, several code patterns
  needed to be changed. Most commonly - calling our Okta.loc function when
  the module is being evaluated.

RESOLVES OKTA-99259


rchild-okta:
Before this commit, we could override properties like this:
```javascript
new OktaSignIn({
  labels: {
    'enroll.call.setup': 'Some new text for this'
  },
  // Separate way to override country properties
  country: {
    'JP': 'new name for Japan'
  }
});
```

After this commit, these are the new language properties (old way is still supported, but deprecated):
```javascript
new OktaSignIn({

  // The new way to do overrides
  text: {
    'en': {
      'enroll.call.setup': 'Some new text for this',
      'country.JP': 'new name for Japan'
    },
    // Other languages can be specified, with their own overrides.
    // You can also pass in a new language here as well, and we will use that
    'jp': {
      // some props
    }
  } 

  // Specify language using a string:
  language: 'en',

  // Specify language using a function:
  language: function (supportedLanguages, userLanguages) {
    // supportedLanguages is the list we have translated, plus any new language
    // added through "text".
    // userLanguages is the list of languages (in order of preference) that
    // we've detected from the browser
    return 'en';
  },

  // By default, this will point to the Okta CDN. You can override it here if:
  // 1) You want to host the assets yourself
  // 2) You are using a new version of the OSW whose labels have not been 
  //    published yet
  // Note: This is where we try to download new languages from (jsonp)
  assetBaseUrl: '/path/to/our/assets'

});
```

raphaellondner-okta:
It would be great if we could have a "smart" browser language default mechanism, where "fr-ca" would default to "fr" if we don't have "fr-ca" as an available translated language (same thing for variants of "zh", "de" and other cultures)

rchild-okta:
@raphaellondner-okta pushed a commit to support aliasing down to the base locales

src/util/FactorUtil.js
```diff
@@ -16,80 +16,80 @@ define(['okta'], function (Okta) {
 
   var factorData = {
     'OKTA_VERIFY': {
-      label: Okta.loc('factor.totpSoft.oktaVerify', 'login'),
-      description: Okta.loc('factor.totpSoft.description', 'login'),
+      label: 'factor.totpSoft.oktaVerify',
```
lboyette-okta: hmm. Guess we made the decision to explicitly not localize factors?

src/util/FactorUtil.js
```diff
@@ -143,11 +143,13 @@ define(['okta'], function (Okta) {
   };
 
   fn.getFactorLabel = function (provider, factorType) {
-    return factorData[fn.getFactorName(provider, factorType)].label;
+    var key = factorData[fn.getFactorName(provider, factorType)].label;
+    return Okta.loc(key, 'login');
```
lboyette-okta: ah, I see

src/util/OAuth2Util.js
```diff
@@ -40,7 +40,7 @@ define(['okta', './Enums', './Errors'], function (Okta, Enums, Errors) {
     _.extend(
       oauthParams,
       _.pick(options, 'clientId', 'redirectUri'),
-      _.pick(options.authParams, 'responseType', 'responseMode', 'display', 'scope', 'scopes'),
```
lboyette-okta: We haven't released a version that declares `scope` to be deprecated; this would be a breaking change. We should probably keep it until v2

src/views/primary-auth/SocialAuth.js
```diff
@@ -69,7 +64,7 @@ define([
           'data-se': attr
         },
         className: 'social-auth-button ' + attr,
-        title: socialAuthIdpLabels[type],
+        title: Okta.loc('socialauth.' + type + '.label'),
```
lboyette-okta: Should this also return a function?

test/unit/helpers/mocks/Util.js
```diff
@@ -73,11 +73,14 @@ function ($, _, Backbone, Q, Duo) {
...
 
-    spyOn($, 'ajax').and.callFake(function() {
+    spyOn($, 'ajax').and.callFake(function(req) {
 
       var xhr = allResponses.shift();
       if (!xhr) {
-        throw new Error('We are making a request that we have not anticipated.');
+        throw new Error(
+          'We are making a request that we have not anticipated: ' +
+          req.type.toUpperCase() + ' ' + req.url
```
lboyette-okta: Why didn't we do this earlier?! 👍

test/unit/helpers/xhr/labels_country_ja.js
```diff
@@ -0,0 +1,7 @@
+define({
+  'status': 200,
+  'responseType': 'json',
+  'response': {
+    'JP': 'JA: country.JP'
+  }
+});
```
lboyette-okta: nit: newline

test/unit/helpers/xhr/labels_login_ja.js
```diff
@@ -0,0 +1,7 @@
+define({
+  'status': 200,
+  'responseType': 'json',
+  'response': {
+    'enroll.call.setup': 'JA: enroll.call.setup'
+  }
+});
```
lboyette-okta: nit: newline

test/unit/spec/LoginRouter_spec.js
```diff
@@ -627,8 +682,392 @@ function (Okta, Q, Backbone, xdomain, SharedUtil, CryptoUtil, CookieUtil, OktaAu
...
+        .then(function () {
+          expectWithBase('https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/9.9.99');
+        });
+      });
+      itp('loads properties from the given baseUrl', function () {
+        return setupLanguage({
+          mockLanguageRequest: 'ja',
+          settings: {
+            language: 'ja',
+            assetBaseUrl: 'http://foo.com'
```
lboyette-okta: ah, I see. We pass our own assetBaseUrl

test/unit/spec/Util_spec.js
```diff
@@ -0,0 +1,53 @@
...
+      it('lowercases all string entries in a given array', function () {
+        var arr = ['Hi', 'THERE', 'i', 'wOUld'],
+            expected = ['hi', 'there', 'i', 'would'];
+        expect(Util.toLower(arr)).toEqual(expected);
+      });
+    });
+
+  });
+
+});
```
lboyette-okta: nit: newline

webpack.common.config.js
```diff
@@ -66,25 +67,6 @@ module.exports = {
     new webpack.ProvidePlugin({
       process: empty,
       setImmediate: empty
-    }),
-
-    new webpack.DefinePlugin({
-      WIDGET_VERSION: JSON.stringify(packageJson.version)
```
lboyette-okta: 👍

Gruntfile.js
```diff
@@ -88,6 +88,19 @@ module.exports = function (grunt) {
...
         ]
       },
+      'i18n-to-target': {
+        files: [
+          {
+            expand: true,
+            cwd: 'node_modules/@okta/i18n/dist/',
+            src: [
+              'json/{login,country}*.json',
+              'properties/{login,country}*.properties'
```
rchild-okta: Good point - will remove properties from target flow

src/models/Settings.js
```diff
@@ -25,6 +32,10 @@ function (Okta, Errors, BrowserFeatures) {
       ConfigError = Errors.ConfigError,
       UnsupportedBrowserError = Errors.UnsupportedBrowserError;
 
+  var assetBaseUrlTpl = Okta.tpl(
+    'https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/{{version}}'
```
rchild-okta: General idea:
- By default, will be serving off the CDN
- In dev, I made a change to the Gruntfile to point assetBaseUrl to our local files
- Anyone that's running off latest master (or a version whose published assets have not hit production) will need to override the assetBaseUrl. This is what we'll need to do as well in core code since we're running off early builds

Do you foresee any problems with this?

*Note:* I do think it will be good to think about a faster deploy-to-cdn process for published artifacts in the near future (along with this, we still have the problem of syncing dev.okta.com docs)

src/util/FactorUtil.js
rchild-okta: Yep, they are localized - I'm just changing the time that we evaluate the Okta.loc call.

src/util/OAuth2Util.js
rchild-okta: Check out settings.js - I'm now handling all deprecations right when the options come in, and converting them to the supported options. It's nicer because it makes internal code more consistent with the way we want it to look.

src/views/primary-auth/SocialAuth.js
rchild-okta: It's already being evaluated after the view has been initialized - the only unsupported pattern is "Okta.loc is evaluated when the module itself is loaded".

src/models/Settings.js
lboyette-okta: Our customers using i18n features in Preview will need to know this, but there's nothing wrong with the solution. Maybe deploying to the cdn on every release, including trex, is a strategy we should consider (at least for sdks).

src/util/OAuth2Util.js
lboyette-okta: gotcha, makes sense

rchild-okta:
@lboyette-okta, @raphaellondner-okta made a small change to support overriding the asset bundle filePath, which is useful (as in the case of okta) for supporting cachebusting:

Instead of **assetBaseUrl**, it is:
```javascript
{
  assets: {
    baseUrl: '/foo/bar',
    rewrite: function (assetPath) {
      // path is relative to baseUrl
      // Example to load login for 'ja': /labels/jsonp/login_ja.jsonp
      return cachebustify(assetPath);
    }
  }
}
```

rchild-okta:
Talked to @karlmcguinness-okta and he suggested renaming "text" to "i18n", so now overrides will look like:
```javascript
{
  i18n: {
    'en': {
      'enroll.call.setup': 'Some new text for this',
      'country.JP': 'new name for Japan'
    },
    // Other languages can be specified, with their own overrides.
    // You can also pass in a new language here as well, and we will use that
    'jp': {
      // some props
    }
  } 
}
```

PR #116, opened 2016-09-29
Resolves OKTA-98474, OKTA-98474

alexeystadnikov-okta:
Windows Hello Beta Changes UX

Resolves OKTA-98474, OKTA-98474

- Added spinner while Windows Hello is authenticating the user.
- Added errors for NotFoundError if Windows Hello factor was enrolled from another PC or was reconfigured on the current.
- Catch AbortError and don't show any error.
- Start Windows Hello verification immediately if Windows Hello is the only configured factor.

screen shot 2016-09-29 at 7 58 08 pm
screen shot 2016-09-29 at 7 55 56 pm
screen shot 2016-09-29 at 7 56 50 pm

@mauriciocastillosilva-okta 
@rchild-okta 

Please, review.

### DIFF
--- a/src/EnrollWindowsHelloController.js
+++ b/src/EnrollWindowsHelloController.js
@@ -15,10 +15,11 @@ define([
   'util/FormController',
   'util/FormType',
   'util/webauthn',
+  'views/shared/Spinner',
   'views/enroll-factors/Footer',
   'views/mfa-verify/WindowsHelloErrorMessageView'
 ],
-function (Okta, FormController, FormType, webauthn, Footer, WindowsHelloErrorMessageView) {
+function (Okta, FormController, FormType, webauthn, Spinner, Footer, WindowsHelloErrorMessageView) {
 
   var _ = Okta._;
 
@@ -60,6 +61,7 @@ function (Okta, FormController, FormType, webauthn, Footer, WindowsHelloErrorMes
         return this.doTransaction(function (transaction) {
           var activation = transaction.factor.activation;
           var user = transaction.user;
+          var model = this;
 
           var accountInfo = {
             rpDisplayName: activation.rpDisplayName,
@@ -79,6 +81,17 @@ function (Okta, FormController, FormType, webauthn, Footer, WindowsHelloErrorMes
               publicKey: creds.publicKey,
               attestation: null
             });
+          })
+          .fail(function (error) {
+            switch (error.message) {
+            case 'AbortError':
+            case 'NotFoundError':
+            case 'NotSupportedError':
+              model.trigger('abort', error.message);
+              return transaction;
+            }
+
+            throw error;
           });
         });
       }
@@ -86,12 +99,16 @@ function (Okta, FormController, FormType, webauthn, Footer, WindowsHelloErrorMes
 
     Form: {
       autoSave: true,
-      hasSavingState: true,
+      hasSavingState: false,
       title: _.partial(Okta.loc, 'enroll.windowsHello.title', 'login'),
       subtitle: function () {
         return webauthn.isAvailable() ? Okta.loc('enroll.windowsHello.subtitle', 'login') : '';
       },
       save: _.partial(Okta.loc, 'enroll.windowsHello.save', 'login'),
+
+      customSavingState: {
+        stop: 'abort'
+      },
 
       modelEvents: function () {
         if (!webauthn.isAvailable()) {
@@ -100,7 +117,8 @@ function (Okta, FormController, FormType, webauthn, Footer, WindowsHelloErrorMes
 
         return {
           'request': '_startEnrollment',
-          'error': '_stopEnrollment'
+          'error': '_stopEnrollment',
+          'abort': '_stopEnrollment'
         };
       },
 
@@ -110,25 +128,65 @@ function (Okta, FormController, FormType, webauthn, Footer, WindowsHelloErrorMes
 
       formChildren: function () {
         var result = [];
+
         if (!webauthn.isAvailable()) {
           result.push(
             FormType.View(
-              { View: WindowsHelloErrorMessageView },
-              { selector: '.o-form-error-container' }
+              { View: new WindowsHelloErrorMessageView(
+                { message: Okta.loc('enroll.windowsHello.error.notWindows', 'login') }) },
+              { selector: '.o-form-error-container'}
             )
           );
         }
+
+        result.push(FormType.View({ View: new Spinner({ model: this.model, visible: false }) }));
+
         return result;
       },
 
       _startEnrollment: function () {
         this.subtitle = Okta.loc('enroll.windowsHello.subtitle.loading', 'login');
+
+        this.model.trigger('spinner:show');
+        this.$('.o-form-button-bar').addClass('hide');
+        this._resetErrorMessage();
+
         this.render();
       },
 
-      _stopEnrollment: function () {
+      _stopEnrollment: function (errorMessage) {
         this.subtitle = Okta.loc('enroll.windowsHello.subtitle', 'login');
+
+        this.model.trigger('spinner:hide');
+        this.$('.o-form-button-bar').removeClass('hide');
+
+        var message;
+        switch (errorMessage){
+        case 'NotSupportedError':
+          message = Okta.loc('enroll.windowsHello.error.notConfiguredHtml', 'login');
+          break;
+        }
+
+        this._resetErrorMessage();
+
+        if (message){
+          var messageView = new WindowsHelloErrorMessageView({
+            message: message
+          });
+
+          this.$('.o-form-error-container').addClass('o-form-has-errors');
+          this.add(messageView, { selector: '.o-form-error-container' });
+          this._errorMessageView = this.last();
+        }
+
         this.render();
+      },
+
+
+      _resetErrorMessage: function () {
+        this._errorMessageView && this._errorMessageView.remove();
+        this._errorMessageView = undefined;
+        this.clearErrors();
       }
     },
 
--- a/src/VerifyWindowsHelloController.js
+++ b/src/VerifyWindowsHelloController.js
@@ -15,16 +15,18 @@ define([
   'util/FormController',
   'util/FormType',
   'util/webauthn',
+  'views/shared/Spinner',
   'views/shared/FooterSignout',
   'views/mfa-verify/WindowsHelloErrorMessageView'
 ],
-function (Okta, FormController, FormType, webauthn, FooterSignout, WindowsHelloErrorMessageView) {
+function (Okta, FormController, FormType, webauthn, Spinner, FooterSignout, WindowsHelloErrorMessageView) {
 
   var _ = Okta._;
 
   return FormController.extend({
     className: 'verify-windows-hello',
     Model: {
+
       save: function () {
         if (!webauthn.isAvailable()) {
           return;
@@ -40,8 +42,8 @@ function (Okta, FormController, FormType, webauthn, FooterSignout, WindowsHelloE
           });
 
           return factor.verify()
-          .then(function (transaction) {
-            var factorData = transaction.factor;
+          .then(function (verifyData) {
+            var factorData = verifyData.factor;
 
             return webauthn.getAssertion(
               factorData.challenge.nonce,
@@ -54,6 +56,17 @@ function (Okta, FormController, FormType, webauthn, FooterSignout, WindowsHelloE
                 clientData: assertion.clientData,
                 signatureData: assertion.signature
               });
+            })
+            .fail(function (error) {
+              switch (error.message) {
+              case 'AbortError':
+              case 'NotFoundError':
+              case 'NotSupportedError':
+                model.trigger('abort', error.message);
+                return transaction;
+              }
+
+              throw error;
             });
           });
         });
@@ -62,11 +75,16 @@ function (Okta, FormController, FormType, webauthn, FooterSignout, WindowsHelloE
 
     Form: {
       autoSave: true,
+      hasSavingState: false,
       title: _.partial(Okta.loc, 'factor.windowsHello', 'login'),
       subtitle: function () {
         return webauthn.isAvailable() ? Okta.loc('verify.windowsHello.subtitle', 'login') : '';
       },
       save: _.partial(Okta.loc, 'verify.windowsHello.save', 'login'),
+
+      customSavingState:{
+        stop: 'abort'
+      },
 
       modelEvents: function () {
         if (!webauthn.isAvailable()) {
@@ -76,6 +94,7 @@ function (Okta, FormController, FormType, webauthn, FooterSignout, WindowsHelloE
         return {
           'request': '_startEnrollment',
           'error': '_stopEnrollment',
+          'abort': '_stopEnrollment',
           'sync': '_successEnrollment'
         };
       },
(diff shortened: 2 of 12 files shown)


rchild-okta:
nit: "Windows Hello **is** not configured" in the error prompt

src/EnrollWindowsHelloController.js
```diff
@@ -110,25 +125,73 @@ function (Okta, FormController, FormType, webauthn, Footer, WindowsHelloErrorMes
...
+          result.push(FormType.View({
+            View: '\
+            <p class="okta-form-subtitle o-form-explain text-align-c">\
+              {{i18n code="enroll.windowsHello.subtitle.loading" bundle="login"}}\
+            </p>',
+            showWhen: {__enrollmentState__: 'loading'}
+          }));
+        }
+
+        result.push(FormType.View({View: '<div class="okta-waiting-spinner hide"></div>'}));
```
rchild-okta: Have you checked with @nbories-okta on the design for the spinning state? We normally use the beacon for loading progress (you can trigger a loading state against it).

src/views/mfa-verify/WindowsHelloErrorMessageView.js
```diff
@@ -18,14 +18,22 @@ define([
 
     var template = '\
       <span class="icon error-24"></span>\
-      <h4><strong>{{i18n code="enroll.windowsHello.error.notWindows" bundle="login"}}</strong></h4>\
+      <h4><strong>{{{message}}}</strong></h4>\
```
rchild-okta: Why do you need the unescaped variant? And if it's necessary, we should explicitly add a comment describing why.

test/unit/spec/EnrollWindowsHello_spec.js
```diff
@@ -131,7 +131,7 @@ function (Okta,
         });
       });
 
-      itp('subtitle changes after submitting the form', function () {
+      /*itp('subtitle changes after submitting the form', function () {
```
rchild-okta: Why is this commented out?

test/unit/spec/webauthn_spec.js
```diff
@@ -72,7 +71,7 @@ function (Q, Expect, webauthn) {
       expect(webauthn.isAvailable()).toBe(true);
     });
 
-    it('makeCredential response has xhr.responseJSON.errorSummary if NotSupportedError is triggered',
+    /*it('makeCredential response has xhr.responseJSON.errorSummary if NotSupportedError is triggered',
```
rchild-okta: Same with this one. In general, I see that there are no tests added - add tests for the functionality you add.

src/EnrollWindowsHelloController.js
lilyhsiao-okta: @rchild-okta  These mocks are from @vincentvoong-okta. @vincentvoong-okta  did you have @nbories-okta  or @bhansen-okta review these? 
I think @alexeystadnikov-okta  can forward with these changes for beta. We can make additional spinner changes later in EA.

rchild-okta: @lilyhsiao-okta it's less work to switch to to existing spinner - all the logic is already there.

@alexeystadnikov-okta a reference example:
https://github.com/okta/okta-signin-widget/blob/b28719fc373a9cd0252ddb7d915bd7e88677d0ac/src/models/PrimaryAuth.js#L81

mauriciocastillosilva-okta: We have this same spinner for U2F, so this is to make Windows Hello use the same design. 
If I remember correctly the one you are talking @rchild-okta is for loading around the beacon, which is more from transition from one state to the other. The spinner here is to show that we are waiting for some external authorization to happen (either U2F or Windows Hello). 
I don't know the answer about Nico and Vince talking about it, but I would assume they did.

rchild-okta: Let's wait on @nbories-okta or @vincentvoong-okta to respond - right now it's unclear to me why we have two different spinners (if it's in U2f, we should switch that one to the beacon spinner as well).

mauriciocastillosilva-okta: Design is here: https://oktawiki.atlassian.net/wiki/display/PM/Windows+10+Hello+Integrated+Experience
nico is on PTO, let's see if Vince can chime in

rchild-okta: In that case, if we're now going to be using it in multiple places (i.e. in these 2 hello controllers and the 2 u2f controllers), let's create a shared view so that we don't have to keep doing the same logic again.

src/EnrollWindowsHelloController.js
```diff
@@ -79,28 +81,41 @@ function (Okta, FormController, FormType, webauthn, Footer, WindowsHelloErrorMes
...
       autoSave: true,
-      hasSavingState: true,
+      hasSavingState: false,
       title: _.partial(Okta.loc, 'enroll.windowsHello.title', 'login'),
-      subtitle: function () {
-        return webauthn.isAvailable() ? Okta.loc('enroll.windowsHello.subtitle', 'login') : '';
-      },
       save: _.partial(Okta.loc, 'enroll.windowsHello.save', 'login'),
 
+      customSavingState:{
```
rchild-okta: nit: space

src/views/mfa-verify/WindowsHelloErrorMessageView.js
alexeystadnikov-okta: https://oktawiki.atlassian.net/wiki/display/PM/Windows+10+Hello+Integrated+Experience?preview=/146506847/151487143/03b%20-%20Win%20Hello%202FA%20Error2.png

Here is the mock with the arrows (&gt;) and the bold words, for which I need to use html markup without escaping.

test/unit/spec/EnrollWindowsHello_spec.js
alexeystadnikov-okta: Subtitle logic was changed to use showWhen functionality, so this test is no longer needed.

test/unit/spec/webauthn_spec.js
alexeystadnikov-okta: New tests added

alexeystadnikov-okta:
- Spinner is moved to the shared view.
- Tests added. One test was removed due to no longer needed.
- Variable should be unescaped due to using html in the i18n translation (https://oktawiki.atlassian.net/wiki/display/PM/Windows+10+Hello+Integrated+Experience?preview=/146506847/151487143/03b%20-%20Win%20Hello%202FA%20Error2.png)

src/VerifyWindowsHelloController.js
```diff
@@ -40,8 +45,8 @@ function (Okta, FormController, FormType, webauthn, FooterSignout, WindowsHelloE
           });
 
           return factor.verify()
-          .then(function (transaction) {
-            var factorData = transaction.factor;
+          .then(function (factorTtransaction) {
```
rchild-okta: factorTransaction

src/VerifyWindowsHelloController.js
```diff
@@ -54,6 +59,17 @@ function (Okta, FormController, FormType, webauthn, FooterSignout, WindowsHelloE
...
                 signatureData: assertion.signature
               });
+            })
+            .fail(function (error) {
+              switch (error.message) {
+              case 'AbortError':
+              case 'NotFoundError':
+              case 'NotSupportedError':
+                model.trigger('abort', error.message);
+                return transaction;
```
rchild-okta: I am wondering what happens here since we're returning a different transaction (the original transaction, not the current state which is the new factorTransaction). Are we still able to continue?

src/views/mfa-verify/WindowsHelloErrorMessageView.js
rchild-okta: Sorry, I meant to add the comment to the code, like here:
https://github.com/okta/okta-signin-widget/blob/master/src/EnrollU2FController.js#L93

test/unit/spec/EnrollWindowsHello_spec.js
```diff
@@ -187,6 +189,42 @@ function (Okta,
...
         });
       });
+
+      itp('does not show error if webauthn.makeCredential fails with AbortError', function () {
+        return emulateWindows('AbortError')
+        .then(setup)
+        .then(function (test) {
+          test.setNextResponse([responseMfaEnrollActivateWebauthn, responseSuccess]);
+          test.form.submit();
+          return tick(test);
```
rchild-okta: We now have Expect statements - in this case, we should use Expect.waitForFormError and then do the checks in the next statement.

test/unit/spec/EnrollWindowsHello_spec.js
```diff
@@ -187,6 +189,42 @@ function (Okta,
...
+        });
+      });
+
+      itp('shows error if webauthn.makeCredential fails with NotSupportedError', function () {
+        return emulateWindows('NotSupportedError')
+        .then(setup)
+        .then(function (test) {
+          test.setNextResponse([responseMfaEnrollActivateWebauthn, responseSuccess]);
+          test.form.submit();
+          return tick(test);
```
rchild-okta: Same here (and for all the following tests) - there are no longer very many cases where we should actually use a tick() - check out Expect.js for the different wait conditions.

test/unit/spec/EnrollWindowsHello_spec.js
```diff
@@ -126,24 +141,11 @@ function (Okta,
         return emulateWindows()
         .then(setup)
         .then(function (test) {
-          expect(test.form.hasErrorNotWindows()).toBe(false);
+          expect(test.form.hasErrorWindowsHello()).toBe(false);
           expect(test.form.submitButton().length).toBe(1);
         });
       });
 
-      itp('subtitle changes after submitting the form', function () {
```
rchild-okta: We are removing this, but we are not replacing it with tests to cover the changed functionality - anything to do with our actual usage of the spinner, or the various views that show up with different enrollment states. This seems like the main purpose of the PR - we should add test coverage for new functionality that we're adding.

rchild-okta:
Also, another note: Can you update the commit message to not include the jiras in the title line? Look at other commits in this repo for the convention:

```
First line < 60 characters so that it shows up correctly in github

Any additional info that you want to add to the message, each line < 80 chars

RESOLVES OKTA-1234
```

src/VerifyWindowsHelloController.js
alexeystadnikov-okta: Fixed

src/VerifyWindowsHelloController.js
alexeystadnikov-okta: Sure, we can.
I have renamed the variable `factorTransaction` to `verifyData` and it should become more understandable.

`factorTransaction` was a request result of the

test/unit/spec/EnrollWindowsHello_spec.js
alexeystadnikov-okta: Done.

test/unit/spec/EnrollWindowsHello_spec.js
alexeystadnikov-okta: Done. Thanks.

alexeystadnikov-okta:
Commit message updated.

src/EnrollWindowsHelloController.js
```diff
@@ -110,25 +126,73 @@ function (Okta, FormController, FormType, webauthn, Footer, WindowsHelloErrorMes
...
+        switch (errorMessage){
+        case 'NotSupportedError':
+          message = Okta.loc('enroll.windowsHello.error.notConfiguredHtml', 'login');
+          break;
+        }
+
+        this._resetErrorMessage();
+
+        if(message){
+          var messageView = new WindowsHelloErrorMessageView();
```
mauriciocastillosilva-okta: nit: can we do something like 
```
new WindowsHelloErrorMessageView({
  message: message
});
```

src/VerifyWindowsHelloController.js
```diff
@@ -94,22 +113,88 @@ function (Okta, FormController, FormType, webauthn, FooterSignout, WindowsHelloE
...
+
+        case 'NotSupportedError':
+          message = Okta.loc('enroll.windowsHello.error.notConfiguredHtml', 'login');
+          break;
+        }
+
+        this._resetErrorMessage();
+
+        if (message) {
+          var messageView = new WindowsHelloErrorMessageView();
```
mauriciocastillosilva-okta: same nit as in Enrollment controller

src/VerifyWindowsHelloController.js
```diff
@@ -94,22 +113,88 @@ function (Okta, FormController, FormType, webauthn, FooterSignout, WindowsHelloE
...
+        }
+
+        result.push(FormType.View({View: Spinner}));
+
         return result;
       },
 
+      postRender: function () {
+        if (this.options.appState.get('factors').length === 1 && !this.model.get('__enrollmentState__')) {
+          this.model.save();
```
mauriciocastillosilva-okta: can you tell me what is this ```postRender``` for? Is it to show Windows Hello modal right away?
If so, this should not be only when windows is the only factor, but instead it should be when windows hello is the selected factor (does not matter if there are more factors or not)
@vincentvoong-okta is this correct?

src/views/shared/Spinner.js
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define(['okta'], function (Okta) {
+
+  return Okta.View.extend({
+    className: 'okta-waiting-spinner hide',
```
mauriciocastillosilva-okta: I think this should start without ```hide```, since it is a shared component. That way if the consumer wants to hide it, he explicitly needs to do so. What do you think?

src/EnrollWindowsHelloController.js
alexeystadnikov-okta: Done

src/VerifyWindowsHelloController.js
alexeystadnikov-okta: Done

src/views/shared/Spinner.js
alexeystadnikov-okta: Done.
Added boolean option `visible` to hide it if needed it the initialize.

node_modules/@okta/i18n/package.json
```diff
@@ -1,40 +1,32 @@
...
+      "@okta/i18n@19.13.0",
+      "/Users/astadnikov/okta/okta-signin-widget"
     ]
   ],
-  "_from": "@okta/i18n@18.7.0",
-  "_id": "@okta/i18n@18.7.0",
+  "_from": "@okta/i18n@19.13.0",
+  "_id": "@okta/i18n@19.13.0",
   "_inCache": true,
+  "_installable": true,
```
mauriciocastillosilva-okta: this has given trouble before, so if Travis has an error about not being able to resolve i18n, you might want to consider removing it

src/VerifyWindowsHelloController.js
mauriciocastillosilva-okta: I think that to know that the factor is the "default" one, we need to verify that it is the first one on the list of factors we get back from the API... not sure we should do that here and behave differently if it is the first one as opposed to be in any other position on the list. @rchild-okta do you have any opinion on this one?

src/views/mfa-verify/WindowsHelloErrorMessageView.js
alexeystadnikov-okta: Done

test/unit/spec/EnrollWindowsHello_spec.js
alexeystadnikov-okta: Done

src/VerifyWindowsHelloController.js
mauriciocastillosilva-okta: After thinking about it a little more, it seems to me that this is something that we will need to add, but not as simple. So for now I think we should leave it as it is (open the modal when WindowsHello is the only factor), and if there are more factors then show the button.

src/EnrollWindowsHelloController.js
```diff
@@ -110,25 +126,75 @@ function (Okta, FormController, FormType, webauthn, Footer, WindowsHelloErrorMes
...
-              { View: WindowsHelloErrorMessageView },
-              { selector: '.o-form-error-container' }
+              { View: new WindowsHelloErrorMessageView(
+                { message: Okta.loc('enroll.windowsHello.error.notWindows', 'login') }) },
+              { selector: '.o-form-error-container'}
             )
           );
         }
+        else {
+          result.push(FormType.View({
```
rchild-okta: Just a note since I didn't catch this in my original review - is there a reason we're waiting on states for the subtitle instead of just rendering? I.e. I would expect this logic to be in the subtitle property, and whenever we change the enrollmentState just to re-render the entire form. 

Is the reason to do it this way because re-rendering with the spinner transitions poorly?

The main reason to do it the re-rendering way is because we are now copying subtitle elements and classes here - if we make a change to how the elements look in the underlying form view, we will need to remember to update it here.

mauriciocastillosilva-okta: I agree, I didn't catch that either
Can we make that change Alexey? It makes sense to have the text in the ```subtitle``` as opposed to copying the css needed for it in a different view.

alexeystadnikov-okta: By default subtitle is displayed above the error messages.
On the mock subtitle is placed below the error. 
https://oktainc.atlassian.net/secure/attachment/69667/07%20-%20Win%20Hello%20Challenge%20Error%20Contact%20Admin.png

rchild-okta: Hmmm, are you sure about that? I can look into the code later, but I just tried the primaryAuth and resetPassword flows and they show the errors below the subtitle.

mauriciocastillosilva-okta: @rchild-okta @alexeystadnikov-okta I think both of you are saying the same thing. And I think what is wrong is the mock.
@vincentvoong-okta we have the pattern of showing the errors below the subtitle, so we should stick to that pattern. Can you update the mocks?

rchild-okta: @mauriciocastillosilva-okta , ah, I did misread - yes, I'm guessing there isn't any reason why we would diverge from how we're doing it elsewhere for this particular screen - @vincentvoong-okta, can you confirm that we should do it the normal way?

alexeystadnikov-okta:
@mauriciocastillosilva-okta
@vincentvoong-okta 

Should I move subtitle to the standard place and make the widget errors during Windows Hello enrollment looks like the screenshot?

screen shot 2016-10-10 at 5 29 25 pm

src/EnrollWindowsHelloController.js
vincentvoong-okta: @rchild-okta @mauriciocastillosilva-okta @alexeystadnikov-okta 

I spoke with the UX team and the error notification belongs below the widget header and above any descriptive text.

After talking to @rchild-okta I suggest implementing the notification consistent with the current pattern and I will open a Jira for the fix.

mauriciocastillosilva-okta:
@alexeystadnikov-okta yes, Vince agreed that we should follow the current pattern.
For the future, a jira will be filed to change the patter in all places.

alexeystadnikov-okta:
@mauriciocastillosilva-okta 
Done.
Changes for showing the subtitle below the error message were reverted in controllers and in tests.

mauriciocastillosilva-okta:
Thanks @alexeystadnikov-okta. Can you please rebase latest master? I'll merge it tomorrow morning

alexeystadnikov-okta:
@mauriciocastillosilva-okta 
Rebased

PR #119, opened 2016-09-30
added validation for phone number field before sending sendCode request

eugenebakaiev-okta:
- field phoneNumber made not required

If user wants to enroll with phone number set while registration, he will be able to do that without entering a phone number again. If user has no phone numbers he will receive a message about it.

Resolves: OKTA-94851


eugenebakaiev-okta:
smsfactor mov

eugenebakaiev-okta:
@rchild-okta, @mauriciocastillosilva-okta, @stsai-okta could you please review?

stsai-okta:
could we add a test to this?

eugenebakaiev-okta:
@stsai-okta, yes I think, I'll add

rchild-okta:
I'm okay with the change (after adding the test), but I am curious about the use case - it doesn't seem very obvious to me that this is a flow the user would go down. Why would the user click the button to send the code if they haven't entered their phone number?

If the phone is already enrolled and the API knows about it, isn't in already in the factor response? And if it's in the factor response, is a better fix in this case to just expose the phone number in the box?

I.e. instead of fixing it by not requiring the phone number, a better fix would be to prepopulate the enrolled phone number in the phone text field.

@vincentvoong-okta, thoughts?

vincentvoong-okta:
@eugenebakaiev-okta 
I agree with @rchild-okta , the flow is not an ideal enduser experience. We may have an existing UI that is similar to the solution Reman described - I would suggest talking to Alex Bovee or John Gronberg on the PM team for next steps.

eugenebakaiev-okta:
We don't have a full number, but have a part of it (e.g. "+1 XXX-XXX-1234"). I can try to use this string as placeholder. Will it be suitable for us?

screen shot 2016-10-10 at 5 10 37 pm

eugenebakaiev-okta:
@rchild-okta, I've commited a POC fix for using number in a placeholder. If it's ok I'll continue to work on this

ahmedalbahar-okta:
Since the user is authenticated at the phone of setting up SMS, and from discussing with Matias from security, we should be able to expose the phone number without needing to hide characters.  

The user experience in my opinion should be to pre-populate the SMS phone number field with phone number set while registration and keep the field editable incase the user would like to use a different phone number at the point of SMS MFA enrollment.

eugenebakaiev-okta:
@abinashbehera-okta  the problem is that backend doesn't return the whole number. It returns the number with hidden characters

rchild-okta:
Ah, right, I remember that now. This seems okay to me - show the placeholder number (with the x's), and if they try to edit it maybe clear out the whole field. @ahmedalbahar-okta/ @nbories-okta , does that work for you?

ahmedalbahar-okta:
Ah ok. I agree with Reman's suggestion. Show the placeholder number (with the x's), and if they try to edit it maybe clear out the whole field.

ahmedalbahar-okta:
Can we make sure the country code doesn't appear twice like in the screenshot?

rchild-okta:
I **think** we might be okay with just parsing it out from the returned string, i.e. just parse out the "+/d" from whatever the backend gives us.

eugenebakaiev-okta:
@ahmedalbahar-okta, sure. And also, should we select the country with the code used in the existing number?

ahmedalbahar-okta:
Ok. Lets put a pause on this fix until we get more customer feedback. This requires more thought on the UX or even if this is something we want to do.

mauriciocastillosilva-okta:
Just as a general note: the same logic we decide to do for SMS we should do for Voice Call.
I like the idea of the placeholder too, but we will need a little more logic. I

If there is a placeholder, then user can click the button and an SMS/Call will happen
If there is no placeholder, we should show an error about the field being blank

The other thing is that if we have a current number, then we will have to parse, as Reman said, the country code and update the country select field.

Those are my 2 cents.

eugenebakaiev-okta:
screen shot 2016-10-17 at 6 22 37 pm

eugenebakaiev-okta:
@rchild-okta, @mauriciocastillosilva-okta please take a look

PR #122, opened 2016-10-07
Localize security questions

mauriciocastillosilva-okta:
Use the localized text for the security question if available

Resolves: OKTA-101579
Bacon: test


src/EnrollQuestionController.js
```diff
@@ -94,7 +94,9 @@ function (Okta, FormController, Footer, TextBox) {
       .then(function(questionsRes) {
         var questions = {};
         _.each(questionsRes, function (question) {
-          questions[question.question] = question.questionText;
+          var localizedQuestion = Okta.loc('security.' + question.question);
```
rchild-okta: Let's write a Util helper function to do this logic so that we don't have to duplicate it

rchild-okta:
lgtm 🚀

PR #123, opened 2016-10-07
Created npm dist file which does not include thirdparty deps

rchild-okta:
Now, when going through the npm flow with require('@okta/okta-signin-widget'),
the file that is required is much smaller, and will not include thirdparty
dependencies. This allows the consumer to use their own versions of deps like
jquery, underscore, backbone, etc.

Other:
- Simplified Gruntfile copy, build, and prerelease commands, and sped up
  build by moving to parallel webpack config execution of release flow
- Moved more vendor files from source to npm:
  - jquery-placeholder
  - qtip2
  - xdomain
- Updated e2e test harness logic to use Handlebars partials
- Moved e2e tests from testing dev version to testing CDN and npm versions
- Fixed bug where we were including files from courage that we were trying
  to ignore (relative paths do not work with alias)

RESOLVES OKTA-102703


rchild-okta:
Some interesting file size stats:

| File | Before commit | After commit |
| ---- | ------------- | ------------ |
| okta-sign-in.min.js (uglified by default) | 615.0k | 577.2k |
| okta-sign-in.min.js, gzipped | 181.0k | 170.8k |
| okta-sign-in.no-jquery.js (uglified by default) | 516.6k | ~ |
| okta-sign-in.no-jquery.js, gzipped | 147.3k | ~ |
| okta-sign-in.entry.js (not uglified, with comments, etc) | ~ | 673.9k |
| okta-sign-in.entry.js, uglified | ~ | 261.2k |
| okta-sign-in.entry.js, uglified + gzipped | ~ | 66.1k | 

Notes:
- We're getting rid of the no-jquery option and replacing it with entry. It's the same idea, except we're removing all the dependencies we can instead of just jquery
- The .entry file is basically the size of "our code". There are still a couple improvements we can make:
  - We still include chosen in source because we have local modifications. We need to upgrade/use a different library from npm and move it out
  - qtip2 is huge (100k) - if we replace this, we'll get some good file size savings
- The .min.js file is what we're serving off the CDN, and includes our code + all dependencies

assets/sass/modules/_qtip.scss
```diff
@@ -36,3 +36,50 @@
     }
   }
 }
+
+/*! Custom qtip styles, based on dark theme */
+.qtip-custom {
```
haishengwu-okta: wonder the scss lint shall generate an warning that combine line 7 and line 41-85 since they all `.qtip-custom` ?

test/e2e/page-objects/OktaHomePage.js
```diff
@@ -0,0 +1,31 @@
...
+
+class OktaHomePage {
+
+  constructor() {
+    this.mainContentEl = $('.main-content');
+  }
+
+  waitForPageLoad() {
+    browser.wait(EC.presenceOf(this.mainContentEl));
+    expect(browser.getCurrentUrl()).toMatch('/app/UserHome');
```
haishengwu-okta: nit: 
verify content like assert user name show up at page header.

webpack.common.config.js
```diff
@@ -1,72 +1,100 @@
...
+      path: TARGET_JS,
+      filename: outputFilename,
+      library: 'OktaSignIn',
+      libraryTarget: 'umd'
+    },
+    resolve: {
+      root: [TARGET_JS],
+      alias: {
+        // General remapping
+        'nls': '@okta/i18n/dist/json',
```
haishengwu-okta: question: I'll assume it's looking for file at `node_modules/@okta/i18n/...`. so how would it work given the `resolve.root` is `TARGET_JS`?

assets/sass/modules/_qtip.scss
rchild-okta: Looks like it's okay (passed the linter) - for this scss change I mostly just copied our overrides from before from inside the vendor css file, to our actual override file.

webpack.common.config.js
rchild-okta: That's one of the nice things about webpack (and the reason to switch from requireJs) - it knows about the `node_modules` folder and how to resolve module paths. So in this case, it resolves it against `resolve.root` + `node_modules`

rchild-okta:
@haishengwu-okta, re "documenting which versions we support" - you're entirely correct that this is an issue.

Currently, we don't actually support them using a different version of the dependencies - this is governed by our package.json, whose dependencies don't support semver ranges (all of our deps are listed as exact versions). When they build with webpack, I think it might actually just include both versions if their's is different (need to check this). Followup work needs to be done to actually see which ranges we can support for each of these dependencies, and then we can increase the ranges in the package.json. Will need to also see if a shrinkwrap affects this at all either. 

This PR is pretty much a first pass - move as many vendor files into npm as possible, cleanup some build stuff, and export a version with no deps included that they can require into their own project (and in this case, they can still get one version of each dep by matching our versions).

I created this jira to followup with this:
https://oktainc.atlassian.net/browse/OKTA-103675

rchild-okta:
@haishengwu-okta pushed another commit to do two things:

1. Added a username check to the test (re your comment)
2. Added the no-jquery dist back in - removing it meant that we had to go down the path of converting `clients/loginpage` to webpack before pulling in any new widget changes. I'm currently working on that in [this jira](https://oktainc.atlassian.net/browse/OKTA-94209), but it turns out it's a bit more work than I originally expected. Will first check in this code, and remove the no-jquery build once we get loginpage rockin' and rollin' on webpack.

assets/sass/modules/_qtip.scss
haishengwu-okta: aha, I think it's because `okta-signin-widget/.scss-lint.yml` disabled many rules..

haishengwu-okta:
@rchild-okta looks great!

src/util/BaseLoginRouter.js
```diff
@@ -98,7 +98,7 @@ function (Okta, Backbone, BrowserFeatures, xdomain, RefreshAuthStateController,
         $.support.cors = true;
         xdomainSlaves = {};
         xdomainSlaves[this.settings.get('baseUrl')] = '/cors/proxy';
-        xdomain.slaves(xdomainSlaves);
+        XDomain.xdomain.slaves(xdomainSlaves);
```
lboyette-okta: I can't wait to get rid of this.

webpack.common.config.js
```diff
@@ -1,72 +1,100 @@
...
-  plugins: [
     // Webpack attempts to add a polyfill for process
     // and setImmediate, because q uses process to see
     // if it's in a Node.js environment
-    new webpack.ProvidePlugin({
-      process: empty,
-      setImmediate: empty
-    })
-  ]
+    node: {
```
lboyette-okta: oh, so there is an explicit way to remove these

PR #124, opened 2016-10-10
Fixes authorization code issues when converting sessionToken

rchild-okta:
- Authorization code flow will now automatically convert sessionToken through
  the redirect flow. Fixes bug where this was done previously through
  okta_post_message, which does not work.
- Setting page:display will now go through the redirect flow for other token
  types as well.

RESOLVES OKTA-103582


yiyangli-okta:
lgtm 🚢 

(sorry, it takes me a while to understand how and why the responseType array is concatenated. )

PR #127, opened 2016-10-18
Fixes Widget if Windows Hello is the only factor.

alexeystadnikov-okta:
Resolves OKTA-104645 (Sign in Widget broken if Windows Hello is the only factor)
- [x] Functionality verified.
- [x] Tests added.

@mauriciocastillosilva-okta 
@ukilon-okta 
@rchild-okta 

Please, review.

### DIFF
--- a/src/VerifyWindowsHelloController.js
+++ b/src/VerifyWindowsHelloController.js
@@ -24,8 +24,11 @@ function (Okta, FormController, FormType, webauthn, Spinner, FooterSignout, Wind
   var _ = Okta._;
 
   return FormController.extend({
-    className: 'verify-windows-hello',
+    className: 'mfa-verify verify-windows-hello',
     Model: {
+      local: {
+        __autoTriggered__: 'boolean'
+      },
 
       save: function () {
         if (!webauthn.isAvailable()) {
@@ -125,7 +128,8 @@ function (Okta, FormController, FormType, webauthn, Spinner, FooterSignout, Wind
       },
 
       postRender: function () {
-        if (this.options.appState.get('factors').length === 1 && !this.model.get('__enrollmentState__')) {
+        if (this.options.appState.get('factors').length === 1 && !this.model.get('__autoTriggered__')) {
+          this.model.set('__autoTriggered__', true);
           this.model.save();
         }
       },
--- a/test/unit/spec/MfaVerify_spec.js
+++ b/test/unit/spec/MfaVerify_spec.js
@@ -1,4 +1,4 @@
-/*jshint maxparams:41, maxstatements:40, camelcase:false, newcap:false */
+/*jshint maxparams:50, maxstatements:40, camelcase:false, newcap:false */
 /*global JSON */
 define([
   'vendor/lib/q',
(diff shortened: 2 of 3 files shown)


src/VerifyWindowsHelloController.js
```diff
@@ -121,7 +124,8 @@ function (Okta, FormController, FormType, webauthn, Spinner, FooterSignout, Wind
       },
 
       postRender: function () {
-        if (this.options.appState.get('factors').length === 1 && !this.model.get('__enrollmentState__')) {
+        if (this.options.appState.get('factors').length === 1 && !this.model.get('__autoTriggered__')) {
```
mauriciocastillosilva-okta: Do we really need the extra prop ```__autoTriggered__```? What I am thinking is that if Windows Hello is the only factor, then we should just open the windows modal. Since it is the only factor, they can't really navigate anywhere else. Or what will be the case where we need to check if ```__autoTriggered__``` is true?

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1915,6 +1919,14 @@ function (Q,
             expect(test.form.subtitleText()).toBe('Signing into Okta...');
           });
         });
+
+        itp('automatically triggers Windows Hello only once', function () {
+          return emulateWindows()
+          .then(setupWebauthnOnly)
+          .then(function (test) {
+            expect(test.router.controller.model.get('__autoTriggered__')).toBe(true);
```
mauriciocastillosilva-okta: I think what we should try to test is if ```model.save``` is called automatically without the need to press any button.

src/VerifyWindowsHelloController.js
alexeystadnikov-okta: The bug was due to the recursion in the logic:
1. The view was added and calls `render()`
2. `render()` calls `postRender()`
3. If Windows Hello is the only factor, `postRender()` calls the `model.save()`
4. `model.save()` triggers the `request` event, that changes the title and calls `render()`.
5. See step 2

To check if it is the very first `postRender` I added the `__autoTriggered__ ` flag.

mauriciocastillosilva-okta: I see. Ok, then I think this is good.

test/unit/spec/MfaVerify_spec.js
mauriciocastillosilva-okta: I think this is the only thing missing, to check for ```save``` being called, and maybe a test to verify that if ```__autoTriggered__``` is true we don't call ```model.save``` again

alexeystadnikov-okta:
Tests updated.
Message updated.

test/unit/spec/MfaVerify_spec.js
```diff
@@ -162,6 +164,8 @@ function (Q,
     var setupOktaPush = _.partial(setup, resAllFactors, { factorType: 'push', provider: 'OKTA' });
     var setupOktaTOTP = _.partial(setup, resVerifyTOTPOnly, { factorType: 'token:software:totp' });
     var setupWebauthn = _.partial(setup, resAllFactors, {  factorType: 'webauthn', provider: 'FIDO' });
+    var setupWebauthnOnly = _.partial(setup, [resRequiredWebauthn, resChallengeWebauthn, resSuccess],
```
mauriciocastillosilva-okta: I think this should be ```_.partial(setup, resRequiredWebauthn, {  factorType: 'webauthn', provider: 'FIDO' });
If you look at line 126 we set there that the next response will be ```resChallengeWebauthn```, so that is taken care of.

alexeystadnikov-okta: Debugging the code I found, that in time when `Expect.waitForMfaVerify()` is done, the controller is already rendered and the `model.save()` is called.
So, to prevent the ajax error, when we make the request without the mock, I should pass all three responses.

mauriciocastillosilva-okta:
Looks good to me, but Travis is failing in a test, which is not yours... my guess is that your test is doing something that it should not.

alexeystadnikov-okta:
Repushed. Green.

PR #128, opened 2016-10-18
Hides button during Windows Hello enrollment and verifying when spinner is displayed

alexeystadnikov-okta:
Resolves OKTA-104478 (Windows Hello: remove button when spinner is displayed)

screen shot 2016-10-18 at 7 02 44 pm

screen shot 2016-10-18 at 7 02 03 pm

@mauriciocastillosilva-okta 
@ukilon-okta 
@rchild-okta 

Please, review.

### DIFF
--- a/src/VerifyWindowsHelloController.js
+++ b/src/VerifyWindowsHelloController.js
@@ -50,12 +50,16 @@ function (Okta, FormController, FormType, webauthn, Spinner, FooterSignout, Wind
               [{ id: factorData.profile.credentialId }]
             )
             .then(function (assertion) {
-              model.trigger('sync');
               return factor.verify({
                 authenticatorData: assertion.authenticatorData,
                 clientData: assertion.clientData,
                 signatureData: assertion.signature
               });
+            })
+            .then(function (data) {
+              model.trigger('sync');
+              model.trigger('signIn');
+              return data;
             })
             .fail(function (error) {
               switch (error.message) {
@@ -95,7 +99,7 @@ function (Okta, FormController, FormType, webauthn, Spinner, FooterSignout, Wind
           'request': '_startEnrollment',
           'error': '_stopEnrollment',
           'abort': '_stopEnrollment',
-          'sync': '_successEnrollment'
+          'signIn': '_successEnrollment'
         };
       },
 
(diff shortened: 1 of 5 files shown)


alexeystadnikov-okta:
Tests added.

src/VerifyWindowsHelloController.js
```diff
@@ -57,6 +57,10 @@ function (Okta, FormController, FormType, webauthn, Spinner, FooterSignout, Wind
                 signatureData: assertion.signature
               });
             })
+            .then(function (data) {
+              model.trigger('signIn');
```
mauriciocastillosilva-okta: not sure I follow what is going on exactly, since this seems to me is not needed due to the ```trigger('sync')``` in line 53, which will show the spinner
And looking a little bit more at this, we should show the spinner and hide the button before we do the call to ```weauthn.getAssertion(``` in line 48, right? 
And then hide the spinner and show the button if the user cancels or there was an error.

src/VerifyWindowsHelloController.js
```diff
@@ -95,7 +99,7 @@ function (Okta, FormController, FormType, webauthn, Spinner, FooterSignout, Wind
           'request': '_startEnrollment',
           'error': '_stopEnrollment',
           'abort': '_stopEnrollment',
-          'sync': '_successEnrollment'
```
mauriciocastillosilva-okta: I think ```sync``` was used in line 53 to start the enrollment. But take a look at my other comment see if that makes sense.

alexeystadnikov-okta: The spinner becomes visible on the `request` event that is at the beginning of the `save` function.
To hide the button bar I have to add the `hide` class to the `'.o-form-button-bar'` and after `render` it becomes visible again.
Unfortunately, If I use the `sync` event my callback is not the last one.
The last one is in the BaseForm.js:
```
this.listenTo(this.model, 'sync', function () {
        if (this.model.get('__edit__')) {
          this.model.set('__edit__', false, {silent: true});
        }
        this.__saveModelState(this.model);
        this.render();
      });
```
and it renders the view again and makes the button visible.

To make it work I trigger the `signIn` event after the `sync` and the button bar stays hidden.

src/VerifyWindowsHelloController.js
alexeystadnikov-okta: Answered below.

PR #105, opened 2016-08-29
OKTA-99168 New Signin Flow does not provide adequate visual indication - violation of Section 508 and WCAG 2

eugenebakaiev-okta:
Added css class for focused text input, added handlers for focusin and focusout events

PR #129, opened 2016-10-20
Add aria-label to the input fields on the signin page.

hor-kanchan-okta:
:bug: Add aria-label to input fields

To be 508 compliance, all form elements must have unique and descriptive labels.
Add an 'aria-label' attribute and use the same value as the placeholder which is a 
localized string.

RESOLVES OKTA-104131

### DIFF
--- a/src/views/shared/TextBox.js
+++ b/src/views/shared/TextBox.js
@@ -52,8 +52,8 @@ function ($, Handlebars, BrowserFeatures, TextBox) {
           <span class="icon input-icon {{params.icon}}"></span>\
         {{/if}}\
       {{/if}}\
-      <input type="{{type}}" placeholder="{{placeholder}}" name="{{name}}" \
-        id="{{inputId}}" value="{{value}}" autocomplete="off"/>\
+      <input type="{{type}}" placeholder="{{placeholder}}" aria-label="{{placeholder}}"\
+        name="{{name}}" id="{{inputId}}" value="{{value}}" autocomplete="off"/>\
     '),
 
     postRender: function () {
--- a/test/unit/spec/TextBox_spec.js
+++ b/test/unit/spec/TextBox_spec.js
@@ -1,0 +1,22 @@
+define([
+  'helpers/util/Expect',
+  'okta',
+  'views/shared/TextBox',
+  'sandbox'
+],
+function (Expect, Okta, TextBox, $sandbox) {
+  Expect.describe('TextBox', function () {
+    it('the value of aria-label is same as the placeholder', function () {
+      var placeHolderValue = 'Test Value 123';
+  
+      var textbox = new TextBox({
+        model: new Okta.BaseModel(),
+        id: Okta._.uniqueId('textbox'),
+        placeholder: placeHolderValue
+      });
+
+      $sandbox.html(textbox.render().el);  
+      expect(textbox.$('#' + textbox.options.inputId).attr('aria-label')).toEqual(placeHolderValue);
+    });
+  });
+});


taylorlaubach-okta:
Lgtm

jhe-okta:
Can you add test?

rchild-okta:
Couple comments:
- Check out [this comment](https://github.com/okta/okta-signin-widget/pull/105#issuecomment-243197622) for some guidelines on commit message formatting
- We are going to be automating release note generation starting this release. Ask @alanzhou-okta about what we need to add for this commit message (background info [here](https://oktawiki.atlassian.net/wiki/display/eng/Open-Source+Automated+Release+Notes)

jhe-okta:
Is there always value for placeholder? Can it be empty?

src/views/shared/TextBox.js
```diff
@@ -52,8 +52,8 @@ function ($, Handlebars, BrowserFeatures, TextBox) {
           <span class="icon input-icon {{params.icon}}"></span>\
         {{/if}}\
       {{/if}}\
-      <input type="{{type}}" placeholder="{{placeholder}}" name="{{name}}" \
-        id="{{inputId}}" value="{{value}}" autocomplete="off"/>\
+      <input type="{{type}}" placeholder="{{placeholder}}" aria-label="{{placeholder}}"\
```
anishashetty-okta: Since this is a visible element. shouldn't we be using the `aria-labelledby` attribute?
(just looked at this [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute))

jhe-okta: Neither of them are visible to me. But it is a good point. Based on what I just read, looks like aria-labelledby is preferred by screen reader.

hor-kanchan-okta: Most of time we have placeholder except views/mfa-verify/YubikeyForm.js
I will modify that file and add the placeholder.

hor-kanchan-okta:
We have placeholder most of time except views/mfa-verify/YubikeyForm.js
I will add the placeholder to that file.

alanzhou-okta:
@hor-kanchan-okta Check out these guidelines for your commit message. Noticeably, we need to add an emoji in front of your commit header in order for our automatic release notes to parse. Thanks!

src/views/shared/TextBox.js
hor-kanchan-okta: aria-labelledby contains ids. I don't have visible elements that it can point to. So aria-labelledby won't work unless I add hidden <div> or <label>

hor-kanchan-okta:
@alanzhou-okta Can you review my comment format?

rchild-okta:
fyi @hor-kanchan-okta you'll need to push another commit to this branch (can't just edit the github comment)

alanzhou-okta:
@hor-kanchan-okta that looks good to me. +1 to reman's comment

test/unit/spec/TextBox_spec.js
```diff
@@ -0,0 +1,23 @@
+define([
+  'underscore',
```
jhe-okta: you can get underscore through Okta._

test/unit/spec/TextBox_spec.js
```diff
@@ -0,0 +1,23 @@
...
+  'okta',
+  'views/shared/TextBox',
+  'sandbox'
+],
+function (_, Expect, Okta, TextBox, $sandbox) {
+  Expect.describe('TextBox', function () {
+    it('the value of aria-label is same as the placeholder', function () {
+      var placeHolderValue = 'Test Value 123';
+  
+      var textbox = new TextBox({
```
jhe-okta: nit: may be we can move the text box initialization to beforeEach, so that we can add more it blocks and they will be benefit from it.

ukilon-okta:
Curios: the spec doesn't really specify that we need an aria-label to the input.
I am guessing an attached `<label>` tag is sufficient.
Does this input has a matching <label> tag? I am guessing it would probably be different than the palce holder

hor-kanchan-okta:
@ukilon-okta 
The input element doesn't have a corresponding label element. I would think using aria-label is better option than creating a hidden label element.

taylorlaubach-okta:
@ukilon-okta My understanding is that one just needs to either wrap the input in the label, or use the "for" attribute on the label and give it a value equal to the id of the input.

rchild-okta:
lgtm 🚀

ukilon-okta:
@taylorlaubach-okta which we do on the framework level

PR #130, opened 2016-10-21
Handle OAuth Error Case

alanzhou-okta:
Show an error message when user is not assigned to OIDC client instead of bubbling up to global error function

Resolves OKTA-104330

### DIFF
--- a/src/util/OAuth2Util.js
+++ b/src/util/OAuth2Util.js
@@ -31,14 +31,20 @@ define(['okta', './Enums', './Errors'], function (Okta, Enums, Errors) {
    * @param params - {idp: 'xxx'} for social auth
    *                 {sessionToken: 'xxx'} for okta idp
    */
-  util.getTokens = function (settings, params) {
+  util.getTokens = function (settings, params, controller) {
 
     function success(result) {
       settings.callGlobalSuccess(Enums.SUCCESS, result);
     }
 
     function error(error) {
-      settings.callGlobalError(new Errors.OAuthError(error.message));
+      // OKTA-104330- Handle error case where user is not assigned to OIDC client
+      if (error.errorCode === 'access_denied') {
+        controller.model.trigger('error', controller.model, {'responseJSON': error});
+        controller.model.appState.trigger('removeLoading');
+      } else {
+        settings.callGlobalError(new Errors.OAuthError(error.message));
+      }
     }
 
     var authClient = settings.getAuthClient(),
--- a/test/e2e/specs/OIDC_spec.js
+++ b/test/e2e/specs/OIDC_spec.js
@@ -57,6 +57,20 @@ describe('OIDC flows', function() {
       expect(oidcApp.getIdTokenUser()).toBe('Saml Jackson');
     });
 
+    it('throws form error if auth client returns with OAuth error', function() {
+      setup({
+        baseUrl: '{{{WIDGET_TEST_SERVER}}}',
+        clientId: 'rW47c465c1wc3MKzHznu',
+        redirectUri: 'http://localhost:3000/done',
+        authParams: {
+          responseType: 'id_token',
+          scope: ['openid', 'email', 'profile', 'address', 'phone']
+        }
+      });
+      primaryAuth.loginToForm('{{{WIDGET_BASIC_USER_5}}}', '{{{WIDGET_BASIC_PASSWORD_5}}}');
+      expect(primaryAuth.getErrorMessage()).toBe('User is not assigned to the client application.');
+    });
+
     it('can login and get a token and id_token', function () {
       setup({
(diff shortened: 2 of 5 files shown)


alanzhou-okta:
screen shot 2016-10-21 at 12 56 06 pm

src/util/OAuth2Util.js
```diff
@@ -31,14 +31,20 @@ define(['okta', './Enums', './Errors'], function (Okta, Enums, Errors) {
...
 
     function success(result) {
       settings.callGlobalSuccess(Enums.SUCCESS, result);
     }
 
     function error(error) {
-      settings.callGlobalError(new Errors.OAuthError(error.message));
+      // OKTA-104330- Handle error case where user is not assigned to OIDC client
+      if (error.errorCode == 'access_denied') {
+        controller.model.trigger('error', controller.model, {'responseJSON': error});
```
rchild-okta: I agree that we're already bleeding internal state into OAuth2Util by exposing the settings object, but I'd prefer not to move even further down this route by passing in controller as well. Better to throw an error here, catch it in RouterUtil, and then call the controller stuff there.

alanzhou-okta: This error is a promise callback, so it doesn't actually bubble back to RouterUtil.

rchild-okta: Ah, I see. Yeah the ideal case would be that we actually return a promise here, and let RouterUtil handle it, but we can do that later since we'd have to fix the rest of this file at the same time. Okay, I'm okay with this change then.

test/e2e/specs/OIDC_spec.js
```diff
@@ -57,6 +57,20 @@ describe('OIDC flows', function() {
...
+        baseUrl: '{{{WIDGET_TEST_SERVER}}}',
+        clientId: 'rW47c465c1wc3MKzHznu',
+        redirectUri: 'http://localhost:3000/done',
+        authParams: {
+          responseType: 'id_token',
+          scope: ['openid', 'email', 'profile', 'address', 'phone']
+        }
+      });
+      primaryAuth.loginToForm('{{{WIDGET_BASIC_USER_5}}}', '{{{WIDGET_BASIC_PASSWORD_5}}}');
+      expect(primaryAuth.$dataSe('o-form-error-conainer').isPresent()).toBe(true);
```
rchild-okta: Hmm, so we need to add this user to Travis to actually test this scenario - I'm guessing that currently the error we're checking for is actually just the "You tried to submit a form without any credentials" type of error message. 

Also, as a side note, it'd be nice to add the $dataSe stuff to the page object rather than doing it here. Ideally at this level the logic would be on the level of "expect(primaryAuth.hasInvalidUserErrorMessage())" or something like that.

lboyette-okta: Just to be clear, `primaryAuth.hasInvalidUserErrorMessage()` should also check the text equals `"User is not assigned to the client application."`

src/util/OAuth2Util.js
```diff
@@ -31,14 +31,20 @@ define(['okta', './Enums', './Errors'], function (Okta, Enums, Errors) {
...
+  util.getTokens = function (settings, params, controller) {
 
     function success(result) {
       settings.callGlobalSuccess(Enums.SUCCESS, result);
     }
 
     function error(error) {
-      settings.callGlobalError(new Errors.OAuthError(error.message));
+      // OKTA-104330- Handle error case where user is not assigned to OIDC client
+      if (error.errorCode == 'access_denied') {
```
rchild-okta: nit: ===

test/e2e/specs/OIDC_spec.js
```diff
@@ -57,6 +57,24 @@ describe('OIDC flows', function() {
...
+        baseUrl: '{{{WIDGET_TEST_SERVER}}}',
+        clientId: 'rW47c465c1wc3MKzHznu',
+        redirectUri: 'http://localhost:3000/done',
+        authParams: {
+          responseType: 'id_token',
+          scope: ['openid', 'email', 'profile', 'address', 'phone']
+        }
+      });
+      primaryAuth.loginToForm('{{{WIDGET_BASIC_USER_5}}}', '{{{WIDGET_BASIC_PASSWORD_5}}}');
+      var element = primaryAuth.errorMessage();
```
rchild-okta: Let's go for this here:
```javascript
primaryAuth.loginToForm(stuff);
expect(primaryAuth.getErrorMessage()).toBe('User is not assigned etc');
```
And add `getErrorMessage` to the PrimaryAuth page, modeled after:
https://github.com/okta/okta-signin-widget/blob/master/test/e2e/page-objects/OIDCAppPage.js#L31

Where the main points are:
- We're putting the page specific logic in the page object
- We're using the protractor expected condition helpers

PR #134, opened 2016-10-28
:bug: Show proper error when trying to enroll or verify U2F factor

alexeystadnikov-okta:
Fix Widget to show proper error when trying to enroll or verify U2F.

screen shot 2016-10-28 at 4 26 54 pm
screen shot 2016-10-28 at 4 28 16 pm

Resolves: OKTA-104645, OKTA-99247, OKTA-99444, OKTA-99440, OKTA-104793

@mauriciocastillosilva-okta 
@anishashetty-okta 
@rchild-okta 

Please, review.


src/util/BrowserFeatures.js
```diff
@@ -58,6 +58,14 @@ define(['underscore'], function (_) {
     return /(msie|trident)/i.test(navigator.userAgent);
   };
 
+  fn.isFirefox = function () {
+    return navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
+  };
+
+  fn.isChrome = function () {
+    return /Chrome/.test(navigator.userAgent) && /Google Inc/.test(navigator.vendor);
+  };
```
mauriciocastillosilva-okta: @anishashetty-okta or @rchild-okta can you just double check that this is the best way to identify Chrome and Firefox?

src/EnrollU2FController.js
```diff
@@ -63,7 +65,7 @@ function (Okta, FormType, FormController, Footer, Q) {
           u2f.register(appId, registerRequests, [], function (data) {
             self.trigger('errors:clear');
             if (data.errorCode && data.errorCode !== 0) {
-              deferred.reject({ responseJSON: {errorSummary: 'Error Code: ' + data.errorCode}});
+              deferred.reject({xhr: {responseJSON: {errorSummary: 'Error Code: ' + data.errorCode}}});
```
mauriciocastillosilva-okta: I am guessing this will change once Vince provides the error messages, correct?

src/EnrollU2FController.js
```diff
@@ -85,21 +87,46 @@ function (Okta, FormType, FormController, Footer, Q) {
...
-        FormType.View({
-          View: '<div class="u2f-instructions"><ol>\
+      formChildren: function () {
+        var result = [];
+        var isSupportedBrowser = BrowserFeatures.isChrome() || BrowserFeatures.isFirefox();
+
+        if (!isSupportedBrowser) {
+          result.push(
+            FormType.View(
+              {View: new HtmlErrorMessageView({message: Okta.loc('enroll.u2f.error.notSupportedBrowser', 'login')})},
```
mauriciocastillosilva-okta: nit: can we add some line break in here to not have all in one line?

src/VerifyU2FController.js
```diff
@@ -46,8 +48,16 @@ function (Okta, FormController, FormType, FooterSignout, Q) {
             var deferred = Q.defer();
             u2f.sign(appId, factorData.challenge.nonce, registeredKeys, function (data) {
               self.trigger('errors:clear');
+              var errorSummary = 'Error Code: ' + data.errorCode;
               if (data.errorCode && data.errorCode !== 0) {
-                deferred.reject({ responseJSON: {errorSummary: 'Error Code: ' + data.errorCode}});
+                switch (data.errorCode) {
+                case 4:
```
mauriciocastillosilva-okta: it might be a good idea to move this ```case``` statement to a common place, since I assume it will be used in Enroll and in Verify

test/unit/spec/EnrollU2F_spec.js
```diff
@@ -65,10 +67,19 @@ function (Okta,
       });
     }
 
+    function mockChrome(isAvailable){
+      BrowserFeatures.isChrome = function () {};
```
mauriciocastillosilva-okta: Since we are spying on the function in the next line, I don't think we need to override it (function already exists in BrowserFeatures)
If I remember correctly, for ```winow.u2f``` I needed to do that since the function didn't exist

test/unit/helpers/dom/EnrollU2FForm.js
```diff
@@ -20,6 +20,10 @@ define(['./Form'], function (Form) {
 
     backLink: function () {
       return this.el('back-link');
+    },
+
+    errorHtml: function () {
```
mauriciocastillosilva-okta: this should be named ```hasErrorHtml```, since it returns a boolean

test/unit/spec/EnrollU2F_spec.js
```diff
@@ -103,7 +113,46 @@ function (Okta,
     });
 
     Expect.describe('Enroll factor', function () {
+      itp('shows error if wrong browser', function () {
+        mockChrome(false);
+        mockFirefox(false);
+
+        return setup().then(function (test) {
+          expect(test.form.errorHtml()).toHaveLength(1);
```
mauriciocastillosilva-okta: nit: might be a good idea to verify that the error message is the correct one, since we have different ones depending on the state
```
expect(test.form.errorHtml().indexOf('The Security Key is only supported for Chrome or Firefox browsers') > -1).toBe(true);
```

test/unit/spec/EnrollU2F_spec.js
```diff
@@ -103,7 +113,46 @@ function (Okta,
...
+        });
+      });
+
+      itp('shows error if Firefox without extension', function () {
+        mockChrome(false);
+        mockFirefox(true);
+        delete window.u2f;
+
+        return setup().then(function (test) {
+          expect(test.form.errorHtml()).toHaveLength(1);
```
mauriciocastillosilva-okta: similar to my previous nit, this could be
```
expect(test.form.errorHtml().indexOf('install the Firefox U2F browser extension before proceeding') > -1).toBe(true);
```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -170,15 +170,36 @@ function (Q,
...
+      BrowserFeatures.isFirefox = function () {};
+      spyOn(BrowserFeatures, 'isFirefox').and.returnValue(isAvailable);
+    }
 
     function setupU2F(options) {
-      if (!window.u2f || !window.u2f.sign) {
-        window.u2f = { sign: function () {} };
+      options || (options = {});
+
+      mockChrome(!options.noChrome);
```
mauriciocastillosilva-okta: can we make the inverse for the options params? It is easier to read
```!noChrome``` -> ```isChrome```
```!noFirefox ``` -> ```isFirefox ```
```!noU2f ``` -> ```hasU2f ```

File: src/util/BrowserFeatures.js (hunk 1 above)
Comment (rchild-okta) [inline src/util/BrowserFeatures.js]: Browser detection via userAgents is notoriously unreliable (see [here](http://webaim.org/blog/user-agent-string-history/) and the deprecated [jquery.browser](https://github.com/gabceb/jquery-browser-plugin/blob/master/dist/jquery.browser.js) plugin, which we've already removed from OSW).

Why can't we do feature detection?

File: src/VerifyU2FController.js [hunk 10]
```diff
@@ -69,25 +79,62 @@ function (Okta, FormController, FormType, FooterSignout, Q) {
...
+      },
       modelEvents: {
         'error': '_showRetry',
         'request': '_hideRetry'
       },
 
-      formChildren: [
-        FormType.View({
-          View: '\
+      formChildren: function () {
```
rchild-okta: I've noticed this before on these hello/u2f commits, but these files are almost all identical (and every PR I see shows more of the same logic being copy/pasted). 

When are we planning on doing this the right way, and moving most of this logic to shared views?

test/unit/spec/EnrollU2F_spec.js
rchild-okta: A bit pedantic, but I'd say that this is not a "nit" - we should write tests for code that we're adding.

src/util/BrowserFeatures.js
alexeystadnikov-okta: I have different mocks for the Firefox and other browsers.
In Chrome feature is available from the versions 38.
In Firefox feature is not available without the specific extension.
So, for Firefox I should check the browser, not feature.

I can refactor the code and remove the isChrome function.

src/EnrollU2FController.js
alexeystadnikov-okta: This change shows the valid error message instead of throwing TypeError when trying to read `responseJSON` from `undefined` inside the code.

src/EnrollU2FController.js
alexeystadnikov-okta: Sure

test/unit/spec/EnrollU2F_spec.js
alexeystadnikov-okta: Done

test/unit/spec/EnrollU2F_spec.js
alexeystadnikov-okta: Added tests for texts.

test/unit/spec/MfaVerify_spec.js
alexeystadnikov-okta: We have more tests with `hasU2f:true` than tests with `hasU2f:false`, so by default, it is better to have `noU2f:false` to avoid duplication of `hasU2f:true`

src/VerifyU2FController.js
alexeystadnikov-okta: They look similar, but not similar and I am not sure if a lot of code can be moved to the shared view.

src/EnrollU2FController.js
```diff
@@ -63,7 +65,13 @@ function (Okta, FormType, FormController, Footer, Q) {
           u2f.register(appId, registerRequests, [], function (data) {
             self.trigger('errors:clear');
             if (data.errorCode && data.errorCode !== 0) {
-              deferred.reject({ responseJSON: {errorSummary: 'Error Code: ' + data.errorCode}});
+              if (data.errorCode === 5) { // TIMEOUT
+                self.trigger('abort');
+                deferred.resolve(transaction);
+              }
+              else {
+                deferred.reject({xhr: {responseJSON: {errorSummary: 'Error Code: ' + data.errorCode}}});
```
mauriciocastillosilva-okta: this should be changed to something like
```
var errorMessage;
switch(data.errorCode) {
  case 1: errorMessage = Okta.loc('u2f.error.unknown');
...
responseJson: { errorSummary: errorMessage + '(' + data.errorCode + ')';
```
In that way we have the errors localized, but still we can see the error code at the end (it might happen that we have the same error message for codes 2-3, so we still need to identify them somehow)

src/VerifyU2FController.js
mauriciocastillosilva-okta: It actually can
From ybukey's website:

> The presented device is not eligible for this request. For a registration request this may mean that the token is already registered, and for a sign request it may mean that the token does not know the presented key handle.

src/VerifyU2FController.js
mauriciocastillosilva-okta: I created https://oktainc.atlassian.net/browse/OKTA-106499 to follow up on it. I think we can refactor some part of it. 
When we started I didn't realize they will share enough logic, so the design since the beginning was to have them independent. My bad.

src/VerifyU2FController.js
```diff
@@ -69,25 +84,60 @@ function (Okta, FormController, FormType, FooterSignout, Q) {
...
+        var result = [];
+
+        if (!window.hasOwnProperty('u2f')){
+          if(BrowserFeatures.isFirefox()){
```
mauriciocastillosilva-okta: nit: this might be easier to read if we do something like
```
var errorCode;
 if(BrowserFeatures.isFirefox()){
  errorCode = 'u2f.error.noFirefoxExtension';
} else if (this.options.appState.get('factors').length === 1) {
  errorCode='verify.u2f.error.notSupportedBrowser.oneFactor';
}
```
and then add the ```HtmlErrorMessageView```

test/unit/spec/MfaVerify_spec.js
mauriciocastillosilva-okta: I don't understand the duplication problem you are talking about... but I still think that it is easier to understand a simple boolean than a double negation
```if (options.u2f) {```vs ```if (!options.noU2f) {```
or 
```setupU2F({u2f: false, firefox: false})``` vs ```setupU2F({noU2f: true, noFirefox: true})```

File: test/unit/helpers/dom/EnrollU2FForm.js (hunk 6 above)
Comment (alexeystadnikov-okta) [inline test/unit/helpers/dom/EnrollU2FForm.js]: With last changes it returns the element, not boolean.

File: src/EnrollU2FController.js (hunk 11 above)
Comment (alexeystadnikov-okta) [inline src/EnrollU2FController.js]: This case, I need texts to localize them.

1 - OTHER_ERROR 
2 - BAD_REQUEST 
3 - CONFIGURATION_UNSUPPORTED
4 - DEVICE_INELIGIBLE 
5 - TIMEOUT 

Codes 1,2,3 depend on the configuration and can be localized with one common message.
For the Code 4 I have a key in the i18n, but don't have texts from @vincentvoong-okta 
For the Code 5 we can avoid showing the error and just reset the form.

File: src/VerifyU2FController.js (hunk 12 above)
Comment (alexeystadnikov-okta) [inline src/VerifyU2FController.js]: You are right.
Done.

**Comment by alexeystadnikov-okta** [comment]
Texts updated.

File: src/VerifyU2FController.js [hunk 13]
```diff
@@ -47,7 +64,9 @@ function (Okta, FormController, FormType, FooterSignout, Q) {
             u2f.sign(appId, factorData.challenge.nonce, registeredKeys, function (data) {
               self.trigger('errors:clear');
               if (data.errorCode && data.errorCode !== 0) {
-                deferred.reject({ responseJSON: {errorSummary: 'Error Code: ' + data.errorCode}});
+                var isOneFactor = self.options.appState.get('factors').length === 1;
+                deferred.reject({xhr: {responseJSON:
+                  {errorSummary: Okta.loc(getErrorMessageKeyByCode(data.errorCode, isOneFactor), 'login')}}});
```
mauriciocastillosilva-okta: nit: use one more line to match where ```deferred.reject``` was open
```
deferred.reject({xhr: {responseJSON:
  {errorSummary: Okta.loc(getErrorMessageKeyByCode(data.errorCode, isOneFactor), 'login')}
}});
```

test/unit/spec/EnrollU2F_spec.js
```diff
@@ -103,7 +123,46 @@ function (Okta,
     });
 
     Expect.describe('Enroll factor', function () {
+      itp('shows error if wrong browser', function () {
+        delete window.u2f;
+        mockFirefox(false);
+
+        return setup().then(function (test) {
+          expect(test.form.errorHtml()).toHaveLength(1);
+          expect(test.form.errorHtml().html()).toEqual(Okta.loc('u2f.error.notSupportedBrowser', 'login'));
```
mauriciocastillosilva-okta: We should check against the actual text instead of calling ```Okta.loc```. Reason is because if there is an error in ```Okta.loc``` or if someone messes up those keys, we will not know about it
Same applies to all the other places you are using ```Okta.loc``` in this test

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1939,6 +1955,28 @@ function (Q,
           });
         });
 
+        itp('shows error if wrong browser', function () {
+          return setupU2F({noU2f: true, noFirefox: true}).then(function (test) {
```
mauriciocastillosilva-okta: I still think this will be better as ```setupU2F({u2f: false, firefox: false})```.

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1939,6 +1955,28 @@ function (Q,
...
+            expect(test.form.el('o-form-error-html')).toHaveLength(1);
+            expect(test.form.el('o-form-error-html').find('strong').html())
+            .toEqual(Okta.loc('u2f.error.notSupportedBrowser', 'login'));
```
mauriciocastillosilva-okta: same comment as in the other test, use the actual string, not ```Okta.loc```

test/unit/spec/EnrollU2F_spec.js
alexeystadnikov-okta: If there is an error in Okta.loc, the tests for Okta.loc should fail.

I am not sure if it is a good idea when for changing texts we have to change not only the i18n, but also the tests.

alexeystadnikov-okta:
Spaces fixed.
`Okta.loc` removed.
`setupU2F({noU2f: true,` changed to `setupU2F({u2f: false,`

PR #137, opened 2016-10-31
:bug: Removes okta css from affecting single-app css

### LINKED ISSUE #126: Widget CSS conflicts with my app
I like that Okta provides this sign-in widget to help me get going, but the use of reset styles and other generic styles makes the provided CSS unusable

https://github.com/okta/okta-signin-widget/blob/okta-signin-widget-1.7.0/assets/sass/common/enduser/_reset.scss
https://github.com/okta/okta-signin-widget/blob/okta-signin-widget-1.7.0/assets/sass/widgets/_mega-drop-down.scss

Specifically, the global `input` styles and the generic `.dropdown` styles conflict with styles already built in my application, forcing me to completely style the Okta widget on my own. Something like this always needs to be scoped under an Okta class so as to not collide with everything else in your clients' custom apps.

alanzhou-okta:
Adds a ScssContainer so css only affects that container
Moves scss imports so they only affect the ScssContainer

Resolves #126

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -2,7 +2,7 @@
 
 scss_files: "assets/sass/**/*.scss"
 plugin_directories: ['.scss-linters']
-exclude: ["assets/sass/widgets/*.scss", "assets/sass/common/**/*.scss"]
+exclude: ["assets/sass/widgets/*.scss", "assets/sass/common/**/*.scss", "assets/sass/_container.scss", "assets/sass/okta-sign-in.scss"]
 
 # List of gem names to load custom linters from (make sure they are already
 # installed)
--- a/assets/sass/_base.scss
+++ b/assets/sass/_base.scss
@@ -1,20 +1,3 @@
-html {
-  height: 100%;
-}
-
-body {
-  font-family: $fonts;
-  font-size: $font-size;
-  font-weight: $font-weight-body;
-  line-height: 1.4;
-  background-color: $secondary-bg-color;
-  min-width: 100%;
-  min-height: 100%;
-  color: $medium-text-color;
-  position: relative;
-  overflow: auto;
-}
-
 a {
   color: $medium-text-color;
 
--- a/assets/sass/_container.scss
+++ b/assets/sass/_container.scss
@@ -1,0 +1,46 @@
+.auth-container {
+
+  // Html and body styles in _base.scss that were moved here
+  font-family: $fonts;
+  font-size: $font-size;
+  font-weight: $font-weight-body;
+  line-height: 1.4;
+  background-color: $secondary-bg-color;
+  min-height: 100%;
+  color: $medium-text-color;
+  position: relative;
+  overflow: auto;
+
+  // Auth-container styles
+  border-radius: 3px;
+  border-style: solid;
+  border-width: 1px;
+  height: auto;
+  margin: 100px auto 8px;
+  width: $container-width;
+  min-width: $container-min-width;
+  @include box-sizing(border-box);
+
+  @include device-mq(x-small) {
+    margin-top: 0;
+    width: 100%;
+  }
+
+  @include max-width-mq($container-width) {
+    width: 100%;
+  }
+
+  @include max-height-mq(750px) {
+    margin-top: 0;
+  }
+}
+
+.auth-container.no-beacon .auth-content {
+  padding-top: 20px;
+}
+
+.auth-container.no-beacon .auth-header {
+  padding-bottom: 30px;
+}
+
+
--- a/assets/sass/common/enduser/_reset.scss
+++ b/assets/sass/common/enduser/_reset.scss
@@ -1,56 +1,56 @@
 // Generated from okta-core sha: 900f683
 // Note: Do not modify this file directly.
 
-html, body, div, span, applet, object, iframe,
-h1, h2, h3, h4, h5, h6, p, blockquote, pre,
-a, abbr, acronym, address, big, cite, code,
-del, dfn, em, img, ins, kbd, q, s, samp,
-small, strike, strong, sub, sup, tt, var,
-b, u, i, center,
-dl, dt, dd, ol, ul, li,
-fieldset, form, label, legend,
-table, caption, tbody, tfoot, thead, tr, th, td {
-  margin: 0;
-  padding: 0;
-  border: 0;
-  outline: 0;
-  font-size: 100%;
-  font: inherit;
-  vertical-align: baseline;
-  background: transparent;
-}
-body {
-  line-height: 1;
-}
-ol, ul {
-  list-style: none;
-}
-blockquote, q {
-  quotes: none;
-}
-blockquote:before, blockquote:after,
-q:before, q:after {
-  content: '';
-  content: none;
-}
-:focus {
-  outline: 0;
-}
-ins {
-  text-decoration: none;
-}
-del {
-  text-decoration: line-through;
-}
-table {
-  border-collapse: collapse;
-  border-spacing: 0;
-}
-input[type=hidden] {
-  display: none !important;
-}
-input[type=checkbox], input[type=radio] {
-  border: 0 !important;
-  margin: 0;
-  padding: 0;
+
+.auth-container {
+  div, span, applet, object, iframe,
+  h1, h2, h3, h4, h5, h6, p, blockquote, pre,
+  a, abbr, acronym, address, big, cite, code,
+  del, dfn, em, img, ins, kbd, q, s, samp,
+  small, strike, strong, sub, sup, tt, var,
+  b, u, i, center,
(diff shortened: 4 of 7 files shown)


assets/sass/okta-sign-in.scss
```diff
@@ -1,32 +1,39 @@
...
-@import 'modules/header';
-@import 'modules/beacon';
-@import 'modules/factors-dropdown';
-@import 'modules/footer';
-@import 'modules/enroll';
-@import 'modules/app-login-banner';
-@import 'modules/okta-footer';
-@import 'ie';
+
+.auth-container-container {
```
rchild-okta: Oh nice, this works.

src/views/shared/ScssContainer.js
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define([
+    'okta'
+  ],
+  function (Okta) {
+    return Okta.View.extend({
+      className: 'auth-container-container',
```
rchild-okta: Do we need to create a new top level div? Ideally we would use our current #okta-sign-in one (which is already supposed to be the root div that we apply all styles to).

alanzhou-okta: So when we do nested scss, it only applies the css to the children elements of the div. So we have to do it to one layer above what we want.

rchild-okta: Do these styles apply to the original parent div?

alanzhou-okta: no, if we just nest it in #okta-sign-in the styles that we apply to that div don't apply.

alanzhou-okta:
Demo: with our current scss 
screen shot 2016-11-02 at 2 12 34 pm 2
As you can see, the other div on the page gets completely overriden by ours.

Updated: 
screen shot 2016-11-02 at 2 04 33 pm 2
The reset styles are only applied within the auth-container, means their css is not overridden.

assets/sass/_container.scss
```diff
@@ -0,0 +1,33 @@
+.auth-container {
```
rchild-okta: We should scope this to #okta-sign-in, i.e.:
```css
#okta-sign-in.auth-container {
}
```

As well as the other places that we have this. This is to answer the question of "what happens if they have their own .auth-container styles?"

assets/sass/_container.scss
```diff
@@ -0,0 +1,33 @@
...
+  @include max-width-mq($container-width) {
+    width: 100%;
+  }
+
+  @include max-height-mq(750px) {
+    margin-top: 0;
+  }
+}
+
+.no-beacon .auth-content {
```
rchild-okta: Does this live on the same level as auth-container? If so, we should also scope this with #okta-sign-in. If not, let's add it inside the body of the .auth-container class.

Actually in general, I am wondering why we need to set the styles to .auth-container - what happens if we just use #okta-signin-in instead?

assets/sass/common/enduser/_reset.scss
```diff
@@ -1,56 +1,59 @@
...
+    margin: 0;
+    padding: 0;
+    border: 0;
+    outline: 0;
+    font-size: 100%;
+    font: inherit;
+    vertical-align: baseline;
+    background: transparent;
+  }
+  body {
```
rchild-okta: So we shouldn't ever be in a position where body is underneath .auth-container - does this change how the widget looks if the line-height is not reset to 1? What happens if the outer page sets body line-height to something else? On our pages that we have multiple lines, does it look funny?

I'm guessing we should probably be putting line-height: 1 in the root element.

assets/sass/okta-sign-in.scss
```diff
@@ -1,32 +1,42 @@
 @import 'compass/css3';
 
 @import 'widgets/jquery.qtip';
```
rchild-okta: Where does this come from? And does it drop styles into the page or is it just a helper file?

assets/sass/okta-sign-in.scss
```diff
@@ -1,32 +1,42 @@
...
-@import 'widgets/mega-drop-down';
-@import 'widgets/chosen';
 
 @import 'common/enduser/reset';
```
rchild-okta: Does this file reset styles as well?

assets/sass/_container.scss
alanzhou-okta: Sass linter complains that we can't use id's

alanzhou-okta: So I haven't tried using a combination of id & class, but only id's will cause the linter to complain

assets/sass/common/enduser/_reset.scss
alanzhou-okta: What if we just delete it? I don't see a way to really keep the body css without having some kind of conflict with the outer page.

assets/sass/_container.scss
alanzhou-okta: Re: same level. Yes, no-beacon is another class of #okta-sign-in

assets/sass/okta-sign-in.scss
alanzhou-okta: This is the only reset style page

assets/sass/_container.scss
rchild-okta: In this case, we know we want to use an id - let's just disable that lint warning in this file and the other one where we'll use the id's

assets/sass/common/enduser/_reset.scss
rchild-okta: We need to remove it, but we still need the style applied somewhere - we should put it on .auth-container.

You should also update the sample you're running to apply a different line-height and verify that w/out our update, it looks different (and with it, it looks like it should)

assets/sass/okta-sign-in.scss
rchild-okta: Oh right, this is the one we're modifying

assets/sass/common/enduser/_reset.scss
alanzhou-okta: I moved all the stuff that would apply to the root elements (html and body) to _container.scss including some stuff from _base.scss. Line-height was overriden by line-height from _base.scss, but _container.scss should now contain all root scss.

.scss-lint.yml
```diff
@@ -2,7 +2,7 @@
 
 scss_files: "assets/sass/**/*.scss"
 plugin_directories: ['.scss-linters']
-exclude: ["assets/sass/widgets/*.scss", "assets/sass/common/**/*.scss"]
+exclude: ["assets/sass/widgets/*.scss", "assets/sass/common/**/*.scss", "assets/sass/_container.scss", "assets/sass/okta-sign-in.scss"]
```
rchild-okta: We shouldn't ignore the whole file - just the line itself, i.e. http://stackoverflow.com/questions/32833542/having-scss-lint-ignore-a-particular-line

assets/sass/_base.scss
```diff
@@ -1,20 +1,3 @@
-html {
```
rchild-okta: Can you try importing this into okta-core and making sure that this does not change the styling of the login page?

assets/sass/_container.scss
```diff
@@ -0,0 +1,48 @@
+#okta-sign-in.auth-container {
+
+  // Html and body styles in _base.scss that were moved here
+  height: 100%;
+  font-family: $fonts;
+  font-size: $font-size;
+  font-weight: $font-weight-body;
+  line-height: 1.4;
+  background-color: $secondary-bg-color;
+  min-width: 100%;
```
rchild-okta: We should remove the properties in this section that are being overridden below.

assets/sass/_container.scss
```diff
@@ -0,0 +1,48 @@
...
+  min-height: 100%;
+  color: $medium-text-color;
+  position: relative;
```
rchild-okta: Also I am wondering if this affects anything - if it doesn't change the appearance, we should remove it.

assets/sass/_container.scss
```diff
@@ -0,0 +1,48 @@
...
+  overflow: auto;
```
rchild-okta: We probably don't need this rule.

alanzhou-okta:
When we modularize our okta code to #okta-sign-in.auth-container, this makes the selector very specific. However, in okta-theme, we have okta specific selectors for #okta-sign-in, and these selectors are not overriding the generic selectors in #okta-sign-in.auth-container. 
There are many scss files that are being used outside the auth-container, and that means we have to import them to okta-core. The problem is there are some that currently may not be being used, but may be used in the future- do we import them or not?
A lot of double imports going on- importing almost exactly the same file in okta-core and okta-sign-in, except the okta-sign-in file has some selectors that are not applicable to the auth-container i.e. html, body.

PR #140, opened 2016-11-17
Upgrade to okta-auth-js 1.6.0

lboyette-okta:
This adds the ability to pass a custom state, nonce, authorizeUrl, and issuer

@rchild-okta 
@eugenebakaiev-okta

### DIFF
--- a/src/util/OAuth2Util.js
+++ b/src/util/OAuth2Util.js
@@ -55,13 +55,18 @@ define(['okta', './Enums', './Errors'], function (Okta, Enums, Errors) {
     _.extend(
       oauthParams,
       _.pick(options, 'clientId', 'redirectUri'),
-      _.pick(options.authParams, 'responseType', 'responseMode', 'display', 'scopes'),
+      _.pick(options.authParams, 'responseType', 'responseMode', 'display', 'scopes', 'state', 'nonce'),
       params
     );
 
     // Extra Options for Social Idp popup window title and id_token response timeout
     extraOptions.popupTitle = Okta.loc('socialauth.popup.title', 'login');
     extraOptions.timeout = options.oAuthTimeout;
+
+    _.extend(
+      extraOptions,
+      _.pick(options.authParams, 'issuer', 'authorizeUrl')
+    );
 
     // Redirect flow - this can be used when logging into an external IDP, or
     // converting the Okta sessionToken to an access_token, id_token, and/or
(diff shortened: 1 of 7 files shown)


src/util/OAuth2Util.js
```diff
@@ -55,14 +55,19 @@ define(['okta', './Enums', './Errors'], function (Okta, Enums, Errors) {
...
-      _.pick(options.authParams, 'responseType', 'responseMode', 'display', 'scopes'),
+      _.pick(options.authParams, 'responseType', 'responseMode', 'display', 'scopes', 'state', 'nonce'),
       params
     );
 
     // Extra Options for Social Idp popup window title and id_token response timeout
     extraOptions.popupTitle = Okta.loc('socialauth.popup.title', 'login');
     extraOptions.timeout = options.oAuthTimeout;
 
+    _.extend(
```
rchild-okta: Yeah it's a bit weird that we have to do this (in general having options and extra options) - I think you've already created a jira for it, but if not can you create one?

test/unit/spec/LoginRouter_spec.js
```diff
@@ -538,23 +538,26 @@ function (Okta, Q, Backbone, XDomain, SharedUtil, CryptoUtil, CookieUtil, OktaAu
...
-            expectedUrl += 'display=' + options.display + '&';
-          }
-          if (options.prompt) {
-            expectedUrl += 'prompt=' + options.prompt + '&';
-          }
-          expectedUrl += '' +
-            'sessionToken=THE_SESSION_TOKEN&' +
-            'scope=openid%20email';
-          expect(url).toBe(expectedUrl);
+        var authorizeUrl = options.authorizeUrl || 'https://foo.com/oauth2/v1/authorize';
```
rchild-okta: It might be nice to setup an e2e test for this as well; I remember you had mentioned looking into it - was there a reason to not also add that test?

src/util/OAuth2Util.js
lboyette-okta: Yup - OKTA-106994

test/unit/spec/LoginRouter_spec.js
lboyette-okta: I thought it'd be necessary, but it wasn't. However, as we're trying to add more coverage, here's a ticket to add tests in a future pr: OKTA-108155

PR #144, opened 2016-11-18
upgrade courage to 1.11.0, a lean courage

yiyangli-okta:


yiyangli-okta:
sorry, I didn't completely follow the rule, I should include the ticket in the branch name. It's related to [OKTA-108218](https://oktainc.atlassian.net/browse/OKTA-108218), let me know if I need to change the branch name and also another PR for the new branch.

rchild-okta:
lgtm! It does look like there is a conflict that needs to be resolved, and you'll probably need to re-push since it looks like this is not running against Travis yet for some reason. But after passing the tests, let me know and I can merge it.

rchild-okta:
Also, we now have a doc discussing the [contributing guidelines](https://github.com/okta/okta-signin-widget/blob/master/CONTRIBUTING.md) - this is mostly related to the commit message where you should add an emoji for the type of commit you're making (in this cased, we would use `:arrow_up:` because we're upgrading a dependency).

yiyangli-okta:
interesting! The Emoji Is the Birth of a New Type of Language

yiyangli-okta:
@rchild-okta the conflict is resolved and tests are passed, could you merge it?

PR #148, opened 2016-12-07
:bug: Conditionally use local storage

stsai-okta:
Do not use local storage for SP initiated flows on non-english embedded Android Browsers

Resolves: OKTA-104858
Bacon: test

### DIFF
--- a/src/util/Bundles.js
+++ b/src/util/Bundles.js
@@ -20,8 +20,9 @@ define([
   'json!nls/login',
   'json!nls/country',
   'util/Logger',
-  'json!config/config'
-], function (_, Q, $, login, country, Logger, config) {
+  'json!config/config',
+  'util/BrowserFeatures'
+], function (_, Q, $, login, country, Logger, config, BrowserFeatures) {
 
   var STORAGE_KEY = 'osw.languages';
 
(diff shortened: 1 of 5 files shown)


jhe-okta:
Looks good to me. ⛩ 
A couple questions:
- what's the difference in local storage and cache? 
- do you know why it local storage is missing from android web view?

src/util/Bundles.js
```diff
@@ -160,17 +161,24 @@ define([
       return Q({});
     }
 
-    var cached = getCachedLanguages();
-    if (cached[language]) {
-      return Q(cached[language]);
+    //local storage does not work correctly with android web views (embeded browers)
+    //so skip the caching and just make the request to get the local info
+    var isAndroidWebView = BrowserFeatures.isAndroidWebView();
```
stsai-okta: fixed

stsai-okta:
where are you seeing that we're using a cache? In this example, we're using local storage as a cache. It looks like local storage has to be enabled by the android app in order for it to work and since this is an SP initiated flow for an app we don't own, we can't set local storage to be true: http://stackoverflow.com/questions/5899087/android-webview-localstorage/5934650#5934650

PR #149, opened 2016-12-12
:arrow_up: bump up version of courage to 1.14.0

stsai-okta:
Resolves: OKTA-110389
Bacon: test

### DIFF
--- a/node_modules/@okta/i18n/package.json
+++ b/node_modules/@okta/i18n/package.json
@@ -1,16 +1,8 @@
 {
   "_args": [
     [
-      {
-        "raw": "@okta/i18n@19.76.0",
-        "scope": "@okta",
-        "escapedName": "@okta%2fi18n",
-        "name": "@okta/i18n",
-        "rawSpec": "19.76.0",
-        "spec": "19.76.0",
-        "type": "version"
-      },
-      "/Users/lboyette/okta/okta-signin-widget"
+      "@okta/i18n@19.76.0",
+      "/Users/stephen.tsai/okta/okta-signin-widget"
     ]
   ],
   "_from": "@okta/i18n@19.76.0",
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -69,6 +69,7 @@
     "axe-webdriverjs": "0.4.0"
   },
   "dependencies": {
+    "@okta/courage": "1.14.0",
```
yiyangli-okta: you don't want to put it under dependencies, because people out of Okta can't access `@okta`'s artifacts, so, undo this line, I guess.

stsai-okta: oh but that was what was generated by install-with-shrinkwrap. Should I manually update the optional-dependencies then?

yiyangli-okta: I think so, but if the developer actually clone the repository, npm install would be good as well since we have put one under `node_modules`, I don't know why `optionalDependencies` though.

mauriciocastillosilva-okta: yeah, this should not be here.
There is already an entry for courage in ```optionalDependencies``` and that is the one we need to update

stsai-okta: okay will do

node_modules/@okta/i18n/package.json
```diff
@@ -1,16 +1,8 @@
...
-        "scope": "@okta",
-        "escapedName": "@okta%2fi18n",
-        "name": "@okta/i18n",
-        "rawSpec": "19.76.0",
-        "spec": "19.76.0",
-        "type": "version"
-      },
-      "/Users/lboyette/okta/okta-signin-widget"
+      "@okta/i18n@19.76.0",
+      "/Users/stephen.tsai/okta/okta-signin-widget"
```
mauriciocastillosilva-okta: which version of npm do you have? just wondering why the structure changed here.

stsai-okta: 3.9.1

PR #154, opened 2016-12-15
:bug: Add title attribute to iframes for Duo factor.

alexeystadnikov-okta:
screen shot 2016-12-15 at 5 31 56 pm

Resolves OKTA-110561
508 compliance: need title on frames - okta-signin-widget

@mauriciocastillosilva-okta 
@lboyette-okta 
Please, review.


hor-kanchan-okta:
Can you add a test for it? Check if Iframe has non-empty title attribute.

alexeystadnikov-okta:
@hor-kanchan-okta 
Tests added.

PR #155, opened 2016-12-19
:bug: Change type from ‘text’ to ‘number’ in textboxes for number vaues.

alexeystadnikov-okta:
Resolves OKTA-79239.

Android:
176bda40-c6db-11e6-9c50-24c8aff98a55
178a3b8e-c6db-11e6-9f8f-0bf67fcb2e5b

iPhone:

iPad:

Alerts were taken by the code in `TextBox:postRender`:
```
input.blur(function () {
          alert(this.value);
        });
```

@mauriciocastillosilva-okta 
@lboyette-okta 

Please, review.

### DIFF
--- a/src/EnrollCallAndSmsController.js
+++ b/src/EnrollCallAndSmsController.js
@@ -249,7 +249,7 @@ function (Okta, FormController, Footer, PhoneTextBox, TextBox, CountryUtil, Form
           placeholder: Okta.loc('mfa.challenge.enterCode.placeholder', 'login'),
           name: 'passCode',
           input: TextBox,
-          type: 'text',
+          type: 'number',
           params: {
             innerTooltip: Okta.loc('mfa.challenge.enterCode.tooltip', 'login')
           },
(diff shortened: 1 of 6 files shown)


marksztainbok-okta:
Looks like just changing the type number allows floating point values and does not actually display the correct keyboard.
http://stackoverflow.com/questions/6178556/phone-numeric-keyboard-for-text-input/31619311#31619311 has a better solution

mauriciocastillosilva-okta:
Can you attach a screenshot? Changes look good to me, but that post that Mark shared says that it might not work.

alexeystadnikov-okta:
Screenshots added.

Also added attributes `pattern` and `inputmode` to the input when `TextBox` has type 'number'.

marksztainbok-okta:
Have you checked on both Android and iOS? Also does a MFA code like 012345 work? From what I've read there may be issues with leading 0's potentially

alexeystadnikov-okta:
Checked with Android and iOS devices (screenshots added).
`01234` works fine. It removes first zero only on Mac/PC if use the up/down arrows.

PR #161, opened 2017-01-12
:bug: Fix checkbox css. Correct the checkbox image height.

hor-kanchan-okta:
Resolves OKTA-112865

Default
default

When focus
focus

When on hover
onhover

When checked but no focus and hover
checked

When checked and on hover
checked and onhover

### DIFF
--- a/assets/sass/common/shared/o-forms/_o-form.scss
+++ b/assets/sass/common/shared/o-forms/_o-form.scss
@@ -444,7 +444,7 @@ only screen and (min-resolution: 2dppx) {
   .custom-checkbox {
     label {
       background: url('../img/ui/forms/checkbox-01@2x.png') no-repeat;
-      background-size: 50px 1055px;
+      background-size: 50px 1155px;
     }
   }
 }


hor-kanchan-okta:
The height should be the image height of assets/img/ui/forms/checkbox-01@2x.png divided by 2.

assets/sass/common/shared/o-forms/_o-form.scss
```diff
@@ -444,7 +444,7 @@ only screen and (min-resolution: 2dppx) {
   .custom-checkbox {
     label {
       background: url('../img/ui/forms/checkbox-01@2x.png') no-repeat;
-      background-size: 50px 1055px;
+      background-size: 50px 1155px;
```
rchild-okta: Double checking - are you going to port this fix to courage as well?

hor-kanchan-okta: No. The courage doesn't have this problem. The checkbox-01@2x.png is not updated on Courage.

rchild-okta: Oh, hmmm, that's confusing - we should probably port that image back to courage as well (doesn't need to be now). Main concern is that at some point we're going to update assets/sass from courage, and we'll just overwrite this change and have a regression.

Can you move this css to one of the widget-specific sass files then? i.e. https://github.com/okta/okta-signin-widget/blob/master/assets/sass/modules/_forms.scss

hor-kanchan-okta: Understood. Will move the checkbox images as well as the related css to widget specific location in a separated PR.

PR #163, opened 2017-01-16
:bug: Fix switching to DUO factor from SMS.

alexeystadnikov-okta:
Fix switching to DUO factor from SMS.
Move Duo.init to postRender.
Resolves: OKTA-100438

ifewhi6wwv

@rchild-okta 
@ukilon-okta
@mauriciocastillosilva-okta 
Please, review.

### DIFF
--- a/src/views/mfa-verify/dropdown/FactorsDropDownOptions.js
+++ b/src/views/mfa-verify/dropdown/FactorsDropDownOptions.js
@@ -24,8 +24,8 @@ define(['okta', 'util/RouterUtil'], function (Okta, RouterUtil) {
         return transaction.prev()
         .then(function (trans) {
           self.trigger('options:toggle');
-          self.options.appState.trigger('navigate', url);
           setTransaction(trans);
+          self.options.appState.trigger('navigate', url);
         });
       } else {
         self.trigger('options:toggle');
--- a/test/unit/spec/MfaVerify_spec.js
+++ b/test/unit/spec/MfaVerify_spec.js
@@ -2321,6 +2321,23 @@ function (Okta,
           });
         });
       });
+      itp('Verify DUO after switching from SMS MFA_CHALLENGE', function () {
+        return setupSMS().then(function (test) {
+          test.setNextResponse(resChallengeSms);
+          test.form.smsSendCode().click();
+          return tick(test);
+        })
+        .then(function (test) {
+          spyOn(Duo, 'init');
+          test.setNextResponse([resAllFactors, resChallengeDuo]);
+          test.beacon.dropDownButton().click();
+          test.beacon.getOptionsLinks().eq(4).click();
+          return Expect.waitForVerifyDuo(test);
+        })
+        .then(function () {
+          expect(Duo.init).toHaveBeenCalled();
+        });
+      });
     });
 
   });


rchild-okta:
@alexeystadnikov-okta, can you explain this change? Specifically, why is fetchInitialData not being called in this flow? Or is it a timing issue?

alexeystadnikov-okta:
Sending SMS changes transaction to new one with status "MFA_CHALLENGE".
After changing the factor it calls `previous` to have "MFA_REQUIRED".

`fetchInitialData` calls `this.model.getInitOptions()` which calls `factor.verify(data)`.

After changing the factor in FactorsDropDownOptions.js it calls
```
 self.options.appState.trigger('navigate', url);
 setTransaction(trans);
```
`navigate` events creates the controller and calls `fetchInitialData` of it, but transaction is old (MFA_CHALLENGE) and doesn't have needed factor, so `factor.verify` throws error.

rchild-okta:
I'm forgetting the timing (you should look into it), but what happens if we switch the order in the lines you pasted, i.e.

```javascript
setTransaction(trans);
self.options.appState.trigger('navigate', url);
```

Does setting the transaction first trigger a new page render (before we call the navigate), which would cause bad UI rendering?

alexeystadnikov-okta:
It works fine with DUO, Question and SMS factors.

Although, it checks `trapAuthResponse` for the previous controller, i.e. when we change SMS to DUO, it checks `MfaVerifyController.trapAuthResponse` and switch to DUO.
Potentially it can create the situation of rendering two controllers one by one (`setTransation` + `navigate`) if `trapAuthResponse` will not stop the `RouterUtil.routeAfterAuthStatusChange` function.

rchild-okta:
Okay, maybe this is an easier fix then (and more in line with how things are setup - i.e. not sure there was a reason for the original ordering). Mostly just need to verify that last point, i.e. we don't get into a situation where trapAuthResponse does not handle something it should.

alexeystadnikov-okta:
Done.
Reverted changes in `VerifyDuoController.js`,
changed lines in `FactorsDropDownOptions.js`
from 
```
 self.options.appState.trigger('navigate', url);
 setTransaction(trans);
```
to
```
setTransaction(trans);
self.options.appState.trigger('navigate', url);
```

rchild-okta:
Can you add a test for this case? I.e. sending the code in SMS and then switching to Duo, and not getting an error.

alexeystadnikov-okta:
@rchild-okta 
I am not sure if it is possible to do with unit tests. Too many components should be initialized for this purpose.

Also, I have found that the same error is triggered if we change the factor with the browser history.
In addition, widget stops working if we switch factor after `factor.verify()` was called (push sent, DUO factor simply opened, a wrong answer on security question, sms code sent, etc.) because changing factor with the browser history doesn't send `factor.prev()` request.
@rchild-okta  Is it a known issue?

rchild-okta:
@alexeystadnikov-okta check out the tests we have currently - they are actually all structured like this. They're not really "unit" tests per se - more functional JS that happen to use jasmine as the runner.

One good one to look at would be [this one](https://github.com/okta/okta-signin-widget/blob/master/test/unit/spec/MfaVerify_spec.js#L2159) (and the other tests in this file).

What do you mean by "change the factor with the browser history"? I.e. are you talking about clicking the back button, or are you saying this doesn't work if we use javascript to call the history object?

alexeystadnikov-okta:
@rchild-okta
Thanks for the example.

I mean browser back button.

rchild-okta:
Interesting - that does sound like a regression if it doesn't work. We have some code here that is supposed to handle the back button:
https://github.com/okta/okta-signin-widget/blob/master/src/util/BaseLoginRouter.js#L266-L278

@hor-kanchan-okta, any thoughts?

alexeystadnikov-okta:
@rchild-okta 
Test added.

hor-kanchan-okta:
@rchild-okta 
I wonder if we set the transition object correct. If we don't, we will mess up the state.
I don't think we ever test this flow.
@alexeystadnikov-okta Could you please file a bug?

alexeystadnikov-okta:
@hor-kanchan-okta 
Created https://oktainc.atlassian.net/browse/OKTA-114121

@rchild-okta 
Do you have any other comments?

alexeystadnikov-okta:
@hor-kanchan-okta 
@rchild-okta 
Do you have any other comments?

rchild-okta:
Just one to double check - can you run this topic build through our okta-core tests to make sure all the enduser login selenium tests pass?

If everything passes, go ahead and 🚢

mauriciocastillosilva-okta:
@rchild-okta how can we do that? I don't think we promote any artifact from topic branches.

rchild-okta:
@mauriciocastillosilva-okta ah, you're right. We don't run topic branches in bacon for OSS repos. In this case, let's use [npm pack](https://docs.npmjs.com/cli/pack) and then run `npm install {{that tarball}}` in okta-core

mauriciocastillosilva-okta:
Will that work running in bacon? Or is that just to install that tarball package locally? I guess to make that work we need to change some other things in package.json and something else...

rchild-okta:
Running enduser tests locally. To run it in bacon - looks like you can point to a tarball in package.json (i.e. check it in and point to it in package.json), else could just check node_modules into source and do it that way (and remove it from package.json so it's not overwritten in the npm install).

alexeystadnikov-okta:
@rchild-okta 
I have run tests locally and they passed.

mauriciocastillosilva-okta:
This is ready to merge, but will be on hold for now until further notice.

alexeystadnikov-okta:
Rebased

PR #164, opened 2017-01-17
Move widget specific checkbox css and images to widget only location

alexeystadnikov-okta:
Resolve OKTA-112900

- Moved changes for OKTA-104129 from `shared/o-forms/_o-form.scss` to `modules/_forms.scss`
- Changed image filename to avoid the conflict.

@hor-kanchan-okta 
@rchild-okta 
@mauriciocastillosilva-okta 
Please, review.


assets/sass/modules/_forms.scss
```diff
@@ -247,3 +247,26 @@
 .chzn-container-active.chzn-with-drop .chzn-single div b {
   background-position: -17px 7px;
 }
+
+.custom-checkbox {
+  label {
+    background-image: url('/img/ui/forms/checkbox.png');
```
rchild-okta: nit: I think we should make it a little more obvious that this is a widget specific file, just in case (for whatever reason) someone creates a checkbox.png in courage and it gets overwritten in a future update. Probably something like checkbox-sign-in-widget.png is fine.

alexeystadnikov-okta:
`checkbox.png` renamed to `checkbox-sign-in-widget.png`
Lint errors fixed.

PR #170, opened 2017-02-08
:bug: Remove posibility of double factor.verify() request in TOTP form.

alexeystadnikov-okta:
Removes possibility to send multiple factor.verify() requests for token factors.
Resolves: OKTA-89133

It was possible to click the submit button while factor.verify() is in the process.
After submitting the form `disabled` property of button was not changed only button class was changed. It was possible to click on it even while it looks like disabled by having the `link-button-disabled` class.

@magizh-okta 
@mauriciocastillosilva-okta 
@rchild-okta 

Please, review.


test/unit/helpers/mocks/Util.js
```diff
@@ -91,18 +91,18 @@ function ($, _, Backbone, Q, Duo) {
...
+        } else {
+          // $.ajax send (jqXHR, textStatus, errorThrown) on failure
+          if (!textOnly) {
+            xhr.responseJSON = xhr.response;
+          }
+          deferred.reject(xhr, null, xhr.response);
         }
-        deferred.reject(xhr, null, xhr.response);
-      }
+      }, xhr.delay || 0);
```
alexeystadnikov-okta: Have to use `xhr.delay` for `deferred.reject` too.

test/unit/spec/MfaVerify_spec.js
```diff
@@ -730,12 +730,40 @@ function (Okta,
...
           })
           .then(function (test) {
             var button = test.form.submitButton();
             var buttonClass = button.attr('class');
+            expect($.ajax.calls.count()).toBe(1);
             expect(buttonClass).not.toContain('link-button-disabled');
+            expect(button.prop('disabled')).toBe(false);
+          });
+        });
+        itp('makes only one request if "verify button" was clicked twice', function () {
```
mauriciocastillosilva-okta: does this test make sense? If we are already disabling the button, not sure we need to test this. Or maybe I am missing something, what is that we are trying to test here?

test/unit/helpers/mocks/Util.js
```diff
@@ -91,18 +91,22 @@ function ($, _, Backbone, Q, Duo) {
       }
 
       var deferred = $.Deferred();
-      if (xhr.status > 0 && xhr.status < 300) {
-        // $.ajax send (data, textStatus, jqXHR) on success
+
+      (function (textOnly) {
```
mauriciocastillosilva-okta: can you please enlighten me on why we need this?

test/unit/spec/MfaVerify_spec.js
alexeystadnikov-okta: Disabling a button is a fix for a problem described in the issue.
This test checks that the problem is fixed to prevent regression.

test/unit/helpers/mocks/Util.js
alexeystadnikov-okta: The function inside the `setTimeout` is called async after the current.
Without the closure function, `textOnly` will always be false.

mauriciocastillosilva-okta: Right, but what was the problem with the previous code? I see the difference is that no matter with ```xhr.status``` we have, now all that code is inside a ```setTimeout```. Why do we need to change that?

alexeystadnikov-okta: The issue was a possibility to create the factor.verify() ajax request while one is already running.
Without making mocks for ajax asynchronous for the error results, not only for success, it was impossible to test such situation.

test/unit/spec/MfaVerify_spec.js
mauriciocastillosilva-okta: but how are you clicking the ```submit``` button for the second time? It is disabled at that time. And the form has errors since the first time you clicked the submit button, right?

test/unit/helpers/mocks/Util.js
mauriciocastillosilva-okta: got it

test/unit/spec/MfaVerify_spec.js
alexeystadnikov-okta: `test.setNextResponse([_.extend({delay: 500}, resInvalid), resInvalid]);`
I use the `delay`, so the button is disabled (before the changes it was not disabled, just has class `link-button-disabled`).
And while the request is in progress I click the button the second time.

I can remove this test if needed and left only tests for the `disabled` property.
This test is showing the problem from the issue.

alexeystadnikov-okta:
@mauriciocastillosilva-okta 
Unneeded test removed.

alexeystadnikov-okta:
Squashed and rebased.

PR #171, opened 2017-02-09
:memo: Add metadata to package.json for D.O.C linkage

joelfranusic-okta:
Add metadata to `package.json` which will enable scripts to transclude documentation from GitHub repositories into developer.okta.comx

Resolves: OKTA-115322
Bacon: test

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,21 @@
     "ci-pkginfo:dataload": "ci-pkginfo -t dataload",
     "installsw": "install-with-shrinkwrap"
   },
+  "okta": {
+    "doc": [
+      {
+        "source": "README.md",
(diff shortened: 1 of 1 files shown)


joelfranusic-okta:
@lboyette-okta - this is my proposal for adding metadata that will allow documentation to be automatically transcluded into developer.okta.com.

Of particular interest is the structure of the JSON that I'm adding. I'm not married to the naming of the `okta` or `doc` fields (my current script uses `oktaDoc`) though I should point out that this structure [might imply that the `okta-auth-js` section in okta-auth-js/package.json](https://github.com/okta/okta-auth-js/blob/master/package.json#L60-L72) might need to be renamed?

PR #174, opened 2017-02-13
:bug: Changed validation message for username and password

eugenebakaiev-okta:
:bug: Changed validation message for username and password

added validation with custom validation messages to username and password fields in PrimaryAuth model

Resolves: OKTA-114899


eugenebakaiev-okta:
@magizh-okta, please take a look

src/models/PrimaryAuth.js
```diff
@@ -37,9 +37,24 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
       }
 
       return {
-        username: ['string', true, username],
+        username: {
+          type: 'string',
+          validate: function (value) {
+            if(!value || value.length < 1) {
+              return 'Please enter a username';
```
magizh-okta: i18n ? also since required is set to true is there a way to set an error message for that ? can u screenshots. I just want to make sure there is only one error message displayed.

eugenebakaiev-okta: errormessage mov


I'll add i18n soon. 

> also since required is set to true is there a way to set an error message for that ?

I haven't found a way to do that.

magizh-okta: If there is no way to just customize message to get to that use same validate function like courage.  /cc @ukilon-okta 
    if (field.required && (!isDefined || _.isNull(value) || value === '')) {

src/models/PrimaryAuth.js
```diff
@@ -37,9 +37,24 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
       }
 
       return {
-        username: ['string', true, username],
+        username: {
+          type: 'string',
+          validate: function (value) {
+            if(!value || value.length < 1) {
```
yiyangli-okta:
```
var value = '123';
console.log(value); //return 123
value.length = 0;
console.log(value); //still return 123
```

use `_.isEmpty` instead.

src/models/PrimaryAuth.js
yiyangli-okta: I think `validate` is good, please change to use i18n, and also add a test to verify `new Model({username: ''}).validate()` will return `{'username' :  '{{your message}}'}`.

src/models/PrimaryAuth.js
ukilon-okta: yay for good oll javascript `if(!value){`

eugenebakaiev-okta:
@yiyangli-okta, @ukilon-okta, @gowthamidommety-okta look please

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -178,6 +179,21 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
 
   Expect.describe('PrimaryAuth', function () {
 
+    Expect.describe('PrimaryAuthModel', function () {
+
+      it('returns username validation error when username is blank', function () {
+        var model = new PrimaryAuth({username: '', password: 'pass'});
+        expect(model.validate().username).toBeDefined();
```
gowthamidommety-okta: Since you are testing for .toEqual(), do you need .toBeDefined()?

eugenebakaiev-okta:
@yiyangli-okta @ukilon-okta @gowthamidommety-okta updated, look again please

yiyangli-okta:
lgtm 🚢

PR #175, opened 2017-02-16
:bug: Hide Sign Out link during MFA based on settings

mauriciocastillosilva-okta:
Settings new prop: hideSignOutLinkInMFA
This is used for App Sign in Flows, when we only need to do MFA. It does not make sense to have a SignOut link in that flow (that sign out will not kill the user session, but only the app session we are trying to login in too. Plus it redirects to the main login form, which also does not make sense)

Resolves: OKTA-114022
Bacon: test

### DIFF
--- a/README.md
+++ b/README.md
@@ -612,9 +612,11 @@ var signIn = new OktaSignIn(config);
   - `fi` - Finnish
   - `fr` - French
   - `hu` - Hungarian
+  - `id` - Indonesian
   - `it` - Italian
   - `ja` - Japanese
   - `ko` - Korean
+  - `ms` - Malaysian
   - `nl-NL` - Dutch
   - `pt-BR` - Portuguese (Brazil)
   - `ro` - Romanian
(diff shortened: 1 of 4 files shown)


rchild-okta:
Do we need a flag or can we just hide it from MFA for all users? I.e. if it doesn't make sense, why even make it an option?

Btw, if you want to add another setting to the widget, make sure that you update the README as well to document it.

mauriciocastillosilva-okta:
I'll update the README, thanks for reminding me (note to myself: I need to keep this in mind for every PR that adds a setting or something that needs doc)

It does not make sense in App MFA, but not in MFA in general... that is why I figured adding the option on the settings was the best way, so it is up to the consumer to set that flag if the sign out link is not needed.

rchild-okta:
So main question I have is  if there's another way to know this information without having to pass the flag in explicitly - i.e. for example when bootstrapping the widget with the per-app stateToken, can it pass us back a signal (or is it already somehow different) that we're doing a step up flow rather than primary auth.

mauriciocastillosilva-okta:
I don't know if we can do it without the flag... the stateToken does not have that info as far as I know, and just by going to the refresh state will not work, since other flows could get there (for example, user settings page to verify password and MFA).

rchild-okta:
Makes sense. Can you talk to the auth team and see if they're open to the idea of giving us context back on refreshToken requests?

mauriciocastillosilva-okta:
@srinivasanagandla-okta take a look at Reman's comment

README.md
```diff
@@ -867,6 +870,14 @@ Options for the [OpenId Connect](http://developer.okta.com/docs/api/resources/oi
     recoveryToken: 'x0whAcR02i0leKtWMZVc'
     ```
 
+## MFA Challenge flow without sign out link
```
rchild-okta: Hmm, good thing we added the doc - looking at this now, would prefer to add this under our feature settings rather than a top level config item.

PR #179, opened 2017-02-23
:seedling: Add async credentials processing

AlainODea:
Allow credentials processing to be asynchronous. This allows the
credentials processing function to do things like AJAX calls if
necessary.

 - Add asyncProcessCreds function to settings

 - Support asyncProcessCreds in Primary Auth Form

 - Preserve device fingerprinting as part of the delegate callback

### DIFF
--- a/src/views/primary-auth/PrimaryAuthForm.js
+++ b/src/views/primary-auth/PrimaryAuthForm.js
@@ -37,27 +37,37 @@ define([
 
     initialize: function () {
       this.listenTo(this, 'save', function () {
+        var self = this;
+        var saveActions = function() {
+          if (self.settings.get('features.deviceFingerprinting')) {
+            DeviceFingerprint.generateDeviceFingerprint(self.settings.get('baseUrl'), self.$el)
+              .then(function (fingerprint) {
+                self.model.set('deviceFingerprint', fingerprint);
+              })
+              .fail(function () {
+                // Keep going even if device fingerprint fails
+              })
+              .fin(function () {
+                self.model.save();
+              });
+          } else {
+            self.model.save();
+          }
+        };
         var processCreds = this.settings.get('processCreds');
-        if (_.isFunction(processCreds)) {
-          processCreds({
+        if (!_.isFunction(processCreds)) {
+          saveActions();
+        } else {
+          var creds = {
             username: this.model.get('username'),
             password: this.model.get('password')
-          });
-        }
-        if (this.settings.get('features.deviceFingerprinting')) {
-          var self = this;
-          DeviceFingerprint.generateDeviceFingerprint(this.settings.get('baseUrl'), this.$el)
-          .then(function (fingerprint) {
-            self.model.set('deviceFingerprint', fingerprint);
-          })
-          .fail(function () {
-            // Keep going even if device fingerprint fails
-          })
-          .fin(function () {
-            self.model.save();
-          });
-        } else {
-          this.model.save();
+          };
+          if (processCreds.length === 2) {
+            processCreds(creds, saveActions);
+          } else {
+            processCreds(creds);
+            saveActions();
+          }
         }
       });
       this.listenTo(this.state, 'change:enabled', function (model, enable) {
--- a/README.md
+++ b/README.md
@@ -568,7 +568,9 @@ var signIn = new OktaSignIn(config);
     }
     ```
 
-- **processCreds:** Synchronous hook to handle the credentials before they are sent to Okta in the Primary Auth, Password Expiration, and Password Reset flows.
+- **processCreds:** Hook to handle the credentials before they are sent to Okta in the Primary Auth, Password Expiration, and Password Reset flows.
+
+    If processCreds takes a single argument it will be executed as a synchonous hook:
 
     ```javascript
     // Passed a creds object {username, password}
@@ -579,6 +581,31 @@ var signIn = new OktaSignIn(config);
         user: creds.username,
         passwordBytes: creds.password,
         keyType: 'KEY_TYPE_PASSWORD_PLAIN'
+      });
+    }
+    ```
+
+    If processCreds takes two arguments it will be executed as an asynchonous hook:
+
+    ```javascript
+    // Passed a creds object {username, password} and a callback for further processing
+    processCreds: function (creds, callback) {
+      // This example demonstrates optional legacy form-based logon
+      $.ajax({
+        method: "POST",
+        url: "/logontype",
+        data: {
+          username : creds.username
+        },
+        success: function (logontype) {
+          if (logontype == "LEGACY") {
+            $('#legacyUser').val(creds.username);
+            $('#legacyPassword').val(creds.password);
+            $('#legacyLogonForm').submit();
+          } else {
+            callback();
+          }
+        }
       });
     }
     ```
--- a/src/PasswordExpiredController.js
+++ b/src/PasswordExpiredController.js
@@ -151,13 +151,20 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, FactorUtil, Tex
     initialize: function () {
       this.listenTo(this.form, 'save', function () {
         var processCreds = this.settings.get('processCreds');
-        if (_.isFunction(processCreds)) {
-          processCreds({
+        if (!_.isFunction(processCreds)) {
+          this.model.save();
+        } else {
+          var creds = {
             username: this.options.appState.get('userEmail'),
             password: this.model.get('newPassword')
-          });
+          };
+          if (processCreds.length === 2) {
+            processCreds(creds, _.bind(this.model.save, this.model));
+          } else {
+            processCreds(creds);
+            this.model.save();
+          }
         }
-        this.model.save();
       });
     }
 
--- a/src/PasswordResetController.js
+++ b/src/PasswordResetController.js
@@ -84,13 +84,20 @@ function (Okta, FormController, FormType, ValidationUtil, FactorUtil, FooterSign
     initialize: function () {
       this.listenTo(this.form, 'save', function () {
         var processCreds = this.settings.get('processCreds');
-        if (_.isFunction(processCreds)) {
-          processCreds({
+        if (!_.isFunction(processCreds)) {
+          this.model.save();
+        } else {
+          var creds = {
             username: this.options.appState.get('userEmail'),
             password: this.model.get('newPassword')
-          });
+          };
+          if (processCreds.length === 2) {
+            processCreds(creds, _.bind(this.model.save, this.model));
+          } else {
+            processCreds(creds);
+            this.model.save();
+          }
         }
-        this.model.save();
       });
     }
 
--- a/test/unit/spec/PasswordExpired_spec.js
+++ b/test/unit/spec/PasswordExpired_spec.js
@@ -24,8 +24,12 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordExpiredForm, Beacon, Expec
   var itp = Expect.itp;
   var tick = Expect.tick;
   var processCredsSpy = jasmine.createSpy('processCredsSpy');
+  var processCredsAsyncSpy = jasmine.createSpy('processCredsAsyncSpy');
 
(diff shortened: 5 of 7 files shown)


AlainODea:
The original idea, work, and testing on this belongs to my colleague Joe Ross. Thank you Joe.

src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -38,26 +38,37 @@ define([
...
-          DeviceFingerprint.generateDeviceFingerprint(this.settings.get('baseUrl'), this.$el)
-          .then(function (fingerprint) {
-            self.model.set('deviceFingerprint', fingerprint);
-          })
-          .fail(function () {
-            // Keep going even if device fingerprint fails
-          })
-          .fin(function () {
+        function postProcessCreds() {
+          if (this.settings.get('features.deviceFingerprinting')) {
```
rchild-okta: It doesn't look like "this" is bound to the correct scope here. 

In general, now that theres a bit more logic (including with DeviceFingerprint), I think we can simplify the callbacks by wrapping this in a promise chain:

```javascript
var self = this;

// Note: Will need to import 'q' as a module up top in the AMD definition
Q.Promise(function (resolve) {
  var processCreds = self.settings.get('processCreds');
  var creds = {
    username: self.model.get('username'),
    password: self.model.get('password')
  };

  if (!_.isFunction(processCreds)) {
    return;
  }

  if (processCreds.length === 2) {
    processCreds(creds, resolve);
  } else {
    processCreds(creds);
    resolve();
  }
})
.then(function () {
  if (!self.settings.get('features.deviceFingerprinting')) {
    return;
  }
  return DeviceFingerprint.generateDeviceFingerprint(self.settings.get('baseUrl'), self.$el)
  .then(function (fingerprint) {
    self.model.set('deviceFingerprint', fingerprint);
  })
  .fail(function () {
    // Keep going even if device fingerprint fails
  });
})
.then(function () {
  self.model.save();
});
```

AlainODea:
@rchild-okta great idea. Much cleaner with the arity check and promises!

src/views/primary-auth/PrimaryAuthForm.js
AlainODea: I will fix this bug.

src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -38,26 +38,37 @@ define([
     initialize: function () {
       this.listenTo(this, 'save', function () {
         var processCreds = this.settings.get('processCreds');
-        if (_.isFunction(processCreds)) {
-          processCreds({
-            username: this.model.get('username'),
-            password: this.model.get('password')
-          });
+        var asyncProcessCreds = this.settings.get('asyncProcessCreds');
```
AlainODea: I will switch to arity checks for determining sync vs async.

src/views/primary-auth/PrimaryAuthForm.js
AlainODea: I will also refactor to promises. I might be a while getting back with the updated/tested version. I'll make sure to test DeviceFingerprint as well to ensure I haven't broken that on the way.

rchild-okta: Great, sounds good

joelfranusic-okta:
@AlainODea and @rchild-okta - just checking in here. It looks like Travis CI is failing ([see line 2383 of the output](https://travis-ci.org/okta/okta-signin-widget/builds/204453649#L2383)). 

What do we need to do to push this forward?

AlainODea:
@joelfranusic-okta I'm testing with the Promises version now. Looks good. I should have an updated version to show soon.

AlainODea:
@joelfranusic-okta this is ready from my perspective. I've tested it to work for our use case.

Travis was failing because I was using `postProcessCreds` before it was defined:
https://travis-ci.org/okta/okta-signin-widget/builds/204453649#L2392

Following @rchild-okta's lead very directly on the Promise-based implementation fixed the issue in Travis and the other code complexity issues I was encountering :)

AlainODea:
@rchild-okta I've addressed the lint issues., but I'm still getting failing tests.

I can't get the test suite or individual specs to run locally. My latest attempt at running an individual spec is:

```bash
npm test "--spec PrimaryAuth_spec.js"
```

PhantomJS launches, but appears to hang. I've let it run for over an hour and it shows no further output. These are the lines it hangs immediately after:

```plain
Running "jasmine:test" (jasmine) task
Testing jasmine specs via PhantomJS
```

AlainODea:
It looks like installing phantomjs or phantomjs-prebuilt externally will break the ability to run tests. I'm not well enough versed in NPM, Jasmine, or PhantomJS internals to get to the root of this to prove it conclusively, but my fix below works for me.

Here is the symptom of the above. From `npm test`, PhantomJS would launch, but would hang:

```plain
Running "jasmine:test" (jasmine) task
Testing jasmine specs via PhantomJS
```

To fix this (caveat emptor: this may break other uses of PhantomJS you have):

```bash
npm remove -g phantomjs
npm remove -g phantomjs-prebuilt
npm remove phantomjs
npm remove phantomjs-prebuilt
```

AlainODea:
Jasmine tests FTW :)

It took me a while to see where I had gone wrong, but it makes total sense in retrospect.

Here were my mistakes that the Jasmine tests thankfully wouldn't let me get away with:

1. I forgot to call **resolve()** if **processCreds** wasn't a function
2. I forgot to integrate changes to PrimaryAuthForm.js from master when I rebased
3. Nesting Promises is tangly. I couldn't get it to work so I reverted to continuation passing style (sorry @rchild-okta)

I really, really tried to get the Promises implementation working, but I ran into several very difficult to debug interactions with other parts of the widget code. It looks like way in which Promises make the execution of the save listener async instead of sync breaks many expectations in the tests. I considered Q.all to wait synchronously for the whole Promise chain, but (as far as I can tell) that relies on having EcmaScript2015 generators (**yield**) in order to work.

AlainODea:
I did a search for usages of **processCreds** and found that Password Expired and Password Reset depend on it and I hadn't updated them to handle the async capability. As a result these capabilities would have broken.

I have integrated those flows and added new Jasmine tests to cover them when used with async **processCreds** functions.

AlainODea:
I dug deeper into why the Promises implementation fails the tests.

Using Promises for PrimaryAuthForm's save listener introduces asynchrony. It does save as expected, but the save can occur after **test.form.submit()** returns.

A _lot_ of tests would need to be modified to tolerate the asynchrony Promises introduce. While I think the new behavior is correct, I'm not in a good position to make that risk analysis call for all Okta users.

In order to support the Promises implementation, all tests that submit the auth form and assert side-effects of its model save would need to have a snippet of code like this injected after `test.form.submit()` and before the assertions to wait until after the save is processed:

```javascript
  return tick(test);
}).then(function(test) {
```

Even attempting that doesn't completely work, because there are cases like **PrimaryAuth_spec:events:beacon loading:does not show beacon-loading animation when primaryAuth fails (no security image)** need to check that the beacon is briefly on and then off after a tick. This is difficult to handle as the model save (which triggers the beacon) would happen asynchronously and the trick above will wind up waiting until after the beacon has already been turned off. Arguably it's functionally irrelevant that the beacon is briefly on since the desired state is that it isn't, but I'm not deeply familiar with that aspect of the widget code.

The current implementation using continuation passing style is unencumbered by these complications as it makes **test.form.submit()** call **this.model.save()** synchronously (before returning) except where async processCreds is supplied. In that case, the asynchrony is expected and handled by the associated unit tests.

AlainODea:
@rchild-okta have I addressed your code review concerns? Is my explicit CPS-based implementation an acceptable alternative to using Promises?

I spent nearly 20 hours trying to make the Promises-based implementation work without luck. I have a manual test solid version of the Promises version on a separate branch:
https://github.com/AlainODea/okta-signin-widget/tree/ao-asyncProcessCreds-Promises

rchild-okta:
@AlainODea, thank you for spending so much time on this.

> I can't get the test suite or individual specs to run locally.

I wonder if this is a Node/NPM version issue - we've got Node 5.8 configured locally as well as on Travis. There have been issues with Phantom before - last time I checked it's not quite a node module, so it hasn't played well with different versions. Something we can look into - I'm guessing we can probably pull in a newer version of jasmine that will point to phantomjs-prebuilt.

> Jasmine tests FTW!

👍 

> I did a search for usages of processCreds and found that Password Expired and Password Reset depend on it and I hadn't updated them to handle the async capability.

Awesome, thanks!

>  It looks like way in which Promises make the execution of the save listener async instead of sync breaks many expectations in the tests.
> ...Using Promises for PrimaryAuthForm's save listener introduces asynchrony. It does save as expected, but the save can occur after test.form.submit() returns.

I see - yes that's a problem. For some context, we originally started with the "tick" idea when the async model was a bit simpler. It looks like we still have some places that we didn't convert over.

Most of the tests should have transitioned to the wait* expect functions ([here](https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/util/Expect.js#L100-L154) and [here](https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/util/Expect.js#L218-L259)) - the nice thing about these functions is that they explicitly wait for an action to happen rather than an arbitrary time with the ticks. We had the exact same thing happen - every time we added another async wait in a core function, we'd either have to make the tick wait longer or add in extra ticks before checking expectations.

In this case, the ideal fix would be to update those tests that are breaking to use [waitForSpyCall](https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/util/Expect.js#L124-L129) with a success spy, i.e. [here in PrimaryAuthSpec](https://github.com/okta/okta-signin-widget/blob/master/test/unit/spec/PrimaryAuth_spec.js#L1641). The general idea (as you mentioned) is that currently the breaking tests assume they can run expectations immediately, but we've added an extra async cycle. If the test behavior is a non-error case, we can use the success function as the wait signal. If the test is testing some error behavior, we can use [waitForFormError](https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/util/Expect.js#L131-L136).

I also took a look at your wip for the promise implementation, some thoughts:

```javascript
.then(Q.Promise(function(resolve) {})
```

There's actually an issue here with the timing - by calling Q.Promise in the then, this block is executing immediately (i.e. concurrently with the previous block, rather than after the previous block is done). Also, since that deviceFingerprint function is already returning a promise, we don't have to construct one ourselves. I'd go ahead and replace this with:

```javascript
.then(function () {
  // Note: Because we're already in a promise chain, this is okay! We don't need to return
  // promises inside of a promise chain (it will automatically be converted over).
  if (!self.settings.get('features.deviceFingerprinting')) {
    return;
  }
  return DeviceFingerprint.generateDeviceFingerprint(self.settings.get('baseUrl'), self.$el)
  .then(function (fingerprint) {
    self.model.set('deviceFingerprint', fingerprint);
  })
  .fail(function () {
    // Keep going even if device fingerprint fails
  });
})
```

> Is my explicit CPS-based implementation an acceptable alternative to using Promises?

Do you think you'll have time to look into trying out my comments above? My thoughts on how to get this commit in:

- If you still have time and are interested in trying this out, go for it!
- If not, I'll try to find someone here to help out (maybe @mauriciocastillosilva-okta, who worked on this part of the code recently?)
- I am a bit hesitant on the CPS implementation because it feels like we're pretty close with the Promise one (which will simplify the code and update those tests that should have already been converted over). That being said, I also don't know which tests are failing - if it affects tests that don't just stop at PrimaryAuth (and that are already using some form of waitForController), I'm open to using CPS - that would lead me to think that updating to promises is more risky.

AlainODea:
@rchild-okta there were failures beyond the PrimaryAuth tests. I'll get a full scope tonight.

I'll incorporate those changes on my other branch and make a separate PR for it so we can see the issue separately in Travis.

I like the **waitFor** approach as it could be used to cleanly and intentionally wait for the save event. I'll take another look.

AlainODea:
#187 is the real deal with Promises. It works. I'm closing this.

PR #183, opened 2017-03-09
Add aria alert for security images.

ravikanodia-okta:
Puts the security image name into a span with aria tags so that it is read out loud by screen readers.

Bacon: test
Resolves: OKTA-115417

### DIFF
--- a/src/models/AppState.js
+++ b/src/models/AppState.js
@@ -32,6 +32,9 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
   var USER_NOT_SEEN_ON_DEVICE = '/img/security/unknown.png';
   var UNDEFINED_USER = '/img/security/default.png';
   var NEW_USER = '/img/security/unknown-device.png';
+  var NEW_USER_IMAGE_DESCRIPTION = '';
+  var UNDEFINED_USER_IMAGE_DESCRIPTION = '';
+  var UNKNOWN_IMAGE_DESCRIPTION = '';
 
   var securityImageUrlTpl = compile('{{baseUrl}}/login/getimage?username={{username}}');
 
@@ -40,7 +43,10 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
 
     // When the username is empty, we want to show the default image.
     if (_.isEmpty(username) || _.isUndefined(username)) {
-      return Q(UNDEFINED_USER);
+      return Q({
+        'securityImage': UNDEFINED_USER,
+        'securityImageDescription': UNDEFINED_USER_IMAGE_DESCRIPTION
+      });
     }
 
     return Q($.get(url)).then(function (res) {
@@ -49,9 +55,16 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         // we want to show the unknown-device security image.
         // We are mapping the server's img url to a new one because
         // we still need to support the original login page.
-        return NEW_USER;
+        return {
+          'securityImage': NEW_USER,
+          'securityImageDescription': NEW_USER_IMAGE_DESCRIPTION
+        };
       }
-      return res.pwdImg;
+      return {
+        'securityImage': res.pwdImg,
+        'securityImageDescription':
+            res.imageDescription || UNKNOWN_IMAGE_DESCRIPTION
+      };
     });
   }
 
@@ -80,7 +93,9 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         this.listenTo(this, 'change:username', function (model, username) {
           getSecurityImage(this.get('baseUrl'), username)
           .then(function (image) {
-            model.set('securityImage', image);
+            model.set('securityImage', image.securityImage);
+            model.set(
+              'securityImageDescription', image.securityImageDescription);
           })
           .fail(function (jqXhr) {
             // Only notify the consumer on a CORS error
@@ -107,6 +122,8 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
       factors: 'object',
       policy: 'object',
       securityImage: ['string', true, UNDEFINED_USER],
+      securityImageDescription:
+          ['string', true, UNDEFINED_USER_IMAGE_DESCRIPTION],
       userCountryCode: 'string',
       userPhoneNumber: 'string',
       factorActivationType: 'string',
--- a/src/views/shared/SecurityBeacon.js
+++ b/src/views/shared/SecurityBeacon.js
@@ -21,6 +21,7 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
     // are hard coded into the css and the value returned by the server
     // is ignored.
     var imgSrc = appState.get('securityImage'),
+        imgDescription = appState.get('securityImageDescription'),
         isUndefinedUser = appState.get('isUndefinedUser'),
         isNewUser = appState.get('isNewUser'),
         isSecurityImage = !isUndefinedUser && !isNewUser;
@@ -36,6 +37,9 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
       return;
     }
     if (isSecurityImage) {
+      // TODO: Newer versions of qtip will remove aria-describedby on their own when destroy() is called.
+      el.removeAttr('aria-describedby');
+      el.find('.auth-beacon-description').text(imgDescription);
       el.css('background-image', 'url(' + _.escape(imgSrc) + ')');
       return;
     }
@@ -120,7 +124,8 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
         <div class="circle right"></div>\
       </div>\
     </div>\
-    <div class="bg-helper auth-beacon auth-beacon-security" data-se="security-beacon">\
+    <div aria-live="polite" role="image" class="bg-helper auth-beacon auth-beacon-security" data-se="security-beacon">\
+      <span class="auth-beacon-description"></span>\
       <div class="okta-sign-in-beacon-border auth-beacon-border js-auth-beacon-border">\
       </div>\
     </div>\
--- a/assets/sass/modules/_beacon.scss
+++ b/assets/sass/modules/_beacon.scss
@@ -76,6 +76,12 @@
   @include box-shadow(0 0 0 15px);
 }
 
+.auth-beacon-description {
+  display: block;
+  height: 0;
+  overflow: hidden;
+}
+
 .undefined-user {
   background-image: url('../img/security/default.png');
 
--- a/test/unit/helpers/xhr/security_image.js
+++ b/test/unit/helpers/xhr/security_image.js
@@ -3,6 +3,7 @@ define({
   "responseType": "json",
   "response": {
     result: 'success',
-    pwdImg: '../../../test/unit/assets/1x1.gif'
+    pwdImg: '../../../test/unit/assets/1x1.gif',
+    imageDescription: 'a single pixel'
   }
 });
--- a/test/unit/helpers/xhr/security_image_fail.js
+++ b/test/unit/helpers/xhr/security_image_fail.js
@@ -3,6 +3,7 @@ define({
   "responseType": "json",
   "response": {
     result: 'success',
-    pwdImg: '/img/security/unknown.png'
+    pwdImg: '/img/security/unknown.png',
+    imageDescription: ''
   }
 });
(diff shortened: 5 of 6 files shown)


src/views/shared/SecurityBeacon.js
```diff
@@ -28,14 +43,19 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
     el.css('background-image', '');
     el.removeClass('new-user undefined-user');
     if (isNewUser) {
+      el.find('.auth-beacon-description').text('no security image');
```
hor-kanchan-okta: We need UX people.
Maybe it makes sense that the screen reader don't speak when there is no security image. Instead of speaking "no security image".

src/views/shared/SecurityBeacon.js
```diff
@@ -15,6 +15,21 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
...
+  function getSecurityImageDescriptionFromUrl(imgSrc) {
+    var parser = document.createElement('a');
+    parser.href = imgSrc;
+    // Everything after the last slash, unless the string ends with a slash,
+    // in which case get everything between the last two slashes.
+    var match = /([^/]*)\/?$/.exec(parser.pathname);
+    var filename = (match.length >= 2) ? match[1] : null;
+    // Everything before the first period (or the entire string if there are
+    // no periods).
+    var basename = /^[^.]*/.exec(filename);
```
hor-kanchan-okta: Can you combine into one regex?

src/views/shared/SecurityBeacon.js
```diff
@@ -15,6 +15,21 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
   var _ = Okta._,
       $ = Okta.$;
 
+  // TODO: Server should send description as part of appState.
+  function getSecurityImageDescriptionFromUrl(imgSrc) {
```
hor-kanchan-okta: Need a jira for the i18n on the security image.

src/views/shared/SecurityBeacon.js
```diff
@@ -28,14 +43,19 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
...
       return;
     }
     if (isUndefinedUser) {
+      el.find('.auth-beacon-description').text('no security image');
       el.addClass('undefined-user');
       return;
     }
     if (isSecurityImage) {
+      // TODO: Newer versions of qtip will remove aria-describedby on their own when destroy() is called.
+      el.removeAttr('aria-describedby');
```
hor-kanchan-okta: Does it affect the screen reader?
I thought the aria-describedby is not working (screen reader ignores it). That's why we workaround it. If the `aria-describedby` works, we should use it instead.

src/views/shared/SecurityBeacon.js
```diff
@@ -120,7 +140,8 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
         <div class="circle right"></div>\
       </div>\
     </div>\
-    <div class="bg-helper auth-beacon auth-beacon-security" data-se="security-beacon">\
+    <div aria-live="polite" role="alert" class="bg-helper auth-beacon auth-beacon-security" data-se="security-beacon">\
```
hor-kanchan-okta: I don't think we should use `role=alert`. The spec said alert should be used for errors and will interrupt user actions. It `role=alert` implies `aria-live=assertive`. The customer asks for `aria-live="polite` specifically.

src/views/shared/SecurityBeacon.js
ravikanodia-okta: qtip sets the aria-describedby property when it's initialized. I don't know whether that property is working correctly or not. qtip.destroy() gets called before this function, but it leaves behind the (now-erroneous) aria-describedby it already set.

src/views/shared/SecurityBeacon.js
ravikanodia-okta: I can change this, but when I use `aria-live="polite"`, I never get the text read by Voiceover.

That said, without having a screen reader expert to consult, I am ok with doing exactly what the customer asked for, under the assumption that they know something about their tools that I do not know.

src/views/shared/SecurityBeacon.js
ravikanodia-okta: I did consider that possibility. We do have a visible callout that says something like "You are connecting to foo.okta.com for the first time". My thought process is that, if we thought that the lack of a security image was important enough to specifically mention to sighted users, it's worth telling a vision-impaired user as well.

src/views/shared/SecurityBeacon.js
ravikanodia-okta: Done.

src/views/shared/SecurityBeacon.js
ravikanodia-okta: Tricky, but done.

src/views/shared/SecurityBeacon.js
mauriciocastillosilva-okta: We also need to localize the text

src/views/shared/SecurityBeacon.js
```diff
@@ -15,6 +15,18 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
...
+  // TODO(OKTA-118980): Server should send description as part of appState.
+  function getSecurityImageDescriptionFromUrl(imgSrc) {
+    var parser = document.createElement('a');
+    parser.href = imgSrc;
+
+    // Get stuff between the last slash, and the first period after that slash.
+    // (if the string ends with a slash, use the second-to-last slash instead.)
+    var match = /^.*\/([^.]+).*\/?$/.exec(parser.pathname);
+    var basename = (match.length >= 2) ? match[1] : null;
+    return basename ? (basename + ' security image') : '';
```
mauriciocastillosilva-okta: we need to localize this string... this is tricky since the image filename is what we are using to describe the image... so maybe we need something like:
```
securityimage.aria.text = {0} security image
securityimage.beer = beer
securityimage.flower = flower
...
```

src/views/shared/SecurityBeacon.js
hor-kanchan-okta: I have a demo in codepen which is using aria-live=polite, does that work for your screen reader?

mauriciocastillosilva-okta: yeah, I don't know what is the best here either... maybe we should ask an expert if we can find one...

From [w3/roles](https://www.w3.org/TR/wai-aria/roles#alert)
> A message with important, and usually time-sensitive, information. 
> Alerts are used to convey messages to alert the user. In the case of audio warnings this is an accessible alternative for a hearing-impaired user.

src/views/shared/SecurityBeacon.js
```diff
@@ -120,7 +137,8 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
         <div class="circle right"></div>\
       </div>\
     </div>\
-    <div class="bg-helper auth-beacon auth-beacon-security" data-se="security-beacon">\
+    <div aria-live="polite" role="alert" class="bg-helper auth-beacon auth-beacon-security" data-se="security-beacon">\
+      <span class="auth-beacon-description" style="height: 0; display: block; overflow: hidden"></span>\
```
mauriciocastillosilva-okta: can you look to see if we already have a class for this? It seems that we only need it hidden, so there should be one already...

src/views/shared/SecurityBeacon.js
ravikanodia-okta: I believe i18n for this is going to require several steps. As you said, we need the properties. And then we need to pipe them through to the client. It looks like SecurityImage can hold alt text, though I'm not sure whether that actually gets populated anywhere (probably not?). SessionCookieManager uses UserSecurityFields, which holds the security image as a string, and is ultimately populated by the Login table which also stores "pwdImg" as a string. At least getimage.jsp should be easy to update once all that is changed.

src/views/shared/SecurityBeacon.js
ravikanodia-okta: The codepen demo works for me. I think there are just too many things going on at once in the actual loginpage. I've tried many, many variants with slightly different arrangements of DOM elements, text() vs html() vs adding/removing a node from the DOM. None of them work for me in this case without `role=alert`.

src/views/shared/SecurityBeacon.js
ravikanodia-okta: The idea here is to make it invisible to the user, but still "displayed" from the browser's perspective. Elements with `display: none` aren't read by screen readers (to the best of my knowledge).

That said, I could add a class with these attributes. Would courage be the right place for it?

src/views/shared/SecurityBeacon.js
mauriciocastillosilva-okta: Thinking about it more, I think the API should return you this info... we can talk about it tomorrow. @hor-kanchan-okta also likes that idea better....

src/views/shared/SecurityBeacon.js
ravikanodia-okta: I decided to leave this out based on the lack of localization and also the absence of a request from the customer.

src/views/shared/SecurityBeacon.js
ravikanodia-okta: After thinking it over, I am going with role="image" because that is what the customer asked for. It doesn't read out loud in Voiceover for me but as I said earlier, I would rather not get ahead of them.

ravikanodia-okta:
Made significant changes (gets imageDescription via getimage) and updated tests.

src/models/AppState.js
```diff
@@ -49,9 +55,15 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
...
         // we still need to support the original login page.
-        return NEW_USER;
+        return {
+          'pwdImg': NEW_USER,
+          'imageDescription': NEW_USER_IMAGE_DESCRIPTION
+        };
       }
-      return res.pwdImg;
+      return {
+        'pwdImg': res.pwdImg,
```
mauriciocastillosilva-okta: nit nit: this could be named ```'image': res.pwdImg,```

src/models/AppState.js
```diff
@@ -107,6 +120,7 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
       factors: 'object',
       policy: 'object',
       securityImage: ['string', true, UNDEFINED_USER],
+      imageDescription: ['string', true, UNDEFINED_USER_IMAGE_DESCRIPTION],
```
mauriciocastillosilva-okta: nit: this could be ```securityImageDescription```

src/views/shared/SecurityBeacon.js
mauriciocastillosilva-okta: so, if I understand it correctly, we can't really set an attribute in the image <div> element because the screenreader doesn't read it, so the workaround is to have a different <span> element with the text describing the image? Does that satisfy the accessibility standards? Asking since usually images have an alt attribute to describe them, or an aria attribute as well in the same element, not in a different one.

src/views/shared/SecurityBeacon.js
```diff
@@ -120,7 +124,8 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
         <div class="circle right"></div>\
       </div>\
     </div>\
-    <div class="bg-helper auth-beacon auth-beacon-security" data-se="security-beacon">\
+    <div aria-live="polite" role="image" class="bg-helper auth-beacon auth-beacon-security" data-se="security-beacon">\
+      <span class="auth-beacon-description" style="height: 0; display: block; overflow: hidden"></span>\
```
mauriciocastillosilva-okta: can we move the styles to a css class and use that here?

src/views/shared/SecurityBeacon.js
ravikanodia-okta: Yes - the screenreader doesn't pay attention to alt text for images loaded as background-image. I believe it does satisfy the standard if it is contained in the same DOM element.

src/models/AppState.js
ravikanodia-okta: changed these up to be nicer

src/models/AppState.js
ravikanodia-okta: same

src/views/shared/SecurityBeacon.js
ravikanodia-okta: done

ravikanodia-okta:
Mau, I'll merge when you give the ship :)

mauriciocastillosilva-okta:
Rebase to the latest since the branch is out of sync, and also change the commit message to follow this guidelines: https://github.com/okta/okta-signin-widget/blob/master/CONTRIBUTING.md

ravikanodia-okta:
Hmm. I did something very, very wrong when attempting to rebase.

ravikanodia-okta:
OK. There we go.

PR #184, opened 2017-03-09
Adds an aria alert to the forgot password form.

ravikanodia-okta:
Whether this gets read out loud or not is unfortunately not very consistent. Voiceover seems to like doing its own thing. In general, typing the username and pressing enter does not get the text read. If the user clicks the reset button with the mouse, or tabs to it and then presses enter, they are likely (but still not guaranteed!) to get the text read.

Bacon: test
Resolves: OKTA-118645

### DIFF
--- a/src/PwdResetEmailSentController.js
+++ b/src/PwdResetEmailSentController.js
@@ -40,6 +40,17 @@ function (Okta, Enums, FormController, FormType) {
       attributes: { 'data-se': 'pwd-reset-email-sent' },
       formChildren: function () {
         return [
+          FormType.View({
+            View: Okta.View.extend({
+              template: '\
+              <span class="accessibility-text" role="status">{{alert}}</span>\
+              ',
+              getTemplateData: function () {
+                return { 'alert': Okta.loc('password.forgot.emailSent.title', 'login') };
+              }
+            })
+          }),
+
           FormType.Button({
             title: Okta.loc('goback', 'login'),
             className: 'button button-primary button-wide',
--- a/src/views/shared/SecurityBeacon.js
+++ b/src/views/shared/SecurityBeacon.js
@@ -39,7 +39,7 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
     if (isSecurityImage) {
       // TODO: Newer versions of qtip will remove aria-describedby on their own when destroy() is called.
       el.removeAttr('aria-describedby');
-      el.find('.auth-beacon-description').text(imgDescription);
+      el.find('.accessibility-text').text(imgDescription);
       el.css('background-image', 'url(' + _.escape(imgSrc) + ')');
       return;
     }
@@ -125,7 +125,7 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
       </div>\
     </div>\
     <div aria-live="polite" role="image" class="bg-helper auth-beacon auth-beacon-security" data-se="security-beacon">\
-      <span class="auth-beacon-description"></span>\
+      <span class="accessibility-text"></span>\
       <div class="okta-sign-in-beacon-border auth-beacon-border js-auth-beacon-border">\
       </div>\
     </div>\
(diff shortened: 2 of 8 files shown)


src/PwdResetEmailSentController.js
```diff
@@ -40,6 +40,17 @@ function (Okta, Enums, FormController, FormType) {
       attributes: { 'data-se': 'pwd-reset-email-sent' },
       formChildren: function () {
         return [
+          FormType.View({
+            View: Okta.View.extend({
+              template: '\
+              <span class="overflow-hidden" style="width:0; height:0; display:block" role="alert">{{alert}}</span>\
```
mauriciocastillosilva-okta: is this going to be the way to solve aria screenreader issues? I see the same span element in the other PR you have for security images, so maybe we should create a view that contains this and add it in all the relevant forms.

src/PwdResetEmailSentController.js
```diff
@@ -40,6 +40,17 @@ function (Okta, Enums, FormController, FormType) {
...
+              ',
+              getTemplateData: function () {
+                return { 'alert': _.partial(Okta.loc, 'password.forgot.emailSent.title', 'login') };
```
mauriciocastillosilva-okta: don't think you need ```_.partial``` here

src/PwdResetEmailSentController.js
ravikanodia-okta: I'm not 100% sure what the ideal way to solve this problem is. My goal is to solve the immediate issue that our customer has raised. I'm hesitant to build a highly generic solution before we have more experience understanding these issues. That said, I can give it a try tonight and see how it works out.

src/PwdResetEmailSentController.js
ravikanodia-okta: Done.

src/PwdResetEmailSentController.js
```diff
@@ -40,6 +40,17 @@ function (Okta, Enums, FormController, FormType) {
       attributes: { 'data-se': 'pwd-reset-email-sent' },
       formChildren: function () {
         return [
+          FormType.View({
+            View: Okta.View.extend({
+              template: '\
+              <span class="overflow-hidden" style="height:0; display:block" role="alert">{{alert}}</span>\
```
mauriciocastillosilva-okta: can we move the styles to a css class so we can reuse it? Don't we have one already that we use for the security image?
Next question is: is 'alert' good? Or do we want 'status'? I'm not sure about this, that is why I am asking

Last thing: If we need to add this again in another Controller we should definitly move it to a separate View, so we can use the same view wherever we need

ravikanodia-okta: Moved the styles to a class, and changed the security image to use that class as well.

The original request from the customer (https://oktainc.atlassian.net/browse/OKTA-114232) doesn't specify a type. I'm ok using "status" with the caveat that we might need to change it if the customer is not happy.

src/views/shared/SecurityBeacon.js
```diff
@@ -39,7 +39,7 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
     if (isSecurityImage) {
       // TODO: Newer versions of qtip will remove aria-describedby on their own when destroy() is called.
       el.removeAttr('aria-describedby');
-      el.find('.auth-beacon-description').text(imgDescription);
```
mauriciocastillosilva-okta: This probably will fail the tests you had for the security beacon.

ravikanodia-okta: Fixed.

ravikanodia-okta:
Added a test for this behavior.

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -884,7 +884,7 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
             success: undefined
           });
           expect($.fn.css).toHaveBeenCalledWith('background-image', 'url(../../../test/unit/assets/1x1.gif)');
-          expect($('.auth-beacon-description').text()).toBe('a single pixel');
+          expect($('.accessibility-text').text()).toBe('a single pixel');
```
mauriciocastillosilva-okta: you should be able to use your newly added method to Form, so this should be
```
expect(test.form.accessibilityText()).toBe('a single pixel');
```

ravikanodia-okta:
Switched the other tests to use Form.accessibilityText()

PR #185, opened 2017-03-09
:bug: Remove anti-phishing tooltip when moving to a new page

mauriciocastillosilva-okta:
- Issue is that the tooltip was not removed when going to unlock/forgot-password pages.

Resolves: OKTA-117900
Bacon: test

### DIFF
--- a/test/unit/helpers/dom/PrimaryAuthForm.js
+++ b/test/unit/helpers/dom/PrimaryAuthForm.js
@@ -66,9 +66,9 @@ define(['jquery', 'underscore', './Form'], function ($, _, Form) {
       return this.tooltipText(SECURITY_BEACON);
     },
 
-    securityImageTooltipDestroyed: function () {
+    isSecurityImageTooltipDestroyed: function () {
       var api = this.tooltipApi(SECURITY_BEACON);
-      return api ? api.destroyed : undefined;
+      return api ? api.destroyed : true;
     },
 
     securityBeacon: function () {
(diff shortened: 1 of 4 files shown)


test/unit/helpers/dom/PrimaryAuthForm.js
```diff
@@ -66,6 +66,11 @@ define(['jquery', 'underscore', './Form'], function ($, _, Form) {
       return this.tooltipText(SECURITY_BEACON);
     },
 
+    securityImageTooltipDestroyed: function () {
```
santhoshbalakrishnan: nit can we change this to isSecurityImageTooltipDestroyed() and make sure it returns a boolean. Would that be better than expecting an undefined for success?

santhoshbalakrishnan:
Also, this does not seem to happen on clicking forgot password now. Was this fixed for forgot password? 

signin-tooltip

test/unit/helpers/dom/PrimaryAuthForm.js
mauriciocastillosilva-okta: will do, no prob

mauriciocastillosilva-okta:
I played a little bit with ```unfocus``` and other arguments for qtip, but the thing is that we need to remove the tooltip when we navigate away from the page, not when we unfocus an element (that is more for regular tooltips when you hover or focus a ?)

Regarding Forgot Password, it seems that it worked most of the time for me as well, but there were times that it didn't. So it is not 100% consistent. For Unlock account it seems that it didn't work every time. 
With this change we have a consistent way of getting rid of that tooltip

santhoshbalakrishnan:
> that is more for regular tooltips when you hover or focus a ?

Thought that was a similar case here. hide:unfocus just closes the element when we click anywhere else. 

This review looks good to me if the above options were not sufficient :)

PR #187, opened 2017-03-14
Alternative Promises-based :seedling: Add async credentials processing

AlainODea:
This is more for discussion and consideration for hand-off to Okta for implementation as an alternative to #179.

Many unit tests still fail.

The path forward suggested by @rchild-okta is to create a **waitFor** function that waits for the PrimaryAuthForm's model save before running assertions.


AlainODea:
Okay, lots is an exaggeration. It fails 5 tests, and the failures are contained within PrimaryAuth_spec :)

AlainODea:
@rchild-okta you win :1st_place_medal: 

I got it working. It was almost entirely synchronous assertions constraints in tests. There were some minor issues with asserting intermediate states. I scrapped those assertions since I believe the tests were overconstrained on intermediate states that aren't meaningful. My assumptions may be wrong so check the tests carefully.

test/unit/spec/PasswordExpired_spec.js
```diff
@@ -24,8 +24,12 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordExpiredForm, Beacon, Expec
   var itp = Expect.itp;
   var tick = Expect.tick;
   var processCredsSpy = jasmine.createSpy('processCredsSpy');
```
rchild-okta: Hmm, strange, not sure why we made these global - feels like this might actually result in hard-to-track bugs if we we didn't reset the spy calls between tests. 

Instead, looking at the test code, feels like we should just define spies for processCreds when we're actually doing a processCreds type test.

For example, in setup, we could add this code:
```javascript
// This demonstrates the new code to add
function setup(settings, res) {
  // convenience so you don't have to do checks for existince later
  settings || (settings = {});

  // Also need a successSpy and use it instead of tick's below
  var successSpy = jasmine.createSpy('successSpy');

  var router = new Router({
    globalSuccessFn: successSpy,
    processCreds: settings.processCreds
    // other stuff
  });

  return Expect.waitForPasswordExpired({
    // other stuff
    successSpy: successSpy
  });
}
```

And then the test would look like this:
```javascript
itp('calls async processCreds function before saving a model', function () {
  const processCredsSpy = jasmine.createSpy().and.callFake(function (creds, cb) {
    cb();
  });
  return setup({ processCreds: processCredsSpy })
  .then(function (test) {
    submitNewPass(test, 'oldpwd', 'newpwd', 'newpwd');
    return Expect.waitForSpyCall(test.successSpy, test);
  })
  .then(function (test) {
    // We know if we got here we invoked the success callback
    expect(processCredsSpy.calls.count()).toBe(1);
    expect(processCredsSpy).toHaveBeenCalledWith({
      username: 'inca@clouditude.net',
      password: 'newpwd'
    }, jasmine.any(Function));
  })
});
```

test/unit/spec/PasswordExpired_spec.js
```diff
@@ -124,14 +128,57 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordExpiredForm, Beacon, Expec
             username: 'inca@clouditude.net',
             password: 'newpwd'
           });
+          return tick();
```
rchild-okta: Calling out that the last comment I added replaced the tick() with a waitForSpy call

test/unit/spec/PasswordExpired_spec.js
```diff
@@ -124,14 +128,57 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordExpiredForm, Beacon, Expec
...
+        }).then(function() {
           expect($.ajax.calls.count()).toBe(1);
         });
       });
+      itp('calls async processCreds function before saving a model', function () {
+        return setup({
```
rchild-okta: nit: we usually format promise chains like this:
```javascript
somePromiseProducingFunction()
.then(function () {
  // some stuff
})
.then(function () {
  // some other stuff
});
```

test/unit/spec/PasswordExpired_spec.js
```diff
@@ -124,14 +128,57 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordExpiredForm, Beacon, Expec
...
+            username: 'inca@clouditude.net',
+            password: 'newpwd'
+          }, jasmine.any(Function));
+          return tick();
+        }).then(function() {
+          expect($.ajax.calls.count()).toBe(1);
+        });
+      });
+      itp('calls async processCreds function and can prevent saving a model', function () {
+        return setup({
```
rchild-okta: I think we should follow the same strategy to dealing with spies as mentioned in my above comment - let me know if you have any further questions about this.

test/unit/spec/PasswordExpired_spec.js
AlainODea: Can't use **const** since it will fail your linter. I'm omitting that.

rchild-okta: Ah, right, yep makes sense

test/unit/spec/PasswordExpired_spec.js
```diff
@@ -114,24 +116,71 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordExpiredForm, Beacon, Expec
...
           submitNewPass(test, 'oldpwd', 'newpwd', 'newpwd');
           expect(processCredsSpy.calls.count()).toBe(1);
           expect(processCredsSpy).toHaveBeenCalledWith({
             username: 'inca@clouditude.net',
             password: 'newpwd'
           });
+          return Expect.waitForSpyCall(test.successSpy);
+        })
+        .then(function() {
+          expect($.ajax.calls.count()).toBe(1);
```
rchild-okta: In this case I'd move the expect processCredsSpy logic to this block (i.e. after we're done waiting for success to be called) - it works because Q.Promise calls its function synchronously, but better to be explicit.

i.e.
```javascript
var processCredsSpy = jasmine.createSpy('processCredsSpy');
return setup({ processCreds: processCredsSpy })
.then(function (test) {
  $.ajax.calls.reset();
  submitNewPass(test, 'oldpwd', 'newpwd', 'newpwd');
  return Expect.waitForSpyCall(test.successSpy);
})
.then(function() {
  expect(processCredsSpy.calls.count()).toBe(1);
  expect(processCredsSpy).toHaveBeenCalledWith({
    username: 'inca@clouditude.net',
    password: 'newpwd'
  });
});
```

test/unit/spec/PasswordExpired_spec.js
```diff
@@ -114,24 +116,71 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordExpiredForm, Beacon, Expec
...
+        });
+      });
+      itp('calls async processCreds function before saving a model', function () {
+        var processCredsSpy = jasmine.createSpy('processCredsSpy');
+        return setup({ processCreds: function (creds, callback) {
```
rchild-okta: nit, format like:
```javascript
return setup({
  processCreds: function (creds, callback) {
  }
})
```

test/unit/spec/PasswordExpired_spec.js
```diff
@@ -114,24 +116,71 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordExpiredForm, Beacon, Expec
...
+          processCredsSpy(creds, callback);
+          callback();
+        }})
+        .then(function (test) {
+          $.ajax.calls.reset();
+          test.setNextResponse(resSuccess);
+          submitNewPass(test, 'oldpwd', 'newpwd', 'newpwd');
+          expect(processCredsSpy.calls.count()).toBe(1);
```
rchild-okta: Would do the same thing here and move this to the next block

test/unit/spec/PasswordExpired_spec.js
```diff
@@ -114,24 +116,71 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordExpiredForm, Beacon, Expec
...
+        .then(function() {
+          expect($.ajax.calls.count()).toBe(0);
+        });
+      });
       itp('saves the new password successfully', function () {
         return setup().then(function (test) {
           $.ajax.calls.reset();
           test.setNextResponse(resSuccess);
           submitNewPass(test, 'oldpassyo', 'boopity', 'boopity');
+          return tick();
```
rchild-okta: Replace with Expect.waitForSpyCall(test.successSpy);

test/unit/spec/PasswordReset_spec.js
```diff
@@ -177,9 +180,58 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordResetForm, Beacon, Expect,
           username: 'administrator1@clouditude.net',
           password: 'newpwd'
         });
+        return Expect.waitForSpyCall(test.successSpy);
```
rchild-okta: Similar comment here about moving the expects down

test/unit/spec/PasswordReset_spec.js
```diff
@@ -188,6 +240,7 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordResetForm, Beacon, Expect,
         test.form.setConfirmPassword('imsorrymsjackson');
         test.setNextResponse(resSuccess);
         test.form.submit();
+        return tick();
```
rchild-okta: Similar comment about waiting on success

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -502,6 +502,8 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
           test.form.setPassword('pass');
           test.setNextResponse(resSuccess);
           test.form.submit();
+          return tick(test);
```
rchild-okta: expect success

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -796,6 +798,8 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
             test.form.setUsername('testuser');
             test.form.setPassword('pass');
             test.form.submit();
+            return tick(test);
```
rchild-okta: expect success

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -836,7 +839,6 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
             test.form.setUsername('testuser');
             test.form.setPassword('pass');
             test.form.submit();
-            expect(test.beacon.isLoadingBeacon()).toBe(true);
```
rchild-okta: I think it's fine to remove this line - end state is *mostly* what we care about

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1108,10 +1110,6 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
           test.form.setPassword('pass');
           test.setNextResponse(resUnauthorized);
           test.form.submit();
-          var button = test.form.submitButton();
-          var buttonClass = button.attr('class');
-          expect(buttonClass).toContain('link-button-disabled');
-          expect(test.form.isDisabled()).toBe(true);
```
rchild-okta: In this case, it looks like the purpose of this test it to verify that we do enter the disabled state. This requires some work on our part - we need to provide a helper method to do something like:
1. Mock ajax request, but don't return the response
2. Have test validate expection, and then call a function to resolve the response promise

Can you comment out this test (with the original code where it is doing these assertions) and write a comment saying why this test no longer works? We'll fix it on our side after this PR gets checked in.

test/unit/spec/PasswordExpired_spec.js
AlainODea: Something is wrong here. This file isn't the latest code on the branch:
https://github.com/AlainODea/okta-signin-widget/blob/ao-asyncProcessCreds-Promises/test/unit/spec/PasswordExpired_spec.js#L157

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1147,31 +1160,85 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
...
+          test.form.setUsername('testuser');
+          test.form.setPassword('pass');
+          test.setNextResponse(resSuccess);
+          test.form.submit();
+          expect(processCredsSpy.calls.count()).toBe(1);
+          expect(processCredsSpy).toHaveBeenCalledWith({
+            username: 'testuser',
+            password: 'pass'
+          }, jasmine.any(Function));
+          return tick();
```
AlainODea: **successSpy** won't be called so **tick()** is unavoidable here.

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1575,6 +1642,8 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
         return setupSocial()
         .then(function (test) {
           test.form.facebookButton().click();
+          return tick(test);
+        }).then(function(test) {
```
AlainODea: Fixing this with **waitForSpyCall**. Wait for it.

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1147,31 +1160,85 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
...
           test.form.setPassword('pass');
           test.setNextResponse(resSuccess);
           test.form.submit();
-          expect(test.form.isDisabled()).toBe(true);
-          return tick();
+          return tick(test);
         })
-        .then(function () {
+        .then(function (test) {
+          expect(test.form.isDisabled()).toBe(true);
```
AlainODea: Fixing this with **waitForSpyCall**. Wait for it.

test/unit/spec/PrimaryAuth_spec.js
AlainODea: This will not work with **waitForSpyCall**. Not sure why. Sticking with **tick()** here.

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1630,7 +1651,8 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
         .then(function (test) {
           test.form.facebookButton().click();
           return tick(test);
-        }).then(function(test) {
+        })
+        .then(function(test) {
```
AlainODea: I could not get this to work with **waitForSpyCall**. It caused wait condition failures. I had to leave it as **tick()**.

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1212,7 +1232,8 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
             password: 'pass'
           }, jasmine.any(Function));
           return tick();
-        }).then(function() {
+        })
+        .then(function() {
```
AlainODea: This is intentionally **tick()** and not **waitForSpyCall** as **successSpy** will not be called as the save is prevented intentionally.

test/unit/spec/PasswordReset_spec.js
```diff
@@ -228,7 +230,8 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordResetForm, Beacon, Expect,
           password: 'newpwd'
         }, jasmine.any(Function));
         return tick();
-      }).then(function() {
+      })
+      .then(function() {
```
AlainODea: This is intentionally **tick()** and not **waitForSpyCall** as **successSpy** will not be called as the save is prevented intentionally.

AlainODea:
Ready for (hopefully) final review @rchild-okta.

johngronberg-okta:
@AlainODea Thank you so much for all of your contributions to Okta! We'd love to send you an Okta Developer t-shirt. What size would you like? The shirts are American Apparel 50/50 blend for sizing consideration. Feel free to email me directly if you prefer!

AlainODea:
@johngronberg-okta my pleasure and thank you. I've reached out to you by email with details.

AlainODea:
@rchild-okta thank you for working with me on this. I learned a ton and will be writing much better JavaScript going forward :)

PR #200, opened 2017-03-29
:seedling: Adds the ability to add a custom Auth Button

gowthamidommety-okta:
This allows users to add a custom Auth Button via new properties when initializing the sign in widget.

Resolves: OKTA-119408

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ Contributors should read our [contributing guidelines](./CONTRIBUTING.md) if the
   * [Username and password](#username-and-password)
   * [Customizing language and text](#customizing-language-and-text)
   * [Changing help links](#changing-help-links)
+  * [Adding Custom Buttons](#adding-custom-buttons)
   * [OpenId Connect](#openid-connect)
   * [Bootstrapping from a recovery token](#bootstrapping-from-a-recovery-token)
   * [Feature flags](#feature-flags)
@@ -775,6 +776,28 @@ helpLinks: {
 - **helpLinks.unlock** - Custom link href for the "Unlock Account" link. For this link to display, `features.selfServiceUnlock` must be set to `true`, and the self service unlock feature must be enabled in your admin settings.
 
 - **helpLinks.custom** - Array of custom link objects `{text, href}` that will be added to the *"Need help signing in?"* section.
+
+## Adding Custom Buttons
+
+You can add custom buttons underneath the login form on the primary auth page by setting the following config options. If you'd like to change the divider text, use the `i18n` config option.
+
+```javascript
+// An example that adds a custom button underneath the login form on the primary auth page
+customButtons: [{
+  title: 'Click Me',
+  className: 'btn-customAuth',
+  click: function() {
+    // clicking on the button navigates to another page
+    window.location.href = 'http://www.example.com';
+  }
+}]
+```
+
+- **customButtons.title** - String that is set as the button text
+
+- **customButtons.className** - Optional class that can be added to the button
+
+- **customButtons.click** - Function that is called when the button is clicked
 
 ## OpenId Connect
 
--- a/src/PrimaryAuthController.js
+++ b/src/PrimaryAuthController.js
@@ -13,12 +13,12 @@
 define([
   'okta',
   'views/primary-auth/PrimaryAuthForm',
-  'views/primary-auth/SocialAuth',
+  'views/primary-auth/CustomButtons',
   'models/PrimaryAuth',
   'shared/util/Util',
   'util/BaseLoginController'
 ],
-function (Okta, PrimaryAuthForm, SocialAuth, PrimaryAuthModel, Util, BaseLoginController) {
+function (Okta, PrimaryAuthForm, CustomButtons, PrimaryAuthModel, Util, BaseLoginController) {
 
   var compile = Okta.Handlebars.compile;
   var _ = Okta._;
@@ -139,10 +139,10 @@ function (Okta, PrimaryAuthForm, SocialAuth, PrimaryAuthModel, Util, BaseLoginCo
 
       this.addListeners();
 
-      // Add SocialAuth view only when the idps are configured. If configured, 'socialAuthPositionTop'
-      // will determine the order in which the social auth and primary auth are shown on the screen.
-      if (options.settings.get('socialAuthConfigured')) {
-        this.add(SocialAuth, {prepend: options.settings.get('socialAuthPositionTop')});
+      // If social auth is configured, 'socialAuthPositionTop' will determine
+      // the order in which the social auth and primary auth are shown on the screen.
+      if (options.settings.get('hasConfiguredButtons')) {
+        this.add(CustomButtons, {prepend: options.settings.get('socialAuthPositionTop')});
       }
       this.add(new Footer(this.toJSON({appState: options.appState})));
 
--- a/src/models/Settings.js
+++ b/src/models/Settings.js
@@ -19,9 +19,10 @@ define([
   'util/BrowserFeatures',
   'util/Util',
   'util/Logger',
+  'util/OAuth2Util',
   'json!config/config'
 ],
-function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
+function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
 
   var DEFAULT_LANGUAGE = 'en';
 
@@ -128,7 +129,10 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
       'helpLinks.help': 'string',
       'helpLinks.forgotPassword': 'string',
       'helpLinks.unlock': 'string',
-      'helpLinks.custom': 'array'
+      'helpLinks.custom': 'array',
+
+      //Custom Buttons
+      'customButtons': ['array', false, []]
     },
 
     derived: {
@@ -223,19 +227,55 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
         },
         cache: true
       },
-      // checks if there are any valid configured idps.
-      socialAuthConfigured: {
-        deps: ['configuredSocialIdps'],
-        fn: function (idps) {
-          return !_.isEmpty(idps);
+      // social auth buttons order - 'above'/'below' the primary auth form (boolean)
+      socialAuthPositionTop: {
+        deps: ['configuredSocialIdps', 'idpDisplay'],
+        fn: function (configuredSocialIdps, idpDisplay) {
+          return !_.isEmpty(configuredSocialIdps) && idpDisplay.toUpperCase() === 'PRIMARY';
         },
         cache: true
       },
-      // social auth buttons order - 'above'/'below' the primary auth form (boolean)
-      socialAuthPositionTop: {
-        deps: ['socialAuthConfigured', 'idpDisplay'],
-        fn: function (socialAuthConfigured, idpDisplay) {
-          return !!(socialAuthConfigured && idpDisplay.toUpperCase() === 'PRIMARY');
+      // merges social auth and custom buttons into one array
+      configuredButtons: {
+        deps: ['configuredSocialIdps', 'customButtons'],
+        fn: function (configuredSocialIdps, customButtons) {
+          var self = this;
+          var buttonArray = [];
+          _.each(configuredSocialIdps, function (idp) {
+            var type = idp.type.toLowerCase();
+            var dataAttr = 'social-auth-' + type + '-button';
+            var socialAuthButton = {
+              id: idp.id,
+              type: idp.type,
+              dataAttr: dataAttr,
+              className: 'social-auth-button ' + dataAttr,
+              title: idp.title = Okta.loc('socialauth.' + type + '.label'),
+              click: function (e) {
+                e.preventDefault();
+                OAuth2Util.getTokens(self, {idp: idp.id});
+              }
+            };
+            buttonArray.push(socialAuthButton);
+          });
+          if(_.isArray(customButtons)) {
+            _.each(customButtons, function(button) {
+              var customConfiguredButton = {
+                title: button.title,
+                className: button.className + ' default-custom-button',
+                click: button.click
+              };
+              buttonArray.push(customConfiguredButton);
+            });
+          }
+          return buttonArray;
+        },
+        cache: true
+      },
+      //checks if there are any configured buttons
+      hasConfiguredButtons: {
+        deps: ['configuredButtons'],
+        fn: function (configuredButtons) {
+          return !_.isEmpty(configuredButtons);
         },
         cache: true
       }
--- a/src/views/primary-auth/CustomButtons.js
+++ b/src/views/primary-auth/CustomButtons.js
@@ -11,9 +11,8 @@
  */
 
 define([
-  'okta',
-  'util/OAuth2Util'
-], function (Okta, OAuth2Util) {
+  'okta'
+], function (Okta) {
 
   var _ = Okta._;
 
@@ -27,16 +26,15 @@ define([
 
   return Okta.View.extend({
 
-    className: 'social-auth',
+    className: 'primary-auth-container',
 
     children: function () {
       var children = [],
-          idProviders = this.settings.get('configuredSocialIdps'),
-          divider = dividerTpl({text: Okta.loc('socialauth.divider.text', 'login')});
+        buttons = this.settings.get('configuredButtons'),
+        divider = dividerTpl({text: Okta.loc('socialauth.divider.text', 'login')});
 
-      // Social Auth IDPs.
-      _.each(idProviders, function (provider) {
-        children.push(this._createButton(provider));
+      _.each(buttons, function (button) {
+        children.push(this._createButton(button));
       }, this);
 
       // If the social auth buttons have to be above the Okta form, the title moves from
@@ -56,21 +54,13 @@ define([
     },
 
     _createButton: function (options) {
-      var type = options.type.toLowerCase(),
-          attr = 'social-auth-' + type + '-button';
-
       return Okta.createButton({
         attributes: {
-          'data-se': attr
+          'data-se': options.dataAttr
         },
-        className: 'social-auth-button ' + attr,
-        title: Okta.loc('socialauth.' + type + '.label'),
-        events: {
-          'click': function (e) {
-            e.preventDefault();
-            OAuth2Util.getTokens(this.settings, {idp: options.id});
-          }
-        }
+        className: options.className,
+        title: options.title,
+        click: options.click
       });
     }
 
--- a/test/unit/spec/PrimaryAuth_spec.js
+++ b/test/unit/spec/PrimaryAuth_spec.js
@@ -127,6 +127,22 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
     });
   }
 
+  function setupAdditionalAuthButton() {
+    var settings = {
+      customButtons: [
+        {
+          title: 'test text',
+          className: 'test-class',
+          click: function (e) {
+            $(e.target).addClass('new-class');
+          },
+          dataAttr: 'test-data'
+        }
+      ]
+    };
+    return setup(settings);
+  }
+
   function waitForBeaconChange(test) {
     return tick() //wait to read value of user input
     .then(tick)   //wait to receive ajax response
@@ -1532,7 +1548,7 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
         };
         return setup(settings).then(function (test) {
           expect(test.form.primaryAuthForm().index()).toBe(0);
-          expect(test.form.socialAuthContainer().index()).toBe(1);
+          expect(test.form.primaryAuthContainer().index()).toBe(1);
           expect(test.form.socialAuthButtons().length).toBe(4);
           expect(test.form.facebookButton().length).toBe(1);
           expect(test.form.googleButton().length).toBe(1);
@@ -1555,7 +1571,7 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
           ]
         };
         return setup(settings).then(function (test) {
-          expect(test.form.socialAuthContainer().index()).toBe(0);
+          expect(test.form.primaryAuthContainer().index()).toBe(0);
           expect(test.form.primaryAuthForm().index()).toBe(1);
           expect(test.form.socialAuthButtons().length).toBe(2);
           expect(test.form.linkedInButton().length).toBe(1);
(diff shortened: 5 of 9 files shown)


src/PrimaryAuthController.js
```diff
@@ -14,11 +14,12 @@ define([
   'okta',
   'views/primary-auth/PrimaryAuthForm',
   'views/primary-auth/SocialAuth',
+  'views/primary-auth/AddAuthButton',
   'models/PrimaryAuth',
   'shared/util/Util',
   'util/BaseLoginController'
 ],
-function (Okta, PrimaryAuthForm, SocialAuth, PrimaryAuthModel, Util, BaseLoginController) {
+function (Okta, PrimaryAuthForm, SocialAuth, AuthButton, PrimaryAuthModel, Util, BaseLoginController) {
```
mauriciocastillosilva-okta: we usually have the same name here as in the 'define' block. So this should be AddAuthButton

src/PrimaryAuthController.js
```diff
@@ -141,9 +142,14 @@ function (Okta, PrimaryAuthForm, SocialAuth, PrimaryAuthModel, Util, BaseLoginCo
 
       // Add SocialAuth view only when the idps are configured. If configured, 'socialAuthPositionTop'
       // will determine the order in which the social auth and primary auth are shown on the screen.
+      var addButtonSettings = this.model.settings.get('features.addButton');
+
       if (options.settings.get('socialAuthConfigured')) {
         this.add(SocialAuth, {prepend: options.settings.get('socialAuthPositionTop')});
       }
+      if (addButtonSettings && addButtonSettings.visibleWhen) {
+        this.add(AuthButton.render(addButtonSettings));
```
mauriciocastillosilva-okta: we should add an AuthButton, not call a render method. This should be
```
this.add(AuthButton);
```

src/models/Settings.js
```diff
@@ -76,6 +76,11 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
       'features.deviceFingerprinting': ['boolean', false, false],
       'features.hideSignOutLinkInMFA' : ['boolean', false, false],
       'features.customExpiredPassword': ['boolean', true, false],
+      'features.addButton': ['object', false, false],
```
mauriciocastillosilva-okta: not sure you need this one, since you are explicitly defining the contents of it in the next few lines

src/views/primary-auth/AddAuthButton.js
```diff
@@ -0,0 +1,49 @@
...
+define([
+    'okta'
+], function (Okta) {
+
+    var dividerTpl = Okta.tpl(
+        '<div class="auth-divider">\
+          <span class="auth-divider-text">{{text}}</span>\
+        </div>');
+
+    var render = function(buttonSettings) {
```
mauriciocastillosilva-okta: no need for this, just return the View. 
If you need to access the settings, you should be able to though the model ```this.model.get('features.addButton.text')``` and so on

src/PrimaryAuthController.js
```diff
@@ -141,9 +142,14 @@ function (Okta, PrimaryAuthForm, SocialAuth, PrimaryAuthModel, Util, BaseLoginCo
 
       // Add SocialAuth view only when the idps are configured. If configured, 'socialAuthPositionTop'
       // will determine the order in which the social auth and primary auth are shown on the screen.
+      var addButtonSettings = this.model.settings.get('features.addButton');
+
       if (options.settings.get('socialAuthConfigured')) {
         this.add(SocialAuth, {prepend: options.settings.get('socialAuthPositionTop')});
       }
+      if (addButtonSettings && addButtonSettings.visibleWhen) {
```
mauriciocastillosilva-okta: we need to evaluate if ```visibleWhen``` is a function or not. If it is, we need to call that function.

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1808,4 +1824,41 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
 
   });
 
+  Expect.describe('Additional Auth Button', function () {
+    itp('does not show the divider or buttons if addButton is not set', function () {
```
mauriciocastillosilva-okta: sp: ```divider and button```

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1808,4 +1824,41 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
...
+      return setup().then(function (test) {
+        expect(test.form.authDivider().length).toBe(0);
+        expect(test.form.additionalAuthButton().length).toBe(0);
+      });
+    });
+    itp('shows divider and buttons if properties passed with visibleWhen true', function () {
```
mauriciocastillosilva-okta: sp: singular ```button```

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1808,4 +1824,41 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
...
+        expect(test.form.additionalAuthButton().length).toBe(1);
+      });
+    });
+    itp('does not show divider and buttons if properties passed with visibleWhen false', function () {
+      return setupAdditionalAuthButton(false).then(function(test){
+        expect(test.form.authDivider().length).toBe(0);
+        expect(test.form.additionalAuthButton().length).toBe(0);
+      });
+    });
+    itp('sets text with property passed', function () {
```
mauriciocastillosilva-okta: one more test will be missing for when ```visibleWhen``` is a function

gowthamidommety-okta:
customauthbutton

src/views/primary-auth/AddAuthButton.js
```diff
@@ -0,0 +1,49 @@
...
+        return Okta.View.extend({
+
+            children: function () {
```
gowthamidommety-okta: Thanks! Will change

README.md
```diff
@@ -957,6 +957,24 @@ features: {
 
 - **features.hideSignOutLinkInMFA** - Hides the sign out link for MFA challenge. Defaults to `false`.
 
+- **features.addButton** - Adds a custom Auth button configurable by the properties below:
+
+    - `text` - String to set button text
```
gowthamidommety-okta: Added a screenshot below. This is only for the Primary Auth page. I don't think we can enable this for any other page.

test/unit/spec/PrimaryAuth_spec.js
gowthamidommety-okta: Will change

test/unit/spec/PrimaryAuth_spec.js
gowthamidommety-okta: Will change

src/PrimaryAuthController.js
mauriciocastillosilva-okta: Main concern with that is that we have a divider, that will show up above the button. If we support an array then how do we know if we add the divider or not? Or if we support adding them in different pages? It seems to me we could do that for v2.

src/PrimaryAuthController.js
gowthamidommety-okta: Makes a lot of sense. Initially I wanted to refactor Social Auth to be more generic, but SocialAuth extending CustomButtons is better

src/PrimaryAuthController.js
gowthamidommety-okta: Refactoring code from discussion we had. So this won't be an issue

src/PrimaryAuthController.js
mauriciocastillosilva-okta: Actually, thinking about it more, we should always add the divider, and we can support an array of buttons.

src/views/primary-auth/AddAuthButton.js
gowthamidommety-okta: Refactoring code from discussion we had. So this won't be an issue

gowthamidommety-okta:
Pushed new changes. Please ignore tests for now

src/PrimaryAuthController.js
```diff
@@ -141,8 +141,9 @@ function (Okta, PrimaryAuthForm, SocialAuth, PrimaryAuthModel, Util, BaseLoginCo
 
       // Add SocialAuth view only when the idps are configured. If configured, 'socialAuthPositionTop'
```
mauriciocastillosilva-okta: you could move this comment to Settings.

src/models/Settings.js
```diff
@@ -238,6 +247,57 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
...
+        cache: true
+      },
+
+      //custom buttons
+      customButtonsOptionsSet: {
+        deps: ['customButtons'],
+        fn: function (customButtons) {
+          _.each(customButtons, function(button) {
+            //add custom auth class name
+            button['className'] = button['className'] + ' default-custom-button';
```
mauriciocastillosilva-okta: I think we should also get rid of any other properties that are sent to us. Basically use a whitelist of what we need. Something like
```
_.pick(button, 'click', 'className', 'title')
```

src/models/Settings.js
```diff
@@ -238,6 +247,57 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
...
+          return customButtons;
+        },
+        cache: true
+      },
+
+      configuredButtons: {
+        deps: ['customButtonsOptionsSet', 'configuredIdpsOptionsSet', 'socialAuthConfigured'],
+        fn: function(customButtonsOptionsSet, configuredIdpsOptionsSet, socialAuthConfigured) {
+          var buttonArray = [];
+          if(socialAuthConfigured) {
```
mauriciocastillosilva-okta: you could also just check the length of ```configuredIdpsOptionsSet``` instead of ```socialAuthConfigured```. And if ```socialAuthConfigured``` is no longer used, you could even get rid of it.

src/PrimaryAuthController.js
gowthamidommety-okta: Yea, i didn't see that. It doesn't fit in settings either. I will get rid of it.

src/models/Settings.js
gowthamidommety-okta: I don't think we can.. Social Auth needs options (properties) sent to the call back.

src/models/Settings.js
gowthamidommety-okta: socialAuthConfigured is used in socialAuthPositionTop. Since we use it in more than one place , I'm making use of it here.

README.md
```diff
@@ -918,6 +918,23 @@ Options for the [OpenId Connect](http://developer.okta.com/docs/api/resources/oi
       authScheme: 'OAUTH2'
     }
     ```
+- **customButtons** - Adds an array of custom Auth buttons configurable by the properties below:
```
rchild-okta: This shouldn't be in the OIDC section, should it? I.e. it could be part of socialAuth, but you could also have a custom button outside of that flow right? Probably just create a new section (like helpLinks before this).

Also, in the style of helplinks, should reword to something like:

You can add custom buttons underneath the login form on the primary auth page by setting the following config option. If you'd like to change the divider text, use the `i18n` config option.

README.md
```diff
@@ -918,6 +918,23 @@ Options for the [OpenId Connect](http://developer.okta.com/docs/api/resources/oi
...
+
+    - `title` - String to set button text
```
rchild-okta: We *mostly* don't word things like this. Actually, this is a good reminder, @mystiberry-okta can you check the text in this PR? This is going to show up in our widget README (and later on, on d.o.c.).

Better perhaps to say something like "the button text"

README.md
```diff
@@ -918,6 +918,23 @@ Options for the [OpenId Connect](http://developer.okta.com/docs/api/resources/oi
...
+    - `className` - Sets a class on the button
```
rchild-okta: Optional class that can be added to the button

README.md
```diff
@@ -918,6 +918,23 @@ Options for the [OpenId Connect](http://developer.okta.com/docs/api/resources/oi
...
+    - `dataAttr` - Sets data-se attribute on the button
```
rchild-okta: I don't think we want to expose this publicly - data-se is our own internal data attr for selenium testing. They'd never need to set this.

README.md
```diff
@@ -918,6 +918,23 @@ Options for the [OpenId Connect](http://developer.okta.com/docs/api/resources/oi
...
+    - `click` - Callback function that is executed when button is clicked
```
rchild-okta: Function that is called when the button is clicked

README.md
```diff
@@ -918,6 +918,23 @@ Options for the [OpenId Connect](http://developer.okta.com/docs/api/resources/oi
...
+
+    ```javascript
+    customButtons: [{
+      title: 'Click Me',
+      className: 'btn-customAuth',
+      dataAttr: 'custom-data',
+      click: function() {
+        //custom callback
```
rchild-okta: nit: space between // and custom 

I do think it's more useful to add a real example, i.e. you could show the code to redirect to a different url (to give more context on what they'll actually be doing. With a comment before it saying that this is an example showing a redirect to bla.

src/PrimaryAuthController.js
```diff
@@ -139,10 +139,11 @@ function (Okta, PrimaryAuthForm, SocialAuth, PrimaryAuthModel, Util, BaseLoginCo
 
       this.addListeners();
 
-      // Add SocialAuth view only when the idps are configured. If configured, 'socialAuthPositionTop'
-      // will determine the order in which the social auth and primary auth are shown on the screen.
-      if (options.settings.get('socialAuthConfigured')) {
-        this.add(SocialAuth, {prepend: options.settings.get('socialAuthPositionTop')});
+      // If social auth is configured, 'socialAuthPositionTop' will determine
+      // the order in which the social auth and primary auth are shown on the screen.
+
```
rchild-okta: nit: remove this newline (i.e. comment should be attached to the code it's talking about)

src/models/Settings.js
```diff
@@ -238,6 +241,57 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
           return !!(socialAuthConfigured && idpDisplay.toUpperCase() === 'PRIMARY');
         },
         cache: true
+      },
+      configuredIdpsOptionsSet: {
+        deps: ['configuredSocialIdps'],
+        fn: function (configuredSocialIdps) {
+          _.each(configuredSocialIdps, function(idp) {
```
rchild-okta: In general, we shouldn't be modifying the object they pass in - low risk, but it's possible they are doing something with it as well and changing the properties on the object could affect the developer's code.

```javascript
return _.map(configuredSocialIdps, function (idp) {
  var type = idp.type.toLowerCase();
  return {
    // the properties + idp properties
    // alternatively, clone the idp and then add these properties to it
  };
});
```

And a nit - no need to use `idp['dataAttr']`, use `idp.dataAttr`

src/models/Settings.js
rchild-okta: Do we need both the idp & customButtons derived properties? Can we just combine this in one? Also, agree, we need to whitelist the properties that are sent to us. Similar to the other comment - but create new objects here, don't modify the existing one they pass in. Doing this, we can just pick and choose (like @mauriciocastillosilva-okta suggested) what we want to take from it).

src/models/Settings.js
rchild-okta: Yeah, just reiterating - since the customButtonOptions and idpOptions are only really used for this (I'm guessing), let's just pack it all in one derived property.

src/views/primary-auth/CustomButtons.js
```diff
@@ -56,21 +55,19 @@ define([
     },
 
     _createButton: function (options) {
-      var type = options.type.toLowerCase(),
-          attr = 'social-auth-' + type + '-button';
-
+      if(options.isSocialAuth) {
+        options.click =  function(e) {
```
rchild-okta: Why not just do the click function in settings.js? In general, better to have things look the same - in this case, SocialAuth is using the generic CustomButtons class, rather than having these edge cases built into this generic view.

src/views/primary-auth/CustomButtons.js
```diff
@@ -27,16 +27,15 @@ define([
...
 
-    className: 'social-auth',
+    className: 'primary-auth-container',
 
     children: function () {
       var children = [],
-          idProviders = this.settings.get('configuredSocialIdps'),
-          divider = dividerTpl({text: Okta.loc('socialauth.divider.text', 'login')});
+        buttons = this.settings.get('configuredButtons'),
+        divider = dividerTpl({text: Okta.loc('authbutton.divider.text', 'login')});
```
rchild-okta: Hmm, I am a little hesitant about this change. It is a breaking change - should think about this more (i.e. if someone updates the widget, is customizing this string, and is using socialAuth, this will result in a change of behavior for them).

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -127,6 +127,22 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
     });
   }
 
+  function setupAdditionalAuthButton() {
```
rchild-okta: General: Since we have e2e tests, might be nice to put a new one in there as well.

src/views/primary-auth/CustomButtons.js
gowthamidommety-okta: It's difficult to have this in Settings.js because the callback references other classes as well. See OAuth2Util.

src/views/primary-auth/CustomButtons.js
gowthamidommety-okta: Spoke to @mauriciocastillosilva-okta about this. I'll change it back to use socialauth.divider.text.

test/unit/spec/PrimaryAuth_spec.js
mauriciocastillosilva-okta: Do this qualifies for an e2e test? At the end we are just calling the callback they sent us, nothing else happens in the widget (and we don't get a token back).

gowthamidommety-okta:
Made more changes. @mauriciocastillosilva-okta @rchild-okta Please review. Some notes:

**Settings.js:** 
- Removed socialAuthConfigured and its dependancies for simplification.
- Did not remove socialAuthPositionTop as it is used in PrimaryAuthController , PrimaryAuthForm and the CustomButtons file.
- Consolidated some of the other new dependancies

**PrimaryAuthForm.js:**
- removed the check for socialAuthConfigured as socialAuthPositionTop is derived from it anyway

**tests**
- Added some tests to check undefined

src/models/Settings.js
```diff
@@ -223,19 +227,65 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
...
+                title: button.title,
+                className: button.className + ' default-custom-button',
+                click: button.click
+              };
+              customConfiguredButtons.push(customConfiguredButton);
+            });
+          }
+          var buttonArray = [];
+          if(!_.isEmpty(socialAuthButtons)) {
+            buttonArray = buttonArray.concat(socialAuthButtons);
```
mauriciocastillosilva-okta: can we, instead of concatenating 2 arrays here, use the same array since the beginning and skip the concat part? At the end we just need to add buttons to the same array and we can return that
```
var buttons = [];
// add social buttons to _buttons_
// add custom buttons to _buttons_
return buttons;
```

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1808,4 +1824,105 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
 
   });
 
+  Expect.describe('Additional Auth Button', function () {
+    itp('does not show the divider and buttons if addButton is not set', function () {
```
mauriciocastillosilva-okta: should read ```...  buttons if settings.customButtons is not set```

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1808,4 +1824,105 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
...
+      return setup().then(function (test) {
+        expect(test.form.authDivider().length).toBe(0);
+        expect(test.form.additionalAuthButton().length).toBe(0);
+      });
+    });
+    itp('show the divider and buttons if addButton set', function () {
```
mauriciocastillosilva-okta: should read ```...  buttons if settings.customButtons is not empty```

gowthamidommety-okta:
Please review again!

src/models/Settings.js
```diff
@@ -223,19 +227,55 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
...
-      // checks if there are any valid configured idps.
-      socialAuthConfigured: {
-        deps: ['configuredSocialIdps'],
-        fn: function (idps) {
-          return !_.isEmpty(idps);
+      // social auth buttons order - 'above'/'below' the primary auth form (boolean)
+      socialAuthPositionTop: {
+        deps: ['configuredSocialIdps', 'idpDisplay'],
+        fn: function (configuredSocialIdps, idpDisplay) {
+          return !!(!_.isEmpty(configuredSocialIdps) && idpDisplay.toUpperCase() === 'PRIMARY');
```
mauriciocastillosilva-okta: nit: not sure if you need !! at the beginning. Both conditions return a boolean...

README.md
```diff
@@ -776,6 +776,28 @@ helpLinks: {
...
+You can add custom buttons underneath the login form on the primary auth page by setting the following config options. If you'd like to change the divider text, use the `i18n` config option.
+
+```javascript
+// An example that adds a custom button underneath the login form on the primary auth page
+customButtons: [{
+  title: 'Click Me',
+  className: 'btn-customAuth',
+  click: function() {
+   // clicking on the button navigates to another page
+   window.location.href = 'http://www.example.com'
```
rchild-okta: nit: spacing and semicolon at the end

README.md
```diff
@@ -776,6 +776,28 @@ helpLinks: {
 
 - **helpLinks.custom** - Array of custom link objects `{text, href}` that will be added to the *"Need help signing in?"* section.
 
+## Adding Custom Buttons
```
rchild-okta: Add to menu up top

src/models/Settings.js
```diff
@@ -223,19 +227,55 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
...
         },
         cache: true
       },
-      // social auth buttons order - 'above'/'below' the primary auth form (boolean)
-      socialAuthPositionTop: {
-        deps: ['socialAuthConfigured', 'idpDisplay'],
-        fn: function (socialAuthConfigured, idpDisplay) {
-          return !!(socialAuthConfigured && idpDisplay.toUpperCase() === 'PRIMARY');
+      //merges social auth and custom buttons into one array
```
rchild-okta: nit: space after comment //

src/models/Settings.js
```diff
@@ -223,19 +227,55 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
...
+              className: 'social-auth-button ' + dataAttr,
+              title: idp.title = Okta.loc('socialauth.' + type + '.label'),
+              click: function (e) {
+                e.preventDefault();
+                OAuth2Util.getTokens(self, {idp: idp.id});
+              }
+            };
+            buttonArray.push(socialAuthButton);
+          });
+          if(_.isArray(customButtons)) {
```
rchild-okta: Do we actually need the check? I.e. I'm guessing our validator should handle it, and the default value is an empty array.

gowthamidommety-okta: When i set customButtons as any string in .widgetrc, it renders a lot of buttons. I'm not sure if this is  a valid use case, but i put in a _.isArray check to stop it. Spoke to Mau about it too, and we decided there is no harm in leaving it in

PR #201, opened 2017-03-30
:seedling: Add show, hide, and remove

lboyette-okta:
@rchild-okta @haishengwu-okta

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,9 @@ Contributors should read our [contributing guidelines](./CONTRIBUTING.md) if the
 * [API](#api)
   * [OktaSignIn](#new-oktasigninconfig)
   * [renderEl](#rendereloptions-success-error)
+  * [hide](#hide)
+  * [show](#show)
+  * [remove](#remove)
   * [on](#onevent-callback-context)
   * [off](#offevent-callback)
   * [session.get](#sessiongetcallback)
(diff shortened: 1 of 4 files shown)


src/widget/OktaSignIn.js
```diff
@@ -80,15 +80,34 @@ var OktaSignIn = (function () {
...
 
+    function hide() {
+      if (router) {
+        router.hide();
+      }
+    }
+
+    function show() {
+      if (router) {
+        router.show();
```
lboyette-okta: I don't think we should throw an error if the router doesn't exist.

README.md
```diff
@@ -230,6 +233,30 @@ signIn.renderEl(
 );
 ```
 
+## hide
+
+Hide the widget, but keep the widget on the DOM.
```
rchild-okta: nit: in the dom

src/widget/OktaSignIn.js
```diff
@@ -80,15 +80,38 @@ var OktaSignIn = (function () {
      * @param success - success callback function
      * @param error - error callback function
      */
+    var router;
     function render(options, success, error) {
-      var router = new LoginRouter(_.extend({}, config, options, {
+      if (router) {
+        throw 'Cannot run two instances of the widget simultaneously. Call remove() first.';
```
rchild-okta: nit, maybe word this like:

An instance of the widget has already been rendered. Call remove() first.

PR #203, opened 2017-03-30
:bug: Isolate widget css from guest page css

lboyette-okta:
@rchild-okta 
@haishengwu-okta

### DIFF
--- a/assets/sass/_base.scss
+++ b/assets/sass/_base.scss
@@ -1,20 +1,3 @@
-html {
-  height: 100%;
-}
-
-body {
-  font-family: $fonts;
-  font-size: $font-size;
-  font-weight: $font-weight-body;
-  line-height: 1.4;
-  background-color: $secondary-bg-color;
-  min-width: 100%;
-  min-height: 100%;
-  color: $medium-text-color;
-  position: relative;
-  overflow: auto;
-}
-
 a {
   color: $medium-text-color;
 
--- a/assets/sass/_container.scss
+++ b/assets/sass/_container.scss
@@ -1,0 +1,52 @@
+// scss-lint:disable IdSelector
+#okta-sign-in {
+// scss-lint:enable IdSelector
+
+  // Html and body styles in _base.scss that were moved here
+  font-family: $fonts;
+  font-size: $font-size;
+  font-weight: $font-weight-body;
+  line-height: 1.4;
+  background-color: $secondary-bg-color;
+  min-height: 100%;
+  color: $medium-text-color;
+  position: relative;
+  overflow: auto;
+
+  // Auth-container styles
+  border-radius: 3px;
+  border-style: solid;
+  border-width: 1px;
+  height: auto;
+  margin: 100px auto 8px;
+  width: $container-width;
+  min-width: $container-min-width;
+  @include box-sizing(border-box);
+
+  @include device-mq(x-small) {
+    margin-top: 0;
+    width: 100%;
+  }
+
+  @include max-width-mq($container-width) {
+    width: 100%;
+  }
+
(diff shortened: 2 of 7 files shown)


lboyette-okta:
I haven't yet tested the affect on okta-core, but @alanzhou-okta opened this PR quite a while ago: https://github.com/okta/okta-core/pull/17361

assets/sass/okta-theme.scss
```diff
@@ -58,7 +58,7 @@
 
 
 // scss-lint:disable IdSelector
-#okta-sign-in {
+#okta-sign-in.auth-container {
```
lboyette-okta: I made the theme more specific. The argument is that themes should be able to override the base at any point.

assets/sass/_base.scss
```diff
@@ -1,20 +1,3 @@
...
-  line-height: 1.4;
-  background-color: $secondary-bg-color;
-  min-width: 100%;
-  min-height: 100%;
-  color: $medium-text-color;
-  position: relative;
-  overflow: auto;
-}
-
 a {
```
rchild-okta: These should also be specific to the widget right (the remaining classes in this file)?

assets/sass/_container.scss
```diff
@@ -0,0 +1,52 @@
+// scss-lint:disable IdSelector
+#okta-sign-in {
+// scss-lint:enable IdSelector
+
+  // Html and body styles in _base.scss that were moved here
+  font-family: $fonts;
+  font-size: $font-size;
+  font-weight: $font-weight-body;
+  line-height: 1.4;
+  background-color: $secondary-bg-color;
```
rchild-okta: Hmm, so some of these don't make sense here (and in general, will just need to be ported to okta-core), i.e:
- Is background-color here for the container, or for the background page?
- min-height - we're setting the height below (and don't really want the widget container to be 100%)
- position: relative and overflow: auto were fixes on the container for embedded scenarios - probably don't belong here, and will definitely need to be tested in okta-core when we port these over

assets/sass/_base.scss
lboyette-okta: I agree

assets/sass/_container.scss
lboyette-okta: - I think originally it came from the page background. `background-color` can be removed
- `min-height` can be removed. What's odd is that this renders normally.
- `position` and `overflow` can be removed

assets/sass/_base.scss
lboyette-okta: yeah, I guess it doesn't make sense to create _container.scss at all. Everything can just be namespaced in _base.scss

lboyette-okta: wait, they are namespaced - they're imported within an `#okta-sign-in` tag in okta-sign-in.scss

assets/sass/_container.scss
```diff
@@ -0,0 +1,52 @@
...
+    width: 100%;
+  }
+
+  @include max-height-mq(750px) {
+    margin-top: 0;
+  }
+}
+
+// scss-lint:disable IdSelector
+#okta-sign-in.no-beacon .auth-content {
```
haishengwu-okta: nested into okta-sign-in namespace?

lboyette-okta: the `.no-beacon` makes it so I can't. However, I can combine the `.auth-content` class into the same namespace with the `.auth-header` below

haishengwu-okta: I thought we could nest it like `&.no-beacon`

```
#foo {
  &.bar .baz {
    color: red;
  }
}
```
 
generates

```
#foo.bar .baz {
  color: red;
}
```

PR #208, opened 2017-04-06
Add shared Form to modify form submit events for correct work in OracleTap mobile app

alexeystadnikov-okta:
Resolves OKTA-118336

screenshot_20170405-150756

Browser inside the Oracle Tap app ignores `e.preventDefault()` and submits form reloading a page.
For other events `e.preventDefault()` works as expected.

Please, review the solution and give your comments.

@rchild-okta 
@mauriciocastillosilva-okta


rchild-okta:
Can you explain what this is doing? I.e. as Mau said, why can't we do this in the event handler itself? Also, your description states that the Tap app ignores preventDefault - how does this PR help with that?

alexeystadnikov-okta:
The form can be submitted by the following ways:
- click on submit button
- hit enter inside a text input
- `form.submit()` in js

For the first two options, it triggers `click` event on a submit button, that we can work with.
So, I can call `__save` on `click`, not on `submit`.
**This is only one solution I have found.**

In the issue they use the only PrimaryAuthForm, so we don't have to change it everywhere.

If my approach is OK, I will fix tests. They are failed because of `form.submit` call triggers only `submit` event without `click` on a submit button.

rchild-okta:
I'm still not following - can you describe the exact scenario where the current code doesn't work?

It sounds like it might be this:
- The user hits enter, which triggers a submit event
- We have a handler in our code somewhere that does a preventDefault on this event, but somehow it doesn't work in this particular Tap app
- The solution is to disable the original click event handler and replace it with a function that just does preventDefault (w/out calling the underlying function)?

alexeystadnikov-okta:
Oracle Tap ignores `preventDefault` for the submit event.
I checked it on a blank page with the simple form with one input.

When the user hits enter, the form triggers `click` event on a first submit button inside the form.
My solution is to move handler for the `submit` event to `click [type=submit]`.

rchild-okta:
Gotcha. Main question for me now is why preventDefault doesn't work for the submit event. Have you tried things like returning false in the handler function (i.e. the old way to prevent default actions).

alexeystadnikov-okta:
Here is my code:
```
<form id="formX" action="" onsubmit="return false" method="POST">
	<input type="text" name="name"/>
	<button id="submitX" type="submit">XXX</button>
</form>
<script type="text/javascript">
document.getElementById('formX').addEventListener('submit', function(e){
	e.preventDefault();
	e.stopPropagation();
	return false;
});
</script>
```

mauriciocastillosilva-okta:
It is really weird that this Oracle Tap does not work with all those different ways to stop the event. But I'm not sure overriding the Form is the right way to go. A quick research tells me that we should try ```stopImmediatePropagation()```, have you try that?

alexeystadnikov-okta:
Tried `stopImmediatePropagation`. Did not help.

alexeystadnikov-okta:
Added alternate solution.

@rchild-okta
@mauriciocastillosilva-okta
Please, take a look.

src/views/shared/Form.js
```diff
@@ -0,0 +1,28 @@
...
+
+define([
+  'okta'
+], function (Okta) {
+
+  var _ = Okta._;
+
+  return Okta.Form.extend({
+    constructor: function (options) {
+      var target = this.attributes.target = _.uniqueId('iframe');
```
mauriciocastillosilva-okta: can we move this to the PrimaryAuthForm.js itself? 

Also, this bug is still tricky to me. Can you explain why adding the iframe helps?

alexeystadnikov-okta: Done.

If we use form.target it submits the form inside the iframe and doesn't redirect the current page to another url.

mauriciocastillosilva-okta:
This looks better, I think this could work
@rchild-okta thoughts?

src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -25,6 +25,13 @@ define([
     saveId: 'okta-signin-submit',
     layout: 'o-form-theme',
 
+    constructor: function (options) {
+      var target = this.attributes.target = _.uniqueId('iframe');
```
rchild-okta: Let's add a comment here explaining why we're doing this.

alexeystadnikov-okta: Comment added.

alexeystadnikov-okta:
In the solution with iframe we don't need to override events.
There should be only one of them and the solution with iframe looks more stable and not so unreliable like with events.

alexeystadnikov-okta:
Test added

PR #209, opened 2017-04-11
:bug: Fixes flakey tests

gowthamidommety-okta:
Fixes a windows hello test.

### DIFF
--- a/test/unit/spec/MfaVerify_spec.js
+++ b/test/unit/spec/MfaVerify_spec.js
@@ -189,22 +189,15 @@ function (Okta,
       Util.mockRouterNavigate(router);
       setNextResponse([resRequiredWebauthn, resChallengeWebauthn, resSuccess]);
       router.refreshAuthState('dummy-token');
-      var $forms = $sandbox.find('.o-form');
-      var forms = _.map($forms, function (form) {
-        return new MfaVerifyForm($(form));
-      });
-      if (forms.length === 1) {
-        forms = forms[0];
-      }
-      var beacon = new Beacon($sandbox);
-      return {
-        router: router,
-        form: forms,
-        beacon: beacon,
-        ac: authClient,
-        setNextResponse: setNextResponse,
-        successSpy: successSpy
-      };
+      return Expect.waitForVerifyWindowsHello()
+      .then(function() {
+        return Expect.waitForSpyCall(successSpy);
+      })
+      .then(function () {
+        return {
+          router: router
+        }
+      })
     }
 
     var setupSecurityQuestion = _.partial(setup, resAllFactors, { factorType: 'question' });
@@ -2032,12 +2025,6 @@ function (Okta,
         itp('automatically triggers Windows Hello only once', function () {
           return emulateWindows()
           .then(setupWebauthnOnly)
-          .then(function (test) {
-            return Expect.waitForVerifyWindowsHello(test);
-          })
-          .then(function (test) {
-            return Expect.waitForSpyCall(test.successSpy, test);
-          })
           .then(function (test) {
             expect(test.router.controller.model.get('__autoTriggered__')).toBe(true);
             spyOn(test.router.controller.model, 'save');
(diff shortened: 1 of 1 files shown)


gowthamidommety-okta:
@mauriciocastillosilva-okta @rchild-okta  Please review

test/unit/spec/MfaVerify_spec.js
```diff
@@ -174,6 +174,39 @@ function (Okta,
       });
     }
 
+    function setupWebauthnOnly() {
```
rchild-okta: Can you explain the reason why we need to break this out into a separate function (which looks very similar to the other setup function)?

gowthamidommety-okta: The other setup function has `Expect.waitForMfaVerify()`, and for this particular test, where Windows Hello is the only factor, it never goes to that controller.

test/unit/spec/MfaVerify_spec.js
```diff
@@ -174,6 +174,39 @@ function (Okta,
...
+      var setNextResponse = Util.mockAjax();
+      var baseUrl = 'https://foo.com';
+      var authClient = new OktaAuth({url: baseUrl, transformErrorXHR: LoginUtil.transformErrorXHR});
+      var successSpy = jasmine.createSpy('success');
+      var router = new Router(_.extend({
```
mauriciocastillosilva-okta: can we refactor part of this function into one that returns the router? And that way less code is duplicated

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2002,9 +2033,15 @@ function (Okta,
           return emulateWindows()
           .then(setupWebauthnOnly)
           .then(function (test) {
+            return Expect.waitForVerifyWindowsHello(test);
```
mauriciocastillosilva-okta: can we move this ```Expect.waitForVerifyWindowsHello``` into ```setupWebauthnOnly````?

test/unit/spec/MfaVerify_spec.js
```diff
@@ -174,6 +174,39 @@ function (Okta,
...
+      Util.registerRouter(router);
+      Util.mockRouterNavigate(router);
+      setNextResponse([resRequiredWebauthn, resChallengeWebauthn, resSuccess]);
+      router.refreshAuthState('dummy-token');
+      var $forms = $sandbox.find('.o-form');
+      var forms = _.map($forms, function (form) {
+        return new MfaVerifyForm($(form));
+      });
+      if (forms.length === 1) {
+        forms = forms[0];
```
mauriciocastillosilva-okta: can you tell me why we need this? This setup function is only used in one test, so we probably know if this form is present or not.

test/unit/spec/MfaVerify_spec.js
gowthamidommety-okta: I think creating another function here may cause a lint error similar to my first commit.

`/home/travis/build/okta/okta-signin-widget/test/unit/spec/MfaVerify_spec.js
  114:33  error  Function has too many statements (35). Maximum allowed is 34  max-statements`
 
Should I create another describe? Is there a better way around it?

test/unit/spec/MfaVerify_spec.js
gowthamidommety-okta: Looking at this, I actually think we don't need any of it. Our test only references router, so returning a reference to that should be enough.

test/unit/spec/MfaVerify_spec.js
gowthamidommety-okta: Will change

test/unit/spec/MfaVerify_spec.js
mauriciocastillosilva-okta: you can change, at the top of the file, the allowed max-statements, and set it to 35, or whatever number you need

gowthamidommety-okta: 👍 done

PR #210, opened 2017-04-12
Update courage to bring in dirty inline validation

santhoshbalakrishnan:
## Reviewers

- @mauriciocastillosilva-okta 
- @alexeystadnikov-okta 
- @rchild-okta 

## Screenshare

signin

signin-forgot

**Upon save**

signin-normal

### DIFF
--- a/node_modules/@okta/jquery.simplemodal/package.json
+++ b/node_modules/@okta/jquery.simplemodal/package.json
@@ -2,13 +2,12 @@
   "_args": [
     [
       "@okta/jquery.simplemodal@1.4.9",
-      "/home/vagrant/okta/okta-signin-widget/node_modules/@okta/courage"
+      "/home/vagrant/okta/okta-signin-widget"
     ]
   ],
   "_from": "@okta/jquery.simplemodal@1.4.9",
   "_id": "@okta/jquery.simplemodal@1.4.9",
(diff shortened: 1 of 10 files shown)


npm-shrinkwrap.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.10.0",
   "dependencies": {
     "@okta/courage": {
-      "version": "2.1.0"
+      "version": "2.5.0-beta.0"
```
rchild-okta: Hmm, I'm not sure if we have a rule for this, but why not publish a new version of courage and include a release version in here? It's not that big of a deal now, but that's the ideal process in the future; doesn't seem like too much extra work to standardize on it.

santhoshbalakrishnan:
@mauriciocastillosilva-okta @rchild-okta Done :)

node_modules/@okta/jquery.simplemodal/package.json
```diff
@@ -20,13 +19,14 @@
     "type": "version"
   },
   "_requiredBy": [
+    "/",
```
mauriciocastillosilva-okta: not sure this is right, since simplemodal is only requred by courage. 
But if Travis is passing it should be ok... at the end / means it could be required by whoever.

PR #211, opened 2017-04-18
:seedling: Handle new UNAUTHENTICATED state

mauriciocastillosilva-okta:
Widget loads PrimaryAuth view when UNAUTHENTICATED is returned

Resolves: OKTA-121731
Bacon: test

### DIFF
--- a/src/models/AppState.js
+++ b/src/models/AppState.js
@@ -482,16 +482,10 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
           return policy && policy.rememberDeviceByDefault;
         }
       },
-      'isUnauthenticatedState' :{
+      'unauthenticatedToken': {
         deps: ['transaction'],
         fn: function (transaction) {
-          return transaction && transaction.status === 'UNAUTHENTICATED';
-        }
-      },
-      'unauthenticatedToken' :{
-        deps: ['isUnauthenticatedState', 'transaction'],
-        fn: function (isUnauthenticatedState, transaction) {
-          if (isUnauthenticatedState && transaction.data) {
+          if (transaction && transaction.status === 'UNAUTHENTICATED' && transaction.data) {
             return transaction.data.stateToken;
           }
           return null;
--- a/src/models/AppState.js
+++ b/src/models/AppState.js
@@ -181,6 +181,12 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
           return res.status === 'MFA_CHALLENGE';
         }
       },
+      'isUnauthenticated': {
+        deps: ['lastAuthResponse'],
+        fn: function (res) {
+          return res.status === 'UNAUTHENTICATED';
+        }
+      },
       'isMfaRejectedByUser': {
         // MFA failures are usually error responses
         // except in the case of Okta Push, when a
@@ -480,12 +486,6 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         deps: ['policy'],
         fn: function (policy) {
           return policy && policy.rememberDeviceByDefault;
-        }
-      },
-      'isUnauthenticatedStatus': {
-        deps: ['transaction'],
-        fn: function (transaction) {
-          return transaction && transaction.status === 'UNAUTHENTICATED';
         }
       }
     },
--- a/src/models/AppState.js
+++ b/src/models/AppState.js
@@ -482,13 +482,10 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
           return policy && policy.rememberDeviceByDefault;
         }
       },
-      'unauthenticatedToken': {
+      'isUnauthenticatedStatus': {
         deps: ['transaction'],
         fn: function (transaction) {
-          if (transaction && transaction.status === 'UNAUTHENTICATED' && transaction.data) {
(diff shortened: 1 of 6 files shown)


src/models/AppState.js
```diff
@@ -481,6 +481,21 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
...
+        }
+      },
+      'unauthenticatedToken' :{
+        deps: ['isUnauthenticatedState', 'transaction'],
+        fn: function (isUnauthenticatedState, transaction) {
+          if (isUnauthenticatedState && transaction.data) {
+            return transaction.data.stateToken;
+          }
+          return null;
+        }
```
gowthamidommety-okta: If `isUnauthenticatedState` is only used in `unauthenticatedToken`, we should probably club this into a single dependency.

src/models/PrimaryAuth.js
```diff
@@ -91,14 +91,18 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
...
           password: password,
           options: {
             warnBeforePasswordExpired: true,
             multiOptionalFactorEnroll: multiOptionalFactorEnroll
           }
-        })
+        };
+        if (this.appState.get('unauthenticatedToken')) {
+          signInArgs.stateToken = this.appState.get('unauthenticatedToken');
+        }
```
gowthamidommety-okta: if `this.appState.get('unauthenticatedToken')` is null, `signInArgs.stateToken` will be undefined. As long as the API handles this, it should be fine :) Maybe write a test for this case to ensure it works.

mauriciocastillosilva-okta: API is good with an empty ```stateToken```, and I added a test for when the stateToken is null in an UNAUTHENTICATED state

mauriciocastillosilva-okta:
@gowthamidommety-okta addressed feedback, take a look again when you can please.

@rchild-okta when you have time, take a look please.

src/models/AppState.js
```diff
@@ -481,6 +481,15 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         fn: function (policy) {
           return policy && policy.rememberDeviceByDefault;
         }
+      },
+      'unauthenticatedToken': {
+        deps: ['transaction'],
+        fn: function (transaction) {
+          if (transaction && transaction.status === 'UNAUTHENTICATED' && transaction.data) {
```
rchild-okta: What's the actual flow on the Okta side? In general, the sign-in-widget shouldn't know about the stateToken, which is handled by AuthJs. In the flows where we bootstrap with a state (i.e. by passing a stateToken into the widget), we pass it to AuthJs via a resume call, but at that point it's loaded into AuthJs and will be attached to any subsequent request. 

If AuthJs is not passing a saved stateToken in the primaryAuth call, it feels like this fix should go there instead (rather than passing it in via the widget).

mauriciocastillosilva-okta: The flow is for login into an app that uses a custom login url (this is configured in the App settings, general tab). 
So the user goes to the deep link for the app login (something like http://org.com/home/salesforce/{sha}) and then we redirect to the custom login, but the url at this times comes with a stateToken. So we bootstrap the widget with such stateToken, and in the first call to the authn api we get back an UNAUTHENTICATED state.
I think we could do that with the AuthJs if we want to keep the stateToken out of the signinwidget world. As of now, my understanding is that for primaryAuth our AuthJs does not include the stateToken. We had a meeting with Karl and Len and all the action items were on the signin widget side.
@lboyette-okta ^

mauriciocastillosilva-okta:
@rchild-okta I implemented this properly now, calling the link we get in the UNAUTHENTICATED response instead of directly calling primaryAuth. Please take a look again

src/models/AppState.js
```diff
@@ -481,6 +481,12 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         fn: function (policy) {
           return policy && policy.rememberDeviceByDefault;
         }
+      },
+      'isUnauthenticatedStatus': {
```
rchild-okta: Nit: Can we call this "isUnauthenticated"? I.e. similar to how we do other status responses up top (i.e. isMfaChallenge, etc). Also might want to just bump this up closer to those, since it's in the same category.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -385,6 +387,27 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, OktaAuth, Util,
         expect(form.hasErrors()).toBe(false);
       });
     });
+    itp('navigates to PrimaryAuth if status is UNAUTHENTICATED', function () {
+      return setup({ stateToken: 'aStateToken' })
+      .then(function (test) {
+        Util.mockRouterNavigate(test.router);
+        test.setNextResponse(resUnauthenticated);
+        test.router.navigate('');
```
rchild-okta: Am curious if this will immediately-ish resolve (without checking for anything) since we're starting on PrimaryAuth. Maybe for testing purposes, it might be better to start out on a different page.

mauriciocastillosilva-okta: good point. I fixed it

PR #212, opened 2017-04-18
Prevent showing tooltip if security image is not visible

alexeystadnikov-okta:
Resolves: OKTA-120613

- Don't show qtip if securityImage is not visible.
- Destroy qtip if window is resized and securityImage become not visible

@mauriciocastillosilva-okta 
@anishashetty-okta 
Please, review.

### DIFF
--- a/src/views/shared/SecurityBeacon.js
+++ b/src/views/shared/SecurityBeacon.js
@@ -41,6 +41,7 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
       el.removeAttr('aria-describedby');
       el.find('.accessibility-text').text(imgDescription);
       el.css('background-image', 'url(' + _.escape(imgSrc) + ')');
+      return;
     }
   }
 
--- a/src/views/shared/SecurityBeacon.js
+++ b/src/views/shared/SecurityBeacon.js
@@ -41,11 +41,14 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
       el.removeAttr('aria-describedby');
       el.find('.accessibility-text').text(imgDescription);
       el.css('background-image', 'url(' + _.escape(imgSrc) + ')');
-      return;
     }
   }
 
   function antiPhishingMessage (image, host, shown) {
+    if(!image.is(':visible')){
+      return;
+    }
+
     // Show the message that the user has not logged in from this device before.
     image.qtip({
       prerender: true,
@@ -64,7 +67,14 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
         viewport: $('body')
       },
       hide: {event: false, fixed: true},
-      show: {event: false, delay: 200}
+      show: {event: false, delay: 200},
+      events: {
+        move: function(event, api) {
+          if (!api.elements.target.is(':visible')) {
+            api.destroy(true);
+          }
+        }
+      }
     });
     image.qtip('toggle', shown);
   }
--- a/src/views/shared/SecurityBeacon.js
+++ b/src/views/shared/SecurityBeacon.js
@@ -46,12 +46,11 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
   }
 
   function antiPhishingMessage (image, host) {
-    function showQtip() {
+    $(window).on('resize.securityBeaconQtip', _.debounce(function () {
       if(image.is(':visible')){
         image.qtip('show');
       }
-    }
-    $(window).on('resize.securityBeaconQtip', _.debounce(showQtip, 300));
+    }, 300));
 
     // Show the message that the user has not logged in from this device before.
     image.qtip({
@@ -84,7 +83,7 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
       }
     });
 
-    showQtip();
+    image.qtip('toggle', image.is(':visible'));
   }
 
   function destroyAntiPhishingMessage(image) {
--- a/src/views/shared/SecurityBeacon.js
+++ b/src/views/shared/SecurityBeacon.js
@@ -45,10 +45,13 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
     }
   }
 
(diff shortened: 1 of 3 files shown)


alexeystadnikov-okta:
@mauriciocastillosilva-okta
@anishashetty-okta

Do you have any comments?

src/views/shared/SecurityBeacon.js
```diff
@@ -64,7 +67,14 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
         viewport: $('body')
       },
       hide: {event: false, fixed: true},
-      show: {event: false, delay: 200}
+      show: {event: false, delay: 200},
+      events: {
+        move: function(event, api) {
```
mauriciocastillosilva-okta: do we need this event? I would think that with the lines you added at the beginning of this function (48-51) it should all work

src/views/shared/SecurityBeacon.js
```diff
@@ -41,11 +41,14 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
       el.removeAttr('aria-describedby');
       el.find('.accessibility-text').text(imgDescription);
       el.css('background-image', 'url(' + _.escape(imgSrc) + ')');
-      return;
```
mauriciocastillosilva-okta: nit: this seems to me doesn't affect the method, so we could probably leave it as it is in case someone in the future adds another ```if``` block

src/views/shared/SecurityBeacon.js
alexeystadnikov-okta: We need it for the case if the window is resized and Beacon image disappears.

src/views/shared/SecurityBeacon.js
alexeystadnikov-okta: Done

src/views/shared/SecurityBeacon.js
mauriciocastillosilva-okta: I see... What if the window is resized again and the security image is visible again? Maybe we should hide/show the qtip based on the same css as we hide/show the security image when we resize the window. Can you look into that please?

src/views/shared/SecurityBeacon.js
```diff
@@ -64,9 +71,25 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
...
+            api.set('hide.effect', false);
+            api.hide();
+            api.set('hide.effect', true);
+          }
+        }
+      }
     });
-    image.qtip('toggle', shown);
+
+    showQtip();
```
mauriciocastillosilva-okta: nit: ```image.qtip('toggle', image.is(':visible'));```

src/views/shared/SecurityBeacon.js
```diff
@@ -45,7 +45,14 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
...
   }
 
-  function antiPhishingMessage (image, host, shown) {
+  function antiPhishingMessage (image, host) {
+    function showQtip() {
+      if(image.is(':visible')){
+        image.qtip('show');
+      }
+    }
+    $(window).on('resize.securityBeaconQtip', _.debounce(showQtip, 300));
```
mauriciocastillosilva-okta: nit: you could avoid creating a new function
```
 _.debounce(image.qtip('toggle', image.is(':visible')), 300)
```

src/views/shared/SecurityBeacon.js
```diff
@@ -110,7 +134,9 @@ define(['okta', 'util/Animations'], function (Okta, Animations) {
       }).then(function () {
         border.addClass('auth-beacon-border');
       }).then(function () {
-        antiPhishingMessage(image, host, hasAntiPhishing);
+        if (hasAntiPhishing) {
```
mauriciocastillosilva-okta: 👍

src/views/shared/SecurityBeacon.js
alexeystadnikov-okta: Anyway, I have to create a function for using inside the `_.debounce`, o why not using it twice?

src/views/shared/SecurityBeacon.js
alexeystadnikov-okta: Commented above.

src/views/shared/SecurityBeacon.js
mauriciocastillosilva-okta: True. It's just that the function is basically doing a one-line call, so I didn't want to add an extra function for something as simple. But either works, so up to you.

alexeystadnikov-okta:
Added tests for qtip.
Tried to avoid adding css to tests, but did not find a better way to test visibility.

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -951,6 +951,63 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
           expect(test.form.securityImageTooltipText()).toEqual('This is the first time you are connecting to foo.com from this browser×');
         });
       });
+      itp('does not show anti-phishing message if security image is hidden', function () {
+        return setup({ features: { securityImage: true }})
+        .then(function (test) {
+          test.form.applyQtipCssRule();
```
mauriciocastillosilva-okta: I don't think we need this... it adds a style so we can verify tooltip is hidden. But we should not test whether the tooltip is hidden or not present (which is the same for testing purposes, it is just not there)
I think we need 2 tests, and without the css rule fix you are adding, the tests will be something like:

1. hide beacon and set wrong username -> waitForBeaconChange -> verify tooltip is not there
2. wrong username -> waitForBeaconChange -> check tooltip is there -> hide beacon and resize -> waitForBeaconChange -> check tooltip is not there


And test for whether the tooltip exists, and if it does, then whether is visible or not.

alexeystadnikov-okta: Done

PR #214, opened 2017-04-19
:bug: Add password history block to the reset password description

alexeystadnikov-okta:
Resolves OKTA-122940

screen shot 2017-04-19 at 6 20 15 pm

@mauriciocastillosilva-okta 
@anishashetty-okta 
Please, review.


test/unit/spec/PasswordReset_spec.js
```diff
@@ -54,6 +55,17 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordResetForm, Beacon, Expect,
       delete settings.policyComplexity;
     }
 
+    if (settings && settings.policyAge) {
```
mauriciocastillosilva-okta: this won't work if you have a test that asks for specific complexity and also policyAge
```
itp('has a valid subtitle if excludeUsername and password age "history" defined', function () {
    return setup({policyComplexity: 'excludeUsername', policyAge: 'history'}).then(function (test) {
```
which is the test I think would be missing.

src/util/FactorUtil.js
```diff
@@ -171,35 +171,49 @@ define(['okta'], function (Okta) {
...
+            requirement;
+        });
 
-    if (requirements.length) {
-      requirements = _.reduce(requirements, function (result, requirement) {
-        return result ?
-          (requirement ? (result + Okta.loc('password.complexity.list.element', 'login', [requirement])) : result) :
-          requirement;
-      });
+        // "password.complexity.history": "Your password cannot be any of your last {0} passwords.",
```
alexeystadnikov-okta: Done

src/util/FactorUtil.js
```diff
@@ -171,35 +171,49 @@ define(['okta'], function (Okta) {
...
+        result.push(Okta.loc('password.complexity.description', 'login', [requirements]));
+      }
+    }
 
-      return Okta.loc('password.complexity.description', 'login', [requirements]);
+    if(policy.age){
```
alexeystadnikov-okta: Done

test/unit/spec/PasswordReset_spec.js
alexeystadnikov-okta: Done

src/util/FactorUtil.js
```diff
@@ -171,35 +171,48 @@ define(['okta'], function (Okta) {
...
+        requirements = _.reduce(requirements, function (result, requirement) {
+          return result ? (result + Okta.loc('password.complexity.list.element', 'login', [requirement])) : requirement;
+        });
+
+        result.push(Okta.loc('password.complexity.description', 'login', [requirements]));
+      }
+    }
 
-      return Okta.loc('password.complexity.description', 'login', [requirements]);
+    if (policy.age) {
```
mauriciocastillosilva-okta: nit: ```if (policy.age && policy.age.historyCount) {```

test/unit/helpers/xhr/PASSWORD_RESET_withComplexity.js
```diff
@@ -7,6 +7,10 @@ define({
     "status": "PASSWORD_RESET",
     "_embedded": {
       "policy": {
+        "age": {
+          "minAgeMinutes": -1,
```
mauriciocastillosilva-okta: do we need ```minAgeMinutes``` here? Doesn't seem to be used...

test/unit/helpers/xhr/PASSWORD_RESET_withComplexity.js
```diff
@@ -7,6 +7,10 @@ define({
...
+          "historyCount": 0
```
mauriciocastillosilva-okta: should we set this to ```7``` here and that way we don't have to add that value in the test itself? Also that will help in case anyone else uses this xhr in some other test

src/util/FactorUtil.js
alexeystadnikov-okta: Done

test/unit/helpers/xhr/PASSWORD_RESET_withComplexity.js
alexeystadnikov-okta: Removed

test/unit/helpers/xhr/PASSWORD_RESET_withComplexity.js
alexeystadnikov-okta: If I will keep here non-zero values I will have to reset them all for testing only one complexity rule.
Do you think it is a better way?

test/unit/spec/PasswordReset_spec.js
```diff
@@ -40,18 +40,23 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordResetForm, Beacon, Expect,
...
       }
-      else {
+      else if (settings.policyComplexity) {
+        var key = settings.policyComplexity;
         responsePolicy.complexity[key] = policyComplexityDefaults[key];
       }
       delete settings.policyComplexity;
+
+      if (settings.policyAge === 'history') {
+        responsePolicy.age.historyCount = 7;
```
mauriciocastillosilva-okta: nit: move the ```7``` to a variable named ```historyCountDefault```, just so we have the default values all in the same place

test/unit/spec/PasswordReset_spec.js
```diff
@@ -148,11 +153,24 @@ function (Q, _, $, OktaAuth, LoginUtil, Util, PasswordResetForm, Beacon, Expect,
...
+      function () {
+        return setup({policyComplexity: 'excludeUsername', policyAge: 'history'}).then(function (test) {
+          expect(test.form.subtitleText()).toEqual('Your password must have no parts of your username.' +
+            ' Your password cannot be any of your last 7 passwords.');
+        });
+      }
+    );
     itp('has a valid subtitle if password complexity is defined with all options', function () {
-      return setup({policyComplexity: 'all'}).then(function (test) {
+      return setup({policyComplexity: 'all', policyAge: 'history'}).then(function (test) {
```
mauriciocastillosilva-okta: can you please add one with just ```policyComplexity: 'all'```, just so we keep testing that scenario too

test/unit/helpers/xhr/PASSWORD_RESET_withComplexity.js
mauriciocastillosilva-okta: I see. then what you have is good 👍

test/unit/spec/PasswordReset_spec.js
alexeystadnikov-okta: Done

test/unit/spec/PasswordReset_spec.js
alexeystadnikov-okta: Done

PR #216, opened 2017-04-25
Use Resend in Voice Call MFA

alexeystadnikov-okta:
screen shot 2017-04-25 at 1 31 11 pm
Resolves OKTA-112736

@mauriciocastillosilva-okta 
@sergiymelnyk-okta 

Please, review.


src/models/Factor.js
```diff
@@ -158,7 +158,7 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
           // However, we currently have a problem with SMS
```
mauriciocastillosilva-okta: can you update the comment two lines above to say ```Only push, sms and call have resend links.```

@gauthamgowrishankar-okta is SMS still broken for resend?

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1437,6 +1437,29 @@ function (Okta,
             expect(test.form.errorBox().length).toBe(0);
           });
         });
+        itp('call resend if call second time', function () {
```
mauriciocastillosilva-okta: nit: ``` posts to resend link if call button is clicked for the second time```

mauriciocastillosilva-okta:
Can you update the comment on ```Factor.js``` please? Then I'll merge it

alexeystadnikov-okta:
What comment?
If you are about a comment about SMS inside `canUseResend`, I will update it with the fix for SMS

mauriciocastillosilva-okta:
Comment in Factor.js line 157. It says ```// Only push and sms have resend links.```, but that is no longer true since now Call also supports resend

alexeystadnikov-okta:
Comment changed

src/models/Factor.js
```diff
@@ -207,7 +207,7 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
         }
 
         // MFA_CHALLENGE
-        else if (this.get('canUseResend') && transaction.resend) {
+        else if (this.get('canUseResend') && !this.get('answer') && transaction.resend) {
```
mauriciocastillosilva-okta: do we need this? It sounds like we can do resend only if it is push or call, and if the answer field is empty. But it should not matter if the answer textbox is empty or not right?

alexeystadnikov-okta: If I use `resend` for validating the answer, backend responses with error:
`{"errorCode":"E0000047","errorSummary":"API call exceeded rate limit due to too many requests.", ...}`
This is why I use resend only to repeat the call or push and not for sending the answer.

mauriciocastillosilva-okta: but you shouldn't use resend to validate the answer. We should just be doing another post to the resend link.
Check the Form see if there is something that is wrong there.

alexeystadnikov-okta: @mauriciocastillosilva-okta 
Could you explain what do you mean?
The main logic is in the model, so I don't understand what to change in the Form.
Moving `resend` to the specific function is not a good idea since PushForm also uses it by calling `save`, so this way there will be a duplicated logic.

mauriciocastillosilva-okta: What I'm trying to figure out is why you need to check if ```answer``` is null. At the end, if the user types something in the answer textbox or not, it has nothing to do with the user clicking the resend button. So I'm confused on why you need to add ```!this.get('answer') ``` to the conditions.

alexeystadnikov-okta: On click `Call` it makes `verify` request.
After we get the code, we fill the `answer` textbox and click `verify`.
Without checking the `answer`, it posts the code with `resend` request, that has error response.
So, to make a `Call` we always have empty `answer` and make the first request with `verify`, others with `resend`.
If we have an `answer`, we always do `verify` request.

PR #217, opened 2017-04-26
Mcs okta 123104 session step up response

mauriciocastillosilva-okta:
When the type is SESSION_STEP_UP, the response object pass to the success callback will now contain data about the target url as well as a function to redirect to such url

Extra (2nd commit): changed 'authentication' to 'authenticate', since that changed in the API for the next link in UNAUTHENTICATED state


src/util/RouterUtil.js
```diff
@@ -112,7 +112,7 @@ function (Okta, Util, OAuth2Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
         return;
       }
 
-      router.settings.callGlobalSuccess(Enums.SUCCESS, {
+      var successData = {
         user: res._embedded.user,
```
manueltanzi-okta: Don't know enough about this, just wondering, should we have a type in successData also for this case to be more consistent?

test/unit/helpers/xhr/SUCCESS_session_step_up.js
```diff
@@ -0,0 +1,48 @@
...
+        "href": "http://foo.okta.com/login/step-up/redirect?stateToken=aStateToken",
+        "hints": {
+          "allow": [
+            "GET"
+          ]
+        }
+      }
+    }
+  }
+});
```
manueltanzi-okta: nit: end line

manueltanzi-okta:
lgtm 🚢

README.md
```diff
@@ -206,6 +206,13 @@ signIn.renderEl(
       console.log(res.user);
       res.session.setCookieAndRedirect('https://acme.com/app');
 
+      // If the widget is not configured for OIDC and the authentication type is SESSION_STEP_UP,
```
rchild-okta: I think it'd be nicer to call out the logic here in the code, rather than through the comment. I.e. explicitly have the:

```javascript
if (res.authTypeOrWhatever === 'SESSION_STEP_UP') {
  res.target.redirect(); 
}
else if (res.thingy === 'SESSION_SSO') {
  // do our other thing
}
```

Also, two other points:
- Were we going to use "target" or something more descriptive like "stepUp"?
- After reading through the docs, I think it'd be nicer to call it something like "finish", "complete", or "return" instead of "redirect".

src/util/Enums.js
```diff
@@ -40,6 +40,12 @@ define({
   // Operations
   PRIMARY_AUTH: 'PRIMARY_AUTH',
   FORGOT_PASSWORD: 'FORGOT_PASSWORD',
-  UNLOCK_ACCOUNT: 'UNLOCK_ACCOUNT'
+  UNLOCK_ACCOUNT: 'UNLOCK_ACCOUNT',
+
+  // Auth Types
+  SESSION_STEP_UP: 'SESSION_STEP_UP',
```
rchild-okta: Can you check with @srinivasanagandla-okta whether we know that the normal mode is going to return an auth type of SESSION_SSO? If so, it'd be nice to include that now in this PR - if it's not currently being returned by the API, we could add it ourselves until it's there.

src/util/RouterUtil.js
```diff
@@ -124,7 +124,24 @@ function (Okta, Util, OAuth2Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
             }));
           }
         }
-      });
+      };
+
+      if (res.type === Enums.SESSION_STEP_UP && res._embedded.target &&
+          res._embedded.target.type === Enums.TARGET_APP) {
```
rchild-okta: Do we actually care about checking the target type? Seems like the proposal should work for any target (i.e. you'd only use the target type to configure what to show in the UI).

rchild-okta: Also, I think in general this code will be nicer if we're not defining the successType and then redefining it. I.e. something like this:

```javascript
const isStepUp = booleanCondition && stuff_below;

var successData = { stuff that is common };

if (isStepUp) {
  successData.stepUp = {
    url: targetUrl,
    redirect: function () {}
  }
}
else {
  successData.session = { foo }
}
```

README.md
mauriciocastillosilva-okta: I'll change the comment to include the ```if``` blocks. Also will change 'redirect' to 'finish'. 

I'm already using 'target', so will keep it unless you prefer 'stepUp'. I don't have a strong opinion for either, I chose 'target' just because the response type is already 'STEP_UP'

src/util/Enums.js
mauriciocastillosilva-okta: API will not return SESSION_SSO just yet, but it is in the roadmap. So for now I will just add it to our response object.

src/util/RouterUtil.js
mauriciocastillosilva-okta: Added

PR #219, opened 2017-05-10
:bug: Use resend link for SMS MFA

alexeystadnikov-okta:
Resolves OKTA-124311

@mauriciocastillosilva-okta 
@manueltanzi-okta 

Please, review.


alexeystadnikov-okta:
@mauriciocastillosilva-okta 

I had to add a check for an empty answer (like in https://github.com/okta/okta-signin-widget/pull/216) to have other tests passed.

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1034,6 +1034,34 @@ function (Okta,
...
+            test.setNextResponse(resChallengeSms);
+            test.form.smsSendCode().click();
+            return tick(test);
+          })
+          .then(function (test) {
+            // One more tick to give it time for become enable
+            return tick(test);
+          })
+          .then(function (test) {
+            test.setNextResponse(resChallengeSms);
```
mauriciocastillosilva-okta: nit: you could reset the calls to $.ajax, like other tets ```$.ajax.calls.reset();```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1034,6 +1034,34 @@ function (Okta,
...
+
+          return setupSMS().then(function (test) {
+            $.ajax.calls.reset();
+            test.setNextResponse(resChallengeSms);
+            test.form.smsSendCode().click();
+            return tick(test);
+          })
+          .then(function (test) {
+            // One more tick to give it time for become enable
+            return tick(test);
```
mauriciocastillosilva-okta: why do we need another tick? We should not need that, or maybe we should wait for something else to be done (instead of a tick, we could try adding a spy and wait for it, or something similar (we have changed some tests already to avoid tick)

mauriciocastillosilva-okta:
we need to rebase on master since this has a conflicting change

test/unit/spec/MfaVerify_spec.js
alexeystadnikov-okta: Done

test/unit/spec/MfaVerify_spec.js
alexeystadnikov-okta: Done

alexeystadnikov-okta:
Branch rebased on last master.
Tests updated

src/views/mfa-verify/CallAndSMSForm.js
```diff
@@ -57,17 +57,17 @@ define(['okta', 'vendor/lib/q', 'views/shared/TextBox'], function (Okta, Q, Text
         className: 'button ' + buttonClassName,
         title: formSubmit,
         click: function () {
-          self.clearErrors();
+          form.clearErrors();
+          this.disable();
+          this.render();
```
mauriciocastillosilva-okta: why did we change this? I think this is causing tests to fail

src/views/mfa-verify/CallAndSMSForm.js
```diff
@@ -33,7 +33,7 @@ define(['okta', 'vendor/lib/q', 'views/shared/TextBox'], function (Okta, Q, Text
     },
 
     initialize: function () {
-      var self = this;
+      var form = this;
```
mauriciocastillosilva-okta: any reason we renamed it? It doesn't affect anything, but just curious... we could leave it as it was

alexeystadnikov-okta: Usage of `self` is far from the definition.
It makes easier to understand what is `self`, is it a form, model or button.

```
self.clearErrors();
this.disable();
this.render();
// To send an OTP to the device, make the same request but use
// an empty passCode
this.model.set('answer', '');
this.model.save()
.then(_.bind(function () {
this.options.title = formSubmitted;
// render and focus on the passcode input field.
self.getInputs().first().render().focus();
return Q.delay(API_RATE_LIMIT);
```
Is it easy to understand what is `self` here?

src/views/mfa-verify/CallAndSMSForm.js
alexeystadnikov-okta: Answered above

src/views/mfa-verify/CallAndSMSForm.js
```diff
@@ -57,19 +57,19 @@ define(['okta', 'vendor/lib/q', 'views/shared/TextBox'], function (Okta, Q, Text
...
+          form.clearErrors();
+          this.disable();
+          this.options.title = formSubmitted;
+          this.render();
           // To send an OTP to the device, make the same request but use
           // an empty passCode
           this.model.set('answer', '');
           this.model.save()
-          .then(_.bind(function () {
-            this.options.title = formSubmitted;
```
mauriciocastillosilva-okta: I think changing the title to whatever formSubmitted is (call or sms) should still be done after save. Before save we haven't really submitted anything. So if there is an error the button will read "Sent" even though nothing was actually sent.
```clearErrors``` makes sense to be moved before saving the model.

File: src/views/mfa-verify/CallAndSMSForm.js (hunk 4 above)
Comment (mauriciocastillosilva-okta) [inline src/views/mfa-verify/CallAndSMSForm.js]: I see. Sounds good

File: src/views/mfa-verify/CallAndSMSForm.js (hunk 5 above)
Comment (alexeystadnikov-okta) [inline src/views/mfa-verify/CallAndSMSForm.js]: In case of error, we also run this function and change the title.

However, if a connection is bad, the button will not be disabled instantly and user is able to click it multiple times.

screen shot 2017-05-18 at 5 26 29 pm

**Comment by alexeystadnikov-okta** [comment]
@mauriciocastillosilva-okta 
Do you have any comments?

File: test/unit/spec/MfaVerify_spec.js [hunk 6]
```diff
@@ -1034,6 +1037,37 @@ function (Okta,
...
+            test.form.smsSendCode().click();
+            return Expect.waitForCss('.sms-request-button.disabled', test);
+          })
+          .then(function (test) {
+            return Expect.waitForCss('.sms-request-button:not(.disabled)', test);
+          })
+          .then(function (test) {
+            $.ajax.calls.reset();
+            test.setNextResponse(resChallengeSms);
+            test.form.smsSendCode().click();
```
mauriciocastillosilva-okta: before click, we should test that the button has the right label
```
expect(test.form.smsSendCode().trimmedText()).toEqual('Re-send code');
```

alexeystadnikov-okta: Done

PR #220, opened 2017-05-11
Add id attribute to username, password and signin button

manueltanzi-okta:
🌱 Add id attribute to username, password and signin button

Add id attribute to username, password and signin button and bump courage version to 2.6.0

Resolves: [OKTA-125825](https://oktainc.atlassian.net/browse/OKTA-125825)

@hor-kanchan-okta @alexeystadnikov-okta @mauriciocastillosilva-okta

### DIFF
--- a/src/views/primary-auth/PrimaryAuthForm.js
+++ b/src/views/primary-auth/PrimaryAuthForm.js
@@ -22,6 +22,7 @@ define([
     className: 'primary-auth-form',
     noCancelButton: true,
     save: _.partial(Okta.loc, 'primaryauth.submit', 'login'),
+    saveId: 'signin-button',
     layout: 'o-form-theme',
 
     // If socialAuth is configured, the title moves from the form to
@@ -73,6 +74,7 @@ define([
         placeholder: Okta.loc('primaryauth.username.placeholder', 'login'),
         name: 'username',
         input: TextBox,
+        inputId: 'user-signin',
         type: 'text',
         params: {
           innerTooltip: {
@@ -87,6 +89,7 @@ define([
         placeholder: Okta.loc('primaryauth.password.placeholder', 'login'),
         name: 'password',
         input: TextBox,
+        inputId: 'pass-signin',
         type: 'password',
         params: {
           innerTooltip: {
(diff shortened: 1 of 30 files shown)


manueltanzi-okta:
@mauriciocastillosilva-okta bump review thank you :)

src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -87,6 +89,7 @@ define([
         placeholder: Okta.loc('primaryauth.password.placeholder', 'login'),
         name: 'password',
         input: TextBox,
+        inputId: 'pass-signin',
```
mauriciocastillosilva-okta: nit: why not ```password-signin```? I think it will be a more descriptive id for this input

hor-kanchan-okta: We try to keep the id same as the old signin page.

rchild-okta:
Why do we need to add these id's? Is the ops monitoring not able to use other types of selectors?

If we have to add ids, they should be more specific - remember that any id we use could collide with any name a developer would choose for their own app. Something like "okta-signin-widget-user", etc (might be enough to just prefix with okta).

src/views/primary-auth/PrimaryAuthForm.js
rchild-okta: If we need consistency, we should change the fields in the old loginform to whatever we end up with here.

hor-kanchan-okta:
From what @garethpaterson-okta told me, the Ops monitoring only use id. It is possible to use other type but it will be inconsistent with the rest of the monitoring script.

For the id, i was trying to keep the same id. So the monitoring script will work for both new and old signin flow. But @rchild-okta has a good point that the id may be collided.
So we have to update the id. Add a prefix.
If necessary, we can update the old signin page ids.
@garethpaterson-okta , will different work for you?

hor-kanchan-okta:
@garethpaterson-okta is ok with different ids. 
@manueltanzi-okta can you change the id to okta-signin-username, okta-signin-password, okta-signin-submit
@rchild-okta I suggest adding the id which will make Ops life easier.

PR #222, opened 2017-05-13
:bug: handle unlock recoveryType before anything else

mauriciocastillosilva-okta:
When status is SUCCESS and we are in recoveryType unlock, we should redirect to the success unlocked view before handling any other SUCCESS possible outcome.

Before we were trying to handle an oauth token exchange for a sessionToken, but during unlock we don't get a sessionToken.

Resolves: OKTA-119625

### DIFF
--- a/src/util/RouterUtil.js
+++ b/src/util/RouterUtil.js
@@ -100,15 +100,15 @@ function (Okta, Util, OAuth2Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
 
     switch (res.status) {
     case 'SUCCESS':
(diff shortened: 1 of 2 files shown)


src/util/RouterUtil.js
```diff
@@ -100,18 +100,18 @@ function (Okta, Util, OAuth2Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
 
     switch (res.status) {
     case 'SUCCESS':
```
edjohns-okta: Can we add a comment indicating that the recovery type check should be the first one made?  That might help us avoid another issue like this.

PR #224, opened 2017-05-15
:bug: Firefox on iOS make Google Auth text readable

eugenebakaiev-okta:
Firefox on iOS make Google Auth text readable

Resolves: OKTA-124746

Reviewers:

@samerfanek-okta

### DIFF
--- a/assets/sass/modules/_enroll.scss
+++ b/assets/sass/modules/_enroll.scss
@@ -403,6 +403,7 @@
 [name="sharedSecret"] {
   text-align: center;
   cursor: text;
+  opacity: 1;
 }
 
 .enroll-activation-link-sent {
--- a/assets/sass/okta-theme.scss
+++ b/assets/sass/okta-theme.scss
@@ -628,6 +628,7 @@
       &[disabled] {
       /* -- Input Fields: disabled input field -- */
         color: $medium-text-color;
+        -webkit-text-fill-color: $medium-text-color;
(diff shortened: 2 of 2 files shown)


eugenebakaiev-okta:
Before: 
img_0006 jpg

After:
img_0005 jpg

assets/sass/modules/_enroll.scss
```diff
@@ -403,6 +403,7 @@
 [name="sharedSecret"] {
   text-align: center;
   cursor: text;
+  opacity: 1;
```
hor-kanchan-okta: `1` is default value. Is the opacity overridden? If so, where?

eugenebakaiev-okta: This input is being located inside the span with class .o-form-disabled, which has a property opacity: 0.4. For some reason Firefox applied this opacity once again to this input which is direct sibling, so we have input with opacity 0.4 too. As a result input was too light. Now it looks fine.

hor-kanchan-okta:
@eugenebakaiev-okta can you update your commit message following 
https://github.com/okta/okta-signin-widget/blob/master/CONTRIBUTING.md

rchild-okta:
FYI @hor-kanchan-okta when merging, you also have an option to update the commit message. So in this case when someone forgets, I think it's good to call it out, but you can also just update it yourself and merge anyway if everything looks okay.

PR #227, opened 2017-05-17
Update generator of password complexity description to use ‘-1’ for false values

alexeystadnikov-okta:
Resolves OKTA-126664

@mauriciocastillosilva-okta 
@richardheng-okta 

Please, review.

### DIFF
--- a/src/util/FactorUtil.js
+++ b/src/util/FactorUtil.js
@@ -186,7 +186,7 @@ define(['okta'], function (Okta) {
 
       var policyComplexity = policy.complexity;
       var requirements = _.map(policyComplexity, function (complexityValue, complexityType) {
-        if (!complexityValue) {
+        if (complexityValue <= 0) { // to skip 0 and -1
           return;
         }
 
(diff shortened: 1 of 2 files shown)


test/unit/helpers/xhr/PASSWORD_RESET_withComplexity.js
```diff
@@ -8,14 +8,14 @@ define({
     "_embedded": {
       "policy": {
         "age": {
-          "historyCount": 0
+          "historyCount": -1
```
mauriciocastillosilva-okta: if I remember correctly this values don't need to change, but instead we need to add a test in ```PasswerReset_spec``` for -1 (that spec overrides the values from this xhr mock)

src/util/FactorUtil.js
```diff
@@ -186,7 +186,7 @@ define(['okta'], function (Okta) {
 
       var policyComplexity = policy.complexity;
       var requirements = _.map(policyComplexity, function (complexityValue, complexityType) {
-        if (!complexityValue) {
+        if (complexityValue <= 0) { // to skip 0 and -1
```
mauriciocastillosilva-okta: In theory, the api won't return complexity as -1. However, this is a good check just in case.

test/unit/helpers/xhr/PASSWORD_RESET_withComplexity.js
mauriciocastillosilva-okta: Talked offline. The tests are actually using these values here, my bad.

PR #229, opened 2017-05-24
:bug: fix e2e tests

mauriciocastillosilva-okta:
Facebook now remembers the username in the login page, so we need to clear it before typing any other username

Resolves: OKTA-127526


mauriciocastillosilva-okta:
We did not. My guess is that we already, somehow, clicked 'Ok' on that page, and that is why facebook remembers the username of the last login for those accounts.

PR #230, opened 2017-05-25
:bug: fix customized labels in some locales

eugenebakaiev-okta:
use lower case language codes for extending i18n bundles with overriden fields
Resolves: OKTA-126768


eugenebakaiev-okta:
@richardheng-okta, please take a look

src/util/Bundles.js
```diff
@@ -198,15 +198,16 @@ define([
...
         this.login = _.extend({}, login, bundles.login);
         this.country = _.extend({}, country, bundles.country);
-        if (parsedOverrides[language]) {
-          _.extend(this.login, parsedOverrides[language]['login']);
-          _.extend(this.country, parsedOverrides[language]['country']);
+        if (parsedOverrides[lowerCaseLanguage]) {
+          _.extend(this.login, parsedOverrides[lowerCaseLanguage]['login']);
+          _.extend(this.country, parsedOverrides[lowerCaseLanguage]['country']);
         }
         this.currentLanguage = language;
```
richardheng-okta: Not sure how this is used, but does this need to be assigned to lowerCaseLanguage?

eugenebakaiev-okta: Actually this is a point of this fix. parsedOverrides is an object which contains overriden i10n fields (comes from org customization). Keys in this object are locale name e.g. 'en', 'fr', 'nl-nl'. The problem is this keys are always in lowercase and the `language` comes to this method in different cases (e.g. for Netherlands it's 'nl-NL'). Those language code is being used wide through the app and comes from okta-core. So I decided to convert it to lower case only in this method to prevent more global changes which could cause regressions.

>

PR #231, opened 2017-05-26
:seedling: Add registration form

hor-kanchan-okta:
Add a registration form base on a static schema.
Update Courage to 2.11.0
Update i18n to 27.26.0

Resolve: OKTA-127382


hor-kanchan-okta:
screen shot 2017-05-26 at 4 16 10 am

When missing password.
screen shot 2017-05-26 at 4 16 49 am

After submit
screen shot 2017-05-26 at 4 17 04 am

npm-shrinkwrap.json
```diff
@@ -2,11 +2,36 @@
   "name": "@okta/okta-signin-widget",
   "version": "1.11.0",
   "dependencies": {
+    "@okta/ci-pkginfo": {
+      "version": "1.2.0"
+    },
+    "@okta/ci-update-package": {
```
hor-kanchan-okta: This should not be here.

hor-kanchan-okta: The [wiki](https://oktawiki.atlassian.net/wiki/display/eng/Sign-in+Widget+Development) said I should remove ci-
- ci-pkginfo
- ci-update-package
- install-with-shrinkwrap
- fs-events

src/views/shared/FooterRegistration.js
```diff
@@ -28,10 +28,13 @@ define([
     },
 
     handleClickEvent: function () {
-      var clickHandler = this.settings.get('registration').click;
-      if (Okta._.isFunction(clickHandler)) {
+      var clickHandler = this.settings.get('registration.click');
```
nikhilvenkatraman-okta: Why are we removing the check for function ?

src/models/Settings.js
```diff
@@ -135,8 +135,9 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
       'customButtons': ['array', false, []],
 
       //Registration
-      'registration': 'object',
-      'registration.click': 'function'
+      'registration.enable': ['boolean', false, false],
```
nikhilvenkatraman-okta: why do we need this enabled flag ? And what is the benefit of adding it here instead of adding it to  features object ?

package.json
```diff
@@ -95,8 +95,8 @@
   "optionalDependencies": {
     "@okta/ci-pkginfo": "1.2.0",
     "@okta/ci-update-package": "1.4.0",
-    "@okta/courage": "2.6.0",
-    "@okta/i18n": "27.13.0",
+    "@okta/courage": "2.11.0",
+    "@okta/i18n": "27.26.0",
```
nikhilvenkatraman-okta: Is there a i18n PR for this ?

src/RegistrationController.js
```diff
@@ -0,0 +1,123 @@
...
+              'description': 'Password'
+            }
+          }
+        }
+      }, {parse:true});
+
+      var properties = schema.properties;
+      var modelProperties = properties.createModelProperties();
+      _.map(modelProperties, function(modelProperty) {
+        modelProperty.required = true;
```
nikhilvenkatraman-okta: can this be defined in the schema itself ?

src/RegistrationController.js
```diff
@@ -0,0 +1,123 @@
...
+      });
+
+      var Model = Okta.Model.extend({
+        settings: this.settings,
+        appState: this.options.appState,
+        props: modelProperties,
+        save: function () {
+          var formSubmitCallback = this.settings.get('registration.submit');
+          if (formSubmitCallback) {
```
nikhilvenkatraman-okta: should we check for formSubmitCallback to be function? We could use that check to throw a message back saying formSubmitCallback is not a function

src/RegistrationEmailSentController.js
```diff
@@ -0,0 +1,49 @@
...
+function (Okta, Enums, FormController, FormType) {
+
+  var _ = Okta._;
+
+  return FormController.extend({
+    className: 'registration-email-sent',
+    Model: function () {
+    },
+
+    Form: {
```
nikhilvenkatraman-okta: :nit Do we have to use a form here since all it has is just message that email is sent. Feels semantically incorrect since we don't have any form inputs.

src/RegistrationController.js
```diff
@@ -0,0 +1,123 @@
...
+      var Schema = BaseSchema.Model.extend({
+        url: 'api/v1/schema',
+        expand: ['schema']
+      });
+      var schema = new Schema({
+        schema: {
+          'properties':{
+            'firstname': {
+              'type': 'string',
+              'description': 'First name',
```
nikhilvenkatraman-okta: When these fields get rendered do they set an aria-label attribute for the inputs ?

README.md
```diff
@@ -824,13 +824,18 @@ You can add a registration button to the primary auth page by setting the follow
 ```javascript
 // An example that adds a registration button underneath the login form on the primary auth page
 registration: {
-  click: function(){
+  enable: true,
+  click: function() {
     window.location.href = 'https://acme.com/sign-up';
+  },
+  submit: function(registrationData) {
```
nikhilvenkatraman-okta: what does  the registrationData look like ?

hor-kanchan-okta: {firstname: "Hor-kan", lastname: "Chan", email: "hor-kan.chan@okta.com", password: "xyz"}

package.json
hor-kanchan-okta: Yes. https://github.com/okta/npm/pull/670

src/RegistrationController.js
hor-kanchan-okta: Unfortunately no. Will file a jira against Courage.

src/RegistrationController.js
hor-kanchan-okta: Will be. When the schema is more defined.

src/RegistrationController.js
hor-kanchan-okta: We clearly state that 'registration.submit' is a function in the model and in the doc. As the function is defined in the option, it can't be changed in the runtime. If we should check it, it should be in the model (Settings.js)

src/RegistrationEmailSentController.js
hor-kanchan-okta: Will change it.

src/models/Settings.js
hor-kanchan-okta: That was a suggestion from Reman. If the developer doesn't override anything, it will be the default behavior.
So instead to turn on a feature by empty json. It will be more explicitly to say `enable: true`

src/views/shared/FooterRegistration.js
hor-kanchan-okta: The model should check for the type. And we don't do check anywhere else.

src/RegistrationEmailSentController.js
hor-kanchan-okta: May do it later. There is a css style only for the form. I need to figure out what they are first.

README.md
```diff
@@ -824,13 +824,18 @@ You can add a registration button to the primary auth page by setting the follow
...
+    // submit the registration data
   }
 }
 ```
+- **registration.enable** - Display the registration section in the primary auth page. Defaults to `false`.
 - **registration.click** - Function that is called when the registration button is clicked
-
+- **registration.submit** - Function that is called when the registration form is submitted.
```
rchild-okta: I think we should consider doing this the way that we talked about a couple days ago - i.e. if we're changing the signature, it's better to do it once (because it will be a breaking change, or at least unclear, in the very next update).

Specifically, there should be two hooks here:

1. updateSchema (or something like that) - step that happens after the schema request is made and they get a chance to change it. In our case, since no endpoint exists yet, we should just pass in our default schema that you defined elsewhere into this function. Goal here is that when we move to the endpoint, their code doesn't need to change, and our code just needs to swap out our hardcoded schema with the result of the request. **Note** this should support async - they should have the option to do async logic and then continue on.

2. submit - hmmm, so feels like there should be a couple behaviors here. Maybe we need another method. What we want is this:

- They can override completely, and just get results of form
- They can return schema + values, which is then passed to our registration endpoint. Should support async as well.

Looking at it now (and I see where you're going with this) - ideally we can do just the submit part and not have it conflict with later changes (i.e. maybe we don't need to do the updateSchema hook yet). If so, I'm okay with that - but what is the plan for how submit behaves (specifically when we're going to pass it through to the RAAS endpoint)?

src/RegistrationController.js
```diff
@@ -0,0 +1,123 @@
...
+              'description': 'Last name',
+              'maxLength': 255
+            },
+            'email': {
+              'type': 'email',
+              'description': 'Email'
+            },
+            'password': {
+              'type': 'password',
+              'description': 'Password'
```
gowthamidommety-okta: Should we localize these descriptions?

hor-kanchan-okta: They should be coming from the backend eventually.
This is the first item in 1B.
https://oktawiki.atlassian.net/wiki/display/PM/Registration+as+a+Service+Dev+Estimates

README.md
```diff
@@ -824,13 +824,18 @@ You can add a registration button to the primary auth page by setting the follow
...
+  enable: true,
+  click: function() {
     window.location.href = 'https://acme.com/sign-up';
+  },
+  submit: function(registrationData) {
+    // submit the registration data
   }
 }
 ```
+- **registration.enable** - Display the registration section in the primary auth page. Defaults to `false`.
```
hor-kanchan-okta: Will move this to `features.registration` as @nikhilvenkatraman-okta points out.

src/RegistrationController.js
```diff
@@ -0,0 +1,123 @@
...
+
+  var Footer = Okta.View.extend({
+    template: '\
+      <a href="#" class="link help" data-se="back-link">\
+      {{i18n code="goback" bundle="login"}}\
+    </a>\
+    ',
+    className: 'auth-footer',
+    events: {
+      'click .help' : function (e) {
```
mauriciocastillosilva-okta: nit: can we add an extra class to that link so this will read something like 
```
'click .back-link' : function (e) {
```
It is just a little confusing to read that we are clicking help to go back.

PR #232, opened 2017-06-01
:fix: Allow Re-push for Okta Verify

alexeystadnikov-okta:
Resolves OKTA-127113

- Remove Timeout error message
- Change button text to "Re-push" after it was enabled after timeout

@mauriciocastillosilva-okta 
Please, review.


mauriciocastillosilva-okta:
We decided to take a different approach, which is to show a message prompting the user to open Okta Verify App.

PR #234, opened 2017-06-04
:seedling: Allows re-push for Okta Verify

gowthamidommety-okta:
Okta verify re-enables submit button after 30 seconds

Resolves: OKTA-127113


src/views/mfa-verify/PushForm.js
```diff
@@ -92,6 +94,10 @@ define(['okta', 'util/CookieUtil'], function (Okta, CookieUtil) {
       if (this.model.isValid()) {
         this.listenToOnce(this.model, 'error', this.setSubmitState, true);
         this.trigger('save', this.model);
+        return Q.delay(API_RATE_LIMIT)
```
magizh-okta: why a return statement ?

src/views/mfa-verify/PushForm.js
```diff
@@ -92,6 +94,10 @@ define(['okta', 'util/CookieUtil'], function (Okta, CookieUtil) {
       if (this.model.isValid()) {
         this.listenToOnce(this.model, 'error', this.setSubmitState, true);
```
magizh-okta: in case of error this.setSubmitState is called and you call it again after 30s as well. SetSubmitState would be called twice but we really need it just once.

src/util/Util.js
```diff
@@ -84,6 +84,12 @@ define(['okta'], function (Okta) {
       .uniq()
       .value();
   };
+  
+  //helper to call setTimeout
+  Util.callAfterTimeout = function (callback, time) {
+    var timeout = setTimeout(callback, time);
```
mauriciocastillosilva-okta: nit: ````return setTimeout(callback, time);````

src/views/mfa-verify/PushForm.js
```diff
@@ -10,11 +10,12 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-define(['okta', 'util/CookieUtil'], function (Okta, CookieUtil) {
+define(['okta', 'util/CookieUtil', 'util/Util'], function (Okta, CookieUtil, Util) {
 
   var _ = Okta._;
   // deviceName is escaped on BaseForm (see BaseForm's template)
   var titleTpl = Okta.Handlebars.compile('{{factorName}} ({{{deviceName}}})');
+  var API_RATE_LIMIT = 30000; //milliseconds
```
mauriciocastillosilva-okta: we should rename this to something like 'TIMEOUT_TO_RETRY'  or 'RETRY_TIMEOUT'

src/views/mfa-verify/PushForm.js
```diff
@@ -58,12 +59,13 @@ define(['okta', 'util/CookieUtil'], function (Okta, CookieUtil) {
         deviceName: this.model.get('deviceName')
       });
     },
-    setSubmitState: function (ableToSubmit) {
+    setSubmitState: function (ableToSubmit, buttonValue) {
```
mauriciocastillosilva-okta: nit: change it to 'buttonLabelCode' or something similar

gowthamidommety-okta:
@mauriciocastillosilva-okta Made changes. I tested the flow by setting up email through postman. clearTimeout on line 98 is never hit, but the timeout does clear. This was my test:
1. Push, button says 'Push Sent!'
2. Switch to email
3. Switch back to Okta Verify
4. Button now says 'Push'
5. I waited for over 30 seconds, it did not change to 'Re-send Push'

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1619,18 +1619,45 @@ function (Okta,
                 });
               });
             });
-            itp('will disable form submit', function () {
+            itp('will re-enable submit when retry timeout reached', function () {
+              spyOn(LoginUtil, 'callAfterTimeout').and.callFake(function() {
```
mauriciocastillosilva-okta: I think you can use ```Util.mockSetTimeout``` instead

src/util/Util.js
```diff
@@ -84,6 +84,11 @@ define(['okta'], function (Okta) {
       .uniq()
       .value();
   };
+  
+  //helper to call setTimeout
+  Util.callAfterTimeout = function (callback, time) {
```
gowthamidommety-okta: I'm using it to help mock the setTimeout in my jasmine test. In the signin widget i see Q used more often than setTimeout, but we do have setTimeout in other places. However, we never used setTimeout with this large of a delay, which made it difficult to test.

src/views/mfa-verify/PushForm.js
```diff
@@ -58,12 +59,13 @@ define(['okta', 'util/CookieUtil'], function (Okta, CookieUtil) {
         deviceName: this.model.get('deviceName')
       });
     },
-    setSubmitState: function (ableToSubmit) {
+    setSubmitState: function (ableToSubmit, buttonLabelCode) {
       var button = this.$el.find('.button');
       this.enabled = ableToSubmit;
       if (ableToSubmit) {
+        var buttonString = buttonLabelCode ? buttonLabelCode : 'oktaverify.send';
```
gowthamidommety-okta: 👍

test/unit/spec/MfaVerify_spec.js
gowthamidommety-okta: Talked to Mau offline about this. Unable to use mockSetTimeout

gowthamidommety-okta:
@mauriciocastillosilva-okta Please review again. Fixed the resend bug and tested the app, seems to be good now.

gowthamidommety-okta:
Noticed that i removed changes suggested in a previous review. Made those changes in the most recent commit

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1904,18 +1904,45 @@ function (Okta,
...
+                }, 10);
+              });
+              return setupOktaPush().then(function (test) {
+                return setupPolling(test, resRejectedPush)
+                .then(tick)
+                .then(function (test) {
+                  expect(test.form.errorMessage()).toBe('You have chosen to reject this login.');
+                  expect(test.form.submitButton().prop('disabled')).toBe(false);
+                  expect(test.form.submitButtonText()).toBe('Send Push');
+                  return tick(test);
```
mauriciocastillosilva-okta: i don't think this 'tick' is doing anything here. Polling sets up to one final response, which is the ```resRejectedPush``` from line `934. After that another tick will not change anything.

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1904,18 +1904,45 @@ function (Okta,
...
+                test.form = test.form[0];
+                test.form.submit();
+                expect(test.form.submitButton().attr('class')).toMatch('link-button-disabled');
+                expect(test.form.submitButton().prop('disabled')).toBe(true);
+                expect(test.form.submitButtonText()).toBe('Push sent!');
+                return tick(test);
+              })
+              .then(function (test) {
+                expect(test.form.submitButton().prop('disabled')).toBe(false);
+                expect(test.form.submitButtonText()).toBe('Re-send Push');
```
mauriciocastillosilva-okta: can we also check that the class is not ```link-button-disabled```?

gowthamidommety-okta: We are checking that the button is not disabled.. Do we need to check the class as well?

test/unit/spec/MfaVerify_spec.js
gowthamidommety-okta: 👍

test/unit/spec/MfaVerify_spec.js
gowthamidommety-okta: Talked offline, will change this to test class as well

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1904,18 +1904,48 @@ function (Okta,
...
+                .then(function (test) {
+                  expect(test.form.errorMessage()).toBe('You have chosen to reject this login.');
+                  expect(test.form.submitButton().attr('class')).not.toMatch('link-button-disabled');
+                  expect(test.form.submitButton().prop('disabled')).toBe(false);
+                  expect(test.form.submitButtonText()).toBe('Send Push');
+                  return test;
                 })
-                .then(function () {
-                  expect($.ajax.calls.count()).toBe(0);
+                .then(function (test) {
```
mauriciocastillosilva-okta: What I meant is that this ```then(function (test) {``` section is probably not needed. At the end it is testing the same thing as lines 1938-1942

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1904,18 +1904,48 @@ function (Okta,
                 });
               });
             });
-            itp('will disable form submit', function () {
```
mauriciocastillosilva-okta: just realize that we probably don't need to remove this test. Button still should be disabled, at least until our timeout

src/views/mfa-verify/PushForm.js
```diff
@@ -58,12 +59,13 @@ define(['okta', 'util/CookieUtil'], function (Okta, CookieUtil) {
         deviceName: this.model.get('deviceName')
       });
     },
-    setSubmitState: function (ableToSubmit) {
+    setSubmitState: function (ableToSubmit, useRetryLabel) {
```
rchild-okta: So, in general, I'm wondering if we can simplify this by doing it at a lower level. Our API now times out at 5 minutes, but we've decided that we should do a client side timeout at 30 seconds. Why not just update the poll code in Factor.js (the place where you added the setTransaction call) to set a timeout response if we've hit 5 of the PUSH_INTERVALs?

If we do it this way, we don't actually need to touch the higher level code (unless you want to change the UI i.e. change the button text, but at that point it should behave the same whether you get the timeout from the server or our client side limit).

gowthamidommety-okta: From what I can tell, trans.poll is only called once in Factor.js. After this authjs sdk takes care of polling. So from that particular code in Factor.js I don't think we know how many PUSH_INTERVALs have passed.

Also, part of the ticket is to change the button text when the client side limit is hit, but yes, this differs from when the server side limit is hit.

mauriciocastillosilva-okta:
Closing it since we decided to show a message instead of allowing re-push

PR #235, opened 2017-06-12
:seedling: update registration schema

hor-kanchan-okta:
The registration schema now contains 'required', 'fieldOrder' and
'passwordComplexity'.

screen shot 2017-06-12 at 8 20 45 am
screen shot 2017-06-12 at 8 21 22 am


src/RegistrationController.js
```diff
@@ -52,42 +61,61 @@ function (Okta, BaseLoginController, FormType, Enums, BaseSchema, SchemaFormFact
...
-              'type': 'password',
-              'description': 'Password'
+            'referrer': {
+              'type': 'string',
+              'description': 'How did you hear about us?',
+              'maxLength': 1024
+            },
+            'password' : {
+              'type' : 'string',
+              'description' : 'Password'
```
nikhilvenkatraman-okta: We use the description for the placeholder on the input ? Should it be called placeholder?

src/RegistrationController.js
```diff
@@ -99,20 +127,38 @@ function (Okta, BaseLoginController, FormType, Enums, BaseSchema, SchemaFormFact
...
       var form = new Form(this.toJSON());
       properties.each(function(schemaProperty) {
-        form.addInput(_.extend(SchemaFormFactory.createInputOptions(schemaProperty), {
-          label: false,
-          'label-top': true,
-          placeholder: schemaProperty.get('description')
-        }));
+        var inputOptions = RegistrationFormFactory.createInputOptions(schemaProperty);
+        var name = schemaProperty.get('name');
+        if (name === 'password' || name === 'login') {
```
manueltanzi-okta: curiosity: login is the email input correct? so, why we need to check password meet complexity also when the user input in email field?

src/RegistrationController.js
```diff
@@ -52,42 +61,61 @@ function (Okta, BaseLoginController, FormType, Enums, BaseSchema, SchemaFormFact
...
       var schema = new Schema({
         schema: {
-          'properties':{
-            'firstname': {
+          'properties': {
+            'firstName': {
               'type': 'string',
-              'description': 'First name',
+              'description': 'First Name',
+              'default': 'Enter your first name',
```
manueltanzi-okta: In which case is default displayed, since we already have description as placeholder? It is used for the errors?

src/models/RegistrationSchema.js
```diff
@@ -0,0 +1,96 @@
...
+
+    constructor: function () {
+      this.properties = new RegistrationSchemaPropertyCollection();
+      Okta.BaseModel.apply(this, arguments);
+    },
+
+    parse: function (resp) {
+      var parsed = BaseSchema.Model.prototype.parse.apply(this, [resp]);
+
+      var requireFields = resp.schema.required;
```
hor-kanchan-okta: According to the latest JSON schema spec, the `required` is a separated.
https://tools.ietf.org/html/draft-wright-json-schema-validation-01#section-6.17

src/util/RegistrationFormFactory.js
```diff
@@ -0,0 +1,43 @@
...
+define([
+  'okta',
+  'shared/views/forms/helpers/SchemaFormFactory'
+], function (Okta, SchemaFormFactory) {
+
+  var _ = Okta._;
+
+  var fnCreateInputOptions = function(schemaProperty) {
+    var inputOptions = SchemaFormFactory.createInputOptions(schemaProperty);
+    if (inputOptions.type === 'select') {
```
hor-kanchan-okta: The `SchemaFormFactory.createInputOptions` always return type.
https://github.com/okta/courage/blob/master/src/views/forms/helpers/SchemaFormFactory.js#L18

src/RegistrationController.js
hor-kanchan-okta: The `default` is ignored in the okta-signin-widget. The API is shared with the other clients. The other clients can decide how to display `default`

manueltanzi-okta: @hor-kanchan-okta thanks

src/RegistrationController.js
samerfanek-okta: Why would you need the ``|| name === 'login'``?

src/util/PasswordComplexityUtil.js
```diff
@@ -0,0 +1,100 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define(['okta'], function (Okta) {
+
+  var _ = Okta._;
+
+  // return true if the complexity is a number and the value is greater than zero
+  var isComplexityEnabledPosNum = function(complexityValue) {
+    return complexityValue > 0;
```
samerfanek-okta: Do you want to also perhaps validate that ``complexityValue`` is indeed a number via something like ``!isNaN(complexityValue)``?

src/RegistrationController.js
```diff
@@ -99,20 +127,38 @@ function (Okta, BaseLoginController, FormType, Enums, BaseSchema, SchemaFormFact
...
+        if (name === 'password' || name === 'login') {
+          inputOptions.events = {
+            'input': function() {
+              checkPasswordMeetComplexities(this.model);
+            }
+          };
+        }
+        form.addInput(inputOptions);
       });
+      form.add(PasswordComplexity.View.extend({passwordComplexity: schema.passwordComplexity}));
```
mauriciocastillosilva-okta: should be 
```
form.add(PasswordComplexity.extend({passwordComplexity: schema.passwordComplexity}))
```

src/views/registration/PasswordComplexity.js
```diff
@@ -0,0 +1,49 @@
...
+    },
+    template: '<div>{{message}}</div>',
+    getTemplateData: function () {
+      return {
+        message: this.message
+      };
+    }
+  });
+
+  var PasswordComplexitySection =  Okta.View.extend({
```
mauriciocastillosilva-okta: this should just return the view ```return Okta.View.extend({```

test/unit/spec/PasswordComplexityUtil_spec.js
```diff
@@ -0,0 +1,293 @@
...
+    describe('default model properties', function () {
+      beforeEach(function () {
+        this.defaultProps = PasswordComplexityUtil.modelProps;
+      });
+
+      it('has 6 properties', function () {
+        expect(_.size(this.defaultProps)).toEqual(6);
+      });
+
+      it('minLength is 8', function () {
```
mauriciocastillosilva-okta: not only checking the size, but more checking all the default props.
Maybe something like ```minLenght is a number, not required and with default size 8```
Same comment for the other tests

test/unit/spec/PasswordComplexityUtil_spec.js
```diff
@@ -0,0 +1,293 @@
...
+        expect(_.size(this.complexities)).toEqual(6);
+      });
+
+      describe('minLength', function () {
+        beforeEach(function () {
+          this.isEnabled = this.complexities['minLength'].isEnabled;
+        });
+
+        it('1 is enabled', function () {
+          expect(this.isEnabled(1)).toBeTruthy();
```
mauriciocastillosilva-okta: nit ```toBe(true)```

src/RegistrationController.js
manueltanzi-okta: for the excludeUsername policy

test/unit/spec/PasswordComplexityUtil_spec.js
```diff
@@ -0,0 +1,293 @@
...
+        it('9 characters password does not meet the complexity', function () {
+          expect(this.doesComplexityMeet('123456789')).toBeFalsy();
+        });
+
+        it('empty password does not meet the complexity', function () {
+          expect(this.doesComplexityMeet('')).toBeFalsy();
+        });
+      });
+
+      describe('minLowerCase is 2 characters', function () {
```
mauriciocastillosilva-okta: can we have all the minLowerCase tests in the same describe block for minLowerCase?
Also, this should not be ```is 2 characters``` but more a describe block inside `minLowerCase` that reads ```complexity validation``` or so

test/unit/spec/RegistrationSchema_spec.js
```diff
@@ -0,0 +1,312 @@
...
+        it('is a string type', function () {
+          expect(this.schema.properties.first().get('type')).toEqual('string');
+        });
+
+        it('have a correct enum', function () {
+          expect(this.schema.properties.first().get('enum')).toEqual(['Apple', 'Banana', 'Cherry']);
+        });
+      });
+
+      Expect.describe('require fields', function () {
```
mauriciocastillosilva-okta: nit: ```required fields```

src/util/PasswordComplexityUtil.js
hor-kanchan-okta: The model should enforce the value is a number. And if the value is great than zero ( at least 1), the rule should be enabled.

src/util/PasswordComplexityUtil.js
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define(['okta'], function (Okta) {
+
+  var _ = Okta._;
+
+  // return true if the complexity is a number and the value is greater than zero
+  var isComplexityEnabledPosNum = function(complexityValue) {
```
hor-kanchan-okta: I think at some point, we may have a rule that is enabled when the value is string or something else. And we want to reuse the function. So I named this function specific for a positive number.

src/models/RegistrationSchema.js
```diff
@@ -0,0 +1,96 @@
...
+        parsed =  _.omit(parsed, 'passwordComplexity');
+      } else {
+        this.passwordComplexity = new PasswordComplexityModel({}, {parse:true});
+      }
+      return parsed;
+    }
+  });
+
+  return {
+    Schema: Schema
```
hor-kanchan-okta: Agree. Will update.

src/RegistrationController.js
```diff
@@ -52,42 +61,61 @@ function (Okta, BaseLoginController, FormType, Enums, BaseSchema, SchemaFormFact
   return BaseLoginController.extend({
     className: 'registration',
     initialize: function() {
-      var Schema = BaseSchema.Model.extend({
-        url: 'api/v1/schema',
-        expand: ['schema']
+      var Schema = RegistrationSchema.Schema.extend({
```
hor-kanchan-okta: Will make the change.

src/views/registration/PasswordComplexity.js
hor-kanchan-okta: Will update.

src/util/PasswordComplexityUtil.js
```diff
@@ -0,0 +1,100 @@
...
+      doesComplexityMeet: function(complexityValue, password, model) {
+        if (_.isEmpty(password)) {
+          return false;
+        }
+        var username = model.get('login');
+        if (_.isEmpty(username)) {
+          return false;
+        }
+
+        return password.indexOf(username) === -1;
```
hor-kanchan-okta: Good point. It doesn't match the backend. Will improve it in OKTA-130242

test/unit/spec/PasswordComplexityUtil_spec.js
hor-kanchan-okta: Will update

hor-kanchan-okta:
@mauriciocastillosilva-okta The schema should be from the backend. But the backend API is not ready yet. The schema is the latest one from AdamL

src/models/RegistrationSchema.js
```diff
@@ -0,0 +1,96 @@
...
+      return modelProperties;
+    }
+  });
+
+  var Schema = BaseSchema.Model.extend({
+    expand: ['schema'],
+
+    constructor: function () {
+      this.properties = new RegistrationSchemaPropertyCollection();
+      Okta.BaseModel.apply(this, arguments);
```
hor-kanchan-okta: I want to set `properties` to my custom collection in the constructor. So the `parse` function can set to the right object.

In the constructor of BaseSchema.Model, it also set `properties` it too. So I have to overwrite the behavior.

hor-kanchan-okta:
@rchild-okta I would address your API comment before public in a different PR .

src/models/RegistrationSchema.js
```diff
@@ -0,0 +1,94 @@
...
+    createModelProperties: function () {
+      var modelProperties = SchemaProperty.Collection.prototype.createModelProperties.apply(this);
+      _.each(modelProperties, function(modelProperty, name) {
+        modelProperty.required = !!this.get(name).get('required');
+      }, this);
+      return modelProperties;
+    }
+  });
+
+  var Schema = BaseSchema.Model.extend({
```
mauriciocastillosilva-okta: nit: usually we do ```return BaseSchema.Model.extend({``` instead of creating a var

src/util/PasswordComplexityUtil.js
mauriciocastillosilva-okta: I see... I think when that time comes we can rename whichever functions we need to something else... so treat it is as a nit, but we don't really know if we will have more functions for checking if the complexity is enabled, or if we can just reuse this same one and check the type of the value within this function

src/models/RegistrationSchema.js
mauriciocastillosilva-okta: it just seems odd that we are extending ```BaseSchema``` but the constructor calls ```BaseModel``` instead... cause then if we enhance or fix something in ```BaseSchema```'s constructor we won't get the fix/enhancement here.
Can we call the parent constructor first and then override this.properties with whatever is needed here?

src/views/registration/PasswordComplexity.js
```diff
@@ -0,0 +1,47 @@
...
+    },
+    template: '<div>{{message}}</div>',
+    getTemplateData: function () {
+      return {
+        message: this.message
+      };
+    }
+  });
+
+  var View =  Okta.View.extend({
```
mauriciocastillosilva-okta: nit: ```return Okta.View.extend({``` so you avoid creating the extra var

test/unit/spec/PasswordComplexityUtil_spec.js
```diff
@@ -0,0 +1,288 @@
...
+          beforeEach(function () {
+            this.doesComplexityMeet = _.partial(this.complexity.doesComplexityMeet, 4);
+          });
+
+          it('a password with 4 numbers meets the complexity', function () {
+            expect(this.doesComplexityMeet('12abCD34')).toBe(true);
+          });
+
+          it('a password with 3 numbers does not meet the complexity', function () {
+            expect(this.doesComplexityMeet('12abCD 4')).toBe(false);
```
mauriciocastillosilva-okta: should we allow spaces in the password? it should probably be the same string but without the space ```'12abCD4'```

test/unit/spec/PasswordComplexityUtil_spec.js
```diff
@@ -0,0 +1,288 @@
...
+          describe('username is empty', function () {
+            beforeEach(function () {
+              var myModel = new (Okta.Model.extend({
+                props: {'login': ['string', false, '']}
+              }))();
+              this.doesComplexityMeet = _.partial(
+                  this.complexities['excludeUsername'].doesComplexityMeet, true, _, myModel);
+            });
+
+            it('any passowrd does not the complexity', function () {
```
mauriciocastillosilva-okta: typo: ```password```

test/unit/spec/RegistrationSchema_spec.js
```diff
@@ -0,0 +1,312 @@
...
+          expect(this.schema.properties.length).toEqual(3);
+        });
+
+        it('field1 is required', function () {
+          expect(this.schema.properties.get('field1').get('required')).toBe(true);
+          expect(this.schema.properties.createModelProperties()['field1'].required).toBe(true);
+        });
+
+        it('field2 is not required', function () {
+          expect(this.schema.properties.get('field2').get('required')).toBeFalsy();
```
mauriciocastillosilva-okta: nit: ```toBe(false)```

src/models/RegistrationSchema.js
hor-kanchan-okta: The issue is `BaseModel.constructor` will invoke `parse()`. If I override the `properties` after calling `BaseSchema`. Then I need some special handling in `parse()`.

test/unit/spec/PasswordComplexityUtil_spec.js
hor-kanchan-okta: I think it is up to the server to allow space or not.  But I will update the test to avoid confusion.

src/models/RegistrationSchema.js
hor-kanchan-okta: Created OKTA-130291 to address this when we move the schema logic to Courage.

PR #236, opened 2017-06-13
:fix: Send validatePhone:false if user insists that the phone number is valid

alexeystadnikov-okta:
screen shot 2017-06-13 at 4 26 59 pm

Resolves OKTA-122419

@mauriciocastillosilva-okta 
@lboyette-okta 
Please, review.


test/unit/spec/EnrollSms_spec.js
```diff
@@ -16,11 +16,12 @@ define([
   'helpers/xhr/MFA_ENROLL_ACTIVATE_success',
   'helpers/xhr/MFA_ENROLL_ACTIVATE_error',
   'helpers/xhr/MFA_ENROLL_ACTIVATE_errorActivate',
+  'helpers/xhr/MFA_ENROLL_ACTIVATE_phone_error.js',
```
mauriciocastillosilva-okta: I think it will be better to rename that file to ```MFA_ENROLL_ACTIVATE_invalid_phone.js```, so we are a little more explicit on what the error is

test/unit/spec/EnrollSms_spec.js
```diff
@@ -16,11 +16,12 @@ define([
...
   'helpers/xhr/SUCCESS',
   'LoginRouter'
 ],
 function (Q, _, $, OktaAuth, LoginUtil, Util, AuthContainer, Form, Beacon, Expect, $sandbox,
-          resAllFactors, resExistingPhone, resEnrollSuccess, resEnrollError, resActivateError,
+          resAllFactors, resExistingPhone, resEnrollSuccess, resEnrollError, resActivateError, resEnrollPhoneError,
```
mauriciocastillosilva-okta: same here, it will be ```resEnrollInvalidPhone```

src/EnrollCallAndSmsController.js
```diff
@@ -114,28 +115,34 @@ function (Okta, FormController, Footer, PhoneTextBox, TextBox, CountryUtil, Form
...
+            var factor = _.findWhere(trans.factors, {
               factorType: self.get('factorType'),
               provider: 'OKTA'
             });
             return factor.enroll({
               profile: profileData
+            })
+            .fail(function (error) {
+              if(error.errorCode === 'E0000098') { // E0000098: "This phone number is invalid."
+                error.xhr.responseJSON.errorSummary += ' If you think the number is valid, please try again.';
```
mauriciocastillosilva-okta: just a reminder to use the localized message once the i18n PR is verified by Vince and merged.

PR #237, opened 2017-06-14
Fix colliding CSS issue

lboyette-okta:
We need to use `inherit_select_classes` and add a special class to our select inputs to ensure only our chosen inputs get the styles https://harvesthq.github.io/chosen/options.html

screen shot 2017-06-14 at 4 16 21 pm

### DIFF
--- a/assets/sass/okta-sign-in.scss
+++ b/assets/sass/okta-sign-in.scss
@@ -1,33 +1,125 @@
-
 @import 'widgets/jquery.qtip';
-@import 'widgets/mega-drop-down';
-@import 'widgets/chosen';
 
-@import 'common/enduser/reset';
 @import 'common/shared/helpers/all';
-@import 'common/admin/icons/all';
 @import 'variables';
-@import 'common/enduser/responsive-variables';
-@import 'common/enduser/helpers';
-@import 'common/shared/o-forms/o-form';
-@import 'common/admin/modules/infobox/infobox';
 @import 'fonts';
 @import 'helpers';
-@import 'base';
-@import 'layout';
-@import 'modules/forms';
-@import 'modules/infobox';
+@import 'common/enduser/responsive-variables';
+@import 'common/enduser/helpers';
+@import 'container';
+
+// scss-lint:disable IdSelector
+#okta-sign-in {
+// scss-lint:enable IdSelector
+  @import 'common/admin/icons/all';
+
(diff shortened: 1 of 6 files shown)


assets/sass/okta-theme.scss
```diff
@@ -57,7 +57,7 @@
 
 
 // scss-lint:disable IdSelector
-#okta-sign-in {
+#okta-sign-in.auth-container {
```
mauriciocastillosilva-okta: I think all will still work without adding ```.auth-container``` since they are the same element, but in any case, it doesn't hurt.

lboyette-okta:
@mauriciocastillosilva-okta We actually don't have a fix for chosen. This is a compromise we're making for now to cover 95% of collisions. Developers using chosen (and qtip) will still have collisions with their own css.

assets/sass/okta-theme.scss
lboyette-okta: The problem is that we changed the specificity of the other classes to be at the ID level. Originally the ID level was enough to override other styles, but now we have to add the class to increase the specificity.

assets/sass/okta-sign-in.scss
```diff
@@ -1,33 +1,125 @@
...
+#okta-sign-in {
+// scss-lint:enable IdSelector
+  @import 'common/admin/icons/all';
+
+  @import 'common/enduser/reset';
+  @import 'widgets/mega-drop-down';
+  @import 'widgets/chosen';
+  @import 'common/shared/o-forms/o-form';
+  @import 'common/admin/modules/infobox/infobox';
+  @import 'base';
```
mauriciocastillosilva-okta: there is one style inside base that might not make sense to be under ```#okta-sign-in```, which is ```html: { height: 100%; }```

assets/sass/okta-sign-in.scss
```diff
@@ -1,33 +1,125 @@
...
+.chzn-container-active {
+  margin: 0;
+  padding: 0;
+  border: 0;
+  outline: 0;
+  font: inherit;
+  vertical-align: baseline;
+  background: transparent;
+  font-size: 13px;
+  font-family: "montserrat", Arial, Helvetica, sans-serif;
```
mauriciocastillosilva-okta: should this be ```$fonts``` instead? Same question for ```font-size``` one line above

PR #238, opened 2017-06-18
experiment

manueltanzi-okta:
@hor-kanchan-okta @samerfanek-okta Need an opinion on this.
I managed to handle the 3 states using the css priority order. I think this way is more elegant.
The other way would be use only one span and change it's class to change the icon, that would mean many addClass and removeClass in the code.

(Don't focus on the colors but on the logic, need to check with UX on Monday for colors.)

What do you think? How would you have done this?

temp

### DIFF
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -128,7 +128,7 @@ function (
       });
       this.model = new Model();
 
-      var checkPasswordMeetComplexities = function(model) {
+      var checkPasswordMeetComplexities = function(model, alert) {
         var password = model.get('password') || '';
         _.each(schema.passwordComplexity.enabledComplexities, function(complexityName) {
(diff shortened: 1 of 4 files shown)


src/RegistrationController.js
```diff
@@ -137,6 +137,9 @@ function (
             ele.addClass('password-complexity-meet');
           } else {
             ele.removeClass('password-complexity-meet');
+            if(alert) {
+              ele.addClass('password-complexity-alert');
```
hor-kanchan-okta: You should be to simplify the logic by limiting to only have one `class` value.
Also the `blur` event always happen after `input` event. That could have impact on your flow too.

manueltanzi-okta: will do it!

Issue #240, opened 2017-06-19
Missing optional dependencies WARNING should be removed


pith:
When installing the okta widget npm is complaining about missing optional dependencies. Do you know how it could be fixed ?

package.json
```
{
  "dependencies": {
    "@okta/okta-signin-widget": "^1.9.0",
    ...
  },
  ...
}

```

`npm install` output:

```
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: @okta/i18n@26.21.0 (node_modules/@okta/okta-signin-widget/node_modules/@okta/i18n):
npm WARN 404 SKIPPING OPTIONAL DEPENDENCY: Not found : @okta/i18n
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: @okta/ci-pkginfo@1.2.0 (node_modules/@okta/okta-signin-widget/node_modules/@okta/ci-pkginfo):
npm WARN 404 SKIPPING OPTIONAL DEPENDENCY: Not found : @okta/ci-pkginfo
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: @okta/ci-update-package@1.4.0 (node_modules/@okta/okta-signin-widget/node_modules/@okta/ci-update-package):
npm WARN 404 SKIPPING OPTIONAL DEPENDENCY: Not found : @okta/ci-update-package
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: @okta/install-with-shrinkwrap@1.6.0 (node_modules/@okta/okta-signin-widget/node_modules/@okta/install-with-shrinkwrap):
npm WARN 404 SKIPPING OPTIONAL DEPENDENCY: Not found : @okta/install-with-shrinkwrap
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: @okta/courage@2.1.0 (node_modules/@okta/okta-signin-widget/node_modules/@okta/courage):
npm WARN 404 SKIPPING OPTIONAL DEPENDENCY: Not found : @okta/courage
```

jmelberg-okta:
Hi @pith. Since they are **optional dependencies**, the install will not fail if they cannot be found.

You can run `npm install --no-optional` if you'd like to skip the optional dependencies.

pith:
This is not the case with the 1.12.0 cf #246. And I don't understand why these dependencies are not on NPM (https://www.npmjs.com/package/@okta/qtip2). Globally passing `--no-optional` don't seems like a clean solution.

Hobbit44:
[...]

PR #249, opened 2017-06-27
Remove optional dependencies since they're checked into GitHub

### LINKED ISSUE #191: Installing Widget doesn't work with Yarn
```bash
yarn add @okta/okta-signin-widgetyarn add v0.21.3
[1/4] 🔍  Resolving packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/@okta%2fci-update-package: Not found".
info If you think this is a bug, please open a bug report with the information provided in "/Users/mraible/angular-okta-example/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

It works fine with npm.

mraible:
Fixes #191 and #240.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,13 +91,5 @@
     "q": "1.4.1",
     "u2f-api-polyfill": "0.4.1",
     "underscore": "1.8.3"
-  },
-  "optionalDependencies": {
-    "@okta/ci-pkginfo": "1.2.0",
-    "@okta/ci-update-package": "1.4.0",
-    "@okta/courage": "2.11.0",
-    "@okta/i18n": "28.15.0",
-    "@okta/install-with-shrinkwrap": "1.7.0",
-    "@okta/qtip2": "3.0.6"
   }
 }


jmelberg-okta:
@rchild-okta @mauriciocastillosilva-okta This fix can help with the issues raised in #191 and #240.

rchild-okta:
Interesting - we ran into issues earlier before because we never just *removed* the dependencies, but had them listed in dependencies/devDependencies. This is actually pretty nice - checked into source, and not in the package.json so npm doesn't try to do a lookup.

One thing to fix - this will break our internal build (specifically, [this code](https://github.com/okta/okta-signin-widget/blob/master/ci-scripts/okta-signin-widget-build.sh#L79)). You need to add [these two lines](https://github.com/okta/stormpath-migration/blob/master/scripts/publish.sh#L7-L8) to that file.

jmelberg-okta:
Closing due to #250

PR #250, opened 2017-06-30
:fire: Removes already checked in optional dependencies

jmelberg-okta:
- :arrow_up: Builds on top of @mraible's work ( #249 ) to make sure internal build does not fail -- per @rchild-okta's comments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,13 +91,5 @@
     "q": "1.4.1",
     "u2f-api-polyfill": "0.4.1",
     "underscore": "1.8.3"
-  },
-  "optionalDependencies": {
-    "@okta/ci-pkginfo": "1.2.0",
-    "@okta/ci-update-package": "1.4.0",
-    "@okta/courage": "2.11.0",
-    "@okta/i18n": "28.25.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -91,13 +91,5 @@
     "q": "1.4.1",
     "u2f-api-polyfill": "0.4.1",
     "underscore": "1.8.3"
-  },
-  "optionalDependencies": {
-    "@okta/ci-pkginfo": "1.2.0",
-    "@okta/ci-update-package": "1.4.0",
-    "@okta/courage": "2.11.0",
-    "@okta/i18n": "28.25.0",
-    "@okta/install-with-shrinkwrap": "1.7.0",
```
rchild-okta: As discussed, also need to remove this from package.json:
https://github.com/okta/okta-signin-widget/blob/master/package.json#L43

sinn1:
Is there a timeline for the next release that will include this? We're experiencing the same issue as #191.

rlueder:
@sinn1 I was able to install the widget using yarn (you can see the merge into master just above your comment).

PR #251, opened 2017-07-05
:bug: Fixes browser back button issue

gowthamidommety-okta:
Okta SignIn widget doesn't work if user clicks on browser back button after verify call

Resolves: OKTA-114121


gowthamidommety-okta:
Closed this as we decided a different approach

PR #253, opened 2017-07-07
:bug: Autopopulates username during On-Prem MFA

gowthamidommety-okta:
Resolves: OKTA-128212

screen shot 2017-07-07 at 10 03 13 am

### DIFF
--- a/src/EnrollOnPremController.js
+++ b/src/EnrollOnPremController.js
@@ -38,7 +38,7 @@ function (Okta, FormType, FormController, Footer, TextBox) {
       var factors = this.options.appState.get('factors');
       var factor = factors.findWhere(_.pick(this.options, 'provider', 'factorType'));
       var profile = factor.get('profile');
-      var credentialId = profile.credentialId || '';
+      var credentialId = (profile && profile.credentialId) ?  profile.credentialId : '';
       return {
         props: {
           credentialId: ['string', true, credentialId],
--- a/src/EnrollOnPremController.js
+++ b/src/EnrollOnPremController.js
@@ -35,9 +35,13 @@ function (Okta, FormType, FormController, Footer, TextBox) {
     },
     Model: function () {
       var provider = this.options.provider;
+      var factors = this.options.appState.get('factors');
+      var factor = factors.findWhere(_.pick(this.options, 'provider', 'factorType'));
+      var profile = factor.get('profile');
+      var credentialId = profile.credentialId || '';
       return {
         props: {
-          credentialId: ['string', true],
+          credentialId: ['string', true, credentialId],
           passCode: ['string', true],
           factorId: 'string'
         },
@@ -95,16 +99,7 @@ function (Okta, FormType, FormController, Footer, TextBox) {
       };
     },
 
-    Footer: Footer,
-
-    initialize: function () {
-      var factors = this.options.appState.get('factors');
-      var factor = factors.findWhere(_.pick(this.options, 'provider', 'factorType'));
-      var profile = factor.get('profile');
-      if(profile && profile['credentialId']) {
-        this.model.set('credentialId', profile['credentialId']);
-      }
-    }
+    Footer: Footer
 
   });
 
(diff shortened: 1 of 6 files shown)


src/EnrollOnPremController.js
```diff
@@ -95,7 +95,16 @@ function (Okta, FormType, FormController, Footer, TextBox) {
       };
     },
 
-    Footer: Footer
+    Footer: Footer,
+
+    initialize: function () {
+      var factors = this.options.appState.get('factors');
+      var factor = factors.findWhere(_.pick(this.options, 'provider', 'factorType'));
```
mauriciocastillosilva-okta: I think this might be simpler if it is added in the Model, around lines 36-44, using the default value for the model prop. Something like
```
var provider = this.options.provider;
var credentiaId = undefined // Get the default value here
return {
  props: {
    credentialId: ['string', true, credentialId],
    passCode: ['string', true],
    factorId: 'string'
  },
```

gowthamidommety-okta: Ok, i'll change this

src/EnrollOnPremController.js
```diff
@@ -35,9 +35,13 @@ function (Okta, FormType, FormController, Footer, TextBox) {
     },
     Model: function () {
       var provider = this.options.provider;
+      var factors = this.options.appState.get('factors');
+      var factor = factors.findWhere(_.pick(this.options, 'provider', 'factorType'));
+      var profile = factor.get('profile');
```
samerfanek-okta: Is it guaranteed that every defined factor has a ``profile``? Just making sure we don't hit an undefined error in line 41 in case ``profile`` turns out to be ``undefined``.

gowthamidommety-okta: If the factor is configured it will always send `profile`. However it's better to put an extra check in. Addressed this in my latest commit

gowthamidommety-okta:
@mauriciocastillosilva-okta Sorry missed your earlier comment. Addressed it now. Please review

test/unit/spec/EnrollOnPrem_spec.js
```diff
@@ -99,6 +99,22 @@ function (Q, _, $, OktaAuth, Util, Form, Beacon, Expect, $sandbox,
             Expect.isTextField(test.form.credentialIdField());
           });
         });
+        itp('autopopulates credentialId text field', function () {
+          return setup().then(function (test) {
+            expect(test.form.getCredentialId()).toEqual('test123');
+          });
+        });
+        itp('does not autopopulate credentialId when profile does not exist', function () {
+          delete resAllFactors['response']['_embedded']['factors'][3]['profile'];
```
mauriciocastillosilva-okta: I think it will be a good idea to have a ```setup``` method that takes care of copying and deleting the object we don't need. Something like ```setupWithoutProfile```

test/unit/spec/EnrollOnPrem_spec.js
```diff
@@ -60,7 +60,25 @@ function (Q, _, $, OktaAuth, Util, Form, Beacon, Expect, $sandbox,
       });
     }
 
-    var setupOnPrem = _.partial(setup, true);
+    var setupOnPrem = _.partial(setup, resAllFactorsOnPrem, true);
+
+    var getResponseNoProfile = function (response, factorType, provider) {
+      var copy = $.extend(true, {}, response);
```
mauriciocastillosilva-okta: we should do a deepCopy similar to what we have in https://github.com/okta/okta-signin-widget/blob/master/test/unit/spec/EnrollChoices_spec.js#L56

nit: can we rename ```copy``` to ```res```  or ```responseCopy```

test/unit/spec/EnrollOnPrem_spec.js
```diff
@@ -60,7 +60,25 @@ function (Q, _, $, OktaAuth, Util, Form, Beacon, Expect, $sandbox,
...
+      var factors = copy['response']['_embedded']['factors'];
+      var factor = _.findWhere(factors, {factorType: factorType, provider: provider});
+      delete factor['profile'];
+      return copy;
+    };
+
+    var setupRsaNoProfile = function () {
+      var copy = getResponseNoProfile(resAllFactors, 'token', 'RSA');
```
mauriciocastillosilva-okta: same nit here, ```var res = ``` and in setupOnPremNoProfile

PR #258, opened 2017-07-10
:bug: Fixes Bacon build script for non-root permissions

jmelberg-okta:
Resolves: OKTA-133184


mauriciocastillosilva-okta:
This will probably won't be needed once we migrate to the new CI

PR #259, opened 2017-07-11
:art: Making new ci scripts.

kelvinzhu-okta:
Fixes okta internal CI scripts and transfers them to our new CI system

Resolves: OKTA-133195


scripts/setup.sh
```diff
@@ -0,0 +1,28 @@
...
+# Install required dependencies
+npm install -g @okta/ci-update-package
+npm install -g @okta/ci-pkginfo
+
+if ! bundle install; then
+  echo "bundle install failed! Exiting..."
+  exit ${FAILED_SETUP}
+fi
+
+if ! npm install --no-optional --unsafe-perm; then
```
kelvinzhu-okta: The no-optional is necessary due to fsevents. --unsafe-perm is necessary due to new CI running as the root user.

package.json
```diff
@@ -90,5 +90,8 @@
     "q": "1.4.1",
     "u2f-api-polyfill": "0.4.1",
     "underscore": "1.8.3"
+  },
+  "optionalDependencies": {
+    "fsevents": "*"
```
mauriciocastillosilva-okta: why do we need the fsevents here? 
We've had problems with this before, so at some point we removed it from npm-shrinkwrap. Then I think someone checked it back in and builds have been working fine so far...

rchild-okta: It is interesting that this is required - usually the dependencies we pull in should have it listed as optional. I'm okay with it if for some reason it's necessary.

scripts/publish.sh
```diff
@@ -0,0 +1,34 @@
...
+if [ -n "$action_branch" ];
+then
+  echo "Publishing from bacon task using branch $action_branch"
+  TARGET_BRANCH=$action_branch
+else
+  echo "Publishing from bacon testSuite using branch $BRANCH"
+  TARGET_BRANCH=$BRANCH
+fi
+
+if ! npm run ci-update-package -- --branch ${TARGET_BRANCH}; then
```
rchild-okta: It looks like this was missed when we removed the optionalDependencies, but since we're installing ci-update-package & ci-pkg-info globally now, two things:

1. Remove these lines from package.json: https://github.com/okta/okta-signin-widget/blob/master/package.json#L41-L42

2. Here, you no longer need to do an "npm run", just call `ci-update-package --branch ${TARGET_BRANCH}` and `ci-pkginfo -t dataload etc` below

package.json
kelvinzhu-okta: Apparently since the fsevents was part of the npm-shrinkwrap, it started failing hard on npm install unless I added this in. I guess the other option would be to remove it from the npm-shrinkwrap? I wasn't sure which was the better way to go.

rchild-okta: Gotcha, yeah the shrinkwrap thing is another problem that we'll need to fix now that npm5 is out, for now let's just go with the optional route.

PR #260, opened 2017-07-12
Moves private modules to packages, and uses package-lock.json

rchild-okta:
- Moves checked in node_modules to packages folder
- Swaps out npm-shrinkwrap with npm5's package-lock.json
- Config changes to read from packages

RESOLVES OKTA-132800


.gitignore
```diff
@@ -1,12 +1,6 @@
 # If a change is made in here, add it to .npmignore as well unless you
 # explicitly want to include it in the published artifact
-node_modules/*
-!node_modules/@okta
-node_modules/@okta/*
-!node_modules/@okta/i18n
-!node_modules/@okta/courage
-!node_modules/@okta/jquery.simplemodal
-!node_modules/@okta/qtip2
+node_modules
```
samerfanek-okta: Do you want to add something like:
```
/packages/*
!packages/@okta
!packages/@okta/*
etc..
```

package.json
```diff
@@ -83,6 +83,7 @@
   "dependencies": {
     "@okta/okta-auth-js": "1.8.0",
     "backbone": "1.2.1",
+    "clipboard": "1.6.1",
```
samerfanek-okta: Is this a new necessary dependency?

Gruntfile.js
```diff
@@ -97,7 +97,7 @@ module.exports = function (grunt) {
           // jquery.qtip.css -> _jquery.qtip.scss
           {
             expand: true,
-            cwd: 'node_modules/@okta/qtip2/dist/',
+            cwd: 'packages/@okta/qtip2/dist/',
```
samerfanek-okta: Would like 444 of the ``retire`` Gruntfile task need to be updated as well?

.gitignore
rchild-okta: Not really - the old gitignore was basically saying "ignore node_modules except for everything under @okta, except for things in @okta/i18n that are not the json/properties files". The things I copied over are exactly what we want to check in (and if they were changed, would need to be checked into source).

package.json
rchild-okta: It is - the one downside to this approach (over using something like [lerna](http://lernajs.io/) which is the direction we should move towards) is that by checking in these dependencies, we have no way of alerting npm about *their* dependencies. In this case, clipboard is the one dependency we need to pull in.

Btw, this is going to be more of an issue as we're currently moving all our courage dependencies out of source and into package.json.

.gitignore
samerfanek-okta: I see, so the ``packages`` directory won't contain extra stuff like the past ``node_modules`` directory did?

Gruntfile.js
rchild-okta: It's a good question - the general purpose of retire.js is to see if we have dependencies that have hit a vulnerability list. This covers both the dependencies we're pulling down (in node_modules), as well as anything in the packages folder. You're right - I need to add packages to this list (will update now).

.gitignore
rchild-okta: Yep, although this will be a **manual** process for now until we get Lerna or write a script - updating one of these dependencies means installing it like normal, and then copying the files over to packages (and in the case of i18n, limiting the copy to the properties files).

PR #265, opened 2017-07-20
:bug: Prevents users from going back during factor verify

gowthamidommety-okta:
Resolves: OKTA-114121

### DIFF
--- a/src/MfaVerifyController.js
+++ b/src/MfaVerifyController.js
@@ -127,6 +127,15 @@ function (Okta, Checkbox, BaseLoginController, CookieUtil, TOTPForm, YubikeyForm
         }
       }
       return false;
+    },
+
+    back: function() {
+      // Empty function on verify controllers to prevent users
+      // from navigating back during 'verify' using the browser's
+      // back button. The URL will still change, but the view will not
+      // We will need to think of a better solution for this, possibly
+      // keeping track of the history of URL's and rendering views
+      // accordingly.
     }
   });
 
--- a/src/MfaVerifyController.js
+++ b/src/MfaVerifyController.js
@@ -133,9 +133,7 @@ function (Okta, Checkbox, BaseLoginController, CookieUtil, TOTPForm, YubikeyForm
       // Empty function on verify controllers to prevent users
       // from navigating back during 'verify' using the browser's
       // back button. The URL will still change, but the view will not
-      // We will need to think of a better solution for this, possibly
-      // keeping track of the history of URL's and rendering views
-      // accordingly.
+      // More details in OKTA-135060.
     }
   });
 
--- a/test/unit/spec/MfaVerify_spec.js
+++ b/test/unit/spec/MfaVerify_spec.js
@@ -2794,7 +2794,7 @@ function (Okta,
         .then(function (test) {
           expectHasRightBeaconImage(test, 'mfa-okta-sms');
           Util.triggerBrowserBackButton();
-          return test;
+          return tick(test);
         })
         .then(function (test) {
           //view is still the same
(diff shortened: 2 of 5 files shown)


test/unit/spec/MfaVerify_spec.js
```diff
@@ -2780,6 +2781,103 @@ function (Okta,
...
+          test.router.start();
+          expectHasRightBeaconImage(test, 'mfa-okta-security-question');
+          test.beacon.dropDownButton().click();
+          clickFactorInDropdown(test, 'SMS');
+          return tick(test);
+        })
+        .then(function (test) {
+          expectHasRightBeaconImage(test, 'mfa-okta-sms');
+          Util.triggerBrowserBackButton();
+          return test;
```
mauriciocastillosilva-okta: I think this should be ```return tick(test);``` so we wait for whatever needs to happen (which in this case is nothing, but that will give enough time in case someone changes the behavior of back in the future.

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2780,6 +2781,103 @@ function (Okta,
...
+          spyOn(Duo, 'init');
+          test.setNextResponse(resChallengeDuo);
+          test.beacon.dropDownButton().click();
+          clickFactorInDropdown(test, 'DUO');
+          return tick(test);
+        })
+        .then(function (test) {
+          expectHasRightBeaconImage(test, 'mfa-duo');
+          Util.triggerBrowserBackButton();
+          return test;
```
mauriciocastillosilva-okta: same here, I think it should be ```return tick(test);```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2780,6 +2781,103 @@ function (Okta,
...
+          expectHasRightBeaconImage(test, 'mfa-okta-security-question');
+          test.setNextResponse(resChallengeWebauthn);
+          test.beacon.dropDownButton().click();
+          clickFactorInDropdown(test, 'WINDOWS_HELLO');
+          return tick(test);
+        })
+        .then(function (test) {
+          expectHasRightBeaconImage(test, 'mfa-windows-hello');
+          Util.triggerBrowserBackButton();
+          return test;
```
mauriciocastillosilva-okta: same here, I think it should be ```return tick(test);```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2780,6 +2781,103 @@ function (Okta,
...
+          spyOn(window.u2f, 'sign');
+          test.setNextResponse(resChallengeU2F);
+          test.beacon.dropDownButton().click();
+          clickFactorInDropdown(test, 'U2F');
+          return tick(test);
+        })
+        .then(function (test) {
+          expectHasRightBeaconImage(test, 'mfa-u2f');
+          Util.triggerBrowserBackButton();
+          return test;
```
mauriciocastillosilva-okta: same here, I think it should be ```return tick(test);```

src/MfaVerifyController.js
```diff
@@ -127,6 +127,15 @@ function (Okta, Checkbox, BaseLoginController, CookieUtil, TOTPForm, YubikeyForm
         }
       }
       return false;
+    },
+
+    back: function() {
+      // Empty function on verify controllers to prevent users
+      // from navigating back during 'verify' using the browser's
+      // back button. The URL will still change, but the view will not
+      // We will need to think of a better solution for this, possibly
```
mauriciocastillosilva-okta: not sure we need to include the part of what we could do in the future. I would leave it out
@rchild-okta thoughts?

rchild-okta: No strong opinion. Slight preference to creating a jira that has this info, and adding the jira number here instead of the "We will need to think of ..."

PR #267, opened 2017-07-25
:seedling: custom sign-out url

mauriciocastillosilva-okta:
If a custom sign out url is provided, 'Sign out' links will redirect to it instead of navigating to the primary auth view.

Resolves: OKTA-129771

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,6 +588,12 @@ var signIn = new OktaSignIn(config);
     helpSupportNumber: '(123) 456-7890'
     ```
 
+- **signOutUrl:** url to redirect when a user clicks the sign out link. If not provided, the widget will navigate to primary auth.
+
+    ```javascript
+    signOutUrl: 'https://www.signmeout.com'
+    ```
+
 ## Username and password
 
 - **username:** Prefills the username input with the provided username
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -588,6 +588,12 @@ var signIn = new OktaSignIn(config);
     helpSupportNumber: '(123) 456-7890'
     ```
 
+- **signOutUrl:** url to redirect when a user clicks the sign out link. If not provided, the widget will navigate to primary auth.
```
rchild-okta: We have a couple different ways of setting links now, I think it'd be nice to:

1. Call this **signOutLink** - we refer to it below as **signOutLinkInMFA**. Also matches helpLinks.
2. Capitalize url since that's what we're doing elsewhere
3. Create a section "Customizing links and buttons" or something like that and group helpLinks, customButtons, registration, and this there.

rchild-okta: I am also curious if we want to move to making it more open by providing a click function instead of just a link that we hardcode to redirect since that's what we've been doing in the latest links/buttons we've been adding.

mauriciocastillosilva-okta: 1, 2 - will do
3 - I think I can create a section for Links and a sections for Buttons. It seems to me that so far we don't need to mixed them together

click function: I'm 50/50 in this one. So far links seems to be overriden with URLs, and buttons seem to have the click function. I think we can keep that pattern?

rchild-okta: Am okay with that and we can shift later when we change all of them at the same time. General idea is that we want everything to be as flexible as possible.

PR #269, opened 2017-07-25
:arrow_up: Upgrades dependency to support OSX Sierra

jmelberg-okta:
Fixes issue with PhantomJS causing `jasmine` tests to hang.

See: [PhantomJS not working on OSX Sierra](https://stackoverflow.com/questions/39739393/phantomjs-not-working-on-os-x-sierra)

Bumps CDN version to `1.13.0`

Verified:
  - Albatross
  - Sierra
  - El Capitan


README.md
```diff
@@ -67,16 +67,16 @@ To use the CDN, include links to the JS and CSS files in your HTML:
 ```html
 <!-- Latest CDN production Javascript and CSS: 1.11.0 -->
```
samerfanek-okta: Nit: This comment needs to be updated as well.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@okta/okta-signin-widget",
-  "version": "1.13.0",
+  "version": "2.1.0",
```
samerfanek-okta: Why are we bumping up this version?

package-lock.json
```diff
@@ -725,20 +725,13 @@
       "dev": true
     },
     "clipboard": {
-      "version": "1.7.1",
-      "resolved": "https://registry.npmjs.org/clipboard/-/clipboard-1.7.1.tgz",
-      "integrity": "sha1-Ng1taUbpmnof7zleQrqStem1oWs=",
+      "version": "1.6.1",
+      "resolved": "https://registry.npmjs.org/clipboard/-/clipboard-1.6.1.tgz",
+      "integrity": "sha1-ZcW2VIEkZrD6q4Lca6Dx0vjkvlM=",
```
samerfanek-okta: Is there a reason for this package downgrade?

samerfanek-okta: Similar comment to ``tiny-emitter`` and ``jasmine-core`` below.

package-lock.json
```diff
@@ -5412,6 +5317,88 @@
         }
       }
     },
+    "phantomjs-prebuilt": {
+      "version": "2.1.14",
```
samerfanek-okta: Just checking, did you verify that this version works on earlier versions than OSX Sierra?

package-lock.json
jmelberg-okta: This was regenerated when we updated the dependencies. Current version is `2.1.0`

package-lock.json
jmelberg-okta: This is a strange change. I'll make sure to revert these back.

jmelberg-okta: RE: `clipboard` -> Bumped version to `1.7.1`. This trickled down and updated `tiny-emitter`.

RE: `jasmine-core`: Oddly, `grunt-contrib-jasmine` depends on a previous version of `jasmine-core`.

Verified that the tests pass 👍

package-lock.json
jmelberg-okta: It appears that our `.travis.yaml` file specifies the OS to be `linux`, so I am not sure if we check different versions of OSX. Curious if we should update the script to include different OS versions?

It appears to be pretty straightforward:
https://docs.travis-ci.com/user/reference/osx/#OS-X-Version

Issue #221, opened 2017-05-11
Browserify Error: Cannot find module './View' from '/private/tmp/okta_signing-widget_pb/node_modules/@okta/okta-signin-widget/dist/js'


bsergean:
(I love this sign in widget, it's quite simple to integrate; my only mistake was that I was using an HTTP connection instead of https and I was having weird errors.)

Here's a simple repro case.

```
$ cd /tmp/okta_signing-widget_pb
okta_signing-widget_pb$ ls
node_modules	package.json	repro.js
okta_signing-widget_pb$ cat package.json 
{
  "dependencies": {
    "@okta/okta-signin-widget": "^1.11.0"
  },
  "devDependencies": {
    "browserify": "^14.3.0"
  }
}
okta_signing-widget_pb$ cat repro.js 
var OktaSignIn = require('@okta/okta-signin-widget')
```

Now If I try to run browserify I get this error:

```
./node_modules/.bin/browserify repro.js 
Error: Cannot find module './View' from '/private/tmp/okta_signing-widget_pb/node_modules/@okta/okta-signin-widget/dist/js'
    at /private/tmp/okta_signing-widget_pb/node_modules/browser-resolve/node_modules/resolve/lib/async.js:55:21
    at load (/private/tmp/okta_signing-widget_pb/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
    at onex (/private/tmp/okta_signing-widget_pb/node_modules/browser-resolve/node_modules/resolve/lib/async.js:92:31)
    at /private/tmp/okta_signing-widget_pb/node_modules/browser-resolve/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:114:15)
```

manueltanzi-okta:
Hi @bsergean !

The problem is that browserify  attempts to resolve paths in require() calls in the already bundled okta-signin-widget, while it should ignore them since the okta-sign-in.entry.js is a standalone file and doesn't need the external modules, they are already bundled in the same file.

This is a browserify bug, it also happens when you try to require another file bundled by browserify (see https://github.com/substack/node-browserify/pull/1151).

A solution to this is to not parse the required okta-signin-widget.
You can use this code to do that:
`browserify --noparse=$PWD/node_modules/@okta/okta-signin-widget/dist/js/okta-sign-in.entry.js repro.js `

PR #279, opened 2017-08-11
:memo: Add documentation to explain how to correctly Browserify the okta-signin-widget module

manueltanzi-okta:
:memo: Add documentation to explain how to correctly Browserify the okta-signin-widget module

When developers tried to browserify an app that require the okta-signin-widget, the bundling failed because Browserify attempts to resolve paths in require() calls in the already bundled okta-signin-widget, while it should ignore them since the okta-sign-in.entry.js is a standalone file and doesn't need the external modules, they are already bundled in the same file. Added documentation to explain how to tell to browserify to not parse the okta-signin-widget module and run correctly.

Resolves: [OKTA-135028](https://oktainc.atlassian.net/browse/OKTA-135028)

@rchild-okta @rajaguruduraisamy-okta @haishengwu-okta @samerfanek-okta

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,16 +151,12 @@ After running `npm install`:
     // Use OktaSignIn
     var signIn = new OktaSignIn(/* configOptions */);
     ```
-3. [Browserify](http://browserify.org/) only: When you want to bundle your application with Browserify, you need to tell it that it doesn't have to parse the okta-signin-widget module. To do this, you need to add this flag when you run the `browserify` command:
-
+    **Note:** If you use [Browserify](http://browserify.org/) to bundle your app, you'll need to use the `--noparse` option:
     ```
-    // Don't parse @okta/okta-signin-widget
-    --noparse=$PWD/node_modules/@okta/okta-signin-widget/dist/js/okta-sign-in.entry.js
-
-    // Example
-    browserify --noparse=$PWD/node_modules/@okta/okta-signin-widget/dist/js/okta-sign-in.entry.js main.js -o bundle.js
+    browserify main.js \
+    --noparse=$PWD/node_modules/@okta/okta-signin-widget/dist/js-okta-sign-in.entry.js \
+    --outfile=bundle.js
     ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -151,6 +151,16 @@ After running `npm install`:
     // Use OktaSignIn
     var signIn = new OktaSignIn(/* configOptions */);
     ```
+3. [Browserify](http://browserify.org/) only: When you want to bundle your application with Browserify, you need to tell it that it doesn't have to parse the okta-signin-widget module. To do this, you need to add this flag when you run the `browserify` command:
```
rchild-okta: I think we can simplify this a bit:

**Note:** If you use [Browserify](http://browserify.org/) to bundle your app, you'll need to use the `--noparse` option: 

```bash
browserify main.js \
  --noparse=$PWD/node_modules/@okta/okta-signin-widget/dist/js-okta-sign-in.entry.js \
  --outfile=bundle.js
```

rchild-okta: Also, can just have it inline rather than creating a new bullet point for it

manueltanzi-okta: Sure, thank you!

README.md
```diff
@@ -151,6 +151,16 @@ After running `npm install`:
...
+
+    ```
+    // Don't parse @okta/okta-signin-widget
+    --noparse=$PWD/node_modules/@okta/okta-signin-widget/dist/js/okta-sign-in.entry.js
+
+    // Example
+    browserify --noparse=$PWD/node_modules/@okta/okta-signin-widget/dist/js/okta-sign-in.entry.js main.js -o bundle.js
+    ```
+    The reason is that if you run the command without this flag, Browserify will attempt to resolve paths in `require()` calls in the already bundled okta-signin-widget, while it should ignore them since the `okta-sign-in.entry.js` is a standalone file and doesn't need the external modules, they are already bundled in the same file.
```
manueltanzi-okta: @rchild-okta should we also remove the reason?

rchild-okta: Yeah, let's remove the reason. My feeling is that the README is more about how to do things rather than an in-depth explanation (especially in the case of how external libraries like browserify work) - if they're curious, I'd imagine they can either do a git blame to see when this commit was added (and see that it's linked to the issue where the reason was given) or just look up what no-parse does.

rchild-okta:
Also, adding a note here that this is related to the issue brought up by #221 - once we merge this doc update, let's close that issue out.

PR #287, opened 2017-08-22
:seedling: Adds option to hide email from recovery flow

gowthamidommety-okta:
Resolves: OKTA-136405

screen shot 2017-08-22 at 10 27 13 am

screen shot 2017-08-22 at 3 25 38 pm


gowthamidommety-okta:
@mauriciocastillosilva-okta @sergiymelnyk-okta Please review

src/ForgotPasswordController.js
```diff
@@ -74,17 +78,25 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, ContactSupport,
           });
         })
         .fail(function () {
-          self.set('factorType', Enums.RECOVERY_FACTOR_TYPE_EMAIL);
```
mauriciocastillosilva-okta: should we remove the whole ```.fail``` code? It seems we don't need it anymore

src/ForgotPasswordController.js
```diff
@@ -74,17 +78,25 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, ContactSupport,
...
-          FormType.Input({
+        /*eslint complexity: [2, 9] */
+        var smsEnabled = this.settings.get('features.smsRecovery');
+        var callEnabled = this.settings.get('features.callRecovery');
+        var emailEnabled = this.settings.get('features.emailRecovery');
+        var noFactorsEnabled = !(smsEnabled || callEnabled || emailEnabled);
+        var formChildren = [];
+
+        if(noFactorsEnabled) {
+          this.trigger('noFactorsEnabled');
```
mauriciocastillosilva-okta: can we just add here the html instead of triggering an event?

src/ForgotPasswordController.js
gowthamidommety-okta: 👍  will do

src/ForgotPasswordController.js
gowthamidommety-okta: I'll change this

src/ForgotPasswordController.js
gowthamidommety-okta: Actually, we need to leave this in. It breaks some tests when i take it out.

gowthamidommety-okta:
@mauriciocastillosilva-okta made changes, please review again

src/ForgotPasswordController.js
mauriciocastillosilva-okta: Then maybe we need to handle the error and show an error message?

gowthamidommety-okta: It never showed an error message before.. I think it is handled at another level

packages/@okta/i18n/dist/properties/login.properties
```diff
@@ -450,6 +450,7 @@ password.forgot.question.title = Answer Forgotten Password Challenge
 password.forgot.question.submit = Reset Password
 password.forgot.sms.notReceived = Didn't receive an SMS? Reset via email
 password.forgot.code.notReceived = Didn't receive a code? Reset via email
+password.forgot.noFactorsEnabled = No password reset options available. Please contact your administrator.
```
hor-kanchan-okta: @ahmedalbahar-okta Do you think UMG will hit this `password.forgot.noFactorsEnabled`?

mauriciocastillosilva-okta: They can't hit this, email is not optional for them. That is a new feature that is in beta.

src/ForgotPasswordController.js
```diff
@@ -93,72 +105,73 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, ContactSupport,
...
         this.listenTo(this.state, 'contactSupport', function () {
           this.add(ContactSupport, '.o-form-error-container');
         });
 
         this.listenTo(this, 'save', function () {
           this.options.appState.set('username', this.model.get('username'));
+          this.model.save();
         });
       },
       addRecoveryFactorButton: function (className, labelCode, factorType, form) {
```
mauriciocastillosilva-okta: nit: rename this to ```createRecoveryFactorButton``` or something similar, since this function no longer adds a button

packages/@okta/i18n/dist/properties/login.properties
gowthamidommety-okta: Just spoke to Srini. He said because this option is folded into password policy, it might be GA. Also, email is enabled by default so unless they disable it, they should not reach that state. @srinivasanagandla-okta do you have anything to add?

src/ForgotPasswordController.js
```diff
@@ -74,17 +78,25 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, ContactSupport,
...
-        var formChildren = [
-          FormType.Input({
+        /*eslint complexity: [2, 9] */
+        var smsEnabled = this.settings.get('features.smsRecovery');
+        var callEnabled = this.settings.get('features.callRecovery');
+        var emailEnabled = this.settings.get('features.emailRecovery');
+        var noFactorsEnabled = !(smsEnabled || callEnabled || emailEnabled);
+        var formChildren = [];
+
+        if(noFactorsEnabled) {
```
mauriciocastillosilva-okta: I think in general we prefer a space in between ```if``` and ```(```, so if you can change them to ```if (``` that will be great

src/UnlockAccountController.js
```diff
@@ -93,45 +104,50 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, ContactSupport,
...
-              }
-            }
-          }, { prepend: true });
-        }
-
         this.listenTo(this.state, 'contactSupport', function () {
           this.add(ContactSupport, '.o-form-error-container');
         });
+      },
+      addRecoveryFactorButton: function (className, labelCode, factorType, form) {
```
mauriciocastillosilva-okta: same nit as in the other file, since this function no longer adds the button itself, but now it returns a button

src/ForgotPasswordController.js
mauriciocastillosilva-okta: Ok. Can you then just add a comment on why we don't need to process the error? That will help in the future to know why that empty fail is there (same comment for the the unlock controller)

src/models/Settings.js
```diff
@@ -79,6 +79,7 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
       'features.hideSignOutLinkInMFA' : ['boolean', false, false],
       'features.customExpiredPassword': ['boolean', true, false],
       'features.registration': ['boolean', false, false],
+      'features.emailRecovery': ['boolean', false, true],
```
mauriciocastillosilva-okta: nit: can you move this next to the other recovery factors (about 9 lines above)?
Also 👍 for setting it to ```true``` by default, and not required, that will make it backwards compatible

packages/@okta/i18n/dist/properties/login.properties
mauriciocastillosilva-okta: @srinivasanagandla-okta and/or @gowthamidommety-okta can we just make sure that the backend sets it to true by default in loginpage_backbone?

gowthamidommety-okta: loginpage_backbone changes were made by @ericcarroll-okta. Just spoke to him, and his main change is here - https://github.com/okta/okta-core/blob/85b87decaebda6974f326a92fec59bff254f33e9/webapp/src/main/resources/META-INF/resources/WEB-INF/jsp/loginpage_backbone.jsp#L224. He said - Yes it is set to true by default. Old policies have it on by default. And it's on if not using group policies. 

I need to uncomment these lines https://github.com/okta/okta-core/blob/85b87decaebda6974f326a92fec59bff254f33e9/webapp/src/main/resources/META-INF/resources/WEB-INF/jsp/loginpage_backbone.jsp#L359 when i update the sign in widget in okta-core. The JIRA for that work is https://oktainc.atlassian.net/browse/OKTA-138947

PR #288, opened 2017-08-30
Generate README reading version from package.json

manueltanzi-okta:
Generate README reading version from package.json

Add grunt task to generate README.md file taking the version from package.json

Resolves : OKTA-135026

@hor-kanchan-okta @anishashetty-okta  @upteam-okta

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -555,6 +555,13 @@ module.exports = function (grunt) {
     ));
   });
 
+  grunt.task.registerTask('generate-readme', 'Generate Readme.md file using package.json', function () {
+    var readmeTpl = grunt.file.read('./README_TPL.md');
+    var version = grunt.file.readJSON('package.json').version;
+    var readme = readmeTpl.replace(/{{version}}/g, version);
+    grunt.file.write('./README.md', readme);
+  });
+
   grunt.task.registerTask('start-server', ['copy:server', 'connect:server']);
   grunt.task.registerTask('start-server-open', ['copy:server', 'connect:open']);
   grunt.task.registerTask('lint', ['scss-lint', 'search:noAbsoluteUrlsInCss', 'eslint']);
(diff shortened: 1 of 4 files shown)


manueltanzi-okta:
RelEng needs to update promote script in order to make this work

scripts/publish.sh
```diff
@@ -15,7 +15,10 @@ else
   TARGET_BRANCH=$BRANCH
 fi
 
-if ! ci-update-package --branch ${TARGET_BRANCH}; then
+if ci-update-package --branch ${TARGET_BRANCH};
+then
+  npm run generate-readme;
```
anishashetty-okta: Do we want to do this everytime we publish?

manueltanzi-okta: we update package.json so we can update also the README.md , but we should push this to github only on promotion

Gruntfile.js
```diff
@@ -555,6 +555,13 @@ module.exports = function (grunt) {
     ));
   });
 
+  grunt.task.registerTask('generate-readme', 'Generate Readme.md file using package.json', function () {
+    var readmeTpl = grunt.file.read('./README_TPL.md');
+    var version = grunt.file.readJSON('package.json').version;
```
anishashetty-okta: probably add a log line about the version being updated:
Something like:
`grunt.log.writeln('Updating version in README.md to ' + version);`

rchild-okta:
Mentioned in IM, but let's hold off on this until we update the CDN process - currently the CDN version won't be available at time of promote because it ships out with the production schedule.

jmelberg-okta:
Any reason we cannot do this now?

haishengwu-okta:
@jmelberg-okta i think we could. I'm wondering if we could have simplify a little bit. e.g instead of generating readme from a template, can we change version in place in current README?

aarongranick-okta:
Closing this PR as we can now update the README using a buildtool: https://github.com/okta/okta-signin-widget/pull/1697

PR #291, opened 2017-09-08
Fix padding on smaller screens when security image is hidden

santhoshbalakrishnan:
### Description

- The specificity of auth-content and auth-header classes were more because of the addition of #okta-sign-in.
- Because of that anything under .can-remove-beacon .auth-* did not affect anything as id selector was more specific.
- Adding #okta-sign-in to .can-remove-beacon to imporove specificity, so that the padding overloads apply on small screens.

### Reviewers

- @rchild-okta 
- @lboyette-okta 
- @mauriciocastillosilva-okta 

### Additional reviewers

- @upteam-okta 

### Screenshare

signin-security

### DIFF
--- a/assets/sass/_layout.scss
+++ b/assets/sass/_layout.scss
@@ -60,7 +60,9 @@
     padding-right: 20px;
   }
 
-  .can-remove-beacon & {
+  // scss-lint:disable IdSelector
+  #okta-sign-in.can-remove-beacon & {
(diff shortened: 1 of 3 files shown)


santhoshbalakrishnan:
@mauriciocastillosilva-okta Ya, I already mentioned it to @lboyette-okta . It's good that this type of SCSS qualifier is used only in this case.

PR #292, opened 2017-09-08
Support SAML ForceAuthn prompting

royalchan-okta:
https://oktainc.atlassian.net/browse/OKTA-140446
https://github.com/okta/okta-core/pull/22786

@mauriciocastillosilva-okta

### DIFF
--- a/src/models/PrimaryAuth.js
+++ b/src/models/PrimaryAuth.js
@@ -24,14 +24,17 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
 
     props: function () {
       var settingsUsername = this.settings && this.settings.get('username'),
+          rememberMeEnabled = this.settings && this.settings.get('features.rememberMe'),
           cookieUsername = CookieUtil.getCookieUsername(),
           remember = false,
           username;
       if (settingsUsername) {
         username = settingsUsername;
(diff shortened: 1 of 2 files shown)


src/models/Settings.js
```diff
@@ -49,6 +49,7 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
       'stateToken': ['string', false, undefined],
       'username' : ['string', false],
       'signOutLink': ['string', false],
+      'isSamlForceAuthn': ['boolean', true, false],
```
mauriciocastillosilva-okta: Do we need to add a new variable? I think we can get away with an existing one, ```features.rememberMe```. If  you pass that ```false``` the widget should hide the checkbox for remember me. 
The reason I am suggesting that is because it seems that this new variable is only used for the same purpose.

royalchan-okta: true, i can just check for the feature in PrimaryAuth.js

mauriciocastillosilva-okta:
@royalchan-okta you have some linting errors (run ```npm run lint``` locally to see what's wrong)

rchild-okta:
@mauriciocastillosilva-okta thanks for the heads up, makes sense (and pr looks good)

PR #293, opened 2017-09-08
:seedling: Supports AD-mastered user password expiry warning

gowthamidommety-okta:
image001


gowthamidommety-okta:
@mauriciocastillosilva-okta @mpeng-okta Please review

packages/@okta/i18n/package.json
```diff
@@ -1,30 +1,31 @@
...
+      "@okta/i18n@34.5.0",
       "/home/vagrant/okta/okta-signin-widget"
     ]
   ],
-  "_from": "@okta/i18n@34.3.0",
-  "_id": "@okta/i18n@34.3.0",
+  "_from": "@okta/i18n@34.5.0",
+  "_id": "@okta/i18n@34.5.0",
   "_inCache": true,
+  "_installable": true,
```
mauriciocastillosilva-okta: this should probably not be here, can you remove it?

packages/@okta/i18n/package.json
```diff
@@ -1,30 +1,31 @@
...
+    "raw": "@okta/i18n@34.5.0",
+    "rawSpec": "34.5.0",
     "scope": "@okta",
-    "spec": "34.3.0",
+    "spec": "34.5.0",
     "type": "version"
   },
   "_requiredBy": [
-    "/"
+    "#USER"
```
mauriciocastillosilva-okta: I have seen this in a few PRs, but if I remember correctly this is not needed, it should be ```/```

src/CustomPasswordExpiredController.js
```diff
@@ -35,6 +35,9 @@ function (Okta, FormController, FormType, Footer, Util) {
         else if (expiringSoon && numDays === 0) {
           return Okta.loc('password.expiring.today', 'login');
         }
+        else if (expiringSoon && (numDays === null || numDays === undefined)) {
```
mauriciocastillosilva-okta: Do we need to check if ```numDays``` is null or underfined? Wouldn't it be enough to just check if ```expiringSoon``` is true?  (same comment for the other controller)
```
else if (expiringSoon) {
```

mpeng-okta: Good point, we only need to check `expiringSoon`, I believe. I checked both expired and non-expired cases to make sure.

gowthamidommety-okta: 👍 Thanks Michael and Mau. I'll change that

PR #294, opened 2017-09-11
:bug: Move SCSS related to security beacon.

santhoshbalakrishnan:
## Description

- Moving the `.can-remove-beacon` to top level, so that it respects the #okta-sign-in closure.

**Original reason:**

- The specificity of auth-content and auth-header classes were more because of the addition of #okta-sign-in.
- Because of that anything under .can-remove-beacon .auth-* did not affect anything as id selector was more specific.
- Adding #okta-sign-in to .can-remove-beacon to imporove specificity, so that the padding overloads apply on small screens.

## Reviewers

- @mauriciocastillosilva-okta 
- @lboyette-okta 
- @rchild-okta 

## Screenshare

signin-security


assets/sass/modules/_okta-footer.scss
```diff
@@ -39,6 +39,23 @@
       }
     }
 
+    a {
```
mauriciocastillosilva-okta: is this for the widget footer? Where did these styles live before?

assets/sass/okta-sign-in.scss
```diff
@@ -32,10 +47,33 @@
   @import 'modules/footer';
   @import 'modules/enroll';
   @import 'modules/app-login-banner';
-  @import 'modules/okta-footer';
   @import 'modules/accessibility';
   @import 'modules/registration';
-  @import 'ie';
+
+  &.can-remove-beacon {
```
mauriciocastillosilva-okta: weren't these better in beacon.scss? Since they are only for the beacon.

assets/sass/okta-sign-in.scss
```diff
@@ -7,6 +7,21 @@
 @import 'common/enduser/responsive-variables';
 @import 'common/enduser/helpers';
 @import 'container';
+@import 'modules/okta-footer';
```
mauriciocastillosilva-okta: Not entirely sure, but should this one be scoped to the widget only, like it was before?

assets/sass/modules/_okta-footer.scss
santhoshbalakrishnan: This is for the <a> links in footer. This is needed here as the footer is outside the #okta-sign-in namespace.

Earlier, it was in _base.scss and it was not namespaced, which is why it worked even in footer as it applied globally to all <a>

assets/sass/okta-sign-in.scss
santhoshbalakrishnan: Unfortunately, the scss lint in okta-core throws an error if the top level selector starts with &, but sign in widget lets it through. This is an error in sass module itself and they suggest updating to 3.0 which I do not want to do at this point.

assets/sass/modules/_okta-footer.scss
mauriciocastillosilva-okta: got it. This is the widget's footer right? Curious why is it outside of the widget namespace?

assets/sass/okta-sign-in.scss
mauriciocastillosilva-okta: oh got it, that answers my question in a comment I just made. Thanks

mauriciocastillosilva-okta: mind filing a jira for later?

assets/sass/okta-sign-in.scss
mauriciocastillosilva-okta: nit: we could move it to beacon file and just add the whole selector, can't we? 
```
#okta-sign-in.can-remove-beacon {
```

santhoshbalakrishnan: well, we can't as beacon scss is imported inside #okta-sign-in and doing the above would make it 

```#okta-sign-in #okta-sign-in.can-remove-beacon {```

which would be wrong :(

assets/sass/okta-sign-in.scss
santhoshbalakrishnan: Done :) https://oktainc.atlassian.net/browse/OKTA-141201

assets/sass/okta-sign-in.scss
mauriciocastillosilva-okta: sad... got it. Thanks for the explanation

PR #296, opened 2017-09-12
WIP : Raas latest changes from oktane branch.

nikhilvenkatraman-okta:
Changes include 
- Merge code from oktane branch (alpha version of sign in widget) 
- Update courage version in sign in widget 
- While updating courage version , found few errors related to webpack.


rchild-okta:
There were some followup items I had discussed with @hor-kanchan-okta about how to clean this up before merging into mainline, is there a plan to do that now that we're past oktane?

nikhilvenkatraman-okta:
@rchild-okta Sorry just coming back to this post hackathon. This PR will be broken down into 2 PR's
- Update Courage version in sign in widget
- Raas related changes 
Currently I am resolving issues with the courage upgrade and I'll get back to Raas changes once the courage upgrade is done. I need to discuss with @hor-kanchan-okta regarding cleanup items.

FYI that there will be more PR's which bring additional functionality to Raas. The first Raas PR's intention is to get to a point (functionality wise) that we had during oktane.

rchild-okta:
I was under the impression that we didn't want to originally check it into master because the code was **not** in a state that we were comfortable with checking in - did something change since then?

nikhilvenkatraman-okta:
Yes, we have moved some generic logic into courage (which is why we need to upgrade courage), the API is more stable now. During oktane since the API was changing constantly we wanted a easy way to consume them via the alpha version of the widget.

PR #298, opened 2017-09-12
Update courage version in sign in widget.

nikhilvenkatraman-okta:
- Updating courage in sign in widget


nikhilvenkatraman-okta:
Need to update courage to a release version instead of a beta version.

PR #301, opened 2017-09-13
Complexity explanation text should include no first/last name

gowthamidommety-okta:
Resolves: OKTA-140359

screen shot 2017-09-26 at 8 20 43 pm

Please note that some tests may change depending on the backend API error message change.

### DIFF
--- a/src/util/FactorUtil.js
+++ b/src/util/FactorUtil.js
@@ -202,6 +202,13 @@ function (Okta, TimeUtil) {
       return Okta.loc(propertiesString, 'login', [displayableTime.time]);
     };
 
+    var setExcludeAttributes = function (policyComplexity) {
+      var excludeAttributes = policyComplexity.excludeAttributes;
+      policyComplexity.excludeFirstName = _.contains(excludeAttributes, 'firstName');
+      policyComplexity.excludeLastName = _.contains(excludeAttributes, 'lastName');
+      return _.omit(policyComplexity, 'excludeAttributes');
+    };
+
     if (policy.complexity) {
       var complexityFields = {
         minLength: {i18n: 'password.complexity.length', args: true},
@@ -209,10 +216,13 @@ function (Okta, TimeUtil) {
         minUpperCase: {i18n: 'password.complexity.uppercase'},
         minNumber: {i18n: 'password.complexity.number'},
         minSymbol: {i18n: 'password.complexity.symbol'},
-        excludeUsername: {i18n: 'password.complexity.no_username'}
+        excludeUsername: {i18n: 'password.complexity.no_username'},
+        excludeFirstName: {i18n: 'password.complexity.no_first_name'},
+        excludeLastName: {i18n: 'password.complexity.no_last_name'}
       };
 
-      var policyComplexity = policy.complexity;
+      var policyComplexity = setExcludeAttributes(policy.complexity);
+
       var requirements = _.map(policyComplexity, function (complexityValue, complexityType) {
         if (complexityValue <= 0) { // to skip 0 and -1
           return;
@@ -230,7 +240,7 @@ function (Okta, TimeUtil) {
           return result ? (result + Okta.loc('password.complexity.list.element', 'login', [requirement])) : requirement;
         });
 
-        result.push(Okta.loc('password.complexity.description', 'login', [requirements]));
+        result.push(Okta.loc('password.complexity.requirements', 'login', [requirements]));
       }
     }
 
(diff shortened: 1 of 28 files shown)


src/util/FactorUtil.js
```diff
@@ -202,17 +202,28 @@ function (Okta, TimeUtil) {
       return Okta.loc(propertiesString, 'login', [displayableTime.time]);
     };
 
-    if (policy.complexity) {
+    var setExcludeAttributes = function (policyComplexity) {
+      var excludeAttributes = policyComplexity.excludeAttributes;
```
rchild-okta: We're slightly changing the behavior here because this code assumes policy.complexity is set - is that always the case?

gowthamidommety-okta: That's a good point. I asked the backend dev and it might be that it is. But because it is a public API it is probably better to revert this

mauriciocastillosilva-okta: yeah, all of this logic should be inside the ```if (policy.complexity)``` below, and maybe just have a helper method to find if an specific attribute is excluded...

test/unit/spec/PasswordReset_spec.js
```diff
@@ -32,13 +32,15 @@ function (Q, _, $, OktaAuth, LoginUtil, SharedUtil, Util, PasswordResetForm, Bea
...
+    var excludeAttributes = getExcludeAttributes(settings.excludeAttributes);
     var policyComplexityDefaults = {
       minLength: 8,
       minLowerCase: 1,
       minUpperCase: 1,
       minNumber: 1,
       minSymbol: 1,
-      excludeUsername: true
+      excludeUsername: true,
+      excludeAttributes: excludeAttributes
```
mauriciocastillosilva-okta: nit: ``` excludeAttributes: getExcludeAttributes(settings.excludeAttributes)```

src/util/FactorUtil.js
gowthamidommety-okta: @mauriciocastillosilva-okta  Do my most recent changes make sense? I looked at it for awhile, but decided to leave setExcludeAttributes because 1. it was clear and 2. it would ignore any other attributes that may be in that array. Please let me know if you see some way for improvement.

mauriciocastillosilva-okta: There is only one thing then: in this method we are adding an```excludeFirstName``` and ```excludeLastName``` to the original ```policy``` object, can we make sure then that does not affect anything else?
As a nit, in line 227 we have ```if (complexityValue <= 0) { // to skip 0 and -1``` which works with true/false, but it is a little odd...

test/unit/helpers/xhr/PASSWORD_EXPIRED_error_complexity.js
```diff
@@ -7,7 +7,7 @@ define({
     "errorLink": "E0000080",
```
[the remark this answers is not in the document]
gowthamidommety-okta: @mpeng-okta Can you please review this API response to see if it is in sync with the backend?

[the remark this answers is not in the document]
mpeng-okta: Yup this error message is in sync with the backend!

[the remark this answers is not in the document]
mauriciocastillosilva-okta: I thought we changed the error summary?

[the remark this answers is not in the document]
gowthamidommety-okta: Yea just checked. The outer errorSummary says: "Update of credentials failed"
The one in errorCauses is correct. I'll make changes

test/unit/spec/PasswordReset_spec.js
```diff
@@ -170,37 +175,63 @@ function (Q, _, $, OktaAuth, LoginUtil, SharedUtil, Util, PasswordResetForm, Bea
...
     });
 
     itp('has a valid subtitle if only password complexity "excludeUsername" defined', function () {
       return setup({policyComplexity: 'excludeUsername'}).then(function (test) {
-        expect(test.form.subtitleText()).toEqual('Your password must have no parts of your username.');
+        expect(test.form.subtitleText()).toEqual('Password requirements: no parts of your username.');
+      });
+    });
+
+    itp('has a valid subtitle when excludeAttributes contains firstName and lastName', function () {
```
mauriciocastillosilva-okta: nit: for this tests, it should read something like ```has a valid subtitle if only excludeAttributes['firstName','lastName'] is defined```

PR #302, opened 2017-09-18
:arrow_up: Update courage version in sign in widget.

nikhilvenkatraman-okta:
- Updated courage version to 3.2.0
- Fixed webpack config file to add aliases for certain components that were being used and removed them from the empty list.
- Fixed i18n bug to point courage bundle to login inside sign in widget as we do not have courage.properties in sign in widget. 
- Fixed tests that were checking for error thrown when sign in widget was bootstrapped with wrong config parameters. This is now changed to throw a log message in courage. 
https://github.com/okta/courage/commit/eb4b5baa59e057846fb8784c37bd21be6426dafa

@mauriciocastillosilva-okta @rchild-okta @samerfanek-okta 

@upteam-okta


src/util/Bundles.js
```diff
@@ -189,6 +189,10 @@ define([
   return {
     login: login,
     country: country,
+    // Courage components within the sign in widget point to courage bundle to look
+    // up i18nkeys. Since we dont have courage.properties inside the sign in widget
+    // we are pointing courage bundle to login.
+    courage: login,
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: Yes , this is a recent change in courage https://github.com/okta/courage/commit/12e2f0cd5273f3a5ac6c906416e16e42cfd74f85#diff-919b0b5e3639e0963eb5c6c3616bbbb0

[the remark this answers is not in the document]
rchild-okta: Ah, I see. This makes sense then - @mauriciocastillosilva-okta, we should probably think about this case in the future. Now that we're not relying on requireJs to handle the localization lookup for us in any package that has translations, might make sense to abstract the bundle lookup to the thing that handles it (right now it's bundles.js).

webpack.common.config.js
```diff
@@ -23,9 +24,13 @@ module.exports = function (outputFilename) {
         'okta/underscore': SHARED_JS + '/util/underscore-wrapper',
         'okta/handlebars': 'handlebars/dist/handlebars',
         'okta/moment': 'moment/moment',
+        'okta/moment-tz': VENDOR + '/lib/moment-timezone-with-data-2010-2020',
         'okta/jqueryui': EMPTY,
         'okta': 'shared/util/Okta',
         'shared/util/Bundles': 'util/Bundles',
+        'mixpanel-browser': SHARED_JS + '/util/mixpanel-factory',
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: We don't need mixpanel for RAAS. In the sign in widget mixpanel was listed in the EMPTY section as we are not using it. I had to add this because an internal courage component was using it and  would thrown an error when it tried to find it.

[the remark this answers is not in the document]
rchild-okta: Which internal component depends on mixpanel? Is it something that's used by the widget?

src/util/Bundles.js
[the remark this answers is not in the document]
mauriciocastillosilva-okta: let's talk about this, but yeah, I didn't think about this case where we don't have courage bundle included.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -164,13 +165,21 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, OktaAuth, Util,
...
+    function expectUnexpectedFieldLog(arg1) {
+      expect(Logger.warn).toHaveBeenCalledWith('Field not defined in schema', arg1);
+    }
+
     it('throws a ConfigError if unknown option is passed as a widget param', function () {
+      spyOn(Logger, 'warn');
       var fn = function () { setup({ foo: 'bla' }); };
-      expect(fn).toThrowError(Errors.ConfigError);
+      expect(fn).not.toThrow(Errors.ConfigError);
+      expectUnexpectedFieldLog('foo');
```
mauriciocastillosilva-okta: I'm not sure we should be changing this tests... the test seems to be here to verify that we throw an error when an invalid config param is passed. Changing this now is not throwing that error, but instead it is just logging a warn message, which I'm guessing, is not an exception anymore, but just a warning.

webpack.common.config.js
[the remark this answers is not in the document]
mauriciocastillosilva-okta: yeah, previously courage required mixpanel, and we just pointed it to EMPTY. Can we do the same here? I'm curious to about which component needs it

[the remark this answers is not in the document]
nikhilvenkatraman-okta: packages/@okta/courage/src/util/Metrics.js in courage is using mixpanel which came with the upgrade. Don't think its used by the widget. Let me try removing mixpanel-browser, selectize, vendor/plugins/jquery.autosuggest-1.4.min to see if I still see the error.

test/unit/spec/LoginRouter_spec.js
nikhilvenkatraman-okta: @mauriciocastillosilva-okta I made these test changes based on this commit to courage https://github.com/okta/courage/commit/eb4b5baa59e057846fb8784c37bd21be6426dafa. I saw similar test fix in Model_spec.js.

webpack.common.config.js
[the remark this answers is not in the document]
nikhilvenkatraman-okta: Took me a while, but I got this error
 ERROR in ./target/js/shared/util/Metrics.js
Module not found: Error: Cannot resolve module 'mixpanel-browser' in /home/vagrant/okta/my-signin-widget/okta-signin-widget/target/js/shared/util
 @ ./target/js/shared/util/Metrics.js 4:0-246:2

Steps: 
1: deleted node_modules
2: npm install
3: npm start

packages/@okta/courage/src/vendor/plugins/jquery.autosuggest-1.4.min.js
```diff
@@ -336,7 +336,8 @@ define(['jquery'], function ($) {
 								formatted = opts.formatList.call(this, this_data, formatted);	
 							}
 							results_ul.append(formatted);
-							delete this_data;
+							//delete this_data;
```
nikhilvenkatraman-okta: This was needed to get rid of this error message. 
ERROR in ./target/js/vendor/plugins/jquery.autosuggest-1.4.min.js
Module build failed: SyntaxError: Deleting local variable in strict mode (341:7)

  339 | 							//delete this_data;
  340 | 							//this_data = null;
> 341 | 							delete this_data;
      | 							^
  342 | 							matchCount++;
  343 | 							if(opts.retrieveLimit && opts.retrieveLimit == matchCount ){ break; }
  344 | 						}

 @ ./target/js/shared/views/forms/inputs/BasePicker.js 3:0-466:2

rchild-okta: Is the change just in this file or is it in courage as well? Should probably address in courage (either by updating it to this code (with a comment why) or disable the linting for that line).

Or, actually, we should probably just ignore linting on vendor files.

webpack.common.config.js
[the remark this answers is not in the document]
rchild-okta: Okay - we probably shouldn't be using Metrics in the widget, can you just point that to empty and see if that resolves this?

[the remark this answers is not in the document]
rchild-okta: Btw, in general, if new components/etc are added to courage but they're not actively being used in the widget, we should add them to the blacklist here.

packages/@okta/courage/src/vendor/plugins/jquery.autosuggest-1.4.min.js
nikhilvenkatraman-okta: I agree that this should be done in courage. Not sure why this error is not thrown in courage.

rchild-okta: It's being ignored:
https://github.com/okta/courage/blob/master/.eslintignore#L1

test/unit/spec/LoginRouter_spec.js
mauriciocastillosilva-okta: right, but is this changing the behavior of the widget? If we pass an invalid config param, what happens now? Do we throw an error so the user knows, or do we just log an error?

webpack.common.config.js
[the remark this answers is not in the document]
nikhilvenkatraman-okta: I had to add `mixpanel-browser` to the EMPTY list instead of `mixpanel`. That seems to get rid of the error. Also moved selectize to EMPTY.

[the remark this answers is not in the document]
rchild-okta: I'd still say that we should axe out dependencies as high up the chain as possible. If we mark metrics as empty, we don't need to mark mixpanel-browser as empty (or any other independent dependencies that it would pull in).

packages/@okta/courage/src/vendor/plugins/jquery.autosuggest-1.4.min.js
nikhilvenkatraman-okta: Should we do the same in sign in widget by adding vendor files to the eslintignore list ?

rchild-okta: Yeah that sounds fine.

webpack.common.config.js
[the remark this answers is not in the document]
nikhilvenkatraman-okta: Looks like Metrics is already marked as empty https://github.com/okta/okta-signin-widget/pull/302/files/e8f77e3cf3c6128cd1f976b62e2abbbecdb7fd0f..4d0063d56cb70dd3a374e9434dc500fec0b0cb08#diff-aa4050a929bf9b482ff53091a6327b9fR63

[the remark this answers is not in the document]
nikhilvenkatraman-okta: Tried moving Metrics and metrics-factory to EMPTY, but it still throws the error. Setting mixpanel-browser to EMPTY fixes the error.

test/unit/spec/LoginRouter_spec.js
nikhilvenkatraman-okta: @rajaguruduraisamy-okta  ^^

rajaguruduraisamy-okta: I believe having extra properties in the courage model will not do any harm. There is no need for this check anymore. We need to check for required and needed params passed alone and any extra unwanted params passed will be discarded right. @rchild-okta ^^ your thoughts.

rchild-okta: This one is probably fine - we originally made the change in courage to be less strict on API changes, i.e. if the API response returns more properties than are explicitly set in the middle, it wouldn't break the UI. This case is a little different - we have been using this as a hard check to let the user know immediately that they're doing something wrong. But, it will log the error to the console (which is where the original error shows up anyway), so I think that should be sufficient.

mauriciocastillosilva-okta: Oh ok, if it is being output to the user then it is good. I thought for some reason that since we don't throw the exception the error will not be immediately visible to the user.

src/util/Bundles.js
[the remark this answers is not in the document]
mauriciocastillosilva-okta: I meant @rchild-okta and we should talk about future enhancements for i18n
The change here looks good to me

webpack.common.config.js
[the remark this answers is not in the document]
rchild-okta: Why is this the case? I.e. if we are not loading metrics, it shouldn't try to load any metrics dependencies. So, is it:

1. Metrics is not being ignored even if we are swapping it out with empty 
2. Another file is requiring metrics

webpack.common.config.js
```diff
@@ -23,9 +24,13 @@ module.exports = function (outputFilename) {
         'okta/underscore': SHARED_JS + '/util/underscore-wrapper',
         'okta/handlebars': 'handlebars/dist/handlebars',
         'okta/moment': 'moment/moment',
+        'okta/moment-tz': VENDOR + '/lib/moment-timezone-with-data-2010-2020',
```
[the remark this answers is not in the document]
rchild-okta: By the way I still have this question - after running the build you should be able to see the size (it's part of the webpack output).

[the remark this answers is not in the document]
nikhilvenkatraman-okta: Sorry, I missed replying to this comment 
okta-sign-in.entry.js  2.34 MB now
okta-sign-in.entry.js  2.25 MB  before

okta-sign-in.min.js  1.12 MB  now 
okta-sign-in.min.js  1.07 MB  before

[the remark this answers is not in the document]
rchild-okta: Cool, not as bad a bump as I thought. Todo for later to make this smaller, since it's a lot larger than when we last pruned.

webpack.common.config.js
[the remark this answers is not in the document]
nikhilvenkatraman-okta: @rchild-okta this is now addressed with the latest commit. I had to move mixpanel to the null loaders section

packages/@okta/courage/src/vendor/plugins/jquery.autosuggest-1.4.min.js
nikhilvenkatraman-okta: Tried adding a eslintignore file. That doesn't seem to fix the problem. Seems like a webpack config issue.

nikhilvenkatraman-okta: Reverting this commit https://github.com/okta/okta-signin-widget/commit/af51f2c73279139c430e4e6c8a4071f66cbe8748 got rid of the error. Might have to ignore this file via Balel's ignore list

webpack.common.config.js
```diff
@@ -104,7 +104,10 @@ module.exports = function (outputFilename) {
         // Babel
         {
           test: /\.js$/,
-          exclude: /node_modules/,
+          exclude: [
+            path.resolve(__dirname, "node_modules"),
```
rchild-okta: nit: singlequotes

Double checking - are we actually using the jquery autosuggest widget anywhere? If we do it this way, it means that we are including the file, but not running it through babel.

nikhilvenkatraman-okta: Its used in views/forms/inputs/BasePicker in courage. Looks like we have a similar exclude list in courage webpack config.
```javascript
      loaders: [
        {
          test: /\.js$/,
          exclude: /(node_modules|vendor|okta\.js)/,
          loader: 'babel-loader',
          query: {
            presets: ['es2015'],
            cacheDirectory: '/tmp/.babel-cache'
          }
        }
      ]
```
@manueltanzi-okta made a similar fix for vkbeautify by setting it to empty `vendor/plugins/vkbeautify.0.99.00.beta': EMPTY`

The same fix seems to work for autosuggest i.e adding ti EMPTy list `'vendor/plugins/vkbeautify.0.99.00.beta': EMPTY`.

rchild-okta: Yeah, so that's the main question - is there any view that is using BasePicker with autosuggest? If there is nothing that uses it, we can go ahead with the empty fix (although once again, we should do it as high up as possible i.e. return empty for BasePicker rather than autosuggest). If there is a view that uses it, then we should go with the general fix, i.e. just ignore the vendor directory rather than calling it out explicitly.

nikhilvenkatraman-okta: I did try moving basePicker to EMPTY list, although the build passes but i see an error when i try to load the widget on the browser. 

```javascript
Uncaught TypeError: BasePicker.extend is not a function
    at Object.<anonymous> (BaseSelectize.js:30)
    at Object.<anonymous> (BaseSelectize.js:1)
    at __webpack_require__ (bootstrap c9e73492f7024617fa23:19)
    at Object.<anonymous> (formsBundle.js:37)
    at __webpack_require__ (bootstrap c9e73492f7024617fa23:19)
    at Object.<anonymous> (viewsBundle.js:8)
    at __webpack_require__ (bootstrap c9e73492f7024617fa23:19)
    at Object.<anonymous> (Okta.js:16)
    at __webpack_require__ (bootstrap c9e73492f7024617fa23:19)
    at Object.<anonymous> (Util.js:14)
```
The error is coming from formsBundle view which is using BaseSelectize which extents BasePicker.

nikhilvenkatraman-okta:
@rchild-okta @mauriciocastillosilva-okta squashed all commits back to 2 again. One for courage update, and another for the additional changes that were needed.

packages/@okta/courage/src/vendor/lib/backbone.js
```diff
@@ -0,0 +1,1873 @@
+//     Backbone.js 1.2.1
```
mauriciocastillosilva-okta: We didn't have backbone before... do we need it or can we get it from node_modules locally, similar as other vendor files?

nikhilvenkatraman-okta: @mauriciocastillosilva-okta I think we do need to list them as dependencies in package.json file but if we delete it from packages folder we should create aliases for them in the webpack.config file like we have for `handlebars: 'handlebars/dist/handlebars` and point the alias to the dist folder. Let me know if that makes sense and I can update the PR.

nikhilvenkatraman-okta: We do use handlebars and underscore so they need to be listed as dependencies in package.json and we already have mappings for them in the webpack.config so i can delete the files  from packages folder. We don't need backbone and clipboard, so i can delete them as well.

rchild-okta: Am curious - is there any reason to delete them? I.e. if we're already aliased to the node_modules we're pulling down, it's not that big of a deal that there's a copy in the packages/vendor folder. 

Mostly thinking that if we can simplify the process to upgrade courage (less manual steps), the better.

mauriciocastillosilva-okta: not any reason, it was my understanding that we didn't need duplicates since we could get them from node_modules. 
But you are right, this will make it easier to update courage in the future.

nikhilvenkatraman-okta: Mostly because it was not there in packages folder before. The way I have it now matches the dependency list we have in courage.

rchild-okta: Great, let's do that then. Btw, is there a doc somewhere outlining the steps we need to update courage @mauriciocastillosilva-okta or @nikhilvenkatraman-okta? If not, we should add this info somewhere.

rchild-okta: Double checking @nikhilvenkatraman-okta - you didn't manually add this to the vendor folder, did you?

nikhilvenkatraman-okta: @rchild-okta I don't know if there is a doc for this. I'll try and write up a wiki doc.

rchild-okta: Okay, it seems like we did - as an aside @nikhilvenkatraman-okta, it would've been better to break out the commits with "things that happen automatically by copying courage over" and "everything else". This one's an easy one to miss - nice catch @mauriciocastillosilva-okta.

We don't need to manually copy dependencies back into vendor (in fact, we're moving away from any checked in source files). That's one of the highlights of webpack - it can consume modules directly from the node_modules folder. We've only got aliases for things like handlebars because we had to make an exception - in that case, the main file that's exported by handlebars isn't the one we want to use.

nikhilvenkatraman-okta: @rchild-okta  The files in packages folder were copied over from node_modules after running the update command

nikhilvenkatraman-okta: @rchild-okta In that case I can delete these `4 new` files copied into packages folder and leave the dependencies in package.json ?

rchild-okta: I don't see backbone.js in the [vendor/lib](https://github.com/okta/courage/tree/master/src/vendor/lib) directory - how did it get there in this PR? Is it part of some build step?

nikhilvenkatraman-okta: Its added to the dependencies list here https://github.com/okta/courage/blob/master/package.json#L100

rchild-okta: Talked over IM, it's getting copied there as part of the build ([here](https://github.com/okta/courage/blob/master/Gruntfile.js#L84)]. In this case, I think it's fine to just include it - the "update courage" step should just be to copy over everything in the node_modules dir, w/out any extra work to delete things we don't need.

nikhilvenkatraman-okta: @rchild-okta  @mauriciocastillosilva-okta Let me know if you have any other comments, if not i can merge.

PR #303, opened 2017-09-20
:bug: set the correct default factor for recovery flows

mauriciocastillosilva-okta:
- Forgot Password and Unlock Account flows were defaulting to email for the recovery factor. If a user press 'Enter' in the for, we were sending a request with email as a factor, even if email was not on the available factors (features.emailRecovery=false).
- The fix is to default to the factor that is first on the list

Resolves: OKTA-141184

### DIFF
--- a/src/ForgotPasswordController.js
+++ b/src/ForgotPasswordController.js
@@ -64,7 +64,7 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, ContactSupport,
     Model: {
       props: {
         username: ['string', true],
-        factorType: ['string', true, Enums.RECOVERY_FACTOR_TYPE_EMAIL]
+        factorType: ['string', true]
       },
       validate: function () {
         return ValidationUtil.validateUsername(this);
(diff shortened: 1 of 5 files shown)


src/ForgotPasswordController.js
```diff
@@ -134,16 +134,19 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, ContactSupport,
...
             this.$el.addClass('forgot-password-call-enabled');
             formChildren.push(this.createRecoveryFactorButton('call-button', 'password.forgot.call',
               Enums.RECOVERY_FACTOR_TYPE_CALL, this));
+            this.setDefaultFactorType(Enums.RECOVERY_FACTOR_TYPE_CALL);
           }
           if (emailEnabled) {
             this.$el.addClass('forgot-password-email-enabled');
             formChildren.push(this.createRecoveryFactorButton('email-button', 'password.forgot.sendEmail',
               Enums.RECOVERY_FACTOR_TYPE_EMAIL, this));
+            this.setDefaultFactorType(Enums.RECOVERY_FACTOR_TYPE_EMAIL);
```
rchild-okta: Double checking - is this the right order of preference (SMS, then call, then email)? 

Edit: Looking at the test cases, this is indeed the intended behavior. Good!

PR #306, opened 2017-09-25
:Bug:Fix signin-widget label customization for locale codes like 'xx-XX'

eugenebakaiev-okta:
Added adjustLocalizationKey method which converts language code to a style which is used in overrides array (eg 'nl-nl', 'nl-NL')

Resolves: OKTA-138149


eugenebakaiev-okta:
@hor-kanchan-okta, @upteam-okta please review this PR

src/util/Bundles.js
```diff
@@ -196,18 +196,49 @@ define([
...
+            return keyPart;
+          }).join('-');
+        }
+      });
+
+      return adjustedLocalizationKey;
+    },
+
     loadLanguage: function (language, overrides, assets) {
       var parsedOverrides = parseOverrides(overrides);
```
hor-kanchan-okta: The other solution is to modify `parseOverrides`. Always return the key in lowercase. The comparison will be case insensitive. Will let @samerfanek-okta to decide

samerfanek-okta: Yeah I think it makes more sense to put it in ``parseOverrides`` and return a lowercase key. Also, can you add unit tests for this covering the various combinations? (For example, 'nl', 'NL', 'nl-nl', 'nl-NL', etc..)

eugenebakaiev-okta:
@hor-kanchan-okta @samerfanek-okta updated PR please look

PR #309, opened 2017-09-28
:Bug: Fix widget height

eugenebakaiev-okta:
-Removed min-height: 100% from #okta-sign-in, this came after removing css collisions. Before this rule was applied to body element.

Resolves: OKTA-142547


eugenebakaiev-okta:
@magizh-okta please look

santhoshbalakrishnan:
CC: @lboyette-okta @robertdamphousse-okta

santhoshbalakrishnan:
@magizh-okta Dont think so. It's not consumed by anyone now. Okta-core has CSS overrides that solve this issue.

PR #315, opened 2017-10-05
:Bug: Fix password validation when no username is provided.

nikhilvenkatraman-okta:
- Fixed password complexity validation when username  field is empty. 
- Added tests for passwordContainsUserName function to check for null values. 

Resolves: OKTA-143940

@hor-kanchan-okta 

@upteam-okta

### DIFF
--- a/test/unit/spec/RegistrationFormFactory_spec.js
+++ b/test/unit/spec/RegistrationFormFactory_spec.js
@@ -130,5 +130,17 @@ function (SchemaProperty, RegistrationFormFactory) {
 
     });
 
+    describe('PasswordContainsUserName', function () {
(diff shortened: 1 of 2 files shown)


test/unit/spec/RegistrationFormFactory_spec.js
```diff
@@ -130,5 +130,17 @@ function (SchemaProperty, RegistrationFormFactory) {
 
     });
 
+    describe('PasswordContainsUserName', function () {
```
nikhilvenkatraman-okta: There are several more tests in registration_spec that test the different combinations of password and username. This test is mainly for null usecase.

PR #323, opened 2017-10-19
Add registration callback hooks.

nikhilvenkatraman-okta:
:seedling: Added preRender, preSubmit, postSubmit callback hooks for registration. 

- preRender hook is used to update the schema sent from the server 
- preSubmit hook is used to update the data that is posted to the server
- postSubmit hook is used to get control post registration post to the server

https://oktawiki.atlassian.net/wiki/spaces/~nikhil.venkatraman/pages/236241695/Sign+in+widget+callback+hooks

Example Usage: 
```javascript
 options['registration'] = {
          'preRender': function(resp, onSuccess, onFailure) {
            setTimeout(function () {
              resp.profileSchema.properties.address = {
                'type': 'string',
                'description': 'Street Address',
                'default': 'Enter your street address',
                'maxLength': 255
              };
              var error = true;
              if (error) {
                var errorObject = {
                  "errorSummary": "Prerender Error: Failed to add address field",
                  "errorCauses": [
                  {
                  "errorSummary": "Prerender Error: Failed to add address field",
                  "reason": "registration.error.address",
                  "resource": "User",
                  "property": "address",
                  "arguments": []
                }]
              }
                onFailure(errorObject);
              } else {
                resp.profileSchema.fieldOrder.push('address');
                onSuccess(resp);
              }
            }, 2000);
          },
          'preSubmit': function (postData, onSuccess, onFailure) {
            setTimeout(function () {
              var error = false;
              if (error) {
                var errorObject = {
                  "errorSummary": "PreSubmit Error: Failed to remove address field"
                }
                onFailure(errorObject);
              }else{
                delete postData.address;
                onSuccess(postData);
              }
            }, 2000);
          },
          'postSubmit': function (response, onSuccess, onFailure) {
              setTimeout(function () {
                var error = false;
                if (error) {
                var errorObject = {
                  "errorSummary": "PostSubmit Error: Failed to submit"
                }
                onFailure();
              } else {
                onSuccess(response);
              }
              }, 2000);
          }

}

```

@hor-kanchan-okta @rchild-okta 
@lboyette-okta @robertdamphousse-okta @richardheng-okta


rchild-okta:
Will take a look more thoroughly later today, but a couple initial questions:
- Why do we need two different ways of doing async (promises/callback)? Unfortunately, we have both patterns in various places in the widget, so we should just choose one. 
- Haven't looked yet, but how do we use the onFailure function? Do they need to format an error message like we'd expect and we'd display it in the UI?
- What's the normal flow for registration after it's submitted? Is it just to pass control to the user in the postSubmit action, or is there some default behavior that we do if no postSubmit function is registered (i.e. call the global success fn)?

nikhilvenkatraman-okta:
- promises/callback :
This is just to support both async and sync use cases. Technically the callback is async but here we use that as the sync option. 
If you were to call another API in the preRender/preSubmit function that comes back after a while you can return promise.

-  onFailure function:
The onFailure right now will just call a globalConfigError letting user know the preRender/preSubmit callback failed. There is a message but its not coming in from the callback. However, we could take a message as an argument form the callback and use that when we call globalConfigError.

-  postSubmit action:
 If there is no postSubmit callback defined it tries to use activation token to perform authClient.signIn or navigates to activation email sent screen.
Only If there is a postSubmit callback defined we call a globalSuccess function   

```javascript

    // registration via activationToken
     this.settings.callGlobalSuccess(Enums.REGISTRATION_COMPLETE, {
       activationToken: activationToken
     });

    //  registration via activation email.
     this.settings.callGlobalSuccess(Enums.ACTIVATION_EMAIL_SENT, {
       username: userName
     });

```
We can call the globalSuccess function even when there is no postSubmit callback if needed.

rchild-okta:
Haven't had time yet to look at the code, but responding to your reply:

> This is just to support both async and sync use cases. Technically the callback is async but here we use that as the sync option.

That's a bit confusing - I wouldn't think of using callbacks as "sync". Is there a reason we need a sync option? Alternatively, since we mostly don't use promises, why not just have the one callback option and call it a day?

> The onFailure right now will just call a globalConfigError letting user know the preRender/preSubmit callback failed

If it's the global, does that mean that it's just a thrown exception? Should we instead be thinking about how to hook their response into our existing error framework?

> postSubmit action

Is there a reason we're not making this a hook as well? I.e. have the option to call a callback to do the default logic. In this case, if they didn't want to do the default action, they could choose not to call the callback.

nikhilvenkatraman-okta:
Changes include in 2nd commit
# Use callbacks support only instead of promises
# postSubmit is also a hook now 
# Use onFailure callback that accepts an error Object (follows the same format as our API errors). Can be used to show form level vs field level errors.
# use globalError to throw RegistrationError
# Move globalSuccess call to registrationComplete controller
# Added tests

src/RegistrationController.js
```diff
@@ -58,19 +62,65 @@ function (
...
-      });
-      
-      var schema = new Schema();
+      var doPostSubmit = _.bind(function () {
+        if (this.model.get('activationToken')) {
+          var authClient = this.model.appState.settings.authClient;
+          authClient.signIn({
+            token: this.model.get('activationToken')
+          })
+          .then(_.bind(function(transaction) {
```
rchild-okta: Why do we need to manually call routeAfterAuthStatusChange? Shouldn't it automatically just hook into the existing widget logic?

src/RegistrationController.js
```diff
@@ -58,19 +62,65 @@ function (
...
     initialize: function() {
       var clientId = this.options.settings.get('clientId');
       var registrationApi = this.options.settings.get('baseUrl')+'/api/v1/registration/'+clientId;
 
-      var Schema = RegistrationSchema.extend({
-        url: registrationApi+'/form'
-      });
-      
-      var schema = new Schema();
+      var doPostSubmit = _.bind(function () {
```
rchild-okta: In general, when using private helper functions, pull them out into the outer scope and pass in the context as a variable, i.e.

```javascript
function doPostSubmit(controller) {
  if (controller.model.get('activationToken')) {
    // etc
  }
}
```

src/RegistrationController.js
```diff
@@ -58,19 +62,65 @@ function (
     save: Okta.loc('registration.form.submit', 'login')
   });
 
+  var DEFAULT_CALLBACK_ERROR = 'We could not process your registration at this time. Please try again later';
```
nikhilvenkatraman-okta: https://github.com/okta/npm/pull/923

src/RegistrationController.js
```diff
@@ -87,53 +137,80 @@ function (
             this.set('activationToken', resp.activationToken);
             delete resp.activationToken;
             return resp;
+          },
+          save: function() {
+            if(_.isFunction(preSubmit)) {
+              //async callback
+              preSubmit(this.attributes, function() {
```
rchild-okta: Let's try to come up with a better way to write this logic. As is, it's pretty tangled up:

- We have a `presubmit` function that they may or may not have defined, so we've two paths
  - Already here, we can simplify by following the pattern we use elsewhere where user input is normalized up-front in the settings.js file (i.e. transformUsername). We shouldn't have to check if they've defined something way down here in the registration controller.
- That calls `postFormSubmit`
- That calls the model save, which then calls `doFormSubmit` 
- That has some logic to once again parse the config input for `postSubmit` (like above, we should normalize this before it even hits this function in settings.js), which then calls `doPostSubmit`
- And then that does some stuff

And each of these places are handling errors, etc. Take a first pass, and if you want some help let's talk about it.

src/RegistrationController.js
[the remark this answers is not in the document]
nikhilvenkatraman-okta: This should be addressed with https://github.com/okta/npm/pull/923

src/RegistrationCompleteController.js
```diff
@@ -32,7 +32,11 @@ function (Okta, Enums, FormController, FormType) {
     className: 'registration-complete',
     Model: function () {
     },
-
+    initialize: function() {
+      this.settings.callGlobalSuccess(Enums.ACTIVATION_EMAIL_SENT, {
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: I did see GlobalSuccess being called inside initialize in a few other places. Will check email-sent flows to see how they handle it.

src/RegistrationController.js
nikhilvenkatraman-okta: Agreed. Will try to move the logic inside the callbacks into settings.

src/RegistrationController.js
```diff
@@ -87,53 +137,80 @@ function (
...
+              form.add(SubSchema.extend({id: 'subschemas-' + name, subSchemas: subSchemas}));
+            }
+          });
+        }
       }, this);
 
-      schema.fetch().then(function(){
-        postSchemaFetch();
+      var Schema = RegistrationSchema.extend({
+        preRender: this.options.settings.get('registration.preRender'),
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: Unfortunately we haven't had that much time to do a lot of re-factor post-oktane. We spent some time cleaning up the API/ integration along with merging the UI changes for oktane branch. Will try to re-factor based on your suggestions here.

cc: @hor-kanchan-okta

src/RegistrationController.js
nikhilvenkatraman-okta: Not a 100% sure about that, need to check . This was existing code that just got moved around.

src/RegistrationController.js
```diff
@@ -61,86 +65,143 @@ function (
...
-          var inputOptions = RegistrationFormFactory.createInputOptions(schemaProperty);
-          var subSchemas = schemaProperty.get('subSchemas');
-          var name = schemaProperty.get('name');
-          form.addInput(inputOptions);
-          if (name === 'password' && subSchemas) {
-            form.add(SubSchema.extend({id: 'subschemas-' + name, subSchemas: subSchemas}));
-          }
-        });
+      // register parse complete event listener
+      schema.on('parseComplete', function(updatedSchema) {
```
rchild-okta: Looking much better with the fetchInitialData, but can we add this part in there as well? That way, we're getting all the async stuff done up front and won't have to deal with it in a callback here.

src/RegistrationController.js
```diff
@@ -61,86 +65,143 @@ function (
...
+      var clientId = this.options.settings.get('clientId');
+      return this.options.settings.get('baseUrl')+'/api/v1/registration/'+clientId;
+    },
+    doPostSubmit: function () {
+      if (this.model.get('activationToken')) {
+        // register via activation token
+        this.settings.callGlobalSuccess(Enums.REGISTRATION_COMPLETE, {
+          activationToken: this.model.get('activationToken')
+        });
+        var authClient = this.model.appState.settings.authClient;
```
rchild-okta: Can you look into why we need to do it this way? I.e. instead of doing it manually here, let's try to use what we do elsewhere with the transaction functions (i.e. startTransaction).

src/RegistrationController.js
nikhilvenkatraman-okta: I had the parseComplete listener inside fetchInitialData, but as discussed on IM had some trouble wiring up the tests. I'll give it another shot after fixing the broken tests.

src/RegistrationController.js
nikhilvenkatraman-okta: Will look into this after I am done with the refactor to move parseComplete listener into fetchInitialData.

src/RegistrationController.js
nikhilvenkatraman-okta: @rchild-okta this is now addressed with my latest commit.

src/RegistrationController.js
```diff
@@ -61,86 +65,141 @@ function (
...
-      schema.fetch().then(function(){
-        postSchemaFetch();
+      if (hideRegisterButton) {
+        this.$el.find('.button-primary').hide();
+      }
+    },
+    fetchInitialData: function () {
+      var self = this;
+      // register parse complete event listener
+      self.state.get('schema').on('parseComplete', function(updatedSchema) {
```
rchild-okta: I think this is probably okay, but I was actually thinking that it would look more like:

- fetchInitialData returns a new promise that is only resolved when parseComplete is finished. It's currently resolved when the schema fetch is done, which won't account for any async logic a consumer adds in their parseSchema hook
- In a preRender function, we can then go ahead and use the existing schema (at this time, we know that it will be ready) to go ahead and do this logic to add all the form elements and fields.

Overall idea being that fetchInitialData handles all async data loading (and parsing), and then we can just do view logic like adding elements, etc in a more normal synchronous way like we do elsewhere.

PR #324, opened 2017-10-19
:bug: Minor UX tweaks for OAuth2 Consent

manueltanzi-okta:
- Break words when client/app name or user name is too long.
- Load Default logo when Client logo is not customized

## Screenshots

Default Logo:
screen shot 2017-10-18 at 6 22 24 pm

screen shot 2017-10-19 at 10 54 46 am

Custom Logo:
screen shot 2017-10-18 at 6 23 56 pm

Resolves: [OKTA-145549](https://oktainc.atlassian.net/browse/OKTA-145549)

@mauriciocastillosilva-okta @hor-kanchan-okta @larsjohansen-okta

### DIFF
--- a/assets/sass/modules/_consent.scss
+++ b/assets/sass/modules/_consent.scss
@@ -33,9 +33,18 @@
     left: 59px;
 
     .client-logo {
+      border-radius: $border-radius-default;
+    }
+
+    .custom-logo {
       width: 70px;
       height: 70px;
-      border-radius: $border-radius-default;
+    }
+
+    .default-logo {
+      width: 50px;
+      height: 50px;
+      padding: 10px;
     }
   }
 
@@ -71,6 +80,7 @@
     padding-top: 20px;
     padding-bottom: 25px;
     text-align: center;
+    word-wrap: break-word;
 
     p {
       font-size: 16px;
--- a/src/views/consent/ConsentBeacon.js
+++ b/src/views/consent/ConsentBeacon.js
@@ -31,15 +31,20 @@ define(['okta'], function (Okta) {
       </div>\
       <div class="logo-wrapper consent-beacon-client">\
         <div class="consent-beacon-border"/>\
-          {{#if clientLogo}}\
-            <img class="client-logo" src="{{clientLogo}}" />\
+          {{#if customLogo}}\
+            <img class="client-logo custom-logo" src="{{customLogo}}" />\
+          {{else}}\
+            <img class="client-logo default-logo" src="{{defaultLogo}}" />\
           {{/if}}\
         </div>\
       </div>\
     ',
 
     getTemplateData: function () {
-      return { clientLogo: this.options.appState.get('targetLogo') && this.options.appState.get('targetLogo').href };
+      return {
+        customLogo: this.options.appState.get('targetLogo') && this.options.appState.get('targetLogo').href,
+        defaultLogo: this.options.appState.get('defaultAppLogo')
+      };
     },
 
     equals: function (Beacon) {
--- a/test/unit/helpers/xhr/CONSENT_REQUIRED.js
+++ b/test/unit/helpers/xhr/CONSENT_REQUIRED.js
@@ -12,10 +12,6 @@ define({
         "label": "Janky App",
         "clientId": "8WUqrTzUG9RyJt2C6Gmm",
         "_links": {
-          "logo": {
-            "href": "https://example.com/logo.png",
-            "type": "image/png"
-          },
           "about": {
             "href": "https://example.com/about.html",
             "type": "text/html"
--- a/test/unit/spec/ConsentRequired_spec.js
+++ b/test/unit/spec/ConsentRequired_spec.js
@@ -20,6 +20,10 @@ function (Q, _, $, OktaAuth, LoginUtil, SharedUtil, Util, ConsentRequiredForm, E
   var itp = Expect.itp;
   var tick = Expect.tick;
 
+  function deepClone(res) {
+    return JSON.parse(JSON.stringify(res));
+  }
+
   function setup(settings, res) {
     settings || (settings = {});
     var successSpy = jasmine.createSpy('successSpy');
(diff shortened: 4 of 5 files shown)


src/views/consent/ConsentBeacon.js
```diff
@@ -31,15 +31,20 @@ define(['okta'], function (Okta) {
       </div>\
       <div class="logo-wrapper consent-beacon-client">\
         <div class="consent-beacon-border"/>\
-          {{#if clientLogo}}\
-            <img class="client-logo" src="{{clientLogo}}" />\
+          {{#if customLogo}}\
+            <img class="client-logo custom-logo" src="{{customLogo}}" />\
+          {{else if defaultLogo}}\
```
mauriciocastillosilva-okta: this will never be empty right? Probably this should be an ```{{else}}``` only, without any other condition. That way we know we will have a logo no matter what.

test/unit/helpers/xhr/CONSENT_REQUIRED.js
```diff
@@ -13,7 +13,6 @@ define({
         "clientId": "8WUqrTzUG9RyJt2C6Gmm",
         "_links": {
           "logo": {
-            "href": "https://example.com/logo.png",
```
mauriciocastillosilva-okta: how does the response look like when there is no logo? What we have makes it kind of confusing I think:
```
"_links": {
    "logo": {
        "type": "image/png"
    },
```

I would have thought that the API will not return you a 'logo' node at all if there is no logo.

larsjohansen-okta:
Could you include screenshots for long first name and long last name as well?

assets/sass/modules/_consent.scss
```diff
@@ -33,9 +33,18 @@
...
+      border-radius: $border-radius-default;
+    }
+
+    .custom-logo {
       width: 70px;
       height: 70px;
-      border-radius: $border-radius-default;
+    }
+
+    .default-logo {
```
mauriciocastillosilva-okta: just a question: why are both logos different size? Just curious on why we don't have a default logo that is the same size as any of the custom logos... (what you have works, but just want to know the reason behind this)

manueltanzi-okta: Is because of the image of our gear logo has no "padding" in the image itself. The custom logo could be literally anything, so is on the customer side decide if the image will have "padding" or not.

src/views/consent/ConsentBeacon.js
manueltanzi-okta: I felt this was safer, is baseUrl always available? If yes, I can put back to just else.

mauriciocastillosilva-okta: even if ```baseUrl``` is not, you are returning a string that is not empty, so default logo will always be there.
And well, ```baseUrl``` is [always available](https://github.com/okta/okta-signin-widget#new-oktasigninconfig), it is the only required config param to bootstrap the widget.

manueltanzi-okta: Good point, thank you

test/unit/helpers/xhr/CONSENT_REQUIRED.js
manueltanzi-okta: Yes exactly like that:
screen shot 2017-10-19 at 10 35 20 am

manueltanzi-okta: @larsjohansen-okta can you confirm if this is the correct behavior for the API response?

larsjohansen-okta: I think that's up to @johngronberg-okta to decide. The natural thing would be to either exclude the "logo" section fully or keep it as it is now, which is what you've shown above.

johngronberg-okta: We should be consistent with the behavior we have for other APIs.

test/unit/spec/ConsentRequired_spec.js
```diff
@@ -58,8 +68,13 @@ function (Q, _, $, OktaAuth, LoginUtil, SharedUtil, Util, ConsentRequiredForm, E
           expect(test.form.userLogo()).toHaveClass('person-16-gray');
         });
       });
-      itp('has the correct client logo', function () {
+      itp('has the correct default logo', function () {
```
mauriciocastillosilva-okta: nit: ```has the default logo if client logo is not provided```

test/unit/helpers/xhr/CONSENT_REQUIRED.js
mauriciocastillosilva-okta: @larsjohansen-okta what is the consistent behavior? Seems to me that a ```logo.type: 'image/png'``` without an href is kind of weird, but you are the api expert
@manueltanzi-okta if api remains the same, this looks good. Otherwise, we just need to change this mock response. It will definitely be ideal if we can have our test response exactly the same as the api.

src/views/consent/ConsentBeacon.js
```diff
@@ -31,15 +31,20 @@ define(['okta'], function (Okta) {
...
           {{/if}}\
         </div>\
       </div>\
     ',
 
     getTemplateData: function () {
-      return { clientLogo: this.options.appState.get('targetLogo') && this.options.appState.get('targetLogo').href };
+      return {
+        customLogo: this.options.appState.get('targetLogo') && this.options.appState.get('targetLogo').href,
+        defaultLogo: this.options.appState.get('baseUrl') + '/img/logos/default.png'
```
mauriciocastillosilva-okta: looking at this again... sorry I didn't see this earlier, but can we move this to the appState model as a derived property? so here we can do
```
 defaultLogo: this.options.appState.get('defaultAppLogo');
```
and we can have '/img/logos/default.png' as a constant there in AppState, similar to other image files we have there.

test/unit/spec/ConsentRequired_spec.js
```diff
@@ -50,6 +54,16 @@ function (Q, _, $, OktaAuth, LoginUtil, SharedUtil, Util, ConsentRequiredForm, E
     return Expect.waitForConsentRequired(settings);
   }
 
+  function setupClientLogo() {
+    var resConsentRequiredClientLogo = deepClone(resConsentRequired);
+    var customLogo = {
+      href: 'https://example.com/logo.png',
```
mauriciocastillosilva-okta: nit nit: ```.com/custom-logo.png``` so when we see the test in line 82 is really obvious that we are expecting a custom logo

src/views/consent/ConsentBeacon.js
manueltanzi-okta: Sure thank you for pointing this out, I wasn't actually sure if this was clean

PR #325, opened 2017-10-19
Disable Register button when password complexity errors are shown

nikhilvenkatraman-okta:
🌱  Disable Register button when password complexity errors are shown  

screen shot 2017-10-18 at 11 27 04 pm

Resolves: OKTA-145577

@hor-kanchan-okta @richardheng-okta

### DIFF
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -115,11 +115,16 @@ function (
         };
 
         // event listeners to toggle registerButton
-        form.listenTo(this.model, 'disableRegister', function(){
+        form.listenTo(this.model, 'disableRegister invalid', function(){
           toggleRegisterButton(true);
         });
-        form.listenTo(this.model, 'enableRegister', function(){
+        form.listenTo(this.model, 'enableRegister valid', function(){
           toggleRegisterButton(false);
+        });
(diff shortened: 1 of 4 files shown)


nikhilvenkatraman-okta:
@hor-kanchan-okta updated this PR to have only changes for OKTA-145577

nikhilvenkatraman-okta:
@hor-kanchan-okta PR updated

hor-kanchan-okta:
I wonder how this work with the required fields.
If I have required field but is empty. Will the "register" button be disabled?

nikhilvenkatraman-okta:
Good catch. I just realized that too. I have to change the listener to also fire the toggle method when model fires "invalid" event. When the form has required fields empty the invalid event is fired. That should do it i guess.

```javascript
        form.listenTo(this.model, 'disableRegister invalid', function(){
           toggleRegisterButton(true);
         });

```

nikhilvenkatraman-okta:
https://okta.box.com/s/l1lvq4v9m2uh26dz4sklt7t3mh5atytu

nikhilvenkatraman-okta:
@hor-kanchan-okta added video to show  required fields use case along

PR #327, opened 2017-10-19
Fire pwd complexity check when user enters username after password

nikhilvenkatraman-okta:
🐛 Fixed bug to trigger password complexity check when user enters username or tabs out of the username field.

Resolves: OKTA-144888

@hor-kanchan-okta @richardheng-okta

### DIFF
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -128,6 +128,13 @@ function (
             form.add(SubSchema.extend({id: 'subschemas-' + name, subSchemas: subSchemas}));
           }
         });
+        // fire password validation when username changes
+        this.listenTo(this.model, 'change:userName', function() {
+          if (this.model.get('userName') && this.model.get('password')) {
+            var passwordField = Okta.$('.registration .okta-form-input-field input[type="password"]');
+            passwordField.trigger('blur');
+          }
+        });
         this.add(form);
         this.footer = new this.Footer(this.toJSON());
         this.add(this.footer);
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -128,13 +128,6 @@ function (
             form.add(SubSchema.extend({id: 'subschemas-' + name, subSchemas: subSchemas}));
           }
         });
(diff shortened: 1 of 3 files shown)


src/util/RegistrationFormFactory.js
```diff
@@ -142,6 +142,16 @@ define([
       inputOptions.params = {
         'icon': 'person-16-gray'
       };
+      inputOptions.events = {
```
hor-kanchan-okta: probably better to move it to the controller. Because you have one field depends on the other field. The controller should be the better place to coordinate the logic.

nikhilvenkatraman-okta: I thought about that, but i kept it here for consistency since we create the inputs in the formFactory and have listed out the events for the password field here as well. I'll try to move it to the controller and see if its better.

nikhilvenkatraman-okta:
@hor-kanchan-okta PR updated

src/RegistrationController.js
```diff
@@ -128,6 +128,13 @@ function (
             form.add(SubSchema.extend({id: 'subschemas-' + name, subSchemas: subSchemas}));
           }
         });
+        // fire password validation when username changes
+        this.listenTo(this.model, 'change:userName', function() {
+          if (this.model.get('userName') && this.model.get('password')) {
+            var passwordField = Okta.$('.registration .okta-form-input-field input[type="password"]');
```
hor-kanchan-okta: Is there a way for the password field to listen to the "change:userName" event?
Also the get rid of the selector too.

nikhilvenkatraman-okta: I would have to try that out. But that would move the logic back to RegistrationFormFactory.js?

nikhilvenkatraman-okta: That does seem to work, and its a much smaller change. I can revert this and move it back to the RegistrationFormFactory.js

nikhilvenkatraman-okta:
@hor-kanchan-okta added test to verify the order in which the fields are filled.

PR #329, opened 2017-10-25
:bug: Fix untraslated strings from error code in okta-signin-widget

manueltanzi-okta:
- Fix untranslated strings from error code in okta-signin-widget
- Add missing tests for transformErrorXHR function

Resolves: [OKTA-144534](https://oktainc.atlassian.net/browse/OKTA-144534)

## Screenshot
screen shot 2017-10-25 at 2 20 53 pm

@hor-kanchan-okta @upteam-okta

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -38,6 +38,8 @@ define(['okta'], function (Okta) {
     // Assuming there is only one field error in a response
     if (xhr.responseJSON && xhr.responseJSON.errorCauses && xhr.responseJSON.errorCauses.length) {
       xhr.responseJSON.errorSummary = xhr.responseJSON.errorCauses[0].errorSummary;
+      // BaseForm will consume errorCauses before errorSummary if it is an array, so, we have to make sure to remove it
+      delete xhr.responseJSON.errorCauses;
     }
     // Replace error messages
     if (!_.isEmpty(xhr.responseJSON)) {
--- a/test/unit/spec/PasswordExpired_spec.js
+++ b/test/unit/spec/PasswordExpired_spec.js
@@ -294,7 +294,7 @@ function (Q, _, $, OktaAuth, LoginUtil, SharedUtil, Util, PasswordExpiredForm, B
         .then(function (test) {
           expect(test.form.hasErrors()).toBe(true);
           expect(test.form.errorMessage()).toBe(
-            'We found some errors. Please review the form and make corrections.'
+            'Old password is not correct'
           );
         });
       });
--- a/test/unit/spec/Util_spec.js
+++ b/test/unit/spec/Util_spec.js
@@ -1,6 +1,61 @@
+/* eslint max-len: [2, 140] */
 define(['util/Util'], function (Util) {
 
   describe('util/Util', function () {
+
+    describe('transformErrorXHR', function () {
+      it('errorSummary shows network connection error when status is 0', function () {
+        var xhr = {
+          'status': 0
+        };
+        Util.transformErrorXHR(xhr);
+        expect(xhr.responseJSON.errorSummary).toEqual('Unable to connect to the server. Please check your network connection.');
+      });
+      it('errorSummary shows internal error when there are no responseJSON and no responseText', function () {
+        var xhr = {
+          'status': 400
+        };
+        Util.transformErrorXHR(xhr);
+        expect(xhr.responseJSON.errorSummary).toEqual('There was an unexpected internal error. Please try again.');
+      });
+      it('errorSummary is set from responseText when there is no responseJSON', function () {
+        var responseText = '{"errorSummary": "errorSummary from responseText"}';
+        var xhr = {
+          'status': 400,
+          'responseText': responseText
+        };
+        Util.transformErrorXHR(xhr);
+        expect(xhr.responseJSON.errorSummary).toEqual('errorSummary from responseText');
+      });
+      it('If there is an errorCauses array and there is no valid error code, get errorSummary from errorCauses array', function () {
+        var errorCauses = [{
+          'errorSummary': 'errorSummary from errorCauses'
+        }];
+        var xhr = {
+          status: 400,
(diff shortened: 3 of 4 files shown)


src/util/Util.js
```diff
@@ -44,6 +44,9 @@ define(['okta'], function (Okta) {
       var errorMsg = Okta.loc('errors.' + xhr.responseJSON.errorCode, 'login');
       if (errorMsg.indexOf('L10N_ERROR[') === -1) {
         xhr.responseJSON.errorSummary = errorMsg;
+        if(xhr.responseJSON.errorCauses[0] && xhr.responseJSON.errorCauses[0].errorSummary) {
```
hor-kanchan-okta: Can you add a comment explain how it solve the problem
i.e.: BaseForm will consume the `errorCauses` if it is an array... That takes the priority of the `errorSummary`.... Delete the `errorCauses` array since we want to show the translated message by `errorCode`...

hor-kanchan-okta:
Th rest is good

src/util/Util.js
mauriciocastillosilva-okta: yes, please add that comment cause I do not understand how this is fixing an error

mauriciocastillosilva-okta: Talked to Manuel offline, makes sense now to me. 
We should just move ```delete xhr.responseJSON.errorCauses;``` to line 41, in the same if block where we are overriding the top errorSummary.

src/views/mfa-verify/InlineTOTPForm.js
```diff
@@ -28,6 +28,9 @@ define(['okta', 'views/shared/TextBox'], function (Okta, TextBox) {
       className: 'button inline-totp-verify',
       title: Okta.loc('mfa.challenge.verify', 'login'),
       click: function () {
+        if(!form.isValid()) {
```
hor-kanchan-okta: Do you need this?

manueltanzi-okta: Yes, this is not the form button so it doesn't trigger this. The property was actually already required, but was not working because of this reason. @mauriciocastillosilva-okta

manueltanzi-okta: They do directly `form.model.save();`, skipping the form validation.

manueltanzi-okta: The factor type is 'token:software:totp', as you can see [here](https://github.com/okta/okta-signin-widget/blob/master/src/models/Factor.js#L174) is already set to required. This is why this is working when I add form validation.

hor-kanchan-okta: I think should be a different PR as this fix is specific to the one-time password.

test/unit/spec/Util_spec.js
```diff
@@ -1,7 +1,62 @@
...
+        expect(xhr.responseJSON.errorSummary).toEqual('Unable to connect to the server. Please check your network connection.');
+      });
+      it('errorSummary shows internal error when there are no responseJSON and no responseText', function () {
+        var xhr = {
+          'status': 400
+        };
+        Util.transformErrorXHR(xhr);
+        expect(xhr.responseJSON.errorSummary).toEqual('There was an unexpected internal error. Please try again.');
+      });
+      it('Set responseJSON to parsed responseText when there is no responseJSON', function () {
```
mauriciocastillosilva-okta: nit: ```error summary is set from responseText when there is no responseJSON```

test/unit/spec/PasswordExpired_spec.js
```diff
@@ -294,7 +294,7 @@ function (Q, _, $, OktaAuth, LoginUtil, SharedUtil, Util, PasswordExpiredForm, B
         .then(function (test) {
           expect(test.form.hasErrors()).toBe(true);
           expect(test.form.errorMessage()).toBe(
-            'We found some errors. Please review the form and make corrections.'
+            'oldPassword: The credentials provided were incorrect.'
```
mauriciocastillosilva-okta: this doesn't seem right... the error is a field error, so it should probably show right below the oldPass field, not as a generic error. My guess is that BaseForm needs the errorCauses for field errors. Can you please check?

manueltanzi-okta: I think that was wrong before and correct now after my change, not the opposite. The correct behavior in this case is that it shows the errorSummary in the errorCause array, since that errorCode doesn't exists in login.properties. Look at the test right after this, it pass the right error as well in errorCauses.

manueltanzi-okta: https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/xhr/PASSWORD_EXPIRED_error_oldpass.js

https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/xhr/PASSWORD_EXPIRED_error_complexity.js

mauriciocastillosilva-okta: I wonder then if that is actually wrong is the mock xhr we have... my concern here is that ```oldPassword: The credentials provided were incorrect.``` is the way the server returns a field error (field ```oldPassword``` has an error of ```The credentials provided were incorrect.```

Can you check a real response for this case and see if our mock xhr is missing something? It might be missing an actual errorSummary

manueltanzi-okta: screen shot 2017-10-31 at 4 04 08 pm

manueltanzi-okta: I think this mock xhr is outdated, right? @mauriciocastillosilva-okta

mauriciocastillosilva-okta: I see. Yes, the mock is not really a true response. Can you update it?

manueltanzi-okta: yepp!

PR #330, opened 2017-10-26
Remove registration.clientId parameter 

nikhilvenkatraman-okta:
Removed clientId parameter from within the registration config object

- Updated registration code to use clientId param from sign in widget config object instead of registration.clientId
- Within the sign in widget authScheme defaults to OAUTH2 
- If registration is enabled we want to use clientId but do not want  OAUTH flow to be triggered once registration is completed.
- We now set authScheme to SESSION if registration is enabled for okta-hosted usecases. This value is set when sign in widget is bootstrapped from loginpage_backbone.jsp. Related okta-core PR
https://github.com/okta/okta-core/pull/23708 
- https://github.com/okta/okta-signin-widget/blob/master/src/models/Settings.js#L200 will return false when clientId is set and authScheme is set to SESSION

Resolves: OKTA-135949

### DIFF
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -61,12 +61,9 @@ function (
   return BaseLoginController.extend({
     className: 'registration',
     initialize: function() {
-      var clientId = this.options.settings.get('registration.clientId');
+      var clientId = this.options.settings.get('clientId');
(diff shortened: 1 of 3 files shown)


nikhilvenkatraman-okta:
@hor-kanchan-okta thanks ! removed references.

PR #339, opened 2017-11-03
Use handlebars wrapper instead of external handlebars

santhoshbalakrishnan:
## Description

Point to Okta handlebars wrapper instead of external handlebars. This solves references to i18n and other helpers in templates.

This also solves the issue where okta-signin-widget did not load on samples.

## Reviewers

- @lboyette-okta 
-  @rchild-okta


santhoshbalakrishnan:
Tested against monolith. Works good!

PR #342, opened 2017-11-08
:bug: Fix field errors from backend in okta-signin-widget

manueltanzi-okta:
Baseform uses the errorSummary in the errorCauses array to display error fields coming from backend. So, we cannot always delete the array, we should delete the array only when the backend doesn't return an error field. Everytime an error field is returned, if there is an errorCode in the response that will be invalid, since it requires the field parameters and we don't pass parameters to the Okta.loc function in okta-signin-widget. So, with this fix, everytime we get an error field from backend the errorCause array will not be deleted and the error field will be displayed correctly.

Resolves: [OKTA-148144](https://oktainc.atlassian.net/browse/OKTA-148144)

## Screenshot
screen shot 2017-11-08 at 11 49 02 am

@mauriciocastillosilva-okta @hor-kanchan-okta

### DIFF
--- a/test/unit/spec/Util_spec.js
+++ b/test/unit/spec/Util_spec.js
@@ -32,14 +32,29 @@ define(['util/Util'], function (Util) {
           'errorSummary': 'errorSummary from errorCauses'
         }];
         var xhr = {
-          status: 400,
-          responseJSON: {
-            errorCauses: errorCauses
+          'status': 400,
+          'responseJSON': {
+            'errorCauses': errorCauses
           }
         };
         Util.transformErrorXHR(xhr);
         expect(xhr.responseJSON.errorSummary).toEqual('errorSummary from errorCauses');
-        expect(xhr.responseJSON.errorCauses).toBeDefined();
+        expect(xhr.responseJSON.errorCauses).toBe(errorCauses);
+      });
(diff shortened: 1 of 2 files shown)


test/unit/spec/Util_spec.js
```diff
@@ -39,6 +39,7 @@ define(['util/Util'], function (Util) {
         };
         Util.transformErrorXHR(xhr);
         expect(xhr.responseJSON.errorSummary).toEqual('errorSummary from errorCauses');
+        expect(xhr.responseJSON.errorCauses).toBeDefined();
```
mauriciocastillosilva-okta: can we check that the object is the same as our ```errorCauses``` from line 31?

Second thing: can we add one more test with an invalid errorCode (like ```errorCode: "1212AB"```) instead of an undefined one? So we can test both scenarios

manueltanzi-okta: yepp!

test/unit/spec/Util_spec.js
```diff
@@ -32,13 +32,29 @@ define(['util/Util'], function (Util) {
           'errorSummary': 'errorSummary from errorCauses'
```
mauriciocastillosilva-okta: description for this test should be ```If there is an errorCauses array and there is no error code, get...```

hor-kanchan-okta:
👍

PR #349, opened 2017-11-14
:bug: Removes spinner in reset password page on fail

santhoshbalakrishnan:
## Description

**Issue:** If security image is disabled, the spinner still remains in the loading state,

**Fix:** Trigger 'removeLoading' to remove the spinner in the case of no security image. Setting loading to 'false' removes the spinner in the case where security image exists.

Also modifying CSS to remove overlap of beacon on the infobox error. Made sure we still look good on other pages as well. Check screenshots.

## Reviewers

- @hor-kanchan-okta 
- @mauriciocastillosilva-okta 
- @lboyette-okta 

## Screenshots

**After**

See that the spinner disappears here:

reset-spinner

Responsiveness with the minor CSS change to auth content padding:

after-css-signin
after-css

**Before**

Note the overlap between beacon and infobox error:

screen shot 2017-11-14 at 2 10 56 pm

### DIFF
--- a/src/RecoveryLoadingController.js
+++ b/src/RecoveryLoadingController.js
@@ -29,6 +29,7 @@ define(['okta', 'util/FormController'], function (Okta, FormController) {
       })
       .fail(function () {
         self.options.appState.trigger('loading', false);
+        self.appState.trigger('removeLoading');
       });
     },
 
--- a/assets/sass/okta-sign-in.scss
+++ b/assets/sass/okta-sign-in.scss
@@ -55,17 +55,13 @@
   @import 'modules/registration';
 
   &.can-remove-beacon {
-    .auth-content {
-      padding-top: 40px;
-    }
-
     @include max-height-mq(550px) {
       .beacon-container {
         display: none;
       }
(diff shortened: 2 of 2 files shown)


src/RecoveryLoadingController.js
```diff
@@ -29,6 +29,7 @@ define(['okta', 'util/FormController'], function (Okta, FormController) {
       })
       .fail(function () {
         self.options.appState.trigger('loading', false);
+        self.appState.trigger('removeLoading');
```
mauriciocastillosilva-okta: is this a problem in other controllers? This will fix the recovery flows, but what about other flows?

santhoshbalakrishnan: Luckily, other places that remove loading change the page itself on app state change. So we do not encounter this anywhere else. The one place where this is possible is primaryAuthController and it already has this `removeLoading` call in it.

santhoshbalakrishnan:
@hor-kanchan-okta Missed adding CSS changes to the commit. Added now.

Issue #320, opened 2017-10-16
NativePromise.resolve is not a function


fehrinjt:
I am running an angular 4 project and the latest 2.3 build is throwing the following errors:

```
Uncaught TypeError: NativePromise.resolve is not a function
    at Object.setNativePromise (zone.js:634)
    at desc.set (zone.js:1028)
    at t.exports (scripts.bundle.js:14)
    at u (scripts.bundle.js:14)
    at Object.<anonymous> (scripts.bundle.js:15)
    at e (scripts.bundle.js:14)
    at Object.<anonymous> (scripts.bundle.js:14)
    at e (scripts.bundle.js:14)
    at Object.<anonymous> (scripts.bundle.js:14)
    at Object.<anonymous> (scripts.bundle.js:14)
zone.js:244 

Uncaught TypeError: global[symbolPromise].resolve is not a function
    at scheduleMicroTask (zone.js:581)
    at ZoneDelegate.webpackJsonp.../../../../zone.js/dist/zone.js.ZoneDelegate.scheduleTask (zone.js:414)
    at Zone.webpackJsonp.../../../../zone.js/dist/zone.js.Zone.scheduleTask (zone.js:236)
    at Zone.webpackJsonp.../../../../zone.js/dist/zone.js.Zone.scheduleMicroTask (zone.js:256)
    at scheduleResolveOrReject (zone.js:871)
    at ZoneAwarePromise.then (zone.js:981)
    at PlatformRef_.webpackJsonp.../../../core/@angular/core.es5.js.PlatformRef_._bootstrapModuleWithZone (core.es5.js:4537)
    at PlatformRef_.webpackJsonp.../../../core/@angular/core.es5.js.PlatformRef_.bootstrapModule (core.es5.js:4522)
    at Object.../../../../../src/main.ts (main.bundle.js:963)
    at __webpack_require__ (inline.bundle.js:55)`
```
**Reverting to 2.1 fixes any issues and it functions normally.**

Angular/system info:
```
@angular/cli: 1.4.4
node: 6.9.5
os: win32 x64
@angular/common: 4.4.4
@angular/compiler: 4.4.4
@angular/core: 4.4.4
@angular/forms: 4.4.4
@angular/http: 4.4.4
@angular/platform-browser: 4.4.4
@angular/platform-browser-dynamic: 4.4.4
@angular/router: 4.4.4
@angular/cli: 1.4.4
@angular/compiler-cli: 4.4.4
[...]

PR #353, opened 2017-11-20
:bug: Fix improper ES6 feature bundling

lboyette-okta:
`babel-polyfill` "is intended to be used in an application rather than a library/tool" - [source](https://babeljs.io/docs/usage/polyfill/)

The runtime transform plugin "is recommended in a library/tool" - [source](https://babeljs.io/docs/plugins/transform-runtime/)

The widget is a library, so I changed the way we bundle ES6 features

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,23 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@okta/okta-auth-js": {
+      "version": "1.11.0",
+      "resolved": "https://registry.npmjs.org/@okta/okta-auth-js/-/okta-auth-js-1.11.0.tgz",
+      "integrity": "sha1-k6/QIE1w1vasBCBmfa/Ff156j/4=",
+      "requires": {
+        "Base64": "0.3.0",
+        "q": "1.4.1",
+        "reqwest": "2.0.5",
+        "tiny-emitter": "1.1.0",
+        "xhr2": "0.1.3"
+      }
+    },
+    "Base64": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/Base64/-/Base64-0.3.0.tgz",
+      "integrity": "sha1-baJhpOgNT6D1xoQlTlvM0Wu9zp8="
+    },
     "abbrev": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.0.tgz",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -43,6 +43,8 @@
     "axe-webdriverjs": "0.4.0",
     "babel-core": "6",
     "babel-loader": "6",
+    "babel-runtime": "^6.26.0",
```
manueltanzi-okta: Should we put this under dependencies? https://babeljs.io/docs/plugins/transform-runtime/

rchild-okta:
Looks great - I do have a question about what they mean by *NOTE: Instance methods such as "foobar".includes("foo") will not work since that would require modification of existing built-ins (Use babel-polyfill for that).*

Does this mean that it will not transpile those types of esNext features, and if we use those we'll be in trouble?

lboyette-okta:
@rchild-okta 

In order to polyfill methods like `[].includes`, you have to modify the Array prototype, which is a global effect. If we could scope babel-polyfill to autodetect and only use polyfills we want, that'd be great! Unfortunately there's no way to reduce the scope of babel-polyfill other than [useBuiltIns](https://babeljs.io/docs/plugins/preset-env/). Unfortunately, useBuiltIns will include the `Promise` polyfill, which was causing zone.js to throw an error: #320 

I've been looking for solutions that will load the polyfills at runtime, but I'm not sure that's a route we want to go.

One alternative is to require end-developers, including us, to include polyfills in the production app that uses the widget. This may be strange to devs who use our cdn. They'd now need to reference a polyfill AND okta-signin-widget.

lboyette-okta:
@rchild-okta As we discussed, we now only include babel-polyfill on assets served from our cdn (`okta-sign-in.min.js` and `okta-sign-in-no-jquery.js`).

package-lock.json
```diff
@@ -1,21 +1,9 @@
 {
   "name": "@okta/okta-signin-widget",
-  "version": "2.2.0",
+  "version": "2.4.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@okta/okta-auth-js": {
```
rchild-okta: Why is authJs being removed from the dependencies?

webpack.common.config.js
```diff
@@ -121,7 +121,8 @@ module.exports = function (outputFilename) {
           exclude: /node_modules/,
           loader: 'babel-loader',
           query: {
-            presets: ['env']
+            presets: ['env'],
+            plugins: ['transform-runtime']
```
rchild-okta: Am curious - if we have transform-runtime + babel-polyfill (for the cdn version), does that mean we have two copies of all the polyfills? In general, what is the difference in file size in the built file between this commit and current master?

package-lock.json
lboyette-okta: weird, dunno. I'll rebuild the package-lock.json

webpack.common.config.js
lboyette-okta: Before
okta-sign-in.min.js - 989 kB
okta-sign-in-no-jquery.js- 891 kB
okta-sign-in.entry.js - 1815 kB

After
okta-sign-in.min.js - 1020 kB
okta-sign-in-no-jquery.js- 921 kB
okta-sign-in.entry.js - 1640 kB

Latest commit
okta-sign-in.min.js - 989 kB
okta-sign-in-no-jquery.js- 891 kB
okta-sign-in.entry.js - 1640 kB

PR #354, opened 2017-11-22
Perf: use latest @okta/courage to register only used form inputs

ukilon-okta:


mauriciocastillosilva-okta:
So this means that we now have ```Okta.js``` that by default includes no inputs, and we also have ```OktaWithInputs.js``` which is what was previously ```Okta.js```?

ukilon-okta:
@mauriciocastillosilva-okta pretty much, yes

mauriciocastillosilva-okta:
out of curiosity, how did you find out which inputs we use in the signin widget? just a grep for ```addInput``` and check the ```type```?

ukilon-okta:
@mauriciocastillosilva-okta pretty much. grep for `type: `, run sanity tests, and groom the unit tests.

mauriciocastillosilva-okta:
looks good 🦃

mauriciocastillosilva-okta:
just one last question: any reason the courage ```package.json``` did not change?

packages/@okta/courage/package.json
```diff
@@ -1,56 +1,61 @@
...
+    "escapedName": "@okta%2fcourage",
     "scope": "@okta",
-    "spec": "3.3.0-beta.12",
-    "type": "version"
+    "rawSpec": "4.1.0-beta.1",
+    "saveSpec": null,
+    "fetchSpec": "4.1.0-beta.1"
   },
   "_requiredBy": [
+    "#USER",
```
mauriciocastillosilva-okta: I still haven't look into if this value matters or not since we are using ```packages``` folder, but if it does it should be kept as ```/``` (long explanation: ```/``` means the i18n module is required by the root module. ```#USER``` means we ran 'npm install' but we did not save the installed module)

jmelberg-okta:
I'm 99% sure this is out of date and can be closed. @santhoshbalakrishnan or @mauriciocastillosilva-okta can you confirm?

mauriciocastillosilva-okta:
yeah, this is. I believe @nikhilvenkatraman-okta merged a change to deal with this. @nikhilvenkatraman-okta can you confirm?

jmelberg-okta:
Thanks all for the update.

I’m closing this PR, as the changes have been included in the commit mentioned above.

jmelberg-okta:
Thanks all for the update.

I’m closing this PR, as the changes have been included in the commit mentioned above.

PR #358, opened 2017-12-06
:memo: Update readme to point to latest cdn url

nikhilvenkatraman-okta:
Update readme to point to latest version of the widget (2.4.0) in the cdn url 

Resolves: OKTA-151620

@hor-kanchan-okta  @rajaguruduraisamy-okta


nikhilvenkatraman-okta:
FYI that we have a new version of the widget 2.5.0 but its not on CDN yet. We already have a ticket to update the read me when it becomes available on CDN https://oktainc.atlassian.net/browse/OKTA-151548

PR #362, opened 2017-12-11
:art: build source-map for okta-signin-in.entry.js

yiyangli-okta:
Issue: dist/ folder is built via `grunt build:release`, it includes all distributable files, however, the source-map file is missing, consequently, any projects consuming okta-signin-widget can't merge the source map and debug the codes in okta-signin-widget;
Fix: fix the webpack config for okta-sign-in.entry.js, and generate a map for it.

### DIFF
--- a/webpack.release.config.js
+++ b/webpack.release.config.js
@@ -64,7 +64,7 @@ cdnConfig.plugins = [
     compress: {
       warnings: false
     },
-    sourceMap: false,
+    sourceMap: true,
     comments: function(node, comment) {
       // Remove other Okta copyrights
       var isLicense = /^!/.test(comment.value) ||
--- a/webpack.common.config.js
+++ b/webpack.common.config.js
@@ -8,6 +8,7 @@ var VENDOR = path.resolve(__dirname, 'packages/@okta/courage/src/vendor');
 module.exports = function (outputFilename) {
   return {
     entry: ['./target/js/widget/OktaSignIn.js'],
+    devtool: 'source-map',
     output: {
       path: TARGET_JS,
       filename: outputFilename,


yiyangli-okta:
screen shot 2017-12-11 at 4 36 19 pm

webpack.release.config.js
```diff
@@ -64,7 +64,7 @@ cdnConfig.plugins = [
     compress: {
       warnings: false
     },
-    sourceMap: false,
+    sourceMap: true,
     comments: function(node, comment) {
```
rajaguruduraisamy-okta: Looks like the developer explicitly mentioned not to send sourceMap.

yiyangli-okta:
1:07 after the change
0:47 before the change

the fun fact is that the more a source code is minified, the bigger the map file is, and consequently, the longer time the build time is, and also the longer time the load/map time is during debugging.

webpack.dev.config.js
```diff
@@ -1,5 +1,4 @@
 var common = require('./webpack.common.config');
 var devConfig = common('okta-sign-in.js');
-devConfig.devtool = '#inline-source-map';
```
mauriciocastillosilva-okta: can we keep the dev build a little faster? Maybe explore other options for [devtool](https://webpack.js.org/configuration/devtool/)

I also heard @richardheng-okta researched this values recently...

yiyangli-okta: https://webpack.js.org/configuration/devtool/ `inline-source-map` is the same as `source-map` in building time, the change affects `grunt webpack:release` only.

mauriciocastillosilva-okta: no, the change to ```webpack.common.config.js``` affects all builds, release, dev and test. (although dev was using ```inline-source-map``` as you pointed out, so the build time is basically the same)... but that is why I was suggesting maybe explore other option for dev... 
I would be ok if we want to do that later as well.

yiyangli-okta: well, I mean, with my change, the build time doesn't change a lot for *dev environment*, because `inline-source-map` is same as `source-map` in building performance. I can't foresee a space to optimize if we want to have a version that allows us to debug. The fact is unchanged with my change: dev would be faster (same faster space with my change) than release for 2 reasons: dev version is not uglified, thus the source-map is faster to build; and dev only need to build one js file, while release need to build 3.

richardheng-okta: If you decide to change it in the future, I have found that Chrome handles inline-source-map a bit better during debugging: https://github.com/webpack/webpack/issues/2145#issuecomment-325441145

yiyangli-okta: okay, I reverted to use `inline-source-map`, thanks for the link, it provides more info than the [documentation of devtool](https://webpack.js.org/configuration/devtool/).

PR #363, opened 2017-12-12
:seedling: Supports Password Re-auth

gowthamidommety-okta:
screen shot 2017-12-13 at 10 42 28 am


assets/sass/okta-theme.scss
```diff
@@ -482,6 +482,18 @@
     }
   }
 
+  .mfa-okta-password {
+    /* -- Factor Icons (large): Password -- */
+    background-image: url('../img/icons/mfa/u2f_70x70.png');
```
gowthamidommety-okta: Using u2f images until we get the one's for password

src/models/Factor.js
```diff
@@ -200,7 +205,11 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
         };
         if (this.get('factorType') === 'question') {
           data.answer = this.get('answer');
-        } else {
+        }
+        else if(this.get('factorType') === 'password') {
```
mauriciocastillosilva-okta: nit: space after ```if```

src/util/FactorUtil.js
```diff
@@ -106,6 +106,12 @@ function (Okta, TimeUtil) {
       iconClassName: 'mfa-okta-security-question',
       sortOrder: 12
     },
+    'PASSWORD': {
+      label: 'factor.password',
+      description: '',
+      iconClassName: 'mfa-okta-password',
+      sortOrder: 13
```
mauriciocastillosilva-okta: I feel this should be higher sort order... probably 11 (so higher than email and question)

src/MfaVerifyController.js
```diff
@@ -114,6 +118,9 @@ function (Okta, Checkbox, BaseLoginController, CookieUtil, TOTPForm, YubikeyForm
     },
 
     trapAuthResponse: function () {
+      if(this.options.factorType === 'password') {
```
mauriciocastillosilva-okta: will this still be needed once the backend returns ```MFA_REQUIRED```?

test/unit/helpers/xhr/MFA_VERIFY_invalid_password.js
```diff
@@ -0,0 +1,16 @@
+define({
+  'status': 403,
+  'responseType': 'json',
+  'response': {
+    'errorCode': 'E0000068',
+    'errorSummary': 'Invalid Passcode\/Answer',
```
mauriciocastillosilva-okta: Is this correct? I would think the server will return ```Invalid Password``` or something along those lines

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2640,6 +2652,160 @@ function (Okta,
...
+          return setupPassword().then(function (test) {
+            expectHasRightBeaconImage(test, 'mfa-okta-password');
+          });
+        });
+        itp('shows the right title', function () {
+          return setupPassword().then(function (test) {
+            expectTitleToBe(test, 'Password');
+          });
+        });
+        itp('has an answer field', function () {
```
mauriciocastillosilva-okta: ```has a password field```

src/util/FactorUtil.js
gowthamidommety-okta: But it will not show up in enrollment right now.. Should we change it once we support enroll?

mauriciocastillosilva-okta: ah right... good point. It doesn't matter for now then.

austinstanhope-okta:
I don't have any comments.  Looks OK to me but I'll defer to @mauriciocastillosilva-okta

src/MfaVerifyController.js
gowthamidommety-okta: Yea, if it's MFA_REQUIRED it will trap the response and will not move on to the normal mfa flow

test/unit/helpers/xhr/MFA_VERIFY_invalid_password.js
gowthamidommety-okta: Yea, it's correct. The error summary is `Invalid Passcode/Answer` while the errorCause is `Password is incorrect`.

packages/@okta/i18n/package.json
```diff
@@ -87,19 +87,19 @@
...
+    "raw": "@okta/i18n@40.2.0",
+    "rawSpec": "40.2.0",
     "scope": "@okta",
-    "spec": "37.4.0",
+    "spec": "40.2.0",
     "type": "version"
   },
   "_requiredBy": [
-    "/"
+    "#USER"
```
mauriciocastillosilva-okta: mind changing this to ```/``` please? I update the [i18n wik](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/238007322/Localization+Dev#Localization(Dev)-Inokta-signin-widget)i with that step

src/MfaVerifyController.js
mauriciocastillosilva-okta: can you explain this to me a little more? I think that if we get ```MFA_REQUIRED``` then it is just like any other factor, so steps are something like

1. bootstrap with a stateToken
2. refreshAuthSate with that stateToken -> returns MFA_REQUIRED
3. RouterUtil navigate to MFA_REQUIRED view, with the first factor on the list, which is password
4. Type password and verify -> response is MFA_REQUIRED again with a different list of factors
5. RouterUtil navigates again to MFA_REQUIRED view, but with a different factor

Is this the flow?

src/views/mfa-verify/PasswordForm.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define(['okta', './inputs/ShowPasswordBox'], function (Okta, PasswordBox) {
+
+  var _ = Okta._;
+
+  return Okta.Form.extend({
+    className: 'mfa-verify-question',
```
mauriciocastillosilva-okta: should this be ```mfa-verify-password```? and maybe we need to update some css class as well

packages/@okta/i18n/package.json
gowthamidommety-okta: Sorry! Forgot that step

src/views/mfa-verify/PasswordForm.js
gowthamidommety-okta: 👍 Missed that. Thanks!

src/MfaVerifyController.js
mauriciocastillosilva-okta: got it. Makes sense... guess it goes back to the initial talk we have about removing ```isMfaRequired``` from the ```if``` block here... which will make this work for more than 1 mfa_required responses.
But I remember there was a problem with that approach too, right?

gowthamidommety-okta: Yea it causes a problem when a factor is in mfa challenge and you switch to another factor in mfa required

src/views/mfa-verify/PasswordForm.js
```diff
@@ -15,7 +15,7 @@ define(['okta', './inputs/ShowPasswordBox'], function (Okta, PasswordBox) {
   var _ = Okta._;
 
   return Okta.Form.extend({
-    className: 'mfa-verify-question',
+    className: 'mfa-verify-password',
```
gowthamidommety-okta: This is only used in tests.. I didnt have a test using it because it is used in tests that change the factor using the dropdown

assets/sass/okta-theme.scss
```diff
@@ -787,6 +787,7 @@
   .mfa-verify-totp-inline { /* -- Page specific top level selectors -- */ }
   .mfa-verify-question { /* -- Page specific top level selectors -- */ }
   .mfa-verify-passcode { /* -- Page specific top level selectors -- */ }
+  .mfa-verify-password { /* -- Page specific top level selectors -- */ }
```
gowthamidommety-okta: Added this to be consistent.

src/MfaVerifyController.js
mauriciocastillosilva-okta: We had an offline conversation about this. This is good and will work perfect for password re-auth, and we will file a jira to make this generic enough to support chained MFA_REQUIRED responses for any factor.

PR #364, opened 2017-12-13
:bug: Changes unauthenticated state route to map to default route

gowthamidommety-okta:
Resolves: OKTA-152402

### DIFF
--- a/src/util/RouterUtil.js
+++ b/src/util/RouterUtil.js
@@ -211,7 +211,7 @@ function (Okta, Util, OAuth2Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
       }
       return;
     case 'UNAUTHENTICATED':
-      router.navigate('signin', { trigger: true });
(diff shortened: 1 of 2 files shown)


gowthamidommety-okta:
Spoke to @srinivasanagandla-okta yesterday. His concern was that when accessing apps, with idp discovery enabled, the state token would not persist. @yuliu-okta @tthompson-okta would this be an issue?

yuliu-okta:
@gowthamidommety-okta:
1. Could you please confirm that the following flow(input user identifier, input password) can use that state token if idp is Okta?
2. For third party idp, when the idp response comes back, we do not know what is the existing state token. But it should not break the flow, it just needs to bootstrap a new state token, is that right? Also we are planning to put the state token into relay state, so that the idp will send the state token back.

gowthamidommety-okta:
@mauriciocastillosilva-okta Added tests. Please review.
@yuliu-okta i'll try testing those flows

PR #367, opened 2017-12-18
Use input with type=tel to automaticaly open numeric keyboard on mobile devices

alexeystadnikov-okta:
screenshot_20171218-181339
screen shot 2017-12-18 at 6 02 56 pm
screen shot 2017-12-29 at 4 06 42 pm

Resolves OKTA-79239

@mauriciocastillosilva-okta 
@eugenebakaiev-okta 
Please, review.

### DIFF
--- a/test/unit/spec/MfaVerify_spec.js
+++ b/test/unit/spec/MfaVerify_spec.js
@@ -801,7 +801,7 @@ function (Okta,
         });
         itp('has a passCode field', function () {
           return setupGoogleTOTP().then(function (test) {
-            expectHasAnswerField(test);
+            expectHasAnswerField(test, 'tel');
           });
         });
         itp('has a masked passCode field for RSA', function () {
@@ -1085,7 +1085,7 @@ function (Okta,
         });
         itp('has a passCode field', function () {
           return setupSMS().then(function (test) {
-            expectHasAnswerField(test);
+            expectHasAnswerField(test, 'tel');
           });
         });
         itp('has remember device checkbox', function () {
@@ -1385,7 +1385,7 @@ function (Okta,
         });
         itp('has a passCode field', function () {
           return setupCall().then(function (test) {
-            expectHasAnswerField(test);
+            expectHasAnswerField(test, 'tel');
           });
         });
         itp('clears the passcode text field on clicking the "Call" button', function () {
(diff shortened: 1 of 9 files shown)


mauriciocastillosilva-okta:
@alexeystadnikov-okta just to confirm, did you try this in iOs and in Android?

test/unit/helpers/util/Expect.js
```diff
@@ -152,9 +152,10 @@ define([
     });
   };
 
-  fn.isTextField = function ($input) {
+  fn.isTextField = function ($input, inputType) {
```
eugenebakaiev-okta: Maybe it would be better to rename this method, as it's purpose changed?

alexeystadnikov-okta: Done. Added another method instead of modification the existing one.

alexeystadnikov-okta:
@mauriciocastillosilva-okta 
Tried this on Android.
Don't have a possibility to test on iOS for the moment.

alexeystadnikov-okta:
Checked on the iOS emulator. Works fine.

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2178,7 +2178,7 @@ function (Okta,
               var form = test.form[1];
               form.inlineTOTPAdd().click();
               expect(form.inlineTOTPAdd().length).toBe(0);
-              Expect.isTextField(form.answerField());
+              Expect.isTelField(form.answerField());
```
mauriciocastillosilva-okta: can we use the same function as in the other tests?  ```expectHasAnswerField```? That way we don't need to implement a new method in ```Expect``` that does the same thing.

alexeystadnikov-okta: Done

test/unit/helpers/util/Expect.js
```diff
@@ -1,4 +1,4 @@
-/* eslint max-statements:[2, 28] */
+/* eslint max-statements:[2, 30] */
```
mauriciocastillosilva-okta: can we revert this? It is probably not needed anymore

alexeystadnikov-okta: Done

PR #369, opened 2017-12-21
:seedling: Send device fingerprint to get security image

mauriciocastillosilva-okta:
Security Image API accepts now the device fingerprint header, which helps fetching the image in devices that have been used before but do not have the proximity cookie stored.
So adding the fingerprint header to the getImage request
- Note: we cannot reuse the same fingerprint twice due to the nonce (it is only valid once), so we generate one for getImage and a different one for primaryAuth

Resolves: OKTA-100304

Screencast:
devicefingerprintingetimage

### DIFF
--- a/src/models/AppState.js
+++ b/src/models/AppState.js
@@ -38,7 +38,7 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
 
   var securityImageUrlTpl = compile('{{baseUrl}}/login/getimage?username={{username}}');
 
-  function getSecurityImage(baseUrl, username) {
+  function getSecurityImage(baseUrl, username, deviceFingerprint) {
     var url = securityImageUrlTpl({ baseUrl: baseUrl, username: username });
 
     // When the username is empty, we want to show the default image.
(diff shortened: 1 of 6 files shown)


test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -689,7 +686,41 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
     });
 
     Expect.describe('Device Fingerprint', function () {
-      itp('does not contain device fingerprint header in primaryAuth if feature is enabled', function () {
+      itp('contains fingerprint header in get security image request if feature is enabled', function () {
```
gowthamidommety-okta: Looks like these two tests will apply to IDPDiscovery as well. Should we add them to IDPDiscovery_spec also?

src/models/AppState.js
```diff
@@ -91,11 +99,11 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
       if (this.settings.get('features.securityImage')) {
         var self = this;
         this.listenTo(this, 'change:username', function (model, username) {
-          getSecurityImage(this.get('baseUrl'), username)
+          getSecurityImage(this.get('baseUrl'), username, model.get('deviceFingerprint'))
```
nikhilvenkatraman-okta: Where does deviceFingerprint get set on the model?

test/unit/spec/PrimaryAuth_spec.js
mauriciocastillosilva-okta: Good point, I'll add them 👍

src/models/AppState.js
mauriciocastillosilva-okta: it is set in ```PrimaryAuthController```, on a few lines that look like ```options.appState.set('username', self.model.get('username'))```.
The model here is the same as ```this```: it is the AppState. So you bring a good point, I should stick to one or the other, so I'll change it to ```this.get('deviceFingerprint')```

PR #370, opened 2017-12-26
Refactor rememberDevice tests to simplify the mock requests

alexeystadnikov-okta:
Removed unnecessary xhr mock files.

@mauriciocastillosilva-okta 
@eugenebakaiev-okta 
Please, review.

Resolves OKTA-93397


test/unit/spec/MfaVerify_spec.js
```diff
@@ -523,36 +514,54 @@ function (Okta,
 
       });
       Expect.describe('Remember device', function () {
+
+        function getResMfaPolicy(options) {
```
mauriciocastillosilva-okta: can we create a ```setupWithMfaPolicy``` function at the top describe level? That seems more consistent with the general approach we've taken for scenarios like this.
And the calls to it will be something like
```
return setupWithMfaPolicy({ minutes: 0 }).then(function (test) {
```
 or
```
return setupWithMfaPolicy({ minutes: 15, byDefault: true }).then(function (test) {
```

alexeystadnikov-okta: Done

test/unit/spec/MfaVerify_spec.js
```diff
@@ -523,36 +514,54 @@ function (Okta,
 
       });
       Expect.describe('Remember device', function () {
+
+        function setupWithMfaPolicy(options) {
```
mauriciocastillosilva-okta: can we move this to somewhere in between lines 181-221? (probably after line 221) This is so all the setup functions are at the top of the test, so if we need to reuse them then they are easier to find

alexeystadnikov-okta: Done

mauriciocastillosilva-okta:
@alexeystadnikov-okta this branch keeps failing in Travis, can you rebase and also look into what the problem is?

PR #373, opened 2018-01-02
:white_check_mark: Add okta-angular tests

lboyette-okta:


package.json
```diff
@@ -27,8 +27,11 @@
     "install:react": "npm install --prefix test/e2e/react-app",
     "prestart:react": "kill -9 $(lsof -t -i:3001 -sTCP:LISTEN) || true",
     "start:react": "npm start --prefix test/e2e/react-app & wait-on http-get://localhost:3001",
-    "pretest:e2e": "npm run install:react && node pretest.js",
-    "test:e2e": "npm run start:react && grunt test-e2e",
+    "install:angular": "npm install --prefix test/e2e/angular-app",
+    "prestart:angular": "kill -9 $(lsof -t -i:4200 -sTCP:LISTEN) || true",
```
jmelberg-okta: Would it make more sense to include the port kill on test completion?

test/e2e/angular-app/prestart.js
```diff
@@ -0,0 +1,13 @@
...
+const target = path.join(__dirname, 'src/environments/environment.ts');
+fs.ensureFileSync(target);
+fs.writeFileSync(target, `
+
+export const environment = {
+  production: true,
+  WIDGET_TEST_SERVER: '${process.env.WIDGET_TEST_SERVER}',
+};
+
+`);
```
jmelberg-okta: Is this needed?

test/e2e/angular-app/src/app/app.component.html
```diff
@@ -0,0 +1,17 @@
+<!-- src/app/app.component.html -->
+
+<link
+href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/2.5.0/css/okta-sign-in.min.css"
```
jmelberg-okta: Not needed for this PR, but we should figure out a good way to pull the version from `package.json` instead of manually updating it every time.

test/e2e/angular-app/src/app/login.component.ts
```diff
@@ -0,0 +1,54 @@
+// src/app/login.component.ts
+
+import { Component, OnInit } from '@angular/core';
+import { Router, NavigationStart} from '@angular/router';
+
+import { OktaAuthService } from '@okta/okta-angular';
+import * as OktaSignIn from '@okta/okta-signin-widget';
```
jmelberg-okta: Hmm. I thought we fixed this with `2.5`. Can we not import the widget via:

```typescript
import OktaSignIn from '@okta/okta-signin-widget';
```

package.json
lboyette-okta: It's possible that they would've started something else on the port we want to use. So, I decided to kill the port *before* I use it, rather than after

test/e2e/angular-app/src/app/app.component.html
robertjd: This can be pulled in from node modules using the `.angular-cli.json` file, see example here: https://github.com/okta/samples-js-angular/blob/first-pass/custom-login/.angular-cli.json#L24

test/e2e/angular-app/prestart.js
lboyette-okta: The `prestart.js` file is needed, because you can't pull environment variables from your angular js files. Anything in your `src` directory counts, so I have to create the environment file before I build.

test/e2e/angular-app/src/app/app.component.html
lboyette-okta: I agree

test/e2e/angular-app/src/app/login.component.ts
lboyette-okta: You can do that in our create-react-app example, because it's a bit looser with how it handles ES6 modules. TypeScript is strict, so you must import like:
`import * as OktaSignIn from '@okta/okta-signin-widget';`

We fixed the need to reference dist:
`import * as OktaSignIn from '@okta/okta-signin-widget/dist/js/okta-sign-in.min.js';`

PR #375, opened 2018-01-04
:fire: Remove handlebars 2.x from Okta sign-in widget

manueltanzi-okta:
Removing from sign in widget handlebars v2.0 imported from Courage.
Sign in widget is already using the 4.x series of Handlebars.
Long term solution will be to copy from Courage only those libraries that are needed by okta-signin-widget.

@hor-kanchan-okta @matiasblanco-okta


matiasblanco-okta:
LGTM

PR #377, opened 2018-01-10
:seedling: Adds support for chains of MFA_REQUIRED responses

gowthamidommety-okta:
Resolves: OKTA-152882

### DIFF
--- a/src/MfaVerifyController.js
+++ b/src/MfaVerifyController.js
@@ -118,11 +118,10 @@ function (Okta, Checkbox, BaseLoginController, CookieUtil, TOTPForm, YubikeyForm
     },
 
     trapAuthResponse: function () {
-      if(this.options.factorType === 'password') {
-        return false;
-      }
-      if (this.options.appState.get('isMfaChallenge') ||
-          this.options.appState.get('isMfaRequired')) {
+      if((this.options.appState.get('isMfaRequired') &&
+          this.options.appState.get('trapMfaRequiredResponse')) ||
+          this.options.appState.get('isMfaChallenge')) {
+        this.options.appState.set('trapMfaRequiredResponse', false);
         return true;
       }
       // update auto push cookie after user accepts Okta Verify MFA
--- a/src/views/mfa-verify/InlineTOTPForm.js
+++ b/src/views/mfa-verify/InlineTOTPForm.js
@@ -37,6 +37,7 @@ define(['okta', 'views/shared/TextBox'], function (Okta, TextBox) {
           // and then call model.save(). If not, we would still be in MFA_CHALLENGE state and
           // verify would result in a wrong request (push verify instead of a TOTP verify).
           if (transaction.status === 'MFA_CHALLENGE' && transaction.prev) {
+            form.options.appState.set('trapMfaRequiredResponse', true);
             return transaction.prev().then(function (trans) {
               setTransaction(trans);
               form.model.save();
--- a/test/unit/spec/MfaVerify_spec.js
+++ b/test/unit/spec/MfaVerify_spec.js
@@ -1,4 +1,4 @@
-/* eslint max-params: [2, 50], max-statements: [2, 41], camelcase: 0 */
+/* eslint max-params: [2, 50], max-statements: [2, 42], camelcase: 0 */
 define([
   'okta',
   'vendor/lib/q',
@@ -2227,6 +2227,7 @@ function (Okta,
               var pushForm = test.form[0],
                   inlineForm = test.form[1];
               return setupPolling(test, resRejectedPush)
+              .then(function () { return tick(test); }) // Final response - REJECTED
               .then(function () {
                 return Expect.waitForFormError(pushForm, {
                   test: test,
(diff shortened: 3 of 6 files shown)


src/MfaVerifyController.js
```diff
@@ -118,11 +118,10 @@ function (Okta, Checkbox, BaseLoginController, CookieUtil, TOTPForm, YubikeyForm
...
     trapAuthResponse: function () {
-      if(this.options.factorType === 'password') {
-        return false;
-      }
-      if (this.options.appState.get('isMfaChallenge') ||
-          this.options.appState.get('isMfaRequired')) {
+      if((this.options.appState.get('isMfaRequired') &&
+          this.options.appState.get('trapMfaRequiredResponse')) ||
+          this.options.appState.get('isMfaChallenge')) {
+        this.options.appState.set('trapMfaRequiredResponse', false);
```
mauriciocastillosilva-okta: I would like to understand this a little more... so before if we were in MFA_REQUIRED or MFA_CHALLENGE, we `trap` the response, meaning we don't let ```RouterUtil``` navigate somewhere else.

With this change, we will trap the reponse when we are either in MFA_CHALLENGE or we are in MFA_REQUIRED and we explicitly set a variable to trap it. This means that if we don't set the variable, we will let RouterUtil navigate to the next view. Is this correct? If so, then my next question is for your changes in ```InlineTOTPForm.js```, why is it setting ```trapMfaRequiredResponse``` to true?

And well, I think in general the idea is (correct me if I am missing something)
- There are 2 flows that come here: from the factors dropdown or from an api response.
- If we interact with the dropdown, we should always trap the response, no matter what status we are in.
- If we get a response that says MFA_REQUIRED, then we should always navigate to a new view (with the exception of the above item involving a dropdown)

src/views/mfa-verify/InlineTOTPForm.js
```diff
@@ -31,6 +31,7 @@ define(['okta', 'views/shared/TextBox'], function (Okta, TextBox) {
         if (!form.isValid()) {
           return;
         }
+        form.options.appState.set('trapMfaRequiredResponse', true);
```
mauriciocastillosilva-okta: I think we should trap the response only if we are in MFA_CHALLENGE because of the ```previous``` call we need to make. 
If we are in MFA_REQUIRED and we type the code and press Verify, that will make a request to the backend, and if the response is MFA_REQUIRED again, we should let RouterUtil do its magic.

test/unit/spec/MfaVerify_spec.js
```diff
@@ -3172,6 +3175,59 @@ function (Okta,
...
+        })
+        .then(function (test) {
+          test.setNextResponse(resAllFactors);
+          test.beacon.dropDownButton().click();
+          clickFactorInDropdown(test, 'QUESTION');
+          return Expect.waitForVerifyQuestion(test);
+        })
+        .then(function () {
+          // trapAuthResponse prevents handleResponseStatus from being called
+          expect(RouterUtil.handleResponseStatus.calls.count()).toBe(0);
```
mauriciocastillosilva-okta: the other way to do this is that we check the ajax calls we make here. There should be only 1, which is the ```previous``` link.

test/unit/spec/MfaVerify_spec.js
```diff
@@ -3172,6 +3175,59 @@ function (Okta,
...
+        });
+      });
+      itp('Verify inline totp with factor Okta Verify Push traps auth response', function () {
+        return setupOktaPush().then(function (test) {
+          return setupPolling(test, resRejectedPush)
+          .then(function () {
+            spyOn(RouterUtil, 'handleResponseStatus').and.callThrough();
+            test.totpForm = new MfaVerifyForm($($sandbox.find('.o-form')[1]));
+            test.totpForm.inlineTOTPAdd().click();
```
mauriciocastillosilva-okta: clicking this link does not make any call or anything right? It only shows the textbox for the user to manually input the code, so not sure this test is needed.

gowthamidommety-okta: Changed this test to call verify

test/unit/spec/MfaVerify_spec.js
gowthamidommety-okta: Spoke to Mau about this offline. ajax calls may not be accurate here.

src/MfaVerifyController.js
gowthamidommety-okta: Spoke to Mau about this offline. The general idea above is correct. Made changes to set trapMfaRequiredResponse only when moving from MFA_CHALLENGE to MFA_REQUIRED state

test/unit/spec/MfaVerify_spec.js
```diff
@@ -3170,6 +3177,62 @@ function (Okta,
       });
     });
 
+    Expect.describe('Supports chains of MFA_REQUIRED responses', function () {
```
mauriciocastillosilva-okta: nit: this should be something like ```MFA_REQUIRED to MFA_REQUIRED```
and the first ```itp```, something like ```renders a new view when we successfully answer the first MFA_REQUIRED response``` or so....
in the end, this will read the whole sentence together, so ```MFA_REQUIRED to MFA_REQUIRED renders a new view when we successfully answer the first MFA_REQUIRED response```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -3170,6 +3177,62 @@ function (Okta,
...
+        return setupOktaPush().then(function (test) {
+          return setupPolling(test, resRejectedPush)
+          .then(function () {
+            spyOn(RouterUtil, 'handleResponseStatus').and.callThrough();
+            test.totpForm = new MfaVerifyForm($($sandbox.find('.o-form')[1]));
+            test.totpForm.inlineTOTPAdd().click();
+            test.totpForm.setAnswer('654321');
+            test.setNextResponse(resAllFactors, resSuccess);
+            test.totpForm.inlineTOTPVerify().click();
+            return tick(test);
```
mauriciocastillosilva-okta: can we change this to an expect for the success spy?

test/unit/spec/MfaVerify_spec.js
gowthamidommety-okta: Hey Mau, not all the tests in this group are MFA_REQUIRED to MFA_REQUIRED. The first tests the password reauth scenario, the second tests MFA_CHALLENGE to MFA_REQUIRED using the drop down and the third tests the inline totp scenario. It is basically testing the main scenarios wrt my changes.

gowthamidommety-okta:
Tested 

- ra-auth (password follow by mfa)
- refresh the page on challenge (choose sms, send the code and then click browser refresh)

PR #383, opened 2018-01-23
:bug: Fix unit tests to run in browser

alexeystadnikov-okta:
screen shot 2018-01-23 at 6 43 52 pm

Resolves OKTA-153354

@mauriciocastillosilva-okta 
Please, review.


test/unit/helpers/dom/EnrollSmsForm.js
```diff
@@ -18,6 +18,10 @@ define(['./Form'], function (Form) {
       return this.selectOption(COUNTRIES_FIELD, countryCode);
     },
 
+    hasCoutriesList: function () {
```
lboyette-okta: `hasCoutriesList` => `hasCountriesList`

alexeystadnikov-okta: Done.
Thanks

mauriciocastillosilva-okta:
@alexeystadnikov-okta Travis is failing, there is one test that does not work 
```
DeviceFingerprint returns a fingerprint if the communication with the iframe is successfull
   Failed: Fingerprint promise incorrectly failed
```
Can you look into that please?

test/unit/spec/DeviceFingerprint_spec.js
```diff
@@ -42,7 +42,7 @@ function ($, Q, Expect, $sandbox, DeviceFingerprint) {
 
     it('returns a fingerprint if the communication with the iframe is successfull', function (done) {
```
mauriciocastillosilva-okta: this test is failing in Travis, take a look please. It might be that the baseurl is not the same.
If you need more ideas about fingerprint tests, Len just added a new [test spec in auth js](https://github.com/okta/okta-auth-js/blob/master/test/spec/fingerprint.js) that you can look at

test/unit/helpers/dom/EnrollTotpBarcodeForm.js
```diff
@@ -68,6 +68,14 @@ define(['./Form'], function (Form) {
 
     clickRefreshLink: function () {
       return this.refreshLink().click();
+    },
+
+    waitForManualSetupLink: function (resolveValue) {
```
mauriciocastillosilva-okta: I can't find where this function is used. If it is not used, please remove it.

test/unit/spec/EnrollTotpController_spec.js
```diff
@@ -697,6 +700,12 @@ function (_, $, Q, OktaAuth, LoginUtil, StringUtil, Util, DeviceTypeForm, Barcod
         });
         itp('has correct fields displayed when different dropdown options selected', function () {
           return enrollOktaPushGoCannotScan()
+          .then(function (test) {
+            return test.manualSetupForm.waitForDropdownElement(test);
+          })
+          .then(function (test) {
+            return test.manualSetupForm.waitForCountryCodeSelect(test);
```
mauriciocastillosilva-okta: do we need both ```waitFor``` calls? I think if the page has a form with a dropdown and a country select, if we wait for one it should be enough.

test/unit/spec/EnrollTotpController_spec.js
```diff
@@ -927,6 +936,7 @@ function (_, $, Q, OktaAuth, LoginUtil, StringUtil, Util, DeviceTypeForm, Barcod
             test.passCodeForm.backLink().click();
             return Expect.waitForManualSetupPush(test);
           })
+          .then(tick)
```
mauriciocastillosilva-okta: what is this ```tick``` for? If we are waiting for the list of factors to enroll, it might be better to use a wait for a specific controller instead. I think we have already that wait: ```waitForEnrollChoices```

test/unit/spec/EnrollWindowsHello_spec.js
```diff
@@ -228,6 +228,7 @@ function (Okta,
           test.form.submit();
           return Expect.waitForSpyCall(webauthn.makeCredential, test);
         })
+        .then(tick)
```
mauriciocastillosilva-okta: same here, we should wait for something more specific. This test seems to be waiting for the spy call, if that is not enough, then let's try to wait for something else, like a controller

test/unit/spec/LoginRouter_spec.js
```diff
@@ -1400,6 +1405,7 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, Logger, OktaAut
             test.router.enrollCall();
             return Expect.waitForEnrollCall(test);
           })
+          .then(tick)
```
mauriciocastillosilva-okta: do we need this ```tick```? we are already waiting for an enroll call. What happens here that we need this?

test/unit/spec/EnrollTotpController_spec.js
alexeystadnikov-okta: We should wait for the Chosen plugin being initialized on both elements.
Sometimes in a browser, it can get not in time.
Have to be sure, that Chosen is ready for both dropdowns.

test/unit/spec/LoginRouter_spec.js
alexeystadnikov-okta: This tick it to give Chosen time to update its UI elements.

test/unit/helpers/dom/EnrollTotpBarcodeForm.js
alexeystadnikov-okta: Removed

test/unit/spec/DeviceFingerprint_spec.js
alexeystadnikov-okta: Fixed

test/unit/spec/EnrollWindowsHello_spec.js
alexeystadnikov-okta: Removed

test/unit/spec/EnrollTotpController_spec.js
alexeystadnikov-okta: Replaced with the waitFor...

test/unit/spec/LoginRouter_spec.js
mauriciocastillosilva-okta: should this then be something like your ```waitForCountryCodeSelect```?

alexeystadnikov-okta: In that test, we check that the value of the select was updated.

Only thing I can `waitFor` is the specific value for the `this.inputWrap(field).find('.chzn-single span').trimmedText()`, which is the same as we expect in the`expect(test.form.selectedCountry()).toBe(countryText);`.

Should I replace the `expect(...selectedCountry()).toBe(countryText)` with the `waitFor` logic?

src/util/DeviceFingerprint.js
```diff
@@ -47,7 +47,7 @@ define(['vendor/lib/q', 'okta/jquery'], function (Q, $) {
 
       function onMessageReceivedFromOkta(event) {
         if (!event || !event.data || event.origin != oktaDomainUrl) {
-          handleError('no data');
+          handleError(event.origin + ' !== ' + oktaDomainUrl);
```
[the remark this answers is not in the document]
alexeystadnikov-okta: Reverted

test/unit/spec/LoginRouter_spec.js
mauriciocastillosilva-okta: ok, it is fine then, just add a comment please
```.then(tick) // Wait for chosen items to render```

PR #388, opened 2018-01-31
:arrow_up: Update i18n files to v41.3.0

manueltanzi-okta:
Resolves: [OKTA-155781]()

@hor-kanchan-okta @samerfanek-okta


packages/@okta/i18n/package.json
```diff
@@ -28,8 +28,8 @@
     "debug": "2.6.9",
     "decamelize": "1.2.0",
     "domelementtype": "1.3.0",
-    "domhandler": "2.3.0",
-    "domutils": "1.5.1",
+    "domhandler": "2.4.1",
+    "domutils": "1.6.2",
```
samerfanek-okta: Are we okay with updating these (and the one on line 59)? Doesn't look related to i18n.

manueltanzi-okta: We are okay, checked with @mauriciocastillosilva-okta

PR #395, opened 2018-02-12
:seedling: Support http only cookie for autoPush

gowthamidommety-okta:
Removes support for autoPush as js readable cookie.
Note: Users will need to re-enroll in auto push with this change

Resolves: OKTA-157892

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,13 +1,13 @@
 {
   "name": "@okta/okta-signin-widget",
-  "version": "2.5.0",
+  "version": "2.6.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
     "@okta/okta-auth-js": {
-      "version": "1.11.0",
-      "resolved": "https://registry.npmjs.org/@okta/okta-auth-js/-/okta-auth-js-1.11.0.tgz",
-      "integrity": "sha1-k6/QIE1w1vasBCBmfa/Ff156j/4=",
+      "version": "1.13.0",
+      "resolved": "https://registry.npmjs.org/@okta/okta-auth-js/-/okta-auth-js-1.13.0.tgz",
+      "integrity": "sha1-9KX3Oku+MquqULuYWkp0rjnHYkg=",
       "requires": {
         "Base64": "0.3.0",
         "q": "1.4.1",
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "webpack": "1.13.1"
   },
   "dependencies": {
-    "@okta/okta-auth-js": "1.11.0",
+    "@okta/okta-auth-js": "1.13.0",
     "babel-polyfill": "^6.23.0",
     "babel-runtime": "^6.26.0",
     "backbone": "1.2.1",
--- a/src/models/AppState.js
+++ b/src/models/AppState.js
@@ -586,6 +586,12 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         fn: function (policy) {
           return policy && policy.rememberDeviceByDefault;
         }
+      },
+      'factorsPolicyInfo' : {
+        deps: ['policy'],
+        fn: function (policy) {
+          return (policy && policy.factorsPolicyInfo) ? policy.factorsPolicyInfo: null;
+        }
       }
     },
 
--- a/src/models/Factor.js
+++ b/src/models/Factor.js
@@ -198,6 +198,7 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
     save: function () {
       var rememberDevice = !!this.get('rememberDevice');
       // Set/Remove the remember device cookie based on the remember device input.
+      var self = this;
 
       return this.doTransaction(function (transaction) {
         var data = {
@@ -211,6 +212,10 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
         }
         else {
           data.passCode = this.get('answer');
+        }
+
+        if (this.pushFactorHasAutoPush()) {
+          data.autoPush = this.get('autoPush');
         }
 
         var promise;
@@ -235,9 +240,17 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
 
         return promise
         .then(function (trans) {
+          var options = {
+            'delay': PUSH_INTERVAL
+          };
           if (trans.status === 'MFA_CHALLENGE' && trans.poll) {
             return Q.delay(PUSH_INTERVAL).then(function() {
-              return trans.poll(PUSH_INTERVAL);
+              if (self.pushFactorHasAutoPush()) {
+                options.autoPush = function() {
+                  return self.get('autoPush');
+                };
+              }
+              return trans.poll(options);
             });
           }
           return trans;
@@ -247,6 +260,10 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
           throw err;
         });
       });
+    },
+
+    pushFactorHasAutoPush: function() {
+      return this.settings.get('features.autoPush') && this.get('factorType') === 'push';
     }
   });
 
--- a/src/views/mfa-verify/PushForm.js
+++ b/src/views/mfa-verify/PushForm.js
@@ -87,7 +87,12 @@ define(['okta', 'util/CookieUtil', 'util/Util'], function (Okta, CookieUtil, Uti
       }
     },
     postRender: function() {
-      if (this.settings.get('features.autoPush') && CookieUtil.isAutoPushEnabled(this.options.appState.get('userId'))) {
+      var factorsPolicyInfo = this.options.appState.get('factorsPolicyInfo');
+      var id = this.model.get('id');
+      var isAutoPushEnabled = (this.settings.get('features.autoPush') && factorsPolicyInfo &&
+                              factorsPolicyInfo[id]) ? factorsPolicyInfo[id]['autoPushEnabled'] : false;
+
+      if (isAutoPushEnabled) {
         this.model.set('autoPush', true);
         // bind after $el has been rendered, and trigger push once DOM is fully loaded
         _.defer(_.bind(this.submit, this));
--- a/test/unit/spec/LoginRouter_spec.js
+++ b/test/unit/spec/LoginRouter_spec.js
@@ -1,11 +1,10 @@
-/* eslint max-params: [2, 32], max-statements: [2, 43], max-len: [2, 180], camelcase:0 */
+/* eslint max-params: [2, 32], max-statements: [2, 44], max-len: [2, 180], camelcase:0 */
 define([
   'okta',
   'vendor/lib/q',
   'backbone',
   'shared/util/Util',
   'util/CryptoUtil',
-  'util/CookieUtil',
   'util/Logger',
   '@okta/okta-auth-js/jquery',
   'helpers/mocks/Util',
@@ -33,7 +32,7 @@ define([
   'helpers/xhr/labels_login_ja',
   'helpers/xhr/labels_country_ja'
 ],
-function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, Logger, OktaAuth, Util, Expect, Router,
+function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Expect, Router,
           $sandbox, PrimaryAuthForm, IDPDiscoveryForm, RecoveryForm, MfaVerifyForm, EnrollCallForm, resSuccess, resRecovery,
           resMfa, resMfaRequiredDuo, resMfaRequiredOktaVerify, resMfaChallengeDuo, resMfaChallengePush,
           resMfaEnroll, errorInvalidToken, resUnauthenticated, resSuccessStepUp, Errors, BrowserFeatures,
@@ -568,8 +567,7 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, Logger, OktaAut
         expect(form.isSecurityQuestion()).toBe(true);
       });
     });
-    itp('checks auto push by default for a returning user', function () {
-      Util.mockCookie('auto_push_' + CryptoUtil.getStringHash('00uhn6dAGR4nUB4iY0g3'), 'true');
+    itp('checks auto push by default for a returning user with autoPush true', function () {
       return setup({'features.autoPush': true})
       .then(function (test) {
         Util.mockRouterNavigate(test.router);
@@ -581,7 +579,8 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, Logger, OktaAut
         expect(form.isPrimaryAuth()).toBe(true);
         // Respond with MFA_REQUIRED
         // Verify is immediately called, so respond with MFA_CHALLENGE
-        test.setNextResponse([resMfaRequiredOktaVerify, resMfaChallengePush]);
+        var resAutoPushTrue = Util.getAutoPushResponse(resMfaRequiredOktaVerify, true);
+        test.setNextResponse([resAutoPushTrue, resMfaChallengePush]);
         form.setUsername('testuser');
         form.setPassword('pass');
         form.submit();
@@ -594,7 +593,7 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, Logger, OktaAut
         expect(form.isPushSent()).toBe(true);
         expect($.ajax.calls.count()).toBe(2);
         Expect.isJsonPost($.ajax.calls.argsFor(1), {
-          url: 'https://foo.com/api/v1/authn/factors/opfhw7v2OnxKpftO40g3/verify',
+          url: 'https://foo.com/api/v1/authn/factors/opfhw7v2OnxKpftO40g3/verify?autoPush=true',
           data: {
             stateToken: 'testStateToken'
           }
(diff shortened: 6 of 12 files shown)


src/models/Factor.js
```diff
@@ -235,9 +240,17 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
...
         .then(function (trans) {
+          var options = {
+            'delay': PUSH_INTERVAL
+          };
           if (trans.status === 'MFA_CHALLENGE' && trans.poll) {
             return Q.delay(PUSH_INTERVAL).then(function() {
-              return trans.poll(PUSH_INTERVAL);
+              if (self.settings.get('features.autoPush') && self.get('factorType') === 'push') {
+                options.autoPush = _.bind(function() {
+                  return self.get('autoPush');
```
mauriciocastillosilva-okta: shouldn't this be ```return this.get('autoPush');```? Or then I don't think we need to bind it...

src/views/mfa-verify/PushForm.js
```diff
@@ -87,7 +87,12 @@ define(['okta', 'util/CookieUtil', 'util/Util'], function (Okta, CookieUtil, Uti
       }
     },
     postRender: function() {
-      if (this.settings.get('features.autoPush') && CookieUtil.isAutoPushEnabled(this.options.appState.get('userId'))) {
+      var factorsPolicyInfo = this.options.appState.get('factorsPolicyInfo');
+      var id = this.model.get('id');
+      var isAutoPushEnabled = (factorsPolicyInfo && factorsPolicyInfo[id]) ?
+                                factorsPolicyInfo[id]['autoPushEnabled'] : false;
+
+      if (this.settings.get('features.autoPush') && isAutoPushEnabled) {
```
mauriciocastillosilva-okta: nit: checking if the feature is enabled before checking the policy
```
var isAutoPushEnabled = false;
if (this.settings.get('features.autoPush') && factorsPolicyInfo && factorsPolicyInfo[id]) {
  isAutoPushEnabled = factorsPolicyInfo[id]['autoPushEnabled'];
}
if (isAutoPushEnabled) {
...
```

test/unit/helpers/xhr/MFA_REQUIRED_allFactors_autoPushTrue.js
```diff
@@ -0,0 +1,279 @@
+define({
```
mauriciocastillosilva-okta: no need to create this file. We should do some function in the tests where we change the json we have in ```MFA_REQUIRED_allFactors.js```, similar to other tests like [this one](https://github.com/okta/okta-signin-widget/blob/master/test/unit/spec/EnrollOnPrem_spec.js#L66-L72)

test/unit/spec/LoginRouter_spec.js
```diff
@@ -581,7 +580,7 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, Logger, OktaAut
         expect(form.isPrimaryAuth()).toBe(true);
         // Respond with MFA_REQUIRED
         // Verify is immediately called, so respond with MFA_CHALLENGE
-        test.setNextResponse([resMfaRequiredOktaVerify, resMfaChallengePush]);
+        test.setNextResponse([resMfaRequiredOktaVerifyAutoPushTrue, resMfaChallengePush]);
```
mauriciocastillosilva-okta: following my other comment, here instead of using ```resMfaRequiredOktaVerifyAutoPushTrue```, just call a function like ```getAutoPushResponse(resMfaRequiredOktaVerify, true)``` that creates the response you need
```
function getAutoPushResponse(response, autoPushEnabled) {
  // clone the response and change the value you need
  var autoPushResponse = _.clone(response);
  autoPushResponse._embedded.policy.factorsPolicyInfo.autoPush = autoPushEnabled;
  return autoPushResponse;
}
```

gowthamidommety-okta: I tried that first, but the autoPushEnabled property is copied by reference when using _.clone. So subsequent tests fail.. If creating another xhr file is not an option, would it make sense to set this property for the test and reset it at the end?

mauriciocastillosilva-okta: What about something like [deepClone](https://github.com/okta/okta-signin-widget/blob/653ceac778903b8189003802f597b9ba9855a552/test/unit/spec/PasswordExpired_spec.js#L30)?

gowthamidommety-okta: That should work! Thanks!

src/models/Factor.js
gowthamidommety-okta: yes, that's right. Thanks for catching this

test/unit/spec/LoginRouter_spec.js
```diff
@@ -33,10 +32,10 @@ define([
   'helpers/xhr/labels_login_ja',
   'helpers/xhr/labels_country_ja'
 ],
-function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, Logger, OktaAuth, Util, Expect, Router,
+function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Expect, Router,
           $sandbox, PrimaryAuthForm, IDPDiscoveryForm, RecoveryForm, MfaVerifyForm, EnrollCallForm, resSuccess, resRecovery,
-          resMfa, resMfaRequiredDuo, resMfaRequiredOktaVerify, resMfaChallengeDuo, resMfaChallengePush,
-          resMfaEnroll, errorInvalidToken, resUnauthenticated, resSuccessStepUp, Errors, BrowserFeatures,
+          resMfa, resMfaRequiredDuo, resMfaRequiredOktaVerify, resMfaChallengeDuo,
+          resMfaChallengePush, resMfaEnroll, errorInvalidToken, resUnauthenticated, resSuccessStepUp, Errors, BrowserFeatures,
```
mauriciocastillosilva-okta: nit: no need to change this line...

src/models/Factor.js
```diff
@@ -213,6 +214,10 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
           data.passCode = this.get('answer');
         }
 
+        if (this.settings.get('features.autoPush') && this.get('factorType') === 'push') {
```
santhoshbalakrishnan: nit: move this into it's own fn, since it's used more than once?

src/models/Factor.js
```diff
@@ -235,9 +240,17 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
...
         return promise
         .then(function (trans) {
+          var options = {
+            'delay': PUSH_INTERVAL
+          };
           if (trans.status === 'MFA_CHALLENGE' && trans.poll) {
             return Q.delay(PUSH_INTERVAL).then(function() {
-              return trans.poll(PUSH_INTERVAL);
+              if (self.settings.get('features.autoPush') && self.get('factorType') === 'push') {
+                options.autoPush = function() {
```
santhoshbalakrishnan: Should this be a fn here?

gowthamidommety-okta: Yea we are implementing this as a function so if the user changes the checkbox value during an active poll, it will catch that change and update the backend. - https://github.com/okta/okta-auth-js/pull/91/files#diff-1a136b779b30ddb4520bd33f89a5c8afR81

src/models/AppState.js
```diff
@@ -586,6 +586,15 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         fn: function (policy) {
           return policy && policy.rememberDeviceByDefault;
         }
+      },
+      'factorsPolicyInfo' : {
+        deps: ['policy'],
+        fn: function (policy) {
+          if(policy && policy.factorsPolicyInfo) {
```
santhoshbalakrishnan: Nit: Can just be something like: return ```(policy && policy.factorsPolicyInfo) ? policy.factorsPolicyInfo: false;```

santhoshbalakrishnan: Also, since `factorsPolicyInfo` is an object, maybe better to return a `null` rather than `false` here in case it does not exist?

src/models/AppState.js
```diff
@@ -586,6 +586,15 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         fn: function (policy) {
           return policy && policy.rememberDeviceByDefault;
         }
+      },
+      'factorsPolicyInfo' : {
```
santhoshbalakrishnan: Nit: It seems like only `autoPushEnabled` is used from this `factorsPolicyInfo` object. If there was no intention to add anything more to this object, maybe just derive `autoPushEnabled` in the model as opposed to the full object?

gowthamidommety-okta: It might be used to pass in information of other factors in the future. Also autoPushEnabled is dependent on the factorID, and appState is loaded before okta-verify is, so I was not able to get the correct value of autoPushEnabled in time

package-lock.json
```diff
@@ -1,13 +1,13 @@
...
+  "version": "2.6.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
     "@okta/okta-auth-js": {
-      "version": "1.11.0",
-      "resolved": "https://registry.npmjs.org/@okta/okta-auth-js/-/okta-auth-js-1.11.0.tgz",
-      "integrity": "sha1-k6/QIE1w1vasBCBmfa/Ff156j/4=",
+      "version": "1.13.0",
+      "resolved": "https://artifacts.aue1d.saasure.com:443/artifactory/api/npm/npm-okta/@okta/okta-auth-js/-/@okta/okta-auth-js-1.13.0.tgz",
```
mauriciocastillosilva-okta: This has to be the public npm registry, otherwise people that do not have access to our internal artifactory won't be able to download auth-js

package.json
```diff
@@ -89,7 +89,7 @@
     "webpack": "1.13.1"
   },
   "dependencies": {
-    "@okta/okta-auth-js": "1.11.0",
+    "@okta/okta-auth-js": "^1.13.0",
```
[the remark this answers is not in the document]
mauriciocastillosilva-okta: good catch. Yeah, let's have an specific version

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1994,6 +2012,24 @@ function (Okta,
               });
             });
           });
+          itp('calls authClient verifyFactor with correct args for autoPush when submitted', function () {
+            return setupOktaPush({'features.autoPush': true}).then(function (test) {
+              $.ajax.calls.reset();
+              setAutoPushCheckbox(test, true);
```
mauriciocastillosilva-okta: do we have one test for when autoPush is not checked?

gowthamidommety-okta: No, i didn't add one because it is covered in auth-js. I'll add them now.

src/views/mfa-verify/PushForm.js
```diff
@@ -87,7 +87,13 @@ define(['okta', 'util/CookieUtil', 'util/Util'], function (Okta, CookieUtil, Uti
       }
     },
     postRender: function() {
-      if (this.settings.get('features.autoPush') && CookieUtil.isAutoPushEnabled(this.options.appState.get('userId'))) {
+      var factorsPolicyInfo = this.options.appState.get('factorsPolicyInfo');
+      var id = this.model.get('id');
+      var isAutoPushEnabled = (this.settings.get('features.autoPush') && factorsPolicyInfo &&
+                              factorsPolicyInfo[id]) ?
+                                factorsPolicyInfo[id]['autoPushEnabled'] : false;
```
mauriciocastillosilva-okta: nit: spacing here is a little weird. Can we move this line to the previous one?
```
var isAutoPushEnabled = (this.settings.get('features.autoPush') && factorsPolicyInfo &&
                        factorsPolicyInfo[id]) ? factorsPolicyInfo[id]['autoPushEnabled'] : false;
```

test/unit/spec/LoginRouter_spec.js
```diff
@@ -175,6 +174,17 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, Logger, OktaAut
       expect(Logger.warn).toHaveBeenCalledWith('Field not defined in schema', arg1);
     }
 
+    var getAutoPushResponse = function (response, factorType, provider, autoPushVal) {
```
mauriciocastillosilva-okta: looking at this again, do we really need the 2nd and 3rd params? They seem to be always ```'push', 'OKTA'```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1994,6 +2012,42 @@ function (Okta,
               });
             });
           });
+          itp('calls authClient verifyFactor with correct args for autoPush when submitted', function () {
```
mauriciocastillosilva-okta: description should probably say auto push is true, something like ```calls authClient verifyFactor with correct args when autoPush is checked```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1994,6 +2012,42 @@ function (Okta,
...
+              expect($.ajax.calls.count()).toBe(1);
+              Expect.isJsonPost($.ajax.calls.argsFor(0), {
+                url: 'https://foo.com/api/v1/authn/factors/opfhw7v2OnxKpftO40g3/verify?autoPush=true',
+                data: {
+                  stateToken: 'testStateToken'
+                }
+              });
+            });
+          });
+          itp('calls authClient verifyFactor with correct args for autoPush when submitted', function () {
```
mauriciocastillosilva-okta: description should probably say auto push is false, something like ```calls authClient verifyFactor with correct args when autoPush is not checked```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -214,8 +212,23 @@ function (Okta,
       return setup(res);
     }
 
+    var getAutoPushResponse = function (response, factorType, provider, autoPushVal) {
```
mauriciocastillosilva-okta: same comment as in the other spec, it seems to me we don't need the 2nd and 3rd params.
```
var getAutoPushResponse = function (response, autoPushVal) {
```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -214,8 +212,23 @@ function (Okta,
...
+
+      policy['factorsPolicyInfo'][id]['autoPushEnabled'] = autoPushVal;
+      return responseCopy;
+    };
+
+    var autoPushTrueResponse =  getAutoPushResponse(resAllFactors, 'push', 'OKTA', true);
+
     var setupSecurityQuestion = _.partial(setup, resAllFactors, { factorType: 'question' });
     var setupGoogleTOTP = _.partial(setup, resAllFactors, { factorType: 'token:software:totp', provider: 'GOOGLE' });
+    var setupGoogleTOTPAutoPushTrue = _.partial(setup, autoPushTrueResponse,
```
mauriciocastillosilva-okta: we could avoid a var here and just do
```
 var setupGoogleTOTPAutoPushTrue = _.partial(setup, getAutoPushResponse(resAllFactors, true),
```
The reason I mention this is because the name of the var ```autoPushTrueResponse``` was confusing to me. We could also change that var to something like ```resAllFactorsAutoPushTrue```, but I think avoiding it might be better

test/unit/spec/LoginRouter_spec.js
gowthamidommety-okta: The way the backend response sends autopush is specific to a factor id. I know other factors might have push (maybe autoPush also?) at some point so i added in factorType and provider. I can remove it for now though and add those in later when the time comes?

test/unit/spec/LoginRouter_spec.js
```diff
@@ -175,6 +174,17 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, Logger, OktaAut
       expect(Logger.warn).toHaveBeenCalledWith('Field not defined in schema', arg1);
     }
 
+    var getAutoPushResponse = function (response, autoPushVal) {
+      var responseCopy = Util.deepCopy(response);
+      var factors = responseCopy['response']['_embedded']['factors'];
+      var factor = _.findWhere(factors, {factorType: 'push', provider: 'OKTA'});
```
santhoshbalakrishnan: nit: Since only ID is needed, you can directly do something like: 
`var id = _.findWhere(factors, {factorType: 'push', provider: 'OKTA'}).id;`

test/unit/spec/LoginRouter_spec.js
```diff
@@ -175,6 +174,17 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, Logger, OktaAut
       expect(Logger.warn).toHaveBeenCalledWith('Field not defined in schema', arg1);
     }
 
+    var getAutoPushResponse = function (response, autoPushVal) {
+      var responseCopy = Util.deepCopy(response);
+      var factors = responseCopy['response']['_embedded']['factors'];
```
santhoshbalakrishnan: nit: Since it is used in 2 places, move `responseCopy['response']['_embedded']` into a var?

test/unit/spec/LoginRouter_spec.js
```diff
@@ -175,6 +174,17 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CookieUtil, Logger, OktaAut
       expect(Logger.warn).toHaveBeenCalledWith('Field not defined in schema', arg1);
     }
 
+    var getAutoPushResponse = function (response, autoPushVal) {
```
santhoshbalakrishnan: Since this is used in 2 places, would it be worth it to move to some util?

gowthamidommety-okta: It doesn't fit in any of the existing helper utils.. I could create a new util file. What do you think?

PR #398, opened 2018-02-15
Upgrade i18n to 42.3.0 to include KO language fixes

santhoshbalakrishnan:
## Description

1. Removing unwanted commas.
2. Making two sentences better in Korean.
3. Updated i18n `package.json` to have `/` as `_requiredBy` field.

Major changes only to this file: https://github.com/santhoshbalakrishnan/okta-signin-widget/blob/6d936ca923c784160c8412f396922fe0c276e08a/packages/%40okta/i18n/dist/properties/login_ko.properties

## Reviewers

- @mauriciocastillosilva-okta 
- @asong-okta 
- @gowthamidommety-okta


packages/@okta/i18n/package.json
```diff
@@ -1,106 +1,80 @@
 {
   "_args": [
     [
-      "@okta/i18n@42.1.0",
-      "/Users/mauricio.castillosilva/okta/okta-signin-widget"
+      {
```
gowthamidommety-okta: Not sure about this change. @mauriciocastillosilva-okta ?

mauriciocastillosilva-okta: I've seen this with a simple string array or an object array, but I have seen no difference on how this affects...
My guess is that this happens because of different versions of npm

gowthamidommety-okta: Talked to Mau offline. This is ok 👍

PR #400, opened 2018-02-21
Change order of Factors

alexeystadnikov-okta:
Resolves OKTA-144166

screen shot 2018-02-21 at 4 19 34 pm

@mauriciocastillosilva-okta 
@yiyangli-okta 

Please, review.

### DIFF
--- a/src/util/FactorUtil.js
+++ b/src/util/FactorUtil.js
@@ -34,84 +34,84 @@ function (Okta, TimeUtil) {
       iconClassName: 'mfa-okta-verify',
       sortOrder: 1
     },
-    'SMS': {
-      label: 'factor.sms',
-      description: 'factor.sms.description',
-      iconClassName: 'mfa-okta-sms',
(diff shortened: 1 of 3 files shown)


src/util/FactorUtil.js
```diff
@@ -32,18 +32,12 @@ function (Okta, TimeUtil) {
       label: 'factor.oktaVerifyPush',
       description: 'factor.push.description',
       iconClassName: 'mfa-okta-verify',
-      sortOrder: 1
-    },
-    'SMS': {
-      label: 'factor.sms',
-      description: 'factor.sms.description',
-      iconClassName: 'mfa-okta-sms',
       sortOrder: 2
```
mauriciocastillosilva-okta: shouldn't this be the same as ```OKTA_VERIFY```? They are at the end the same factor

alexeystadnikov-okta: Fixed

PR #404, opened 2018-03-05
:bug: Fixes Okta Verify polling issue when switching factors

magizh-okta:
* We continute to poll even after user navigates to another factor after
triggering a MFA verify. This was due to the fact that the transaction state not being set
when sing MFA verify until the poll complted with a success/failure.

The fix here is that the transaction is updated when we get the first response from server.

Polling stops when you switch factors

bugfix2

### DIFF
--- a/src/models/Factor.js
+++ b/src/models/Factor.js
@@ -200,7 +200,7 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
       // Set/Remove the remember device cookie based on the remember device input.
       var self = this;
 
-      return this.doTransaction(function (transaction) {
+      return this.manageTransaction(function (transaction, setTransaction) {
         var data = {
           rememberDevice: rememberDevice
         };
@@ -243,6 +243,8 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
           var options = {
             'delay': PUSH_INTERVAL
           };
+          setTransaction(trans);
+          // In Okta verify case we initiate poll.
           if (trans.status === 'MFA_CHALLENGE' && trans.poll) {
             return Q.delay(PUSH_INTERVAL).then(function() {
               if (self.pushFactorHasAutoPush()) {
@@ -250,14 +252,11 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
                   return self.get('autoPush');
                 };
               }
-              return trans.poll(options);
+              return trans.poll(options).then(function(trans) {
+                setTransaction(trans);
+              });
             });
           }
-          return trans;
-        })
-        .fail(function (err) {
-          // Clean up the cookie on failure.
-          throw err;
         });
       });
     },
--- a/test/unit/spec/MfaVerify_spec.js
+++ b/test/unit/spec/MfaVerify_spec.js
@@ -328,12 +328,13 @@ function (Okta,
       // Spy on backup factor model for TOTP, since TOTP is special
       var model = isTotp ? controller.model.get('backupFactor') : controller.model;
       spyOn(model, 'trigger').and.callThrough();
+      spyOn(model, 'setTransaction').and.callThrough();
       spyOn(controller.options.appState, 'set').and.callThrough();
       spyOn(RouterUtil, 'routeAfterAuthStatusChange').and.callThrough();
     }
 
     // Expect -
-    // 1. model triggers the setTransaction event
+    // 1. setTransaction on model is called with transaction
     // 2. controller sets the transaction property on the appState
     // 3. routerAfterAuthStatusChange is called with the right parameters (success response)
     function expectSetTransaction(router, res, isTotp) {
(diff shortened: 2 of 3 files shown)


src/models/Factor.js
```diff
@@ -244,20 +244,21 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
             'delay': PUSH_INTERVAL
           };
           if (trans.status === 'MFA_CHALLENGE' && trans.poll) {
+            // In verify case since we poll we dont return the trans
```
mauriciocastillosilva-okta: nit: just a little bit more explicit.
```
When the factor has a poll link, we don't return the transaction here. 
Rather we set it and start polling.
```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2150,13 +2151,28 @@ function (Okta,
...
-            itp('on SUCCESS, polling stops and form is submitted', function () {
+            itp('on SUCCESS, sets transaction, polling stops and form is submitted', function () {
               return setupOktaPush().then(function (test) {
+                spyOn(test.router.controller.model, 'setTransaction').and.callThrough();
                 spyOn(test.router.settings, 'callGlobalSuccess');
                 return setupPolling(test, resSuccess)
-                .then(function () { return tick(test); }) // Final tick - SUCCESS
+                .then(function () {
+                  return tick(test);
+                }) // Final tick - SUCCESS
```
mauriciocastillosilva-okta: comment should be in line above right? line 2160, after the ```tick```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2150,13 +2151,28 @@ function (Okta,
...
                 .then(function () {
                   expect(test.router.settings.callGlobalSuccess).toHaveBeenCalled();
+                  // One after first poll returns and one after polling finished.
+                  expect(test.router.controller.model.setTransaction.calls.count()).toBe(2);
```
mauriciocastillosilva-okta: we should check that ```setTransaction``` was called with the right arguments. First should be for the MFA_CHALLENGE status, and second one should be for SUCCESS if I'm reading this test correctly

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2150,13 +2151,28 @@ function (Okta,
...
+                });
+              });
+            });
+            itp('sets transaction state to MFA_CHALLENGE before poll', function () {
+              return setupOktaPush().then(function (test) {
+                $.ajax.calls.reset();
+                test.setNextResponse(resChallengePush);
+                test.form[0].submit();
+                tick().then(function() {
```
mauriciocastillosilva-okta: this tests usually return tick
```
  return tick(test);
})
.then(function (test) {
  expect(test.router.controller.model.appState.get('transaction').status).toBe('MFA_CHALLENGE');
}
```

src/models/Factor.js
```diff
@@ -244,20 +244,21 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
...
+                setTransaction(trans);
+              });
             });
           }
-          return trans;
-        })
-        .fail(function (err) {
-          // Clean up the cookie on failure.
-          throw err;
+          setTransaction(trans);
```
rchild-okta: Can you move the setTransaction to before the if? The logic should be:
- We always set the current transaction (first)
- If it's a poll, we then start the poll and update the transaction when it's done

src/models/Factor.js
```diff
@@ -244,20 +244,21 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
...
-              return trans.poll(options);
+              return trans.poll(options).then(function(trans) {
+                setTransaction(trans);
+              });
             });
           }
-          return trans;
-        })
-        .fail(function (err) {
-          // Clean up the cookie on failure.
```
rchild-okta: I am curious - what is this for? Is it okay just to remove it?

magizh-okta: We have a fail case in BaseLoginModel https://github.com/okta/okta-signin-widget/pull/404/files#diff-733a4f9ae0ae362416e8e89d8ef73db7R54 which has other checks anyway  this feels it is unnecessary. I would add additional check https://github.com/okta/okta-signin-widget/pull/404/files#diff-733a4f9ae0ae362416e8e89d8ef73db7R42

src/models/Factor.js
magizh-okta: I moved the setTranasction outside If and the code now is readable. Just added a small comment now.

src/models/Factor.js
```diff
@@ -243,21 +243,20 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
...
                   return self.get('autoPush');
                 };
               }
-              return trans.poll(options);
+              return trans.poll(options).then(function(trans) {
+                setTransaction(trans);
+              });
             });
           }
-          return trans;
```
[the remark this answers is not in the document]
magizh-okta: No we call setTransaction now. The reason we were returning was because doTransaction was setting it for us.

src/models/Factor.js
magizh-okta: Also verified that this was a stale comment https://github.com/okta/okta-signin-widget/commit/7ad708db07845d89a8ad2cb84fac1d424b3a57e0#diff-14b6f8f7494d563336c6490bab399799R72

PR #406, opened 2018-03-09
Fixing e2e tests

mauriciocastillosilva-okta:
Problem seems to be that we are trying to do something before the build is done. So the fix is to run build:release before starting e2e tests

Another problem is that there was a breaking change in angular/router, so fixing the version to before the breaking change, which is 5.2.7

### DIFF
--- a/test/e2e/angular-app/package.json
+++ b/test/e2e/angular-app/package.json
@@ -20,7 +20,7 @@
     "@angular/http": "^5.0.0",
     "@angular/platform-browser": "^5.0.0",
     "@angular/platform-browser-dynamic": "^5.0.0",
-    "@angular/router": "^5.0.0",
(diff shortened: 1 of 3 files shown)


test/e2e/angular-app/package.json
```diff
@@ -20,7 +20,7 @@
     "@angular/http": "^5.0.0",
     "@angular/platform-browser": "^5.0.0",
     "@angular/platform-browser-dynamic": "^5.0.0",
-    "@angular/router": "^5.0.0",
+    "@angular/router": "5.2.7",
```
jmelberg-okta: For more context - this issue was discovered in [okta-oidc-js](https://github.com/okta/okta-oidc-js/issues/140) and extends down to AuthJS.

PR #407, opened 2018-03-13
Move responsiveness code from okta-core to okta-theme in signin-widget.

santhoshbalakrishnan:
## Description

Styles are being moved from https://github.com/okta/okta-core/blob/master/clients/loginpage/sass/loginpage-theme.scss#L43

## Reviewers

- @rchild-okta 
- @mauriciocastillosilva-okta 

## Screenshare

siw-resp

### DIFF
--- a/assets/sass/okta-theme.scss
+++ b/assets/sass/okta-theme.scss
@@ -381,7 +381,7 @@
       background-image: url('../img/icons/mfa/voicecall@2x.png');
     }
   }
-  
+
   .mfa-email-30,
   .enroll-factor-row .mfa-okta-email {
     /* -- Factor Icons (small): EMAIL -- */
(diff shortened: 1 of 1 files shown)


bretterer:
Any way to make this so the logo remains in the center?

santhoshbalakrishnan:
@brianretterer-okta Good thing to think about! We do support that as long as the logo is of supported size. This is a bit more than that, why is why it didn't work. Also, it started breaking only below the size of iphone 4, which I do not know if we still support. However, that is a good thing to add. But, it's a little bit out of the scope of the current PR, where we concentrated on the widget itself aligning proper on okta-theme.

assets/sass/okta-theme.scss
```diff
@@ -820,3 +820,33 @@
...
+/*  Remove the background image and borders on a smaller window size */
+@include max-width-mq(600px) {
+  // scss-lint:disable ImportantRule
+  .login-bg-image {
+    background-image: none !important;
+    background-color: #fff !important;
+    filter: unset !important;
+  }
+
+  /* See line 59. Unfortunately, we have to use the selector this way to override at small screen sizes. */
```
mauriciocastillosilva-okta: nit: Might be just enough to say we use the same selector than in line 59?

santhoshbalakrishnan: Haha okay!

santhoshbalakrishnan: Changed to this: Using the same selector as line 59 to override styles at small screen sizes.

mauriciocastillosilva-okta: sounds good. Once you push your changes this is ready, but let me know if you need me to review again.

PR #410, opened 2018-03-20
Signin widget change to use org level registration policy.

uvartak-okta:
🌱 Signin widget change to use org level registration policy.

Updated registrationControlller to use new registration endpoint. Added unit tests.
Resolves OKTA-160857

Reviewers:
@hor-kanchan-okta @nikhilvenkatraman-okta

### DIFF
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -70,7 +70,7 @@ function (
       // setup schema
       var Schema = RegistrationSchema.extend({
         settings: this.options.settings,
-        url: this.getRegistrationApiBase()+'/form'
+        url: this.getRegistrationSchemaApi()+'/form'
       });
       var schema = new Schema();
       this.state.set('schema', schema);
@@ -84,7 +84,7 @@ function (
         return this.options.settings.get('baseUrl');
       }
     },
-    getRegistrationApiBase: function() {
+    getRegistrationSchemaApi: function() {
       if (this.options.settings.get('policyId')) {
         return this.options.settings.get('baseUrl') + '/api/v1/registration/' + this.options.settings.get('policyId');
       } else {
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -70,7 +70,7 @@ function (
       // setup schema
       var Schema = RegistrationSchema.extend({
         settings: this.options.settings,
-        url: this.getRegistrationPolicyApi()+'/form'
+        url: this.options.settings.get('baseUrl') + '/api/v1/registration/form'
       });
       var schema = new Schema();
       this.state.set('schema', schema);
@@ -85,11 +85,7 @@ function (
       return apiUrl;
     },
     getRegistrationPolicyApi: function(policyId) {
-      if (policyId) {
-        return this.options.settings.get('baseUrl') + '/api/v1/registration/' + policyId;
-      } else {
-        return this.options.settings.get('baseUrl') + '/api/v1/registration';
-      }
+      return this.options.settings.get('baseUrl') + '/api/v1/registration/' + policyId;
     },
     doPostSubmit: function () {
       if (this.model.get('activationToken')) {
@@ -185,7 +181,7 @@ function (
       self.state.get('schema').on('parseComplete', function(updatedSchema) {
         var modelProperties = updatedSchema.properties.createModelProperties();
         if (!updatedSchema.error) {
-          self.settings.set('defaultPolicyId', updatedSchema.properties.policyId);
+          self.settings.set('defaultPolicyId', updatedSchema.properties.defaultPolicyId);
         }
         // create model
         self.model = self.createRegistrationModel(modelProperties);
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -70,14 +70,26 @@ function (
       // setup schema
       var Schema = RegistrationSchema.extend({
         settings: this.options.settings,
-        url: this.getRegistrationApiUrl()+'/form'
+        url: this.getRegistrationApiBase()+'/form'
       });
       var schema = new Schema();
       this.state.set('schema', schema);
     },
     getRegistrationApiUrl: function() {
-      var clientId = this.options.settings.get('clientId');
-      return this.options.settings.get('baseUrl')+'/api/v1/registration/'+clientId;
+      if (this.options.settings.get('policyId')) {
+        return this.options.settings.get('baseUrl') + '/api/v1/registration/' + this.options.settings.get('policyId');
+      } else if (this.settings.get('policyId')) {
+        return this.options.settings.get('baseUrl') + '/api/v1/registration/' + this.settings.get('policyId');
+      } else {
+        return this.options.settings.get('baseUrl');
+      }
+    },
+    getRegistrationApiBase: function() {
+      if (this.options.settings.get('policyId')) {
+        return this.options.settings.get('baseUrl') + '/api/v1/registration/' + this.options.settings.get('policyId');
+      } else {
+        return this.options.settings.get('baseUrl') + '/api/v1/registration';
+      }
     },
     doPostSubmit: function () {
       if (this.model.get('activationToken')) {
@@ -172,6 +184,9 @@ function (
       // register parse complete event listener
       self.state.get('schema').on('parseComplete', function(updatedSchema) {
         var modelProperties = updatedSchema.properties.createModelProperties();
+        if (!updatedSchema.error) {
+          self.settings.set('policyId', updatedSchema.properties.policyId);
+        }
         // create model
         self.model = self.createRegistrationModel(modelProperties);
         // create form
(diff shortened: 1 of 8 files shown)


test/unit/spec/Registration_spec.js
```diff
@@ -120,7 +122,7 @@ function (Q, _, $, OktaAuth, Backbone, Util, Expect, Beacon, RegForm, RegSchema,
       form: form,
       beacon: beacon,
       ac: authClient,
-      setNextResponse: setNextResponse
+      setNextResponse: setNextResponse,
```
nikhilvenkatraman-okta: nit: extra comma

src/RegistrationController.js
```diff
@@ -70,14 +70,26 @@ function (
...
-        url: this.getRegistrationApiUrl()+'/form'
+        url: this.getRegistrationApiBase()+'/form'
       });
       var schema = new Schema();
       this.state.set('schema', schema);
     },
     getRegistrationApiUrl: function() {
-      var clientId = this.options.settings.get('clientId');
-      return this.options.settings.get('baseUrl')+'/api/v1/registration/'+clientId;
+      if (this.options.settings.get('policyId')) {
```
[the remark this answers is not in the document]
uvartak-okta: this.options.settings.get('policyId') is initialized from html. In future when we allow multiple org level policies you can tie the widget to specific org level policy. this.settings.get('policyId') is the default org level policy that we fetch from okta-core.

src/RegistrationController.js
```diff
@@ -70,14 +70,26 @@ function (
...
+        return this.options.settings.get('baseUrl') + '/api/v1/registration/' + this.options.settings.get('policyId');
+      } else if (this.settings.get('policyId')) {
+        return this.options.settings.get('baseUrl') + '/api/v1/registration/' + this.settings.get('policyId');
+      } else {
+        return this.options.settings.get('baseUrl');
```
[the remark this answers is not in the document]
uvartak-okta: we do not need it.. however i was not sure how to return error.

src/RegistrationController.js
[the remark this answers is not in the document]
nikhilvenkatraman-okta: Hmmm this seems very confusing. Currently `this.options.settings.get('clientId')` is set via the config (i.e html) and points to the `clientId` in https://github.com/okta/okta-signin-widget/blob/master/src/models/Settings.js#L123. So we should do the same for `policyId` that is initialized from the config/html.

From what you are saying it seems like we have 
1: PolicyId defined from the html (similar to clientId)
2: Default policy Id 

Is the default `policyId` used if `policyId` is not initialized from the html ?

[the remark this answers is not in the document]
uvartak-okta: yep

src/RegistrationController.js
```diff
@@ -70,14 +70,26 @@ function (
       // setup schema
       var Schema = RegistrationSchema.extend({
         settings: this.options.settings,
-        url: this.getRegistrationApiUrl()+'/form'
+        url: this.getRegistrationApiBase()+'/form'
```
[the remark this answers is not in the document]
uvartak-okta: done

src/RegistrationController.js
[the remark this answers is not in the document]
uvartak-okta: yes.. one other benefit is - this should allow us to keep widget backward compatible with RAAS. This is not a requirement  i.e. we are not actively looking for RAAS regressions with newer code.. but widget side of things should work fine.

src/models/RegistrationSchema.js
```diff
@@ -71,6 +71,7 @@ function (Okta, BaseSchema, SchemaProperty) {
           BaseSchema.Model.prototype.parse.apply(self, [resp]);
           resp = parseResponseData(resp);
         }
+        self.properties.policyId = resp.policyId;
```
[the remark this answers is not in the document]
uvartak-okta: done

test/unit/helpers/dom/RegistrationForm.js
```diff
@@ -4,6 +4,7 @@ define(['./Form'], function (Form) {
   var LASTNAME_FIELD = 'lastName';
   var EMAIL_FIELD = 'userName';
   var PASSWORD_FIELD = 'password';
+  var REFRRER_FIELD = 'referrer';
```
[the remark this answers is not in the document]
uvartak-okta: done

test/unit/spec/Registration_spec.js
uvartak-okta: done

src/RegistrationController.js
```diff
@@ -172,6 +184,9 @@ function (
       // register parse complete event listener
       self.state.get('schema').on('parseComplete', function(updatedSchema) {
         var modelProperties = updatedSchema.properties.createModelProperties();
+        if (!updatedSchema.error) {
+          self.settings.set('policyId', updatedSchema.properties.policyId);
```
nikhilvenkatraman-okta: Are we trying to overwrite the `policyId` when a default `policyId` is present here ?

It feels like we should create another variable called `defaultPolicyId` in Settings.js to keep them separate and set it here when we want to use the default `policyId`. Also for this case I feel it would be ideal if the API actually sent the key as `defaultPolicyId`.

src/RegistrationController.js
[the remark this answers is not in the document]
nikhilvenkatraman-okta: We could simplify the code here to something like 

```
     getRegistrationPolicyApi: function(policyId) {
     	return this.options.settings.get('baseUrl') + '/api/v1/registration/' + policyId;
     } 

     // default policyId
     var defaultPolicyId = this.settings.get('defaultPolicyId');
     // org policyId
     var orgPolicyId = this.options.settings.get('policyId');
     var apiUrl = defaultPolicyId ? getRegistrationPolicyApi(defaultPolicyId) : getRegistrationPolicyApi(orgPolicyId);
     return apiUrl;

```

src/RegistrationController.js
```diff
@@ -70,14 +70,26 @@ function (
...
+      var orgPolicyId = this.options.settings.get('policyId');
+      var apiUrl = defaultPolicyId ? this.getRegistrationPolicyApi(defaultPolicyId) :
+        this.getRegistrationPolicyApi(orgPolicyId);
+      return apiUrl;
+    },
+    getRegistrationPolicyApi: function(policyId) {
+      if (policyId) {
+        return this.options.settings.get('baseUrl') + '/api/v1/registration/' + policyId;
+      } else {
+        return this.options.settings.get('baseUrl') + '/api/v1/registration';
```
nikhilvenkatraman-okta: why do we need to return `this.options.settings.get('baseUrl') + '/api/v1/registration';` in the else ?

src/RegistrationController.js
```diff
@@ -172,6 +184,9 @@ function (
       // register parse complete event listener
       self.state.get('schema').on('parseComplete', function(updatedSchema) {
         var modelProperties = updatedSchema.properties.createModelProperties();
+        if (!updatedSchema.error) {
```
nikhilvenkatraman-okta: Not sure if we need this if check as we already are checking for `updatedSchema.error` below on line 200.

src/models/RegistrationSchema.js
```diff
@@ -61,6 +61,7 @@ function (Okta, BaseSchema, SchemaProperty) {
           this.properties.comparator = 'sortOrder';
           this.properties.sort();
         }
+        this.properties.policyId = resp.policyId;
```
nikhilvenkatraman-okta: Isn't this the default policy Id ? Should we just call it `this.properties.defaultPolicyId` ? The difference is just that the server sends this instead of it being set the html. It would be nice if we can also add a small comment here for that.

src/RegistrationController.js
uvartak-okta: fixed

src/RegistrationController.js
uvartak-okta: this is needed for a unit test that does not have policy populated

nikhilvenkatraman-okta: seems odd that we are making a code change for the test. Have we already tried to mock the server response?

src/util/RegistrationFormFactory.js
```diff
@@ -125,10 +125,10 @@ define([
     var inputOptions = SchemaFormFactory.createInputOptions(schemaProperty);
     if (inputOptions.type === 'select') {
       inputOptions = _.extend(inputOptions, {
-        label: schemaProperty.get('description')
+        label: schemaProperty.get('title')
```
nikhilvenkatraman-okta: Are we going to use `schemaProperty.get('title')` going forward to match with the API? Will this be a problem if we try to use an older version of the widget?

uvartak-okta: Confirmed with UX that they always expected title and not description to be shown as placeholder/label.

src/RegistrationController.js
uvartak-okta: hmm.. my bad.. testing it again after removing this check and all tests pass. I have removed this check

uvartak-okta:
Ack.. created JIRA https://oktainc.atlassian.net/browse/OKTA-163598 to update readme.

PR #411, opened 2018-03-21
:seedling: Send relayState param with registration post call.

nikhilvenkatraman-okta:
- Sign in widget now sends a relayState param with registration post call which is the fromURI value of window.location.href 

Resolves : OKTA-152289

Reviewers: 
@hor-kanchan-okta @uvartak-okta

### DIFF
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -127,6 +127,19 @@ function (
         }
       });
     },
+    getJsonFromUrl: function(url) {
+      var query = url.substr(1);
+      var result = {};
+      query.split('&').forEach(function(part) {
+        var item = part.split('=');
+        result[item[0]] = item[1];
+      });
+      return result;
+    },
+    getRelayStateData: function () {
+      var urlParams = this.getJsonFromUrl(window.location.search);
+      return urlParams.fromURI || '';
+    },
     createRegistrationModel: function (modelProperties) {
       var self = this;
       var Model = Okta.Model.extend({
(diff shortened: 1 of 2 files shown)


uvartak-okta:
🚢 relayState parameter looks good.

src/RegistrationController.js
```diff
@@ -127,6 +127,21 @@ function (
         }
       });
     },
+    // @param url (Eg: ?fromURI=%2Fapp%2FUserHome&query=blah)
+    // returns {fromURI:'%2Fapp%2FUserHome', query: blah}
+    getJsonFromUrl: function(url) {
+      var query = url.substr(1);
```
hor-kanchan-okta: what happen if substr is empty?

src/RegistrationController.js
```diff
@@ -139,7 +154,10 @@ function (
         },
         toJSON: function() {
           var data = Okta.Model.prototype.toJSON.apply(this, arguments);
-          return {userProfile: data};
+          return {
+            userProfile: data,
+            relayState: self.getRelayStateData()
```
hor-kanchan-okta: Does that mean we always return the `relatState` even it is empty?

nikhilvenkatraman-okta: yeah, I had asked @iancorcoran-okta about the this.

src/RegistrationController.js
hor-kanchan-okta: From the spec
https://www.w3schools.com/jsref/jsref_substr.asp
```If start is negative or larger than the length of the string, start is set to 0```
So empty string will be fine

nikhilvenkatraman-okta: yep, when query is empty it will return empty object.

PR #414, opened 2018-03-27
Less bash-specific instructions

nbarbettini:
Removing bash semantics from the Developing the widget section.


nbarbettini:
cc @santhoshbalakrishnan @vijetmahabaleshwar-okta

jmelberg-okta:
Fixed with #461

PR #415, opened 2018-03-27
Cross-platform package scripts

nbarbettini:
Removed bash-only things like `&&` from package.json scripts.

- [x] Use `rimraf` to replace `rm -rf`
- [x] Split up tasks into smaller chunks and use `npm-run-all` instead of `&` or `&&`
- [ ] Find a cross-platform replacement for `kill -9`

Can be rebased after #414 is merged.


haishengwu-okta:
#1 had been done
#2 seem low priority

Issue #355, opened 2017-11-28
Undeclared variable error- vkbeautify.0.99.00.beta.js:74 Uncaught ReferenceError: ix is not defined


chidanandaboligadde:
I just cloned the widget code and stared the application (npm start). The browser displays empty screen with console error showing: 
vkbeautify.0.99.00.beta.js:74 Uncaught ReferenceError: ix is not defined

I used the the widget few weeks back and it was working fine.  How do I disable the 'strict use' mode to get rid of this error?

santhoshbalakrishnan:
Hi @chidanandaboligadde , I just tried the same and it seems to work fine for me. Long shot, but did you maybe miss `npm install` after fetching the latest code? 

Can you also give more info about node, npm, OS and browser versions? It might also be worth trying with any of your browser extensions disabled. Also, where do see you the 'strict use' mode?

chidanandaboligadde:
$ npm --version
5.3.0

OS: Windows 7 Enterprise
Browsers: Google Chrome and IE11

I got rid of the error by introducing 'var' key word for the variables in the for loop in okta-signin-widget\packages\@okta\courage\src\vendor\plugins\vkbeautify.0.99.00.beta.js

lazytesting:
I had the same issue (and same fix works for me).
Running on Windows with Node 9.3.0 and NPM 5.5.1

santhoshbalakrishnan:
With https://github.com/okta/okta-signin-widget/pull/416, this should be solved. Also tested to work on Windows 7, Node 10.7.0, npm 6.1.0.

<img width="1383" alt="screen shot 2018-07-25 at 11 30 58 am" src="https://user-images.githubusercontent.com/24683447/43220165-3b8d1518-8ffe-11e8-95dc-7add2076b56b.png">

Although, to get it working on Windows, I had to modify this line https://github.com/okta/okta-signin-widget/blob/master/package.json#L23 to 

```
"clean": "rimraf target && rimraf dist",
```

and also do `npm install rimraf`. Will treat that as a separate PR soon.

PR #416, opened 2018-03-27
Exclude vkbeautify from build

nbarbettini:
This works with only line 45 added. Line 110 is not necessary, but I think it should be cleaned up (removed) if I'm understanding the build process correctly.

This still needs: testing to make sure the widget really really doesn't need the `vkbeautify` package. 😄 
cc @santhoshbalakrishnan @rchild-okta


santhoshbalakrishnan:
This might solve https://github.com/okta/okta-signin-widget/issues/355 as well.

jmelberg-okta:
Closing with respect to #497

PR #426, opened 2018-04-10
honor idp discovery feature for all other paths

haishengwu-okta:
- don't whitelist for primary auth v.s. idp discovery auth but only relies on the feature to determine whether run idp discovery flow or primary auth flow.
- therefore shift the logic to backend to set up the auth flow.
- video: https://okta.box.com/s/l640apxbvql99s9h7dqb8qfs16nvk27i
- OKTA-163556

@mauriciocastillosilva-okta @yuliu-okta 
^ @hor-kanchan-okta

### DIFF
--- a/test/unit/spec/LoginRouter_spec.js
+++ b/test/unit/spec/LoginRouter_spec.js
@@ -395,7 +395,7 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Exp
         expect(form.isIDPDiscovery()).toBe(true);
       });
     });
-    itp('navigates to IDPDiscovery for /app/salesforce/{id}/sso/saml when features.idpDiscovery is false', function () {
+    itp('navigates to PrimaryAuth for /app/salesforce/{id}/sso/saml when features.idpDiscovery is false', function () {
       return setup({'features.idpDiscovery': false})
       .then(function (test) {
         Util.mockRouterNavigate(test.router);
@@ -407,7 +407,7 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Exp
         expect(form.isPrimaryAuth()).toBe(true);
       });
     });
-    itp('navigates to PrimaryAuth for /any/other when features.idpDiscovery is true', function () {
+    itp('navigates to IDPDiscovery for /any/other when features.idpDiscovery is true', function () {
       return setup({'features.idpDiscovery': true})
       .then(function (test) {
         Util.mockRouterNavigate(test.router);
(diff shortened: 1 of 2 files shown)


test/unit/spec/LoginRouter_spec.js
```diff
@@ -383,11 +383,47 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Exp
...
-        test.router.navigate('login/default');
+        test.router.navigate('/app/salesforce/abc123sef/sso/saml');
+        return Expect.waitForIDPDiscovery();
+      })
+      .then(function () {
+        var form = new IDPDiscoveryForm($sandbox);
+        expect(form.isIDPDiscovery()).toBe(true);
+      });
+    });
+    itp('navigates to IDPDiscovery for /app/salesforce/{id}/sso/saml when features.idpDiscovery is false', function () {
```
yuliu-okta: navigates to primaryAuth

test/unit/spec/LoginRouter_spec.js
```diff
@@ -383,11 +383,47 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Exp
...
+        Util.mockRouterNavigate(test.router);
+        test.router.navigate('/app/salesforce/abc123sef/sso/saml');
+        return Expect.waitForPrimaryAuth();
+      })
+      .then(function () {
+        var form = new PrimaryAuthForm($sandbox);
+        expect(form.isPrimaryAuth()).toBe(true);
+      });
+    });
+    itp('navigates to PrimaryAuth for /any/other when features.idpDiscovery is true', function () {
```
yuliu-okta: navigates to IdP Discovery

haishengwu-okta:
@mauriciocastillosilva-okta 
thanks for bring it up and that's exactly we want. (unless we missed any cases)

- this is how the [feature variable set from jsp](https://github.com/okta/okta-core/blob/master/webapp/src/main/resources/META-INF/resources/WEB-INF/jsp/loginpage_backbone.jsp#L377-L380) and essentially it comes [from controller](https://github.com/okta/okta-core/blob/master/server/ui/src/main/java/com/saasure/controller/enduser/login/LoginController.java#L337)
- any url path will hit corresponding controller first and it does imply the dependency on backend to set the feature variable properly if want to support idp discovery for that url.
- basically we shift whitelist/blacklist logic from widget to backend, which I think it's better because `features.idpDiscovery` is single source of truth to enable idp discovery for widget.

haishengwu-okta:
@yuliu-okta 
widget is a SPA hence the initial value will retain while navigation unless refresh/redirect.

PR #431, opened 2018-04-17
Show registration link at idp discovery identifier page

haishengwu-okta:
- support RaaS when idp discovery is enabled
- video: https://okta.box.com/s/olz3hsws3rzgqvdj8ty9fp18kxm040t5
- OKTA-154930

@yuliu-okta @mauriciocastillosilva-okta 
@santhoshbalakrishnan @uvartak-okta 
@yingli-okta

### DIFF
--- a/test/unit/spec/IDPDiscovery_spec.js
+++ b/test/unit/spec/IDPDiscovery_spec.js
@@ -1053,16 +1053,20 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryF
           expect(test.form.registrationContainer().length).toBe(0);
         });
       });
-      itp('does not show the registration button if features.registration is false', function () {
-        var registration =  {
-        };
-        return setupRegistrationButton(null, registration).then(function (test) {
+      itp('does not show the registration button if features.registration is not set', function () {
+        var registration = {};
+        return setupRegistrationButton(undefined, registration).then(function (test) {
           expect(test.form.registrationContainer().length).toBe(0);
         });
       });
-      itp('show the registration button if settings.registration.enable is true', function () {
-        var registration =  {
-        };
+      itp('does not show the registration button if features.registration is false', function () {
+        var registration = {};
+        return setupRegistrationButton(false, registration).then(function (test) {
+          expect(test.form.registrationContainer().length).toBe(0);
+        });
+      });
+      itp('show the registration button if registration.enable is true', function () {
+        var registration = {};
         return setupRegistrationButton(true, registration).then(function (test) {
           expect(test.form.registrationContainer().length).toBe(1);
           expect(test.form.registrationLabel().length).toBe(1);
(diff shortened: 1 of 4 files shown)


test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1036,5 +1046,47 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryF
...
+      });
+      itp('does not show the registration button if features.registration is false', function () {
+        var registration =  {
+        };
+        return setupRegistrationButton(null, registration).then(function (test) {
+          expect(test.form.registrationContainer().length).toBe(0);
+        });
+      });
+      itp('show the registration button if settings.registration.enable is true', function () {
+        var registration =  {
```
santhoshbalakrishnan: Is this a valid usecase? Curious if this can happen actually.

yiyangli-okta:
🚢

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1036,5 +1046,47 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryF
...
+    Expect.describe('Registration Flow', function () {
+      itp('does not show the registration button if features.registration is not set', function () {
+        return setup().then(function (test) {
+          expect(test.form.registrationContainer().length).toBe(0);
+        });
+      });
+      itp('does not show the registration button if features.registration is false', function () {
+        var registration =  {
+        };
+        return setupRegistrationButton(null, registration).then(function (test) {
```
mauriciocastillosilva-okta: description says `features.registration` is false, but here we are passing `null`. Should we change it to `false` or change the test description to match the test?

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1036,5 +1046,47 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryF
...
+          expect(test.form.registrationContainer().length).toBe(0);
+        });
+      });
+      itp('show the registration button if settings.registration.enable is true', function () {
```
mauriciocastillosilva-okta: we should say ```if features.registration is enabled```, cause we don't really have a ```settings.registration.enable``` value

test/unit/spec/IDPDiscovery_spec.js
haishengwu-okta: my understanding is RaaS and IDP discovery are sort of decoupled feature (thanks to Okta is always to be default idp) therefore we allow registration either primary auth or idp auth.

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1053,16 +1053,20 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryF
           expect(test.form.registrationContainer().length).toBe(0);
         });
       });
+      itp('does not show the registration button if features.registration is not set', function () {
```
mauriciocastillosilva-okta: should probably say ```is undefined``` since the test above is for when ```it is not set```

PR #434, opened 2018-04-19
:bug: Use email instead of userName for activation email message

uvartak-okta:
New self service registration logic does not use userName attribute. Instead it uses email and optionally login attributes. It is safer to use email to display registration completion messages.

Resolves OKTA-166442

### DIFF
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -106,7 +106,7 @@ function (
         });
       } else {
         // register via activation email
-        this.model.appState.set('username', this.model.get('userName'));
+        this.model.appState.set('username', this.model.get('email'));
         this.model.appState.trigger('navigate', 'signin/register-complete');
       }
     },
@@ -115,7 +115,7 @@ function (
       this.model.attributes = postData;
       Backbone.Model.prototype.save.call(this.model).then(function() {
         var activationToken = self.model.get('activationToken');
-        var postSubmitData = activationToken ? activationToken : self.model.get('userName');
+        var postSubmitData = activationToken ? activationToken : self.model.get('email');
         self.settings.postSubmit(postSubmitData, function() {
           self.doPostSubmit();
         }, function(errors) {
--- a/src/util/RegistrationFormFactory.js
+++ b/src/util/RegistrationFormFactory.js
@@ -79,6 +79,9 @@ define([
     if (_.isString(regex)) {
       if (regex === '^[#/userName]') {
         var username = model.get('userName');
+        if (username == undefined) {
+          username = model.get('email');
+        }
         var password = value;
         return !passwordContainsUserName(username, password);
       } else {
--- a/src/util/RegistrationFormFactory.js
+++ b/src/util/RegistrationFormFactory.js
@@ -78,12 +78,9 @@ define([
     
     if (_.isString(regex)) {
       if (regex === '^[#/userName]') {
-        var username = model.get('userName');
         // with email as login enabled, we only have email populated.
         // Therefore we fallback and run validation with email attribute.
-        if (!model.has('userName')) {
-          username = model.get('email');
-        }
+        var username = model.has('userName') ? model.get('userName'): model.get('email');
         var password = value;
         return !passwordContainsUserName(username, password);
       } else {
--- a/src/util/RegistrationFormFactory.js
+++ b/src/util/RegistrationFormFactory.js
@@ -79,7 +79,7 @@ define([
     if (_.isString(regex)) {
       if (regex === '^[#/userName]') {
         var username = model.get('userName');
-        if (username == undefined) {
+        if (!model.has('userName')) {
           username = model.get('email');
         }
         var password = value;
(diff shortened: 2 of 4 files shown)


nikhilvenkatraman-okta:
Can we add a test that will validate the text on the verification email sent screen ?

src/RegistrationController.js
```diff
@@ -106,7 +106,7 @@ function (
         });
       } else {
         // register via activation email
-        this.model.appState.set('username', this.model.get('userName'));
+        this.model.appState.set('username', this.model.get('email'));
```
nikhilvenkatraman-okta: I saw this in the description `New self service registration logic does not use userName attribute. Instead it uses email and optionally login attributes`. Do we need to handle login attribute as well ? Will email always be available ?

uvartak-okta: when email as login feature is enabled, only email is available. Therefore email looks like a common attribute that we can use irrespective of email as login feature

hor-kanchan-okta:
@uvartak-okta There are few places referring to `userName`. Are you saying they should be changed to `email`?
https://github.com/okta/okta-signin-widget/blob/2.8/src/util/RegistrationFormFactory.js#L81
https://github.com/okta/okta-signin-widget/blob/2.8/test/unit/spec/Registration_spec.js#L38
https://github.com/okta/okta-signin-widget/blob/2.8/test/unit/spec/RegistrationFormFactory_spec.js#L66

uvartak-okta:
@hor-kanchan-okta widget codebase uses username in model that does not have to change.

nikhilvenkatraman-okta:
@uvartak-okta a follow-up on @hor-kanchan-okta's question, if email as login feature is enabled will the  `userName` attribute be used at all ?

test/unit/spec/Registration_spec.js
```diff
@@ -652,6 +652,41 @@ function (Q, _, $, OktaAuth, Backbone, Util, Expect, Beacon, RegForm, RegSchema,
           expect(setting.registration.postSubmit).toHaveBeenCalled();
         });
       });
+      itp('calls postSubmit call onSuccess assert username is same as email', function () {
```
nikhilvenkatraman-okta: nit: May want to change the name of the test to something like `asserts username is same as email and verifies  text on registration complete view`

uvartak-okta:
If email as login is enabled, registration policy does not contain userName field. Email is always present (whether feature flag is on or off). Therefore we are using email in registration verification message.

nikhilvenkatraman-okta:
In that case do we need to add the feature flag check here https://github.com/okta/okta-signin-widget/blob/2.8/src/util/RegistrationFormFactory.js#L81 ? And update the tests ?

src/util/RegistrationFormFactory.js
```diff
@@ -79,6 +79,9 @@ define([
     if (_.isString(regex)) {
       if (regex === '^[#/userName]') {
         var username = model.get('userName');
+        if (username == undefined) {
```
hor-kanchan-okta: use `===`

haishengwu-okta: it shall be sufficient to do `if (username)` 🔹

hor-kanchan-okta: It matters if both `userName` and `email` exists in the policy.

haishengwu-okta: theoretically we shall consider `null` which means `username == undefined` seems fine. but wonder we shall be more explicit like `isString and is not empty`?
anyway, minor thing.

uvartak-okta: so whats the verdict? change to === or keep it unchanged?

hor-kanchan-okta: Right, we should consider `null` as well. Then the best is to do `model.has('userName')`

src/util/RegistrationFormFactory.js
```diff
@@ -79,6 +79,9 @@ define([
     if (_.isString(regex)) {
       if (regex === '^[#/userName]') {
         var username = model.get('userName');
+        if (!model.has('userName')) {
```
nikhilvenkatraman-okta: nit: Do we need a small comment here, describing the fallback to `email` ?

test/unit/spec/Registration_spec.js
```diff
@@ -642,14 +692,49 @@ function (Q, _, $, OktaAuth, Backbone, Util, Expect, Beacon, RegForm, RegSchema,
...
+          test.form.setUserName('test@example.com');
           test.form.setPassword('Abcd1234');
           test.form.setFirstname('firstName');
           test.form.submit();
           var model = test.router.controller.model;
           spyOn(Backbone.Model.prototype, 'save').and.returnValue($.Deferred().resolve());
           model.save();
+          test.router.navigate('signin/register-complete', {trigger: true});
           expect(setting.registration.postSubmit).toHaveBeenCalled();
+          expect(model.get('userName')).toBe('test@example.com');
```
nikhilvenkatraman-okta: Not sure if this expect `expect(model.get('userName')).toBe('test@example.com');` is adding any value, since we set the userName and just validate that its set ?

PR #435, opened 2018-04-19
:seedling: adding logic for identifier first auth flows

mauriciocastillosilva-okta:
- Added new features.identifierFirst config to bootstrap the widget in identifier first auth mode
- PrimaryAuth has the logic not to show password field or not, depending on the config option for identifier first
- Factors dropdown shows Password correctly now

Resolves: OKTA-166235

Screenshots:
identifierfirst-login

identifierfirst-dropdownoptions


mauriciocastillosilva-okta:
Updated so `identifierFirst` and `idpDiscovery` work together. 
- if `identifierFirst` config is true, then `idpDiscovery` has to be true as well
- After we know Okta is the idp, if `identifierFirst` is true, we will call authClient.signIn with the username only

src/LoginRouter.js
```diff
@@ -154,7 +154,7 @@ function (BaseLoginRouter,
     ],
 
     defaultAuth: function() {
-      if(this.settings.get('features.idpDiscovery')) {
+      if(this.settings.get('features.idpDiscovery') || this.settings.get('features.identifierFirst')) {
```
rchild-okta: Do we still need this?

mauriciocastillosilva-okta: ah, good point. We do not need it, and probably removing it will also enforce better the dependency. Will remove.

PR #436, opened 2018-04-21
:seedling: passwordlessAuth support without idpDiscovery

mauriciocastillosilva-okta:
- Renamed identifierFirst config to passwordlessAuth, so it is clear what is the purpose
- idpDiscovery and passwordlessAuth do not depend on each other, they can be enabled independently
- With passwordlessAuth widget will render a username-only form

Resolves: OKTA-166235

passwordlessauth-true


src/models/Settings.js
```diff
@@ -312,11 +312,6 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
       else if (BrowserFeatures.corsIsNotSupported()) {
         this.callGlobalError(new UnsupportedBrowserError(Okta.loc('error.unsupported.cors')));
       }
-      else if (options.features) {
-        if (options.features.identifierFirst && !options.features.idpDiscovery) {
-          this.callGlobalError(new ConfigError(Okta.loc('error.invalid.identifierFirst')));
```
rchild-okta: Followup for this PR or later - we should remove this property from i18n as well

PR #442, opened 2018-05-01
:seedling: Expose AuthJS on SignIn object

jmelberg-okta:
### Description
Exposes the underlying `@okta/okta-auth-js` object used when initializing the widget.

Currently the Widget wraps AuthJS and exposes a subset of methods needed by developers. This PR aims to expose the entire object for the most flexibility.

### Resolves
[OKTA-168960](https://oktainc.atlassian.net/browse/OKTA-168960) & [REQ-13006](https://oktainc.atlassian.net/browse/REQ-13006)

### DIFF
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -130,10 +130,18 @@ var OktaSignIn = (function () {
       .fail(error);
     }
 
+    /**
+     * Returns the Okta Auth JS auth client.
+     */
+    function getAuthClient() {
+      return authClient;
+    }
+
     // Properties exposed on OktaSignIn object.
     return {
       renderEl: render,
       signOut: closeSession,
+      authClient: getAuthClient(),
       idToken: {
         refresh: refreshIdToken
       },
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -130,18 +130,11 @@ var OktaSignIn = (function () {
       .fail(error);
     }
 
-    /**
-     * Returns the Okta Auth JS auth client.
-     */
-    function getAuthClient() {
-      return authClient;
-    }
-
     // Properties exposed on OktaSignIn object.
     return {
       renderEl: render,
       signOut: closeSession,
-      authClient: getAuthClient(),
+      authClient: authClient,
       idToken: {
         refresh: refreshIdToken
       },
--- a/test/unit/spec/OktaSignIn_spec.js
+++ b/test/unit/spec/OktaSignIn_spec.js
@@ -1,0 +1,78 @@
+/* eslint max-params:[0, 2], max-len:[2, 180] */
+define([
+  'widget/OktaSignIn',
+  'helpers/util/Expect'
+],
+function (Widget, Expect) {
+
+  var url = 'https://foo.com';
+  var signIn = new Widget({
+    baseUrl: url
+  });
+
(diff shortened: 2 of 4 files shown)


src/widget/OktaSignIn.js
```diff
@@ -130,10 +130,18 @@ var OktaSignIn = (function () {
       .fail(error);
     }
 
+    /**
+     * Returns the Okta Auth JS auth client.
+     */
+    function getAuthClient() {
+      return authClient;
```
jmelberg-okta: My thinking could be off here, so I'd appreciate some feedback. 

I wanted to keep this as a `getter` so we could easily modify the `authClient` object as needed. For example, if we decide to omit the `fingerPrint` method, we _could_ easily transform the `authClient` before returning it.

robertjd: Technically you can use [property getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) to solve the same problem.  If we feel like we want to hide auth JS functionality in the future we should be having a bigger conversation about why we're doing this in the first place.

jmelberg-okta: Property getters would be great if this library already used that design, so I'm hesitant on updating this PR to reflect that.

Regarding hiding AuthJS functionality - I'm not sure I see a valid case **today** for needing to do so, but pulling it out into a `getter` allows us to in the future.

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -0,0 +1,78 @@
+/* eslint max-params:[0, 2], max-len:[2, 180] */
```
santhoshbalakrishnan: Would let @mauriciocastillosilva-okta comment on this test. There was a specific use case around how unit tests are being used in the widget.

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -0,0 +1,78 @@
...
+    baseUrl: url
+  });
+
+  Expect.describe('OktaSignIn initialization', function () {
+    Expect.describe('At the root level', function () {
+      it('has a renderEl method', function () {
+        expect(signIn.renderEl).toBeDefined();
+      });
+      it('has a signOut method', function () {
+        expect(signIn.renderEl).toBeDefined();
```
mauriciocastillosilva-okta: this should be `signIn.signOut` instead of `renderEl`

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -0,0 +1,78 @@
...
+      });
+      it('has an authClient method', function () {
```
mauriciocastillosilva-okta: more line an `authClient object` I think, right?

test/unit/spec/OktaSignIn_spec.js
mauriciocastillosilva-okta: so, on your case Santhosh, you were changing a model, so it is better to test the functionality of that change in a flow where it is used
But this change is for the widget object we create, which needs to return some objects/functions... I think this is good. We could add later e2e tests for each of the functions the widget object has, but I think this is a good start

Does that make sense?

santhoshbalakrishnan: Still a little fuzzy to me :( Will sync up offline. But we're good to go on this PR!

PR #443, opened 2018-05-01
:seedling: Adds relayState to sign in widget config parameters

santhoshbalakrishnan:
## Description

- Added relayState to widget config. Eventually, it will be an optional state value that is persisted for the lifetime of the authentication transaction.
- Added tests for Settings Model.

## Reviewers

- @rchild-okta 
- @mauriciocastillosilva-okta 
- @uvartak-okta 
- @hor-kanchan-okta 
- @manueltanzi-okta

### DIFF
--- a/src/ForgotPasswordController.js
+++ b/src/ForgotPasswordController.js
@@ -72,8 +72,7 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, Util, ContactSu
       },
       save: function () {
         var self = this;
-        var urlParams = Util.getJsonFromUrl(window.location.search);
-        var relayState = urlParams.fromURI || '';
+        var relayState = this.settings.get('relayState');
         this.startTransaction(function(authClient) {
           return authClient.forgotPassword({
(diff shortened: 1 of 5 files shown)


rchild-okta:
Agree with @mauriciocastillosilva-okta - I would expect this PR to actually do something with the relay state. Specifically, updating [this logic](https://github.com/okta/okta-signin-widget/blob/7dded84452e04f5e2d5c6ae6e2b776bb2cb54609/src/ForgotPasswordController.js#L75-L83) to use it.

mauriciocastillosilva-okta:
yeah, here is an example of a [test](https://github.com/santhoshbalakrishnan/okta-signin-widget/blob/8c38bd01bee00e749511735edd3c4d380251c091/test/unit/spec/ForgotPassword_spec.js#L226) that has uses the relayState

santhoshbalakrishnan:
@rchild-okta @mauriciocastillosilva-okta Had a chat with @uvartak-okta. He'll add this change as a part of his so that we have all the change in one place.

PR #444, opened 2018-05-02
:seedling: Capture app context in relayState property for registration & recovery flows.

uvartak-okta:
This change captures app context in relayState property. This relayState is consumed by registration and recovery flows. Updated tests.

Resolves OKTA-168553

### DIFF
--- a/src/ForgotPasswordController.js
+++ b/src/ForgotPasswordController.js
@@ -72,8 +72,7 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, Util, ContactSu
       },
       save: function () {
         var self = this;
-        var urlParams = Util.getJsonFromUrl(window.location.search);
-        var relayState = urlParams.fromURI || '';
+        var relayState = this.settings.get('relayState');
         this.startTransaction(function(authClient) {
           return authClient.forgotPassword({
             username: self.settings.transformUsername(self.get('username'), Enums.FORGOT_PASSWORD),
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -129,10 +129,6 @@ function (
         }
       });
     },
-    getRelayStateData: function () {
-      var urlParams = Util.getJsonFromUrl(window.location.search);
-      return urlParams.fromURI || '';
-    },
     createRegistrationModel: function (modelProperties) {
       var self = this;
       var Model = Okta.Model.extend({
@@ -147,7 +143,7 @@ function (
           var data = Okta.Model.prototype.toJSON.apply(this, arguments);
           return {
             userProfile: data,
-            relayState: self.getRelayStateData()
+            relayState: this.settings.get('relayState')
           };
         },
         parse: function(resp) {
--- a/src/util/RouterUtil.js
+++ b/src/util/RouterUtil.js
@@ -118,9 +118,12 @@ function (Okta, Util, OAuth2Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
 
       var successData = {
         user: res._embedded.user,
-        type: res.type,
-        relayState: res.relayState
+        type: res.type
       };
+
+      if (res.relayState) {
+        successData.relayState = res.relayState;
+      }
 
       if (res.type === Enums.SESSION_STEP_UP) {
         var targetUrl = res._links && res._links.next && res._links.next.href;
--- a/src/util/RouterUtil.js
+++ b/src/util/RouterUtil.js
@@ -118,7 +118,8 @@ function (Okta, Util, OAuth2Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
 
       var successData = {
         user: res._embedded.user,
-        type: res.type
+        type: res.type,
+        relayState: res.relayState
       };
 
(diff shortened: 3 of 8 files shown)


src/ForgotPasswordController.js
```diff
@@ -72,8 +72,7 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, Util, ContactSu
       },
       save: function () {
         var self = this;
-        var urlParams = Util.getJsonFromUrl(window.location.search);
-        var relayState = urlParams.fromURI || '';
+        var relayState = this.settings.get('relayState');
```
mauriciocastillosilva-okta: this breaks the previous functionality right? I'm guessing the settings param is preferred over the url, but we should still have a criteria here to chose the relayState in that order
1 if in settings, use it
2 not in settings, use the one from url
3 no relayState? then empty string

Or is there a reason we don't want to support the url relayState anymore?

src/RegistrationController.js
```diff
@@ -147,7 +143,7 @@ function (
           var data = Okta.Model.prototype.toJSON.apply(this, arguments);
           return {
             userProfile: data,
-            relayState: self.getRelayStateData()
+            relayState: this.settings.get('relayState')
```
mauriciocastillosilva-okta: same comment here, we are not supporting the url relayState. Are we guaranteeing we are not breaking any flow? I rather have the url as a second option (look at my other comment)

src/util/RouterUtil.js
```diff
@@ -118,7 +118,8 @@ function (Okta, Util, OAuth2Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
 
       var successData = {
         user: res._embedded.user,
-        type: res.type
+        type: res.type,
+        relayState: res.relayState
```
mauriciocastillosilva-okta: is this always defined? We are adding it to all success callbacks, so do we know if all SUCCESS api responses have this value? if not all have the value, are we ok with sending undefined/empty string in the response?
A consumer of the widget will see in the data we pass to the success function the usual data + an relayState (empty or with a value). So just want to be sure that is the intention

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -230,20 +233,13 @@ function (Q, _, $, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
             data: {
               'username': 'foo',
               'factorType': 'EMAIL',
-              'relayState': ''
```
mauriciocastillosilva-okta: I see... now since this is undefined we don't get an empty string. It seems good, but we are again breaking a current pattern, are we good with it? This one might be trickier if someone out there is comparing for an empty string, and now we are not even returning the `relayState` key in the object...

src/util/RouterUtil.js
uvartak-okta: main-loginpage.js check is:

if (featureRequiresStepUp && res.stepUp) {
      res.stepUp.finish();
    } else if(res.idpDiscovery) {
      res.idpDiscovery.redirectToIdp(config.fromUri);
    } else if (res.relayState) {
      res.session.setCookieAndRedirect(res.relayState);
    } else {
      // Normal flow: Navigate to the sessionCookieRedirect endpoint
      res.session.setCookieAndRedirect(config.redirectUri);
    }

this will work only if relayState is populated.

test/unit/spec/ForgotPassword_spec.js
uvartak-okta: @santhoshbalakrishnan  can you comment on this? I think we are good.. but I will let Santhosh confirm.

src/ForgotPasswordController.js
mauriciocastillosilva-okta: okay... but then more to the point of *not* removing the functionality that already existed right? Unless as I was asking that is the intention

santhoshbalakrishnan: This piece of code was added 14 days back and is not an existing functionality in any widget release.

test/unit/spec/ForgotPassword_spec.js
santhoshbalakrishnan: While making the POST call to authClient, it removes any undefined parameter. If relayState was undefined, it's best to leave it undefined. If that is not sent with the original request, it won't be added back in the response.

src/ForgotPasswordController.js
nikhilvenkatraman-okta: Yes, as @santhoshbalakrishnan mentioned the code to get fromURI form `window.location.search` was added recently but after recent discussions this change was preferred.

src/util/RouterUtil.js
nikhilvenkatraman-okta: @uvartak-okta @mauriciocastillosilva-okta 
Would it be best to do a check to see the if relayState is present and only then add to the successData object ?

nikhilvenkatraman-okta:
``` 
var successData = {
        user: res._embedded.user,
        type: res.type
};
if (res.relayState) {
   successData.relayState = res.relayState;
}

```

nikhilvenkatraman-okta:
rebase before merge!

PR #446, opened 2018-05-04
:bug: Clean up currentLanguage in Bundles on remove()

joelfranusic-okta:
This allows for changing of `i18n` strings after `signInWidget.remove();` has been called.


jmelberg-okta:
@joelfranusic-okta,

Thanks for this change! Can you add a test to this to ensure these changes work correctly?

I'd probably start by updating the [`basic_spec`](https://github.com/okta/okta-signin-widget/blob/master/test/e2e/specs/basic_spec.js) to change the `i18n` settings, then assert a label on the page has updated.

@mauriciocastillosilva-okta and/or @nikhilvenkatraman-okta - Any other pointers for testing this kind of change?

jmelberg-okta:
Included with #501

PR #447, opened 2018-05-04
Updates logger to extend warnings outside of Okta's window.okta

jmelberg-okta:
### Description
- Updates extended `Logger.js` utility methods to use a custom `window.console`. Previously, the Widget used Courage's `Logger` class, which performes a `window.okta` evaluation before logging to the console. This made console logs visible **only** when `window.okta` was defined - which appears to only be for Okta's login page.
  ```
    if (window.console && window.okta && window.okta.debug) {
      window.console[level].apply(window.console, args);
    }
  ```
- This provides the ground work necessary for adding a new **debug** mode to the widget for development builds.

**Note:** Most of the `Logger` class was extracted from Courage.

Resolves: [OKTA-169946](https://oktainc.atlassian.net/browse/OKTA-169946)


test/unit/spec/LoginRouter_spec.js
```diff
@@ -170,21 +170,9 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Exp
       });
     }
 
-    function expectUnexpectedFieldLog(arg1) {
-      expect(Logger.warn).toHaveBeenCalledWith('Field not defined in schema', arg1);
-    }
-
-    it('throws a ConfigError if unknown option is passed as a widget param', function () {
```
jmelberg-okta: The description of this test (and below a bit) do not match the assertions.

Since we do not inject additional params anywhere, we should be OK **not** throwing an error here. As an alternative, we *could* add conditional logic [here](https://github.com/okta/okta-signin-widget/blob/master/src/widget/OktaSignIn.js#L87).

test/unit/spec/LoginRouter_spec.js
```diff
@@ -290,22 +278,15 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Exp
...
-      spyOn(Logger, 'warn');
       var fn = function () {
-        setup({ globalErrorFn: errorSpy, foo: 'bar' });
+        setup({ globalErrorFn: errorSpy, baseUrl: undefined });
       };
       expect(fn).not.toThrow();
-      expectUnexpectedFieldLog('foo');
+      var err = errorSpy.calls.mostRecent().args[0];
+      expect(err.name).toBe('CONFIG_ERROR');
+      expect(err.message).toEqual('"baseUrl" is a required widget parameter');
```
jmelberg-okta: Updated assertions to match the test description

test/unit/spec/LoginRouter_spec.js
```diff
@@ -5,7 +5,7 @@ define([
   'backbone',
   'shared/util/Util',
   'util/CryptoUtil',
-  'util/Logger',
+  'shared/util/Logger',
```
jmelberg-okta: Since we are relying on Courage's logger, we should reference it directly.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -174,7 +174,7 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Exp
       expect(Logger.warn).toHaveBeenCalledWith('Field not defined in schema', arg1);
     }
 
-    it('throws a ConfigError if unknown option is passed as a widget param', function () {
+    it('logs a ConfigError error if unknown option is passed as a widget param', function () {
       spyOn(Logger, 'warn');
```
jmelberg-okta: Fun fact: We're **not** actually logging here.

There is a [_known_](https://oktainc.atlassian.net/browse/OKTA-170705) issue when instantiating the `BaseRouter`, where we're logging `Field not defined in schema: ${var}` for every widget param (`baseUrl`, `assets`, etc.).

Once that is resolved, we need to update these tests.

jmelberg-okta: They will not. The warning `Field not defined in schema` will only be visible when `window.okta && window.okta.debug` are defined, as it is still using Courage's `Logger` class.

With the new additions in this PR, any `Logger.X` method called within the Widget's source will now emit a console notification. We have only two usages of that today:

- [Use "scopes" instead of "scope"](https://github.com/okta/okta-signin-widget/blob/d168cf84402b503b17a7b4b1e967a6bf02ed03da/src/models/Settings.js#L443)
- [Use "i18n" instead of "labels" and "country"](https://github.com/okta/okta-signin-widget/blob/d168cf84402b503b17a7b4b1e967a6bf02ed03da/src/models/Settings.js#L449)

test/unit/spec/LoginRouter_spec.js
jmelberg-okta: I wanted to make this a bit more clear in our tests that we are expecting a `console.warn` message from our model validation inside of Courage, as it only get passed up to a user/developer in a specific environment. Our new `Logger` class does **not** call this method.

Instead of removing/modifying the existing tests that make this assertion, do you have any recommendations?

test/unit/spec/LoginRouter_spec.js
mauriciocastillosilva-okta: got it, thanks for the explanation. model validation still uses Courage's Logger, makes sense now 👍

test/unit/spec/LoginRouter_spec.js
mauriciocastillosilva-okta: I think this is good. I would add a comment explaining this, so we know why we use Courage's Logger instead of the local.

PR #452, opened 2018-05-11
Adds dev logging for all dev builds

jmelberg-okta:
## Description
Publishes `okta-sign-in.js`, a non-minified version of the Widget with helpful console warnings.

Bootstrapping the widget will produce the following ⚠️ message for developers:
    screen shot 2018-05-11 at 3 22 55 pm

**Goal**: Follow the [Vue.js](https://github.com/vuejs/vue/blob/dev/dist/README.md#explanation-of-build-files) approach explaining their **build files**, with clear indication on when/how to use [**Development vs. Production Mode**](https://github.com/vuejs/vue/blob/dev/dist/README.md#development-vs-production-mode
). 

## Whats Next
Update CommonJS and ES Module builds to preserve `process.env.NODE_ENV` to determine whether to run in **Development** or **Production** mode.

### File Size
| Asset | Size |
| :----: | :---: |
| `okta-sign-in.min.js` | 1 MB  |
| `okta-sign-in.min.js.map` |  6.8 MB |
| `okta-sign-in.js` |  2.67 MB  |
| `okta-sign-in.js.map` |  3.68 MB |


test/unit/spec/OktaSignIn_spec.js
```diff
@@ -1,13 +1,31 @@
...
+  });
+
+  Expect.describe('Debug Mode', function () {
+    it('logs a warning message on page load', function () {
+      var debugMessage =
+      `
+        The Okta Sign-In Widget is running in development mode.
+        When you are ready to publish your app, embed the minified version to turn on production mode.
+        See: https://developer.okta.com/code/javascript/okta_sign-in_widget#cdn
+      `;
```
robertjd: Can we move these strings into our i18n library?

src/widget/OktaSignIn.js
```diff
@@ -167,6 +167,14 @@ var OktaSignIn = (function () {
     OktaAuth = require('@okta/okta-auth-js/jquery');
     Util = require('util/Util');
     LoginRouter = require('LoginRouter');
+    Logger = require('util/Logger');
+
+    Logger.warn(
```
bdemers: Can this be move into `util/logger.js` The actual logging concerns ideally shouldn't leak into this class?

jmelberg-okta: I'm interpreting "logging concerns" a couple of different ways, so hopefully I address your comment below.

We want to output this log statement **only** when the SIW is instantiated. Therefore, adding this message to the `Logger` class constructor would produce this error message whenever it is imported and used in another file.

Or - are you saying pull out the message into a separate method: For example - `Logger.initialWarning()`. I do not have a preference with this one, but this design does seem like an anti-pattern. I'll defer to @robertjd for a decision here.

In the future, this string will be localized and reduced to something similar to:
```javascript
Logger.warn(loc('console.developer.inital.warning'));
```
^ Where the `loc()` param is a key lookup.

bdemers: it seems like an implementation detail of the logger class (not the sign in widget). When the logger object is created this message could be recorded. (meaning you wouldn't need a `initialWarning` function)

jmelberg-okta: Gotcha. Problem with this approach is whenever this class is imported, we'd log that error.  Take the following example:

```
1. Initialize the widget (Log #1)
2. Pass in unsupported parameter (Log #2)
3. Resort the the "default" success/error handler (Log #3)
...
```

```javascript
// Widget instantiation code

// Logs #1 and #2
var signIn = new OktaSignIn({
  baseUrl: 'https://foo.com',
  fakeKey: 'bar'
});

// Logs #3
signIn.renderEl({el: '#domElement'});
```
Each scenario numbered above is handled in separate files, requiring us to import in the `Logger` class.

webpack.common.config.js
```diff
@@ -2,7 +2,7 @@ var path      = require('path');
 var EMPTY     = 'widget/empty';
 var TARGET_JS = path.resolve(__dirname, 'target/js/');
 var SHARED_JS = TARGET_JS + '/shared';
-var VENDOR = path.resolve(__dirname, 'packages/@okta/courage/src/vendor');
```
rchild-okta: Interesting - why isn't this caught in lint (no-undefined-vars)?

webpack.plugins.config.js
```diff
@@ -0,0 +1,53 @@
+var webpack = require('webpack');
+var _       = require('underscore');
+
+var plugins = {
+  envPlugin: function(type) {
+    // Sepecify which environment variables
+    return new webpack.DefinePlugin({
+      'process.env': {
```
rchild-okta: Related to the other comment - let's just use `DEBUG` instead.

This is actually a bit strange - the expectation from the other line is that we're using the [EnvironmentPlugin](https://webpack.js.org/plugins/environment-plugin/), which is a shortcut for the DefinePlugin and env variables. In that case I'd expect that there'd be changes to our build commands to set different env variables when running different stages, and the question would be if we wanted to use node variables in client side code. But, here we're not doing that, and also having to drop into (or explicitly define) process.env variables.

webpack.plugins.config.js
```diff
@@ -0,0 +1,53 @@
+var webpack = require('webpack');
```
rchild-okta: Currently, the other `webpack.*.config.js` files follow the pattern of exporting webpack config objects - common.config is the base, and the others extend it. This one's different - it's exporting helper functions that return pieces of config.

Let's create a `webpack` directory in `buildtools/`, move this there, and rename it to something like `plugins.js`. In general, there are a bunch of things we can do to clean these webpack files, and this is a good start.

webpack.plugins.config.js
```diff
@@ -0,0 +1,53 @@
...
+var _       = require('underscore');
+
+var plugins = {
```
rchild-okta: It might be nicer to just export a function with configuration options rather than having to pick and choose. The way it's currently setup:
- There's a default plugins specified in common
- But, it's overwritten in several places
- Other places extend

It's a bit confusing to try to figure out which plugins are exposed where. Instead, let's just set the plugins at each end state, where config would look something like this:

```javascript
plugins: plugins({
  debug: false,
  uglify: true, // and if uglify, that means that we add the banner automatically
  // etc
})
```

src/widget/OktaSignIn.js
jmelberg-okta: @rchild-okta - In practice, it make sense to include it in `Logger.js` and output the message once that file loads. A problem arises when we're attempting to validate that `Logger.warn` was called via our jasmine tests.

After some work, I wasn't able to capture this call via a `spyOn` method, as `Logger.warn` origionally gets called from `Expect.js`.

webpack.common.config.js
jmelberg-okta: Seems like a few of our rules weren't being applied to files located at the root. I've updated this PR to now include `webpack.*` files.

src/util/Logger.js
```diff
@@ -13,7 +13,8 @@
 define(function () {
 
   function log(level, args) {
-    if (window.console) {
+    // Only log statements in development mode
+    if (DEBUG) {
```
jmelberg-okta: Based on which versions of the Widget [we support](https://help.okta.com/en/prod/Content/Topics/Miscellaneous/Platforms_Browser_OS_Support.htm), we don't need the `window.console` check anymore.

src/widget/OktaSignIn.js
jmelberg-okta: Per offline convo, we're going to keep this `Logger` statement in its current location

test/unit/spec/OktaSignIn_spec.js
nbarbettini: I don't think we should spend time translating our debug messages. It would be a lot of extra work. If you feel really strongly about this @robertjd let me know.

robertjd: Ah, so I wasn't necessarily suggesting that we need to translate these.  I was just looking for a place where we could enumerate the strings so that we aren't repeating them.

buildtools/webpack/plugins.js
```diff
@@ -0,0 +1,80 @@
+/* global module module */
+/*eslint camelcase: ["error", {properties: "never"}]*/
+
+var fs      = require('fs');
```
rchild-okta: nit - we're on node 8 so can use es6, so here:
```javascript
// const + destructuring assignments
const { readFileSync } = require('fs');
```

buildtools/webpack/plugins.js
```diff
@@ -0,0 +1,80 @@
...
+var webpack = require('webpack');
+var _       = require('underscore');
+
+var license = fs.readFileSync('src/widget/copyright.txt', 'utf8');
```
rchild-okta: This is somewhat ambiguous because it depends on where this is being run (root dir) - nicer to be explicit i.e.

```javascript
const { join } = require('path');

const licence = readFileSync(join(__dirname, '../../src/widget/copyright.txt'), 'utf8');
```

buildtools/webpack/plugins.js
```diff
@@ -0,0 +1,80 @@
...
+
+var devMode = function(mode = true) {
```
rchild-okta: For functions generally prefer using named, i.e.

```javascript
function devMode(mode = true) {
  // etc
}
```

buildtools/webpack/plugins.js
```diff
@@ -0,0 +1,80 @@
...
+};
+
+var uglify = function() {
+  return new webpack.optimize.UglifyJsPlugin({
+    compress: {
+      warnings: false,
+      // Drop all console.* and Logger statements
+      drop_console: true,
+      drop_debugger: true,
+      pure_funcs: [
```
rchild-okta: nice! Is there a way to do Logger.*?

buildtools/webpack/plugins.js
```diff
@@ -0,0 +1,80 @@
...
+        return comment.value.indexOf(exception) !== -1;
+      });
+
+      return (isLicense || isException) && !isOkta;
+    }
+  });
+};
+
+var banner = function(license) {
+  // Add a single Okta license after removing others
```
rchild-okta: Since we're only using license here, let's just add it here rather than having the global var

buildtools/webpack/plugins.js
```diff
@@ -0,0 +1,80 @@
...
+  return new webpack.BannerPlugin(license);
+};
+
+function plugin(options = {}) {
```
rchild-okta: nit: name of file is "plugins" but we're exporting "plugin" - it doesn't affect anything, but it's better to be consistent.

buildtools/webpack/plugins.js
```diff
@@ -0,0 +1,80 @@
...
+  var plugins = [];
+  if (options.uglify) {
```
rchild-okta: Are we ever going to want debug + uglify? If not, could just pass in env (production, debug) or just a boolean isProduction.

Gruntfile.js
```diff
@@ -493,6 +495,7 @@ module.exports = function (grunt) {
         'copy:e2e',
         'copy:e2e-pages',
         'exec:build-e2e-app',
+        'exec:build-e2e-dev-app',
```
rchild-okta: What is actually happening here? Is it overwriting the production app bundle with the dev bundle?

buildtools/webpack/plugins.js
jmelberg-okta: I wish! Unfortunately we need to be explicit about the methods here.

rchild-okta: In that case, we should probably add in all the other ones as well right? We might not be using them now, but it will be a surprise to the developer that adds an info log line and it shows up in prod.

jmelberg-okta: Thats fair. When I was thinking about this earlier I choose **not** to add more time to the build, as `pure_funcs` for all the `Logger` methods take an extra 12 seconds. But a confused developer is probably worse in this case.

test/unit/spec/OktaSignIn_spec.js
nbarbettini: Ah, I see. I'd advocate for them staying here, since it's really only in the code base once (not counting tests). But I will defer to @jmelberg-okta 🙂

webpack.dev.config.js
```diff
@@ -1,5 +1,10 @@
-var common = require('./webpack.common.config');
+/* global module */
+
+var common    = require('./webpack.common.config');
 var devConfig = common('okta-sign-in.js');
+var plugins   = require('./buildtools/webpack/plugins');
```
rchild-okta: nit: require statements should be above others (group them together) - there's a lint rule for this, but looks like we haven't enabled it here yet.

jmelberg-okta: There are a few common eslint rules we aren't including, so I'll submit a new PR with `.eslintrc` updates soon!

PR #460, opened 2018-05-19
:arrow_up: update i18n and improve U2F errors

mauriciocastillosilva-okta:
- Update i18n to latest
- Improve U2F errors

U2F Instructions:
widget-u2f-instructions

U2F Not supported:
widget-u2f-notsupported

U2F not supported and only factor:
widget-u2f-notsupported-onefactor


src/VerifyU2FController.js
```diff
@@ -42,7 +41,7 @@ function (Okta, FormController, FormType, FooterSignout, Q, HtmlErrorMessageView
   }
 
   return FormController.extend({
-    className: 'verify-u2f',
+    className: 'mfa-verify verify-u2f',
```
nikhilvenkatraman-okta: Curious, why we need to add this class name `mfa-verify`

mauriciocastillosilva-okta: It was missing. In theory our verify views should have it (https://github.com/okta/okta-signin-widget/blob/master/assets/sass/okta-theme.scss#L796)

PR #463, opened 2018-05-23
 :bug: Fix concurrent fingerprinting

magizh-okta:
The signin widget runs the device fingerprinting code in an iframe and proceeds when the iframe posts back to the parent an event with type `FingerprintAvailable`. The problem is that the widget assumes that one one iframe will exist at a time and therefore only one `FingerprintAvailable` event will occur. This is a bad assumption. It is possible for multiple requests to request fingerprints at the same time but all requests will listen for the same `FingerprintAvailable` event regardless of which iframe generated it. This means that all concurrent fingerprint requests will generate unique fingerprints but all will use the first one that is returned. Because the fingerprints have one time tokens, this will cause failures

Checking iframe contentWindow with event sources makes sure that we only process messages received from the right iframe.

### DIFF
--- a/src/util/DeviceFingerprint.js
+++ b/src/util/DeviceFingerprint.js
@@ -10,11 +10,14 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-/* eslint complexity: [2, 7] */
+/* eslint complexity: [2, 8] */
 
 define(['vendor/lib/q', 'okta/jquery'], function (Q, $) {
 
   return {
+    _isMessageFromCorrectSource: function($iframe, event) {
+      return event.source === $iframe[0].contentWindow;
+    },
     generateDeviceFingerprint: function (oktaDomainUrl, element) {
       if (!navigator.userAgent) {
         return Q.reject('user agent is not defined');
@@ -23,6 +26,7 @@ define(['vendor/lib/q', 'okta/jquery'], function (Q, $) {
       }
 
       var deferred = Q.defer();
+      var self = this;
 
       // Create iframe
       var $iframe = $('<iframe>', {
@@ -46,6 +50,9 @@ define(['vendor/lib/q', 'okta/jquery'], function (Q, $) {
       }
 
       function onMessageReceivedFromOkta(event) {
+        if (!self._isMessageFromCorrectSource($iframe, event)) {
+          return;
+        }
         if (!event || !event.data || event.origin != oktaDomainUrl) {
           handleError('no data');
           return;
--- a/src/util/DeviceFingerprint.js
+++ b/src/util/DeviceFingerprint.js
@@ -15,13 +15,17 @@
 define(['vendor/lib/q', 'okta/jquery'], function (Q, $) {
 
   return {
+    __getUserAgent: function() {
+      return navigator.userAgent;
+    },
     _isMessageFromCorrectSource: function($iframe, event) {
       return event.source === $iframe[0].contentWindow;
     },
     generateDeviceFingerprint: function (oktaDomainUrl, element) {
-      if (!navigator.userAgent) {
+      var userAgent = this.__getUserAgent();
+      if (!userAgent) {
         return Q.reject('user agent is not defined');
-      } else if (isWindowsPhone()) {
+      } else if (isWindowsPhone(userAgent)) {
         return Q.reject('device fingerprint is not supported on Windows phones');
       }
 
@@ -34,8 +38,8 @@ define(['vendor/lib/q', 'okta/jquery'], function (Q, $) {
       });
       $iframe.appendTo(element);
 
-      function isWindowsPhone() {
-        return navigator.userAgent.match(/windows phone|iemobile|wpdesktop/i);
+      function isWindowsPhone(userAgent) {
+        return userAgent.match(/windows phone|iemobile|wpdesktop/i);
       }
 
       function removeIframe() {
--- a/src/util/DeviceFingerprint.js
+++ b/src/util/DeviceFingerprint.js
@@ -15,14 +15,14 @@
 define(['vendor/lib/q', 'okta/jquery'], function (Q, $) {
 
   return {
-    __getUserAgent: function() {
+    getUserAgent: function() {
       return navigator.userAgent;
     },
-    _isMessageFromCorrectSource: function($iframe, event) {
+    isMessageFromCorrectSource: function($iframe, event) {
       return event.source === $iframe[0].contentWindow;
     },
     generateDeviceFingerprint: function (oktaDomainUrl, element) {
-      var userAgent = this.__getUserAgent();
+      var userAgent = this.getUserAgent();
       if (!userAgent) {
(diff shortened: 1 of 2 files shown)


src/util/DeviceFingerprint.js
```diff
@@ -10,11 +10,14 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-/* eslint complexity: [2, 7] */
+/* eslint complexity: [2, 8] */
 
 define(['vendor/lib/q', 'okta/jquery'], function (Q, $) {
 
   return {
+    _isMessageFromCorrentIframe: function($iframe, event) {
```
dannavarro-okta: nit: Did you mean`Current` instead of `Corrent`? Either way, what if we just generalize the method name to `_isMessageFromCorrectSource`

src/util/DeviceFingerprint.js
```diff
@@ -23,6 +26,7 @@ define(['vendor/lib/q', 'okta/jquery'], function (Q, $) {
       }
 
       var deferred = Q.defer();
+      var self = this;
```
dannavarro-okta: why do you need to do this? Can't you just call `this.METHOD` or is this a weird js thing?

test/unit/spec/DeviceFingerprint_spec.js
```diff
@@ -25,6 +26,14 @@ function ($, Q, Expect, $sandbox, DeviceFingerprint) {
       navigator.__defineGetter__('userAgent', function () { return userAgent; });
     }
 
+    function byPassIframeCheck() {
```
dannavarro-okta: nit: `bypassMessageSourceCheck`. Bypass is one word

test/unit/spec/DeviceFingerprint_spec.js
```diff
@@ -104,5 +116,21 @@ function ($, Q, Expect, $sandbox, DeviceFingerprint) {
...
+          done.fail('Fingerprint promise should not have been resolved');
+        })
+        .fail(function () {
+          done.fail('Fingerprint promise should not have been rejected');
+        });
+      Util.mockSetTimeout();
+      setTimeout(function() {
+        // give it time to check if promise resolves or rejects.
+        done();
+      }, 1000);
```
dannavarro-okta: does this mean wait for 1 second before succeeding? Is that a guaranteed time for the promise to return? Seems strange to have a wait in a unit test

src/util/DeviceFingerprint.js
dannavarro-okta: cool thanks for clarifying 👍

test/unit/spec/DeviceFingerprint_spec.js
magizh-okta: no we dont actually wait 1sec. We fake waiting for 1 sec using the mock Util.

src/util/DeviceFingerprint.js
```diff
@@ -10,11 +10,14 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-/* eslint complexity: [2, 7] */
+/* eslint complexity: [2, 8] */
 
 define(['vendor/lib/q', 'okta/jquery'], function (Q, $) {
 
   return {
+    _isMessageFromCorrectSource: function($iframe, event) {
```
mauriciocastillosilva-okta: we don't really need to return this function right? We can move it above the `return` statement...
Oh, or are you doing this just for the tests? that is also why you actually added this function instead of just adding the logic inside `onMessageReceivedFromOkta` right?

test/unit/spec/DeviceFingerprint_spec.js
```diff
@@ -19,18 +18,21 @@ function ($, Q, Expect, $sandbox, DeviceFingerprint) {
       window.postMessage(JSON.stringify(message), '*');
     }
 
-    function mockUserAgent(userAgent) {
-      navigator = new Object(); // eslint-disable-line no-global-assign
-      navigator.__proto__ || (navigator.__proto__ = _navigator);
-      navigator.__defineGetter__('userAgent', function () { return userAgent; });
+    function mockUserAgentCheck(userAgent) {
+      spyOn(DeviceFingerprint, '__getUserAgent').and.callFake(function() {
```
magizh-okta: The prev impl of this was wrong. The reset in afterEach dint happen correctly. Seems like it is not good to modify window.navigator for test and hence moved getting userAgent to a separate function to mock it correctly.

test/unit/spec/DeviceFingerprint_spec.js
```diff
@@ -19,18 +18,21 @@ function ($, Q, Expect, $sandbox, DeviceFingerprint) {
       window.postMessage(JSON.stringify(message), '*');
     }
 
-    function mockUserAgent(userAgent) {
-      navigator = new Object(); // eslint-disable-line no-global-assign
-      navigator.__proto__ || (navigator.__proto__ = _navigator);
-      navigator.__defineGetter__('userAgent', function () { return userAgent; });
+    function mockUserAgentCheck(userAgent) {
```
mauriciocastillosilva-okta: I think we can leave this as `mockUserAgent` since it is still doing that for the test purposes

test/unit/spec/DeviceFingerprint_spec.js
```diff
@@ -104,5 +109,23 @@ function ($, Q, Expect, $sandbox, DeviceFingerprint) {
...
+      mockIFrameMessages(true);
+      DeviceFingerprint.generateDeviceFingerprint(baseUrl, $sandbox)
+        .then(function () {
+          done.fail('Fingerprint promise should not have been resolved');
+        })
+        .fail(function (reason) {
+          done.fail('Fingerprint promise should not have been rejected. ' +  reason);
+        });
+      Util.mockSetTimeout();
+      setTimeout(function() {
```
mauriciocastillosilva-okta: can we use a `tick()` here instead?

src/util/DeviceFingerprint.js
```diff
@@ -10,28 +10,36 @@
...
-/* eslint complexity: [2, 7] */
+/* eslint complexity: [2, 8] */
 
 define(['vendor/lib/q', 'okta/jquery'], function (Q, $) {
 
   return {
+    __getUserAgent: function() {
+      return navigator.userAgent;
+    },
+    _isMessageFromCorrectSource: function($iframe, event) {
```
mauriciocastillosilva-okta: either add this one with `__` as a suffix or change `__getUserAgent` to have only one `_`. I vote for 2 underscores...

Although, now that I think about this, we don't really have a pattern here in the widget repo for naming "internal" or "private" functions. so I will probably just leave the names without any suffix. Plus they are not actually "private", since we are also returning them.

test/unit/spec/DeviceFingerprint_spec.js
mauriciocastillosilva-okta: Glad you find this issue!

PR #464, opened 2018-05-23
:white_check_mark: stabilize call-resend test

mauriciocastillosilva-okta:
Test is sometimes failing in Travis, so instead of waiting for a tick, which is not really predictable, I change the logic to wait for specific changes in the UI.

Resolves: OKTA-172535

### DIFF
--- a/test/unit/spec/MfaVerify_spec.js
+++ b/test/unit/spec/MfaVerify_spec.js
@@ -1645,19 +1645,27 @@ function (Okta,
         });
         itp('posts to resend link if call button is clicked for the second time', function () {
           Util.speedUpPolling();
-
           return setupCall().then(function (test) {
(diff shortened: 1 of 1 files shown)


test/unit/spec/MfaVerify_spec.js
```diff
@@ -1603,19 +1603,27 @@ function (Okta,
         });
         itp('posts to resend link if call button is clicked for the second time', function () {
           Util.speedUpPolling();
-
           return setupCall().then(function (test) {
-            $.ajax.calls.reset();
```
magizh-okta: Not sure why we need to remove this though ?

jmelberg-okta:
@mauriciocastillosilva-okta ^ Any estimate on when you'll be able to revisit this?

PR #466, opened 2018-05-29
:seedling: Use default success/error handlers when not provided

jmelberg-okta:
### Description
In the past, we've seen a number of developers omit adding `success` or `error` handlers when rendering the Widget.

Previously, an error was thrown to the console when a `success` handler was **not** provided. If an `error` handler was missing, the developer would not be notified of any error caught by the Widget's `globalErrorFn`.

```javascript
const signIn = new OktaSignIn({ baseUrl: 'https://{yourOktaDomain}' });
signIn.renderEl({
  el: '#widget-container'
},
function success(res) {
  console.log(`Response: ${res}`);
});
// If a recoverable error occurs, this error will be lost
```

This PR adds a **default** `success` handler (see flows below) and a **default** `error` handler to enhance the Widget's DX:

```javascript
const signIn = new OktaSignIn({ baseUrl: 'https://{yourOktaDomain}' });
signIn.renderEl({ el: '#widget-container' });

// Errors are passed via console warning messages
```

#### Default `success` handler
For various [OpenID Connect](https://github.com/okta/okta-signin-widget/#openid-connect) flows, the Widget never enters the `success` handler, redirecting the end-user to another application path.

See these sample applications as a reference: [Angular](https://github.com/okta/samples-js-angular/blob/master/custom-login/src/app/login/login.component.ts#L51-L62), [React](https://github.com/okta/samples-js-react/blob/master/custom-login/src/Login.jsx#L47-L58), [Vue](https://github.com/okta/samples-js-vue/blob/master/custom-login/src/components/Login.vue#L53-L64).

From a high-level:
- `mysite.com/login`
  - When the Widget is configured with `display: page`, it will initiate the OAuth 2.0 redirect flow
- :leftwards_arrow_with_hook: Redirect to app
- `mysite.com/callback`
  - Handle token parsing from `redirectUri`

#### Default `error` handler
We use a custom `Logger` class to output errors thrown by the Widget if the `error` handler is not provided. This is intended for running in **development** mode, as when the Widget is built for production, the `Logger.warn` messages will be stripped out.

### Resolves
[OKTA-171139](https://oktainc.atlassian.net/browse/OKTA-171139)

> See this Wiki for more details: [Improving Sign-In Widget DX](https://oktawiki.atlassian.net/wiki/spaces/~nate.barbettini/pages/299368755/Improving+Sign-In+Widget+DX)

### DIFF
--- a/src/models/Settings.js
+++ b/src/models/Settings.js
@@ -308,9 +308,6 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
       if (!options.baseUrl) {
         this.callGlobalError(new ConfigError(Okta.loc('error.required.baseUrl')));
       }
-      else if (!options.globalSuccessFn) {
-        this.callGlobalError(new ConfigError(Okta.loc('error.required.success')));
-      }
       else if (BrowserFeatures.corsIsNotSupported()) {
         this.callGlobalError(new UnsupportedBrowserError(Okta.loc('error.unsupported.cors')));
       }
--- a/src/util/BaseLoginRouter.js
+++ b/src/util/BaseLoginRouter.js
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-/* eslint max-params: [2, 13], max-statements: [2, 18] */
+/* eslint max-params: [2, 14], max-statements: [2, 18] */
 // BaseLoginRouter contains the more complicated router logic - rendering/
 // transition, etc. Most router changes should happen in LoginRouter (which is
 // responsible for adding new routes)
@@ -27,10 +27,12 @@ define([
   './RouterUtil',
   './Animations',
   './Errors',
-  'util/Bundles'
+  'util/Bundles',
+  'util/Logger'
 ],
-function (Okta, Backbone, BrowserFeatures, RefreshAuthStateController, Settings, Header,
-          SecurityBeacon, AuthContainer, AppState, RouterUtil, Animations, Errors, Bundles) {
+function (Okta, Backbone, BrowserFeatures, RefreshAuthStateController, Settings,
+          Header, SecurityBeacon, AuthContainer, AppState, RouterUtil, Animations,
+          Errors, Bundles, Logger) {
 
   var _ = Okta._,
       $ = Okta.$;
@@ -74,6 +76,16 @@ function (Okta, Backbone, BrowserFeatures, RefreshAuthStateController, Settings,
     Events:  Backbone.Events,
 
     initialize: function (options) {
+      // Create a default success and/or error handler if
+      // one is not provided.
+      if (!options.globalSuccessFn) {
+        options.globalSuccessFn = function () {};
+      }
+      if (!options.globalErrorFn) {
+        options.globalErrorFn = function(err) {
+          Logger.warn(err);
+        };
+      }
       this.settings = new Settings(_.omit(options, 'el', 'authClient'), { parse: true });
       this.settings.setAuthClient(options.authClient);
 
--- a/src/util/BaseLoginRouter.js
+++ b/src/util/BaseLoginRouter.js
@@ -83,7 +83,7 @@ function (Okta, Backbone, BrowserFeatures, RefreshAuthStateController, Settings,
       }
       if (!options.globalErrorFn) {
         options.globalErrorFn = function(err) {
-          Logger.warn(err);
+          Logger.error(err);
         };
       }
       this.settings = new Settings(_.omit(options, 'el', 'authClient'), { parse: true });
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -5,7 +5,6 @@ var OktaSignIn = (function () {
   var config  = require('json!config/config'),
       _ = require('okta/underscore');
 
-
   function getProperties(authClient, LoginRouter, Util, config) {
 
     /**
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -5,6 +5,7 @@ var OktaSignIn = (function () {
   var config  = require('json!config/config'),
       _ = require('okta/underscore');
 
+
   function getProperties(authClient, LoginRouter, Util, config) {
 
     /**
--- a/test/unit/spec/LoginRouter_spec.js
+++ b/test/unit/spec/LoginRouter_spec.js
@@ -6,6 +6,7 @@ define([
   'shared/util/Util',
   'util/CryptoUtil',
   'shared/util/Logger',
+  'util/Logger',
   '@okta/okta-auth-js/jquery',
   'helpers/mocks/Util',
   'helpers/util/Expect',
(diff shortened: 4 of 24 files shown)


src/widget/OktaSignIn.js
```diff
@@ -5,6 +5,7 @@ var OktaSignIn = (function () {
   var config  = require('json!config/config'),
       _ = require('okta/underscore');
 
+
```
santhoshbalakrishnan: nit: remove new line

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -5,11 +5,11 @@ define([
   'util/Logger'
 ],
 function (Widget, Expect, Logger) {
-  var signIn;
+  var signIn, devLogger;
   var url = 'https://foo.com';
 
   beforeEach(function(){
-    spyOn(Logger, 'warn');
+    devLogger = spyOn(Logger, 'warn');
```
santhoshbalakrishnan: Would prefer leaving it as before, so that we are clear about what needs to be called right on the `expect` statement.

webpack.test.config.js
```diff
@@ -3,8 +3,9 @@
 var path          = require('path');
 var _             = require('underscore');
 var commonConfig  = require('./webpack.common.config');
-var testConfig    = commonConfig('main-tests.js');
 var plugins       = require('./buildtools/webpack/plugins');
+var testConfig    = commonConfig('main-tests.js');
```
santhoshbalakrishnan: Is this just a cosmetic change?

test/unit/spec/LoginRouter_spec.js
```diff
@@ -189,29 +189,77 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Exp
...
+        return Expect.waitForPrimaryAuth();
+      })
+      .then(function () {
+        var form = new PrimaryAuthForm($sandbox);
+        expect(form.isPrimaryAuth()).toBe(true);
+      });
+    });
+    itp('uses a default globalSuccessFn if a null globalSuccessFn is passed as a widget param', function () {
+      return setup({ globalSuccessFn: null })
+      .then(function (test) {
```
santhoshbalakrishnan: Super nit: you can just move these into a method on its own and just pass `null` and `undefined` as parameters. Or even pass in the `{ globalSuccessFn: null }` or `{ globalErrorFn: undefined }` config.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -32,11 +33,11 @@ define([
   'helpers/xhr/labels_login_ja',
   'helpers/xhr/labels_country_ja'
 ],
-function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Expect, Router,
-          $sandbox, PrimaryAuthForm, IDPDiscoveryForm, RecoveryForm, MfaVerifyForm, EnrollCallForm, resSuccess, resRecovery,
-          resMfa, resMfaRequiredDuo, resMfaRequiredOktaVerify, resMfaChallengeDuo, resMfaChallengePush,
-          resMfaEnroll, errorInvalidToken, resUnauthenticated, resSuccessStepUp, Errors, BrowserFeatures,
-          labelsLoginJa, labelsCountryJa) {
+function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, DevLogger, OktaAuth, Util, Expect, Router,
```
santhoshbalakrishnan: Can we rename the `'shared/util/Logger',` to `courageLogger` and leave this as `Logger`?

webpack.test.config.js
jmelberg-okta: More of an `eslint` change. All of our imports are now at the top of the page (`commonConfig` slipped in there before).

test/unit/spec/OktaSignIn_spec.js
jmelberg-okta: This is required due to the update in [`Expect.js`](https://github.com/okta/okta-signin-widget/pull/466/files#diff-9f742f61fcd34d4a03a359547d1eebfcR31), allowing us to reassign the `Logger.warn` spy multiple times.

test/unit/spec/LoginRouter_spec.js
jmelberg-okta: Cleaned this up quite a bit!

src/models/Settings.js
```diff
@@ -308,9 +308,6 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
       if (!options.baseUrl) {
         this.callGlobalError(new ConfigError(Okta.loc('error.required.baseUrl')));
       }
-      else if (!options.globalSuccessFn) {
```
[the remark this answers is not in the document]
jmelberg-okta: This is a fair point.  We'd need to validate `authParams.display === 'page'` here, since this performs the redirect flow.

I think the idea moving forward is "Hey, the `success` handler isn't required for some flows, but it might be for yours."

@nbarbettini - Should we revisit adding the `Logger` when we **don't** have these callback methods?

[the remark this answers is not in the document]
nbarbettini: I don't want to add too much logic to these checks. Today, the success handler is required for authN flows and not required for OIDC flows, but that could change in the future.

What we're saying with this change is: if you don't add a success handler, nothing will happen on success. If you want to do something on success, add a success handler. I think that is clear and makes sense.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -170,48 +170,75 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Exp
...
     it('has the correct error message if el is not passed as a widget param', function () {
       var fn = function () { setup({ el: undefined }); };
-      expect(fn).toThrowError('"el" is a required widget parameter');
+      expect(fn).not.toThrow();
+      var err = Logger.warn.calls.mostRecent().args[0];
+      expect(err.name).toBe('CONFIG_ERROR');
+      expect(err.message).toEqual('"el" is a required widget parameter');
     });
-    it('throws a ConfigError if baseUrl is not passed as a widget param', function () {
+    it('logs a ConfigError if baseUrl is not passed as a widget param', function () {
```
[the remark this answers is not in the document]
nbarbettini: @jmelberg-okta What ends up being sent to the console here, a warning?

src/util/BaseLoginRouter.js
```diff
@@ -74,6 +76,16 @@ function (Okta, Backbone, BrowserFeatures, RefreshAuthStateController, Settings,
     Events:  Backbone.Events,
 
     initialize: function (options) {
+      // Create a default success and/or error handler if
+      // one is not provided.
+      if (!options.globalSuccessFn) {
+        options.globalSuccessFn = function () {};
+      }
+      if (!options.globalErrorFn) {
+        options.globalErrorFn = function(err) {
```
robertjd: Our default error handler should send the errors to console.error (see definition in wiki link from REQ-13008).

If we are using our default error handler, I believe we should log a development-mode warning that lets the developer know we are using a default error handler.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -170,48 +170,75 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, Logger, OktaAuth, Util, Exp
...
+      return expectPrimaryAuthRender({ globalSuccessFn: undefined });
+    });
+    itp('uses a default globalSuccessFn if a null globalSuccessFn is passed as a widget param', function () {
+      return expectPrimaryAuthRender({ globalSuccessFn: null });
     });
-    it('throws a ConfigError if globalSuccessFn is not passed as a widget param', function () {
-      var fn = function () { setup({ globalSuccessFn: undefined }); };
-      expect(fn).toThrowError(Errors.ConfigError);
+    itp('uses a default globalErrorFn if an undefined globalErrorFn is passed as widget param', function () {
+      return expectPrimaryAuthRender({ globalErrorFn: undefined });
```
robertjd: I'm not sure that this test is actually testing that the default handler is used? This test feels more like "assert that things work if I omit the handler during setup" ?

src/util/BaseLoginRouter.js
jmelberg-okta: We [decided against](https://oktawiki.atlassian.net/wiki/spaces/~nate.barbettini/pages/299368755/Improving+Sign-In+Widget+DX?focusedCommentId=496699714#comment-496699714) adding `console` messages when these are omitted. As an example, the following would log a message when we'd expect it not to:

```javascript
const signIn = new OktaSignIn({ baseUrl: 'https://{yourOktaDomain}' });
signIn.renderEl({ el: '#widget-container' });
```

test/unit/spec/LoginRouter_spec.js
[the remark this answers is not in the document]
jmelberg-okta: This is a misleading test - as `okta-auth-js` throws an error before the Widget has a chance to evaluate this.

I've updated this PR to expect an `AuthSdkError`, instead of asserting a `console` warning that will **never** be seen.

src/util/BaseLoginRouter.js
robertjd: Ok, that decision makes sense for my second statement about warnings.  I still want us to change the default error logger so that it sends the errors to console.error()

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -5,11 +5,11 @@ define([
   'util/Logger'
 ],
 function (Widget, Expect, Logger) {
-  var signIn, devLogger;
+  var signIn;
```
jmelberg-okta: Reverts change implemented in a [previous PR commit](https://github.com/okta/okta-signin-widget/pull/466/commits/7fa57dc3d16e2cdf68aae33d5d9b6be3d578a79a#r191577368)

sharstream:
This issue has been fixed in my own app using this setup https://developer.okta.com/docs/api/getting_started/enabling_cors#granting-cross-origin-access-to-websites. Thanks.

PR #468, opened 2018-05-31
:bug: Adds remember device checkbox for u2f

gowthamidommety-okta:
Resolves: OKTA-171665

When checking rememberDevice checkbox:

u2f_remembertrue

When unchecking rememberDevice checkbox:

u2f_rememberfalse

### DIFF
--- a/src/VerifyU2FController.js
+++ b/src/VerifyU2FController.js
@@ -19,10 +19,11 @@ define([
   'util/FormType',
   'views/shared/FooterSignout',
   'vendor/lib/q',
+  'util/FactorUtil',
   'views/mfa-verify/HtmlErrorMessageView',
   'u2f-api-polyfill'
 ],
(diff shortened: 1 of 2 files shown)


src/VerifyU2FController.js
```diff
@@ -62,14 +74,16 @@ function (Okta, FormController, FormType, FooterSignout, Q, HtmlErrorMessageView
             var deferred = Q.defer();
             u2f.sign(appId, factorData.challenge.nonce, registeredKeys, function (data) {
               self.trigger('errors:clear');
+              var rememberDevice = !!self.get('rememberDevice');
```
mauriciocastillosilva-okta: should this be in the `else` block?

gowthamidommety-okta: yes, you're right. It's better in the else block.

PR #472, opened 2018-06-02
:seedling: Adds helpful message when tokens are not parsed

jmelberg-okta:
### Description
Provides a console warning (**development mode**) when tokens have not been removed from the URL, after the Widget has been rendered. This *hint* is to ensure developers parse and store the tokens correctly.

screen shot 2018-06-01 at 5 08 48 pm

### Resolves
[OKTA-171143](https://oktainc.atlassian.net/browse/OKTA-171143)

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -11,7 +11,7 @@
  */
 
 /* eslint complexity: [2, 8], max-depth: [2, 3] */
-define(['okta'], function (Okta) {
+define(['okta', 'util/Logger'], function (Okta, Logger) {
 
   var Util = {};
   var _ = Okta._;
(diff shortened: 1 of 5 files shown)


nbarbettini:
How long does the widget wait before warning? The hash might be updated a few ms after initial page load.

jmelberg-okta:
@nbarbettini,

It is an immediate warning - given the page + Widget are in a specific state.

There are a few different states the Widget can be in where a developer may have failed to remove the tokens.

1. Instantiation
    ```javascript
    const signIn = new OktaSignIn({config})
    // Setup a timeout here, to alert the developer that they
    // have configured the widget again w/o removing tokens
    ```
    This seems problematic, as we would only alert **if** the Widget is constructed again.

2. `renderEl`
     ```javascript
      oktaSignIn.renderEl({ '#widget-container' })
     ```
    Here, we know that they're attempting to re-render the Widget with tokens in the URL. At this point, we know that they forgot to remove the tokens - and will attempt to sign-in again. This is the most optimal spot to place it.

3. Success handler
     ```javascript
    oktaSignIn.renderEl({ 
      '#widget-container'
    }, function success(response) {
      // Attach a timeout to success handler behind-the-scenes
    });
    ```
    Again, this is problematic as it is unlikely that we do not escape the `success` handler within the allotted timeout period.

Does this help provide more clarification?

robertjd:
@nbarbettini see this section of the README, which shows how we tell people to grab tokens: https://github.com/okta/okta-signin-widget#oidc-tokenparsetokensfromurlsuccess-error

PR #474, opened 2018-06-07
use new endpoint to social auth when none OIDC mode

haishengwu-okta:
## Description

- social login today only works in OIDC mode which doesn't well support the customization signin page.
- a new endpoint was added last week (/sso/idp/:id) to allow social idp to be used as standard IdP, meaning could be used at none OIDC mode.
- this commit is to implement social login using the new endpoint at none OIDC mode

## Video

- https://okta.box.com/s/kvxc3dmsfn4071os9tp4itrspa5nckuh

## JIRA

- [OKTA-159649](https://oktainc.atlassian.net/browse/OKTA-159649)

### DIFF
--- a/src/models/Settings.js
+++ b/src/models/Settings.js
@@ -20,9 +20,10 @@ define([
   'util/Util',
   'util/Logger',
   'util/OAuth2Util',
+  'shared/util/Util',
   'json!config/config'
 ],
-function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
+function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, SharedUtil, config) {
 
   var DEFAULT_LANGUAGE = 'en';
 
@@ -31,6 +32,7 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
       oauthRedirectTpl = Okta.tpl('{{origin}}');
 
   var _ = Okta._,
+      $ = Okta.$,
       ConfigError = Errors.ConfigError,
       UnsupportedBrowserError = Errors.UnsupportedBrowserError;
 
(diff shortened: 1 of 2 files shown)


src/models/Settings.js
```diff
@@ -275,7 +277,16 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
               },
               click: function (e) {
                 e.preventDefault();
-                OAuth2Util.getTokens(self, {idp: idp.id});
+                if (self.get('oauth2Enabled')) {
```
mauriciocastillosilva-okta: I can't remember exactly the order of how derived props are computed, but I think we should add `oauth2Enabled` as a dependency in line 263 `deps: ['configuredSocialIdps', 'customButtons', 'oauth2Enabled'],`, so we know that it is computed before `configuredButtons`

haishengwu-okta: Good catch! Thanks!

mauriciocastillosilva-okta: Maybe... but in `OAuth2Util` you don't really get any derived property I think, it is either regular props or the `authClient`. 
Since I don't know exactly how we compute derived props, I think it is safer to move it to deps. I assume that if a derived prop depends on another derived prop, the ones that have no dependency will be computed first.

haishengwu-okta: - actually derived props are computed when construct the model.
- the diff with regular props are derived props adapts to changes from regular props. (update derived props when deps changed)
- now the question is if `oauth2Enabled` has been changed, shall we update `configuredButtons` and will it impact the click handler? yes for 1 but properly no for 2 since once the `configuredButtons` have been rendered, we didn't re-render upon changes.
- theoretically it would be more accurate to always the derived prop at click handler cause it could always get latest value v.s. adding the `oauth2Enabled` as another dependency which would become a closure and won't be updated after first render.
- I'm thinking is we shall't create event handlers at Model but View and maybe shall refactor the code at some point?

haishengwu-okta:
@larsjohansen-okta we don't have 'central page' for endpoints and usually at Model, we tend to define endpoint template at top and at this case, I'm a little bit hesitate since it's being used once at click handler.

src/models/Settings.js
mauriciocastillosilva-okta: > theoretically it would be more accurate to always the derived prop at click handler cause it could always get latest value

good point. I think then this is good... as you said, we probably shouldn't have click handlers here. But I agree, that is a different task than we can deal with later.

PR #475, opened 2018-06-07
Send accept language header if widget configured with language.

nikhilvenkatraman-okta:
- When the widget is bootstrapped with a language in the config option, we want to send that language code as a part of the Accept-Language header while making API requests.
- This is so that the server returns an API error localized in the widgets configured language which can be different than the browser language setting.

Video shows before and after results 
https://okta.box.com/s/c7a0ztrdeo7c496575hkm6s7iagvk7yj

Resolves: OKTA-168338

@santhoshbalakrishnan @mauriciocastillosilva-okta @hor-kanchan-okta

### DIFF
--- a/src/models/Settings.js
+++ b/src/models/Settings.js
@@ -324,10 +324,14 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, SharedUtil
       }
     },
 
-    setAuthClient: function (authClient) {
+    setAcceptLanguageHeader: function (authClient) {
       if (authClient && authClient.options && authClient.options.headers) {
         authClient.options.headers['Accept-Language'] = this.get('languageCode');
       }
+    },
+
+    setAuthClient: function (authClient) {
+      this.setAcceptLanguageHeader(authClient);
       this.authClient = authClient;
     },
 
--- a/src/models/Settings.js
+++ b/src/models/Settings.js
@@ -325,6 +325,9 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, SharedUtil
     },
 
     setAuthClient: function (authClient) {
+      if (authClient && authClient.options && authClient.options.headers) {
+        authClient.options.headers['Accept-Language'] = this.get('languageCode');
+      }
       this.authClient = authClient;
     },
 
--- a/test/unit/spec/LoginRouter_spec.js
+++ b/test/unit/spec/LoginRouter_spec.js
@@ -71,7 +71,7 @@ function (Okta, Q, Backbone, SharedUtil, CryptoUtil, CourageLogger, Logger, Okta
     function setup(settings) {
       var setNextResponse = Util.mockAjax();
       var baseUrl = 'https://foo.com';
-      var authClient = new OktaAuth({url: baseUrl});
+      var authClient = new OktaAuth({url: baseUrl, headers: {}});
       var eventSpy = jasmine.createSpy('eventSpy');
       var router = new Router(_.extend({
         el: $sandbox,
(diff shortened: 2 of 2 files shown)


src/models/Settings.js
```diff
@@ -314,6 +314,9 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
     },
 
     setAuthClient: function (authClient) {
+      if (authClient && authClient.options && authClient.options.headers) {
+        authClient.options.headers['Accept-Language'] = this.get('languageCode');
```
mauriciocastillosilva-okta: I think `languageCode` always returns a value right? (default 'en') So if the user does not set a language, then `get('languageCode')` will return English by default. This could be a problem if the browser is in a different language than English, cause the widget will get the browser language, but the API will be called with 'en'

nikhilvenkatraman-okta: Hey @mauriciocastillosilva-okta if the locale is not set via widget config but instead we use browser setting to change the locale to something other than english,  `languageCode ` will get that value set in the browser's language setting. Internally while calculating `languageCode ` we call `BrowserFeatures.getUserLanguages()` https://github.com/okta/okta-signin-widget/blob/master/src/util/BrowserFeatures.js#L58 

Screenshot below shows widget not configured with any locale but browser locale set to spanish, and `languageCode` picking up the browser setting

screen shot 2018-06-08 at 7 05 38 am

santhoshbalakrishnan: We do not have to add to the existing list of accept-language header value? Is replacing fine?

src/models/Settings.js
```diff
@@ -314,6 +314,9 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
     },
 
     setAuthClient: function (authClient) {
+      if (authClient && authClient.options && authClient.options.headers) {
```
santhoshbalakrishnan: Can we move this part into its own function? `setLanguageForWidget` maybe? So that anyone who's using the widget can just comment that out if needed.

src/models/Settings.js
nikhilvenkatraman-okta: Discussed over slack: 

### TLDR; 
Add another fallback locale appended along with `languageCode` while setting `Accept-Language` header.


### Details:
Typically the browser sends a list of languages with their weights as a part of `Accept-Language` header 

Eg: `Accept-Language: en,es;q=0.9,fr;q=0.8,el;q=0.7,ru;q=0.6,nl;q=0.5,pt-BR;q=0.4,pt;q=0.3,es-US;q=0.2,en-US;q=0.1,pt-PT;q=0.1,zh;q=0.1,zh-CN;q=0.1,zh-TW;q=0.1,zh-HK;q=0.1`)

Which means english is preferred local for response, but es is the next preferred locale on the list.

For this PR, instead of just sending the `languageCode`, we will try to also send a fallback locale along with the `languageCode`. 

This fallback locale will be the first locale from the list of browser languages with a weight of `0.9`.

Ideally we should not have to have to use the fallback locale since the `languageCode` will default to `en` if user tries to set it to a unsupported locale.

Example value 
authClient.options.headers['Accept-Language'] = `es, en;q=0.9`

Where `es` is the languageCode configured in the widget and the preferred locale for the response, but en is the fallback locale.

@santhoshbalakrishnan @mauriciocastillosilva-okta @hor-kanchan-okta let me know you guys think.

rchild-okta: What's the reason for the fallback? It seems like the point of this PR is to get consistency between the widget language and any text sent back in an API response - if the widget thinks it should display in English (because they didn't override or because they overrode with an unsupported language), why send another option?

There are probably some other edge cases to think about:
- What if they specify their own translations for a language we don't support? This will work for anything that's bundled in the widget, but would be one of the cases where we don't have translations in the backend - is this the reason for the fallback?
  - And if this is the case, are we okay with changing the fallback from what it currently is (english) to the next-best option from the browser?
-  Might not be relevant, but the widget is currently setup in a way that language can be changed on the fly - in this commit it only gets set when initializing the auth client. Might be worth thinking about making this more dynamic in authJs.

santhoshbalakrishnan: > This fallback locale will be the first locale from the list of browser languages with a weight of 0.9

For clarity, would change that to `first locale from the list of browser languages *that we support*`

Will wait for @mauriciocastillosilva-okta and @hor-kanchan-okta 's comments.

santhoshbalakrishnan: @rchild-okta 

>And if this is the case, are we okay with changing the fallback from what it currently is (english) to the next-best option from the browser?

Ya, that's what we discussed. More like next best option that we have in our supported list of languages.

mauriciocastillosilva-okta: @santhoshbalakrishnan this is what we were wondering. As @nikhilvenkatraman-okta pointed out, languageCode considers the browser languages in its logic, so if the config language is invalid, we check then if the browser languages are supported. And if none are, then we default to en.
@nikhilvenkatraman-okta can you double check we have a test for languageCode where the config lang is invalid so we return the browser lang?

nikhilvenkatraman-okta: @mauriciocastillosilva-okta https://github.com/okta/okta-signin-widget/blob/master/test/unit/spec/LoginRouter_spec.js#L1378

mauriciocastillosilva-okta:
I think this is fine. The other place where we set headers is in [OktaSignIn](https://github.com/okta/okta-signin-widget/blob/master/src/widget/OktaSignIn.js#L182), but at that point I don't think we have access to the Settings yet. So I think it is good as is. But let's wait for Reman's comment

nikhilvenkatraman-okta:
@santhoshbalakrishnan @mauriciocastillosilva-okta I did discuss with @rchild-okta about this over slack and he was ok to keep it here for now. The use-case when we would want to move this over to auth js might be when we have to dynamically switch the locale which might be some thing to think about for the future.

PR #485, opened 2018-06-20
:bug: clear input field if changing RSA pin

mauriciocastillosilva-okta:
RSA returns a 409 error when changing the pin number. At this point we need to clear the input field so we don't have the previous value in it to avoid possible errors of users changing the pin to something different than intented

Resolves: OKTA-165849

clearinputduringchangepin

### DIFF
--- a/src/EnrollOnPremController.js
+++ b/src/EnrollOnPremController.js
@@ -72,6 +72,16 @@ function (Okta, FormType, FormController, Footer, TextBox) {
         noButtonBar: true,
         autoSave: true,
         className: getClassName(provider),
+        initialize: function() {
+          this.listenTo(this.model, 'error', _.bind(function (source, error) {
+            if (error && error.status === 409) {
+              // 409 means we are in change pin, so we should clear out answer input
+              this.model.set('passCode', '');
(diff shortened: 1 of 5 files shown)


src/EnrollOnPremController.js
```diff
@@ -72,6 +72,18 @@ function (Okta, FormType, FormController, Footer, TextBox) {
         noButtonBar: true,
         autoSave: true,
         className: getClassName(provider),
+        initialize: function() {
+          if (isRSA(provider)) {
+            this.listenTo(this.model, 'error', _.bind(function (source, error) {
+              if (error && error.status === 409) {
+                // 409 means we are in change pin, so we should clear out answer input
+                this.model.set('passCode', '');
+                this.$('.o-form-input-name-passCode input').val('');
```
gowthamidommety-okta: couldn't we trigger 'change' and that would update the model as well?

mauriciocastillosilva-okta: would have to be `change` in the input right? I don't have the input here, and probably it is simpler to set the value as it is, as opposed to find the input. Does that sounds right?

PR #486, opened 2018-06-25
:seedling:  Add showPassword event to okta-signin-widget

manueltanzi-okta:
Add showPassword event to okta-signin-widget

Resolves: [OKTA-175374](https://oktainc.atlassian.net/browse/OKTA-175374)

@hor-kanchan-okta @eugenebakaiev-okta

### DIFF
--- a/src/PrimaryAuthController.js
+++ b/src/PrimaryAuthController.js
@@ -97,6 +97,9 @@ function (Okta, PrimaryAuthForm, CustomButtons, FooterRegistration, PrimaryAuthM
       },
       'focusout input': function (e) {
         $(e.target.parentElement).removeClass('focused-input');
+      },
+      'click .button-show': function () {
+        this.trigger('passwordRevealed');
       }
     },
 
--- a/test/unit/spec/PrimaryAuth_spec.js
+++ b/test/unit/spec/PrimaryAuth_spec.js
@@ -598,6 +598,31 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
           expect(test.form.passwordToggleContainer().length).toBe(1);
         });
       });
+      itp('Triggers a passwordRevealed event when show password button is clicked', function () {
+        return setup({ 'features.showPasswordToggleOnSignInPage': true }).then(function (test) {
+          var eventSpy = jasmine.createSpy('eventSpy');
+          test.router.on('passwordRevealed', eventSpy);
+          test.form.setPassword('testpass');
+          test.form.setUsername('testuser');
+          expect(test.form.passwordToggleContainer().length).toBe(1);
+          test.form.passwordToggleShowContainer().click();
+          expect(eventSpy).toHaveBeenCalled();
+        });
+      });
+      itp('Does not trigger a passwordRevealed event when hide password button is clicked', function () {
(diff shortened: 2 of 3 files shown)


eugenebakaiev-okta:
Looks good

src/PrimaryAuthController.js
```diff
@@ -97,6 +97,12 @@ function (Okta, PrimaryAuthForm, CustomButtons, FooterRegistration, PrimaryAuthM
       },
       'focusout input': function (e) {
         $(e.target.parentElement).removeClass('focused-input');
+      },
+      'click .button-show': function () {
+        var track = this.settings.get('tracker.track');
+        if(track) {
```
eugenebakaiev-okta: Nit: maybe it would be nice to also check that `track` is function? With `_.isFunction()` for example

src/PrimaryAuthController.js
```diff
@@ -97,6 +97,9 @@ function (Okta, PrimaryAuthForm, CustomButtons, FooterRegistration, PrimaryAuthM
       },
       'focusout input': function (e) {
         $(e.target.parentElement).removeClass('focused-input');
+      },
+      'click .button-show': function () {
```
rchild-okta: Is there a reason we're using a namespace 'sign-in-widget-event' rather than a named event like 'showPassword'?

src/PrimaryAuthController.js
```diff
@@ -97,6 +97,9 @@ function (Okta, PrimaryAuthForm, CustomButtons, FooterRegistration, PrimaryAuthM
...
+        this.trigger('showPassword');
```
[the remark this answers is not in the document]
manueltanzi-okta: @nbarbettini We want the event from hidden to show. Will add in the Events section 👍

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -598,6 +598,17 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
...
       });
+      itp('Triggers a showPassword event when show password button is clicked', function () {
+        return setup({ 'features.showPasswordToggleOnSignInPage': true }).then(function (test) {
+          var eventSpy = jasmine.createSpy('eventSpy');
+          test.router.on('showPassword', eventSpy);
+          test.form.setPassword('testpass');
+          test.form.setUsername('testuser');
+          expect(test.form.passwordToggleContainer().length).toBe(1);
+          test.form.passwordToggleShowContainer().click();
+          expect(eventSpy).toHaveBeenCalled();
```
jmelberg-okta: Since this event is intended to **only** fire when going from hidden -> shown state, we need to have a test that asserts the opposite case.

This makes the behavior explicitly clear (and will catch devs who might change [`click .button-show`](https://github.com/okta/okta-signin-widget/pull/486/files#diff-19b949c50e99d13eb55db7490ef2effdR101) to `click .button-hide` and expect the same result).

src/PrimaryAuthController.js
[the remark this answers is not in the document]
rchild-okta: Agree with Nate on the past tense for the message. How about `passwordRevealed`?

nbarbettini:
In the future, we may add an event when the password is hidden too. I agree with @rchild-okta that `passwordRevealed` is nice, because then we could have a complementary `passwordHidden` event later on.

PR #489, opened 2018-06-27
UX Improvements for password toggle

nikhilvenkatraman-okta:
- Make target area for eyeicon bigger so that its easier to click on mobile and add cursor pointer for desktop [OKTA-173752](https://oktainc.atlassian.net/browse/OKTA-173752)

https://okta.box.com/s/2uklsmquhui5rr3mmxqkmfqv6l4re0pw

- Added padding to input to address issue for long passwords [OKTA-177309](https://oktainc.atlassian.net/browse/OKTA-177309)

Before: 
screen shot 2018-06-29 at 1 32 38 pm

After: 
screen shot 2018-06-29 at 1 32 46 pm

- Wrapped login page specific style overrides under `.primary-auth` classname which fixes and applies the style overrides only to login page where we have tooltip and need more space on the right [OKTA-175118](https://oktainc.atlassian.net/browse/OKTA-175118)
Before: 

screen shot 2018-06-29 at 1 35 16 pm

After: 

screen shot 2018-06-29 at 1 40 03 pm

@hor-kanchan-okta @johnintrater-okta @mauriciocastillosilva-okta @eugenebakaiev-okta


assets/sass/common/shared/o-forms/_o-form.scss
```diff
@@ -99,15 +99,21 @@
       box-sizing: border-box;
     }
 
+    input.password-with-toggle {
```
nikhilvenkatraman-okta: This classname is added from PasswordBox component in courage. https://github.com/okta/courage/pull/475

assets/sass/okta-theme.scss
```diff
@@ -733,6 +738,10 @@
       &[type="password"] {
       /* -- Input Fields: password input field -- */
         font-family: Verdana, sans-serif;
+
+        &:placeholder-shown {
```
nikhilvenkatraman-okta: Fixes [OKTA-176404](https://oktainc.atlassian.net/browse/OKTA-176404) . This is because we use a different font for password field https://github.com/okta/okta-signin-widget/blob/master/assets/sass/okta-theme.scss#L735. Setting the font to `proxima` when we field is empty to avoid the jump seen in the label.

nikhilvenkatraman-okta: https://okta.box.com/s/138r94bkg0lb34asgasf97abzhs9pdmo

nikhilvenkatraman-okta: This is specific to the widget. We are doing this because we use a different font for the password field in the widget https://github.com/okta/okta-signin-widget/blob/master/assets/sass/okta-theme.scss#L735.

johnintrater-okta:
👍

nikhilvenkatraman-okta:
@hor-kanchan-okta yes, this is not supported in IE, and I need to confirm if this was an issue in IE to begin with. But if it is then I feel we could take it as a known issue, since we are not compromising functionality. @johnintrater-okta thoughts?

johnintrater-okta:
Nothing new on my end @nikhilvenkatraman-okta

assets/sass/okta-theme.scss
mauriciocastillosilva-okta: wait, so all our fonts in the widget are the same EXCEPT for the password field? Any chance we can then make all consistent?

nikhilvenkatraman-okta:
@hor-kanchan-okta @johnintrater-okta 
I found  that IE has its own implementation of password toggle to password fields
https://okta.box.com/s/gmj0isa6kk6egglnlln9z2ja5ywa7h68 . 

Although there is a way to disable it via css, so that it does not interfere with our implementation. 

This is how it looks on IE 11 with the native password toggle disabled 
 https://okta.box.com/s/phq5odg0nk12v0ewzvp53793w3uquc47

assets/sass/okta-theme.scss
nikhilvenkatraman-okta: Unfortunately yes. We have had a long discussion about this already on slack (show_password_toggle). We cant make the password field's font the same as the other inputs as that font (proxima) makes the bullets look really tiny. 

screen shot 2018-06-29 at 11 14 28 am

packages/@okta/courage/README.md
```diff
@@ -179,8 +179,8 @@ Command        | Description
 -------------- | --------------
 **yarn lint** | Runs *scss-lint*, *jshint*, and *eslint*
 **yarn test** | Runs the jasmine unit tests in the cli.
-**yarn test -- --br** | Runs the jasmine unit tests in the browser
+**yarn test --br** | Runs the jasmine unit tests in the browser
 **yarn start** | Start the playground server to run on port 3000 by default. Load a url in browser with the specified port to use the playground, ex. localhost:3000.
-**yarn test:e2e** | Runs end-2-end selenium tests.
-**yarn test:e2e -- --specs test/selenium/spec/NAME.js** | Runs single end-2-end selenium test
+**yarn test:e2e** | Runs end-2-end selenium tests. You **must** use a Chrome version [supported by the ChromeDriver version](http://chromedriver.chromium.org/downloads) we use.
```
mauriciocastillosilva-okta: this line doesn't read right
`You **must** use a Chrome version [supported by the ChromeDriver version](http://chromedriver.chromium.org/downloads) we use.` the last `we use.` seems out of place.
I think if what we want is to have a link of where to find a chrome version, we can do that 
`You **must** use a Chrome version (http://chromedriver.chromium.org/downloads).`

Also, is that entirely true? this doesn't work with other browsers? @rchild-okta or @jmelberg-okta do you guys know?

nikhilvenkatraman-okta: created https://oktainc.atlassian.net/browse/OKTA-178636 to track

Issue #488, opened 2018-06-27
Proposal: Default token handler


nbarbettini:
### Background

When the Widget is used to sign in users to a frontend application, tokens are returned from Okta in the fragment and this code is used to grab them:

```js
  signIn.token.parseTokensFromUrl(
    function success(res) {
      signIn.tokenManager.add('id_token', res[0]);
      signIn.tokenManager.add('access_token', res[1]);
    },
    function error(err) {
      console.log('handle error', err);
    }
  );
```

This code rarely needs to change. To reduce the code a developer needs to write (and reduce the chance of copypasta errors), I propose adding a built-in handler.

### DefaultTokenHandler

Define `OktaSignIn.prototype.DefaultTokenHandler()` with this behavior:

- Store access tokens via `tokenManager.add('access_token')`
- Store ID tokens via `tokenManager.add('id_token')`

### Changes to `parseTokensFromUrl`

If the second parameter of the function (`error`) is omitted, use the default error handler added in #466 (logging to the console).

#### Example

```js
signIn.token.parseTokensFromUrl(OktaSignIn.DefaultTokenHandler);
```

robertjd:
@nbarbettini just to clarify, are you expecting the developer to place this line in their application:

```javascript
signIn.token.parseTokensFromUrl(OktaSignIn.DefaultTokenHandler);
```

? That is what I would like to see, because it's explicit.

The current PR #494 tries to do it automatically, such that doing this:

```javascript
signIn.token.parseTokensFromUrl();
```
Has the magic side affect of running the default token handler.  I don't think this magic is clear and should be avoided.

jmelberg-okta:
@robertjd - Can you tell me more about the benefit of binding this to the root object as opposed to keeping it internal?

I'd imagine that this method would only be used once, so binding it to the `OktaSignIn` object seems a bit odd to me.

With many of the changes coming into the Widget, it seems like we're moving into a world where we're handling many of the well-known use cases for developers. If i'm understanding correctly, the goal of this type of change is to cleanup the code required for [basic OIDC use](https://developer.okta.com/code/javascript/okta_sign-in_widget#sign-in-to-your-application):

```javascript
var signIn = new OktaSignIn({ /* config */ });

if (!signIn.token.hasTokensInUrl()) {
  signIn.renderEl({el: '#widget-container'});
} else {
  signIn.token.parseTokensFromUrl(
    function success(res) {
      // Add the token to tokenManager to automatically renew the token when needed
[...]

Issue #487, opened 2018-06-27
Proposal: Convenience methods for signing in


nbarbettini:
### Background
The Sign-in Widget can be used for many use cases, but the two most common are:
* Starting an Okta session (most likely followed by redirecting to the Okta dashboard)
* Signing a user into your own application

To accomplish these two tasks, you shouldn't need to understand the full Widget configuration model. I propose adding two convenience methods that will make these tasks much easier.

### showSignInToGetTokens

Define the options interface

```ts
interface SignInToAuthorizationServerOptions {
  authorizationServerId?: string;
  clientId: string;
  redirectUri?: string;
  getIdToken?: boolean;
  getAccessToken?: boolean;
  scope?: string | string[];
}
```

and `OktaSignIn.showSignInToGetTokens(options: SignInToAuthorizationServerOptions)` that performs the following "translation":

```js
let config = {
  clientId = options.clientId,
  redirectUri = options.redirectUri || "{current origin}",
  authParams: {
    issuer: options.authorizationServerId,
    responseType: buildResponseTypeArray(options.getIdToken || true, options.getAccessToken || false),
    responseMode: 'fragment'
    display: 'page',
    scopes: options.scope
  }
};

OktaSignIn.renderEl(config, DefaultSuccessHandler, DefaultErrorHandler);
```

#### Example

With this new convenience method, the [sign in to your application example](https://developer.okta.com/code/javascript/okta_sign-in_widget#sign-in-to-your-application) from the Widget Guide becomes much shorter:

```js
var widget = new OktaSignIn({
  baseUrl: 'https://{yourOktaDomain}',
  el: '#widget-container'
});

if (!signIn.token.hasTokensInUrl()) {
  widget.showSignInToGetTokens({
    clientId: '${clientId}'
    getAccessToken: true
  });

} else {
  signIn.token.parseTokensFromUrl(...);
}
```

### showSignInToStartOktaSession

Define the options interface

```ts
interface LoginToOktaOptions {
  redirectUri?: string;
}
```

and `OktaSignIn.showSignInToStartOktaSession(options: SignInToOktaOptions)` that performs the following "translation":

```js
let redirectingSuccessHandler = (res) => {
  if (!options.redirectUri) return;

  if (res.status === 'SUCCESS') {
    res.session.setCookieAndRedirect(options.redirectUri);
  }

OktaSignIn.renderEl(config, redirectingSuccessHandler, DefaultErrorHandler);
```

#### Example

[...]

PR #494, opened 2018-07-06
:seedling: Adds default success/error handlers for parseFromUrl

jmelberg-okta:
### Description
Exposes a new method `parseAndStoreTokensFromUrl`, which uses a `default` success/error handler functions for the `parseTokensFromUrl` method.

One of the most common use cases is to use the [OAuth 2.0 Implicit Grant](https://oauth.net/2/grant-types/implicit/) to return an access and identity tokens. Previously, consumers of the widget needed parse these tokens from the URL similar to:

```javascript
signIn.token.parseTokensFromUrl(
  function success(res) {
    // Tokens are returned in an ordered Array, based on the order requested:
    // responseType: ['id_token', 'token']
    signIn.tokenManager.add('id_token', res[0]);
    signIn.tokenManager.add('access_token', res[1]);
  },
  function error(err) {
    // handle error
  }
);
```

This change removes the requirement for success and/or error handlers, so developers can simply use:

```javascript
signIn.token.parseAndStoreTokensFromUrl()
.then(function() {
  const idToken = signIn.tokenManager.get('idToken');
  console.log(idToken);
});

// Optionally, token storage keys can be utilized
signIn.token.parseAndStoreTokensFromUrl({
  accessToken: 'my-access-token-key',
  idToken: 'my-id-token-key'
})
.then(function() {
  const idToken = signIn.tokenManager.get('my-id-token-key');
  console.log(idToken);
});
```

Breakdown of primary commits:
- [`8cc9b3d`](https://github.com/okta/okta-signin-widget/pull/494/commits/8cc9b3d76ad78d956950d471c058960c56ae9c09) - Utilize new `WIDGET_CLIENT_ID` environment variable
- [`9abc37b`](https://github.com/okta/okta-signin-widget/pull/494/commits/9abc37b331464e72cd9686be7cd294424240c148) - Adds new `parseAndStoreTokensFromUrl` method

#### Additional Changes
- Moved `WIDGET_CLIENT_ID` to an environment variable to simplify local e2e testing. Follow-up work needs to be completed to make this easier to use for external contributors.

### Resolves
- #488 
- [OKTA-178048](https://oktainc.atlassian.net/browse/OKTA-178048)


src/util/Handlers.js
```diff
@@ -0,0 +1,47 @@
+/*!
+ * Copyright (c) 2015-2016, Okta, Inc. and/or its affiliates. All rights reserved.
```
robertjd: Wrong dates

src/util/Handlers.js
```diff
@@ -0,0 +1,47 @@
...
+  Handlers.defaultSuccessTokenHandler = function(tokenManager) {
+    /**
+     * Default success handler which is normally invoked when
+     * `parseTokenFromUrl` is called. This success handler will
+     * automatically store the tokens with the keys "accessToken" and
+     * "idToken", if returned in the response.
+     */
+    return function (res) {
+      var tokens = Array.isArray(res) ? res : [res];
+      for (var i = 0; i < tokens.length; ++i) {
```
robertjd: Perhaps personal pref, but`tokens.forEach(token => {})` feels more straightforward?

src/util/Handlers.js
```diff
@@ -0,0 +1,47 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define(['util/Logger'], function (Logger) {
+
+  var Handlers = {};
+
+  Handlers.defaultSuccessTokenHandler = function(tokenManager) {
```
jmelberg-okta: Open question - do we make the key values configurable? I personally like the idea of hardcoding these to `idToken` and `accessToken`, as it'll align with our README documentation and samples.

test/e2e/templates/basic-dev.tpl
```diff
@@ -1,15 +1,8 @@
...
-  'baseUrl': '{{{WIDGET_TEST_SERVER}}}'
-};
-var oktaSignIn = new OktaSignIn(options);
+var oktaSignIn;
 
-oktaSignIn.renderEl(
-  { el: '#okta-login-container' },
-  function (res) {
-    if (res.status === 'SUCCESS') {
-      res.session.setCookieAndRedirect(options.baseUrl + '/app/UserHome');
```
robertjd: Sanity check, do we still have a test that asserts that this flow is possible? (Do set cookie and redirect after the .renderEl() success handler called)

README.md
```diff
@@ -409,8 +409,9 @@ else {
 
 Parses the access or ID Tokens from the url after a successful authentication redirect. This is used when `authParams.display = 'page'`.
 
-- `success` - Function that is called after the tokens have been parsed and validated
-- `error` - Function that is called if an error occurs while trying to parse or validate the tokens
+- `success` *(optional)*- Function called after the tokens have been parsed and validated. If omitted, a default handler stores the tokens, where they can be returned via the [tokenManager.get](#oidc-tokenmanagergetkey) method under the keys `accessToken` and/or `idToken`.
```
robertjd: See my questions about this in #488

src/util/Handlers.js
jmelberg-okta: This was mostly a speed decision. Storing the tokens already takes a bit of time due to their size, so I wanted to try and keep performance up as high as possible.

```
// Store accessToken and idToken

// foreEach
33752.566650390625ms

// for loop
20372.998779296875ms
```

test/e2e/templates/basic-dev.tpl
jmelberg-okta: We do. Those are still covered using the `basic` template

src/util/Handlers.js
robertjd: Tell me what those numbers are telling me/ how you got them? There are performance tradeoffs with foreach vs for, but I'd be surprised if a two element array, evaluated once per app instance, causes any significant impact? This minimal use case is why I was suggesting a readability improvement over a possible performance issue.

jmelberg-okta: The numbers above are execution times for storing an `accessToken` and `idToken` into `localStorage`. The first, uses the `forEach` method as a replacement for the line in question [here](https://github.com/okta/okta-signin-widget/pull/494/files/53f106e9feaf0fc0ebd5bcef5c6932d9bdabd6fb#diff-e86f1699ebb182f99bfd831860aeeb29R26), while the second is the current state.

Both calls were measured using `console.time` and `console.timeEnd`.

Since we're looking at a `40%` time difference between the two, it makes the most sense to me to choose the basic loop - especially since this is happening during a state transition.

src/util/Handlers.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+define(['util/Logger'], function (Logger) {
+
+  var Handlers = {};
+
+  Handlers.defaultSuccessTokenHandler = function(tokenManager, keys = {}) {
+    /**
+     * Default success handler which is normally invoked when
+     * `parseTokenFromUrl` is called. This success handler will
```
robertjd: This comment needs to be updated per new changes

README.md
```diff
@@ -32,6 +32,7 @@ Contributors should read our [contributing guidelines](./CONTRIBUTING.md) if the
   * [session.close](#sessionclosecallback)
   * [token.hasTokensInUrl](#oidc-tokenhastokensinurl)
   * [token.parseTokensFromUrl](#oidc-tokenparsetokensfromurlsuccess-error)
+  * [token.parseAndStoreTokensFromUrl](#oidc-tparseAndStoreTokensFromUrl-tokenStorageKeys)
```
robertjd: Looks like the fragment has a typo?

README.md
```diff
@@ -450,6 +452,48 @@ else {
...
+// ID Token in the URL.
+else {
+  signIn.token.parseAndStoreTokensFromUrl()
+  .then(function() {
+    const idToken = signIn.tokenManager.get('idToken');
+    console.log(idToken);
+  });
+
+  // Optionally - use a different key for the idToken
+  signIn.token.parseAndStoreTokensFromUrl({ idToken: 'id-token-key' })
```
robertjd: Should we be updating our example code in https://github.com/okta/okta-signin-widget#rendereloptions-success-error to use this new method?

src/util/Handlers.js
```diff
@@ -0,0 +1,53 @@
...
+
+  Handlers.defaultErrorHandler = function() {
+    /**
+     * Default error handler that outputs the
+     * response using the window.console in "error" form.
+     */
+    return function (err) {
+      Logger.error(err.toString());
+    };
+  };
```
robertjd: I'm curious to know why this returns a function, rather than just accepting `err` and handling it?

README.md
jmelberg-okta: Ideally, yes. However I'm going to hold off on doing that for now - at least until these changes get published externally.

Due to the way we publish the Widget, we need to be careful about including README changes for functionality that isn't available on both the CDN and registry.

src/util/Handlers.js
jmelberg-okta: We're expecting `success` and `error` handlers to be in the form of functions, since they're passed directly into our promise resolver. See [`token.parseFromUrl`](https://github.com/okta/okta-auth-js#tokenparsefromurloptions) in AuthJS.

README.md
```diff
@@ -450,6 +452,47 @@ else {
...
+if (!signIn.token.hasTokensInUrl()) {
+  signIn.renderEl({el: '#osw-container'});
+}
+
+// The user has redirected back after authenticating and has their access and/or
+// ID Token in the URL.
+else {
+  signIn.token.parseAndStoreTokensFromUrl()
+  .then(function() {
+    const idToken = signIn.tokenManager.get('idToken');
```
santhoshbalakrishnan: Would making idToken in Caps make it look like it's constant?

signIn.tokenManager.get('ID_TOKEN');

This way even the next part where we add alias can look more explicit:

signIn.token.parseAndStoreTokensFromUrl({ 'ID_TOKEN': 'my-id-token' })

jmelberg-okta: Great idea! When I was thinking about this originally, I wanted to stick with the camelcase convention, but this is MUCH more explicit. 👍

README.md
```diff
@@ -450,6 +452,51 @@ else {
...
+    console.log(idToken);
+  });
+
+  // Optionally - use a different key for the tokens
+  signIn.token.parseAndStoreTokensFromUrl({
+    ACCESS_TOKEN: 'access-token-key',
```
nbarbettini: If I want to store things in the tokenManager with my own keys, why not just use `parseTokensFromUrl`? I think `parseAndStoreTokensFromUrl` should be opinionated.

src/util/Handlers.js
```diff
@@ -0,0 +1,51 @@
...
+        if (tokens[i].idToken) {
+          tokenManager.add(tokenKeys.idToken, tokens[i]);
+        } else if (tokens[i].accessToken) {
+          tokenManager.add(tokenKeys.accessToken, tokens[i]);
+        }
+      }
+    };
+  };
+
+  Handlers.defaultErrorHandler = function(err) {
```
nbarbettini: Can we reuse this for what we added in https://github.com/okta/okta-signin-widget/pull/466?

src/util/Handlers.js
```diff
@@ -0,0 +1,51 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define(['util/Logger'], function (Logger) {
+
+  var Handlers = {};
+
+  Handlers.defaultSuccessTokenHandler = function(tokenManager, keys = {}) {
```
nbarbettini: Looking ahead to #487, could this also be used to handle tokens received via `success(res)` in a silent login flow?

README.md
jmelberg-okta: Developers would need to re-implement logic that we're already providing. By extending this method to optionally take in keys for the tokens, they'd be equipped to:

1. Support multiple tokens of a single type
2. Namespace tokens

src/util/Handlers.js
jmelberg-okta: Internally, yes. This would apply to any error handler currently with the Widget. In #466, we [hardcoded the default handler](https://github.com/okta/okta-signin-widget/pull/466/files#diff-d57a8731b033a3e50aa86ec27435a102R85) to ensure a global change in the function wouldn't cause unexpected changes (i.e change from `.error` to `.warn`).

The previous design pattern forced developers to include their own `success` and `error` handlers, where this new approach (BYO or use ours) provides developers with significantly more flexibility.

src/util/Handlers.js
jmelberg-okta: Correct - with the assumption that a new interface is exposed.

We run into this problem currently where exposed Widget APIs expect callbacks to handle `async` methods (Why this is treated as an `async` function is another thing to look into).

PR #495, opened 2018-07-09
:bug: Prevents XSS when hovering over help icons

gowthamidommety-okta:
Resolves: OKTA-175902

NOTE: This is for the sign-in-widget side of changes. (From the JIRA - Escape the values that are controlled by the user in the UI) The other recommendation will be taken care of in a separate PR for okta-core.

screen shot 2018-07-10 at 3 35 47 pm

screen shot 2018-07-10 at 3 35 51 pm

### DIFF
--- a/src/EnrollOnPremController.js
+++ b/src/EnrollOnPremController.js
@@ -89,7 +89,7 @@ function (Okta, FormType, FormController, Footer, TextBox) {
             type: 'text',
             placeholder: Okta.loc('enroll.onprem.username.placeholder', 'login', [vendorName]),
             params: {
-              innerTooltip: Okta.loc('enroll.onprem.username.tooltip', 'login', [vendorName])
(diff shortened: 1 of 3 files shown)


src/EnrollOnPremController.js
```diff
@@ -89,7 +89,7 @@ function (Okta, FormType, FormController, Footer, TextBox) {
             type: 'text',
             placeholder: Okta.loc('enroll.onprem.username.placeholder', 'login', [vendorName]),
```
mauriciocastillosilva-okta: placeholder doesn't have a problem?

gowthamidommety-okta: I've been testing this and it seems ok. Looks like Handlebars escapes the placeholder param passed (in courage).

mauriciocastillosilva-okta: ah right. `placeholder` is used in the template... but tooltip is not, it is actually passed to qtip.
Makes sense then why it is not needed here.

travismorrow-okta:
LGTM

PR #504, opened 2018-07-25
Moved rm -rf to rimraf for Windows builds to work

santhoshbalakrishnan:
## Description

- Currently npm run clean uses `rm -rf`. This blocks us from building sign in widget repo on any Windows machine.

## Reviewers

- @rchild-okta 
- @jmelberg-okta

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@okta/okta-signin-widget",
-  "version": "2.9.0",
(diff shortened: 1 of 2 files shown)


santhoshbalakrishnan:
Thanks for pointing that out! This one is a subset of that change. We can get this in first so that we can get Windows builds to working stage quickly and then clean the tasks up.

PR #516, opened 2018-08-08
Fix for Okta-165796 - cancel polling if factor was switched

magizh-okta:
* If factor was switched before initial delay the poll still hapened.
Now we listen on factor switch and then clear the timout causing the delay.

* Once we go past the initial delay we stop listening on the factor change because
cancelling poll is handled by AuthJS once the poll is initiated.
eg) Once previous is invoked as a result of switching factor AuthJS cancells poll.

pushfix
pushfix2

Original issue 

pushfix3

Rejection of promise

pushfix4


test/unit/spec/MfaVerify_spec.js
```diff
@@ -394,6 +394,10 @@ function (Okta,
     }
 
     function setupPolling(test, finalResponse) {
+      // This is to reduce delay beofre initiaing polling in the tests.
```
mauriciocastillosilva-okta: typo: `before`

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2301,6 +2305,76 @@ function (Okta,
...
+            itp('starts poll after a delay of 6000ms', function () {
+              return setupOktaPush().then(function (test) {
+                spyOn(LoginUtil, 'callAfterTimeout').and.callFake(function() {
+                  // reducing the timeout to 100 so that test is fast.
+                  return setTimeout(arguments[0], 100);
+                });
+                $.ajax.calls.reset();
+                test.setNextResponse(resChallengePush);
+                test.form[0].submit();
+                return tick(test).then(function() {
```
mauriciocastillosilva-okta: we should use, if possible, some `Expect.waitFor*` instead of `tick`. That is preferred since it is more deterministic. Maybe wait for button to be disabled or an spy being called
Applies for all the new tests

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2301,6 +2305,76 @@ function (Okta,
...
+                  expect(LoginUtil.callAfterTimeout.calls.argsFor(1)[1]).toBe(6000);
+                  var deferred = Q.defer();
+                  expect(test.router.controller.model.appState.get('transaction').status).toBe('MFA_CHALLENGE');
+                  var transaction = test.router.controller.model.appState.get('transaction');
+                  spyOn(transaction, 'poll');
+                  expect(transaction.poll).not.toHaveBeenCalled();
```
mauriciocastillosilva-okta: is this needed? If you just spied on `poll` on the line before, then there is no other place where this would be called, right?

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2301,6 +2305,76 @@ function (Okta,
...
+                  expect(test.form.submitButton().prop('disabled')).toBe(false);
+
+                  // Setup responses
+                  $.ajax.calls.reset();
+                  test.setNextResponse([resChallengePush, resChallengePush, resSuccess]);
+
+                  // Click submit
+                  test.form.submit();
+                  spyOn(test.router.controller.model, 'stopListening').and.callThrough();
+                  return tick()
```
mauriciocastillosilva-okta: instead of all this `ticks`, can we do something similar to the above tests?

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2360,6 +2434,10 @@ function (Okta,
             });
             itp('re-enables submit and displays an error when request fails', function () {
               function setupFailurePolling(test) {
+                // This is to reduce delay beofre initiaing polling in the tests.
```
mauriciocastillosilva-okta: typo: `before`

test/unit/spec/MfaVerify_spec.js
santhoshbalakrishnan: typo: `initiating`

test/unit/spec/MfaVerify_spec.js
santhoshbalakrishnan: typo: `initiating` 😄

test/unit/spec/MfaVerify_spec.js
magizh-okta: Yeah I guess more ticks are not needed here. Modified the test.

src/models/Factor.js
```diff
@@ -249,7 +250,14 @@ function (Okta, Q, factorUtil, BaseLoginModel) {
           setTransaction(trans);
           // In Okta verify case we initiate poll.
           if (trans.status === 'MFA_CHALLENGE' && trans.poll) {
-            return Q.delay(PUSH_INTERVAL).then(function() {
+            const deferred = Q.defer();
+            const initiatePollTimout = Util.callAfterTimeout(deferred.resolve, PUSH_INTERVAL);
+            self.listenToOnce(self.options.appState, 'factorSwitched', () => {
+              clearTimeout(initiatePollTimout);
```
mauriciocastillosilva-okta: should reject the promise in addition to canceling the timeout?

src/models/BaseLoginModel.js
```diff
@@ -12,9 +12,10 @@
...
   'okta',
-  'vendor/lib/q'
+  'vendor/lib/q',
+  'util/Enums',
 ],
-function (Okta, Q) {
+function (Okta, Q, Enums) {
 
   var _ = Okta._;
   var KNOWN_ERRORS = [
```
santhoshbalakrishnan: Interesting why `AuthPollStopError` was added here since we will return before that. Curious why it was added here. Should we add `AUTH_STOP_POLL_INITIATION_ERROR` in case this array is going to be used elsewhere, since this is also going to be a known error?

magizh-okta: Lol yeah this has to be cleaned up.  I checked its usage and seems like it doesnt make sense to add it to Known errors array.

PR #519, opened 2018-08-10
Change save button value to be function in IDPDiscoveryForm due to i18n bundle lazy loading.

haishengwu-okta:
language bundle will be loaded via jsonp request after loading the javascript file which execute the expression `Okta.loc(...)` at first place. 

`IDPDiscoveryForm`

```
define([
  'okta',
  'views/primary-auth/PrimaryAuthForm'
], function (Okta, PrimaryAuthForm) {

  var _ = Okta._;

  return PrimaryAuthForm.extend({
    className: 'idp-discovery-form',
    save: Okta.loc('oform.next', 'login'),
...
```

OKTA-183830

### DIFF
--- a/src/views/idp-discovery/IDPDiscoveryForm.js
+++ b/src/views/idp-discovery/IDPDiscoveryForm.js
@@ -19,7 +19,9 @@ define([
 
   return PrimaryAuthForm.extend({
     className: 'idp-discovery-form',
-    save: Okta.loc('oform.next', 'login'),
+    save: function () {
+      return Okta.loc('oform.next', 'login');
+    },
     saveId: 'idp-discovery-submit',
 
     initialize: function () {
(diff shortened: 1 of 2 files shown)


src/views/idp-discovery/IDPDiscoveryForm.js
```diff
@@ -19,7 +19,9 @@ define([
 
   return PrimaryAuthForm.extend({
     className: 'idp-discovery-form',
-    save: Okta.loc('oform.next', 'login'),
```
hor-kanchan-okta: You can try `_.partial(Okta.loc, 'oform.next', 'login')`

haishengwu-okta:
forgot to add screenshot ;)

screen shot 2018-08-10 at 1 13 42 pm

nbarbettini:
I am interested in @mauriciocastillosilva-okta's question too:

> Why is this a problem in IdP Form and not in any other form? What is special about the idp form?

rchild-okta:
FYI @mauriciocastillosilva-okta and @nbarbettini, after the change to [support dynamically configured locale bundles](https://github.com/okta/okta-signin-widget/pull/115), all forms need to keep this in mind. If it's in a template, it will be calculated at the right time (after the bundle is loaded). But, there are these cases where our pattern for calling `Okta.loc` at module load time will not work.

This does bring up a good point though - we should probably write an eslint rule to catch these upfront, or write some helper function to do this behind the scenes.

PR #521, opened 2018-08-15
:seedling: Added Typing Biometrics

magizh-okta:
* Added ability to track user typing pattern behind a flag trackTypingPattern.
* Using external TypingDNA library https://github.com/TypingDNA/TypingDnaRecorder-JavaScript
Checking in it in as it is not available as a npm dist.

@rpamidimarri-okta  @chrisleung-okta  
The X-Typing-Pattern header is 
1) **null** when - there is any error in capturing the pattern or in cases where usernams are autofilled. 
2) **undefined** when flag is off
3) present - if flag is on and data was captured successfully

Let me know if this is okay. 

screen shot 2018-08-14 at 3 50 05 pm

Edited  username
typing

### DIFF
--- a/src/models/PrimaryAuth.js
+++ b/src/models/PrimaryAuth.js
@@ -160,13 +160,15 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
     },
 
     doPrimaryAuth: function (authClient, signInArgs, func) {
-      var deviceFingerprintEnabled = this.settings.get('features.deviceFingerprinting');
+      var deviceFingerprintEnabled = this.settings.get('features.deviceFingerprinting'),
+          typingPatternEnabled = this.settings.get('features.trackTypingPattern');
+
       // Add the custom header for fingerprint if needed, and then remove it afterwards
       // Since we only need to send it for primary auth
       if (deviceFingerprintEnabled) {
         authClient.options.headers['X-Device-Fingerprint'] = this.appState.get('deviceFingerprint');
       }
-      if (this.settings.get('features.trackTypingPattern')) {
+      if (typingPatternEnabled) {
         authClient.options.headers['X-Typing-Pattern'] = this.appState.get('typingPattern');
       }
       var self = this;
@@ -175,6 +177,10 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
         if (deviceFingerprintEnabled) {
           delete authClient.options.headers['X-Device-Fingerprint'];
           self.appState.unset('deviceFingerprint'); //Fingerprint can only be used once
+        }
+        if (typingPatternEnabled) {
+          delete authClient.options.headers['X-Typing-Pattern'];
+          self.appState.unset('deviceFingerprint');
         }
       });
     }
--- a/src/models/PrimaryAuth.js
+++ b/src/models/PrimaryAuth.js
@@ -163,7 +163,7 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
       var deviceFingerprintEnabled = this.settings.get('features.deviceFingerprinting'),
           typingPatternEnabled = this.settings.get('features.trackTypingPattern');
 
-      // Add the custom header for fingerprint if needed, and then remove it afterwards
+      // Add the custom header for fingerprint, typing pattern if needed, and then remove it afterwards
       // Since we only need to send it for primary auth
       if (deviceFingerprintEnabled) {
         authClient.options.headers['X-Device-Fingerprint'] = this.appState.get('deviceFingerprint');
@@ -180,7 +180,7 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
         }
         if (typingPatternEnabled) {
           delete authClient.options.headers['X-Typing-Pattern'];
-          self.appState.unset('deviceFingerprint');
+          self.appState.unset('typingPattern');
         }
       });
     }
--- a/src/models/PrimaryAuth.js
+++ b/src/models/PrimaryAuth.js
@@ -109,13 +109,11 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
       if (this.appState.get('isUnauthenticated') && !this.settings.get('features.passwordlessAuth')) {
         primaryAuthPromise = this.doTransaction(function (transaction) {
           var authClient = this.appState.settings.authClient;
-          return this.doPrimaryAuth(authClient, signInArgs,
-                                    transaction.authenticate);
+          return this.doPrimaryAuth(authClient, signInArgs, transaction.authenticate);
         });
       } else {
         primaryAuthPromise = this.startTransaction(function (authClient) {
-          return this.doPrimaryAuth(authClient, signInArgs,
-                                    _.bind(authClient.signIn, authClient));
+          return this.doPrimaryAuth(authClient, signInArgs, _.bind(authClient.signIn, authClient));
         });
       }
 
--- a/src/models/PrimaryAuth.js
+++ b/src/models/PrimaryAuth.js
@@ -92,8 +92,7 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
     save: function () {
       var username = this.settings.transformUsername(this.get('username'), Enums.PRIMARY_AUTH),
           remember = this.get('remember'),
-          lastUsername = this.get('lastUsername'),
-          deviceFingerprintEnabled = this.settings.get('features.deviceFingerprinting');
+          lastUsername = this.get('lastUsername');
 
       this.setUsernameCookie(username, remember, lastUsername);
 
@@ -106,15 +105,16 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
       var signInArgs = this.getSignInArgs(username);
 
       var primaryAuthPromise;
+
       if (this.appState.get('isUnauthenticated') && !this.settings.get('features.passwordlessAuth')) {
         primaryAuthPromise = this.doTransaction(function (transaction) {
           var authClient = this.appState.settings.authClient;
-          return this.doPrimaryAuth(authClient, deviceFingerprintEnabled, signInArgs,
+          return this.doPrimaryAuth(authClient, signInArgs,
                                     transaction.authenticate);
         });
       } else {
         primaryAuthPromise = this.startTransaction(function (authClient) {
-          return this.doPrimaryAuth(authClient, deviceFingerprintEnabled, signInArgs,
+          return this.doPrimaryAuth(authClient, signInArgs,
                                     _.bind(authClient.signIn, authClient));
         });
       }
@@ -161,11 +161,15 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
       }
     },
 
-    doPrimaryAuth: function (authClient, deviceFingerprintEnabled, signInArgs, func) {
+    doPrimaryAuth: function (authClient, signInArgs, func) {
+      var deviceFingerprintEnabled = this.settings.get('features.deviceFingerprinting');
       // Add the custom header for fingerprint if needed, and then remove it afterwards
       // Since we only need to send it for primary auth
       if (deviceFingerprintEnabled) {
         authClient.options.headers['X-Device-Fingerprint'] = this.appState.get('deviceFingerprint');
+      }
+      if (this.settings.get('features.trackTypingPattern')) {
+        authClient.options.headers['X-Typing-Pattern'] = this.appState.get('typingPatern');
       }
       var self = this;
       return func(signInArgs)
--- a/src/models/PrimaryAuth.js
+++ b/src/models/PrimaryAuth.js
@@ -169,7 +169,7 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
         authClient.options.headers['X-Device-Fingerprint'] = this.appState.get('deviceFingerprint');
       }
       if (this.settings.get('features.trackTypingPattern')) {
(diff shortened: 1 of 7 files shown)


test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -885,6 +888,72 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
       });
     });
 
+    Expect.describe('Typing biometrics', function () {
+      itp('does not contain typing pattern header in primary auth request if feature is disabled', function () {
+        return setup({ features: { trackTypingPattern: false }})
+        .then(function (test) {
+          $.ajax.calls.reset();
+          test.form.setUsername('testuser');
```
ruthadrineda-okta: Can you also add a test for what X-Typing-Pattern header contains when username is null and when username is just a space char? Or will the sign-in widget catch these cases before getTypingPattern() gets called?

src/vendor/TypingDNA.js
```diff
@@ -0,0 +1,1258 @@
+/**
+ * TypingDNA - Typing Biometrics JavaScript API
+ * https://api.typingdna.com/scripts/typingdna.js
```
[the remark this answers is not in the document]
magizh-okta: Dont want to make courage heavier than what it is now and as of now there is no use case in courage.

src/util/TypingUtil.js
```diff
@@ -0,0 +1,34 @@
...
+define(['vendor/TypingDNA'], function (TypingDNA) {
+  var tdna;
+  return {
+    track: function(selector) {
+      try {
+        tdna = new TypingDNA();
+        tdna.addTarget(selector);
+        tdna.start();
+      } catch (e) {
+        // Issues in typing should not stop Primary auth.
```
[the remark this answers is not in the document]
magizh-okta: No need to log here. When we get a null value for the header we will naturally know something went wrong. We will be setting up splunk dashboards for that.

test/unit/spec/PrimaryAuth_spec.js
magizh-okta: when username is null form cannot be submitted so as you said sign-in widget catch these cases before getTypingPattern() gets called. When username is just a space it would be treated as another character and typing pattern would be computed.  I am not seeing a need to testing this unless I am missing something.

ruthadrineda-okta: Just curious if X-Typing-Pattern will have an empty value or have the corresponding pattern for the space character. What would be better I think is if the sign-in widget trimmed the trailing space, caught the empty username case, and not need to make subsequent calls to getTypingPattern(), etc.

chrisleung-okta: According to the TypingDNA api and what I've seen, the X-Typing-Pattern header can only produce either one of 'null' or the typing pattern. If there is no typing information that can be caught (like when username is empty or autofilled), TypingDNA produces 'null'.

magizh-okta: @ruthadrineda-okta  agreed! I think limiting the submission of a space should be considered in general even in courage. (Will bring it up in UI office hours and create a JIRA if required). But outside the scope of this JIRA. For now space isa  valid character and honestly getTypingPattern is not expensive that we need to worry about it for typing patterns. 
https://oktawiki.atlassian.net/wiki/spaces/eng/pages/236645615/UICore+Office+Hours+Meeting+Notes

src/vendor/TypingDNA.js
```diff
@@ -0,0 +1,1258 @@
...
+ * https://typingdna.com/scripts/typingdna.js (alternative)
+ *
+ * @version 2.15
+ * @author Raul Popa
+ * @copyright SC TypingDNA SRL, http://typingdna.com
+ * @license http://www.apache.org/licenses/LICENSE-2.0
```
magizh-okta: @rchild-okta  I had talked about this with you earlier. Wanted to double check if it is okay to add this library here. Should we notify legal. Mau was mentioning that we maintain a list of things external libs we use under a particular license and we might want to add this to that list.

src/models/PrimaryAuth.js
```diff
@@ -106,15 +105,16 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
       var signInArgs = this.getSignInArgs(username);
 
       var primaryAuthPromise;
+
       if (this.appState.get('isUnauthenticated') && !this.settings.get('features.passwordlessAuth')) {
         primaryAuthPromise = this.doTransaction(function (transaction) {
           var authClient = this.appState.settings.authClient;
-          return this.doPrimaryAuth(authClient, deviceFingerprintEnabled, signInArgs,
+          return this.doPrimaryAuth(authClient, signInArgs,
```
mauriciocastillosilva-okta: nit: we probably can have all arguments in 1 line. Same for line 117

src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -167,8 +173,13 @@ define([
       } else if (!this.settings.get('features.passwordlessAuth')) {
         this.getInputs().toArray()[1].focus();
       }
-    }
+    },
 
+    postRender: function() {
+      if (this.settings.get('features.trackTypingPattern')) {
+        TypingUtil.track('okta-signin-username');
```
mauriciocastillosilva-okta: 2 questions here
- is the selector assumed as an id? Asking since we don't pass `#` or `.` or anything specifying what type of selector it is
- do we want to do this in `postRender`? or do we need to start tracking on focus for the username field?

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -885,6 +888,72 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
...
+          test.form.setPassword('pass');
+          test.setNextResponse(resSuccess);
+          spyOn(TypingUtil, 'track').and.callFake(function (target) {
+            expect(target).toBe('okta-signin-username');
+          });
+          test.form.submit();
+          return tick();
```
mauriciocastillosilva-okta: can we replace this with a different wait? maybe `return Expect.waitForSpyCall(test.successSpy)`
(trying to avoid `tick`s if possible since ticks are not really deterministic)

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -885,6 +888,72 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
...
+          test.form.setPassword('pass');
+          test.setNextResponse(resSuccess);
+          spyOn(TypingUtil, 'track').and.callFake(function (target) {
+            expect(target).toBe('okta-signin-username');
+          });
+          spyOn(TypingUtil, 'getTypingPattern').and.callFake(function () {
+            return typingPattern;
+          });
+          test.form.submit();
+          return tick();
```
mauriciocastillosilva-okta: same comment here about not using `tick` and use `Expect.waitForSpyCall` instead.
Same for other tests that are added

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -885,6 +888,72 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthFo
...
+          spyOn(TypingUtil, 'getTypingPattern').and.callFake(function () {
+            return null;
+          });
+          test.form.submit();
+          return tick();
+        })
+        .then(function () {
+          expect($.ajax.calls.count()).toBe(1);
+          var ajaxArgs = $.ajax.calls.argsFor(0);
+          expect(ajaxArgs[0].headers['X-Typing-Pattern']).toBe(null);
```
mauriciocastillosilva-okta: does it make a difference in our api if it is null or undefined?

magizh-okta: This is more for splunk dashboards. Just wanted to capture the cases where flag is tuned on but typing patter was not captured.

src/vendor/TypingDNA.js
magizh-okta: Checked with Reman looks like apache is whitelisted and we dont have to worry about this. Also Raghu had talked with Security to check if it is okay to use this library. Will check with both again before we go beta.

src/models/Settings.js
```diff
@@ -87,6 +87,7 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, SharedUtil
       'features.idpDiscovery': ['boolean', false, false],
       'features.passwordlessAuth': ['boolean', false, false],
       'features.showPasswordToggleOnSignInPage': ['boolean', false, false],
+      'features.trackTypingPattern': ['boolean', false, false],
```
rpamidimarri-okta: where is the map from `trackTypingPattern` to the actual feature name `TYPING_PATTERN_COLLECTION`?

rpamidimarri-okta:
> The X-Typing-Pattern header is

null when - there is any error in capturing the pattern or in cases where usernams are autofilled.
undefined when flag is off
present - if flag is on and data was captured successfully

@magizh-okta makes sense and the change looks good to me.

src/models/Settings.js
magizh-okta: That would happen in Okta-core when we bootstrap the sign in widget. That is how every feature is mapped.

rpamidimarri-okta: cool, thanks

src/models/PrimaryAuth.js
```diff
@@ -161,12 +159,16 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
       }
     },
 
-    doPrimaryAuth: function (authClient, deviceFingerprintEnabled, signInArgs, func) {
+    doPrimaryAuth: function (authClient, signInArgs, func) {
+      var deviceFingerprintEnabled = this.settings.get('features.deviceFingerprinting');
       // Add the custom header for fingerprint if needed, and then remove it afterwards
       // Since we only need to send it for primary auth
```
mauriciocastillosilva-okta: I see that we are not removing the header after primary auth...  As far as I remember `authClient.options.headers` persist after this call, so we will be sending an empty header in all the requests after this one. Thoughts? Might be a good idea since we only need it for primaryAuth when we capture the username pattern.

src/util/TypingUtil.js
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+define(['vendor/TypingDNA'], function (TypingDNA) {
+  var tdna;
+  return {
+    track: function(selector) {
```
mauriciocastillosilva-okta: nit: since this works with an element `id`, maybe we should name the argument `selectorId`

src/models/PrimaryAuth.js
magizh-okta: Good catch ! Fixed now.

src/models/PrimaryAuth.js
```diff
@@ -161,19 +159,29 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
       }
     },
 
-    doPrimaryAuth: function (authClient, deviceFingerprintEnabled, signInArgs, func) {
+    doPrimaryAuth: function (authClient, signInArgs, func) {
+      var deviceFingerprintEnabled = this.settings.get('features.deviceFingerprinting'),
+          typingPatternEnabled = this.settings.get('features.trackTypingPattern');
+
       // Add the custom header for fingerprint if needed, and then remove it afterwards
```
mauriciocastillosilva-okta: can we update this comment saying it is for `typingPattern` too?

src/models/PrimaryAuth.js
```diff
@@ -161,19 +159,29 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
...
       var self = this;
       return func(signInArgs)
       .fin(function () {
         if (deviceFingerprintEnabled) {
           delete authClient.options.headers['X-Device-Fingerprint'];
           self.appState.unset('deviceFingerprint'); //Fingerprint can only be used once
         }
+        if (typingPatternEnabled) {
+          delete authClient.options.headers['X-Typing-Pattern'];
+          self.appState.unset('deviceFingerprint');
```
mauriciocastillosilva-okta: this should be unsetting `typingPattern` instead of `deviceFingerprint`

src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -167,8 +173,9 @@ define([
       } else if (!this.settings.get('features.passwordlessAuth')) {
         this.getInputs().toArray()[1].focus();
       }
+      if (this.settings.get('features.trackTypingPattern')) {
+        TypingUtil.track('okta-signin-username');
```
mauriciocastillosilva-okta: just checking: if that selector id does not exist in the dom, it returns null right? so all works fine in that case.

magizh-okta: If Id is not valid it tracks only clicks and special character press. Doest really stop authn though.

https://github.com/TypingDNA/TypingDnaRecorder-JavaScript#typingdnaaddtargetelement_id

mauriciocastillosilva-okta: ok... we can keep an eye on this. Otherwise we can move this logic to line 172, inside the `if` block for when username is empty and the user needs to type it.

PR #522, opened 2018-08-16
:seedling: Supporting IdP Discovery Shortname

petermiller-okta:
Resolves: OKTA-184692

Reviewers: @mauriciocastillosilva-okta @haishengwu-okta @yuliu-okta

### DIFF
--- a/src/models/IDPDiscovery.js
+++ b/src/models/IDPDiscovery.js
@@ -30,7 +30,6 @@ function (Okta, PrimaryAuthModel, CookieUtil, Enums, Util) {
       return {
         username: {
           type: 'string',
-          format: 'email',
           validate: function (value) {
             if(_.isEmpty(value)) {
               return Okta.loc('model.validation.field.blank', 'login');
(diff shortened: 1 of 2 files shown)


src/models/IDPDiscovery.js
```diff
@@ -50,7 +49,7 @@ function (Okta, PrimaryAuthModel, CookieUtil, Enums, Util) {
       var username = this.settings.transformUsername(this.get('username'), Enums.IDP_DISCOVERY),
           remember = this.get('remember'),
           lastUsername = this.get('lastUsername'),
-          resource = 'acct:' + username,
+          resource = 'okta:acct:' + username,
```
mauriciocastillosilva-okta: does this mean that previous versions of the widget won't work anymore for idp discovery? Since those versions will be sending `acct:username` without the `okta:` prefix

petermiller-okta: https://github.com/okta/okta-core/pull/29042 will support both schemas

PR #523, opened 2018-08-22
webauthn factor enduser flow

yumingcao-okta:
Introducing a new webauthn factor which is going to supercede the legacy Windows Hello factor and U2F factor. The factor interacts with `webauthn` factorType in `/authn` API. 

We use U2F labels for now until we figure out the new texts. The factor only support U2F devices for now, and will support other authenticators like Windows Hello in later phase.

This factor requires users to have a webauthn supported browser, non-webauthn browser users needs to fallback to legacy U2F client library for enrollment and verification, this will be addressed separately.

Backend PR: https://github.com/okta/okta-core/pull/29034


src/util/FactorUtil.js
```diff
@@ -155,8 +157,15 @@ function (Okta, TimeUtil) {
       return 'QUESTION';
     }
     if (provider === 'FIDO' && factorType === 'webauthn') {
-      return 'WINDOWS_HELLO';
-    }
+      if (this.settings != null && this.settings.get('features.webauthn') != null) {
+        webauthnEnabled = this.settings.get('features.webauthn');
```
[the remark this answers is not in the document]
yumingcao-okta: hmm. upon successful primary auth, this function is called in two paths: one when `Factor.factorName` is being derived, and one when `Factor.factorLabel` model is derived, it doesn't seem to have access to `this.settings` for the second path, any advice?

src/EnrollWebauthnController.js
```diff
@@ -0,0 +1,157 @@
...
+        });
+      },
+
+      activate: function () {
+        this.set('__enrolled__', true);
+        this.trigger('errors:clear');
+
+        return this.doTransaction(function (transaction) {
+          var activation = transaction.factor.activation;
+          var deferred = Q.defer();
```
rchild-okta: Nit: No need to have a separate defer for this. Since we're already creating a promise with `navigator.credentials.create`, we can just use that:

```js
return navigator.credentials.create({ publicKey: options })
  .then(function () {
    return transaction.activate(/* stuff */);
  })
  .catch(function () {
    throw new Error({ xhr: 'stuff' });
  });
```

rchild-okta: Also, no need for all the names of "publicKey" - we go from activation, to options, to publickey. Why not just:

```js
var publicKey = _.extend({}, transaction.factor.activation, {
  challenge: stuff,
  user: {
    id: stuff
  }
});
```

src/EnrollWebauthnController.js
```diff
@@ -0,0 +1,157 @@
...
+
+    Form: {
+      title: _.partial(Okta.loc, 'enroll.u2f.title', 'login'),
+      save: _.partial(Okta.loc, 'enroll.u2f.save', 'login'),
+      noCancelButton: true,
+      hasSavingState: false,
+      autoSave: true,
+      className: 'enroll-u2f-form',
+      noButtonBar: function () {
+        return navigator.credentials.create == null;
```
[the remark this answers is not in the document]
rchild-okta: Agree - there are a lot of places in this PR where we're depending on functionality that's not x-browser compatible. There needs to be more checks. For example, guessing this line (and all places that don't check for navigator.credentials first) will throw an error in non-chrome browsers.

src/EnrollWebauthnController.js
```diff
@@ -0,0 +1,157 @@
...
+        else {
+          //There is html in enroll.u2f.general2 in our properties file, reason why is unescaped
+          result.push(FormType.View({
+            View: '<div class="u2f-instructions"><ol>\
+          <li>{{{i18n code="enroll.u2f.general2" bundle="login"}}}</li>\
+          <li>{{i18n code="enroll.u2f.general3" bundle="login"}}</li>\
+          </ol></div>'
+          }));
+
+          result.push(FormType.View({
```
rchild-okta: Is there a reason this is a separate view from the previous one?

src/EnrollWebauthnController.js
```diff
@@ -0,0 +1,157 @@
...
+            <div data-se="u2f-waiting" class="okta-waiting-spinner"></div>\
+          </div>'
+          }));
+        }
+
+        return result;
+      },
+
+      _startEnrollment: function () {
+        this.$('.u2f-instructions').addClass('hide');
```
rchild-okta: Is there a reason you're using the class hide for these two, and using the method hide for o-form-button-bar?

And in general, since we're triggering these ourselves, could just squash this down to one `toggle` function and use the [toggleClass](http://api.jquery.com/toggleclass/) method with params for toggling true or false.

src/VerifyWebauthnController.js
```diff
@@ -0,0 +1,177 @@
...
+            factorType: 'webauthn',
+            provider: 'FIDO'
+          });
+
+          var self = this;
+          return factor.verify().then(function (transaction) {
+            var factorData = transaction.factor;
+            self.trigger('request');
+
+            var deferred = Q.defer();
```
rchild-okta: Same note here about the unnecessary deferred

src/VerifyWebauthnController.js
```diff
@@ -0,0 +1,177 @@
...
+            var challengeBytes = CryptoUtil.strToBin(factorData.challenge.challenge);
+            var options = factorData.challenge;
+
+            options.allowCredentials = [{
+              type: "public-key",
+              id: idBytes
+            }];
+            options.challenge = challengeBytes;
+
+            navigator.credentials.get({ "publicKey": options }).then(function (assertion) {
```
rchild-okta: nit: singlequotes; would've expected lint to catch this

src/VerifyWebauthnController.js
```diff
@@ -0,0 +1,177 @@
...
+              return factor.verify({
+                clientData: CryptoUtil.binToStr(assertion.response.clientDataJSON),
+                authenticatorData: CryptoUtil.binToStr(assertion.response.authenticatorData),
+                signatureData: CryptoUtil.binToStr(assertion.response.signature),
+                rememberDevice: rememberDevice
+              })
+              .then(deferred.resolve);
+            }).catch(function (err) {
+              self.trigger('errors:clear');
+                deferred.reject({
```
rchild-okta: nit: spacing

src/VerifyWebauthnController.js
```diff
@@ -0,0 +1,177 @@
...
+      },
+      modelEvents: {
+        'request': '_startEnrollment',
+        'error': '_stopEnrollment'
+      },
+
+      formChildren: function () {
+        var result = [];
+
+        // TODO: handle non-webauthn browser flow
```
rchild-okta: What do you mean by this? Is this referring to us not handling browsers that do not have webauthn support (but are being served a webauthn factors response and will break when they render this view)?

If yes, agree - but, we should do this before merging.

src/VerifyWebauthnController.js
```diff
@@ -0,0 +1,177 @@
...
+      noButtonBar: function () {
+        return navigator.credentials.create == null;
+      },
+      modelEvents: {
+        'request': '_startEnrollment',
+        'error': '_stopEnrollment'
+      },
+
+      formChildren: function () {
+        var result = [];
```
rchild-okta: nit: clearer to call this children so it's clearer in the downstream form what we're doing

src/util/FactorUtil.js
```diff
@@ -21,6 +21,8 @@ function (Okta, TimeUtil) {
 
   var fn = {};
 
+  var webauthnEnabled;
```
rchild-okta: nit: No need to put this up here - just put it closer to where you're using it

src/util/FactorUtil.js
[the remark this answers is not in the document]
rchild-okta: Haven't dug into the code, but wouldn't your description be a problem? I.e. in one of those paths, for the same factor, this will resolve to 'U2F' when it has access to settings, and 'WINDOWS_HELLO'.

src/util/FactorUtil.js
```diff
@@ -155,8 +157,15 @@ function (Okta, TimeUtil) {
       return 'QUESTION';
     }
     if (provider === 'FIDO' && factorType === 'webauthn') {
-      return 'WINDOWS_HELLO';
-    }
+      if (this.settings != null && this.settings.get('features.webauthn') != null) {
+        webauthnEnabled = this.settings.get('features.webauthn');
+      }
+      if (webauthnEnabled) {
+        return 'U2F';
```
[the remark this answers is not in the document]
rchild-okta: Just curious - are Windows Hello and WebAuthn compatible? This may be an extreme edge case, but what happens if an org turns off the windows_hello flag and enables the new webauthn flag?

yumingcao-okta:
@rchild-okta your description of UI/backend behavior is accurate. Regarding non webauthn supported browsers, we will need to send u2f js result to `webauthn` API and use clientData.type as a differentiator as we discussed in email thread. I would like to address that in a separate PR.

src/util/FactorUtil.js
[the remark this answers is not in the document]
yumingcao-okta: We'll figure out whether to migrate existing windows hello factor enrollments to webauthn factor when we add Hello support for the new factor. As for the ff case, definitely a weird case, I'll follow up on it. (also tracked in https://oktainc.atlassian.net/browse/OKTA-186817).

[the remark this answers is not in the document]
mauriciocastillosilva-okta: we should work on that jira soon, so we don't risk having an issue if both factors are enabled. Do we know when that jira will be worked on?

src/util/webauthn.js
```diff
@@ -63,6 +63,12 @@ function (Okta, Q) {
     getAssertion: getAssertion,
     isAvailable: function () {
       return window.hasOwnProperty('msCredentials');
+    },
+    isNewApiAvailable: function () {
+      return navigator && navigator.credentials && navigator.credentials.create;
+    },
+    isWebauthnOrU2fAvailable: function () {
+      return this.isNewApiAvailable() || window.hasOwnProperty('u2f');
```
mauriciocastillosilva-okta: can we have a function to check `window.hasOwnProperty('u2f')`? It seems that it is used in a few places in the U2FControllers (enroll and verify)... maybe it goes to FidoUtil? Asking so we have only in one place the logic to know if `u2f` is available or not

test/unit/helpers/xhr/MFA_CHALLENGE_webauthn.js
```diff
@@ -0,0 +1,47 @@
...
+      },
+      "factor": {
+        "id": "webauthnFactorId",
+        "factorType": "webauthn",
+        "provider": "FIDO",
+        "vendorName": "FIDO",
+        "profile": {"credentialId": "vdCxImCygaKmXS3S_2WwgqF1LLZ4i_2MKYfAbrNByJOOmSyRD_STj6VfhLQsLdLrIdgvdP5EmO1n9Tuw5BawZw"},
+        "_embedded": {
+          "challenge": {
+            "challenge": "kygOUtSWURMv_t_Gj71Y",
```
mauriciocastillosilva-okta: Probably this was discussed in the api review, but it seems a little redundant to have `challenge.challenge`... it used to be `challenge.nonce`.

test/unit/helpers/xhr/MFA_CHALLENGE_webauthn.js
```diff
@@ -0,0 +1,47 @@
...
+            "extensions": {
+              "appid": "https://foo.com"
+            }
+          }
+        }
+      },
+      "policy": {"allowRememberDevice": false, "rememberDeviceLifetimeInMinutes": 0, "rememberDeviceByDefault": false}
```
mauriciocastillosilva-okta: nit: each property in different line

test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_webauthn.js
```diff
@@ -0,0 +1,68 @@
...
+        "factorType":"webauthn",
+        "provider":"FIDO",
+        "vendorName":"FIDO",
+        "_embedded":{
+          "activation":{
+            "rp": {
+              "name": "acme"
+            },
+            "user": {
+              "id": "00u1212qZXXap6Cts0g4",
```
mauriciocastillosilva-okta: might want to change to test data

test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_webauthn.js
```diff
@@ -0,0 +1,68 @@
...
+        }
+      },
+      "factor": {
+        "id":"fuf52dhWPdJAbqiUU0g4",
+        "factorType":"webauthn",
+        "provider":"FIDO",
+        "vendorName":"FIDO",
+        "_embedded":{
+          "activation":{
+            "rp": {
```
mauriciocastillosilva-okta: just for me to learn, what is `rp`?

test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_webauthn.js
```diff
@@ -0,0 +1,68 @@
...
+              "type": "public-key",
+              "alg": -7
+            }],
+            "challenge": "G7bIvwrJJ33WCEp6GGSH",
+            "authenticatorSelection": {
+              "authenticatorAttachment": "cross-platform",
+              "requireResidentKey": false,
+              "userVerification": "preferred"
+            },
+            "u2fParams": {
```
mauriciocastillosilva-okta: would like to know the reason this is not one level above, `activation.appId` since it seems it is useful in general to know what is the app

test/unit/spec/EnrollWebauthn_spec.js
```diff
@@ -0,0 +1,304 @@
...
+        deferred.reject({ message: 'something went wrong' });
+        return deferred.promise;
+      });
+    }
+
+    Expect.describe('Header & Footer', function () {
+      itp('displays the correct factorBeacon', function () {
+        return setup().then(function (test) {
+          expect(test.beacon.isFactorBeacon()).toBe(true);
+          expect(test.beacon.hasClass('mfa-u2f')).toBe(true);
```
mauriciocastillosilva-okta: shouldn't this be `mfa-webauthn`?

test/unit/spec/EnrollWebauthn_spec.js
```diff
@@ -0,0 +1,304 @@
...
+      itp('shows instructions and a register button', function () {
+        spyOn(webauthn, 'isWebauthnOrU2fAvailable').and.returnValue(true);
+        return setup().then(function (test) {
+          Expect.isVisible(test.form.enrollInstructions());
+          Expect.isVisible(test.form.submitButton());
+        });
+      });
+
+      itp('shows a waiting spinner and devices images after submitting the form', function () {
+        spyOn(webauthn, 'isWebauthnOrU2fAvailable').and.returnValue(true);
```
mauriciocastillosilva-okta: can this be part of `mockWebauthnSuccessRegistration();`?

test/unit/spec/EnrollWebauthn_spec.js
```diff
@@ -0,0 +1,304 @@
...
+                type: 'public-key',
+                alg: -7
+              }],
+              challenge: CryptoUtil.strToBin('G7bIvwrJJ33WCEp6GGSH'),
+              authenticatorSelection: {
+                authenticatorAttachment: 'cross-platform',
+                requireResidentKey: false,
+                userVerification: 'preferred'
+              },
+              u2fParams: {
```
mauriciocastillosilva-okta: do we need this in the data for `navigator.credentials.create`? If not, we should remove it

test/unit/spec/FidoUtil_spec.js
```diff
@@ -0,0 +1,11 @@
+define([
+  'helpers/util/Expect',
+  'util/FidoUtil'
+],
+function (Expect, FidoUtil) {
+  Expect.describe('FidoUtil', function () {
+    it('Gets U2F version', function () {
+      expect(FidoUtil.getU2fVersion()).toEqual('U2F_V2');
+    });
+  });
```
mauriciocastillosilva-okta: I know it was just a refactor, but can we add a couple of simple tests for the other 2 functions in FidoUtil?

test/unit/spec/VerifyWebauthn_spec.js
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  function mockWebauthnSignPending() {
+    spyOn(navigator.credentials, 'get').and.returnValue(Q.defer().promise);
+  }
+
+  function setupWebauthnFactor(options) {
+    options || (options = {});
+
+    spyOn(webauthn, 'isWebauthnOrU2fAvailable').and.returnValue(options.u2fSupported == true);
```
mauriciocastillosilva-okta: should be 3 `=`, `options.u2fSupported === true`

test/unit/spec/VerifyWebauthn_spec.js
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  function expectTitleToBe(test, desiredTitle) {
+    expect(test.form.titleText()).toBe(desiredTitle);
+  }
+
+  Expect.describe('Webauthn Factor', function () {
+    itp('shows the right beacon and title for webauthn', function () {
+      return setupWebauthnFactor({webauthnSupported: true, u2fSupported: true}).then(function (test) {
+        expectHasRightBeaconImage(test, 'mfa-u2f');
```
mauriciocastillosilva-okta: should be `mfa-webauthn`

test/unit/spec/VerifyWebauthn_spec.js
```diff
@@ -0,0 +1,271 @@
...
+          data: {
+            clientData: testClientData,
+            signatureData: testSignature,
+            stateToken: 'testStateToken'
+          }
+        });
+      });
+    });
+  });
+});
```
mauriciocastillosilva-okta: nit: new line

src/models/Settings.js
```diff
@@ -77,6 +77,7 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
       'features.callRecovery': ['boolean', true, false],
       'features.emailRecovery': ['boolean', false, true],
       'features.windowsVerify': ['boolean', true, false],
+      'features.webauthn': ['boolean', true, false],
```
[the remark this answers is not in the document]
yumingcao-okta: This is the easiest way I could think of, this flag tells a `webauthn` factorType represents a  webauthn factor instead of a Windows Hello factor.

src/EnrollWebauthnController.js
```diff
@@ -0,0 +1,186 @@
...
+              })
+              .fail(function (error) {
+                throw {
+                  xhr: {responseJSON: {errorSummary: error.message}}
+                };
+              });
+          } else {
+            // verify via legacy u2f js for non webauthn supported browser
+            var registerRequests = [{
+              version: FidoUtil.getU2fVersion(),
```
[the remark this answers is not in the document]
yumingcao-okta: It's a client side constant string instead of loading from API response since webauthn API does not provide `version` field.

src/util/FactorUtil.js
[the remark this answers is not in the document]
yumingcao-okta: yup, it's pending review

test/unit/helpers/xhr/MFA_CHALLENGE_webauthn.js
yumingcao-okta: Yeah, the upper level `challenge` is to be consistent with other factor API structure, the nested `challenge` is to be consistent with webauthn protocol (https://www.w3.org/TR/webauthn/#dictdef-publickeycredentialrequestoptions).

test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_webauthn.js
yumingcao-okta: it stands for `webauthn relying party`: https://www.w3.org/TR/webauthn/#webauthn-relying-party

test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_webauthn.js
yumingcao-okta: During enroll process, `appId` is only needed for users whose browser does not support webauthn API, where we fallback to legacy u2f js library. I put legacy u2f related variables under `u2fParams`.

test/unit/spec/EnrollWebauthn_spec.js
yumingcao-okta: See my comment https://github.com/okta/okta-signin-widget/pull/523#discussion_r227161299 for it's usage.

test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_webauthn.js
mauriciocastillosilva-okta: got it. Why didn't we explicitly named it `relyingParty`?

test/unit/spec/EnrollWebauthn_spec.js
yumingcao-okta: Webauthn factor is reusing U2F images and texts for beacon.

test/unit/spec/VerifyWebauthn_spec.js
yumingcao-okta: Webauthn factor is reusing U2F images and texts for beacon.

test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_webauthn.js
yumingcao-okta: This id value is generated from my local test environment.

test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_webauthn.js
yumingcao-okta: This is to be consistent with webauthn protocol, so that we can feed the entire `activation` object to browser's `navigator.credential.create` without making any changes.

src/util/FidoUtil.js
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+/* eslint complexity:[2, 9] */
+
+define(function () {
+  var fn = {};
+
+  fn.getU2fEnrollErrorMessageKeyByCode = function (errorCode) {
```
mauriciocastillosilva-okta: I think it will be better if we change these 2 functions for U2F error messages to return the actual message instead of the key, since everywhere these functions are used is just to feed the key to an `Okta.loc` fn
```
fn.getU2fEnrollErrorMessage = function (errorCode) {
  ...
  return Okta.loc(key, 'login');
}

File: src/util/FidoUtil.js [hunk 29]
```diff
@@ -0,0 +1,56 @@
...
+      case 3:
+        return 'u2f.error.badRequest';
+      case 4:
+        return 'u2f.error.unsupported';
+      case 5:
+        return 'u2f.error.timeout';
+    }
+  };
+
+  fn.getU2fVerifyErrorMessageKeyByCode = function (errorCode, isOneFactor) {
```
mauriciocastillosilva-okta: and in the same line of my other comment, this can be
```
fn.getU2fVerifyErrorMessage = function (errorCode, isOneFactor) {
  ...
  return Okta.loc(key, 'login');
}
```

src/models/Settings.js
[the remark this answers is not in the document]
mauriciocastillosilva-okta: ok. @rchild-okta just double checking with you name of this new config is good.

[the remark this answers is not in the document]
rchild-okta: Yep, seems reasonable to me.

PR #526, opened 2018-08-28
Okta 186322 Remove the call to processCreds at IdP discovery form submit

haishengwu-okta:
`processCreds` hook is intend to record Username/Password. 
The use case from monolith is for ChromeBook [OKTA-69142](https://oktainc.atlassian.net/browse/OKTA-69142).
However, the IdP Discovery (identity first flow) only has username but not password and doesn't need to invoke the `processCreds` function.

cc @federations-okta

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -36,6 +36,7 @@ module.exports = function (grunt) {
         'test/unit/helpers/**/*.js',
         'test/**/**/*.js',
         '!test/e2e/react-app/**/*.js',
+        '!test/e2e/angular-app/**/*.js',
         '!test/unit/helpers/xhr/*.js',
         '!test/unit/vendor/*.js',
         'webpack.*.js'
--- a/src/views/idp-discovery/IDPDiscoveryForm.js
+++ b/src/views/idp-discovery/IDPDiscoveryForm.js
@@ -25,13 +25,7 @@ define([
     saveId: 'idp-discovery-submit',
 
     initialize: function () {
-      this.listenTo(this, 'save', function () {
-        var creds = {
-          username: this.model.get('username')
-        };
-        this.settings.processCreds(creds)
-        .then(_.bind(this.model.save, this.model));
-      });
+      this.listenTo(this, 'save', _.bind(this.model.save, this.model));
       this.stateEnableChange();
     },
 
--- a/test/unit/spec/IDPDiscovery_spec.js
+++ b/test/unit/spec/IDPDiscovery_spec.js
@@ -115,8 +115,10 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryF
   }
 
   function waitForWebfingerCall(test) {
-    //wait for the webfinger call
-    return Expect.waitForSpyCall(test.ac.webfinger, test);
+    return tick() // wait for the webfinger call cycle finish (promise -> then -> final)
+    .then(function () {
+      return Expect.waitForSpyCall(test.ac.webfinger, test);
+    });
   }
 
   function transformUsername(name) {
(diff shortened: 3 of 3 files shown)


test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -924,14 +926,11 @@ function (_, $, Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryF
...
-          expect(processCredsSpy.calls.count()).toBe(1);
-          expect(processCredsSpy).toHaveBeenCalledWith({
-            username: 'testuser@clouditude.net'
-          });
+          expect(processCredsSpy.calls.count()).toBe(0);
           expect(test.ac.webfinger).toHaveBeenCalled();
         });
       });
-      itp('calls async processCreds function before saving a model', function () {
+      itp('does not call async processCreds function before saving a model', function () {
```
santhoshbalakrishnan: Don't think we need two tests here. We only worry about whether it's called or not.

src/views/idp-discovery/IDPDiscoveryForm.js
```diff
@@ -25,13 +25,7 @@ define([
     saveId: 'idp-discovery-submit',
 
     initialize: function () {
-      this.listenTo(this, 'save', function () {
-        var creds = {
-          username: this.model.get('username')
-        };
-        this.settings.processCreds(creds)
```
santhoshbalakrishnan: I thought that the point of this was that the widget consumer can modify the username before it gets saved. Is that not the case anymore? When does `processCreds` gets called then?

haishengwu-okta: This is where monolith provides the hook.
https://github.com/okta/okta-core/blob/master/clients/loginpage/src/main-loginpage.js#L127-L145

I think the username/password are pass as value (read-only) to `processCreds` hook hence it's supposed not to be modified.

santhoshbalakrishnan: Even if it's read only, where would be the place customers can hook on to if they host the widget themselves? Since this is idp discovery, isn't a call back to processCreds expected with just the username?

haishengwu-okta: @santhoshbalakrishnan 
Good question. AFAIK, there are no use cases today that customers would run some callback logic only on username at IdP discovery login flow.
ping @yuliu-okta @petermiller-okta to confirm.

haishengwu-okta: `processCreds` still being used at Primary Auth flow.

Gruntfile.js
```diff
@@ -36,6 +36,7 @@ module.exports = function (grunt) {
         'test/unit/helpers/**/*.js',
         'test/**/**/*.js',
         '!test/e2e/react-app/**/*.js',
+        '!test/e2e/angular-app/**/*.js',
```
jmelberg-okta: Thank you for this!

src/views/idp-discovery/IDPDiscoveryForm.js
yuliu-okta: According to the widget doc, `processCreds` is to handle credentials and should be only called in Primary Auth, Password Expiration, and Password Reset flows. If the identifier first flow needs a callback in the future, we might need to provide a new callback instead of `processCreds` since it is not credentials.

mauriciocastillosilva-okta: I think this makes sense. If Okta is the IdP, then we will go to PrimaryAuth and there we will be able to send username/password to the callback, which will do the logic for chromeOs. 
If Okta is not the IdP, then I don't know how this works, but not sure that is supported

A question I have is when we have passwordless, then I don't think we can support chromeOS login the way it is right now...

haishengwu-okta: @mauriciocastillosilva-okta 
yeah, I believe it won't work when we have passwordless. who's the PMs owning passwordless? we might want to give heads up in case they're not aware yet.

test/unit/spec/IDPDiscovery_spec.js
mauriciocastillosilva-okta: +1

src/views/idp-discovery/IDPDiscoveryForm.js
mauriciocastillosilva-okta: I will let him know, since I've been working in that feature.

PR #529, opened 2018-08-30
Change the IDP discovery redirect to other IDP from GET to POST request.

haishengwu-okta:
In IDP Discovery flow, the endpoint to next IdP could be very long (typically a inbound SAML request).
Therefore, instead of doing GET (browser redirect) which has length limit, we changed to POST.

for OKTA-182115

- @mauriciocastillosilva-okta @santhoshbalakrishnan 
- @yuliu-okta  @federations-okta

### DIFF
--- a/src/IDPDiscoveryController.js
+++ b/src/IDPDiscoveryController.js
@@ -11,16 +11,15 @@
  */
 
 define([
-  'okta',
   'PrimaryAuthController',
   'models/PrimaryAuth',
   'views/idp-discovery/IDPDiscoveryForm',
   'models/IDPDiscovery',
-  'views/shared/Footer',
-  'util/BaseLoginController'
+  'util/BaseLoginController',
+  'util/Util'
 ],
-function (Okta, PrimaryAuthController, PrimaryAuthModel, IDPDiscoveryForm, IDPDiscoveryModel,
-          Footer, BaseLoginController) {
+function (PrimaryAuthController, PrimaryAuthModel, IDPDiscoveryForm, IDPDiscoveryModel,
+          BaseLoginController, Util) {
 
   return PrimaryAuthController.extend({
     className: 'idp-discovery',
@@ -65,6 +64,10 @@ function (Okta, PrimaryAuthController, PrimaryAuthModel, IDPDiscoveryForm, IDPDi
           this.options.appState.set('disableUsername', true);
           this.options.appState.trigger('navigate', 'signin');
         }
+      });
+
+      this.listenTo(this.model, 'goToOtherIdpAuth', function (url) {
+        Util.postToUrl(url, this.$el);
       });
     }
 
--- a/src/IDPDiscoveryController.js
+++ b/src/IDPDiscoveryController.js
@@ -67,7 +67,7 @@ function (PrimaryAuthController, PrimaryAuthModel, IDPDiscoveryForm, IDPDiscover
       });
 
       this.listenTo(this.model, 'goToOtherIdpAuth', function (url) {
-        Util.postToUrl(url, this.$el);
+        Util.redirectToUrlWithPost(url, this.$el);
       });
     }
 
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -15,7 +15,14 @@ define(['okta', 'util/Logger'], function (Okta, Logger) {
 
   var Util = {};
   var _ = Okta._;
-
+  var $ = Okta.$;
+  var hiddenPostFormTpl = Okta.tpl(
+    '<form method="POST" action="{{action}}" style="display:none;">' +
+    '{{#each inputs}}'+
+    '<input type="hidden" name="{{name}}" value="{{value}}">' +
+    '{{/each}}'+
+    '</form>');
+  
   Util.hasTokensInHash = function (hash) {
     return /((id|access)_token=)/i.test(hash);
   };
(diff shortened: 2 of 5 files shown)


src/util/Util.js
```diff
@@ -15,7 +15,14 @@ define(['okta', 'util/Logger'], function (Okta, Logger) {
 
   var Util = {};
   var _ = Okta._;
-
+  var $ = Okta.$;
+  var hiddenPostFormTpl = Okta.tpl(
+    '<form method="POST" action="{{action}}" style="display:none;">' +
+    '{{#each inputs}}'+
```
santhoshbalakrishnan: nit: formatting.

src/util/Util.js
```diff
@@ -101,6 +108,34 @@ define(['okta', 'util/Logger'], function (Okta, Logger) {
...
+  // what it does is actually create a hidden form
+  // and fill values from the url (base url, query parameters)
+  // and submit the form.
+  Util.postToUrl = function (url, $el) {
+    var parts = url.split('?'),
+        baseUrl = parts[0],
+        queryString = parts[1],
+        queryParams,
+        form;
+    var formData = {
```
santhoshbalakrishnan: Why not just add it to previous `var` declaration?

src/IDPDiscoveryController.js
```diff
@@ -66,6 +65,10 @@ function (Okta, PrimaryAuthController, PrimaryAuthModel, IDPDiscoveryForm, IDPDi
           this.options.appState.trigger('navigate', 'signin');
         }
       });
+
+      this.listenTo(this.model, 'goToOtherIdpAuth', function (url) {
+        Util.postToUrl(url, this.$el);
```
mauriciocastillosilva-okta: just thinking... could we append the form to the `body` of the page so we avoid passing the extra `$el` param and maybe even the extra event `goToOtherIdpAuth`?

src/util/Util.js
```diff
@@ -101,6 +108,34 @@ define(['okta', 'util/Logger'], function (Okta, Logger) {
     Logger.warn(`\n${message.replace(/^(\s)+/gm, '')}`);
   };
 
+  // Heleper function to submit a url via POST request.
+  // what it does is actually create a hidden form
+  // and fill values from the url (base url, query parameters)
+  // and submit the form.
+  Util.postToUrl = function (url, $el) {
```
santhoshbalakrishnan: Can we rename this to be more specific: something like `createPostFormAndSubmit` ?

src/IDPDiscoveryController.js
haishengwu-okta: good question and it's what I thought initially then I'm slightly leaning towards to not pollute the `body` especially when customer host the widget.
although it actually doesn't quite matter given we're leaving the page anyway after submit..

test/unit/spec/Util_spec.js
```diff
@@ -133,6 +135,41 @@ define(['util/Util', 'util/Logger'], function (Util, Logger) {
       });
     });
 
+    describe('postToUrl', function () {
+      beforeEach(function () {
+        spyOn($.fn, 'submit');
+      });
+
+      it('post the from only when no query parameters', function () {
```
santhoshbalakrishnan: Why `only when`?

src/util/Util.js
haishengwu-okta: hmm, why would you prefer more specific naming that indicates a little more implementation detail?
I thought `postToUrl` explain what it does?

src/IDPDiscoveryController.js
mauriciocastillosilva-okta: yeah... at the end it doesn't matter... so I guess either way works, I was just thinking a way to avoid the event and then handling it here as opposed to do everything on the view... but I agree with the point of not polluting the page's body. So sounds good, we can keep it as you have it

src/util/Util.js
mauriciocastillosilva-okta: Maybe `redirectWithForm` or `redirectToUrlWithPost`? I like the simplicity of `postToUrl`, but to Santhosh's point I think `postToUrl` is probably to general.... or maybe keep `postToUrl` but not have this method in a Util file since it is very specific for IdpDiscovery?

test/unit/spec/Util_spec.js
```diff
@@ -133,6 +135,41 @@ define(['util/Util', 'util/Logger'], function (Util, Logger) {
       });
     });
 
+    describe('postToUrl', function () {
+      beforeEach(function () {
+        spyOn($.fn, 'submit');
+      });
+
+      it('post the from when no query parameters', function () {
```
mauriciocastillosilva-okta: typo `post the form` not the `from`, and maybe `post the form without query parameters`

test/unit/spec/Util_spec.js
```diff
@@ -133,6 +135,41 @@ define(['util/Util', 'util/Logger'], function (Util, Logger) {
...
+        expect(container.children().size()).toBe(1);
+
+        var form = $(container.children()[0]);
+        expect(form[0].outerHTML).toBe(
+          '<form method="POST" action="http://test.abc.com/" style="display:none;"></form>'
+        );
+        expect(form.submit.calls.count()).toBe(1);
+      });
+
+      it('post the from when has query parameters', function () {
```
mauriciocastillosilva-okta: same typo here `form`... maybe this also could be `post the form with query parameters`

src/util/Util.js
santhoshbalakrishnan: ^ exactly my thoughts.

haishengwu-okta: vote for `redirectToUrlWithPost`

src/util/Util.js
```diff
@@ -101,6 +108,33 @@ define(['okta', 'util/Logger'], function (Okta, Logger) {
     Logger.warn(`\n${message.replace(/^(\s)+/gm, '')}`);
   };
 
+  // Heleper function to submit a url via POST request.
```
larsjohansen-okta: `Heleper` -> `Helper`

haishengwu-okta:
close as planned change. find detail in the JIRA

PR #534, opened 2018-09-06
Fix flaky OIDC e2e test.

haishengwu-okta:
- wait for bookmark menu before clicking

### DIFF
--- a/test/e2e/page-objects/FacebookPage.js
+++ b/test/e2e/page-objects/FacebookPage.js
@@ -22,8 +22,10 @@ class FacebookPage {
   }
 
   logout() {
+    const bookmarkElemId = '#bookmarks_jewel';
     browser.get('https://m.facebook.com/');
-    $('#bookmarks_jewel').click();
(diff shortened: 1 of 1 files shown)


test/e2e/page-objects/FacebookPage.js
```diff
@@ -22,8 +22,10 @@ class FacebookPage {
   }
 
   logout() {
+    const bookmarkElemId = '#bookmarks_jewel';
```
jmelberg-okta: nit: Since we're wrapping the value of `bookmarkElemId` into a jquery selector, would it make more sense to do that up front? Essentially:

```javascript
const bookmarkElem = $('#bookmarks_jewel');
...
browser.wait(EC.presenceOf(bookmarkElem), 5000);
bookmarkElem.click();
```

haishengwu-okta: my impression when using selenium Java is select element early on likely to get staleness exception. might not be the case in protractor and would give a try when I touch this code again. ;)

PR #285, opened 2017-08-22
:art: Move to use Yarn in okta-signin-widget

manueltanzi-okta:
:art: Move to use Yarn in okta-signin-widget

- Add yarn.lock.
- Delete package-lock.json.
- Update Readme to point out that we use yarn as our node dependency manager.
- Update ci scripts to use yarn as well.

Resolves: [OKTA-137898](https://oktainc.atlassian.net/browse/OKTA-137898)

@rchild-okta @ukilon-okta @rajaguruduraisamy-okta @kelvinzhu-okta

PR #540, opened 2018-09-14
Generates a courage bundle for signin widget

haishengwu-okta:
## Description

1. consolidate dependency to courage by using functions from `Okta` namespace
2. add a new packages `@okta/courage-for-signin-widget`
  - has a webpackg scripts to generate a bundled file
  - thereafter, the upgrade process would be 1) upgrade courage package 2) run the build script
3. remove source code package at `@okta/courage`

### Before change

```
-rw-r--r--   1 haisheng.wu  staff   981622 Sep 12 21:42 okta-sign-in-no-jquery.js
-rw-r--r--   1 haisheng.wu  staff  6408042 Sep 12 21:42 okta-sign-in-no-jquery.js.map
-rw-r--r--   1 haisheng.wu  staff  1749293 Sep 12 21:42 okta-sign-in.entry.js
-rw-r--r--   1 haisheng.wu  staff  2585882 Sep 12 21:42 okta-sign-in.entry.js.map
-rw-r--r--   1 haisheng.wu  staff  2847909 Sep 12 21:42 okta-sign-in.js
-rw-r--r--   1 haisheng.wu  staff  3931307 Sep 12 21:42 okta-sign-in.js.map
-rw-r--r--   1 haisheng.wu  staff  1079945 Sep 12 21:42 okta-sign-in.min.js
-rw-r--r--   1 haisheng.wu  staff  7232125 Sep 12 21:42 okta-sign-in.min.js.map
```

### After change

increased a little

```
drwxr-xr-x  10 haisheng.wu  staff      340 Sep 13 22:41 .
drwxr-xr-x   8 haisheng.wu  staff      272 Sep 13 22:41 ..
-rw-r--r--   1 haisheng.wu  staff  1077746 Sep 13 22:41 okta-sign-in-no-jquery.js
-rw-r--r--   1 haisheng.wu  staff  6990336 Sep 13 22:41 okta-sign-in-no-jquery.js.map
-rw-r--r--   1 haisheng.wu  staff  1780243 Sep 13 22:41 okta-sign-in.entry.js
-rw-r--r--   1 haisheng.wu  staff  2668434 Sep 13 22:41 okta-sign-in.entry.js.map
-rw-r--r--   1 haisheng.wu  staff  3038607 Sep 13 22:41 okta-sign-in.js
-rw-r--r--   1 haisheng.wu  staff  4200455 Sep 13 22:41 okta-sign-in.js.map
-rw-r--r--   1 haisheng.wu  staff  1176131 Sep 13 22:41 okta-sign-in.min.js
-rw-r--r--   1 haisheng.wu  staff  7815611 Sep 13 22:41 okta-sign-in.min.js.map
```


packages/@okta/courage-for-signin-widget/README.md
```diff
@@ -0,0 +1,8 @@
+generates a Courage version for signin-widget
+
+## Steps
+
+0. make sure VPN is connected
+1. upgrade courage version. e.g. `yarn upgrade @okta/courage@4.4.0-beta.xxxxxx`
+2. `yarn build`
+3. `dit/okta-for-signin-widget.js` may be changed depending on the change set since last upgrade.
```
nikhilvenkatraman-okta: nit: `dist`?

test/unit/main.js
```diff
@@ -1,7 +1,8 @@
-require('okta/underscore');
-require('okta/handlebars');
+// require('okta/underscore');
+// require('okta/handlebars');
```
mauriciocastillosilva-okta: we can remove this comments I think

src/util/CountryUtil.js
```diff
@@ -11,12 +11,12 @@
  */
 
 define([
-  'okta/underscore',
-  'shared/util/Bundles',
+  'okta',
+  './Bundles',
```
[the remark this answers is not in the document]
haishengwu-okta: I don't think so because prior this change, `shared/util/Bundles` is alias to widget's `util/Bundle` hence no functional changes.

src/widget/OktaSignIn.js
```diff
@@ -186,9 +187,9 @@ var OktaSignIn = (function () {
 
     // Modify the underscore, handlebars, and jquery modules
     // Remove once these are explicitly required in Courage
-    require('okta/underscore');
-    require('okta/handlebars');
-    require('okta/jquery');
+    // require('okta/underscore');
+    // require('okta/handlebars');
+    // require('okta/jquery');
```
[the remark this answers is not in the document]
haishengwu-okta: yeah, they are removed at 2nd commit.

test/unit/main.js
haishengwu-okta: ditto.

packages/@okta/courage-for-signin-widget.js
```diff
@@ -0,0 +1,11371 @@
+(function webpackUniversalModuleDefinition(root, factory) {
```
nikhilvenkatraman-okta: Similar to what @mauriciocastillosilva-okta asked on slack. Should this be done within courage as opposed to doing it here in the widget? Feels like the consumer should not be  responsible for this . Additionally I am not sure if the current work of moving courage to okta-ui has any affect on this.

haishengwu-okta: why would think consumer should not be responsible for this?
yeah, it could be done in courage and I'm slightly leaning toward in signin widget as the first step. (didn't add complexity to Courage plus improve the signin widget a little bit)
we could move to courage or another okta-ui package after courage has been moved to okta-ui. 
- probably another office hour topic?

nikhilvenkatraman-okta: This is great first step and maybe worth another office hour discussion. Only reason I feel this could belong in courage is
- Since this is in the open-source sign in widget repo and the courage framework is only internal to okta, all this bundling logic only makes sense to okta.
- Comparing it to how we have option to bundle  courage with `okta.js` and `OktaWithInputs ` or in the sign-in-widget we have  `cdnConfig` and `noJqueryConfig `, feels like this could be another option where we ship courage for sign in widget as another option.

> we could move to courage or another okta-ui package after courage has been moved to okta-ui.`. 

Yeah, if we do eventually want this to live in  okta-ui as a package or under courage as a package? hopefully this would make it easier to move later.

packages/@okta/courage-for-signin-widget/README.md
```diff
@@ -0,0 +1,8 @@
+generates a Courage version for signin-widget
```
jmelberg-okta: Lets make sure READMEs have a header:

```
# @okta/courage-for-signin-widget

Generates a custom version of Courage for the Sign-In Widget

## Steps
...
```

packages/@okta/courage-for-signin-widget/README.md
```diff
@@ -0,0 +1,8 @@
...
+
+## Steps
+
+0. make sure VPN is connected
+1. upgrade courage version. e.g. `yarn upgrade @okta/courage@4.4.0-beta.xxxxxx`
+2. `yarn build`
```
jmelberg-okta: Do we want to mix `yarn` and `npm`? At the repository root, we use `npm` for all build/start instructions.

As a future task, lets update all Sign-In Widget docs to use `yarn` instead of `npm`.

packages/@okta/courage-for-signin-widget/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "courage-for-signin-widget",
+  "version": "1.0.0",
+  "description": "generates a Courage version for signin-widget",
+  "main": "src/OktaForSigninWidget.js",
```
jmelberg-okta: Okta for Courage for the widget?

`CourageForSigninWidget` might be more clear.

packages/@okta/courage-for-signin-widget/src/OktaForSigninWidget.js
```diff
@@ -0,0 +1,131 @@
...
+
+    Okta.registerInput('text', TextBox);
+    Okta.registerInput('password', PasswordBox);
+    Okta.registerInput('checkbox', CheckBox);
+    Okta.registerInput('radio', Radio);
+    Okta.registerInput('select', Select);
+
+    return Okta;
+  });
+
```
jmelberg-okta: nit: Extra newline

packages/@okta/courage-for-signin-widget/webpack.config.js
```diff
@@ -0,0 +1,86 @@
+/* global module __dirname */
+const path = require('path');
+const EMPTY = 'src/empty';
+const SHARED_JS = path.resolve(__dirname, 'node_modules/@okta/courage/src');
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
```
jmelberg-okta: Can this be a destructured import? 

```javascript
const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
```

src/util/Bundles.js
```diff
@@ -11,19 +11,21 @@
...
   'json!nls/login',
   'json!nls/country',
   'util/Logger',
   'json!config/config',
   'util/BrowserFeatures'
-], function (Okta, Q, login, country, Logger, config, BrowserFeatures) {
+], function ($, _, Handlebars, Q, login, country, Logger, config, BrowserFeatures) {
 
-  var { _, $ } = Okta;
+  // var { _, $ } = Okta;
```
jmelberg-okta: Can you help me understand why in other files we use underscore and jquery from `okta`, but here we are explicitly using the dependencies?

haishengwu-okta: the `Okta` object is empty at generating `signin-widget` bundle. I think it's because of cycle dependency.
- `courage-for-widget` bundle declares that `shared/util/Bundle` as external
- when integrate with signin widget (webpack.common.config.js), it says `shared/util/Bundle` is alias to `util/Bundle`, which needs to load `okta` (`courage-for-widget.js`), which then asks for alias `util/Bundle` again.

packages/@okta/courage-for-signin-widget.js
jmelberg-okta: Just my two cents -

The Widget is a consumer of Courage, so it should be responsible for any/all bundling logic around it. Just as any other library uses a component package, it too is responsible to filtering out what it needs and/or doesn't need.

If we move this logic within Courage, it **does** simplify the build steps + dependencies for the Widget. It could simply import `@okta/courage@4.4.0-beta.XYZ/dist/CourageForSigninWidget.js`, and be done.

That being said, I'm +1 for having an office hours discussion around this. Further, I'm curious what versioning strategy for "internal packages" can be. With both implementations, we're forced to manually update the courage version, while a flexible range might be a viable feature.

packages/@okta/courage-for-signin-widget/README.md
haishengwu-okta: I guess I'm used to use `yarn` (especially my intention for this package is for internal).
I could change it to `npm` or if we're comfortable to use `yarn` at the repo root, I'll create a follow up task to consolidate all npm usage to yarn plus the package-lock.json to yarn.lock.

jmelberg-okta: @manueltanzi-okta has #285  already for this, lets get him to close it out :)

packages/@okta/courage-for-signin-widget.js
```diff
@@ -0,0 +1,11372 @@
+/*! DO NOT MODIFY THIS FILE MANUALLY. IT IS GENERATED FROM PACKAGE @OKTA/COURAGE-FOR-SIGNIN-WIDGET. */
```
jmelberg-okta: Not a blocker, but we should consider removing comments and uglifying this file. We'd reduce this file down to about `140000` bytes, an improvement by nearly 63%.

packages/@okta/courage-for-signin-widget/README.md
```diff
@@ -0,0 +1,11 @@
+# @okta/courage-for-signin-widget
+
+Generates a custom version of Courage for the Sign-In Widget. 
+It's intend for internal developers and please open a ticket if any questions.
+
+## Steps
+
+0. make sure VPN is connected
```
jmelberg-okta: nit: Capitalize the beginning of each ordered list item

mauriciocastillosilva-okta: +1

packages/@okta/courage-for-signin-widget/README.md
```diff
@@ -0,0 +1,11 @@
...
+1. upgrade courage version. e.g. `yarn upgrade @okta/courage@4.4.0-beta.xxxxxx`
```
mauriciocastillosilva-okta: let's change this to use release versions. It is my understanding we should try to use release versions of i18n and courage in the SignIn Widget, and use beta versions internally only
```
yarn upgrade @okta/courage@4.4.0.xxxxx
```

packages/@okta/courage-for-signin-widget/README.md
```diff
@@ -0,0 +1,11 @@
...
+2. `yarn build`
+3. check file `../okta-for-signin-widget.js` which may be changed depending on the change set since last upgrade
```
mauriciocastillosilva-okta: A couple of things in this line
- name of the file should be updated to `courage-for-signin-widget.js`
- we can re-word this to a little more explicit instructions like
```
commit file ..okta/courage-for-signin-widget.js if it has changes. This file will have changes only if the relevant components of courage that we consume in the Sign-In Widget have changed.
```

packages/@okta/courage-for-signin-widget/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "courage-for-signin-widget",
+  "version": "1.0.0",
+  "description": "generates a Courage version for signin-widget",
+  "main": "src/CourageForSigninWidget.js",
+  "scripts": {
+    "build": "webpack"
+  },
+  "author": "",
+  "license": "MIT",
```
mauriciocastillosilva-okta: we use `Apache 2` in our sign in widget (https://github.com/okta/okta-signin-widget/blob/master/package.json#L6)
Should this be apache too? I'm not 100% sure, maybe @jmelberg-okta knows...

packages/@okta/courage-for-signin-widget/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "dependencies": {
+    "@okta/courage": "4.4.0-beta.2818bde",
```
mauriciocastillosilva-okta: following my comment above, we should try use release versions here instead of beta (beta can be use internally). Or we can talk about it and change it as well. Technically it doesn't matter much, but we decided that long time ago just to keep our open source repo using release versions

packages/@okta/courage-for-signin-widget/README.md
haishengwu-okta: (replied below)

packages/@okta/courage-for-signin-widget/package.json
jmelberg-okta: IMO lets just keep it simple - We can remove the `author` and `license` fields

packages/@okta/courage-for-signin-widget/package.json
haishengwu-okta: good point and I'm not aware that we like our open source repo to use release version.
especially for courage, turns out we didn't make very clear when / what to be release.
will make change to use release version and like to chat more about it.

haishengwu-okta: @mauriciocastillosilva-okta yeah, added to uicore office hour topic.

packages/@okta/courage-for-signin-widget.js
nbarbettini: I also think this logic to build a custom version of Courage for Widget consumption should live in the Widget repo. The folks maintaining the Widget will have a better idea of what the Widget requires from Courage than the folks maintaining Courage. My $0.02 🙂

webpack.common.config.js
```diff
@@ -1,14 +1,13 @@
...
 var path      = require('path');
-var EMPTY     = 'widget/empty';
 var TARGET_JS = path.resolve(__dirname, 'target/js/');
-var SHARED_JS = TARGET_JS + '/shared';
+var LOCAL_PACKAGES = path.resolve(__dirname, 'packages/');
 
 // Return a function so that all consumers get a new copy of the config
 module.exports = function (outputFilename) {
   return {
-    entry: ['./src/util/RegisterInputs.js', './target/js/widget/OktaSignIn.js'],
```
mauriciocastillosilva-okta: do we still need https://github.com/okta/okta-signin-widget/blob/master/src/util/RegisterInputs.js ? Since you are removing it here and I see that packages/@okta/courage-for-signin-widget/src/CourageForSigninWidget.js already does the same, I thought it was worth asking

Gruntfile.js
```diff
@@ -74,10 +74,6 @@ module.exports = function (grunt) {
           // Source Files
           {expand: true, cwd: 'src/', src: ['**'], dest: JS + '/'},
 
-          // Courage files
-          {expand: true, cwd: 'packages/@okta/courage/src/', src: ['**'], dest: JS + '/shared/'},
-          {expand: true, cwd: 'packages/@okta/courage/src/vendor', src: ['**'], dest: JS + '/vendor/'},
```
mauriciocastillosilva-okta: do we need to copy the new `courage-for-signin-widget.js` to `target/js` folder?

webpack.common.config.js
haishengwu-okta: ah, good point. I don't think we need it.

Gruntfile.js
haishengwu-okta: hmm, either way works.
I thought it's unnecessary to copy but rather refer to the file directly.

PR #547, opened 2018-09-20
Adds getTransaction method

### LINKED ISSUE #490: Transactional information about the target from the underlying client
### I'm submitting a
  - [ ] bug report
  - [X] feature request

### Background info

Expose the auth-js client or the transactional information so I can get information about the state of the widget.

When the widget is loaded in Okta Hosted, or from a token (state, activation, otherwise) it would be great to be able to get the information out of the transaction:

`transaction.target.clientId`

For example this is how you do it with auth-js:

```
var authJs = new OktaAuth({url: orgUrl});
authJs.tx.resume({stateToken: '00oQMxcSBiopj6YLMTCVlhR'})
  .then(function(transaction) {
    console.log('current client:', transaction.target.clientId, transaction);
  }).fail(function(error){
    console.log('err', error)
  });
```

jmelberg-okta:
### Description

Provides developers the ability to see authentication transaction information given a `stateToken`.

#### Resolves

#490

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Contributors should read our [contributing guidelines](./CONTRIBUTING.md) if the
   * [remove](#remove)
   * [on](#onevent-callback-context)
   * [off](#offevent-callback)
+  * [getTransaction](#gettransactionstatetoken)
   * [session.get](#sessiongetcallback)
   * [session.refresh](#sessionrefreshcallback)
   * [session.close](#sessionclosecallback)
@@ -362,6 +363,27 @@ signIn.off('pageRendered');
 // Unsubscribe the onPageRendered listener from the 'pageRendered' event
 signIn.off('pageRendered', onPageRendered);
 ```
+
+## getTransaction(stateToken)
+
+Returns [authentication transaction information](https://developer.okta.com/docs/api/resources/authn#transaction-model) given a `stateToken`. If a `stateToken` is not provided, it will fetch the current token from a saved browser cookie.
+
+```javascript
+signIn.getTransaction(stateToken)
+.then(function (transaction) {
+  console.log(transaction.status);
+  console.log(transaction.user.profile.login);
+  // PASSWORD_EXPIRED
+  // user@example.com
+})
+.catch(function (err) {
+  console.log(error.message);
+});
+```
+
+When the `getTransaction` method resolves, it returns a **transaction** object that encapulates the [state authentication flow](https://developer.okta.com/docs/api/resources/authn#transaction-model). The **transaction** contains metadata about the current state, and provides methods that can be used to progress into another state.
+
+> For a list of examples, see the [`transaction`](https://github.com/okta/okta-auth-js/#locked_out) documentation.
 
 ## session.get(callback)
 
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ signIn.off('pageRendered', onPageRendered);
 
 ## getTransaction(stateToken)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -363,6 +364,27 @@ signIn.off('pageRendered');
 signIn.off('pageRendered', onPageRendered);
 ```
 
+## getTransaction(stateToken)
+
+Returns [authentication transaction information](https://developer.okta.com/docs/api/resources/authn#transaction-model) given a `stateToken`. If a `stateToken` is not provided, it will fetch the current token from a saved browser cookie.
```
manueltanzi-okta: Maybe I remember wrong, but from the conversations didn't you guys decided to go for 2 different functions? one that use the stateToken as input and one that use the one from the cookie?

jmelberg-okta: We decided that having two methods didn’t make sense, as relying on the browser cookie isn’t a feasible use case at this stage. However, the `resume` method queries for the token if one is not provided, so we made sure to document that here.

Further, the only viable use case where you’d want to rely on the stateToken cookie is after primary authentication (in the middle of a Widget view). We are going to encourage devs to listen to change events if they need it mid-auth.

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -89,5 +96,35 @@ function (Widget, Expect, Logger) {
...
+      it('resolves to an error if a transaction cannot resume', function (done) {
+        signIn.getTransaction()
+        .then(function(){/* Should never reach this */})
+        .catch(function(err) {
+          expect(err.name).toBe('AuthSdkError');
+          expect(err.errorSummary).toBe('No transaction to resume');
+          done();
+        });
+      });
+      it('calls the authentication api with a stateToken', function (done) {
```
robertjd: It may be simpler to mock `authJs.resume` if possible, and just assert that what we pass to `getTransaction()` gets passed to the `authJS.resume()` transaction mock.  Similarly, with a `authJs.resume` mock, you could mock the return value and assert it makes it's way back through `getTransaction()`.

src/widget/OktaSignIn.js
```diff
@@ -163,6 +163,9 @@ var OktaSignIn = (function () {
      * @returns {Promise} - Returns a promise for an object containing the transaction information
      */
     function getTransaction(stateToken) {
+      if (!stateToken) {
+        throw 'A state token is required.';
```
jmelberg-okta: In the past, we've neglected language support for thrown error messages. Should we add a todo item for making sure these are localized?

src/widget/OktaSignIn.js
```diff
@@ -156,6 +156,19 @@ var OktaSignIn = (function () {
...
 
+    /**
+     * Returns authentication transaction information given a stateToken.
+     * @param {String} stateToken - Ephemeral token that represents the current state of an authentication
+     *                              or recovery transaction
+     * @returns {Promise} - Returns a promise for an object containing the transaction information
+     */
+    function getTransaction(stateToken) {
+      if (!stateToken) {
+        throw 'A state token is required.';
```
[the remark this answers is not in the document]
jmelberg-okta: Good catch. Updated the `render` method to correctly throw as well.

README.md
```diff
@@ -363,6 +364,27 @@ signIn.off('pageRendered');
 signIn.off('pageRendered', onPageRendered);
 ```
 
+## getTransaction(stateToken)
+
+Returns [authentication transaction information](https://developer.okta.com/docs/api/resources/authn#transaction-model) given a `stateToken`.
```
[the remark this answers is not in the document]
jmelberg-okta: We could probably find a way to make this more clear, but we do implicitly state it is required:

> given a `stateToken`

As a future item, we should restructure our README docs to use types - which will also solve the "requirement" issue.

PR #548, opened 2018-09-20
Adds OIDC Factor

gowthamidommety-okta:
- Completed enroll for OIDC

OIDC Enroll:

oidcenroll

OIDC Verify:

oidcverify

Resolves: OKTA-186350


magizh-okta:
Please check if this has to be updated https://github.com/okta/okta-signin-widget/blob/975ceb8f8122fa25e2eca4cf74d6204fe0484c9e/test/unit/helpers/util/Expect.js#L240

gowthamidommety-okta:
Checked! We don't need to change it because it is the same view for both oidc and saml

PR #554, opened 2018-10-05
Upgrade grunt and some of plugins

haishengwu-okta:
## Description

- upgrade grunt and some plugins (low hanging fruits ;))
- no for jasmine, eslint which involve more changes and will come as later PR)
- upgrade grunt-sass and grunt-postcss cause NO changes to generated css.
- minor fix to readme
- removed `grunt-contrib-requirejs` and `grunt-contrib-jshint`


nikhilvenkatraman-okta:
@haishengwu-okta while you are here can you also update the read me to point to latest CDN version of widget `https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/2.13.0/js/okta-sign-in.min.js`

package.json
```diff
@@ -55,27 +55,26 @@
     "cssnano": "3.10.0",
     "fs-extra": "1.0.0",
     "git-rev-sync": "1.4.0",
-    "grunt": "0.4.5",
-    "grunt-cli": "0.1.13",
-    "grunt-contrib-connect": "0.11.2",
-    "grunt-contrib-copy": "0.8.0",
+    "grunt": "^1.0.3",
```
nikhilvenkatraman-okta: We are moving from specific versions to being more flexible by allowing the most recent major version now ? I guess its ok for grunt, but I think we are not consistent about this pattern.

haishengwu-okta: yeah, we're not quite consistent. I'm being more sloppy since we have the lock file.

package.json
```diff
@@ -55,27 +55,26 @@
...
-    "grunt-exec": "0.4.6",
-    "grunt-json-generator": "0.1.0",
-    "grunt-postcss": "0.8.0",
-    "grunt-retire": "0.3.12",
-    "grunt-sass": "2.0.0",
+    "grunt-exec": "^3.0.0",
+    "grunt-json-generator": "^0.1.0",
+    "grunt-postcss": "^0.9.0",
+    "grunt-retire": "^1.0.7",
+    "grunt-sass": "^3.0.1",
```
jmelberg-okta: Given the scope of fixing outdated  dependencies + `npm audit` warnings, I think this change makes a lot of sense.

One thing we should consider prioritizing is moving from grunt to webpack. We should be able to use the [`sass-loader`](https://github.com/webpack-contrib/sass-loader) instead.

package.json
```diff
@@ -55,27 +55,26 @@
...
     "git-rev-sync": "1.4.0",
-    "grunt": "0.4.5",
-    "grunt-cli": "0.1.13",
-    "grunt-contrib-connect": "0.11.2",
-    "grunt-contrib-copy": "0.8.0",
+    "grunt": "^1.0.3",
+    "grunt-cli": "^1.3.1",
+    "grunt-contrib-connect": "^2.0.0",
+    "grunt-contrib-copy": "^1.0.0",
     "grunt-contrib-jasmine": "1.1.0",
```
jmelberg-okta: I'm sure you discovered this, but upgrading to `grunt-contrib-jasmine` to `2.0.2` gives us a log of benefits:
- es6 support
- Chrome headless

Of course, this requires a bit of updating on our end (I took a quick look at it a few months back with #520) - but something we should explore soon!

haishengwu-okta: for sure! both (move to webpack and upgrade jasmine plus karma) are all in radar!
and I think upgrade grunt plugin would be helpful at paving the way.

PR #556, opened 2018-10-08
:bug: fixes the validation error disappear quickly

haishengwu-okta:
## Description

It's about timing.. 
Both inline validation and the 'send via email' button would trigger clear form error and validation.
Although inline validation comes a bit later (_.delay) and it supposes to work fine.
However the actual validation is throttled.

## Solution

There could be 2 possible solution

1. delay/throttle (200ms) the save event like what `BaseForm#save does`
2. disable the inline validation

I choose the 2nd since inline validation doesn't seem useful and don't want to introduce more timer.

## Screencasts

- https://okta.box.com/s/qx0jhm2xogoxnjgsmalrzv6588ur4y5a

### DIFF
--- a/src/ForgotPasswordController.js
+++ b/src/ForgotPasswordController.js
@@ -105,6 +105,7 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, Util, ContactSu
             name: 'username',
             input: TextBox,
             type: 'text',
+            inlineValidation: false,
             params: {
               innerTooltip: Okta.loc('password.forgot.email.or.username.tooltip', 'login'),
               icon: 'person-16-gray'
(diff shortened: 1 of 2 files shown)


haishengwu-okta:
@jmelberg-okta 
good catch! but we don't assert props if I understand correctly
Will explore other options.

haishengwu-okta:
@mauriciocastillosilva-okta 
thanks for looking and another option is to delay the validate on click, like what default save does
https://github.com/okta/okta-ui/blob/master/packages/courage/src/views/forms/BaseForm.js#L214

mauriciocastillosilva-okta:
i see... so 3 different ways to fix the issue, but I think you are solving the actual problem, which is that we are validating twice. 
thanks!

Issue #542, opened 2018-09-17
Token manager config


mohsenvafa:
### I'm submitting a
  - [x ] bug report

### Background info

In OktaSignIn's Token Manager config, when you set the Token Manager to **sessionStorage**, it doesn't write the token into session storage. 

```javascript
var signIn = new OktaSignIn(
  {
    baseUrl: 'https://domain.com',
    tokenManager: {
       storage: 'sessionStorage'
     }
  }
);
```

### Expected behavior

The token should be written into Session Storage. 

### What went wrong?
The token is still written in Local Storage

### Steps to reproduce

As mentioned above, the token manager config is not applied. 

### Your environment

  - Okta Sign-In Widget Version: @okta/okta-signin-widget = ^2.6.0
          @okta/okta-angular = 0.0.13
  - Browser: Chrome latest version
  - OS: Windows 10
  - Language: Type Script

niranjan-challa92:
Even I am facing the same issue with 
@okta/okta-signin-widget = 2.10.0, 
Angular = 5.2, 
okta-angular 1.0.1
Windows 7 environment

jmelberg-okta:
Hi @mohsenvafa & @niranjan-challa92 -

The Widget does not currently provide the ability to override the default `tokenManager` object, but this is certainly a useful addition. That being said, we do [accept PRs](https://github.com/okta/okta-signin-widget/blob/master/CONTRIBUTING.md)! 😄 

As a starting point, this can be added into [the `authClient` instantiation](https://github.com/okta/okta-signin-widget/blob/master/src/widget/OktaSignIn.js#L213-L221) with the [`tokenManager` config](https://github.com/okta/okta-auth-js/#the-tokenmanager) (if it exists).

I'm going to rename this issue accordingly and label it as a `Feature Request`.

mohsenvafa:
Hi @jmelberg-okta,

I created a pull request #560 to address this issue, but I see CI has failed. Am I missing something?

aarongranick-okta:
@mohsenvafa The Signin Widget accepts all configuration available to `okta-auth-js`, via the `authParams` config object. This object will be passed directly to `okta-auth-js`. See all available options for the token manager: https://github.com/okta/okta-auth-js#the-tokenmanager

Please let us know if your issue is resolved.

PR #560, opened 2018-10-11
feat: Add ability to override token manager configuration #542

mohsenvafa:
@jmelberg-okta, This PR is to address feature #542

### DIFF
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -173,6 +173,7 @@ var OktaSignIn = (function () {
       headers: {
         'X-Okta-User-Agent-Extended': 'okta-signin-widget-' + config.version
       },
+		  tokenManager: options.tokenManager,    
       clientId: options.clientId,
       redirectUri: options.redirectUri
     });


stueccles:
This would be a very welcome addition, one i need for instance :). The PR looks approved and passed a year ago but remains open. Any chance of getting it merged in?

src/widget/OktaSignIn.js
```diff
@@ -216,6 +216,7 @@ var OktaSignIn = (function () {
       headers: {
         'X-Okta-User-Agent-Extended': 'okta-signin-widget-' + config.version
       },
+      tokenManager: options.tokenManager,
```
aarongranick-okta: Is this PR out of date with master? The SIW currently accepts ALL options under `authParams` and passes them to `OktaAuth`. This includes the `TokenManager` section. https://github.com/okta/okta-signin-widget/blob/88c158ef70b1c2e2323264d5e2c136550320675f/src/widget/OktaSignIn.js#L148

stueccles: Yeah I figured that out afterwards by trawling the code and it works! Thanks. I would suggest putting this is in the docs rather than update this PR

PR #561, opened 2018-10-12
:seedling: Adds config option for relayState in PrimaryAuth

mauriciocastillosilva-okta:
Added a new config option (sendRelayStateInPrimaryAuth) that when true, will add the relayState to the primary auth request data

Resolves: OKTA-193359

Note: this is intentionally not documented in README since it is not even Beta yet


mauriciocastillosilva-okta:
@nikhilvenkatraman-okta this is mainly for risk mitigation: the change is for primaryAuth, and we don't want to change all primaryAuth requests. Not all of request need `relayState`... actually just a couple of flows need it (when we have custom factors)

mauriciocastillosilva-okta:
It's been 2 years since this PR, so closing it now since this work is probably no longer relevant

PR #567, opened 2018-10-22
feat: Adds new 'navigated' event

jmelberg-okta:
### Description

Renames `pageRendered` event to `navigated`, and returns the following `context` information:

```javascript
signIn.on('navigated', (context) => {
  // Assume view transitions between Primary Auth and MFA Push
  console.log(context);
  /**
   * {
   *   controller: 'mfa-verify'
   * }
   */
});
```

> Note: Since there is no *good* way to deprecate events, we're going to label the existing `pageRendered` event as deprecated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1326,7 +1326,7 @@ Triggered when the widget transitions to a new page and animations have finished
 
 - **view** - Current page path
 - **controller** - Current controller name
-- **transaction** - Current authentication [transaction](https://github.com/okta/okta-auth-js/#transactionstatus) (if available)
+- **transaction** - Current [authentication transaction](https://developer.okta.com/docs/api/resources/authn#transaction-model) (if available)
 
 ```javascript
 signIn.on('navigated', function (context) {
--- a/README.md
+++ b/README.md
@@ -1332,7 +1332,14 @@ Events published by the widget. Subscribe to these events using [on](#onevent-ca
       // {
       //   view: '/signin/verify/okta/push',
       //   controller: 'mfa-verify',
-      //   transaction: AuthTransaction
+      //   transaction: {
+      //     status: 'MFA_REQUIRED',
+      //     user: {
+      //       id: "00u..",
+      //       ...
+      //     }
+      //     ...
+      //   }
       // }
     });
     ```
@@ -1353,7 +1360,7 @@ Events published by the widget. Subscribe to these events using [on](#onevent-ca
       });
     });
     ```
-- **pageRendered** (*Deprecated*) - triggered when the widget transitions to a new page and animations have finished.
+- **pageRendered** (*Deprecated in favor of **navigated***) - triggered when the widget transitions to a new page and animations have finished.
 
     ```javascript
     signIn.on('pageRendered', function (data) {
--- a/README.md
+++ b/README.md
@@ -1325,6 +1325,7 @@ Events published by the widget. Subscribe to these events using [on](#onevent-ca
 Triggered when the widget transitions to a new page and animations have finished. Returns a `context` object containing the following properties:
 
 - **controller** - Current controller name
+- **stateToken** (*optional*) - An ephemeral token mapped to the current state of an authenticataion transaction.
 
 ```javascript
 // Overriding the "Back to Sign In" click action on the Forgot Password page
@@ -1337,6 +1338,43 @@ signIn.on('navigated', function (context) {
     e.preventDefault();
     e.stopPropagation();
     // Custom link behavior
+  });
+});
+```
+
+```javascript
+// Request authentication transaction information
+signIn.on('navigated', function (context) {
+  if (context.controller !== 'mfa-verify') {
+    return;
+  }
+  signIn.getTransaction(context.stateToken)
+  .then(function (transaction) {
+    console.log(transaction);
+    // {
+    //   status: 'MFA_REQUIRED',
+    //   user: {
+    //     id: '00ugti3kwafWJBRIY0g3',
+    //     profile: {
+    //       login: 'isaac@example.org',
+    //       ...
+    //     },
+    //   },
+    //   factors: [{
+    //     id: 'ufsigasO4dVUPM5O40g3',
+    //     provider: 'OKTA',
+    //     factorType: 'question',
+    //     profile: {
+    //       question: 'disliked_food',
+    //       questionText: 'What is the food you least liked as a child?'
+    //     },
+    //     verify: function(options) { /* returns another transaction */ }
+    //   }],
+    //   ...
+    // }
+  })
+  .catch(function (err) {
+    console.error(err);
   });
 });
 ```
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1316,12 +1320,26 @@ features: {
...
+- **navigated** - triggered when the widget transitions to a new page and animations have finished.
+
+    ```javascript
+    signIn.on('navigated', function (context) {
+      // Assume view transitions between Primary Auth and MFA Push
+      console.log(context);
+      // {
+      //   view: '/signin/verify/okta/push',
+      //   controller: 'mfa-verify',
+      //   transaction: AuthTransaction
```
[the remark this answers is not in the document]
jmelberg-okta: Good point. I"ll make this example a bit more clear.

src/util/BaseLoginController.js
```diff
@@ -101,6 +101,12 @@ define(['okta', 'q'], function (Okta, Q) {
     postRenderAnimation: function() {
       // Event triggered after a page is rendered along with the classname to identify the page
       this.trigger('pageRendered', {page: this.className});
+      var context = {
+        view: window.location.pathname,
+        controller: this.className,
+        transaction: this.options.appState.get('transaction')
+      };
+      this.trigger('navigated', _.pick(context, _.identity));
```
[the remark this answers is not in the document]
jmelberg-okta: This will remove the `transaction` key if it is `undefined`. Figured this was a cleaner solution to:

```javascript
var context = {
  view: window.location.pathname,
  controller: this.className
};

var transaction = this.options.appState.get('transaction');
if (transaction) {
  context.transaction = transaction;
}

this.trigger('navigated', context);
```

jmelberg-okta:
@haishengwu-okta - In regards to your question on introducing a new event:

The existing `pageRendered` event wasn't super clear for developers. Renaming this to `navigated` implies that when an end-user transitions from one view to the next, a `navigated` event will be fired. This will align closer to the new event convention Nate is proposing (e.g. `ready`).

src/util/BaseLoginController.js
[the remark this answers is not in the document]
haishengwu-okta: hmm, why would remove `transaction` when it's `undefined`?
It seems doesn't hurt and by keeping it's a little easier to express the type of `context`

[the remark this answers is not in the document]
jmelberg-okta: Mostly a style thing - I don't have a set opinion on this, but do like that I will not need to account for this undefined values in my tests :)

Per your suggestion, should I simplify this method to:

 ```javascript
var context = {
  view: window.location.pathname,
  controller: this.className,
  transaction: this.options.appState.get('transaction')
};

this.trigger('navigated', context);

// Tests
expect(test.navigatedSpy).toHaveBeenCalledWith({
  view: '/context.html',
  controller: 'primary-auth',
  transaction: undefined
});
```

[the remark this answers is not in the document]
Comment (haishengwu-okta) [reply src/util/BaseLoginController.js]: gotcha. either way works for me and slightly prefer to keep `undefined`.

File: README.md [hunk 3]
```diff
@@ -354,8 +354,12 @@ Subscribe to an event published by the widget.
...
 
 ```javascript
-signIn.on('pageRendered', function (data) {
-  console.log(data);
+signIn.on('navigated', function (context) {
+  console.log(context);
+  // {
+  //   view: '/',
+  //   controller: 'primary-auth'
+  // }
```
haishengwu-okta: - add `transaction` field
- I wonder if it's worth to use a actual event as example here cause it's sort of duplicated with "events" section. Something like

```
 - `event` - [Event](#events) to subscribe to
 - `callback` - Function to call when the event is triggered. That function takes one parameter optional `context`. More details see [Event](#events).

signIn.on('EVENT_NAME', function (context) { ... });

```

README.md
```diff
@@ -370,11 +374,11 @@ Unsubscribe from widget events. If no callback is provided, unsubscribes all lis
...
 
-// Unsubscribe all listeners that have been registered to the 'pageRendered' event
-signIn.off('pageRendered');
+// Unsubscribe all listeners that have been registered to the 'navigated' event
+signIn.off('navigated');
 
-// Unsubscribe the onPageRendered listener from the 'pageRendered' event
-signIn.off('pageRendered', onPageRendered);
+// Unsubscribe the onNavigated listener from the 'navigated' event
+signIn.off('navigated', onNavigated);
```
haishengwu-okta: ditto.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -539,6 +542,31 @@ function (Okta, Q, Logger, OktaAuth, Util, Expect, Router,
...
+        expect(test.navigatedSpy.calls.count()).toBe(2);
+        expect(test.navigatedSpy.calls.allArgs()[0]).toEqual([{
+          view: '/context.html',
+          controller: 'refresh-auth-state',
+          transaction: undefined
+        }]);
+        expect(test.navigatedSpy.calls.allArgs()[1]).toEqual([{
+          view: '/context.html',
+          controller: 'primary-auth',
+          transaction: jasmine.objectContaining({
```
haishengwu-okta: is there chance to do exact match or is it because too much noise coming from the Transaction class

README.md
jmelberg-okta: Great idea. I'm going to shorten the content in this section, and point our readers to the `#Events` section.

test/unit/spec/LoginRouter_spec.js
jmelberg-okta: Pushed an update to assert the entire object. Good reminder that we should check for as many values as possible.

haishengwu-okta: also I'm wondering if we expose too much information by having
- `authenticate` function (probably others per different transaction
- `data` the raw response

was the intention is expose as much information as possible and let user do whatever they want?
my worry is too flexible unless it's a clear request from users (I may miss some information)

src/util/BaseLoginController.js
```diff
@@ -101,6 +101,9 @@ define(['okta', 'q'], function (Okta, Q) {
     postRenderAnimation: function() {
       // Event triggered after a page is rendered along with the classname to identify the page
       this.trigger('pageRendered', {page: this.className});
+      this.trigger('navigated', {
+        controller: this.className
```
haishengwu-okta: Slightly prefer ‘page’.

src/util/BaseLoginController.js
```diff
@@ -101,6 +101,13 @@ define(['okta', 'q'], function (Okta, Q) {
     postRenderAnimation: function() {
       // Event triggered after a page is rendered along with the classname to identify the page
       this.trigger('pageRendered', {page: this.className});
+
+      // If a stateToken exists, return it in the event context
+      var transaction = this.options.appState.get('transaction') && this.options.appState.get('transaction').data;
```
haishengwu-okta: could create a derived prop `stateToken` in AppState and will be just `this.options.appState.get('stateToken')`?

PR #575, opened 2018-10-30
feat: Adds beforeNavigated event

jmelberg-okta:
### Description

Adds new `beforeNavigate` event.

```javascript
signIn.on('beforeNavigate', (data, context) => {
  // Assume view is transitioning between Primary Auth and MFA Push
  console.log(data);
  /**
   * {
   *   toController: 'mfa-verify'
   * }
   */
  console.log(context);
  /**
   * {
   *   controller: 'primary-auth'
   * }
   */
});
```

### DIFF
--- a/src/util/BaseLoginRouter.js
+++ b/src/util/BaseLoginRouter.js
@@ -231,6 +231,13 @@ function (Okta, BrowserFeatures, Settings,
           return;
         }
 
+        // Emit a beforeNavigate event with the current and next controller name
+        this.trigger(
+          'beforeNavigate',
+          { toController: this.controller.className },
+          { controller: oldController.className }
+        );
+
         return Animations.swapPages({
           $parent: this.el,
           $oldRoot: oldController.$el,
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ You can learn more on the [Okta + JavaScript][lang-landing] page in our document
 - [Events](#events)
   - [navigated](#navigated)
   - [pageRendered](#pagerendered)
+  - [beforeNavigate](#beforenavigate)
   - [passwordRevealed](#passwordrevealed)
 - [Building the Widget](#building-the-widget)
   - [The `.widgetrc` config file](#the-widgetrc-config-file)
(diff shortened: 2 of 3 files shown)


src/util/BaseLoginRouter.js
```diff
@@ -231,6 +231,13 @@ function (Okta, BrowserFeatures, Settings,
           return;
         }
 
+        // Emit a beforeNavigate event with the current and next controller name
+        this.trigger(
+          'beforeNavigate',
```
haishengwu-okta: 1. feels like the naming convention is not kind of consistent. ?

`navigated` / `beforeNavigate`
v.s.
`afterNavigate` / `beforeNavigate`

2. also `fromController` and `toController`?

jmelberg-okta: Yeah - Since `navigated` is essentially emitting "after navigation occurs", I wonder if it would make more sense to change these into:

- `beforeNavigate` -> `beforeNavigation`
- `controller` -> `fromController`

To me, this is much easier to understand:

```javascript
signIn.on('beforeNavigation', (data, context) => {
  // Assume view is transitioning between Primary Auth and MFA Push
  console.log(data);
  // { toController: 'mfa-verify' }
  console.log(context);
  // { fromController: 'primary-auth' }
});
```

haishengwu-okta: What I was thinking is the pattern `preRender` and `postRender` we have in Courage and iirc `before*` and `after*` is another sort of convention. (`beforeNavigation`, `afterNavigation`)
Even though `navigated` does emit "after navigation occurs", I feel with prefix it's easer to grasp.

Let's see what @rchild-okta thought?

rchild-okta: Is there a master list of the events we want to add for this first pass? It'd be easier to come up with the pattern we want with that in hand. In general, agree with @haishengwu-okta  that before/after (or some common prefixes, suffixes) is more flexible than depending on grammar for tenses. It'd be good to get this right up front - would be a good topic for the design meeting on Friday.

PR #577, opened 2018-11-05
feat: Adds ready and afterRender events

jmelberg-okta:
### Description

Begins implementation around the new design pattern for events. This PR introduces a new event `ready` and renames the existing event `pageRendered` to `afterRender`.

The new event design adheres to the following naming conventions:

- All standard events are triggered **before** the DOM has rendered, and allows developer to "hook into" our controller as a blocking operation. This naming convention will adhere to all events, excluding any obvious outliers. For example, `ready`.
- If an event is prefixed with `after`, the DOM has rendered and the Widget has completed all controller and view logic. This will allow developers to update the UI, log messages, and/or navigate elsewhere under specific scenarios.

### Reviewers

- @haishengwu-okta 
- @manueltanzi-okta 
- @nbarbettini

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,10 @@ You can learn more on the [Okta + JavaScript][lang-landing] page in our document
   - [Bootstrapping from a recovery token](#bootstrapping-from-a-recovery-token)
   - [Feature flags](#feature-flags)
 - [Events](#events)
+  - [ready](#ready)
+  - [afterRender](#afterrender)
+  - [pageRendered](#pagerendered)
+  - [passwordRevealed](#passwordrevealed)
 - [Building the Widget](#building-the-widget)
   - [The `.widgetrc` config file](#the-widgetrc-config-file)
   - [Build and test commands](#build-and-test-commands)
@@ -354,30 +358,28 @@ Subscribe to an event published by the widget.
 
 - `event` - [Event](#events) to subscribe to
 - `callback` - Function to call when the event is triggered
-- `context` - Optional context to bind the callback to
 
 ```javascript
-signIn.on('pageRendered', function (data) {
-  console.log(data);
-});
+// Handle a 'ready' event using an onReady callback
+signIn.on('ready', onReady);
 ```
 
 ### off
 
 Unsubscribe from widget events. If no callback is provided, unsubscribes all listeners from the event.
 
-- `event` - Optional event to unsubscribe from
+- `event` - Optional [event](#events) to unsubscribe from
 - `callback` - Optional callback that was used to subscribe to the event
 
 ```javascript
 // Unsubscribe all listeners from all events
 signIn.off();
 
-// Unsubscribe all listeners that have been registered to the 'pageRendered' event
-signIn.off('pageRendered');
+// Unsubscribe all listeners that have been registered to the 'ready' event
+signIn.off('ready');
 
-// Unsubscribe the onPageRendered listener from the 'pageRendered' event
-signIn.off('pageRendered', onPageRendered);
+// Unsubscribe the onReady listener from the 'ready' event
+signIn.off('ready', onReady);
 ```
 
 ### getTransaction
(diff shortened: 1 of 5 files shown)


src/util/BaseLoginController.js
```diff
@@ -100,7 +100,10 @@ define(['okta', 'q'], function (Okta, Q) {
 
     postRenderAnimation: function() {
       // Event triggered after a page is rendered along with the classname to identify the page
+      // TODO: Deprecate this event in the next major version in favor of 'render'
```
robertjd: Let's use a development mode warning to declare that this is deprecated.

jmelberg-okta: I spent a bit of time looking into doing this, but I don't think it is possible.

The Widget would need to listen to the `pageRendered` event itself, where the handler would then emit another `pageRendered` event. Of course, this introduces an endless event loop.

As long as we document the event's deprecation through our README and CHANGELOG, we should be able to encourage devs to move away from it before `3.0`.

src/util/BaseLoginController.js
```diff
@@ -100,7 +100,10 @@ define(['okta', 'q'], function (Okta, Q) {
 
     postRenderAnimation: function () {
       // Event triggered after a page is rendered along with the classname to identify the page
+      // TODO: Deprecate this event in the next major version in favor of 'render'
       this.trigger('pageRendered', {page: this.className});
+
+      this.trigger('render', { controller: this.className });
```
manueltanzi-okta: Sorry to come back to naming, but I think this should be `rendered` and not `render`. As a developer, I would expect that an event named `render` is triggered when the `render` function is called, not in the `postRender`.

README.md
```diff
@@ -1319,23 +1321,63 @@ features: {
...
+Triggered when the widget transitions to a new page and animations have finished.
+
+```javascript
+signIn.on('pageRendered', function (data) {
+  console.log(data);
+  // { page: 'forgot-password' }
+});
+```
+
+### passwordRevealed
```
manueltanzi-okta: nit: Wonder if we should deprecate this in favor of an `afterRevealPassword` event.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -539,6 +542,21 @@ function (Okta, Q, Logger, OktaAuth, Util, Expect, Router,
...
+      return setup({ stateToken: 'aStateToken' })
+        .then(function (test) {
+          Util.mockRouterNavigate(test.router);
+          test.setNextResponse(resUnauthenticated);
+          test.router.navigate('/app/sso');
+          return Expect.waitForPrimaryAuth(test);
+        })
+        .then(function (test) {
+          expect(test.router.navigate).toHaveBeenCalledWith('', { trigger: true });
+          expect(test.renderSpy.calls.count()).toBe(2);
```
manueltanzi-okta: nit: `.toHaveBeenCalledTimes(2);` (also the ones below)

README.md
```diff
@@ -1319,23 +1321,63 @@ features: {
...
-    });
-    ```
-- **passwordRevealed** - triggered when the show password button is clicked.
+Triggered when the widget is ready to accept user input for the first time. Returns a `context` object containing the following properties:
+
+- **controller** - Current controller name
+
+```javascript
+signIn.on('ready', function (context) {
+  if (context.controller === 'primary-auth') {
```
haishengwu-okta: could `ready` event happen at other context rather than `primary-auth`?

jmelberg-okta: Yep. If the Widget is passed in a `recovery`, `activation`, or `stateToken`, the context will be mapped to that page. For example:

Example

README.md
jmelberg-okta: Good call. I'm going to track that with [OKTA-198666](https://oktainc.atlassian.net/browse/OKTA-198666) and will pick that up next.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -71,20 +71,23 @@ function (Okta, Q, Logger, OktaAuth, Util, Expect, Router,
       var baseUrl = 'https://foo.com';
       var authClient = new OktaAuth({url: baseUrl, headers: {}});
       var eventSpy = jasmine.createSpy('eventSpy');
+      var renderSpy = jasmine.createSpy('renderSpy');
```
haishengwu-okta: nit:
`afterRenderCallback` or `afterRenderHandler`

README.md
haishengwu-okta: then let's add a test other than the ready page is primary auth?

jmelberg-okta: Done! Added a few more for coverage.

PR #578, opened 2018-11-06
:art: Fixes flaky unit test

haishengwu-okta:
- Fixes several flakiness in `MfaVerify_spec.js`, `EnrollTotpController_spec.js`, `LoginRouter_spec.js`.
- Add `deprecate` annotation to `tick` help function

### DIFF
--- a/test/unit/helpers/util/Expect.js
+++ b/test/unit/helpers/util/Expect.js
@@ -152,13 +152,6 @@ define([
     return fn.wait(condition, resolveValue);
   };
 
-  fn.waitForEnrollChoices = function (resolveValue) {
-    var condition = function () {
-      return $('.enroll-choices').length > 0;
-    };
-    return fn.wait(condition, resolveValue);
-  };
-
   fn.wait = function (condition, resolveValue, timeout) {
     function check(success, fail, triesLeft) {
       if (condition()) {
--- a/test/unit/helpers/util/Expect.js
+++ b/test/unit/helpers/util/Expect.js
@@ -139,7 +139,7 @@ define([
   };
 
   /**
-   * Use this function to wait for a error view which has top level class '.okta-form-infobox-error'.
+   * Use this function to wait for an error view which has top level class '.okta-form-infobox-error'.
    */
   fn.waitForFormError = function (form, resolveValue) {
     var condition = function () {
@@ -149,7 +149,7 @@ define([
   };
 
   /**
-   * Use this function to wait for a error view which has top level class '.okta-infobox-error'.
+   * Use this function to wait for an error view which has top level class '.okta-infobox-error'.
    */
   fn.waitForFormErrorBox = function (form, resolveValue) {
     var condition = function () {
--- a/test/unit/helpers/util/Expect.js
+++ b/test/unit/helpers/util/Expect.js
@@ -86,6 +86,10 @@ define([
 
   fn.fitp = runTest.bind({}, fit);
 
+  /**
+   * @deprecated This is very hacky rather solving the actual reason of flakiness.
+   *             Please slack eng-ui and think twice before you want to use this function.
+   */
   fn.tick = function (returnVal) {
     var deferred = Q.defer();
     // Using four setTimeouts to remove flakiness (some tests need an extra
--- a/test/unit/helpers/util/Expect.js
+++ b/test/unit/helpers/util/Expect.js
@@ -138,6 +138,9 @@ define([
     return fn.wait(condition, resolveValue);
   };
 
+  /**
+   * Use this function to wait for a error view which has top level class '.okta-form-infobox-error'.
+   */
   fn.waitForFormError = function (form, resolveValue) {
     var condition = function () {
       return form.hasErrors();
@@ -145,6 +148,9 @@ define([
     return fn.wait(condition, resolveValue);
   };
 
+  /**
+   * Use this function to wait for a error view which has top level class '.okta-infobox-error'.
+   */
   fn.waitForFormErrorBox = function (form, resolveValue) {
     var condition = function () {
       return form.errorBox().length > 0;
--- a/test/unit/helpers/util/Expect.js
+++ b/test/unit/helpers/util/Expect.js
@@ -87,8 +87,8 @@ define([
   fn.fitp = runTest.bind({}, fit);
 
   /**
-   * @deprecated This is very hacky rather solving the actual reason of flakiness.
-   *             Please slack eng-ui and think twice before you want to use this function.
+   * @deprecated This function is non deterministic and can affect the output of the tests
+   *             Instead use any of the Expect.wait* functions.
    */
   fn.tick = function (returnVal) {
     var deferred = Q.defer();
@@ -141,6 +141,20 @@ define([
   fn.waitForFormError = function (form, resolveValue) {
     var condition = function () {
       return form.hasErrors();
+    };
+    return fn.wait(condition, resolveValue);
+  };
+
+  fn.waitForFormErrorBox = function (form, resolveValue) {
+    var condition = function () {
+      return form.errorBox().length > 0;
+    };
+    return fn.wait(condition, resolveValue);
+  };
+
+  fn.waitForEnrollChoices = function (resolveValue) {
+    var condition = function () {
+      return $('.enroll-choices').length > 0;
     };
     return fn.wait(condition, resolveValue);
   };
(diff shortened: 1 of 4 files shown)


test/unit/helpers/util/Expect.js
```diff
@@ -86,6 +86,10 @@ define([
 
   fn.fitp = runTest.bind({}, fit);
 
+  /**
+   * @deprecated This is very hacky rather solving the actual reason of flakiness.
+   *             Please slack eng-ui and think twice before you want to use this function.
+   */
```
jmelberg-okta: Best option for us is to work toward removing this method entirely. Can we create a follow-up JIRA to remove `tick` from these tests?

test/unit/spec/MfaVerify_spec.js
```diff
@@ -263,12 +263,15 @@ function (Okta,
...
       if (options.u2f) {
         window.u2f = {
-          sign: function () {
-          }
+          sign: jasmine.createSpy('window-u2f-spy')
         };
-        spyOn(window.u2f, 'sign');
         if (options && options.signStub) {
-          window.u2f.sign.and.callFake(options.signStub);
+          window.u2f.sign.and.callFake(function () {
```
jmelberg-okta: Could we simplify this to use `spyOn` instead since we aren't asserting the spy gets called above?

```javascript
spyOn(window.u2f, 'sign').and.callFake(function () { ... });
```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2855,7 +2858,15 @@ function (Okta,
             return Expect.waitForSpyCall(webauthn.getAssertion, test);
           })
           .then(function (test) {
-            expect(test.form.el('o-form-error-html').length).toBe(1);
+            return Expect.wait(() => {
```
jmelberg-okta: Since we're moving toward a pattern where the `errorBox` can be populated with dynamic messages (custom views, events, etc) - I wonder if we should add a shared `Expect` method called `waitForErrorBox` that will handle this for us.

test/unit/helpers/util/Expect.js
haishengwu-okta: There are 28 test files using `tick` and I'm not sure if a single JIRA would be useful or 28 tickets.

test/unit/spec/MfaVerify_spec.js
haishengwu-okta: we do actually. when `u2f` option is true, we are expecting `u2f.sign` exists but may not test the logic after user tap the key.

test/unit/spec/MfaVerify_spec.js
haishengwu-okta: I doubt it will introduce some confusion because we have `waitForFormError` today.
The reason it doesn't here is because the error was added manually rather from Courage hence the error doesn't have expected class.
Basically `.okta-form-infobox-error` v.s. `.okta-infobox-error`.

Maybe we shall when it appears in many places.

test/unit/helpers/util/Expect.js
jmelberg-okta: 😱 

Gotcha. I'm wondering if adding this deprecation notice is enough to encourage developers **not** to use this method. For now, we could probably get away with catching this during the review stage.

test/unit/spec/MfaVerify_spec.js
jmelberg-okta: Sounds good!

test/unit/helpers/util/Expect.js
mauriciocastillosilva-okta: I like the idea of deprecating this function, so devs don't use it anymore. But I think we should probably change the comment, since this is open source and eng-ui slack channel is only internal
Can we say something like
```
@deprecated This function is non deterministic and can affect the output of the tests
Instead use any of the Expect.wait* functions 
```

test/unit/spec/EnrollTotpController_spec.js
```diff
@@ -862,7 +862,7 @@ function (Okta, OktaAuth, LoginUtil, Util, DeviceTypeForm, BarcodeForm,
             Expect.isVisible(test.linkSentConfirmation.smsSentMsg());
             expect(test.linkSentConfirmation.getMsgText().indexOf('+14152554668') >= 0).toBe(true);
             test.originalAjax = Util.resumeEnrollFactorPoll(test.ac, test.originalAjax, resAllFactors);
-            return tick(test);
+            return Expect.wait(() => $('.enroll-choices').length > 0, test);
```
mauriciocastillosilva-okta: should this be `Expect.waitForEnrollChoices();`?

test/unit/spec/MfaVerify_spec.js
mauriciocastillosilva-okta: with this change we have `return Expect.waitForSpyCall(webauthn.getAssertion, test);` followed by another wait `return Expect.wait(() => {`
Do we need both or can we get rid of the first one and then just assert that it was called (without a wait block)
```
  return Expect.waitForFormErrors(test.form, test);
})
.then(function (test) {
  expect(webauthn.getAssertion).toHaveBeenCalled();
  expect(test.form.errorBox().length).toBe(1);
  ...
```

test/unit/spec/MfaVerify_spec.js
```diff
@@ -263,12 +263,15 @@ function (Okta,
...
-          sign: function () {
-          }
+          sign: jasmine.createSpy('window-u2f-spy')
         };
-        spyOn(window.u2f, 'sign');
         if (options && options.signStub) {
-          window.u2f.sign.and.callFake(options.signStub);
+          window.u2f.sign.and.callFake(function () {
+            var signArgs = arguments;
+            window.u2f.tap = function () {
```
mauriciocastillosilva-okta: 👍

test/unit/helpers/util/Expect.js
haishengwu-okta: @mauriciocastillosilva-okta @jmelberg-okta 
I hope it helps a little bit and we shall deny any code (new or updating exists) that still uses `tick`.
I'm all ears for any ideas that helps fixing flakyness. :)

test/unit/helpers/util/Expect.js
```diff
@@ -141,6 +145,20 @@ define([
     return fn.wait(condition, resolveValue);
   };
 
+  fn.waitForFormErrorBox = function (form, resolveValue) {
```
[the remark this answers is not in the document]
haishengwu-okta: https://github.com/okta/okta-signin-widget/pull/578#discussion_r231221978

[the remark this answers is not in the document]
mauriciocastillosilva-okta: got it 👍 
Do you think a comment on top of both functions will help devs to know which function to use where?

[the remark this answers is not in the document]
haishengwu-okta: sounds like a plan! will add.

test/unit/helpers/util/Expect.js
```diff
@@ -134,13 +138,26 @@ define([
     return fn.wait(condition, resolveValue);
   };
 
+  /**
+   * Use this function to wait for a error view which has top level class '.okta-form-infobox-error'.
```
jmelberg-okta: a error -> an error

test/unit/helpers/util/Expect.js
```diff
@@ -134,13 +138,26 @@ define([
...
+   */
   fn.waitForFormError = function (form, resolveValue) {
     var condition = function () {
       return form.hasErrors();
     };
     return fn.wait(condition, resolveValue);
   };
 
+  /**
+   * Use this function to wait for a error view which has top level class '.okta-infobox-error'.
```
jmelberg-okta: a error -> an error

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2315,30 +2318,38 @@ function (Okta,
               });
             });
             itp('starts poll after a delay of 6000ms', function () {
-              return setupOktaPush().then(function (test) {
-                spyOn(LoginUtil, 'callAfterTimeout').and.callFake(function() {
-                  // reducing the timeout to 100 so that test is fast.
-                  return setTimeout(arguments[0], 100);
+              var callAfterTimeoutStub;
+
```
jmelberg-okta: nit: Remove the newline please

haishengwu-okta: actually it was added deliberately

jmelberg-okta: Is there a reason for it? This seems like a one-off from the layout of our other specs.

PR #579, opened 2018-11-07
Added iframe with src when creating - devicefingerprint flow

magizh-okta:
* Creating the iframe without source and appending to the dom causes  IE embedded browser to crash in office
    2016 and 2010. (We are adding src later as a attribute).
* Moved to creating iframe with source and appending.

Tested on chrome mac, IE 10, embedded browser. 

screen shot 2018-11-07 at 11 19 21 am
screen shot 2018-11-07 at 11 19 44 am
screen shot 2018-11-07 at 11 22 22 am
screen shot 2018-11-07 at 11 23 20 am

inexcelwin10

### DIFF
--- a/src/util/DeviceFingerprint.js
+++ b/src/util/DeviceFingerprint.js
@@ -32,12 +32,7 @@ define(['q', 'okta'], function (Q, Okta) {
 
       var deferred = Q.defer();
       var self = this;
-
-      // Create iframe
-      var $iframe = $('<iframe>', {
-        style: 'display: none;'
-      });
-      $iframe.appendTo(element);
+      var $iframe;
 
       function isWindowsPhone(userAgent) {
         return userAgent.match(/windows phone|iemobile|wpdesktop/i);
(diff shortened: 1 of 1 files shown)


dannavarro-okta:
@magizh-okta did you confirm that the fingerprint is actually sent in these flows? As in, I see that this unblocks logins, I'm just curious if the rest of the fingerprint flow succeeds

magizh-okta:
screen shot 2018-11-06 at 10 08 17 pm
@dannavarro-okta  here is the fidler trace showing the authn call is made with fingerprint.

src/util/DeviceFingerprint.js
```diff
@@ -86,8 +82,11 @@ define(['q', 'okta'], function (Q, Okta) {
 
       // Attach listener
       window.addEventListener('message', onMessageReceivedFromOkta, false);
-      // Load devicefingerprint page inside the iframe
```
haishengwu-okta: why remove the comment? I wonder if shall add a bit more comments regarding why loading devicefingureprint inside a iframe?

magizh-okta: Added back and also as we spoke lets deal with refactoring this file in a later PR.

PR #581, opened 2018-11-13
:art: Upgrade backbone to 1.3.3

haishengwu-okta:
- backbone was upgrade in courage last year https://github.com/okta/courage/commit/e7dcca7c9ae65391813b376eb3925d1b2143226e
- it's non-breaking change and doesn't seem a good that we shall leave OSW behind.


haishengwu-okta:
@rajaguruduraisamy-okta 
yes, this is the version widget's using. https://github.com/okta/okta-signin-widget/blob/master/packages/%40okta/courage-for-signin-widget/package.json

PR #583, opened 2018-11-13
Deprecates grunt eslint

haishengwu-okta:
- remove `grunt-eslint`
- add `eslint-5.9` as Dev dependency
- simplified the `.eslintrc.js` to exclude rules have been enabled by default.
- run through `eslint --fix` for most of them (mostly space changes) and a few manual fix
- **ignore whitespace** would dramatically reduce the change set.


src/util/BaseLoginRouter.js
```diff
@@ -29,8 +29,8 @@ define([
   'util/Logger'
 ],
 function (Okta, BrowserFeatures, Settings,
-          Header, SecurityBeacon, AuthContainer, AppState, RouterUtil, Animations,
-          Errors, Bundles, Logger) {
+  Header, SecurityBeacon, AuthContainer, AppState, RouterUtil, Animations,
+  Errors, Bundles, Logger) {
```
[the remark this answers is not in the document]
haishengwu-okta: yeah, I guess it comes from auto-fix. We could make change as long as having chance to make changes to those files but probably not at this migration PR. (avoid code as much as possible)

.eslintignore
```diff
@@ -0,0 +1,13 @@
...
+build2
+dist
+target
+test-reports
+
+packages
+scripts
+assets
+vendor
+test/unit/helpers/xhr/
```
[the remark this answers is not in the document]
haishengwu-okta: yes and will have lots more changes (eg. double quote -> single quote) hence I leave as it's today.
my best guess it's because those are playload copy from somewhere hence it's not really JavaScript code.

will follow up at: https://oktainc.atlassian.net/browse/OKTA-197768

src/util/BaseLoginRouter.js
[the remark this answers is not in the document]
haishengwu-okta: yeah and convert to ES6 would save the effort.

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -377,23 +377,23 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, Router,
...
         });
         itp('does not show the Finish button if there are no required factors and no ' +
             'optional factors have been enrolled',
-          function () {
-            return setupWithAllOptionalNoneEnrolled().then(function (test) {
-              expect(test.form.submitButton().length).toBe(0);
-            });
+        function () {
+          return setupWithAllOptionalNoneEnrolled().then(function (test) {
+            expect(test.form.submitButton().length).toBe(0);
```
manueltanzi-okta: nit: I wonder if we can add some linting rule to force the use some syntactic sugar in tests expectations (ex.: `.toHaveLength(0)`).

src/EnrollCustomFactorController.js
```diff
@@ -34,21 +34,21 @@ function (Okta, FormController, Footer) {
...
-            setTransaction(trans);
-            var url = this.appState.get('enrollCustomFactorRedirectUrl');
-            if(url !== null) {
-              Util.redirect(url);
-            }
-          })
-          .fail(function (err) {
-            throw err;
-          });
+            .then((trans) => {
```
manueltanzi-okta: nit: we could add the rule to remove parenthesis on single param arrow functions

haishengwu-okta: I think you meant https://eslint.org/docs/rules/arrow-parens
Will enable at next refactor phase if it's preferred.

test/unit/spec/EnrollChoices_spec.js
manueltanzi-okta: Is also part of `jasmine-jquery`

haishengwu-okta: hmm, interesting! will dig into a little bit after `grunt` deprecation.

PR #584, opened 2018-11-14
Revert ":art: Upgrade backbone to 1.3.3 (#581)"

haishengwu-okta:
This reverts commit a54a7a8683d489f714689621c75febceee3d2670.

## Description

- upgrade to 1.3.3 caused underscore has been bundled twice.
- looks like backbone depends on 'underscore >= 1.8.3' which will always try to pull latest '1.9.1' at this time
- but sign-in widget explicitly depends on underscore-1.8.3

revert changes and cycle back later (probably after jQuery work when we have better idea at either upgrade both or include both within courage bundle file)

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "@okta/okta-auth-js": "1.17.0",
     "babel-polyfill": "^6.23.0",
     "babel-runtime": "^6.26.0",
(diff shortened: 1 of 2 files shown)


haishengwu-okta:
@jmelberg-okta it indeed odd

PR #586, opened 2018-11-15
:art: Upgrade to webpack3

haishengwu-okta:
1. Upgrades to webpack3 for okta-signin-widget
2. Upgrades to webpack3 for courage-for-sign-in-widget package
3. Adds remove-strict plugin for test due to 
    - `navigator.credential` is read only and not allow to override at strict mode (the problem wasn't surfaced up due to webpack1 'use strict' enabling bug)
    - will follow up at https://oktainc.atlassian.net/browse/OKTA-197783
4. Migrates sample server to use webpack dev server
    - _another improvement would be combining signin-widget and sample javascript so that any change to signin widget source code will be picked up automatically by dev server._
    - the goal of this PR is focusing upgrade hence do as less refactor.
5. Tested at (https://okta.box.com/s/jfju1k1w7d0pommcidpdsg26w9gt44gs)
    - [x] monolith selenium (bacon green)
    - [x] IE desktop
    - [x] embedded browsers


Gruntfile.js
```diff
@@ -364,16 +293,6 @@ module.exports = function (grunt) {
...
         keepalive: true
       },
-      server: {
-        options: {
-          open: false
-        }
-      },
-      open: {
-        options: {
-          open: true
```
jmelberg-okta: I personally like the auto-open behavior - is there a reason why we decided to remove it?

buildtools/sample-server/index.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
```
jmelberg-okta: What do you think about moving the `sample-server` outside of `build-tools` and into the root of this project? That way, the playground is in a more obvious location for developers who are wanting to make changes to `index.html`

buildtools/sample-server/main.js
```diff
@@ -0,0 +1,73 @@
...
+    rememberMe: true,
+    multiOptionalFactorEnroll: true
+  },
+  // Host the assets (i.e. jsonp files) locally
+  assets: {
+    baseUrl: '/'
+  }
+};
+
+const options = Object.assign({}, defaultOptions, widgetRc.widgetOptions);
```
jmelberg-okta: nit: Can we move the above logic out into another file?

To me, this file acts as the blueprint for rendering the Widget, where the code in this file should be nearly identical to our README, quickstarts, and sample apps. Adding configuration defaults above  _might_ confuse devs when trying to decide **where** to put their Widget config.

webpack.sample.server.config.js
```diff
@@ -0,0 +1,76 @@
+/* global module __dirname */
+
+const path = require('path');
+const fs = require('fs');
+const CopyWebpackPlugin = require('copy-webpack-plugin');
+const TARGET = path.resolve(__dirname, 'target');
```
jmelberg-okta: nit: Newline between require statements and assignments.

buildtools/sample-server/main.js
rchild-okta: Isn't this closer to what a developer would actually write though? I.e. bootstrap the widget with some options, and then call render. In the easy cases not sure how many other developers would split out widget config to a separate file.

buildtools/sample-server/main.js
```diff
@@ -0,0 +1,73 @@
+/* global OktaSignIn */
+/* eslint no-console: 0 */
+
+import widgetRc from '../../.widgetrc';
+
+const defaultOptions = {
+  baseUrl: 'http://rain.okta1.com:1802',
```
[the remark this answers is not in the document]
rchild-okta: Good call; guessing @haishengwu-okta is just using the same options in the previous rev of the devServer, but probably good to change it now to something invalid that the developer will need to pass in upfront.

buildtools/sample-server/main.js
```diff
@@ -0,0 +1,73 @@
+/* global OktaSignIn */
+/* eslint no-console: 0 */
+
+import widgetRc from '../../.widgetrc';
```
rchild-okta: Just curious - this is automatically resolved to using the json loader?

buildtools/sample-server/main.js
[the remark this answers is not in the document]
haishengwu-okta: hmm, good point. I'm thinking leave a real org make more sense. For example the org used for signin widget e2e test?

buildtools/sample-server/main.js
haishengwu-okta: yeah, basically.
```
{
        test: /\.json$|\.widgetrc/,
        loader: 'json-loader'
      },
```

Gruntfile.js
haishengwu-okta: hah, I guess it's more personal preference whether turn on auto-open.
will enable it in the webpack-dev-server config

buildtools/sample-server/index.html
haishengwu-okta: I actually didn't have an opinion whether it shall live in its own folder and what could be good name (eg. `example`, `sample-app`,  `sample-server`)
Hence keep in buildtools for simplicity for now.

buildtools/sample-server/main.js
haishengwu-okta: - maybe the `defaultOptions` is not best naming but just `widgetOptions` 

```
Object.assign({}, widgetOptions, widgetRc.widgetOptions)
```

- another approach is to move all those `defaults` to `webpack.sample.server.config` where generates those into default `widgetrc`.

jmelberg-okta: +1 on bringing defaults into the `widgetrc` file.

The case I'm trying to avoid is having developers update the `defaultOptions` instead of the `.widgetrc` file. We saw this a few times within our `oidc-js` repo, and needed to bring the default logic into a shared config to reduce friction: [`.oidc-config.js`](https://github.com/okta/okta-oidc-js/blob/master/.oidc.config.js).

buildtools/sample-server/main.js
rchild-okta: Hah! I checked the wrong config file - common.config rather than playground.config

buildtools/sample-server/main.js
rchild-okta: Yeah, I suppose there's no reason _not_ to have a default .widgetrc already in place, with the pre-configured defaults. Would simplify this build logic and also simplify conceptually - just change the one and it'll show up.

buildtools/sample-server/main.js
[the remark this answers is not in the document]
haishengwu-okta: changed to `your-org.okta.com`

buildtools/sample-server/index.html
haishengwu-okta: moved to `playground`

packages/@okta/courage-for-signin-widget/webpack.config.js
```diff
@@ -36,13 +26,10 @@ const webpackConfig = {
     // node_modules at root directory.
     path: path.resolve(__dirname, '../'),
     filename: `${DIST_FILE_NAME}.js`,
-    libraryTarget: 'umd',
-    library: DIST_FILE_NAME
+    libraryTarget: 'commonjs2'
```
jmelberg-okta: Are we ok changing the target to `commonjs2`? This seems like we're introducing a breaking change.

playground/main.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    // Unlock account flow
+    if (res.status === 'UNLOCK_ACCOUNT_EMAIL_SENT') {
+      alert('SUCCESS: Unlock account email sent');
+      return;
+    }
+
+    //IDP Discovery
+    if (res.status == 'IDP_DISCOVERY') {
```
jmelberg-okta: nit: `===`

webpack.playground.config.js
```diff
@@ -0,0 +1,72 @@
...
+const fs = require('fs');
+
+const TARGET = path.resolve(__dirname, 'target');
+const PLAYGROUND = path.resolve(__dirname, 'playground');
+const DEFAULT_SERVER_PORT = 3000;
+const WIDGET_RC = '.widgetrc';
+
+let widgetRc = {
+  widgetOptions: {
+    baseUrl: 'https://your-org.okta.com',
```
jmelberg-okta: The placeholder text we use for all open source repos is `https://{yourOktaDomain}`, as it can imply a vanity URL, preview orgs and prod orgs.

Can we change this value to be consistent with our API docs + other repos?

packages/@okta/courage-for-signin-widget/webpack.config.js
haishengwu-okta: I think it's fine because this file is only being used by signin widget internally. `umd` target is for broader audience.

jmelberg-okta: Awesome - just wanted to be sure.

PR #588, opened 2018-11-15
:bug: Fixes the flaky FIDO factor verify

haishengwu-okta:


test/e2e/specs/OIDC_spec.js
```diff
@@ -112,7 +112,8 @@ describe('OIDC flows', function () {
 
   });
 
-  describe('Social IDP', function () {
+  // https://oktainc.atlassian.net/browse/OKTA-197861
+  xdescribe('Social IDP', function () {
```
jmelberg-okta: Can we move only disable the `'can login and get a "code" using the redirect flow'` test? The others seem to be working properly still.

haishengwu-okta: @jmelberg-okta actually I'm seeing 2 of them.

jmelberg-okta: Discussed offline - due to what appears to be a Social Login bug, we're going disable the Social IDP tests until we can investigate further.

PR #591, opened 2018-11-16
Supporting custom button in idp discovery controller

petermiller-okta:
screen shot 2018-11-16 at 3 07 26 pm

https://oktainc.atlassian.net/browse/OKTA-198129

@okta/idx 
@mauriciocastillosilva-okta @nikhilvenkatraman-okta @gowthamidommety-okta

### DIFF
--- a/src/IDPDiscoveryController.js
+++ b/src/IDPDiscoveryController.js
@@ -17,10 +17,11 @@ define([
   'views/idp-discovery/IDPDiscoveryForm',
   'models/IDPDiscovery',
   'views/shared/Footer',
-  'util/BaseLoginController'
+  'util/BaseLoginController',
+  'views/primary-auth/CustomButtons'
 ],
 function (Okta, PrimaryAuthController, PrimaryAuthModel, IDPDiscoveryForm, IDPDiscoveryModel,
-  Footer, BaseLoginController) {
+  Footer, BaseLoginController, CustomButtons) {
 
   return PrimaryAuthController.extend({
     className: 'idp-discovery',
@@ -39,6 +40,12 @@ function (Okta, PrimaryAuthController, PrimaryAuthModel, IDPDiscoveryForm, IDPDi
       BaseLoginController.apply(this, arguments);
 
       this.addListeners();
+
+      // If social auth is configured, 'socialAuthPositionTop' will determine
+      // the order in which the social auth and primary auth are shown on the screen.
+      if (options.settings.get('hasConfiguredButtons')) {
+        this.add(CustomButtons, {prepend: options.settings.get('socialAuthPositionTop')});
+      }
 
       this.addFooter(options);
 
--- a/test/unit/spec/IDPDiscovery_spec.js
+++ b/test/unit/spec/IDPDiscovery_spec.js
@@ -134,6 +134,64 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, B
     return name;
   }
 
+  function setupWithCustomButtons () {
+    var settings = {
+      customButtons: [
+        {
+          title: 'test text',
+          className: 'test-class',
+          click: function (e) {
+            $(e.target).addClass('new-class');
+          },
+          dataAttr: 'test-data'
+        }
+      ]
+    };
+    return setup(settings);
+  }
+
+  function setupWithCustomButtonsWithIdp () {
+    var settings = {
+      customButtons: [
+        {
+          title: 'test text',
+          className: 'test-class',
+          click: function (e) {
+            $(e.target).addClass('new-class');
+          }
+        }
+      ],
+      idps: [
+        {
+          type: 'FACEBOOK',
+          id: '0oaidiw9udOSceD1234'
+        }
+      ]
+    };
+    return setup(settings);
+  }
+
+  function setupWithoutCustomButtonsAndWithIdp () {
+    var settings = {
+      customButtons: undefined,
+      idps: [
+        {
+          type: 'FACEBOOK',
+          id: '0oaidiw9udOSceD1234'
+        }
+      ]
+    };
+    return setup(settings);
+  }
+
+  function setupWithoutCustomButtonsWithoutIdp () {
+    var settings = {
+      customButtons: undefined,
+      idps: undefined
+    };
+    return setup(settings);
+  }
+
   var setupWithTransformUsername = _.partial(setup, {username: 'foobar', transformUsername: transformUsername});
   var setupWithTransformUsernameOnUnlock = _.partial(setup, {transformUsername: transformUsernameOnUnlock});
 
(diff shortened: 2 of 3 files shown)


taylorlaubach-okta:
What's the UI for configuring this behavior to happen? Which config caused the text "PIV Card" to end up on the button?

petermiller-okta:
@taylorlaubach-okta when we enable `X509_BASED_AUTH`, we show the button on the regular username/password page. Now that we are supporting X509 for Idp Discovery, we would like the button to appear on the "identifier first" page as well

taylorlaubach-okta:
Got it. That makes sense. Thanks!

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1133,6 +1149,106 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, B
         });
       });
     });
-
+    Expect.describe('Additional Auth Button', function () {
+      itp('does not show the divider and buttons if settings.customButtons is not set', function () {
```
nikhilvenkatraman-okta: Can we add another test that checks the if customButtons is set we render the PIV button, the user name field and next button on the identity first page.

src/IDPDiscoveryController.js
```diff
@@ -40,6 +41,12 @@ function (Okta, PrimaryAuthController, PrimaryAuthModel, IDPDiscoveryForm, IDPDi
 
       this.addListeners();
 
+      // If social auth is configured, 'socialAuthPositionTop' will determine
+      // the order in which the social auth and primary auth are shown on the screen.
+      if (options.settings.get('hasConfiguredButtons')) {
+        this.add(CustomButtons, {prepend: options.settings.get('socialAuthPositionTop')});
```
mauriciocastillosilva-okta: this will add all custom buttons the user configures. In our case, that means PIV and also social idps. Just want to confirm that is what we want for the IdpDiscovery login form

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1133,6 +1149,106 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, B
...
+            }
+          ]
+        };
+        return setup(settings).then(function(test){
+          expect(test.form.authDivider().length).toBe(1);
+          expect(test.form.additionalAuthButton().length).toBe(1);
+          expect(test.form.facebookButton().length).toBe(1);
+        });
+      });
+      itp('does not display custom auth when it is undefined', function () {
```
gowthamidommety-okta: Should be `does not display custom buttons when it is undefined`?

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -134,6 +134,22 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, B
     return name;
   }
 
+  function setupAdditionalAuthButton() {
```
gowthamidommety-okta: nit: can this be `setupWithCustomButtons`?

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1133,6 +1149,106 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, B
...
+      });
+      itp('clickHandler is called when button is clicked', function () {
+        return setupAdditionalAuthButton().then(function(test){
+          expect(test.form.additionalAuthButton().hasClass('new-class')).toBe(false);
+          test.form.additionalAuthButton().click();
+          expect(test.form.additionalAuthButton().hasClass('new-class')).toBe(true);
+        });
+      });
+      itp('displays social auth and custom buttons', function () {
+        var settings = {
```
gowthamidommety-okta: Can we have a function for this (and the other tests) as well, similar to `setupAdditionalAuthButton`?

src/IDPDiscoveryController.js
petermiller-okta: @mauriciocastillosilva-okta, I don't see social auth buttons appearing like PIV on the regular login page...is there something else I need to configure other than having `SOCIAL_AUTH` enabled as well as configuring a social idp?

haishengwu-okta: In additions, if we really going to this, the `constructor` looks quite similar to `PrimiaryAuthController.constructor` except how it creates `this.model`.

I wonder if you could refactor the parent construct to extra the difference into function, e.g `createModel`, and override in this child class.

test/unit/spec/IDPDiscovery_spec.js
petermiller-okta: Will add `expect(test.form.forgotPasswordLinkVisible()).toBe(false);` at line 1203

src/IDPDiscoveryController.js
petermiller-okta: @haishengwu-okta can we do without the refactor for this PR? I made the changes to get it to work, but it caused ~25 test failures

petermiller-okta: @mauriciocastillosilva-okta yeah that's the plan https://oktainc.atlassian.net/browse/OKTA-184467

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1133,6 +1186,80 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, B
...
+      });
+      itp('clickHandler is called when button is clicked', function () {
+        return setupWithCustomButtons().then(function(test){
+          expect(test.form.additionalAuthButton().hasClass('new-class')).toBe(false);
+          test.form.additionalAuthButton().click();
+          expect(test.form.additionalAuthButton().hasClass('new-class')).toBe(true);
+        });
+      });
+      itp('displays social auth and custom buttons', function () {
+        //here
```
nikhilvenkatraman-okta: is this comment needed ?

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1133,6 +1186,80 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, B
...
+          ]
+        };
+        return setup(settings).then(function(test){
+          expect(test.form.authDivider().length).toBe(1);
+          expect(test.form.additionalAuthButton().length).toBe(0);
+          expect(test.form.facebookButton().length).toBe(1);
+        });
+      });
+      itp('does not display social auth when it is undefined', function () {
+        //here
```
nikhilvenkatraman-okta: same here

test/unit/spec/IDPDiscovery_spec.js
petermiller-okta: bleh, no

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -134,6 +134,59 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, B
...
+      customButtons: [
+        {
+          title: 'test text',
+          className: 'test-class',
+          click: function (e) {
+            $(e.target).addClass('new-class');
+          }
+        }
+      ],
+      idps: undefined
```
haishengwu-okta: looks like this setup is quite same to `setupWithCustomButtons`. (`idps: undefined` shall be same as omit the property).
I'm curious what difference you found?

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1133,6 +1186,78 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, B
...
+            }
+          ]
+        };
+        return setup(settings).then(function(test){
+          expect(test.form.authDivider().length).toBe(1);
+          expect(test.form.additionalAuthButton().length).toBe(0);
+          expect(test.form.facebookButton().length).toBe(1);
+        });
+      });
+      itp('does not display social auth when it is undefined', function () {
```
mauriciocastillosilva-okta: this should read `does not display social auth when idps is undefined`

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -134,6 +134,59 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, B
...
+        {
+          type: 'FACEBOOK',
+          id: '0oaidiw9udOSceD1234'
+        }
+      ]
+    };
+    return setup(settings);
+  }
+
+  function setupWithCustomButtonsWithoutIdp() {
```
mauriciocastillosilva-okta: I think that this is equivalent to `setupWithCustomButtons`, so we can probably remove it and use that one instead

test/unit/spec/IDPDiscovery_spec.js
petermiller-okta: I just pulled out this config from multiple tests to be shared

PR #593, opened 2018-11-21
:art: migrates custom grunt task to build scripts

haishengwu-okta:
1. Upgrade some dependencies to latest (cssnano, grunt-cli, grunt-sass, time-grunt)
2. Migrate generate-phone-codes script 
    - It has been broken a while ago and it's not used anywhere in build process. 
    - Turns out there is small change with latest google phone code library (VA code was changed from 379 to 39)
    - Will follow up separately. OKTA-198793
3. Migrate language config generate
4. Migrate generate language jsonp
5. remove `app-root-path` and `grunt-json-generator` which are not used at all.


buildtools/commands/generate-phone-codes.js
```diff
@@ -0,0 +1,50 @@
...
+      // http://www.listofcountriesoftheworld.com/country-dialling-codes.html
+      // Note: TF, BV, and HM do not have phone codes
+      regionCodesToCallingCodeMap['GS'] = 500;
+      regionCodesToCallingCodeMap['PN'] = 64;
+      regionCodesToCallingCodeMap['AQ'] = 672;
+      regionCodesToCallingCodeMap['UM'] = 1;
+      regionCodesToCallingCodeMap['AN'] = 599;
+
+      const contents = 'define(' + JSON.stringify(regionCodesToCallingCodeMap, null, 2) + ')';
+      writeFileSync(OUTPUT_FILE, contents);
```
jmelberg-okta: Doesn't have to be today, but we should probably change how this currently works.

We're currently fetching some URL from `master` (should be tied to a SHA instead), and then writing to a file once the phone codes have been generated. Ideally we will have an **update** task that can be run to assert that the checked in Google lib matches our current `countryCallingCodes.js` file, instead of doing this every time `yarn start` is run.

Maybe we update the description of OKTA-198793 to include changing behavior too?

package.json
```diff
@@ -21,9 +21,19 @@
   "main": "./dist/js/okta-sign-in.entry.js",
   "scripts": {
     "clean": "rimraf target && rimraf dist && rimraf build2",
-    "lint:report": "grunt lint && stylelint assets/sass/ > build2/OSW-stylelint-checkstyle.xml --custom-formatter node_modules/stylelint-checkstyle-formatter || true && eslint -f checkstyle -o build2/OSW-eslint-checkstyle.xml . || true",
-    "lint": "grunt lint && stylelint assets/sass/ && eslint .",
-    "test:dev": "grunt generate-config && karma start",
+    "generate-phone-codes": "node buildtools generate-phone-codes",
+    "generate-config": "node buildtools generate-language-config",
+    "generate-jsonp": "node buildtools generate-language-jsonp",
```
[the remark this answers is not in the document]
haishengwu-okta: - `generate-phone-codes` actually isn't invoked anywhere.
- `generate-config` and `generate-jsonp` are required by server. 
- `retirejs` is probably not required (could be moved to lint or prepublish)

Doesn't seem big deal hence I left what it was for now.

buildtools/commands/generate-phone-codes.js
haishengwu-okta: - why should be tied to SHA instead of master?
- what's the advantage to have an *update* task to do assertion (I'll the *update* task would also need to download file and parse it) rather than generating every time?

package.json
```diff
@@ -21,9 +21,19 @@
...
+    "retirejs": "retire --jspath src -p",
```
mauriciocastillosilva-okta: will this work in all environments? I don't see `retire` in the list of `devDependencies` or `dependencies`... we still have `grunt-retire` there though... maybe we need to remove `grunt-retire` and add `retire`?

package.json
```diff
@@ -21,9 +21,19 @@
...
+    "lint:eslint": "eslint .",
+    "lint:eslint:report": "eslint -f checkstyle -o build2/OSW-eslint-checkstyle.xml .",
+    "lint:stylelint": "stylelint assets/sass/",
+    "lint:stylelint:report": "stylelint --custom-formatter node_modules/stylelint-checkstyle-formatter assets/sass/ > build2/OSW-stylelint-checkstyle.xml",
+    "lint:grunt": "grunt lint",
+    "lint:grunt:report": "grunt lint --checkstyle",
+    "lint": "mkdir build2 && run-p -c lint:eslint lint:grunt lint:stylelint",
+    "lint:report": "mkdir build2 && run-p -c lint:eslint:report lint:grunt:report lint:stylelint:report",
```
mauriciocastillosilva-okta: we are generating individual reports right? Will this affect bacon? I know we run our tests in travis, but might be worth checking if bacon was also looking for some report...

package.json
haishengwu-okta: good catch!!
I probably mess up when merging where there shall be changes that remove `grunt-retire` and add `retire` as devDependency.

It works now because `grunt-retire` pull `retire` as dependency.

buildtools/commands/generate-phone-codes.js
```diff
@@ -0,0 +1,50 @@
...
+const METADATA_URI = 'https://raw.githubusercontent.com/googlei18n/libphonenumber/master/javascript/i18n/phonenumbers/metadata.js';
+const ROOT_DIR = resolve(__dirname, '../../');
+const OUTPUT_FILE = resolve(ROOT_DIR, 'src/util/countryCallingCodes.js');
+
+exports.command = 'generate-phone-codes';
+exports.describe = 'Downloads latest metadata from google libphonenumber and generates ' +
+  'json we use to map region codes to country calling codes';
+exports.handler = () => {
+  console.log('Downloading latest metadata from google libphonenumber...');
+  axios.get(METADATA_URI)
```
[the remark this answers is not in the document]
haishengwu-okta: No particular reason but the simpler interface of `axios` than raw the node's https utility.

package.json
haishengwu-okta: looks like it's not working for a while and added the fix.

scripts/lint.sh
```diff
@@ -4,12 +4,14 @@ source $OKTA_HOME/$REPO/scripts/setup.sh
...
 export TEST_RESULT_FILE_DIR="${REPO}/build2"
+echo $TEST_SUITE_TYPE > $TEST_SUITE_TYPE_FILE
+echo $TEST_RESULT_FILE_DIR > $TEST_RESULT_FILE_DIR_FILE
 
 if ! npm run lint:report; then
   echo "lint failed! Exiting..."
-  exit ${TEST_FAILURE}
+  # PUBLISH_TYPE_AND_RESULT_DIR_BUT_ALWAYS_FAIL will invoke bacon to parse results and display number of tests
+  # and exit with failure, displaying all failed tests in BACON modal.
+  exit ${PUBLISH_TYPE_AND_RESULT_DIR_BUT_ALWAYS_FAIL}
```
mauriciocastillosilva-okta: is this the preferred exit value in bacon? I actually don't know but would like to learn

haishengwu-okta: That's what I remember from most recent build script refactor which I thought is a good practice.

PR #594, opened 2018-11-22
feat: Adds afterError event for transactions

jmelberg-okta:
### Description

This PR focuses on authentication, MFA, and recovery errors that are returned via the `/authn` API, and triggered as `transactionErrors`. OAuth errors are captured in the `globalErrorHandler` and will be considered at a later point in time.

This PR is broken into multiple commits:

1. Adds `afterError` event for transactions
    - Some API failures were not returned as Error types, causing the event pattern to fail. This commit introduces two new errors to capture the error context.
    - Some controllers use `functions` for deciding which `className` to return. This was accounted for in our event.
2. Tests
    - Tests for all known possible cases where an `error` could be thrown during authentication.
    - **Excludes**: Enrollment in Windows Hello, as we're purposefully swallowing errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1334,7 +1334,7 @@ signIn.on('ready', function (context) {
 
 ### afterError
 
-The widget will handle most types of errors - for example, if the user enters an invalid password or there are issues authenticating. To capture authentication errors after they have been handled and rendered by the Widget, listen to the `afterError` event. Returns `context` and `error` objects containing the following properties:
+The widget will handle most types of errors - for example, if the user enters an invalid password or there are issues authenticating. To capture an authentication state change error after it is handled and rendered by the Widget, listen to the `afterError` event. Returns `context` and `error` objects containing the following properties:
 
 - `context`:
   - **controller** - Current controller name


src/util/BaseLoginController.js
```diff
@@ -104,6 +105,16 @@ define(['okta', 'q'], function (Okta, Q) {
...
       this.trigger('afterRender', { controller: this.className });
+
+      this.listenTo(this, 'transactionError', function (err) {
+        if (!err.statusCode) {
+          // Bring the statusCode to the top-level of the Error
+          err.statusCode = err.xhr && err.xhr.status;
+        }
+        // Some controllers return the className as a function - process it here:
+        var className = typeof this.className === 'function' ? this.className() : this.className;
+        this.trigger('afterError', { error: err }, { controller: className });
```
mauriciocastillosilva-okta: I am a little confused... we will trigger the event with 2 objects:
```
1: { error: { statusCode: code, message: 'some_message' } }
2: { controller: 'name' }
```
in the README we are saying that we will return 2 objects, an `error` and a `context`, so shouldn't we be returning
```
1: { statusCode: code, message: 'some_message' },
2: { controller: 'name' }
```
It seems to me we don't need to put the error object inside an `error` property, but just return it itself...
```
this.trigger('afterError', err, { controller: className });
```

src/util/BaseLoginController.js
```diff
@@ -104,6 +105,16 @@ define(['okta', 'q'], function (Okta, Q) {
       this.trigger('pageRendered', {page: this.className});
 
       this.trigger('afterRender', { controller: this.className });
+
+      this.listenTo(this, 'transactionError', function (err) {
```
mauriciocastillosilva-okta: can't remember exactly, so can you check if this happens after we "normalize" some errors with the logic in https://github.com/okta/okta-signin-widget/blob/4e0c5d2823eb6a283f8e6f216eabf7e6eeba1eca/src/util/Util.js#L23?

src/util/Errors.js
```diff
@@ -42,12 +42,30 @@ define(['okta', './Enums'], function (Okta, Enums) {
   }
   AuthStopPollInitiationError.prototype = new Error();
 
+  function U2FError (err) {
+    this.name = Enums.U2F_ERROR;
+    this.message = err.xhr.responseJSON.errorSummary;
+    this.errorSummary = err.xhr.responseJSON.errorSummary;
```
mauriciocastillosilva-okta: since both `message` and `errorSummary` are the same, do we need both? I guess in the future some errors might not have the same content, but not sure if we want to just duplicate the data.

test/unit/spec/EnrollCall_spec.js
```diff
@@ -537,6 +557,31 @@ function (Q, Okta, OktaAuth, LoginUtil, Util, Form, Beacon, Expect, $sandbox,
...
           });
       });
+      itp('triggers an afterError event if error response on verification', function () {
+        return setupAndSendValidCode()
+          .then(function (test) {
+            Q.stopUnhandledRejectionTracking();
+            test.setNextResponse(resActivateError);
+            test.form.setCode(123);
+            test.form.submit();
+            return tick(test);
```
mauriciocastillosilva-okta: can we change all the `ticks` in the tests you are adding for `Expect.waitForFormError`? `tick`s have the potential of adding flakiness.

test/unit/spec/MfaVerify_spec.js
```diff
@@ -3052,6 +3128,35 @@ function (Okta,
...
+              expect(test.afterErrorHandler).toHaveBeenCalledTimes(1);
+              expect(test.afterErrorHandler.calls.allArgs()[0]).toEqual([
+                {
+                  error: jasmine.objectContaining({
+                    name: 'U2F_ERROR',
+                    message: 'There was an error with the U2F request. Try again or select another factor.'
+                  })
+                },
+                {
+                  controller: 'mfa-verify verify-u2f'
```
mauriciocastillosilva-okta: might be for a follow-up jira... should we have just one name for the controller? I think only MFAVerifyController has this problem...

test/unit/spec/RecoveryQuestion_spec.js
```diff
@@ -239,6 +242,30 @@ function (Okta, OktaAuth, Util, RecoveryQuestionForm, Beacon, Expect, Router,
           expect(test.form.errorMessage()).toBe('The recovery question answer did not match our records.');
         });
     });
+    itp('triggers an afterError event if there is an error submitting the answer', function () {
+      return setup()
+        .then(function (test) {
+          test.setNextResponse(resError);
+          test.form.setAnswer('4444');
+          test.form.submit();
+          return tick(test);
```
mauriciocastillosilva-okta: here too... In general for the tests that we expect errors, let's use `Expect.waitForFormError`

test/unit/spec/UnlockAccount_spec.js
```diff
@@ -336,6 +355,18 @@ function (Q, Okta, OktaAuth, Util, AccountRecoveryForm, Beacon, Expect, Router,
             expect(test.form.errorMessage()).toBe('You do not have permission to perform the requested action');
           });
       });
+      itp('triggers an afterError event if sending email results in an error', function () {
+        return setup()
+          .then(function (test) {
+            test.setNextResponse(resError);
+            test.form.setUsername('foo');
+            test.form.sendEmail();
+            return tick(test);
```
mauriciocastillosilva-okta: same for the tests in this spec. lets use `Expect.waitForFormError`

src/util/BaseLoginController.js
jmelberg-okta: You're correct! This was a *rookie* mistake on my end. Fixed with the next commit.

test/unit/spec/MfaVerify_spec.js
jmelberg-okta: Yeah - we have nested names for a few controllers currently. We're going to be looking at this with a major version rev of the Widget. I'll create a JIRA to track this though

src/util/BaseLoginController.js
jmelberg-okta: Great question - we modify the `xhr` object by replacing the the following:

```javascript
 xhr.responseJSON = {
  errorSummary: 'something'
};
```

Normalization occurs on the request response, so the `err` object received here is already transformed.

README.md
```diff
@@ -1335,6 +1336,35 @@ signIn.on('ready', function (context) {
...
+- `error`:
+  - **name** - Name of the error triggered
+  - **message** - Error message
+  - **statusCode** - HTTP status code  (if available)
+  - **xhr** - HTTP response (if available)
+- `context`:
+  - **controller** - Current controller name
+
+```javascript
+signIn.on('afterError', function (error, context) {
```
haishengwu-okta: curious why you like to split `controller` to another variable but not keep in `error` object?

src/util/BaseLoginController.js
haishengwu-okta: curious your thoughts on why built another layer to handle the error (trigger `transactionError` at line 58 and handles it at line 109) instead of invoke the handler at line 58 directly?

src/util/BaseLoginController.js
```diff
@@ -55,6 +55,7 @@ define(['okta', 'q'], function (Okta, Q) {
       }, this);
       var transactionErrorHandler = _.bind(function (err) {
         this.options.appState.set('transactionError', err);
+        this.trigger('transactionError', err);
```
haishengwu-okta: Turns out `transactionError` will eventually invoke `callGlobalError` at certain cases.
In additions, `callGlobalError` is invoked directly when error happened at some case (search `callGlobalError` through code base)
So I'm wondering we introduces some confusion to developers because there are two error handlers basically. (globalError handler and afterError handler)

README.md
jmelberg-okta: Primarily to be consistent with all general purpose events, as there will always be a `context` object with information related to the current state. To me, merging `context` with `error` overloads the `error` object.

manueltanzi-okta: Make sense to split them, however, I think if we want to be consistent in general we should keep the common returned objects in the same positions. 
Just for example, if we have a general callback function that use `context`, we cannot use for both render events and errors event, because in afterRender `context` would be the first parameter, while in afterError event `context` would be the second parameter.

jmelberg-okta: Good point - I originally thought the more important param *should* be first, but I do like the idea of consistency between all events.

Updated in a follow-up commit.

src/util/BaseLoginController.js
jmelberg-okta: When I was originally testing this change, we needed to emit the `afterError` event **after** some code execution path. Turns out, we were firing this event a little too early.

I've pushed a new commit with changes to address your concern above. Please see [27e63a4](https://github.com/okta/okta-signin-widget/pull/594/commits/27e63a4abe99b398f1384dc6e680c994ee3df87b)

src/util/BaseLoginController.js
jmelberg-okta: Yep - this is a known problem (or feature), with these types of events. I think it makes sense to let the current execution flow continue (callback to error handler, throw error, etc), **and** emit the event. Moving forward with our blocking events, I'd expect us to no longer carry out the `globalError` behavior.

src/models/BaseLoginModel.js
```diff
@@ -41,6 +41,8 @@ function (Okta, Q, Enums) {
           }
           self.trigger('setTransactionError', err);
           self.trigger('error', self, err.xhr);
+          // Trigger an event once the previous 'error' event is handled
+          self.trigger('afterError', err);
```
jmelberg-okta: Once `self.trigger('error')` is called, the errorBox is rendered to the DOM. Since the `afterError` event is intended to give developers the ability to override the errorbox message, we need to trigger a special event here.

README.md
```diff
@@ -286,12 +286,8 @@ signIn.renderEl(
   },
 
   function error(err) {
-    // The widget will handle most types of errors - for example, if the user
-    // enters an invalid password or there are issues authenticating.
-    //
     // This function is invoked with errors the widget cannot recover from:
-    // 1. Known errors: CONFIG_ERROR, UNSUPPORTED_BROWSER_ERROR, OAUTH_ERROR
-    // 2. Uncaught exceptions
```
jmelberg-okta: This was too vague, as the widget fails in most uncaught exception cases. Removing this statement from our README as it doesn't apply.

src/models/BaseLoginModel.js
```diff
@@ -58,8 +58,8 @@ function (Okta, Q, Enums) {
               err.name === Enums.AUTH_STOP_POLL_INITIATION_ERROR) {
             return;
           }
-          self.trigger('setTransactionError', err);
           self.trigger('error', self, err.xhr);
+          self.trigger('setTransactionError', err);
```
jmelberg-okta: Changing this order grants us the ability to add additional listener logic to the `transactionError` event being emitted (see [`BaseLoginController`](https://github.com/okta/okta-signin-widget/pull/594/files#diff-e7366467f1d90e02ae84e1dab0578a09)), as we know that the error has been rendered to the page.

The `setTransactionError` and `error` events are not dependent on one another, so this change is safe.

src/util/BaseLoginRouter.js
```diff
@@ -250,6 +250,11 @@ function (Okta, BrowserFeatures, Settings,
                   }
                 });
                 this.appState.unset('flashError');
+                this.controller.trigger(
```
jmelberg-okta: This `flashError` happens only when the `stateToken` expires, and is used only in this place. However, this requires us to trigger another `afterError` event with the cached transaction error.

Therefore, if a user's state times out, we're going to fire this event two times (see [`LoginRouter_spec`](https://github.com/okta/okta-signin-widget/pull/594/files#diff-1fa99203134b9819901ba981b75f4af7)).

README.md
haishengwu-okta: 1. did you figure out what are `uncaught exceptions` or there are actually no others?
2. how about error in RegistrationController where invokes the globalError callback?

src/util/BaseLoginRouter.js
haishengwu-okta: why would trigger `afterError` twice at this case sounds fine?
I wonder if we shall only trigger `afterError` once.?

README.md
jmelberg-okta: > did you figure out what are uncaught exceptions or there are actually no others?
I wan't able to discover any occurrences where an unknown error would slip through and need to be handled via the `globalError`. We're explicit in most code execution paths that we're going to either:
a) Let the Widget handle the error (trigger an `error` event on the model/controller)
b) Use the `callGlobalError` method when we see a `Registration`, `OAuth`, `Config` and `UNSUPPORTED_BROWSER` errors.

Regarding registration, I'll add that to the list of known errors above. Sound good?

haishengwu-okta:
as stretch goal, we may want to extra the error handler in `RouterUtil.routeAfterAuthStatusChange` to another function for code decouple.

src/models/BaseLoginModel.js
haishengwu-okta: hmm, the only soft dependency I'm seeing is when INVALID_TOKEN_EXCEPTION

before:
- navigate back to primary auth hence `model.trigger('error')` actually wont happen

after
- model.trigger('error') hence form renders error banner
- navigate back to primary auth

sounds correct? though it may not be big concern.

src/util/BaseLoginController.js
```diff
@@ -55,6 +55,13 @@ define(['okta', 'q'], function (Okta, Q) {
       }, this);
       var transactionErrorHandler = _.bind(function (err) {
         this.options.appState.set('transactionError', err);
+        if (!err.statusCode) {
```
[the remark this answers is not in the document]
jmelberg-okta: Good point - I'll verify that if we get back a CORS error, we'll immediately return.

src/models/BaseLoginModel.js
jmelberg-okta: After writing a few more test cases for this, it turns out that there are some cases where we're creating two banners after reordering these. I'll go back to the drawing board 😞

haishengwu-okta: why/how it renders two banner after your reordering?

jmelberg-okta: There was an extra `trigger('error')` being called in a failure handler. I'll leave a comment on my next commit with more details.

mauriciocastillosilva-okta: will look at the future comment, this is an interesting topic. Looking at `RouterUtil` I see the same as Haisheng was point out, and seems that `flashError` is the special handling for it... so we might need to do something there as well.

src/util/RouterUtil.js
```diff
@@ -79,6 +79,18 @@ function (Okta, OAuth2Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
...
+    if (err) {
+      if (!err.statusCode) {
+        // Bring the statusCode to the top-level of the Error
+        err.statusCode = err.xhr && err.xhr.status;
+      }
+      // Some controllers return the className as a function - process it here:
+      var className = typeof router.controller.className === 'function'
+        ? router.controller.className()
+        : router.controller.className;
+      router.controller.trigger('afterError', { controller: className }, err)
```
mauriciocastillosilva-okta: how will this work for INVALID_TOKEN_EXCEPTION? we trigger this event, but right after they handle it, we navigate to a different page... maybe when INVALID_TOKEN_EXCEPTION is the error, we need to trigger `afterError` in the same code block where we handle `flashError`?
And then this logic can go below the special INVALID_TOKEN_EXCEPTION handler

jmelberg-okta: Yep, that is one approach - and I touched on that topic [here](https://github.com/okta/okta-signin-widget/pull/594#discussion_r237955733).

Users can enter an `INVALID_TOKEN_EXCEPTION ` state on any view/controller, where the proposed behavior is:

1. Trigger an `afterError` event on the controller the user is currently on
2. Set the current `flashError` state so we can use this later
    - Note: In the future, we can change how this works. By using the `afterError` event ourselves, we can handle this without needing to store the state and navigate.
3. Once navigating back to primary auth, trigger a new `error` event to build the banner with the expired token text.
   - Note: This message can be customized via `i18n` properties

To me, it seems that firing this event **where** it occurs is the right answer, as opposed to firing a new `afterError` event once we navigate back to primary auth.

What do you think?

README.md
```diff
@@ -285,12 +286,8 @@ signIn.renderEl(
...
+    // Known errors: CONFIG_ERROR, UNSUPPORTED_BROWSER_ERROR, OAUTH_ERROR, REGISTRATION_FAILED
```
haishengwu-okta: @nikhilvenkatraman-okta 
Would you add more colors to why we invoke globalError handler when registration failed?
Given this is not doc'ed, I wonder if it's safe to refactor to use `afterError` event?

src/util/RouterUtil.js
haishengwu-okta: 1. feels like we shall remove logic to line 103
2. and only trigger `afterError` for flashError when it's rendered. (after navigated back to primary)
3. since you already touch this `routeAfterAuthStatusChange`, why not let's extra the error handling logic into separated dedicated method

mauriciocastillosilva-okta: I don't know if we need to remove 103, but I think this `if` block should go after handling INVALID_TOKEN_EXCEPTION... and that will go well with our description of `afterError`, since we will be triggering it AFTER the widget handled it (which in this case it means navigating to a different page)

haishengwu-okta: sorry I meant `moved` not `remove`..

README.md
nikhilvenkatraman-okta: @haishengwu-okta `REGISTRATION_FAILED` is invoked when the regCallbacks fails. We have `parseSchema `,`preSubmit ` and `postSubmit` hooks for registration. We invoke the global error because at this point the user registration fails and they cant proceed further. Looking at the readme for `afterError ` event, I don't see a problem with refactoring the regCallbacks to use `afterError ` event.

src/util/RouterUtil.js
mauriciocastillosilva-okta: yeah, I think we are saying the same thing 👍 
Does that make sense @jmelberg-okta ?

jmelberg-okta: Got it @mauriciocastillosilva-okta and @haishengwu-okta! Good advice. I'll make that change now.

jmelberg-okta: Fixed with [`bfeae8d`](https://github.com/okta/okta-signin-widget/pull/594/commits/bfeae8dbf0950a79d3d5345e4e55ceb24d4c8f68)

src/util/RouterUtil.js
```diff
@@ -79,28 +79,30 @@ function (Okta, OAuth2Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
...
-      router.controller.trigger('afterError', { controller: className }, err)
+    if (err && !err.statusCode) {
+      // Bring the statusCode to the top-level of the Error
+      err.statusCode = err.xhr && err.xhr.status;
     }
 
     // Token has expired - no longer valid. Navigate back to primary auth.
     if (err && err.errorCode === ErrorCodes.INVALID_TOKEN_EXCEPTION) {
-      router.appState.set('flashError', Okta.loc('error.expired.session'));
+      router.appState.set('flashError', err);
```
jmelberg-okta: We need to pass in the `err` here instead of the custom message, in order for us to trigger the correct `afterError` event.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -476,7 +476,7 @@ function (Okta, Q, Logger, OktaAuth, Util, Expect, Router,
           expect(test.afterErrorHandler).toHaveBeenCalledTimes(1);
           expect(test.afterErrorHandler.calls.allArgs()).toEqual([
             [
-              { controller: 'recovery-question' },
+              { controller: 'primary-auth' },
```
jmelberg-okta: Now, we're going to always fire the `afterError` event from the primary-auth controller, as users are navigated back to this page when the `flashError` is rendered.

test/e2e/util/util.js
```diff
@@ -13,6 +13,12 @@
 /* global module, browser */
 var util = module.exports = {};
 
+var EC = protractor.ExpectedConditions;
+
 util.loadTestPage = function (pageName) {
   browser.get('http://localhost:3000/' + pageName + '.html');
 };
+
+util.waitForElement = function (el) {
```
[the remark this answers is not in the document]
jmelberg-okta: Easy to miss due to the number of changed files, but we actually use it inside of [`basic_spec`](https://github.com/okta/okta-signin-widget/pull/594/files#diff-93e23e71a6d3f5fc58c65c5dff5ae5c2R69) to know when the error banner has been rendered.

test/unit/spec/EnrollCall_spec.js
```diff
@@ -263,6 +265,22 @@ function (Q, Okta, OktaAuth, LoginUtil, Util, Form, Beacon, Expect, $sandbox,
...
       });
+      itp('triggers an afterError event and does not go to next step if error response', function () {
+        return setupAndSendInvalidCode().then(function (test) {
+          expectCallButton(test);
+          expect(test.afterRenderHandler).toHaveBeenCalledTimes(1);
+          expect(test.afterRenderHandler.calls.allArgs()[0]).toEqual([
+            {
+              controller: 'enroll-call'
+            },
+            jasmine.objectContaining({
```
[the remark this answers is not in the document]
jmelberg-okta: Because we're returning an `Error` type, we need to assert properties on the object via `objectContaining`. Exact matches unfortunately do not work with error types.

Alternatively, we can import in all known error types and then attempt to match, but we're still going to need to assert our known values the same way we're currently doing so.

```javascript
const errorContext = test.afterErrorHandler.calls.allArgs()[0][1]);
expect(errorContext instanceOf SomeError);
expect(errorContext.name).toBe('SomeError');
```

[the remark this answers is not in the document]
jmelberg-okta: @haishengwu-okta: Great points! I should have communicated this earlier.

1. Most errors we're capturing have XHR values. This is the response value via the API, but isn't always available. As an example, some OAuth error will be captured via the URL instead of an XHR request.
   - Another finding (and recorded [here](https://oktainc.atlassian.net/browse/OKTA-200208)), some of our tests aren't using the same behavior as the actual source code. In many places, we're using [`transformErrorXhr`](https://github.com/okta/okta-signin-widget/blob/4e0c5d2823eb6a283f8e6f216eabf7e6eeba1eca/src/util/Util.js#L23-L55) to extract an `errorCause` from the response, and use it for the `errorSummary`. When attempting to assert the `xhr` in some test files, we're going to need to refactor some of the assertions.
2. Correct. OAuth errors within AuthJS and inherently the Widget do not have `xhr` fields. I've noted this property as *optional* in our [README](https://github.com/okta/okta-signin-widget/pull/594/files#diff-04c6e90faac2675aa89e2176d2eec7d8R1345)
3. Yep. I think having all errors be extensions of the `Error` type is the right call. If a developer wants to know which line/file the error occurs, they can always output and parse the `Error.stack` field.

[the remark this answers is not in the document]
haishengwu-okta: 1. I guess we shall assert `xhr` if it does exists. e.g in EnrollU2f_spec, it does have `xhr`.
2. It has more fields than we doc'ed (name, message, statusCode, xhr). I'm hesitate we shall populate extra fields than what we doc'ed
3. I doubt developers need the `Error.stack`. When I say "developers" I meant who use pre-built signin-widget and I cant think a reason that they want to know which line/file the error occurs given the `name/message/statusCode/xhr` shall already give sufficient information. 
(I'm pick at we shall do what we said. hah)

thoughts?

src/util/Util.js
```diff
@@ -103,13 +103,14 @@ define(['okta', 'util/Logger'], function (Okta, Logger) {
 
   // Trigger an afterError event
   Util.triggerAfterError = function (controller, err) {
-    if (!err.statusCode) {
+    if (err.xhr && err.xhr.status) {
```
jmelberg-okta: Changing this logic to only include the `statusCode` when present, instead of possibly setting this value to `undefined`.

test/unit/helpers/mocks/Util.js
```diff
@@ -313,6 +314,12 @@ function (Okta, Q, Duo) {
     return responseCopy;
   };
 
+  fn.transformMockXHR = function (xhr) {
+    // The xhr needs to be transformed without modifying the existing response
+    var xhrClone = JSON.parse(JSON.stringify(xhr));
+    return LoginUtil.transformErrorXHR(xhrClone);
+  }
```
jmelberg-okta: In order to validate that our `xhr` objects contain the expected information, we need to do two things:
1. Run it through the [`transformErrorXHR`](https://github.com/okta/okta-signin-widget/blob/master/src/widget/OktaSignIn.js#L215) method, similar to what the Widget does by default. This maps the request object correctly.
2. Clone the `xhr` by using the `JSON.parse` and `JSON.stringify` hack to avoid overriding the original value. Without this, the API response we're taking from `helpers/xhr/*` will be modified and reused throughout our tests.

I was unable to find a better way to do this, so I'm open to suggestions.

src/util/Util.js
```diff
@@ -103,13 +103,14 @@ define(['okta', 'util/Logger'], function (Okta, Logger) {
...
       // Bring the statusCode to the top-level of the Error
       err.statusCode = err.xhr && err.xhr.status;
```
haishengwu-okta: - what if `err.statusCode` is already exists, will it be override unexpectedly or not?
- probably just do `statusCode = err.xhr.status`

src/util/Util.js
```diff
@@ -103,13 +103,14 @@ define(['okta', 'util/Logger'], function (Okta, Logger) {
 
   // Trigger an afterError event
   Util.triggerAfterError = function (controller, err) {
```
haishengwu-okta: nit:
make sure `err` is not null: `(controller, err = {})`

src/util/Util.js
jmelberg-okta: Good question. `statusCode` isn't a top-level key in any of our errors, so we do not need to worry about an overwrite.

Will update line `108` to copy over `err.xhr.status` since we're checking above for the presence of `xhr`.

test/unit/helpers/mocks/Util.js
haishengwu-okta: I usually prefer to assert with plain object (if it's not too verbose at this case)
Pros is if any mistakes / changes in `transformErrorXHR`, test would be able to catch. (an extreme example is if `transformErrorXHR` return `{}`, test would still pass but apparently that's not the right implementation)
Const is of course is very verbose to type of the actual `xhr` value in test.

README.md
haishengwu-okta: shall we briefly mention that other errors will pop up through `afterEvent`?

test/e2e/util/util.js
```diff
@@ -13,6 +13,12 @@
...
+  browser.wait(EC.presenceOf(el), 1000);
```
haishengwu-okta: Can we do 'wait for something to be visible' rather fighting with time.
(I thought I added at first place but apparently I forgot ...)

test/e2e/specs/basic_spec.js
```diff
@@ -57,4 +59,14 @@ describe('Basic flows', function () {
...
 
+  it('modifies the error banner using an afterError event', function () {
+    // Ensure the widget exists
+    var el = element(by.css('#okta-sign-in'));
+    expect(el.isDisplayed()).toBe(true);
+
+    primaryAuth.loginToForm('foo', 'bar');
+    var errorBox = element(by.className('okta-form-infobox-error'));
+    util.waitForElement(errorBox);
+    expect(errorBox.getText()).toBe('Custom Error!');
```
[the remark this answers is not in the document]
jmelberg-okta: Might be overkill due to the other tests. But I've updated this function to remove the unused `context` and `error` vars to abide by lint rules

test/unit/spec/MfaVerify_spec.js
```diff
@@ -482,7 +482,7 @@ function (Okta,
           name: 'AuthApiError',
           message: message,
           statusCode: code,
-          xhr: Util.transformMockXHR(resError)
+          xhr: resError
```
haishengwu-okta: nit:
shall we follow same pattern in other tests that write out the actual error?

jmelberg-okta: Yes! My sleep deprivation caused me to skip over a few. I'll update that now. Thanks!

jmelberg-okta: Alright, I guess I wasn't too crazy. We're passing in the error object when calling this function `expectError` or `expectErrorEvent`.

Here are the usages:

- [Google TOTP](https://github.com/okta/okta-signin-widget/pull/594/files#diff-4b2ed50d042dededa6f2af06dc054199R941)
- [RSA TOTP](https://github.com/okta/okta-signin-widget/pull/594/files#diff-4b2ed50d042dededa6f2af06dc054199R1023)
- [SMS](https://github.com/okta/okta-signin-widget/pull/594/files#diff-4b2ed50d042dededa6f2af06dc054199R1441)
- [Call](https://github.com/okta/okta-signin-widget/pull/594/files#diff-4b2ed50d042dededa6f2af06dc054199R1717)
- [Email](https://github.com/okta/okta-signin-widget/pull/594/files#diff-4b2ed50d042dededa6f2af06dc054199R2018)
- [Custom SAML Factor](https://github.com/okta/okta-signin-widget/pull/594/files#diff-4b2ed50d042dededa6f2af06dc054199R3234)
- [Custom OIDC Factor](https://github.com/okta/okta-signin-widget/pull/594/files#diff-4b2ed50d042dededa6f2af06dc054199R3319)
- [Password](https://github.com/okta/okta-signin-widget/pull/594/files#diff-4b2ed50d042dededa6f2af06dc054199R3452)

README.md
jmelberg-okta: We already describe that [below](https://github.com/okta/okta-signin-widget/pull/594/files#diff-04c6e90faac2675aa89e2176d2eec7d8R1337). I've changed this text to do a better job of calling this out.

Also, I've created a [follow-up task](https://github.com/okta/okta.github.io/issues/2527) in our "guides" to showcase this new behavior.

PR #596, opened 2018-11-29
:art: Remove jquery.simplemodal source code.

haishengwu-okta:


haishengwu-okta:
@jmelberg-okta 
regarding the css, turns out those css are copying from courage.
probably want leave it for now until we refactor the css so that it's easier to do comparison with the one from courage.

PR #598, opened 2018-11-29
WIP-v1: internalize jQuery

haishengwu-okta:
- Bundle jQuery within courage-for-signin-widget hence qtip and backbone which depend on jQuery
- Split out ajax related logic from `Bundles.js` to LanguageUtil because of cycle dependency (Okta --> Bundles --cannot--> Okta(`.$`))
- `okta-auth-js/jquery` wouldn't work because there is not more jquery in dependency tree. Created a customized auth-js version. (might be port back to auth-js)


haishengwu-okta:
close since we probably wont take this approach.

PR #599, opened 2018-11-29
:bug: Don't copy webpack bundle analyzer result to dist

haishengwu-okta:


Gruntfile.js
```diff
@@ -75,7 +75,7 @@ module.exports = function (grunt) {
           {
             expand: true,
             cwd: JS,
-            src: ['okta-sign-in*'],
+            src: ['okta-sign-in*', '!*.html'],
```
jmelberg-okta: Would it be better if we called out this file by name instead of using a wildcard?

manueltanzi-okta: Imho should be fine since we want only js files and this one exclude the multiple html generated files. An even better one could be exclude everything without js extension.

haishengwu-okta: Actually my first attempt is `['okta-sign-in*.js', 'okta-sign-in*.js.map']` but feels like `'!*.html'` seems simpler. :)

PR #600, opened 2018-11-29
:bug: Fixes the webpack3 config naming for module.rules

haishengwu-okta:
- After upgrade to webpack3, the naming for module loaders is `module.rules`.

### DIFF
--- a/webpack.release.config.js
+++ b/webpack.release.config.js
@@ -39,14 +39,6 @@ entryConfig.externals = {
   'u2f-api-polyfill': true,
   'underscore': true
 };
-// Add transform-runtime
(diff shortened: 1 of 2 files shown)


webpack.release.config.js
```diff
@@ -39,14 +39,6 @@ entryConfig.externals = {
   'u2f-api-polyfill': true,
   'underscore': true
 };
-// Add transform-runtime
-entryConfig.module.loaders = entryConfig.module.loaders || [];
```
haishengwu-okta: This was added by https://github.com/okta/okta-signin-widget/pull/353 and turns out `transform-runtime` was added twice since it's already added in `webpack.common.config.js`

PR #604, opened 2018-12-05
Use form submit for redirects

nbarbettini:
This PR:

* Replaces `window.location` with a dynamic form submit for a few redirect calls within the critical login path
* Makes it clearer in code where the Widget modifies `window.location`
* Adds tests

Needs to be reviewed, and also squashed!

### DIFF
--- a/src/util/RedirectUtil.js
+++ b/src/util/RedirectUtil.js
@@ -1,0 +1,61 @@
+/*!
+ * Copyright (c) 2015-2018, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define(function () {
+  var fn = {};
+
+  /**
+   * Redirects the browser to a new URL by directly setting window.location.
+   */
+  fn.setWindowLocationTo = function (url) {
+    window.location = url;
+  };
+
+  /**
+   * Redirects the browser to a new URL without using window.location.
+   */
+  fn.redirectTo = function (url) {
+    var form = this.buildDynamicForm(url);
+    document.body.appendChild(form);
+    form.submit();
+  };
+
+  fn.buildDynamicForm = function (url) {
+    var form = document.createElement('form');
+    form.method = 'get';
+    form.setAttribute('style', 'display: none;');
+
+    var chunks = (url || '').split('?');
+    var targetUrl = chunks[0];
+    var query = chunks[1];
+
+    form.action = targetUrl;
+
+    if (query && query.length) {
+      var queryParts = query.split('&');
+
+      for (var i = 0; i < queryParts.length; i++) {
+        var input = document.createElement('input');
+        var parameterParts = queryParts[i].split('=');
+
+        input.name = parameterParts[0];
+        input.value = decodeURIComponent(parameterParts[1]);
+        input.type = 'hidden';
+        form.appendChild(input);    
+      }
+    }
+
+    return form;
+  };
+
+  return fn;
+});
--- a/src/util/RedirectUtil.js
+++ b/src/util/RedirectUtil.js
@@ -30,14 +30,13 @@ define(function () {
   };
 
   fn.buildDynamicForm = function (url) {
-    var form = document.createElement('form');
-    form.method = 'get';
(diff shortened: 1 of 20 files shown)


src/util/RedirectUtil.js
```diff
@@ -0,0 +1,61 @@
...
+  };
+
+  fn.buildDynamicForm = function (url) {
+    var form = document.createElement('form');
+    form.method = 'get';
+    form.setAttribute('style', 'display: none;');
+
+    var chunks = (url || '').split('?');
+    var targetUrl = chunks[0];
+    var query = chunks[1];
```
haishengwu-okta: nit:
generates url before form creation.

src/util/RedirectUtil.js
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define(function () {
+  var fn = {};
+
+  /**
+   * Redirects the browser to a new URL by directly setting window.location.
+   */
+  fn.setWindowLocationTo = function (url) {
```
haishengwu-okta: - Prefer to keep naming convention `redirectTo` for window location change redirection because it's more common  perspective IMHO.
- And name the other more explicitly like `redirectToWithFormGet`

test/unit/spec/RedirectUtil_spec.js
```diff
@@ -0,0 +1,101 @@
...
+        document.body.appendChild(frame);
+
+        frame.contentDocument.open();
+        frame.contentDocument.write("<html><body></body></html>");
+        frame.contentDocument.close();
+        
+        return frame;
+      }
+
+      var wait = 25;
```
haishengwu-okta: why needs wait?

src/util/RedirectUtil.js
```diff
@@ -0,0 +1,61 @@
...
+    window.location = url;
+  };
+
+  /**
+   * Redirects the browser to a new URL without using window.location.
+   */
+  fn.redirectTo = function (url) {
+    var form = this.buildDynamicForm(url);
+    document.body.appendChild(form);
```
haishengwu-okta: We shall try to append form to widget main container rather than body.
I think we have general guide line that do not populate `body` as much as possible.

test/unit/spec/RedirectUtil_spec.js
nbarbettini: The navigation is not synchronous, it takes a few cycles for the iframe URL to change after `form.submit()`. In my testing 10ms was too low, 25ms seemed to be okay.

haishengwu-okta: Probably need to dig into more (maybe to mock the timer).
Fighting with time is always not a good practice at first place.

nbarbettini:
Rebased and responded to @haishengwu-okta's feedback.

src/util/RedirectUtil.js
nbarbettini: I don't know how to append to the Widget container from inside this method. If you do, feel free. 😄 

As is, the form is totally invisible and is only in the DOM for a few ms before `submi()` is called.

src/util/RedirectUtil.js
nbarbettini: Sure. I'll leave final naming up to you.

src/util/RouterUtil.js
```diff
@@ -146,11 +146,13 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
...
           setCookieAndRedirect: function (redirectUrl) {
-            CourageUtil.redirect(sessionCookieRedirectTpl({
+            var url = sessionCookieRedirectTpl({
               baseUrl: router.settings.get('baseUrl'),
               token: encodeURIComponent(res.sessionToken),
               redirectUrl: encodeURIComponent(redirectUrl)
-            }));
+            });
+
+            RedirectUtil.redirectTo(url);
```
rchild-okta: We should add a comment here and in the other usage on 140 describing why we're doing a form get here rather than a normal redirect (once we've renamed the methods to be more clear as per @haishengwu-okta's suggestion).

src/util/RedirectUtil.js
haishengwu-okta: basically pass `this.$el` from View/Controller to this redirect function.

src/util/RedirectUtil.js
```diff
@@ -0,0 +1,71 @@
...
+    var splitOnFragment = (url || '').split('#');
+    var fragment = splitOnFragment[1];
+    
+    var splitOnQuery = (splitOnFragment[0] || '').split('?');
+    var query = splitOnQuery[1];
+
+    var targetUrl = splitOnQuery[0];
+    if (fragment) {
+      targetUrl += '#' + fragment;
+    }
```
gowthamidommety-okta: nit: Would it make sense to pull lines 33-42 out into a separate function that returns targetUrl and query?

haishengwu-okta:
note:
Find design doc at wiki titled `Widget+changes+to+fix+Android+Chrome+redirect+bug`

PR #605, opened 2018-12-05
Updates README and CHANGELOG

jmelberg-okta:
### Description

README and CHANGELOG updates to be merged upon the release of `@okta/okta-signin-widget@2.15.0`.

PR #606, opened 2018-12-08
:seedling: Add colors.brand to config

manueltanzi-okta:
### Description

- Added `colors.brand` configuration for developers to control color of primary button via js config.
- Format supported 6 characters hex (ex.: "#008000").
- TODO: update README.md to expose this new functionality.

### Video

config_colors_brand

### Reviewers

@haishengwu-okta @jmelberg-okta @robertjd @nbarbettini


src/util/ColorsUtil.js
```diff
@@ -0,0 +1,63 @@
...
+    var css = template(colors);
+    var head = document.head || document.getElementsByTagName('head')[0];
+    var style = document.createElement('style');
+
+    style.id = 'config-colors';
+    style.type = 'text/css';
+
+    if (style.styleSheet) {
+      // IE8 and below.
+      style.styleSheet.cssText = css;
```
jmelberg-okta: nit: We can remove this conditional check, as we no longer support IE8.

src/util/ColorsUtil.js
```diff
@@ -0,0 +1,63 @@
...
+      var c = parseInt(hex.substr(i * 2, 2), 16);
+      c = Math.round(Math.min(Math.max(0, c + (c * lum)), 255)).toString(16);
+      newHex += ('00' + c).substr(c.length);
+    }
+    return newHex;
+  };
+
+  var template = function (colors) {
+    return `
+      #okta-sign-in.auth-container .button-primary,
```
jmelberg-okta: Use of multi-line template literals oftentimes causes unexpected indentations.

Can we bring the indentation to the far left, so the return object will be:

```css
#okta-sign-in.auth-container .button-primary,
#okta-sign-in.auth-container .button-primary:active,
#okta-sign-in.auth-container .button-primary:focus { background: abc123; }
#okta-sign-in.auth-container .button-primary:hover { background: 0.05; }
#okta-sign-in.auth-container .button.button-primary.link-button-disabled {
  background: abc123;
  opacity: 0.5;
}
```

Instead of: 
```css
      #okta-sign-in.auth-container .button-primary,
      #okta-sign-in.auth-container .button-primary:active,
      #okta-sign-in.auth-container .button-primary:focus { background: abc123; }
      #okta-sign-in.auth-container .button-primary:hover { background: 0.05; }
      #okta-sign-in.auth-container .button.button-primary.link-button-disabled {
        background: abc123;
        opacity: 0.5;
      }
```

src/util/BaseLoginRouter.js
```diff
@@ -206,6 +207,13 @@ function (Okta, BrowserFeatures, Settings,
           .done();
       }
 
+      if (this.settings.get('colors.brand') && !ColorsUtil.isLoaded()) {
```
jmelberg-okta: Should we add a comment here stating that the intended use of this functionality is that it **only runs once**?

src/util/ColorsUtil.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    head.appendChild(style);
+  };
+
+  fn.isLoaded = function () {
+    return !!document.getElementById('config-colors');
+  };
+
+  return fn;
```
jmelberg-okta: There are cases where developers are calling `signIn.remove()` and then regenerating the widget with `renderEl` conditionally. I believe this use case will have the side-effect of creating multiple style scripts, right?

I'd suggest adding a cleanup method that gets triggered on [`signIn.remove`](https://github.com/okta/okta-signin-widget/blob/master/src/util/BaseLoginRouter.js#L300-L305).

test/e2e/templates/basic.tpl
```diff
@@ -1,6 +1,9 @@
 {{#> cdnLayout}}
 var options = {
-  'baseUrl': '{{{WIDGET_TEST_SERVER}}}'
+  'baseUrl': '{{{WIDGET_TEST_SERVER}}}',
+  'colors': {
```
jmelberg-okta: Since this is an optional config param, I'd suggest **not** including it in the template.

What do you think about using [this type of helper method](https://github.com/okta/okta-signin-widget/blob/master/test/e2e/specs/basic_spec.js#L45-L56) instead?

src/models/Settings.js
```diff
@@ -156,7 +156,10 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
       'consent.cancel': 'function',
 
       //IDP Discovery
-      'idpDiscovery.requestContext': 'string'
+      'idpDiscovery.requestContext': 'string',
+
+      //Colors
+      'colors.brand': 'string'
```
haishengwu-okta: add readme and the color format

src/util/BaseLoginRouter.js
haishengwu-okta: probably could move `ColorsUtil.isLoaded` check to `ColorsUtil.addStyle` where is smart enough to add style or not base on whether it's already loaded.

src/util/ColorsUtil.js
haishengwu-okta: Looked at `@mixin button-template` and turns out it defines corresponding text color and border color.
Are we comfortable to omit both? Maybe Edbury could provide more help.

src/util/ColorsUtil.js
```diff
@@ -0,0 +1,63 @@
...
+      #okta-sign-in.auth-container .button.button-primary.link-button-disabled {
+        background: ${colors.brand};
+        opacity: 0.5;
+      }
+    `;
+  };
+
+  fn.addStyle = function (colors) {
+    var css = template(colors);
+    var head = document.head || document.getElementsByTagName('head')[0];
```
haishengwu-okta: In general prefer to add the element inside widget main container to avoid populate global document.
Did you get chance to verify if it works?

src/util/ColorsUtil.js
manueltanzi-okta: Will not create multiple styles since `isLoaded` would still return `true` but will not update with the new config if they pass a new color. Good catch!

test/e2e/templates/basic.tpl
manueltanzi-okta: Make sense 👍

src/util/ColorsUtil.js
manueltanzi-okta: I could add style for the border-color since in the `@mixin` it depends from primary button background color. 
However, the text color does not depend on the primary button background color, so I am not sure if we should include also that in this PR, since we may add a variable to control that later.
If we really want to generate also the primary button text color here, we could computationally generate a text color that has a good contrast with the primary button background color passed in config.

@nbarbettini @edburyenegren-okta opinions?

src/util/BaseLoginRouter.js
manueltanzi-okta: The reason I put it there is to be consistent with the similar pattern we use for i18n Bundle (https://github.com/okta/okta-signin-widget/blob/646bfaf7f467cdfea5cee3828a3fe2f19721aa0c/src/util/BaseLoginRouter.js#L199). If we are ok breaking this pattern, I can move the check to the `ColorsUtil.addStyle` function.

src/util/ColorsUtil.js
manueltanzi-okta: Make sense, will move inside widget main container 👍

src/models/Settings.js
manueltanzi-okta: Will add the format 👍 
For the README, we want to update when the new version that includes the feature is published.
CC: @jmelberg-okta

jmelberg-okta: Yep. Because we do not have a set release cadence, we've decided to separate README changes from the feature work. As an example, you can see the `afterEvents` documentation picked up with the `2.15.0` CHANGELOG PR (#605).

src/util/ColorsUtil.js
manueltanzi-okta: Yes, now that I moved it to be inside the widget yes, so this should not be a concern anymore. @jmelberg-okta was referring to first commit when style was appended to `body`.

src/models/Settings.js
jmelberg-okta: Does our model validation fail if you pass in `rgb(255,0,0)`? If not, maybe we check to see if the color is prepended with a `#`.

test/unit/spec/ColorsUtil_spec.js
```diff
@@ -0,0 +1,56 @@
...
+      });
+      it('returns true if config-color style was added', function () {
+        var colors = {
+          brand: '#008000'
+        };
+        ColorsUtil.addStyle(colors);
+        expect(ColorsUtil.isLoaded()).toBe(true);
+      });
+    });
+  });
```
jmelberg-okta: Lets add tests for the following cases too:

1. Using `rgb` decimal instead of a hex value will result in a thrown configuration error
2. Using a color keyword (`red`) will throw a configuration error

src/models/Settings.js
manueltanzi-okta: Will add model validation 👍

src/util/ColorsUtil.js
nbarbettini: We should do 2 things for a good developer experience:

1. For developers who want advanced customization of Widget look & feel, make it possible to modify the primary button text color as a Sass variable. @manueltanzi-okta can you create an issue to track this?

2. For developers who just want a quick & easy customization (and are not editing Sass), automatically make the primary button text color darker if the `colors.brand` color is very light.

manueltanzi-okta: Created JIRA for SASS change: https://oktainc.atlassian.net/browse/OKTA-202207
Will add computed text color in this PR.

manueltanzi-okta: As per discussion in SIW design meeting, we are going to skip the text color for now.

src/util/Enums.js
```diff
@@ -12,6 +12,8 @@
 
 define({
 
+  WIDGET_CONTAINER_ID: 'okta-sign-in',
```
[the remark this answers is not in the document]
manueltanzi-okta: Yepp! However @haishengwu-okta already prepared that change in https://github.com/okta/okta-signin-widget/pull/613/files , this is why I haven't changed here.

manueltanzi-okta:
Updated README.md and CHANGELOG.md in https://github.com/okta/okta-signin-widget/pull/605/commits/0f92bba6cf09751c110afdd9227995ee9e9a93ab

src/util/ColorsUtil.js
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define(['util/Enums'], function (Enums) {
+  var fn = {};
+
+  var validate = function(colors) {
```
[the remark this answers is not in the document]
nbarbettini: Does eslint pick up stuff like this?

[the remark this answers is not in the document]
manueltanzi-okta: Yepp, haven't run the linter locally for that commit, however travis was correctly failing

src/util/ColorsUtil.js
```diff
@@ -0,0 +1,67 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define(['util/Enums'], function (Enums) {
+  var fn = {};
+
+  var validate = function (colors) {
+    if (!colors.brand.match(/^#[0-9A-Fa-f]{6}$/)) {
```
haishengwu-okta: shall we follow same pattern like [this](https://github.com/okta/okta-signin-widget/blob/master/src/models/Settings.js#L320-L322) to fail earlier?

src/util/ColorsUtil.js
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  fn.addStyle = function (colors) {
+    validate(colors);
+
+    var css = template(colors);
+    var main = document.getElementById(Enums.WIDGET_CONTAINER_ID);
+    var style = document.createElement('style');
+
+    style.id = 'config-colors';
```
haishengwu-okta: nit:
maybe: `okta-signin-widget-config-colors`

src/util/ColorsUtil.js
```diff
@@ -0,0 +1,67 @@
...
+  background: ${colors.brand};
+  opacity: 0.5;
+}
+    `;
+  };
+
+  // visible for testing
+  fn.lighten = function (hex, lum) {
+    lum = lum || 0;
+    hex = String(hex).replace(/[^0-9a-f]/gi, '');
```
haishengwu-okta: nit:
I guess we could do substring since only `hex` is always hex string.

test/unit/spec/ColorsUtil_spec.js
```diff
@@ -0,0 +1,86 @@
...
+  'util/Enums'
+],
+function (Okta, Expect, ColorsUtil, Enums) {
+  var $ = Okta.$;
+
+  var normalize = function (text) {
+    return text.replace(/\s+/g, ' ');
+  };
+
+  Expect.describe('ColorsUtil', function () {
```
haishengwu-okta: I doubt we need `Expect.describe` but simple `describe` would work best.
(looks like it has already inconsistent usage)

test/unit/spec/ColorsUtil_spec.js
```diff
@@ -0,0 +1,86 @@
...
+          brand: '#008000'
+        };
+        ColorsUtil.addStyle(colors);
+        expect(ColorsUtil.isLoaded()).toBe(true);
+      });
+    });
+
+    Expect.describe('lighten', function () {
+      it('returns a lighter color if a lum parameter is passed', function () {
+        expect(ColorsUtil.lighten('#008000', 0.3)).toBe('#00a600');
```
haishengwu-okta: maybe test other colors (e.g. 080000, 0000EE, C0C0C0) for diversity.

src/models/Settings.js
```diff
@@ -323,6 +323,9 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
       if (!options.baseUrl) {
         this.callGlobalError(new ConfigError(Okta.loc('error.required.baseUrl')));
       }
+      else if (options.colors && options.colors.brand && !options.colors.brand.match(/^#[0-9A-Fa-f]{6}$/)) {
+        this.callGlobalError(new ConfigError('colors.brand has to be in 6 digits Hex format'));
```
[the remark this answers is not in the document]
manueltanzi-okta: FYI, the reason I haven't localized is because this message is developer facing and not enduser facing.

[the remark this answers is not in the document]
manueltanzi-okta: I can add to be consistent, but not so useful since nothing else for developers is localized.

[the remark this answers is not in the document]
mauriciocastillosilva-okta: We should be consistent... in summary I think if the consumer (developer or not) adds an i18n supported lang to the bootstrap config, we should return all messages in that language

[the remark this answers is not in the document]
manueltanzi-okta: Added to i18n 👍

PR #609, opened 2018-12-12
Adds MFA_VERIFY errors to afterError event

jmelberg-okta:
### Description

Triggers an `afterError` event on specific MFA error responses to handle the following end-user scenarios:

1. User fails **to respond** to the push notification
2. User **rejects** the push notification

> Note: This is needed for [OKTA-198042](https://oktainc.atlassian.net/browse/OKTA-198042)

This feature will be consumed immediately by the login page. [WIP here](https://github.com/okta/okta-core/pull/31368).

### Resolves

- [OKTA-198465](https://oktainc.atlassian.net/browse/OKTA-198465)


src/MfaVerifyController.js
```diff
@@ -110,6 +110,11 @@ function (Okta, BaseLoginController, TOTPForm, YubikeyForm, SecurityQuestionForm
         this.listenTo(this.model, 'change:rememberDevice', function (model, rememberDevice) {
           model.get('backupFactor').set('rememberDevice', rememberDevice);
         });
+
+        // Listen to specific errors and bubble them up
+        this.listenTo(this.model, 'errors:verify', function (model, err) {
```
jmelberg-okta: Using `errors:verify` as the event name to be consistent with other [Backbone events](http://backbonejs.org/#Events-catalog). Open to changing this internal event if needed.

haishengwu-okta: remind me why we cant listen to `this.model # error` event transform as `afterError` ?

jmelberg-okta: Good question. We can - however we'd need to validate the return value for a model `error` event in the process. This is just to be more explicit.

For example, to listen to the errors we need:

```javascript
this.listenTo(this.model, 'error', function (model, err) {
  if (!model || !err) { return }

  // We emit an 'error' event for other cases, so we
  // need to check for the error type before emitting
  if (err.name === 'MFA_VERIFY_ERROR') {
    this.trigger(
      'afterError',
      { controller: this.className },
      {
        name: 'MFA_VERIFY_ERROR',
        message: err.responseJSON.errorSummary
      }
    );
  }
});
```

haishengwu-okta: hmm, I guess the pro is we don't need to maintain specific error event. what's your thoughts?

src/MfaVerifyController.js
```diff
@@ -110,6 +110,11 @@ function (Okta, BaseLoginController, TOTPForm, YubikeyForm, SecurityQuestionForm
...
+          this.trigger('afterError', { controller: this.className }, err);
```
haishengwu-okta: feels like we shall build abstract layer (util function) for `this.trigger('afterError',...` to be a util function so that we only need to maintain the data structure for `afterError` at one place.

jmelberg-okta: We do! We have `Util.triggerAfterError` that will account for some of the defaults. I'll make sure this gets updated.

src/MfaVerifyController.js
haishengwu-okta: - we wont need `once` and `bind` at this case, so I'm not sure how that problem is related.
- IMHO, 'event' system add complexity and makes debug harder. and now it's going to have two events when error happened which I thought makes things worse.

jmelberg-okta: Pushed another commit which uses the existing `error` event.

It does add some more complexity and requires us to create a new Error type, but I'm assuming that is a good pattern to follow moving forward.

Let me know what you think!

src/views/mfa-verify/PushForm.js
```diff
@@ -39,15 +39,24 @@ define(['okta', 'util/CookieUtil', 'util/Util'], function (Okta, CookieUtil, Uti
...
           }
         }
       );
       this.listenTo(this.options.appState, 'change:isMfaTimeout',
         function (state, isMfaTimeout) {
           this.setSubmitState(isMfaTimeout);
           if (isMfaTimeout) {
-            this.showError(Okta.loc('oktaverify.timeout', 'login'));
+            this.showError({
+              name: Enums.MFA_VERIFY_ERROR,
```
mauriciocastillosilva-okta: would it make more sense to have a value for session timeout? 
```
name: Enums.SESSION_TIMEOUT,
```
or 
```
name: Enums.SESSION_EXPIRED,
```
the reason I am asking is that an MFA_VERIFY_ERROR sounds to me that something went wrong while verifying a factor.

PR #612, opened 2018-12-18
Passing along login_hint for all idps rather than just SAML

petermiller-okta:
We would like to pass along the login_hint value for more than just SAML idps

@okta/idx 
@nikhilvenkatraman-okta

### DIFF
--- a/src/models/IDPDiscovery.js
+++ b/src/models/IDPDiscovery.js
@@ -71,9 +71,7 @@ function (Okta, PrimaryAuthModel, CookieUtil, Enums) {
                   redirectToIdp: function (redirectUrl) {
                     if(res.links && res.links[0] && res.links[0].href) {
                       var queryParams = {fromURI: redirectUrl};
-                      if (res.links[0].properties && res.links[0].properties['okta:idp:type'] === 'SAML2') {
-                        queryParams['login_hint'] = username;
-                      }
+                      queryParams['login_hint'] = username;
                       var url = res.links[0].href + Util.getUrlQueryString(queryParams);
                       Util.redirect(url);
                     }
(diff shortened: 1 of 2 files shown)


src/models/IDPDiscovery.js
```diff
@@ -71,7 +71,7 @@ function (Okta, PrimaryAuthModel, CookieUtil, Enums) {
                   redirectToIdp: function (redirectUrl) {
                     if(res.links && res.links[0] && res.links[0].href) {
                       var queryParams = {fromURI: redirectUrl};
-                      if (res.links[0].properties && res.links[0].properties['okta:idp:type'] === 'SAML2') {
+                      if (res.links[0].properties) {
```
haishengwu-okta: - curious when `.properties` would be null?
- and when `.properties` exists, is it guaranteed that `.properties['okta:idp:type']` exists as well?

petermiller-okta: I don't think it would ever be null. These properties include the idp type (okta, saml2, facebook, google, etc) and the idp id (for those idps that have an ID). Maybe it wouldn't hurt to just always do `queryParams['login_hint'] = username`

src/models/IDPDiscovery.js
```diff
@@ -71,9 +71,7 @@ function (Okta, PrimaryAuthModel, CookieUtil, Enums) {
                   redirectToIdp: function (redirectUrl) {
                     if(res.links && res.links[0] && res.links[0].href) {
                       var queryParams = {fromURI: redirectUrl};
-                      if (res.links[0].properties && res.links[0].properties['okta:idp:type'] === 'SAML2') {
-                        queryParams['login_hint'] = username;
-                      }
+                      queryParams['login_hint'] = username;
```
petermiller-okta: This needs to be like this, since `login_hint` can't be camel cased in line 73

PR #613, opened 2018-12-19
:bug: Add form submit for redirection and guard by feature flag

haishengwu-okta:
## Description

- Mostly from https://github.com/okta/okta-signin-widget/pull/604
- Add ability to redirect via form submit for IdP Discovery, Session Step up and Session Redirect flow behind FF `features.redirectByFormSubmit`
- Add `favicon.ico` for playground app

## Screencasts

- browser: https://okta.box.com/s/da65bo7kbq1nbvy92mkf8uh2bzpd67g7
- android: https://okta.box.com/s/w3kkbairlr7wdgnui5rmwhn5kd0881zd
- android (idp discovery flow): https://okta.box.com/s/tbxfdh2vzfkh2que3mrjux8bw2x8fmgx

## Additional Reviewers

- @mauriciocastillosilva-okta 
- @gowthamidommety-okta 
- @rchild-okta


src/util/Enums.js
```diff
@@ -12,6 +12,8 @@
 
 define({
 
+  WIDGET_CONTAINER_ID: 'okta-sign-in',
```
nbarbettini: How does this relate to `options.el`? The Widget may not always be rendered in an element with id `#okta-sign-in`.

manueltanzi-okta: `#okta-sign-in` is not external to the widget, it's the top level element of the widget itself. `WIDGET_CONTAINER_ID` name might be confusing tho.

haishengwu-okta: It's exactly what Maunel said.
screen shot 2019-01-02 at 9 11 19 am

src/models/IDPDiscovery.js
```diff
@@ -66,14 +67,18 @@ function (Okta, PrimaryAuthModel, CookieUtil, Enums) {
               this.trigger('goToPrimaryAuth');
             }
             else {
+              var redirectFn = this.settings.get('features.redirectByFormSubmit')
```
manueltanzi-okta: Wonder if we should move this logic inside the Util so we don't have to replicate everytime

src/util/Util.js
```diff
@@ -113,6 +147,36 @@ define(['okta', 'util/Logger'], function (Okta, Logger) {
...
 
+  /**
+   * Why redirect via Form get rather using `window.location.href`?
+   * At the time of writing, Chrome (<72) in Android would block window location change
+   * at following case
+   * 1. An AJAX is trigger because of user action.
+   * 2. 5+ seconds passed without any further user interaction.
+   * 3. User takes an action results in window location change.
+   *
+   * Luckily we discovered that uses Form submit would work around this problem
```
manueltanzi-okta: nit: *using

src/util/Util.js
```diff
@@ -113,6 +147,36 @@ define(['okta', 'util/Logger'], function (Okta, Logger) {
...
+   * even though it changed window location.
+   *
+   * Check the commit history for more details.
+   */
+  Util.redirectWithFormGet = function (url) {
+    if (!url) {
+      Logger.error(`Cannot redirect to empty URL: (${url})`);
```
manueltanzi-okta: Curious, is there an advantage to pass the empty url?

haishengwu-okta: not really. it might (not quite likely) be helpful when it shows undefined or null or empty string.

src/models/IDPDiscovery.js
haishengwu-okta: I figured it could be an derived property in `models/Settings.js`, sounds good?

manueltanzi-okta: Yepp!

haishengwu-okta:
@manueltanzi-okta good question on e2e testing. I actually didn't figure out good approach. From user's perspective, changes are transparent - navigated to other site with URL modified. And once navigated to 'other site', there is no chance to inspect the DOM to confirm a Form is created.

nbarbettini:
> From user's perspective, changes are transparent - navigated to other site with URL modified.

@haishengwu-okta I would argue that is a "pass", since it verifies the most important thing: that the Widget is working correctly.

src/util/Util.js
```diff
@@ -11,11 +11,45 @@
...
-define(['okta', 'util/Logger'], function (Okta, Logger) {
+define(['okta', './Logger', './Enums'], function (Okta, Logger, Enums) {
 
   var Util = {};
   var _ = Okta._;
 
+  var buildInputForParameter = function (name, value) {
+    var input = document.createElement('input');
+    input.name = name;
+    input.value = decodeURIComponent(value);
```
jmelberg-okta: We should probably validate that `value` isn't a malformed URI, otherwise this assignment will throw an error.

src/util/Util.js
```diff
@@ -113,6 +147,36 @@ define(['okta', 'util/Logger'], function (Okta, Logger) {
...
+   */
+  Util.redirectWithFormGet = function (url) {
+    if (!url) {
+      Logger.error(`Cannot redirect to empty URL: (${url})`);
+      return;
+    }
+
+    var mainContainer = document.getElementById(Enums.WIDGET_CONTAINER_ID);
+    if (!mainContainer) {
+      Logger.error('Cannot find okta-sign-in container append to which a form');
```
jmelberg-okta: If we hit this conditional, the user will not be redirected to Okta and/or a defined step-up URL.Any reason **not** to use the "global error" handler, as the Widget cannot recover from this state?

src/util/Util.js
haishengwu-okta: there are tests in Util_spec to test malformed value. are those what you meant or any examples you might think of?

src/util/Util.js
haishengwu-okta: actually good question!
My initial thought was those 2 error are more for programming safety rather runtime exception (user has nothing to do with cause nor fix it) hence I don't even think of `callGlobalErrorFn`.
hmm, turns of cause some inconsistency.

src/util/Util.js
jmelberg-okta: Hmm. Maybe I missed it, but it doesn't look like `Util_spec` is asserting failures for invalid URLs.

I'm guessing this is a fringe case because we're running `decodeURIComponent` on the query string only, but wouldn't the following case would result in a thrown error?

```javascript
setCookieAndRedirect("https://foo.com?q=%E0%A4%A")
```

^ Looking at the code execution, we're going to attempt to `buildDynamicForm` with the encoded value returned from the [`sessionCookieRedirectTpl`](https://github.com/okta/okta-signin-widget/blob/2e662d97ef63a37c55bf2bd90c8af3373a98311a/src/util/RouterUtil.js#L34-L37). Seems like this would error, right?

Of course, let me know if I'm missing something.

src/util/Util.js
jmelberg-okta: Agreed - I think your decision to use the `Logger` is correct, we're just a bit inconsistent :)

Alternatively, we can also trigger an `afterError` event, so developers can respond accordingly.

This might require a bit more investigation, so what do you think about adding it to [OKTA-200220](https://oktainc.atlassian.net/browse/OKTA-200220)?

src/util/Util.js
haishengwu-okta: hmm, what I found is `setCookieAndRedirect("https://foo.com/.....")` would translate to
`redirectFn('http://example.com/?redirectURI=https%3A%2F%2Ffoo.com%3Fq%3D%25E0%25A4%25A')`
which would create a form like
```
'<form method="get" style="display: none;" action="http://example.com">' +
'<input name="redirectURI" type="hidden" value="https://foo.com?q=%E0%A4%A">' +
'</form>'
```
hence the redirect function would not throw any errors and be able to redirect but okta side may not able to interpret the redirect URI.
I tested with `redirectWithFormGet` feature on and off with following code and at both cases are able to redirect back to okta side (though it's error from okta side)

```
res.session.setCookieAndRedirect(signinWidgetOptions.baseUrl + '/app/UserHome?q=%E0%A4%A');
```

src/util/Util.js
haishengwu-okta: I wouldn't think it's worth to be a `afterError` event and more prefer `callGlobalError` if do want to bubble up the error.

haishengwu-okta: I'm debating myself whether it's good idea to trigger an event or failed silently (log to console) to developer.
Didn't have a good answer hence add comment to the ticket.
Thanks for raising the question!

nbarbettini: Isn't this an error that should never happen? Since the Widget is in control of `WIDGET_CONTAINER_ID`. If this check fails, then something is really wrong and the Widget is likely not even in the DOM.

There's no need to emit `afterError` because there's nothing a developer could do. JavaScript console makes sense to me.

nbarbettini:
Double checking our earlier decision: the flag that controls this feature is default false/off, correct?

haishengwu-okta:
@nbarbettini yes.
```

PR #614, opened 2018-12-20
Use 'idempotent-babel-polyfill' instead of babel-polyfill

lolney:
Babel-polyfill causes issues when it's instantiated more than once, e.g. when it's used with Webpack as well as an included package (see https://github.com/babel/babel-loader/issues/401). It looks like you partially addressed this with https://github.com/okta/okta-signin-widget/pull/353, but babel-polyfill still gets included in the js files that you serve on your CDN.

This PR replaces it with `idempotent-babel-polyfill`, which guards against requiring babel-polyfill multiple times.

### DIFF
--- a/webpack.release.config.js
+++ b/webpack.release.config.js
@@ -43,7 +43,7 @@ entryConfig.plugins = plugins({ isProduction: false, analyzerFile: 'okta-sign-in
 
 // 2. cdnConfig
 var cdnConfig = config('okta-sign-in.min.js');
-cdnConfig.entry.unshift('babel-polyfill');
+cdnConfig.entry.unshift('idempotent-babel-polyfill');
 cdnConfig.plugins = plugins({ isProduction: true, analyzerFile: 'okta-sign-in.min.analyzer' });
 
 // 3. noJqueryConfig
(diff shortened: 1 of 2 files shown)


webpack.release.config.js
```diff
@@ -43,7 +43,7 @@ entryConfig.plugins = plugins({ isProduction: false, analyzerFile: 'okta-sign-in
 
 // 2. cdnConfig
 var cdnConfig = config('okta-sign-in.min.js');
-cdnConfig.entry.unshift('babel-polyfill');
+cdnConfig.entry.unshift('idempotent-babel-polyfill');
```
haishengwu-okta: probably need to update the `devConfig` below as well.

lolney: Whoops, yep. I guess I should also commit the yarn.lock and keep babel-polyfill in the dev dependencies, since it's used in the karma config.

lolney: Hmm, after doing that it looks like karma still runs the tests (verified by inspecting in Chrome), but doesn't output results to the terminal.

Btw, another option might be to move to Babel v7+ and use the `noConflict` mode: https://github.com/babel/babel/blob/master/CHANGELOG.md#rocket-new-feature-12

PR #618, opened 2019-01-08
Upgrade Courage and i18n versions

willgantt-okta:
## Description
These version upgrades were prompted by a [customer urgent](https://oktainc.atlassian.net/browse/OKTA-199716) in which users were unable to register when `countryCode` or `preferredLanguage` were added as fields to the registration form. However, given that it has been close to three months since the last Courage version upgrade, other changes have been brought in as well. I have included video showing successful registration using the new widget below.

Bacon: https://bacon.aue1e.internal/#!/commits/okta-signin-widget/sha/d419a8b72bbd8c9559e4215eca077bc9f368cc43

## Video
**Registration with the new widget**
registrationwithnewwidget

### DIFF
--- a/packages/@okta/courage-for-signin-widget/package.json
+++ b/packages/@okta/courage-for-signin-widget/package.json
@@ -7,7 +7,7 @@
     "build": "webpack"
   },
   "devDependencies": {
-    "@okta/courage": "4.4.0",
+    "@okta/courage": "4.6.0-beta.2098.gc581fa3",
     "webpack": "^3.5.4",
     "webpack-bundle-analyzer": "^3.0.2"
   }
--- a/packages/@okta/courage-dist/okta.js
+++ b/packages/@okta/courage-dist/okta.js
@@ -1,4 +1,4 @@
-/*! THIS FILE IS GENERATED FROM PACKAGE @okta/courage@4.4.0 */
+/*! THIS FILE IS GENERATED FROM PACKAGE @okta/courage@4.6.0-beta.2098.gc581fa3 */
 module.exports =
 /******/ (function(modules) { // webpackBootstrap
 /******/ 	// The module cache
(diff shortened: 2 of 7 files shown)


packages/@okta/courage-for-signin-widget/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "webpack"
   },
   "devDependencies": {
-    "@okta/courage": "4.4.0",
+    "@okta/courage": "4.6.0-beta.2050.g7ec8178",
```
mauriciocastillosilva-okta: can we use a release version? Generally speaking, we've been using only release versions for the sign-in widget

src/EnrollCallAndSmsController.js
```diff
@@ -101,7 +101,7 @@ function (Okta, FormController, Footer, PhoneTextBox, TextBox, CountryUtil, Form
         self.trigger('errors:clear');
 
         if(!phoneNumber.length) {
-          self.trigger('invalid', self, {'phoneNumber': 'model.validation.field.blank'});
+          self.trigger('invalid', self, {'phoneNumber': 'model.validation.this.field.blank'});
```
mauriciocastillosilva-okta: if there is anyone providing a custom message for `model.validation.field.blank`, it won't work anymore. So I think we need to release note that we change the `key` name, so when someone updates to the latest sign in widget, they are aware (@rchild-okta is this right? Or de we need some more communication about it)

packages/@okta/courage-for-signin-widget/package.json
mauriciocastillosilva-okta: discussed offline, we are good with beta version

src/EnrollCallAndSmsController.js
rchild-okta: That's correct - if we're changing the property names, it would be a breaking change (exactly as you're describing).

+ @haishengwu-okta @robertdamphousse-okta

haishengwu-okta: good catch!!
Assume you're talking about developer override key via https://github.com/okta/okta-signin-widget#language-and-text.
we probably have to do so.

cc @nbarbettini

mauriciocastillosilva-okta: yes, that is the one I am talking about.

manueltanzi-okta: @haishengwu-okta @rchild-okta @mauriciocastillosilva-okta Imho, if that is the case, it is more worth to do an i18n breaking change keeping the name of the key and changing the value (assuming the value has to change for every occurrence) rather than doing a breaking change to the widget changing the key, so we are not affecting any external developers. CC: @robertdamphousse-okta @nbarbettini

haishengwu-okta:
@willgantt-okta I think we shall have unit test for verifying the problem has been fixed.

willgantt-okta:
@haishengwu-okta I'm testing the latest widget changes out in the monolith and will add that test soon.

packages/@okta/courage-for-signin-widget/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "webpack"
   },
   "devDependencies": {
-    "@okta/courage": "4.4.0",
+    "@okta/courage": "^4.6.0-beta.2093.ga073567",
```
mauriciocastillosilva-okta: we should keep this as an specific version (the `^` means it is that version or above, s we don't need that)

PR #619, opened 2019-01-09
:seedling: Support next link in any SUCCESS state

manueltanzi-okta:
### Description
- In any SUCCESS state add a `next` function to `successData` which redirect to `next.href`, if available, instead of  the`setCookieAndRedirect` function.

Part of: [OKTA-197158](https://oktainc.atlassian.net/browse/OKTA-197158)

### Reviewers
@haishengwu-okta @royalchan-okta @tthompson-okta

### DIFF
--- a/src/util/RouterUtil.js
+++ b/src/util/RouterUtil.js
@@ -124,7 +124,7 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
 
       var successData = {
         user: res._embedded.user,
-        type: res.type
+        type: res.type || Enums.SESSION_SSO
       };
 
       if (res.relayState) {
@@ -133,17 +133,12 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
 
       var redirectFn = router.settings.get('redirectUtilFn');
 
-      if (res.type === Enums.SESSION_STEP_UP) {
-        var targetUrl = res._links && res._links.next && res._links.next.href;
-        successData.stepUp = {
-          url: targetUrl,
-          finish: function () {
-            redirectFn(targetUrl);
-          }
+      var nextUrl = res._links && res._links.next && res._links.next.href;
+      if (nextUrl) {
+        successData.next = function () {
+          redirectFn(nextUrl);
         };
       } else {
-        // Add the type for now until the API returns it.
-        successData.type = Enums.SESSION_SSO;
         successData.session = {
           token: res.sessionToken,
           setCookieAndRedirect: function (redirectUrl) {
--- a/src/util/RouterUtil.js
+++ b/src/util/RouterUtil.js
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-/* eslint complexity: [2, 30],max-statements: [2, 30] */
+/* eslint complexity: [2, 35], max-statements: [2, 30] */
 define([
   'okta',
   './OAuth2Util',
@@ -133,12 +133,23 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
 
       var redirectFn = router.settings.get('redirectUtilFn');
 
-      var nextUrl = res._links && res._links.next && res._links.next.href;
-      if (nextUrl) {
+      var nextUrl = res._links &&
+                    ((res._links.original && res._links.original.href) || (res._links.next && res._links.next.href));
+      if (res.type === Enums.SESSION_STEP_UP) {
+        var targetUrl = res._links && res._links.next && res._links.next.href;
+        successData.stepUp = {
+          url: targetUrl,
+          finish: function () {
+            redirectFn(targetUrl);
+          }
+        };
+      } else if (nextUrl) {
         successData.next = function () {
           redirectFn(nextUrl);
         };
       } else {
+        // Add the type for now until the API returns it.
+        successData.type = Enums.SESSION_SSO;
         successData.session = {
           token: res.sessionToken,
           setCookieAndRedirect: function (redirectUrl) {
--- a/test/unit/spec/LoginRouter_spec.js
+++ b/test/unit/spec/LoginRouter_spec.js
@@ -278,17 +278,21 @@ function (Okta, Q, Logger, Errors, BrowserFeatures, WidgetUtil,
         });
     });
     itp('has a success callback which correctly implements the setCookieAndRedirect function', function () {
+      var spied = { };
+      spied.successFn = function (resp) {
+        if (resp.status === 'SUCCESS') {
+          resp.session.setCookieAndRedirect('http://baz.com/foo');
+        }
+      };
+      spyOn(spied, 'successFn').and.callThrough();
       spyOn(SharedUtil, 'redirect');
-      var successSpy = jasmine.createSpy('successSpy');
-      return setup({ globalSuccessFn: successSpy })
+      return setup({ globalSuccessFn: spied.successFn })
         .then(function (test) {
           test.setNextResponse(resSuccess);
           test.router.refreshAuthState('dummy-token');
-          return Expect.waitForSpyCall(successSpy);
+          return Expect.waitForSpyCall(spied.successFn);
         })
         .then(function () {
-          var setCookieAndRedirect = successSpy.calls.mostRecent().args[0].session.setCookieAndRedirect;
-          setCookieAndRedirect('http://baz.com/foo');
           expect(SharedUtil.redirect).toHaveBeenCalledWith(
             'https://foo.com/login/sessionCookieRedirect?checkAccountSetupComplete=true' +
           '&token=THE_SESSION_TOKEN&redirectUrl=http%3A%2F%2Fbaz.com%2Ffoo'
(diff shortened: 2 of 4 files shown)


src/util/RouterUtil.js
```diff
@@ -133,17 +133,12 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
...
+      var nextUrl = res._links && res._links.next && res._links.next.href;
+      if (nextUrl) {
+        successData.next = function () {
+          redirectFn(nextUrl);
         };
       } else {
-        // Add the type for now until the API returns it.
-        successData.type = Enums.SESSION_SSO;
         successData.session = {
           token: res.sessionToken,
```
manueltanzi-okta: Do we need `token` also in the case we have a `next.href`? If that's the case, will add, if not, will leave like this.

src/util/RouterUtil.js
```diff
@@ -133,17 +133,12 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
...
-      if (res.type === Enums.SESSION_STEP_UP) {
-        var targetUrl = res._links && res._links.next && res._links.next.href;
-        successData.stepUp = {
-          url: targetUrl,
-          finish: function () {
-            redirectFn(targetUrl);
-          }
+      var nextUrl = res._links && res._links.next && res._links.next.href;
+      if (nextUrl) {
+        successData.next = function () {
```
manueltanzi-okta: I opted for `.next` instead of `finish`, and since we don't use `stepUp.url` anywere in okta-core, I decided to totally remove the wrapping object instead of just renaming it. What do you think about this? Feel free to propose better names/structure.

src/util/RouterUtil.js
royalchan-okta: nope, the session token is unnecessary

src/util/RouterUtil.js
royalchan-okta: that's completely fine

src/util/RouterUtil.js
```diff
@@ -133,17 +133,12 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
...
+          redirectFn(nextUrl);
         };
       } else {
-        // Add the type for now until the API returns it.
-        successData.type = Enums.SESSION_SSO;
```
royalchan-okta: just making sure it's entirely safe to remove this, though I also assumed it wasn't used for anything.

manueltanzi-okta: @royalchan-okta I haven't removed, it's here: https://github.com/okta/okta-signin-widget/pull/619/files#diff-1d1f2f1a166e47fd617d3b5e9dfe4b74R127 ^^
However, that's a good point! Should we remove it? Is the API now guaranteed to always return a res.type?

src/util/RouterUtil.js
```diff
@@ -124,7 +124,7 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
 
       var successData = {
         user: res._embedded.user,
-        type: res.type
+        type: res.type || Enums.SESSION_SSO
```
haishengwu-okta: theoretically this changed the logic.
before: type is always either STEP_UP or SESSION_SSO.
after: type is STEP_UP or SESSION_SSO or anything when `res.type` is not null.

Am curious what else the `res.type` could be beside STEP_UP and SESSION_SSO from API.
This will change the expectation in case developers rely on the type.

src/util/RouterUtil.js
```diff
@@ -133,17 +133,12 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
 
       var redirectFn = router.settings.get('redirectUtilFn');
 
-      if (res.type === Enums.SESSION_STEP_UP) {
-        var targetUrl = res._links && res._links.next && res._links.next.href;
-        successData.stepUp = {
```
haishengwu-okta: Also strictly speaking this is breaking change (`successData` which is parameter to global success function).
Although we're fine with breaking change, is it better to add deprecate notes/logger instead of removing until 3.0?

Basically add `successData.next` and keep `stepUp` logic unchanged until 3.0 and go ahead to use `successData.next` from monolith.

thoughts?

src/util/RouterUtil.js
manueltanzi-okta: You right, was thinking under the assumption that the comment `// Add the type for now until the API returns it.` means that the type was `null` (except for `SESSION_STEP_UP`) and not something else.

src/util/RouterUtil.js
manueltanzi-okta: Good point here 👍 , it is a breaking change. I can keep the `stepUp` logic and remove it when we go to 3.0 as you said. Wonder if we can already modify the `README.md` (https://github.com/okta/okta-signin-widget/blame/master/README.md#L247) to encourage to use the new way (but as you said keep the old way working) or if we should wait 3.0 also for that, what's your opinion @haishengwu-okta @nbarbettini?

src/util/RouterUtil.js
haishengwu-okta: would be better to get confirmed regarding than merely trust comments

src/util/RouterUtil.js
haishengwu-okta: sounds good idea to update README

src/util/RouterUtil.js
royalchan-okta: no, unfortunately not, as some older flows will omit it because it isn't even set in the state token in okta-core

src/util/RouterUtil.js
royalchan-okta: it can be `SESSION_STEP_UP` or `CREDENTIAL_VERIFY` i believe. Probably better to avoid changing behavior for now. It's a bit of a shame because we don't really use the value correctly or effectively, but we can worry about that later.

tthompson-okta: We are adding new `type` values for IdX though:

* `LOGIN`
* `ENROLL_NEW_USER`
* `RECOVER_LOGIN`

src/util/RouterUtil.js
```diff
@@ -133,17 +133,12 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
...
-          url: targetUrl,
-          finish: function () {
-            redirectFn(targetUrl);
-          }
+      var nextUrl = res._links && res._links.next && res._links.next.href;
```
tthompson-okta: Hmm, it's not actually going to be a "next" link but rather an "original" link:

```json
"_links": {
        "original": {
            "href": "https://your-org.okta.com/login/idx/redirect?stateToken=01OCl7uyAUC4CUqHsObI9bvFiq01cRFgbnpJQ1bz82",
            "hints": {
                "allow": [
                   "GET"
                ]
            }
        }
    }
```

tthompson-okta: Actually -- we need to debate this.

@tomabbott-okta @wdawson-okta @robertdamphousse-okta @karlmcguinness-okta 

We have been moving away from the "next" link pattern and moving towards more explicit link names which are the actual operation names. So instead of:
```json
"_links":{
      "next":{
         "name":"original",
         "href":"https://atko.oktapreview.com/login/step-up/redirect?stateToken=00quAZYqYjXg9DZhS5UzE1wrJuQ6KKb_kzOeH7OGB5",
         "hints":{
            "allow":[
               "GET"
            ]
         }
      }
   }
```
you'd have:
```json
"_links": {
        "original": {
            "href": "https://your-org.okta.com/login/idx/redirect?stateToken=01OCl7uyAUC4CUqHsObI9bvFiq01cRFgbnpJQ1bz82",
            "hints": {
                "allow": [
                   "GET"
                ]
            }
        }
    }
```
This is how MFA_REQUIRED state is [already modeled in the Auth API.](https://developer.okta.com/docs/api/resources/authn#response-example-for-primary-authentication-with-public-application-factor-challenge). In general we've basically had explicit link operation names in any embedded structures, and then use the `next` pattern in the top-level links.

IMO we need to move away from the `next` pattern altogether and just always offer explicit operation names as the link names.

wdawson-okta: Agree there might be multiple paths to a SUCCESS. That, to me, is the root of why we need to figure out a different pattern than "next".

tthompson-okta: So while we're here then, could we make the widget handle both `next` and `original` the same way (which is: ignore any `sessionToken` and just redirect to the link)? That should keep it backward compatible with the [AuthN Session Step-Up](https://developer.okta.com/docs/api/resources/authn#sp-initiated-step-up-authentication) functionality, which I'm still hoping we can deprecate but for now needs to keep working.

manueltanzi-okta: Good for me 👍

src/util/RouterUtil.js
```diff
@@ -133,12 +133,22 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
...
+            redirectFn(targetUrl);
+          }
+        }
+      } else if (nextUrl) {
         successData.next = function () {
           redirectFn(nextUrl);
         };
       } else {
+        // Add the type for now until the API returns it.
+        successData.type = Enums.SESSION_SSO;
```
manueltanzi-okta: @haishengwu-okta Added this back here, so will not be breaking change.
I would also keep the `||` condition in line https://github.com/okta/okta-signin-widget/pull/619/commits/3f7564633e4a8124290eeca6ac8211f4522c19e7#diff-1d1f2f1a166e47fd617d3b5e9dfe4b74R127 so that the in new if case (original or next link) will behave in that new way, since we will have new response.type in future, and we probably don't want to force `Enums.SESSION_SSO` everytime, what do you think?

src/util/RouterUtil.js
manueltanzi-okta: Thinking more about `README.md` changes (thanks @robertdamphousse-okta!) we will have to wait to update it, since old behavior (`stepUp` format) still has priority on new behavior (original/next format) and new APIs format that returns original link are still under FF/not ready, so developers cannot really use the new way yet.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -356,6 +358,88 @@ function (Okta, Q, Logger, Errors, BrowserFeatures, WidgetUtil,
           );
         });
     });
+    itp('for success with an original link, success callback data contains a next function that redirects to original.href', function () {
+      spyOn(SharedUtil, 'redirect');
+      var successSpy = jasmine.createSpy('successSpy');
+      return setup({ stateToken: 'aStateToken', globalSuccessFn: successSpy })
```
haishengwu-okta: shall we set up `globalSuccessFn` as normal function 
```
function(resp) {
  if (resp.type === SOMETHING && resp.next) {
    resp.next();
  }
}
```

src/util/RouterUtil.js
haishengwu-okta: shall we override the `type` to `Enums.SESSION_SSO` as same as the `else` block rather than at line 127?

src/util/RouterUtil.js
haishengwu-okta: I'm kind of hesitate and prefer revisit the code when the future comes.

src/util/RouterUtil.js
royalchan-okta: You'll actually need to update the README since it calls out these values

manueltanzi-okta: I agree with you that we should update the README, but we cannot do it in this PR and until we move to use only the new way (see discussion with @tthompson-okta in the other comment where we decided to keep it backward compatible), because right now to ensure retrocompatibility, the old way (`.stepUp.finish()`) needs to have still priority on the new one in the if conditions. If we update README and developers start to use `.next`, they will still receive the response from SESSION_STEP_UP with the `stepUp.finish` function and not the new `.next`, so their code would not work.

src/util/RouterUtil.js
royalchan-okta: Just as an FYI, I think the true ideal situation is that the widget does not care what the `type` is at all, as it should not need to behave differently.

PR #623, opened 2019-01-22
Enable enrollment of Password Factor

kenwang-okta:
Resolves: OKTA-206190

Enable the enrollment of Password Factor in the Okta Sign in Widget.

Screenshots here (See last 3 screenshots):
https://oktawiki.atlassian.net/wiki/spaces/eng/pages/630128829/Password+as+Factor+-+UI+and+i18n+changes

Video:
https://okta.box.com/s/pyu8e2n0no0mrit6qy0p409j3jtnzlvm


test/unit/spec/EnrollChoices_spec.js
```diff
@@ -567,14 +567,14 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, Router,
         return setup(resAllFactors).then(function (test) {
           var factorList = test.form.getFactorList();
           expect(factorList).toEqual(['OKTA_VERIFY', 'U2F', 'WINDOWS_HELLO', 'YUBIKEY', 'GOOGLE_AUTH',
-            'SMS', 'CALL', 'QUESTION', 'DUO', 'SYMANTEC_VIP', 'RSA_SECURID', 'GENERIC_SAML', 'GENERIC_OIDC']);
```
kenwang-okta: Question for UI folks: What actually determines the factorList order? It doesn't seem to match the order in MFA_ENROLL_allFactors.js.

Does it even matter?

src/EnrollPasswordController.js
```diff
@@ -0,0 +1,82 @@
...
+function (Okta, FormController, ValidationUtil, FactorUtil, Footer, TextBox) {
+
+  var _ = Okta._;
+
+  return FormController.extend({
+    className: 'enroll-password',
+    Model: {
+      props: {
+        password: ['string', true],
+        confirmPassword: ['string', true]
```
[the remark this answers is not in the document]
mauriciocastillosilva-okta: second argument is for 'required' value. If true, the model will validate this property to not be empty

test/unit/spec/EnrollPassword_spec.js
```diff
@@ -0,0 +1,199 @@
...
+        expect(test.beacon.hasClass('mfa-okta-password')).toBe(true);
+      });
+    });
+    itp('does not allow autocomplete', function () {
+      return setup().then(function (test) {
+        expect(test.form.getPasswordAutocomplete()).toBe('off');
+        expect(test.form.getConfirmPasswordAutocomplete()).toBe('off');
+      });
+    });
+    itp('has an password field', function () {
```
mauriciocastillosilva-okta: `has a password field`

test/unit/spec/EnrollPassword_spec.js
```diff
@@ -0,0 +1,199 @@
...
+              provider: 'OKTA',
+              profile: {
+                password: 'somepassword'
+              },
+              stateToken: 'testStateToken'
+            }
+          });
+        });
+    });
+    itp('validates password and confirmPassword fields and errors before the request', function () {
```
mauriciocastillosilva-okta: nit: `validates password and confirmPassword cannot be empty`

src/EnrollPasswordController.js
```diff
@@ -0,0 +1,82 @@
...
+              password: this.get('password')
+            }
+          });
+        });
+      }
+    },
+
+    Form: {
+      autoSave: true,
+      title: _.partial(Okta.loc, 'enroll.password.setup', 'login'),
```
mauriciocastillosilva-okta: I don't see this `enroll.password.setup` key in login.properties... is there a different PR for it? We need to that it along with the rest of the strings in this PR

kenwang-okta:
@mauriciocastillosilva-okta thanks for the review. A video of the feature is here: https://okta.box.com/s/pyu8e2n0no0mrit6qy0p409j3jtnzlvm

src/EnrollPasswordController.js
kenwang-okta: Yep, here is the pr https://github.com/okta/npm/pull/1172

kenwang-okta:
@mauriciocastillosilva-okta addressed the nits.

src/EnrollPasswordController.js
```diff
@@ -0,0 +1,82 @@
...
+              password: this.get('password')
+            }
+          });
+        });
+      }
+    },
+
+    Form: {
+      autoSave: true,
+      title: 'enroll.password.setup',
```
mauriciocastillosilva-okta: we should use `_.partial` here. This is because we load other languages on demand, so having a function here guarantees that the label will be correct
```
title: _.partial(Okta.loc, 'enroll.password.setup', 'login'),
```

test/unit/spec/EnrollChoices_spec.js
mauriciocastillosilva-okta: It matters. We have a sort order defined in [FactorUtil](https://github.com/okta/okta-signin-widget/blob/master/src/util/FactorUtil.js#L24)

src/EnrollPasswordController.js
kenwang-okta: My mistake, did some hard coding of the message for testing purposes and accidentally committed that. Should be fixed now.

test/unit/spec/EnrollChoices_spec.js
kenwang-okta: Ah, I see. Side note, does it make sense to you to change the order so that password is first, given its importance? Or is it alright to leave it in the order that it is now?

src/util/FactorUtil.js
```diff
@@ -194,8 +194,7 @@ function (Okta, TimeUtil) {
   };
 
   fn.getFactorDescription = function (provider, factorType) {
-    var key = factorData[fn.getFactorName.apply(this, [provider, factorType])].description;
```
haishengwu-okta: curious why made this refactor?

test/unit/spec/EnrollChoices_spec.js
mauriciocastillosilva-okta: This is intentional, password is not first since it is not one of the strongest factors. We've had a few discussions and iterations of this with PMs. Take a look at our docs for authentication methods https://help.okta.com/en/prod/Content/Topics/Security/MFA.htm

src/util/FactorUtil.js
kenwang-okta: There was a request to not have a description for password, but we could not just pass '' or else we would see an i18nERROR[] (or something similar to that), since we are passing a non existent message key to Okta.loc.

Thus we need to move the definition up a stack so that we either pass a valid Okta.loc message, or empty string there.

kenwang-okta:
@mauriciocastillosilva-okta @haishengwu-okta I've addressed all concerns please let me know if there are any additional requested changes

src/util/FactorUtil.js
haishengwu-okta: - I didn't see `password` has added into this factor list, did I miss anything?
- you could also do
```
var key = ....description`
return _.isEmpty(key) ? '' : okta.loc(key, 'login');
```
- the main question is `okta.loc` has to be invoke after language bundle loaded. I'm not user your refactor would break the sequencing or not.

kenwang-okta:
Did not need.

PR #626, opened 2019-02-05
Add support for new stateToken states on idx pipeline

nikhilvenkatraman-okta:
- Added support for new stateToken state in the widget, which include `FACTOR_REQUIRED`,`FACTOR_CHALLENGE` and `PROFILE_REQUIRED`
- Added support for `UNAUTHENTICATED` state with passwordLess auth in the widget with new `idxStateToken`. This includes the changes that were removed from [this](https://github.com/okta/okta-signin-widget/pull/627) PR
- Created new route, view, controller for `PROFILE_REQUIRED`.
- `FACTOR_REQUIRED` and `FACTOR_CHALLENGE` (email and password factor) states render the same view in the widget. 
-  **Note** : Currently when the widget goes from `MFA_REQUIRED` to `MFA_CHALLENGE` state, the API sends a `prev` link in the response. This is used by the widget to make an API call to when the widget gets bootstrapped in the `MFA_CHALLENGE` state to make an API call to get `MFA_REQUIRED` state and render that view.
- In the new pipeline, we  just render then `FACTOR_REQUIRED` view if widget is bootstrapped with `FACTOR_CHALLENGE`, as API response for `FACTOR_REQUIRED`  does not contain `prev` link.
- With the new idx pipeline for `FACTOR-REQUIRED` state, with email as a factor, we show a screen with just an option to email a magic link. There is no `otp code input-box` and hence no `Verify` button on the screen ( see screenshot).
screen shot 2019-02-05 at 1 16 40 pm

- Added `EmailMagicLinkForm` view for `FACTOR_REQUIRED`, with email as factor.
- On `SUCCESS`, the widget will preform redirect to `redirect_uri` using the `idtoken`. 

Resolves:
 [OKTA-202668](https://oktainc.atlassian.net/browse/OKTA-202668)
 [OKTA-210077](https://oktainc.atlassian.net/browse/OKTA-210077)
 [OKTA-209004](https://oktainc.atlassian.net/browse/OKTA-209004)

@mauriciocastillosilva-okta @haishengwu-okta @manueltanzi-okta @okta/idx


src/models/ProfileSchema.js
```diff
@@ -0,0 +1,50 @@
...
+],
+function (Okta) {
+  var _ = Okta._;
+  var { BaseSchema, SchemaProperty } = Okta.internal.models;
+
+  var ProfileSchemaPropertyCollection = SchemaProperty.Collection.extend({
+    createModelProperties: function () {
+      var modelProperties = SchemaProperty.Collection.prototype.createModelProperties.apply(this);
+      _.each(modelProperties, function (modelProperty, name) {
+        this.get(name).attributes.title = this.get(name).get('displayName');
```
nikhilvenkatraman-okta: This is needed to show the placeholder text inside input fields.

nikhilvenkatraman-okta: screen shot 2019-01-31 at 11 49 26 am

src/models/ProfileSchema.js
```diff
@@ -0,0 +1,50 @@
...
+],
+function (Okta) {
+  var _ = Okta._;
+  var { BaseSchema, SchemaProperty } = Okta.internal.models;
+
+  var ProfileSchemaPropertyCollection = SchemaProperty.Collection.extend({
+    createModelProperties: function () {
+      var modelProperties = SchemaProperty.Collection.prototype.createModelProperties.apply(this);
+      _.each(modelProperties, function (modelProperty, name) {
+        this.get(name).attributes.title = this.get(name).get('label');
```
nikhilvenkatraman-okta: This is needed to show the placeholder text inside input fields.

screen shot 2019-02-14 at 3 13 01 pm

src/models/EnrollUser.js
```diff
@@ -0,0 +1,75 @@
...
+define([
+  'okta',
+  'models/ProfileSchema',
+  './BaseLoginModel'
+],
+function (Okta, ProfileSchema, BaseLoginModel) {
+  var {_} = Okta;
+
+  return BaseLoginModel.extend({
+    props: this.properties,
```
haishengwu-okta: I doubt this is necessary (or work properly)?
`this` refers to the context of callback function of `define` hence `this.properties` is undefined.

src/views/enrollUser/EnrollUserForm.js
```diff
@@ -0,0 +1,37 @@
...
+], function (Okta, RegistrationFormFactory) {
+  return Okta.Form.extend({
+    layout: 'o-form-theme',
+    autoSave: true,
+    noCancelButton: true,
+    title: Okta.loc('registration.form.title', 'login'),
+    save: Okta.loc('registration.form.submit', 'login'),
+    initialize: function (options) {
+      this.options = options || {};
+      var schema = this.options.appState.get('schema');
```
haishengwu-okta: I'm thinking we shall be able to set `schema` as part of `EnrollUser Model` rather than `appState`.

src/views/factor-verify/EmailMagicLinkForm.js
```diff
@@ -0,0 +1,50 @@
...
+  var _ = Okta._;
+  var API_RATE_LIMIT = 30000; //milliseconds
+
+  return PassCodeForm.extend({
+    className: 'factor-verify-passcode',
+    initialize: function () {
+      // for FACTOR_REQUIRED with email magic link we dont need to show otp code input field and verify button
+      var form = this;
+      this.title = this.model.get('factorLabel');
+      //TODO: Temp fix for demo. FACTOR_REQUIRED after sign up is missing the profile object in API response
```
haishengwu-okta: - any jira for the todo?
- feels like we create another derived field in `appState` to read `lastAuthResponse.login`?

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -30,10 +32,10 @@ define([
 ],
 function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Beacon, PrimaryAuth,
   Router, BrowserFeatures, Errors, DeviceFingerprint, TypingUtil, Expect, resSecurityImage,
-  resSecurityImageFail, resSuccess, resUnauthenticated, resLockedOut, resPwdExpired, resUnauthorized,
+  resSecurityImageFail, resSuccess, resUnauthenticated, resUnauthenticatedIdx, resFactorRequired, resLockedOut, resPwdExpired, resUnauthorized,
   resNonJson, resInvalidText, resThrottle, resPasswordlessUnauthenticated, $sandbox) {
 
-  var { _, $ } = Okta;
+  var { _, $} = Okta;
```
haishengwu-okta: inconsistent spacing. surprised you didn't get eslint error.

src/views/factor-verify/EmailMagicLinkForm.js
nikhilvenkatraman-okta: Actually the TODO is on the server side once the API returns the right response, we should just be able to do `var email = this.model.get('email')` to get the email.

src/models/EnrollUser.js
nikhilvenkatraman-okta: not necessary

src/views/factor-verify/EmailMagicLinkForm.js
haishengwu-okta: at that point, do we need to remove the else condition (` || ...`)

nikhilvenkatraman-okta: Yes https://oktainc.atlassian.net/browse/OKTA-211618

haishengwu-okta: please put in the comment as well

src/models/EnrollUser.js
```diff
@@ -0,0 +1,62 @@
...
+      }
+      return postData;
+    },
+    getEnrollFormData: function () {
+      return this.manageTransaction(function (transaction) {
+        return transaction.enroll();
+      });
+    },
+    save: function () {
+      var data = Okta.Model.prototype.toJSON.apply(this, arguments);
```
haishengwu-okta: guess shall be able to do `BaseLoginModel.prototype.toJSON.apply...`.

test/unit/spec/EnrollUser_spec.js
```diff
@@ -0,0 +1,158 @@
...
+      setNextResponse(resUnauthenticatedIdx);
+      return Expect.waitForPrimaryAuth(settings);
+    } else {
+      setNextResponse(resProfileRequiredUpdate);
+      return Expect.waitForEnrollUser(settings);
+    }
+  }
+
+  Expect.describe('Enroll User', function () {
+    Expect.describe('Enroll User Form', function () {
```
haishengwu-okta: `Expect.describe` has done extra setup hence running one `Expect.describe` at top level is sufficient.

src/views/factor-verify/EmailMagicLinkForm.js
```diff
@@ -0,0 +1,50 @@
...
+  var API_RATE_LIMIT = 30000; //milliseconds
+
+  return PassCodeForm.extend({
+    className: 'factor-verify-passcode',
+    initialize: function () {
+      // for FACTOR_REQUIRED with email magic link we dont need to show otp code input field and verify button
+      var form = this;
+      this.title = this.model.get('factorLabel');
+      //TODO: OKTA-211618 Temp fix for demo. FACTOR_REQUIRED after sign up is missing the profile object in API response
+      var email = this.model.get('email') || this.options.appState.get('lastAuthResponse')._embedded.user.profile.login;
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: I did not want to do it, since we will be removing that bit of code when we address 
OKTA-211618

src/models/Factor.js
```diff
@@ -223,10 +223,18 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
         if (this.pushFactorHasAutoPush()) {
           data.autoPush = this.get('autoPush');
         }
-
+        // for FACTOR_REQUIRED with email magic link we dont need to show otp code input field,
+        // hence we dont need to post the otp code to the server
+        if (this.options.appState.get('isFactorRequired') && this.get('factorType') === 'email') {
```
mauriciocastillosilva-okta: I think that instead of deleting the property we just set a few lines above, we can probably add some logic to not set it in the first place (this will go in line 219). Something like
```
...
else if (this.addPasscode()) {
  data.passCode = this.get('answer');
}
```

src/models/PrimaryAuth.js
```diff
@@ -106,12 +106,19 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
...
-      if (this.appState.get('isUnauthenticated') && !this.settings.get('features.passwordlessAuth')) {
-        primaryAuthPromise = this.doTransaction(function (transaction) {
-          var authClient = this.appState.settings.authClient;
-          return this.doPrimaryAuth(authClient, signInArgs, transaction.authenticate);
-        });
-      } else {
+      if (this.appState.get('isUnauthenticated')) {
+        var authClient = this.appState.settings.authClient;
+        // bootstrapped with stateToken
+        if (!this.settings.get('features.passwordlessAuth')) {
```
mauriciocastillosilva-okta: I feel like this can be simplified... seems to me conditions for passwordless are not needed
Currently you have
```
if not passwordless
  -> transaction.authenticate
else if idxToken
  -> transaction.login
else
  -> transaction.authenticate
```
can we avoid checking for passwordless? Seems that it is irrelevant when we are in unauthenticated state.
```
if idxToken
 -> transaction.login
else
 -> transaction.authenticate
```

src/util/RouterUtil.js
```diff
@@ -169,6 +169,10 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
         router.navigate('signin/consent', {trigger: true});
       }
       return;
+    // We want the same view for FACTOR_REQUIRED & FACTOR_CHALLENGE
+    // In the new idx pipeline FACTOR_CHALLENGE API response does not contain a prev link
```
mauriciocastillosilva-okta: is this going to be true? Or are the backend folks revisiting this? Asking to see if it makes sense to keep the comment or not

src/views/enrollUser/Footer.js
```diff
@@ -0,0 +1,33 @@
+/*!
```
mauriciocastillosilva-okta: will it make sense to add this file to `views/shared` as a `FooterBackLink.js` or similar? Seems it is generic enough that it can be reused

src/views/factor-verify/EmailMagicLinkForm.js
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+/* eslint complexity: [2, 7] */
+define(['okta', 'q', 'views/mfa-verify/PassCodeForm'], function (Okta, Q, PassCodeForm) {
+
+  var subtitleTpl = Okta.Handlebars.compile('({{subtitle}})');
+  var _ = Okta._;
+  var API_RATE_LIMIT = 30000; //milliseconds
+
+  return PassCodeForm.extend({
```
mauriciocastillosilva-okta: probably we shouldn't extend `PassCodeForm` since this form doesn't need a passcode. `PassCodeForm` doesn't have any functions/attributes that we can use here, so it is probably better to extend `Okta.Form.extend`

src/EnrollUserController.js
```diff
@@ -0,0 +1,59 @@
...
+],
+function (
+  Okta,
+  EnrollUser,
+  BaseLoginController,
+  EnrollUserForm,
+  Footer,
+) {
+  var _ = Okta._;
+  return BaseLoginController.extend({
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: whats the rule for extending from `FormController` ? I wanted to do the API call to fetch some data inside the initialize method if user is trying to sign up from an unauthenticated state, hence i extended `BaseLoginController`

src/EnrollUserController.js
```diff
@@ -0,0 +1,59 @@
...
+    className: 'enroll-user',
+    initialize: function (options) {
+      // If user is unauthenticated and starts enroll flow make a post call to transition state to PROFILE_REQUIRED
+      this.isEnrollWithLoginIntent = !!(this.options.appState.get('lastAuthResponse').status === 'UNAUTHENTICATED');
+      this.options = options || {};
+      // create model
+      this.model = new EnrollUser(this.options);
+      if (this.isEnrollWithLoginIntent) {
+        this.model.getEnrollFormData()
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: Yes, thats correct. The first call to `enroll` is just with the stateToken that does the state transition to `PROFILE_REQUIRED` and the second call is the save that has the profile data.

src/util/RouterUtil.js
nikhilvenkatraman-okta: Yeah, I think in the IDX world we are moving away from the pattern of making a call to prev link.

src/EnrollUserController.js
```diff
@@ -0,0 +1,59 @@
...
+                !!this.options.appState.get('profileSchemaAttributes').createNewAccount);
+              this.renderForm();
+            }
+          },this));
+      } else {
+        this.renderForm();
+      }
+    },
+    renderForm: function () {
+      // create form
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: removed

src/MfaVerifyController.js
```diff
@@ -44,7 +45,11 @@ function (Okta, BaseLoginController, TOTPForm, YubikeyForm, SecurityQuestionForm
       case 'sms':
       case 'call':
       case 'email':
-        View = PassCodeForm;
+        if (this.options.appState.get('isFactorRequired') && factorType === 'email'){
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: fixed

src/models/PrimaryAuth.js
nikhilvenkatraman-okta: that should work too !

test/unit/helpers/xhr/FACTOR_CHALLENGE_email.js
```diff
@@ -2,42 +2,51 @@ define({
   "status": 200,
   "responseType": "json",
   "response": {
-    "stateToken": "01MMFqbYA4dwgGi0t5I78wIhcvHS1pOxKJE_B6jiUk",
-    "expiresAt": "2015-09-27T01:14:24.573Z",
+    "stateToken": "01bfpkAkRyqUZQAe3IzERUqZGOfvYhX83QYCQIDnKZ",
```
mauriciocastillosilva-okta: nit: since you re changing for test data, you could also have this as `testToken`

src/models/Factor.js
```diff
@@ -212,7 +212,11 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
         return {'password': Okta.loc('error.password.required')};
       }
     },
-
+    addPasscode: function () {
```
mauriciocastillosilva-okta: nit: `needsPasscode`, otherwise I will assume the functions adds the passcode itself

src/models/Factor.js
```diff
@@ -212,7 +212,11 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
...
+      // for new idx pipeline with email magic link we dont need to show otp code input field,
```
mauriciocastillosilva-okta: nit: I think `// we don't need passcode in email with magic link` should be enough

src/EnrollUserController.js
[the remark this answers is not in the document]
royalchan-okta: Is this the situation where you hit the sign on page, you are in `UNAUTHENTICATED`, and you manually click "Sign Up" as a new user? In this case, there is a blank enrollment call that will transition the state token into `PROFILE_REQUIRED` and get you back the form data.

test/unit/helpers/xhr/FACTOR_CHALLENGE_email.js
nikhilvenkatraman-okta: I want the ixd stateToken to have `01` in it, so that it evaluates the appState variable `isIdxStateToken` correctly.

src/EnrollUserController.js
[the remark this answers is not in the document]
nikhilvenkatraman-okta: Yeah, the widget does get bootstrapped in an `UNAUTHENTICATED` state here. Like Royal mentioned, it is only when the user hits manually `sign-up`, we make this API call to the backend which transitions the stateToken to `PROFILE_REQUIRED ` and use the form fields in the API response to render the enroll user form.

src/models/Factor.js
nikhilvenkatraman-okta: fixed

src/EnrollUserController.js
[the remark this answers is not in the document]
mauriciocastillosilva-okta: Got it. Makes sense, we make that call to know what profile data we need when the user wants to sign up for an account.

test/unit/spec/EnrollUser_spec.js
```diff
@@ -0,0 +1,157 @@
...
+  'helpers/xhr/SUCCESS',
+  'helpers/xhr/UNAUTHENTICATED_IDX'
+],
+function (Okta, OktaAuth, LoginUtil, Util, EnrollUserForm, Expect, Router,
+  $sandbox, resProfileRequiredUpdate, resProfileRequiredNew, resSuccess, resUnauthenticatedIdx) {
+
+  var { _, $, Backbone } = Okta;
+  var itp = Expect.itp;
+
+  function setup (settings, resUnauthenticatedIdx) {
```
mauriciocastillosilva-okta: I don't see a call with `settings` argument, so we can probably skip it
and there is only one call with `resUnauthenticatedIdx`. I think we should have 2 functions
```
function setupEnroll
function setupUnauthenticated()
```
or the other option is to refactor this one to something like
```
function setup (isUnauthenticated) {
  nextResponse = isUnathetnticated ? resUnauthenticatedIdx : resProfileRequiredUpdate;
```

test/unit/spec/EnrollUser_spec.js
```diff
@@ -0,0 +1,157 @@
...
+        test.setNextResponse(resProfileRequiredNew);
+        test.form.$('.registration-link').click();
+        return Expect.waitForEnrollUser(test);
+      })
+        .then(function (test) {
+          expect(test.form.formInputs('streetAddress').length).toEqual(1);
+          expect(test.form.formInputs('streetAddress').hasClass('okta-form-input-field input-fix')).toBe(true);
+          expect(test.form.formInputs('employeeId').length).toEqual(1);
+          expect(test.form.formInputs('employeeId').hasClass('okta-form-input-field input-fix')).toBe(true);
+          return Expect.waitForEnrollUser(test);
```
mauriciocastillosilva-okta: why do we need this `wait`? I don't see any submit form happening or something that will required a wait.

test/unit/spec/EnrollUser_spec.js
```diff
@@ -0,0 +1,157 @@
...
+        })
+        .then(function (test) {
+          $.ajax.calls.reset();
+          test.setNextResponse(resSuccess);
+          var model = test.router.controller.model;
+          model.set('streetAddress', 'street address');
+          model.set('employeeId', '1234');
+          spyOn(Backbone.Model.prototype, 'save').and.returnValue($.Deferred().resolve());
```
mauriciocastillosilva-okta: can this be `EnrollUser` model instead?

test/unit/spec/EnrollUser_spec.js
```diff
@@ -0,0 +1,157 @@
...
+      });
+    });
+
+    itp('makes call to enroll if isEnrollWithLoginIntent is true and then renders the right fields based on API response', function () {
+      return setup(null, resUnauthenticatedIdx).then(function (test) {
+        test.setNextResponse(resProfileRequiredNew);
+        test.form.$('.registration-link').click();
+        return Expect.waitForEnrollUser(test);
+      })
+        .then(function (test) {
```
mauriciocastillosilva-okta: spacing is off here. Other places that you use `.then` it has the same indentation as the previous line (see lines 100 or 110)

src/models/EnrollUser.js
```diff
@@ -0,0 +1,61 @@
...
+      this.appState = this.options.appState;
+    },
+    constructPostData: function (profileAttributes) {
+      var postData = {
+        'registration': {
+          'profile': profileAttributes
+        }
+      };
+      // send createNewAccount flag for new user creation
+      if (this.get('createNewAccount')) {
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: @iancorcoran-okta can probably add more details here, but I believe there is a plan in the future to have a separate endpoint for [enroll](https://gist.github.com/tthompson-okta/57ecf1b9fd68e68ff77967e2b86ed00b/bbccdbd88658b0a3739ac177873448a915c2bb79#request-with-registration-profile) and [update](https://gist.github.com/tthompson-okta/57ecf1b9fd68e68ff77967e2b86ed00b/bbccdbd88658b0a3739ac177873448a915c2bb79#request-with-profile-updates)
```
           { 
              "name": "enroll",
              "href":"https://your-org.okta.com/api/v1/authn/enroll",
              "hints":{
                  "allow":[
                     "POST"
                  ]
              }
            }
```
 ```
            { 
              "name": "updateProfile",
              "href": "https://your-org.okta.com/api/v1/authn/enroll/update",
              "hints": {
                 "allow": [
                    "POST"
                 ]
              }
            }
```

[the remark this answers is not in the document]
Comment (iancorcoran-okta) [reply src/models/EnrollUser.js]: Yeah we will have separate `/authn/enroll` and `/authn/enroll/update` endpoints. We can revisit whether that makes the `createNewAccount` flag superfluous after the beta deadline

File: test/unit/spec/EnrollUser_spec.js (hunk 24 above)
Comment (nikhilvenkatraman-okta) [reply test/unit/spec/EnrollUser_spec.js]: I removed the spy and just set model.set('createNewAccount', true/false) (depending on if its enroll vs update) and called model.save(). That seems to have worked.

File: test/unit/spec/EnrollUser_spec.js (hunk 23 above)
Comment (nikhilvenkatraman-okta) [reply test/unit/spec/EnrollUser_spec.js]: removed

File: test/unit/spec/EnrollUser_spec.js (hunk 25 above)
Comment (nikhilvenkatraman-okta) [reply test/unit/spec/EnrollUser_spec.js]: i moved the `.then` to the same line

```       return setup(null, resUnauthenticatedIdx).then(function (test) {
        test.setNextResponse(resProfileRequiredNew);
        test.form.$('.registration-link').click();
        return Expect.waitForEnrollUser(test);
      }).then(function (test) {
        $.ajax.calls.reset();
        expect(test.form.formInputs('streetAddress').length).toEqual(1);
        expect(test.form.formInputs('streetAddress').hasClass('okta-form-input-field input-fix')).toBe(true);
        expect(test.form.formInputs('employeeId').length).toEqual(1);
        expect(test.form.formInputs('employeeId').hasClass('okta-form-input-field input-fix')).toBe(true);
        var model = test.router.controller.model;
        model.set('streetAddress', 'street address');
        model.set('employeeId', '1234');
        model.set('createNewAccount', true);
        model.save();
        return Expect.waitForEnrollUser(test);
      }).then(function () {
        Expect.isJsonPost($.ajax.calls.argsFor(0), {
          url: 'https://foo.okta.com/api/v1/authn/enroll',
          data: {
            'registration': {
              'createNewAccount': true,
              'profile': {
                'streetAddress': 'street address',
                'employeeId': '1234'
              }
            },
            'stateToken': '01nDL4wRHu-dLvUHUj1QCA9r5P1n5dw6WJ_voGPFWB'
          }
        });
      }); ```

File: src/EnrollUserController.js (hunk 15 above)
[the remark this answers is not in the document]
Comment (haishengwu-okta) [reply src/EnrollUserController.js]: It has some handy things helping create form inputs and footer. looks like it doesn't hurt since `FormController extends BaseLoginController`

File: src/models/ProfileSchema.js [hunk 27]
```diff
@@ -0,0 +1,46 @@
...
+define([
+  'okta'
+],
+function (Okta) {
+  var _ = Okta._;
+  var { BaseSchema } = Okta.internal.models;
+
+  return BaseSchema.Model.extend({
+    expand: ['schema'],
+    constructor: function () {
```
mauriciocastillosilva-okta: we don't need to override this since we are not doing anything extra

src/models/AppState.js
```diff
@@ -143,6 +143,7 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
     local: {
       baseUrl: 'string',
       lastAuthResponse: ['object', true, {}],
+      profileSchema: ['object', true, {}],
```
mauriciocastillosilva-okta: I don't think we need this attribute. From what I can see, we are storing a full api response here, which we already have in `lastAuthResponse`.

src/models/AppState.js
```diff
@@ -321,6 +334,15 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
           return res._links.next.href;
         }
       },
+      'profileSchemaAttributes': {
```
mauriciocastillosilva-okta: we have a `policy` attribute, that might contain already what you need and we can avoid this derived property
If it does not, this should be dependent on `lastAuthResponse` I think. Also the name is a little misleading since it contains the full registration object, not just the schema attributes.

nikhilvenkatraman-okta: ah, good point. I'll re-use that.

src/EnrollUserController.js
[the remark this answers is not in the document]
nikhilvenkatraman-okta: ok, I am leaving this as `BaseLoginController`

src/EnrollUserController.js
```diff
@@ -0,0 +1,56 @@
...
+  Okta,
+  EnrollUser,
+  BaseLoginController,
+  EnrollUserForm,
+  FooterWithBackLink,
+) {
+  var _ = Okta._;
+  return BaseLoginController.extend({
+    className: 'enroll-user',
+    initialize: function (options) {
```
mauriciocastillosilva-okta: just realized about this... sorry this is late in the review. This should probably be in `fetchInitialData`, which is part of the `BaseLoginController`. That function is called when we call render on the controller, and we only render after `fetchInitialData` promise is fulfilled. This is similar to [RegistrationController](https://github.com/okta/okta-signin-widget/blob/master/src/RegistrationController.js)

So we can return either a promise that is resolved immediatly if we are not in unauthenticated, or we can return the `getEnrollFormData` promise.

src/EnrollUserController.js
```diff
@@ -0,0 +1,56 @@
...
+      // create model
+      this.model = new EnrollUser(this.options);
+      // If user is unauthenticated and starts enroll flow make a post call to transition state to PROFILE_REQUIRED
+      if (this.options.appState.get('isUnauthenticated')) {
+        this.model.getEnrollFormData()
+          .then(_.bind(function (response) {
+            if (response && response.data) {
+              this.options.appState.setAuthResponse(response.data);
+              //this.options.appState.set('lastAuthResponse', response.data);
+              this.model.set('createNewAccount',
```
mauriciocastillosilva-okta: don't think we need to do this here.. our model has access to `appState`, so we can get the value directly from there.

src/EnrollUserController.js
```diff
@@ -0,0 +1,56 @@
...
+    initialize: function (options) {
+      this.options = options || {};
+      // create model
+      this.model = new EnrollUser(this.options);
+      // If user is unauthenticated and starts enroll flow make a post call to transition state to PROFILE_REQUIRED
+      if (this.options.appState.get('isUnauthenticated')) {
+        this.model.getEnrollFormData()
+          .then(_.bind(function (response) {
+            if (response && response.data) {
+              this.options.appState.setAuthResponse(response.data);
```
mauriciocastillosilva-okta: a more common pattern is to do `model.setTransaction(response.data)`.

src/EnrollUserController.js
```diff
@@ -30,21 +30,16 @@ function (
...
-              this.renderForm();
-            }
-          },this));
-      } else {
-        this.renderForm();
-      }
+      return this.model.getEnrollFormData()
+        .then(_.bind(function (response) {
+          if (response && response.data) {
+            this.model.setTransaction(response);
```
haishengwu-okta: why not move `setTransaction` to model since you moved the other logic?

src/models/Factor.js
```diff
@@ -241,18 +241,16 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
...
-          transaction.status === 'FACTOR_CHALLENGE' ||
           this.appState.get('promptForFactorInUnauthenticated')) {
           var factor = _.findWhere(transaction.factors, {
             id: this.get('id')
           });
           promise = factor.verify(data);
         }
 
-        // MFA_CHALLENGE
+        // MFA_CHALLENGE/ FACTOR_CHALLENGE
```
haishengwu-okta: 👍

src/EnrollUserController.js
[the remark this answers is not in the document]
haishengwu-okta: is FormController not working or you'd prefer to BaseLoginController? maybe add comment regarding your decision. (I see most of page use FormController.)

src/views/enrollUser/EnrollUserForm.js
```diff
@@ -0,0 +1,38 @@
...
+  return Okta.Form.extend({
+    layout: 'o-form-theme',
+    autoSave: true,
+    noCancelButton: true,
+    title: Okta.loc('registration.form.title', 'login'),
+    save: Okta.loc('registration.form.submit', 'login'),
+    initialize: function (options) {
+      this.options = options || {};
+      this.schema =
+      new ProfileSchema({ profileSchemaAttributes: this.options.appState.get('policy').registration.profile });
```
haishengwu-okta: the indent looks incorrect. no linting error?

test/unit/spec/MfaVerify_spec.js
```diff
@@ -114,7 +118,7 @@ function (Okta,
     'GENERIC_SAML': 14,
     'GENERIC_OIDC': 15
   };
-
+  /* eslint max-statements: [2, 50]*/
```
haishengwu-okta: isn't this override working https://github.com/okta/okta-signin-widget/blob/master/test/unit/.eslintrc.js#L10?

src/views/enrollUser/EnrollUserForm.js
nikhilvenkatraman-okta: hmmm, I din't get any lint errors

test/unit/helpers/xhr/FACTOR_CHALLENGE_email.js
```diff
@@ -46,6 +46,35 @@ define({
         "client": { "id": "0oat59f8prmWrik500g3", "name": "IDX", "_links": {} }
       }
     },
-    "_links": { "cancel": { "href": "http://foo.okta.com/api/v1/authn/cancel", "hints": { "allow": ["POST"] } } }
+    "_links": {
```
nikhilvenkatraman-okta: Updated this since the API is now sending a resend link

src/views/enrollUser/EnrollUserForm.js
nikhilvenkatraman-okta: @haishengwu-okta I don't have a strong opinion. I could add a es lint rule to allow it on the same line or use a variable `profile` to hold `this.options.appState.get('policy').registration.profile`.

src/models/EnrollUser.js
```diff
@@ -29,15 +30,23 @@ function (Okta, BaseLoginModel) {
...
+      if (this.appState.get('policy').registration.createNewAccount) {
         postData.registration['createNewAccount'] = true;
       }
       return postData;
     },
     getEnrollFormData: function () {
-      return this.manageTransaction(function (transaction) {
-        return transaction.enroll();
-      });
+      if (this.options.appState.get('isUnauthenticated')) {
```
mauriciocastillosilva-okta: I feel that this `if` belongs more in the controller, like you had it before. `model.getEnrollFormData` will always make a request to get that data, which sounds right

nikhilvenkatraman-okta: We dont have to make a call if its not unauthenticated, because the state would beI am using `isUnauthenticated ` to decide if a call needs to be made to fetch the data. If its not unauthenticated then i just resolve using `deferred.resolve(this.appState.get('lastAuthResponse'));`. If I move this check to the controller I feel like I would  have to add a check in the model to see if lastAuthResponse was `PROFILE_REQUIRED`

nikhilvenkatraman-okta: Actually I would have to move stuff into the controller 

``` 
fetchInitialData: function () {
      // If user is unauthenticated and starts enroll flow make a post call to transition state to PROFILE_REQUIRED
      if (this.options.appState.get('isUnauthenticated')) {
        return this.model.getEnrollFormData()
          .then(_.bind(function () {
            this.renderForm();
          }, this));
      } else {
        var deferred = Q.defer();
        deferred.resolve(this.options.appState.get('lastAuthResponse'));
        this.renderForm();
        return deferred.promise;
      }
    },
```
and the model would look like 
``
    getEnrollFormData: function () {
      return this.manageTransaction(function (transaction, setTransaction) {
        return transaction.enroll().then(function (trans) {
          setTransaction(trans);
        });
      });
    },

```
In my approach I was trying to abstract all this logic from the controller.

File: src/EnrollUserController.js [hunk 39]
```diff
@@ -0,0 +1,57 @@
...
+      this.options = options || {};
+      // create model
+      this.model = new EnrollUser(this.options);
+    },
+    fetchInitialData: function () {
+      // If user is unauthenticated and starts enroll flow make a post call to transition state to PROFILE_REQUIRED
+      if (this.options.appState.get('isUnauthenticated')) {
+        return this.model.getEnrollFormData()
+          .then(_.bind(function () {
+            this.renderForm();
```
mauriciocastillosilva-okta: I believe `BaseLoginController` will do `controller.render` after `fetchInitialData` succeeds, so maybe you don't have to be explicit about rendering the form here, and you just have to return the promise... you could give it a try

src/EnrollUserController.js
```diff
@@ -0,0 +1,57 @@
...
+          }, this));
+      } else {
+        var deferred = Q.defer();
+        deferred.resolve(this.options.appState.get('lastAuthResponse'));
+        this.renderForm();
```
mauriciocastillosilva-okta: and if my other comment is correct, then here you can just return `BaseLoginController.prototype.fetchInitialData.call`

src/views/enrollUser/EnrollUserForm.js
haishengwu-okta: it's fine to keep it for now. might be lint rule config inappropriately.

src/EnrollUserController.js
nikhilvenkatraman-okta: @mauriciocastillosilva-okta I was able to get it working by moving the rendering logic the logic inside `postRender` ?
```
    fetchInitialData: function () {
      // If user is unauthenticated and starts enroll flow make a post call to transition state to PROFILE_REQUIRED
      if (this.options.appState.get('isUnauthenticated')) {
        return this.model.getEnrollFormData();
      } else {
        var deferred = Q.defer();
        deferred.resolve(this.options.appState.get('lastAuthResponse'));
        return deferred.promise;
      }
    },
    postRender: function () {
      var form = new EnrollUserForm(this.toJSON());
      this.add(form);
      this.add(new FooterWithBackLink(this.toJSON()));
      this.addListeners();
    }
```

nikhilvenkatraman-okta: After debugging a little more, I found that i'll have to use `trapAuthResponse` when `isProfileRequired` is true.  When user clicks sign up in an `UNAUTHENTICATED` state, it switches the route to `signin/enroll-user`, then we do the fetch to get `PROFILE_REQUIRED` form data and eventually do a `setTransaction(trans)`, which calls routerUtil and tries to change the route to `signin/enroll-user` again.

PR #627, opened 2019-02-08
Post stateToken in request for passwordLessAuth flows

nikhilvenkatraman-okta:
- Currently for passwordLessAuth flows with widget does not post stateToken with the request. This is causing the widget to losethe stateToken it was bootstrapped with initially.
- As per my understanding, AuthJS takes care of adding the stateToken to the request as long as we call the right method.
- I found that the reason we are not posting stateToken for passwordLess auth flow is because 
the widget is calling [authClient.signIn](https://github.com/okta/okta-signin-widget/blob/master/src/models/PrimaryAuth.js#L116) which eventually calls 
 [this](https://github.com/nikhilvenkatraman-okta/okta-auth-js/blob/master/lib/clientBuilder.js#L175) method in authJS. (Note clientBuilder.js file is now removed)
- Ideally the widget should call the `transaction.authenticate` (authenticate comes back in the links object) and Auth JS will make sure the state Token gets added to the request [here](https://github.com/okta/okta-auth-js/blob/master/lib/tx.js#L208).

Resolves [OKTA-209004](https://oktainc.atlassian.net/browse/OKTA-209004)

@royalchan-okta @mauriciocastillosilva-okta @tthompson-okta @haishengwu-okta


src/models/PrimaryAuth.js
```diff
@@ -106,12 +106,21 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
...
+            var authClient = this.appState.settings.authClient;
+            return this.doPrimaryAuth(authClient, signInArgs, transaction.authenticate);
+          });
+        } else {
+          primaryAuthPromise = this.performTransaction(authClient, signInArgs);
+        }
+      }
+      else {
+        //normal username/password flow
         primaryAuthPromise = this.startTransaction(function (authClient) {
```
nikhilvenkatraman-okta: I am not sure if we need this else here, but I am leaving it here for now.

src/models/PrimaryAuth.js
```diff
@@ -129,6 +138,17 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
         }, this));
     },
 
+    performTransaction: function (authClient, signInArgs) {
+      this.doTransaction(function (transaction) {
+        // if its the newStateToken we need to post to login endpoint with key as identifier
```
royalchan-okta: nit: comments should be consistent with other naming, so `idxStateToken`

src/models/PrimaryAuth.js
```diff
@@ -141,6 +161,13 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
       if (!this.settings.get('features.passwordlessAuth')) {
         signInArgs.password = this.get('password');
       }
+
+      // if its an idx stateToken, we send the parameter as identifier to login API
+      if (this.appState.get('isIdxStateToken')) {
+        signInArgs.identifier = signInArgs.username;
+        delete signInArgs.username;
+        delete signInArgs.options;
```
royalchan-okta: maybe a comment on why we delete these options as well

src/models/PrimaryAuth.js
```diff
@@ -106,12 +106,21 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
...
-        });
-      } else {
+      var authClient = this.appState.settings.authClient;
+
+      if (this.appState.get('isUnauthenticated')) {
+        // bootstrapped with stateToken
+        if (!this.settings.get('features.passwordlessAuth')) {
+          primaryAuthPromise = this.doTransaction(function (transaction) {
+            var authClient = this.appState.settings.authClient;
+            return this.doPrimaryAuth(authClient, signInArgs, transaction.authenticate);
```
royalchan-okta: out of curiosity, why does this flow return immediately and not register all the fail and finish callbacks?

haishengwu-okta:
can we post stateToken regardless new or old pipeline?

src/models/PrimaryAuth.js
nikhilvenkatraman-okta: Its similar to other flows, where we return a promise which  `primaryAuthPromise` is set to and the  
success,fail callbacks are handled https://github.com/okta/okta-signin-widget/blob/98fbdc84e4898a71a70f9866a9ace972dc83fc0d/src/models/PrimaryAuth.js#L130

royalchan-okta:
@haishengwu-okta the new pipeline always expects a state token. 

The old pipeline actually handled a state token for almost all authn endpoints except identifier first with passwordless. Currently it looks safe to do so for identifier first for passwordless, but there's another JIRA open to investigate whether it's correctly using the token since it never expected one to be passed in.

src/models/PrimaryAuth.js
nikhilvenkatraman-okta: Looks like this flow is hit for login flow when widget is not bootstrapped with a stateToken.

src/models/AppState.js
```diff
@@ -254,6 +254,15 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
           return res._embedded.user.id;
         }
       },
+      'isIdxStateToken': {
+        deps: ['lastAuthResponse'],
+        fn: function (res) {
+          if (!res.stateToken) {
+            return null;
```
mauriciocastillosilva-okta: should this return `false`?

src/models/PrimaryAuth.js
mauriciocastillosilva-okta: we need this when there is no `stateToken` in the widget bootstrap config (at that point in time we don't have any response from the api yet, so we need this as the default login)

src/models/PrimaryAuth.js
```diff
@@ -106,12 +106,21 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
 
       var primaryAuthPromise;
 
-      if (this.appState.get('isUnauthenticated') && !this.settings.get('features.passwordlessAuth')) {
-        primaryAuthPromise = this.doTransaction(function (transaction) {
-          var authClient = this.appState.settings.authClient;
-          return this.doPrimaryAuth(authClient, signInArgs, transaction.authenticate);
-        });
-      } else {
+      var authClient = this.appState.settings.authClient;
```
mauriciocastillosilva-okta: this should probably be inside the `isUnauthenticated` if, since it is only used there

src/models/PrimaryAuth.js
```diff
@@ -106,12 +106,21 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
...
+
+      if (this.appState.get('isUnauthenticated')) {
+        // bootstrapped with stateToken
+        if (!this.settings.get('features.passwordlessAuth')) {
+          primaryAuthPromise = this.doTransaction(function (transaction) {
+            var authClient = this.appState.settings.authClient;
```
mauriciocastillosilva-okta: and we can get rid of this, since we already have it from line 109 (which should move inside the if isUnathenticated)

src/models/PrimaryAuth.js
```diff
@@ -106,12 +106,21 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
...
+            return this.doPrimaryAuth(authClient, signInArgs, transaction.authenticate);
+          });
+        } else {
+          primaryAuthPromise = this.performTransaction(authClient, signInArgs);
```
mauriciocastillosilva-okta: at this point, we know we are in UNAUTHENTICATED state and that we have passwordless enabled. This method should be named something like `doPrimaryAuthForPasswordless`

src/models/PrimaryAuth.js
mauriciocastillosilva-okta: yeah... or maybe there should be a main if/else block to add whichever data we need depending on `isIdxStateToken`, instead of deleting options we aded a few lines above

src/models/PrimaryAuth.js
nikhilvenkatraman-okta: yes, forgot to remove this

src/models/AppState.js
nikhilvenkatraman-okta: I think both would work, but i can change it to a boolean value to be consistent

haishengwu-okta:
@royalchan-okta I was thinking if widget could always post `stateToken` hence git rid of complicity at determine whether it's idx token from UI.

nikhilvenkatraman-okta:
@haishengwu-okta currently we use `isIdxStateToken` check  to post to a different API i.e `login` instead of `authenticate` during passwordLess flows, and could potentially use this in the future to do stuff specific for the new pipeline.

src/models/PrimaryAuth.js
mauriciocastillosilva-okta: Does the new pipeline supports primaryAuth without passwordless? If `isIdxStateToken` is true here, this will break since we need to call `translation.login` instead.

src/models/PrimaryAuth.js
```diff
@@ -106,12 +106,19 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
...
-        primaryAuthPromise = this.doTransaction(function (transaction) {
-          var authClient = this.appState.settings.authClient;
-          return this.doPrimaryAuth(authClient, signInArgs, transaction.authenticate);
-        });
-      } else {
+      if (this.appState.get('isUnauthenticated')) {
+        var authClient = this.appState.settings.authClient;
+        // bootstrapped with stateToken
+        if (!this.settings.get('features.passwordlessAuth')) {
+          primaryAuthPromise = this.doTransaction(function (transaction) {
```
haishengwu-okta: line 113-115 is identical to `doPrimaryAuthForPasswordless`. 
Did you miss some logic to pass `stateToken` or did I miss something here?

src/models/PrimaryAuth.js
```diff
@@ -129,14 +136,19 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
...
+
     getSignInArgs: function (username) {
       var multiOptionalFactorEnroll = this.get('multiOptionalFactorEnroll');
-      var signInArgs = {
-        username: username,
-        options: {
-          warnBeforePasswordExpired: true,
-          multiOptionalFactorEnroll: multiOptionalFactorEnroll
-        }
+      var signInArgs = {};
```
haishengwu-okta: looks identical with before. better to rollback?

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -114,10 +115,12 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
     return setup(settings, requests, true);
   }
 
-  function setupPasswordlessAuth (requests) {
-    return setup({ 'features.passwordlessAuth': true }, requests)
+  function setupPasswordlessAuth (requests, refreshState, isIdxToken) {
```
haishengwu-okta: don't need `isIdxToken`

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1956,6 +1959,28 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
             expect(test.form.el('factor-question').length).toEqual(1);
           });
       });
+      itp('calls transaction.authenticate with username, options and stateToken', function () {
+        return setupPasswordlessAuth(null, true, false).then(function (test) {
+          $.ajax.calls.reset();
+          test.form.setUsername('testuser@test.com');
+          test.form.submit();
+          return tick(test);
```
haishengwu-okta: `tick` is bad practice. try to use `waitFor*` utilies (https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/util/Expect.js)

src/models/PrimaryAuth.js
nikhilvenkatraman-okta: I can clean this up. But basically when bootstrapped with a stateToken with or without passwordless auth we need to call `transaction.authenticate`, and auth JS will make sure we are posting the stateToken

src/models/PrimaryAuth.js
```diff
@@ -106,12 +106,16 @@ function (Okta, BaseLoginModel, CookieUtil, Enums) {
...
+      if (this.appState.get('isUnauthenticated')) {
+        var authClient = this.appState.settings.authClient;
+        // bootstrapped with stateToken
+        // with or without passwordless auth we want to call transaction.authenticate
         primaryAuthPromise = this.doTransaction(function (transaction) {
-          var authClient = this.appState.settings.authClient;
           return this.doPrimaryAuth(authClient, signInArgs, transaction.authenticate);
         });
-      } else {
+      }
```
mauriciocastillosilva-okta: nit: we could keep this `else` without an extra line break...

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -91,9 +91,10 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
     var beacon = new Beacon($sandbox);
     router.on('afterError', afterErrorHandler);
     if (refreshState) {
+      var stateToken = 'aStateToken';
       Util.mockRouterNavigate(router);
       setNextResponse(resUnauthenticated);
-      router.refreshAuthState('aStateToken');
+      router.refreshAuthState(stateToken);
```
mauriciocastillosilva-okta: nit: this also could just remain as it was before, no other place is reusing the created `stateToken` variable

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1956,6 +1959,28 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
             expect(test.form.el('factor-question').length).toEqual(1);
           });
       });
+      itp('calls transaction.authenticate with username, options and stateToken', function () {
+        return setupPasswordlessAuth(null, true, false).then(function (test) {
```
mauriciocastillosilva-okta: what is the 3rd argument `false` for? I see only 2 arguments in `setupPasswordlessAuth` function

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -114,10 +115,12 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
...
-  function setupPasswordlessAuth (requests) {
-    return setup({ 'features.passwordlessAuth': true }, requests)
+  function setupPasswordlessAuth (requests, refreshState) {
+    return setup({ 'features.passwordlessAuth': true }, requests, refreshState)
       .then(function (test){
-        Util.mockRouterNavigate(test.router);
+        if (!refreshState) {
+          Util.mockRouterNavigate(test.router);
+        }
         test.setNextResponse(resPasswordlessUnauthenticated);
```
mauriciocastillosilva-okta: if `refreshState` is true, then `setup` will set the next response to `resUnauthenticated`, and then here we set up another next response to `resPasswordlessUnauthenticated`... should this be inside the `if`?

test/unit/spec/PrimaryAuth_spec.js
nikhilvenkatraman-okta: Not needed, removed. It was the holding a flag for `idxStateToken`.

test/unit/spec/PrimaryAuth_spec.js
nikhilvenkatraman-okta: I think this is needed (tried moving it inside the if, but got some test failures). `resPasswordlessUnauthenticated` is the response form next step of `passwordlessAuth`.

mauriciocastillosilva-okta: ah yeah... so `refreshState` is for when we have a stateToken, so we refresh the state before anything else in the flow. Then this `resPasswordlessUnauthenticated` is for when you actually submit the form with username/identifier only. All good 👍

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1956,6 +1958,28 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
             expect(test.form.el('factor-question').length).toEqual(1);
           });
       });
+      itp('calls transaction.authenticate with username, options and stateToken', function () {
```
mauriciocastillosilva-okta: I think we should rename this test to something that explicitly says we have stateToken from the config. Maybe `calls transaction.authenticate with stateToken when stateToken is in the bootstrap config`?

nikhilvenkatraman-okta: will update this to `calls transaction.authenticate with the same stateToken that the widget was bootstrapped with, in the config object`.

nikhilvenkatraman-okta:
merged as a part of https://github.com/okta/okta-signin-widget/pull/626

PR #628, opened 2019-02-08
feat: Adds basic general IDP option

swiftone:
We already support Identity Providers (IdP) from four hardcoded sources (Google, Facebook, Microsoft, and Linked In) - offering buttons with specific text and styling for IdPs of those specified "type"s.  

This PR makes the following changes:
* If an IDP is listed with an unknown type or no type, it is treated as a generic IDP (see below).  Previous behavior: unknown types were silently skipped
* Any IDP config can included a `className` property that will be added to the class property. Previous behavior: the known types all got their specific classes with no option to add others.  Those specific classes will still be applied after this change. 
* If a generic IDP has a `text` property, that text will be the text of the button.  If the `text` property is missing, default text is added to say they need to configure it.  Previous Behavior: The non-generic IDPs already have i18n defined text.   A later PR to add i18n-customization options for generic IDPs is expected.
```
    "idps": [
      { "type": "FACEBOOK", "id": "0oar25ZnMM5LrpY1O0g3"},
      { "id": "0oar25ZnMM5LrpY1O0g3", "text": "I'm a button!"},
      { "id": "totallyfakeidpidentr", "text": "I'm also a button!", "className": "demonstrated" },
      { "id": "totallyfakeidpidentr" }
    ],
```
screen shot 2019-02-20 at 8 34 28 am


src/models/Settings.js
```diff
@@ -287,7 +290,7 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
               dataAttr: dataAttr,
               className: 'social-auth-button ' + dataAttr,
               title: function () {
-                return Okta.loc('socialauth.' + type + '.label');
+                return idp.text || Okta.loc('socialauth.' + type + '.label');
```
haishengwu-okta: in your example, `type` could have space (am guessing allowing other characters). Would it still work when type has space?
I wonder if it'll make life easier to only allow characters and digits?

src/models/Settings.js
```diff
@@ -254,12 +254,15 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
...
-      // filters the idps passed into the widget to include only the ones we support.
+      // Adjusts the idps passed into the widget based on if they get explicit support
       configuredSocialIdps: {
         deps: ['idps'],
         fn: function (idps) {
-          return _.filter(idps, function (idp) {
-            return _.contains(supportedIdps, idp.type.toLowerCase());
+          return _.map(idps, function (idp) { 
+            if( ! ( idp.type && _.contains(supportedIdps, idp.type.toLowerCase()) )) {
+              idp.type = 'general-idp';
```
haishengwu-okta: Am I reading wrong that the logic here is to override `type` to `general-idp` as long as not one of those 4 supported types? Then what's the point allow developer to specify `type`.

src/models/Settings.js
swiftone: Ah!  Yes, I established what the criteria would be...and then utterly failed to enforce them.  Good catch.

src/models/Settings.js
swiftone: Largely to have something to hang the future expected features on - custom CSS classes, potential i18n keys, etc, but it could be something that gets dropped.  Easier to have customers keep using something they have than to retroactively add it after dropping it.  @nbarbettini  - any thoughts?

src/models/Settings.js
swiftone: Oh yes - I didn't enforce since I'm not using the type yet, so this also falls under the other question (re: wherefore type)

src/models/Settings.js
```diff
@@ -279,15 +288,12 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
...
           _.each(configuredSocialIdps, function (idp) {
-            var type = idp.type.toLowerCase();
-            var dataAttr = 'social-auth-' + type + '-button';
             var socialAuthButton = {
               id: idp.id,
-              type: idp.type,
-              dataAttr: dataAttr,
-              className: 'social-auth-button ' + dataAttr,
+              dataAttr: idp.dataAttr,
+              className: 'social-auth-button ' + idp.className,
```
haishengwu-okta: shall move the `className` concatenation to line 267?

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1996,56 +2000,119 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
...
+          expect(buttons.eq(1).hasClass('example-class')).toBe(false);
+          expect(buttons.eq(1).hasClass('other-class')).toBe(false);
+
+          expect(buttons.eq(2).hasClass('social-auth-google-button')).toBe(false);
+          expect(buttons.eq(2).hasClass('social-auth-facebook-button')).toBe(false);
+          expect(buttons.eq(2).hasClass('example-class')).toBe(false);
+          expect(buttons.eq(2).hasClass('other-class')).toBe(true);
+        });
+      });
+      itp('displays generic idp buttons', function () {
```
haishengwu-okta: nit:
more explicitly like `displays generic idp buttons for not default supported type`?

src/models/Settings.js
swiftone: After talking with Nate - type is optional and is no longer mentioned for generic IDPs, and I've added a `className` property to allow for custom styling.  i18n will be dealt with later when we have a consistent approach.

swiftone: Revised: type is no longer required.  Unknown types will be treated just like absent types (though we might add some console logging to that effect in a later round).  Type is meaningful only for the existing four hardcoded options.

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1996,56 +2000,119 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
...
         };
         return setup(settings).then(function (test) {
-          expect(test.form.hasSocialAuthDivider()).toBe(true);
-          expect(test.form.socialAuthButtons().length).toBe(4);
-          expect(test.form.facebookButton().length).toBe(1);
-          expect(test.form.googleButton().length).toBe(1);
-          expect(test.form.microsoftButton().length).toBe(1);
-          expect(test.form.linkedInButton().length).toBe(1);
+          var buttons = test.form.socialAuthButtons();
+          expect(buttons.eq(0).hasClass('social-auth-linkedin-button')).toBe(true);
```
manueltanzi-okta: nit: `.toHaveClass` (also in the other places)

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1996,56 +2000,119 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
...
+          idps: [
+            {
+              id: '0oaidiw9udOSceD1111',
+              text: 'Not default text'
+            }
+          ]
+        };
+        return setup(settings).then(function (test){
+          var buttons = test.form.socialAuthButtons();
+          expect(buttons.eq(0).text()).toBe('Not default text');
```
manueltanzi-okta: nit: `.toHaveText` (also in the other places)

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1228,6 +1262,27 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm,
       });
     });
     Expect.describe('Additional Auth Button', function () {
+      itp('does not display custom buttons when it is undefined', function () {
+        return setupWithoutCustomButtonsAndWithIdp().then(function (test){
+          expect(test.form.authDivider().length).toBe(1);
```
manueltanzi-okta: nit: `. toHaveLength ` (also in the other places)

src/models/Settings.js
```diff
@@ -254,12 +254,25 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
...
+          return _.map(idps, function (idp) { 
+            var type = idp.type && idp.type.toLowerCase();
+            if ( !( type && _.contains(supportedIdps, type) ) ) {
+              type = 'general-idp';
+              idp.text = idp.text || '{ Please provide a text value }';
+            }
+
+            idp.className = [
+              'social-auth-button',
+              type ? 'social-auth-' + type + '-button ' : '',
```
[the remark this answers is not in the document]
brettritter-okta: good catch

PR #629, opened 2019-02-14
Send device fingerprint only if cookie is not present in getImage call.

magizh-okta:
* Currently to retrieve the Beacon image in SIW, we send the device fingerprint and
cookie. If cookie is present we can avoid computing and sending the device fingerprint.(saves 2 api calls).

* Verified if getimage has fingerprint in (first login and expired cookie case) and not when proximity cookie is present.

* authn call still sends device fingerprint and remains unaffected.

* Introduced a feature flag in SIW to tell if we need to compute device fingerprint for getImage call.

**Auth n call has device fingerprint**
screen shot 2019-02-14 at 1 21 05 pm

**First Get image call - device fingerprint is set when there is no proximity cookie** 
screen shot 2019-02-14 at 1 21 30 pm

**Get image call when proximity cookie is present**
screen shot 2019-02-14 at 1 21 46 pm

e2e flow - first login and next login 
siw

### DIFF
--- a/src/PrimaryAuthController.js
+++ b/src/PrimaryAuthController.js
@@ -18,7 +18,7 @@ define([
   'models/PrimaryAuth',
   'views/shared/Footer',
   'util/BaseLoginController',
-  'util/DeviceFingerprint'
+  'util/DeviceFingerprint',
 ],
 function (Okta, PrimaryAuthForm, CustomButtons, FooterRegistration, PrimaryAuthModel,
   Footer, BaseLoginController, DeviceFingerprint) {
@@ -76,8 +76,8 @@ function (Okta, PrimaryAuthForm, CustomButtons, FooterRegistration, PrimaryAuthM
 
     events: {
       'focusout input[name=username]': function () {
-        // Get DeviceFingerprint so we can use it for security image and for primary auth
-        if (this.settings.get('features.deviceFingerprinting')) {
+        if (this.settings.get('features.deviceFingerprinting') &&
+            this.settings.get('features.useDeviceFingerprintForSecurityImage')) {
           var self = this;
           DeviceFingerprint.generateDeviceFingerprint(this.settings.get('baseUrl'), this.$el)
             .then(function (fingerprint) {
--- a/test/unit/spec/IDPDiscovery_spec.js
+++ b/test/unit/spec/IDPDiscovery_spec.js
@@ -554,7 +554,8 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, 
     });
 
     Expect.describe('Device Fingerprint', function () {
-      itp('contains fingerprint header in get security image request if feature is enabled', function () {
+      itp(`contains fingerprint header in get security image request if deviceFingerprinting
+        is true (useDeviceFingerprintForSecurityImage defaults to true)`, function () {
         spyOn(DeviceFingerprint, 'generateDeviceFingerprint').and.callFake(function () {
           var deferred = Q.defer();
           deferred.resolve('thisIsTheDeviceFingerprint');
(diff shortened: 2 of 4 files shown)


src/util/CookieUtil.js
```diff
@@ -32,5 +32,13 @@ define(['okta'], function (Okta) {
     Cookie.removeCookie(LAST_USERNAME_COOKIE_NAME, { path: '/' });
   };
 
+  fn.isProximityCookiePresent = function () {
+    try {
+      return document.cookie && document.cookie.indexOf('proximity_') !== -1;
+    } catch(e) {
```
haishengwu-okta: what exception could happen or merely safe check for whatever reason?

src/PrimaryAuthController.js
```diff
@@ -18,7 +18,7 @@ define([
   'models/PrimaryAuth',
   'views/shared/Footer',
   'util/BaseLoginController',
-  'util/DeviceFingerprint'
+  'util/DeviceFingerprint',
```
TianGan-okta: we may not need comma here

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -560,7 +560,8 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm,
           deferred.resolve('thisIsTheDeviceFingerprint');
           return deferred.promise;
         });
-        return setup({ features: { securityImage: true, deviceFingerprinting: true }})
+        return setup({ features: { securityImage: true, deviceFingerprinting: true,
```
mauriciocastillosilva-okta: we have tests for 
- `deviceFingerprinting: true` and `useDeviceFingerprintForSecurityImage: true`
- `deviceFingerprinting: false` and `useDeviceFingerprintForSecurityImage: false`

I think we should add one more here for 
- `deviceFingerprinting: true` and `useDeviceFingerprintForSecurityImage: false`

that way we cover the 3 scenarios in this test (similar to the other test you are changing)

README.md
```diff
@@ -1329,6 +1329,8 @@ features: {
 
 - **features.showPasswordToggleOnSignInPage** - End users can now toggle visibility of their password on the Okta Sign-In page, allowing end users to check their password before they click Sign In. This helps prevent account lock outs caused by end users exceeding your org's permitted number of failed sign-in attempts. Note that passwords are visible for 30 seconds and then hidden automatically. Defaults to `false`.
 
+- **features.useDeviceFingerprintForSecurityImage** - Uses device fingerprint to retrieve security image. Useful in cases where cookies are cleared off when the app is closed (e.g. embedded browsers). Defaults to `false`.
```
mauriciocastillosilva-okta: should we mention that this feature depends on `deviceFingerprinting`? (which I don't think is documented here...)

cc: @robertdamphousse-okta we are adding a new feature to the list

test/unit/spec/IDPDiscovery_spec.js
haishengwu-okta: what about     
* `deviceFingerprinting: false` and `useDeviceFingerprintForSecurityImage: true`

magizh-okta: @mauriciocastillosilva-okta  @haishengwu-okta  I have the same cases tested in Primary_auth spec. 
Dint want to repeat all the flows for idp since it would go through the same flow. 

https://github.com/okta/okta-signin-widget/pull/629/files#diff-dc06304da4eb7ee5361d32a9f5850bb0R973

mauriciocastillosilva-okta: IdpDiscoveryForm extends from PrimaryAuthForm, so you have a point. We already have tests that are somehow duplicated for both specs, which is not ideal. Probably we should look into refactoring these tests in the future to remove all the dups.
But with the current state of things, it won't hurt having the tests in both specs...

src/PrimaryAuthController.js
magizh-okta: Not a problem. We add trailing commas in these cases so when we add the next item we dont forget to add the comma. Linter is passing so no worries I guess.

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -554,7 +554,8 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm,
     });
 
     Expect.describe('Device Fingerprint', function () {
-      itp('contains fingerprint header in get security image request if feature is enabled', function () {
+      itp(`contains fingerprint header in get security image request if both features(
```
mauriciocastillosilva-okta: nit: can we duplicate this test so we can have
- explicitly set `useDeviceFingerprintForSecurityImage=true`
- check that the default of `useDeviceFingerprintForSecurityImage` is true

so basically
- "contains fingerprint header in get security image request if both features are enabled"
- "contains fingerprint header in get security image request if deviceFingerprinting is true (deviceFingerprinting defaults to true)`

PR #635, opened 2019-02-22
Adding tooltip to identifier first widget

petermiller-okta:
Resolves: OKTA-205616

Adding tooltip to username field. Password field tooltip already exists

username
password

@okta/idx


petermiller-okta:
Would like to get @nikhilvenkatraman-okta's approval as well...and then a merge please :)

Issue #631, opened 2019-02-20
Google Idp button icon


alexpx:
:information_source: **If you have a question**, please post it on the [Okta Developer Forum](https://devforum.okta.com) instead. Issues in this repository are reserved for bug reports and feature requests.

### I'm submitting a
  - [*] bug report
  - [ ] feature request

### Background info

Right now the Google Idp button icon has the Google plus icon.
Since this API is deprecated and you have already released a fix for that in the Okta OIDC flow, I expect the icon to be changed as well.

### Steps to reproduce

Just add Google as Idp provider and check the button  icon

### Your environment

  - Okta Sign-In Widget Version: 2.15.0

haishengwu-okta:
@alexpx thanks for reporting and we're working on. stay tune.

PR #636, opened 2019-02-25
Fix google-icon

haishengwu-okta:
- Replace google+ icon with google brand icon.
- Fixes https://github.com/okta/okta-signin-widget/issues/631
- Screenshot

localhost_3000_


assets/sass/_variables.scss
```diff
@@ -49,9 +49,9 @@ $form-divider-color: #e8e8e8;
 $facebook-icon-bg-color: #304879; // fb brand color - #3B5998
 $facebook-label-bg-color: #406094;
 $facebook-border-color: #375685;
-$google-icon-bg-color: #d43722; // google brand color - #DC4E41
-$google-label-bg-color: #de4937;
-$google-border-color: #b63c33;
+$google-icon-bg-color: #fff;
+$google-label-bg-color: $button-bg-color;
```
manueltanzi-okta: imo we should not use the generic variable button background color, since someone that change that variable value to customize widget style doesn't probably want to also affect the Google idp button, which is more an ad hoc style for Google

haishengwu-okta: good question. I wouldn't be worry about it for now because
- generic IdP uses it as well
- there is an new design for all IdP buttons (OKTA-199876)

santhoshbalakrishnan:
@haishengwu-okta Is the google logo on a white bg but the button has a gray bg? Can we get a transparent Google logo?

santhoshbalakrishnan:
Also, can we use SVGs? https://codepen.io/bsanth/pen/BbaZWW

I got the assets from here: https://developers.google.com/identity/branding-guidelines

haishengwu-okta:
@santhoshbalakrishnan 
- The bg-color of google icon is white.
- nice suggestion and finding on the assets!
  I'd defer to @natemacinnes-okta to see if want to move forward or not since we do have an redesign for all social buttons. (OKTA-199876)

nbarbettini:
SVGs are nice but PNGs are okay too.

@haishengwu-okta  We should get as close as possible to Google's branding guidelines here since a future redesign is at least a few months out.

haishengwu-okta:
Update:
- push another commit to use image from google standard assets.
- video: https://okta.box.com/s/8gxwsyrshfifjrgot8prr5q0zolda8wy

santhoshbalakrishnan:
Can we not still build our own button? I feel like this new button does not fit well with the widget. The branding guideline allows us to use our own buttons, we do not have to strictly use their button in the asset. We can use the Google logo and then create our own button. If we use the Google button, it cannot be i18n'ed. Here is their doc about building a custom button: https://developers.google.com/identity/sign-in/web/build-button

haishengwu-okta:
@santhoshbalakrishnan we build our own google-sign-in button, it doesn't get font hence it doesn't look quite right. Also I doubt we want to load the font just for the button atm.
google-icon-css

santhoshbalakrishnan:
@haishengwu-okta Your box video shows a different button? - which is why I posted that comment.

haishengwu-okta:
@santhoshbalakrishnan 
the video shows the solution using image but I drop the custom button which doesn't get font.

nbarbettini:
I agree with the direction of building the button with CSS instead of an image. 

If you have any more cycles @haishengwu-okta, it would be nice to:
* Use dark gray (rather than black) text to more closely match Google's example
* Use a gray drop-shadow behind the button rather than a black border (also to match their example)

haishengwu-okta:
Given another look at 2nd example of https://developers.google.com/identity/sign-in/web/build-button,
it pretty much create its own style except using the google logo.
Therefore, here is another update
1. Use same style of Custom IdP button as Alex D suggested in the JIRA
2. Change `bg-color` to be white all the time.
3. Use logo from google: https://developers.google.com/identity/images/g-logo.png

widget-fix-google-icon-v3

alexdahl-okta:
Catching up on this: how did we end up with styles and assets different from [OKTA-199876](https://oktainc.atlassian.net/browse/OKTA-199876?focusedCommentId=800384&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-800384)?

haishengwu-okta:
@alexdahl-okta because of https://developers.google.com/identity/branding-guidelines

haishengwu-okta:
google-icon-20x20

santhoshbalakrishnan:
@alexdahl-okta Overall looks good, but a couple of questions:

1. Are we using the assets from Google's press kit? 
2. There is a white background behind the `G` logo. Is that intended? Can that be transparent or can we make the button white?

Asking this since social logins are more consumer facing, do they expect to see a familiar Google sign in button?

alexdahl-okta:
@santhoshbalakrishnan 

1. Yep, exported from Google's sign in assets.
2. White background behind the G is intended, and required by Google. But a light gray button background is a better fit with our other buttons (two-tone for all, same button color as generic).

It's an interesting balance – end-users should be able to quickly identify a provider's branding, but should also be able to understand that all the buttons are part of a set (and have a cohesive visual experience). For this one specifically, it's extremely similar to Google's default button, so I don't expect end-users to have trouble recognizing it.

After this quick fix, we'll eventually do another pass to redesign all providers' buttons more cohesively.

assets/sass/_variables.scss
```diff
@@ -49,9 +49,9 @@ $form-divider-color: #e8e8e8;
...
+$google-border-color: darken($button-bg-color, 22%);
```
santhoshbalakrishnan: Can we actually get a hex here? It's easier to categorize and switch to DS when we have it.

santhoshbalakrishnan:
Thank for the explanation @alexdahl-okta

assets/sass/_variables.scss
haishengwu-okta: I'm not too worry about that now since it's probably got an redesign before DS hit SIW.

Issue #639, opened 2019-02-28
Update handlebars


Z3r021:
### Background info

NPM is reporting a vulnerability with handlebars.

### Expected behavior

handlebars updated to >- 4.0.13

### What went wrong?

=== npm audit security report ===
High - Prototype Pollution
Package: handlebars
Patched In: >=4.0.13
Dependency Of: @okta/okta-signin-widget   
Path: @okta/okta-signin-widget > handlebars 
More Info: https://npmjs.com/advisories/755   

### Your environment

  - Okta Sign-In Widget Version: 2.15.0

swiftone:
Thanks for the note, we'll put this update on our list.

Z3r021:
@swiftone, any speculation on target for next release? Trying to determine if we go with a forked version for now or wait, if the release is not too far off. I am new to contributing, so their may be something posted somewhere answer this, I was not able to find it if there was one. :)

fuzzthink:
When will 2.18 be published to npm? Unable to install handlebars fix without version published to npm.

swiftone:
@fuzzthink A teammate is literally tweaking the release notes now, so while it might take some time to trickle through through our CI process, 2.17 (not 2.18) which will have this change should be on npm Very Soon (tm).

haishengwu-okta:
@fuzzthink `2.17` has been published. let us know if any more questions.
https://www.npmjs.com/package/@okta/okta-signin-widget/v/2.17.0

PR #640, opened 2019-02-28
chore handlebars: bumps version to 4.0.13

Z3r021:
fixes vulnerability in handlebars

resolves: #639


Z3r021:
I am not certain on the versioning strategy for okta-signin-widget. Out of habit I had updated it (using the strategy we use), but reverted it.

PR #642, opened 2019-03-05
feat: Adds afterError events for OAUTH_ERROR and REGISTRATION_FAILED

manueltanzi-okta:
### Description

- Remove global error handler for OAUTH_ERROR and REGISTRATION_FAILED in favor of afterError events

**Note:** This PR contains mostly breaking changes, this is why it’s a draft PR. Will be merged on the release of okta-signing-widget 3.0.

### Reviewers

@haishengwu-okta @jmelberg-okta @robertjd @nbarbettini

### DIFF
--- a/src/IDPDiscoveryController.js
+++ b/src/IDPDiscoveryController.js
@@ -44,10 +44,12 @@ function (Okta, PrimaryAuthController, PrimaryAuthModel, IDPDiscoveryForm, IDPDi
       // If social auth is configured, 'socialAuthPositionTop' will determine
       // the order in which the social auth and primary auth are shown on the screen.
       if (options.settings.get('hasConfiguredButtons')) {
-        // CustomButtons needs current controller as parameter
         this.add(CustomButtons, {
           prepend: options.settings.get('socialAuthPositionTop'),
-          options: { currentController: this }
+          options: {
+            // To trigger an afterError event, we require the current controller
+            currentController: this
+          }
         });
       }
 
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -160,7 +160,7 @@ function (
       });
 
       //throw registration error
-      var errMsg = error.callback? error.callback+':'+ error.errorSummary: error.errorSummary;
+      var errMsg = error.callback ? error.callback + ':' + error.errorSummary : error.errorSummary;
       Util.triggerAfterError(this, new Errors.RegistrationError(errMsg));
 
       if (hideRegisterButton) {
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -17,7 +17,8 @@ define([
   'util/Enums',
   'util/RegistrationFormFactory',
   'views/registration/SubSchema',
-  'util/Errors'
+  'util/Errors',
+  'util/Util'
 ],
 function (
   Okta,
@@ -27,7 +28,8 @@ function (
   Enums,
   RegistrationFormFactory,
   SubSchema,
-  Errors
+  Errors,
+  Util
 ) {
 
   var { _, Backbone } = Okta;
@@ -110,7 +112,7 @@ function (
         var responseJSON = err.responseJSON;
         if (responseJSON && responseJSON.errorCauses.length) {
           var errMsg = responseJSON.errorCauses[0].errorSummary;
-          self.settings.callGlobalError(new Errors.RegistrationError(errMsg));
+          Util.triggerAfterError(self, new Errors.RegistrationError(errMsg));
         }
       });
     },
@@ -157,9 +159,9 @@ function (
         responseJSON: error
       });
 
-      //throw global error
+      //throw registration error
       var errMsg = error.callback? error.callback+':'+ error.errorSummary: error.errorSummary;
-      this.settings.callGlobalError(new Errors.RegistrationError(errMsg));
+      Util.triggerAfterError(this, new Errors.RegistrationError(errMsg));
 
       if (hideRegisterButton) {
         this.$el.find('.button-primary').hide();
--- a/src/util/OAuth2Util.js
+++ b/src/util/OAuth2Util.js
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-define(['okta', './Enums', './Errors'], function (Okta, Enums, Errors) {
+define(['okta', './Enums', './Errors', './Util'], function (Okta, Enums, Errors, Util) {
 
   var util = {};
   var _ = Okta._;
@@ -42,9 +42,8 @@ define(['okta', './Enums', './Errors'], function (Okta, Enums, Errors) {
       if (error.errorCode === 'access_denied') {
         controller.model.trigger('error', controller.model, {'responseJSON': error});
         controller.model.appState.trigger('removeLoading');
-      } else {
-        settings.callGlobalError(new Errors.OAuthError(error.message));
       }
+      Util.triggerAfterError(controller, new Errors.OAuthError(error.message), settings);
     }
 
     var authClient = settings.getAuthClient(),
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -146,7 +146,7 @@ define(['okta', './Logger', './Enums'], function (Okta, Logger, Enums) {
     var error = _.pick(err, 'name', 'message', 'statusCode', 'xhr');
     controller.trigger('afterError', { controller: className }, error);
     // Logs to console only in dev mode
-    Logger.info('controller: ' + className + ', error: ' + error);
+    Logger.warn('controller: ' + className + ', error: ' + error);
   };
 
   /**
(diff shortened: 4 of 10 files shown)


src/models/Settings.js
```diff
@@ -303,7 +303,7 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
               click: function (e) {
                 e.preventDefault();
                 if (self.get('oauth2Enabled')) {
-                  OAuth2Util.getTokens(self, {idp: idp.id});
+                  OAuth2Util.getTokens(self, {idp: idp.id}); // TODO
```
manueltanzi-okta: @haishengwu-okta @jmelberg-okta Question: this one does not have/pass a Controller. What you guys think we should use in this case?

jmelberg-okta: Given that OAuth errors can occur in a variety of cases, it might make sense to  map all OAuth errors to a controller named `oauth`. I'm thinking of the following scenarios:

1. User is going through the authentication/authorization flow for the first time:
    - The primary auth page is rendered and an error occurs because the user is **not** assigned to the application. While this happens on the primary auth page, we're technically redirecting back from Okta in an error state.
2. The user's token is no longer valid, and we attempt to renew it via `signIn.tokenManager.renew`. If an error occurs in that state, we technically **do not** have a controller present.

src/util/OAuth2Util.js
```diff
@@ -42,9 +42,8 @@ define(['okta', './Enums', './Errors'], function (Okta, Enums, Errors) {
       if (error.errorCode === 'access_denied') {
         controller.model.trigger('error', controller.model, {'responseJSON': error});
         controller.model.appState.trigger('removeLoading');
-      } else {
-        settings.callGlobalError(new Errors.OAuthError(error.message));
       }
+      Util.triggerAfterError(controller, new Errors.OAuthError(error.message), settings);
```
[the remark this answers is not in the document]
manueltanzi-okta: Good point! What does @nbarbettini thinks about this?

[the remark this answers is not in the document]
nbarbettini: We should differentiate between configuration errors (usually at startup time), and runtime errors. If I configure the Widget with a missing client ID or an empty scopes array, I would expect an exception to be thrown early.

src/models/Settings.js
haishengwu-okta: iirc, `configuredButtons` are used at either primary auth or identity first auth page. and curious what is `this` context at this click function and if could get something out of `this`?

manueltanzi-okta: Created https://github.com/okta/okta-signin-widget/pull/652 to move logic for creating button options.

@haishengwu-okta `this` context has a className parameter with this kind of value: `"social-auth-button social-auth-facebook-button`.
Unfortunately, `this` context does not have the controller.

manueltanzi-okta: Updated https://github.com/okta/okta-signin-widget/pull/652 so we can have a controller :D

src/util/OAuth2Util.js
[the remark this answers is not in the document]
manueltanzi-okta: As per our discussion, added a dev logger

src/IDPDiscoveryController.js
```diff
@@ -44,7 +44,11 @@ function (Okta, PrimaryAuthController, PrimaryAuthModel, IDPDiscoveryForm, IDPDi
       // If social auth is configured, 'socialAuthPositionTop' will determine
       // the order in which the social auth and primary auth are shown on the screen.
       if (options.settings.get('hasConfiguredButtons')) {
-        this.add(CustomButtons, {prepend: options.settings.get('socialAuthPositionTop')});
+        // CustomButtons needs current controller as parameter
```
jmelberg-okta: nit: I'd suggest moving this comment inside of the `options` object, since it only pertains to the `afterError` event.

Something like:

```js
this.add(CustomButtons, {
  options: {
    // To trigger an afterError event, we require the current controller
    currentController: this 
  }
});
```

src/RegistrationController.js
```diff
@@ -157,9 +159,9 @@ function (
         responseJSON: error
       });
 
-      //throw global error
-      var errMsg = error.callback? error.callback+':'+ error.errorSummary: error.errorSummary;
-      this.settings.callGlobalError(new Errors.RegistrationError(errMsg));
+      //throw registration error
+      var errMsg = error.callback ? error.callback + ':' + error.errorSummary : error.errorSummary;
```
jmelberg-okta: Thank you for this! 😄

src/models/Settings.js
```diff
@@ -286,55 +284,10 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, OAuth2Util, config) {
...
-          return buttonArray;
-        },
-        cache: true
-      },
-      //checks if there are any configured buttons
-      hasConfiguredButtons: {
-        deps: ['configuredButtons'],
-        fn: function (configuredButtons) {
-          return !_.isEmpty(configuredButtons);
+          return !_.isEmpty(configuredSocialIdps) || !_.isEmpty(customButtons);
```
jmelberg-okta: Is this an intended change for this PR? Or is this for #652?

src/util/Util.js
```diff
@@ -145,6 +145,8 @@ define(['okta', './Logger', './Enums'], function (Okta, Logger, Enums) {
     var className = _.isFunction(controller.className) ? controller.className() : controller.className;
     var error = _.pick(err, 'name', 'message', 'statusCode', 'xhr');
     controller.trigger('afterError', { controller: className }, error);
+    // Logs to console only in dev mode
+    Logger.info('controller: ' + className + ', error: ' + error);
```
jmelberg-okta: Shouldn't this be using the `Logger.error` property instead of `info`?

src/models/Settings.js
manueltanzi-okta: Correct, need to rebase! I just merged today 👍

src/util/Util.js
manueltanzi-okta: Unfortunately, `Logger.error` logs also in prod: https://github.com/okta/okta-signin-widget/blob/master/src/util/Logger.js#L18

src/RegistrationController.js
manueltanzi-okta: 🎉

src/util/Util.js
jmelberg-okta: Want to meet in the middle and do `Logger.warn`?

manueltanzi-okta: ook! 👍

haishengwu-okta: logger.warn or logger.log

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -84,13 +88,45 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm,
...
+          id: '0oaidiw9udOSceD1234'
+        }
+      ]
+    }, settings))
+      .then(function (test) {
+        spyOn(window, 'open').and.callFake(function () {
+          test.oidcWindow = {closed: false, close: jasmine.createSpy()};
+          return test.oidcWindow;
+        });
+        return tick(test);
```
haishengwu-okta: use `Expect.wait*` instead of `tick`

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -1410,14 +1446,44 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm,
...
+            var callback = args[1];
+            callback.call(null, {
+              origin: 'https://foo.com',
+              data: {
+                state: OIDC_STATE,
+                error: 'OAuth Error',
+                error_description: 'Message from server'
+              }
+            });
+            return tick(test);
```
haishengwu-okta: use `Expect.wait*` instead of `tick`

test/unit/spec/LoginRouter_spec.js
```diff
@@ -1141,14 +1140,19 @@ function (Okta, Q, Logger, Errors, BrowserFeatures, WidgetUtil,
...
-            expect(errorSpy.calls.count()).toBe(1);
-            var err = errorSpy.calls.argsFor(0)[0];
-            expect(err instanceof Errors.OAuthError).toBe(true);
-            expect(err.name).toBe('OAUTH_ERROR');
-            expect(err.message).toBe('Invalid value for client_id parameter.');
+          .then(function (test) {
+            expect(test.afterErrorHandler).toHaveBeenCalledTimes(1);
+            expect(test.afterErrorHandler.calls.allArgs()[0]).toEqual([
+              {
+                controller: 'mfa-verify'
```
haishengwu-okta: curious why it's `mfa-verify`

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -2469,14 +2468,19 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
                 error_description: 'Message from server'
               }
             });
-            return tick();
+            return tick(test);
```
haishengwu-okta: refactor to `Expect.wait*`

test/unit/spec/LoginRouter_spec.js
manueltanzi-okta: Because `setupOAuth2` goes use mfa-verify https://github.com/okta/okta-signin-widget/blob/54bee338d94c35bc3709f69c403090cf645e7ee6/test/unit/spec/LoginRouter_spec.js#L132

PR #643, opened 2019-03-06
test: Adds tests for multiple u2f

gowthamidommety-okta:
Adds tests for last used factor and multiple u2f when another factor is set up

Resolves: OKTA-212651

### DIFF
--- a/test/unit/spec/MfaVerify_spec.js
+++ b/test/unit/spec/MfaVerify_spec.js
@@ -1,4 +1,4 @@
-/* eslint max-params: [2, 50], max-statements: [2, 49], camelcase: 0 */
+/* eslint max-params: [2, 50], max-statements: [2, 50], camelcase: 0 */
 define([
   'okta',
   'q',
@@ -22,6 +22,7 @@ define([
   'helpers/xhr/MFA_REQUIRED_oktaPassword',
   'helpers/xhr/MFA_REQUIRED_U2F',
   'helpers/xhr/MFA_REQUIRED_multipleU2F',
+  'helpers/xhr/MFA_REQUIRED_multipleFactors',
   'helpers/xhr/MFA_CHALLENGE_duo',
   'helpers/xhr/MFA_CHALLENGE_sms',
   'helpers/xhr/MFA_CHALLENGE_call',
@@ -68,6 +69,7 @@ function (Okta,
   resPassword,
   resU2F,
   resMultipleU2F,
+  resMultipleFactors,
   resChallengeDuo,
   resChallengeSms,
   resChallengeCall,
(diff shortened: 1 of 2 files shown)


test/unit/spec/MfaVerify_spec.js
```diff
@@ -4144,6 +4154,152 @@ function (Okta,
         });
       });
 
+      Expect.describe('Multiple Security Keys (U2F) are setup with Security Question', function () {
```
mauriciocastillosilva-okta: should we name this `Multiple Security Keys (U2F) and one more factor are setup` since we don't really care if it is security question or something else?

test/unit/helpers/xhr/MFA_REQUIRED_multipleU2FmultipleFactors.js
```diff
@@ -0,0 +1,114 @@
+define({
```
mauriciocastillosilva-okta: do we have a `MFA_REQUIRED_multipleU2F.js` already? or can we name this file just that?

gowthamidommety-okta: We have that file. It's for setup with only multiple u2f

test/unit/spec/MfaVerify_spec.js
gowthamidommety-okta: 👍

test/unit/spec/MfaVerify_spec.js
```diff
@@ -4144,6 +4154,152 @@ function (Okta,
...
+        itp('shows the right beacon for Security Key (U2F)', function () {
+          return setupU2F({u2f: true, initResponse: resMultipleU2FMultipleFactors, multipleU2F: true}).then(function (test) {
```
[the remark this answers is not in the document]
gowthamidommety-okta: Yea.. i was debating that. I'll create one

test/unit/spec/MfaVerify_spec.js
```diff
@@ -288,9 +290,9 @@ function (Okta,
         delete window.u2f;
       }
 
-      return setup(options.oneFactor ? resU2F : resAllFactors)
+      return setup(options.initResponse ? options.initResponse : resAllFactors)
```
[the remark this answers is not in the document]
gowthamidommety-okta: 👍

test/unit/helpers/xhr/MFA_REQUIRED_multipleU2FmultipleFactors.js
gowthamidommety-okta: I think i will rename this file to multiple factors, so we can re-use it for the different factors that support cardinality

mauriciocastillosilva-okta: oh, so the idea is that we can add more factors here that support cardinality? I like that, so we don't have to add one file for each factor

PR #645, opened 2019-03-07
refactor: Remove jQuery from public dependency

haishengwu-okta:
## Description

- Check in jQuery source which has security vule been patched.
There are two approach to internalize jQuery, bundle within the `courage-for-sign-in-widget` or as individual file. (Essentially both check in source code)
See https://github.com/okta/okta-signin-widget/pull/598 for former approach and apparently later one way simpler. Since the end goal to upgrade jQuery, "individual file" comes as decent short term solution.
- Delete `clipboard` from dependency as not used at all.
- Move `okta-auth-js/jquery` and `backbone` devDependency (**breaking change**). 
    - The reason is because both expect `jquery` as public dependency.


package.json
```diff
@@ -107,16 +107,14 @@
...
     "webpack-bundle-analyzer": "^3.0.2",
+    "@okta/okta-auth-js": "~2.3.1",
     "webpack-dev-server": "2.11.3"
   },
   "dependencies": {
-    "@okta/okta-auth-js": "~2.3.1",
     "babel-polyfill": "^6.23.0",
     "babel-runtime": "^6.26.0",
     "backbone": "1.2.1",
-    "clipboard": "1.6.1",
```
haishengwu-okta: `clipboard` is used nowhere.

package.json
```diff
@@ -107,16 +107,14 @@
     "webdriver-manager": "12.0.6",
     "webpack": "^3.5.4",
     "webpack-bundle-analyzer": "^3.0.2",
+    "@okta/okta-auth-js": "~2.3.1",
```
haishengwu-okta: backbone and auth-js/jquery depends on `jquery` hence has to move to devDependency.

manueltanzi-okta: Probably not a big issue, but we should mention that this library will not work properly when using `npm install --production` or `yarn install --production`.

haishengwu-okta: I'm thinking mentioned in the 3.0 release notes so that developers would be aware

manueltanzi-okta: Should we also add in the README?

haishengwu-okta: probably not IMHO

manueltanzi-okta: 🚢

packages/@okta/courage-for-signin-widget/yarn.lock
```diff
@@ -127,10 +127,24 @@ array-flatten@1.1.1:
   version "1.1.1"
   resolved "https://artifacts.aue1d.saasure.com:443/artifactory/api/npm/npm-okta-master/array-flatten/-/array-flatten-1.1.1.tgz#9a5f699051b1e7073328f2a008968b64ea2955d2"
 
+array-union@^1.0.1:
```
jmelberg-okta: We had a conversation about the resolved URLs before, but need to revisit this again since its been picking up traction. Can we remove this file for now?

package.json
jmelberg-okta: This seems odd, as AuthJS only has a devDependency on jQuery. Why would that be causing an install issue for us?

PR #647, opened 2019-03-08
fix: change polling to 4 seconds

mauriciocastillosilva-okta:
It has been reported that Okta Verify Push takes too long to update it's status, so changing the poll interval to 4 sec (2 seconds less than current interval)

### DIFF
--- a/src/models/Factor.js
+++ b/src/models/Factor.js
@@ -23,7 +23,7 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
 
   // Note: Keep-alive is set to 5 seconds - using 5 seconds here will result
   // in network connection lost errors in Safari and IE.
-  var PUSH_INTERVAL = 6000;
+  var PUSH_INTERVAL = 4000;
 
   var Factor = BaseLoginModel.extend({
     extraProperties: true,
--- a/src/models/Factor.js
+++ b/src/models/Factor.js
@@ -21,8 +21,8 @@ define([
 function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
   var _ = Okta._;
 
-  // Note: Keep-alive is set to 5 seconds - using 5 seconds here will result
-  // in network connection lost errors in Safari and IE.
+  // Avoid setting interval to same value as keep-alive time (5 seconds) because it
+  // caused an occasional issue with network connection lost errors in Safari and IE
   var PUSH_INTERVAL = 4000;
 
   var Factor = BaseLoginModel.extend({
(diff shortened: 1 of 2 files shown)


src/models/Factor.js
```diff
@@ -23,7 +23,7 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
 
   // Note: Keep-alive is set to 5 seconds - using 5 seconds here will result
   // in network connection lost errors in Safari and IE.
-  var PUSH_INTERVAL = 6000;
+  var PUSH_INTERVAL = 4000;
```
haishengwu-okta: looks like reverting previous fix (5 --> 6). so 4000 work now while you testing?

mauriciocastillosilva-okta: yes. All info is in the jira, but basically with 5s we hit a problem where the server closed the connection and we are trying to send another poll. It didn't happen often, but it happened. So the older fix was to go above the keep-alive time, to 6s. That meant that the odds of the issue happening were very low. 
I did some testing with 4, 5 and 6 secs about 2 weeks ago, and I was able to see the issue with 5 sec, but not with 4 or 6. This lead me to think that if we have a different interval, either below or above, the odds of running into the issue are very low, which is what we want. And 4s will solve a problem with some customers saying that it takes too long to update the status of the factor verification

swiftone: Perhaps update the comment above to say something like "avoid being too close to the keep-alive time of 5 seconds because it caused an occasional issue with network connection lost errors in Safari and IE"?

santhoshbalakrishnan: Can you update the comment to say that having exactly 5 secs is the issue and not > or < 5?

Also, what happens when the initial 4 secs runs up? We wait exponentially? With 6 being the initial number, would we give more time to the user to verify rather than having 4?

CC: @magizh-okta

mauriciocastillosilva-okta: will update the comment 👍 

We are using the same wait time for the Initial wait (before we start polling) as for the poll itself. So we will check every 4 seconds to see if we have an answer to the push notification. The bug states _"customer reported takes anywhere from 5 to 10 seconds..."_, so with 4 seconds hopefully users will either get the answer in the first 2 polls (8 sec max, 4 sec in the base case).

santhoshbalakrishnan: Changes look good to me, but out of curiosity, does this affect how long a user can take to accept a push or just an internal timeout to see if a push was accepted within the allowed time?  Do we know the average time a user takes to accept a push?

mauriciocastillosilva-okta: it does not affect how long a user can take to accept the push. That is handled in the backend, and the timeout value is in the backend too. The Sign-In widget will keep polling until the api returns a timeout error in the response (`{ factorResult: "TIMEOUT" }` )
The only thing this affects is how often we poll to see if the user has answered the push notification.

PR #648, opened 2019-03-08
Replace Okta.Handlebars.compile by Okta.tpl.

haishengwu-okta:
- `Okta.tpl` is sort of able to 'cache' compiled template. ([more details](https://github.com/okta/okta-ui/blob/master/packages/courage/src/util/TemplateUtil.js#L15))
- also better encapsulation.

### DIFF
--- a/src/models/AppState.js
+++ b/src/models/AppState.js
@@ -26,7 +26,6 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
 
   var _ = Okta._;
   var $ = Okta.$;
-  var compile = Okta.Handlebars.compile;
 
   var DEFAULT_APP_LOGO = '/img/logos/default.png';
   var USER_NOT_SEEN_ON_DEVICE = '/img/security/unknown.png';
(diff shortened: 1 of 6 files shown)


haishengwu-okta:
Notes:
There is [one last piece to be fixed](https://github.com/okta/okta-signin-widget/blob/master/src/util/Bundles.js#L16) before we could remove `handlebars` from dependency list.
The reason is `Bundles` are depended by `Okta.js` (it's all because how i18n works in the framework).
    - https://github.com/okta/okta-signin-widget/blob/master/webpack.common.config.js#L25
    - https://github.com/okta/okta-signin-widget/blob/master/packages/%40okta/courage-for-signin-widget/webpack.config.js#L21
But `Bundles` is essentially the place providing i18n key/pair and doesn't necessarily deal with loading bundles remotely, I'm thinking we could do following refactoring
- move i18n jsonp fetching logic into another file like `I18NUtil`, which could depend on `okta`.
- change the `Bundles.getLanguage` to `Bundles.setLanguage` which only override `this.login` and `this.country` by using value fetched from jsonp.
- change `BaseLoginRouter` to `I18NUtil.loadLanguage` then `Bundles.setLanguage`

PR #651, opened 2019-03-11
Add support for multiple factors

nikhilvenkatraman-okta:
- Add supports for multiple factors on Idx pipeline. There are no visual changes to the widget screens (except for email factor which is magic link)
- Following are the factors supported for beta
1. SMS
2. Voice
3. Email
4. Password
5. Security Question
6. Google Authenticator
7. Okta Verify (with and without push)
- This PR also refactors some existing tests so that we can re-use them to run with the new Idx pipeline API.

@haishengwu-okta @mauriciocastillosilva-okta @okta/idx 

Resolves: [OKTA-213530](https://oktainc.atlassian.net/browse/OKTA-213530)

### DIFF
--- a/src/models/AppState.js
+++ b/src/models/AppState.js
@@ -258,10 +258,9 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         }
       },
       'hasMultipleFactorsAvailable': {
-        deps: ['lastAuthResponse', 'factors'],
-        fn: function (res, factors) {
-          if (res.status !== 'MFA_REQUIRED' && res.status !== 'MFA_CHALLENGE'
-            && res.status !== 'UNAUTHENTICATED') {
+        deps: ['factors', 'isMfaRequired', 'isMfaChallenge', 'isUnauthenticated'],
+        fn: function (factors, isMfaRequired, isMfaChallenge, isUnauthenticated) {
+          if (!isMfaRequired && !isMfaChallenge && !isUnauthenticated) {
             return false;
           }
           return factors && factors.length > 1;
--- a/test/unit/helpers/xhr/FACTOR_REQUIRED_oktaPassword.js
+++ b/test/unit/helpers/xhr/FACTOR_REQUIRED_oktaPassword.js
@@ -1,0 +1,52 @@
+define({
+  "status": 200,
+  "responseType": "json",
+  "response": {
+    "stateToken": "01bfpkAkRyqUZQAe3IzERUqZGOfvYhX83QYCQIDnKZ",
+    "expiresAt": "2019-06-11T21:33:24.778Z",
+    "status": "FACTOR_REQUIRED",
+    "_embedded": {
+      "user": {
+        "id": "00uhn6dAGR4nUB4iY0g3",
+        "profile": {
+          "login": "administrator1@clouditude.net",
+          "firstName": "Add-Min",
+          "lastName": "O'Cloudy Tud",
+          "locale": "en_US",
+          "timeZone": "America\/Los_Angeles"
+        }
+      },
+      "policy": {
+        "allowRememberDevice": true,
+        "rememberDeviceLifetimeInMinutes": 0,
+        "rememberDeviceByDefault": false
+      },
+      "factors":[
+        {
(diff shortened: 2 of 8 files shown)


src/models/AppState.js
```diff
@@ -261,7 +261,8 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         deps: ['lastAuthResponse', 'factors'],
         fn: function (res, factors) {
```
haishengwu-okta: looks like could depend on `isMfaRequired`, `isMfaChallenge` and `isUnauthenticated`.

nikhilvenkatraman-okta: Thanks for reviewing. I was planning on making that change. I am still making some changes/fixing some tests. Will send out for review once i am done.

src/models/AppState.js
```diff
@@ -258,10 +258,9 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         }
       },
       'hasMultipleFactorsAvailable': {
-        deps: ['lastAuthResponse', 'factors'],
-        fn: function (res, factors) {
-          if (res.status !== 'MFA_REQUIRED' && res.status !== 'MFA_CHALLENGE'
-            && res.status !== 'UNAUTHENTICATED') {
+        deps: ['factors', 'isMfaRequired', 'isMfaChallenge', 'isUnauthenticated'],
+        fn: function (factors, isMfaRequired, isMfaChallenge, isUnauthenticated) {
+          if (!isMfaRequired && !isMfaChallenge && !isUnauthenticated) {
```
nikhilvenkatraman-okta: We also want to check for `FACTOR_REQUIRED ` and `FACTOR_CHALLENGE`.  `isMfaRequired ` is true when status `MFA_REQUIRED` or `FACTOR_REQUIRED`. Same is the case for  `isMfaChallenge `

test/unit/spec/MfaVerify_spec.js
```diff
@@ -564,1304 +581,1663 @@ function (Okta,
...
+      itp('sets the transaction error on the appState on error response', function () {
+        return setupFn().then(function (test) {
+          mockTransactions(test.router.controller);
+          test.setNextResponse(resInvalidTotp);
+          test.form.setAnswer('wrong');
+          test.form.submit();
+          return Expect.waitForFormError(test.form, test);
+        })
+          .then(function (test) {
+            expectSetTransactionError(test.rou
```
mauriciocastillosilva-okta: not sure about the name. The only reference to this right now is in [AppState](https://github.com/okta/okta-signin-widget/blob/master/src/models/AppState.js#L288) for `isIdxStateToken`, so maybe we should name these tests `SMS with idxStateToken` (applies for all)

test/unit/helpers/xhr/FACTOR_REQUIRED_oktaPassword.js
```diff
@@ -0,0 +1,52 @@
+define({
+  "status": 200,
+  "responseType": "json",
+  "response": {
+    "stateToken": "01bfpkAkRyqUZQAe3IzERUqZGOfvYhX83QYCQIDnKZ",
+    "expiresAt": "2019-06-11T21:33:24.778Z",
+    "status": "FACTOR_REQUIRED",
+    "_embedded": {
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: yeah, they are not being used in code right now.

test/unit/spec/MfaVerify_spec.js
nikhilvenkatraman-okta: hmm Idx pipeline is a more generic name for the new pipeline. I don't want to use `isIdxStateToken` in the test since it is a very specific derived attribute name that tells us that the stateToken has a `01` prefix. Idx pipeline is also short hand for the actual FF which is `IDENTITY_EXPERIENCE_PIPELINE`

mauriciocastillosilva-okta: Got it. Then let's just name all the tests the same `{factor} on IdX Pipeline` (or whichever combination of upper/lower case if prefered

PR #653, opened 2019-03-14
Hide `back to sign-in` depends on FF

iuriipetrov-okta:
Resolves: OKTA-212309

### DIFF
--- a/src/ForgotPasswordController.js
+++ b/src/ForgotPasswordController.js
@@ -30,9 +30,11 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, Util, ContactSu
 
   var Footer = Okta.View.extend({
     template: '\
-      <a href="#" class="link help js-back" data-se="back-link">\
-        {{i18n code="goback" bundle="login"}}\
-      </a>\
+      {{#unless hideBackToSignInForReset}}\
+        <a href="#" class="link help js-back" data-se="back-link">\
+          {{i18n code="goback" bundle="login"}}\
+        </a>\
+      {{/unless}}\
       {{#if helpSupportNumber}}\
       <a href="#" class="link goto js-contact-support">\
         {{i18n code="mfa.noAccessToEmail" bundle="login"}}\
@@ -52,7 +54,10 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, Util, ContactSu
       }
     },
     getTemplateData: function () {
-      return this.settings.pick('helpSupportNumber');
+      return {
+        hideBackToSignInForReset: this.settings.get('features.hideBackToSignInForReset'),
+        helpSupportNumber: this.settings.get('helpSupportNumber')
+      };
     },
     back: function () {
       this.state.set('navigateDir', Enums.DIRECTION_BACK);
--- a/src/PasswordResetController.js
+++ b/src/PasswordResetController.js
@@ -80,7 +80,6 @@ function (Okta, FormController, FormType, ValidationUtil, FactorUtil, FooterSign
         ];
       }
     },
-    Footer: FooterSignout,
 
     initialize: function () {
       this.listenTo(this.form, 'save', function () {
@@ -91,6 +90,10 @@ function (Okta, FormController, FormType, ValidationUtil, FactorUtil, FooterSign
         this.settings.processCreds(creds)
           .then(_.bind(this.model.save, this.model));
       });
+
+      if (!this.settings.get('features.hideBackToSignInForReset')) {
+        this.addFooter(FooterSignout);
+      }
     }
 
   });
--- a/src/PwdResetEmailSentController.js
+++ b/src/PwdResetEmailSentController.js
@@ -39,7 +39,7 @@ function (Okta, Enums, FormController, FormType) {
       noButtonBar: true,
       attributes: { 'data-se': 'pwd-reset-email-sent' },
       formChildren: function () {
-        return [
+        let children = [
           FormType.View({
             View: Okta.View.extend({
               template: '\
@@ -49,9 +49,11 @@ function (Okta, Enums, FormController, FormType) {
                 return { 'alert': Okta.loc('password.forgot.emailSent.title', 'login') };
               }
             })
-          }),
+          })
+        ];
 
-          FormType.Button({
+        if (!this.settings.get('features.hideBackToSignInForReset')) {
+          children.push(FormType.Button({
             title: Okta.loc('goback', 'login'),
             className: 'button button-primary button-wide',
             attributes: {'data-se': 'back-button'},
@@ -65,8 +67,10 @@ function (Okta, Enums, FormController, FormType) {
                   self.options.appState.trigger('navigate', '');
                 });
             }
-          })
-        ];
+          }));
+        }
+
+        return children;
       }
     },
 
@@ -74,8 +78,7 @@ function (Okta, Enums, FormController, FormType) {
       this.settings.callGlobalSuccess(Enums.FORGOT_PASSWORD_EMAIL_SENT, {
         username: options.appState.get('username')
       });
-    }
-
+    },
   });
 
 });
--- a/src/RecoveryQuestionController.js
+++ b/src/RecoveryQuestionController.js
@@ -80,7 +80,11 @@ function (Okta, FormController, FormType, FooterSignout, TextBox) {
         ];
       }
     },
-    Footer: FooterSignout
+    initialize: function () {
+      if (!this.settings.get('features.hideBackToSignInForReset')) {
+        this.addFooter(FooterSignout, {linkText: Okta.loc('goback', 'login'), linkClassName: ''});
+      }
+    }
 
   });
 
--- a/src/models/Settings.js
+++ b/src/models/Settings.js
@@ -79,6 +79,7 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
       'features.preventBrowserFromSavingOktaPassword': ['boolean', true, true],
       'features.deviceFingerprinting': ['boolean', false, false],
       'features.hideSignOutLinkInMFA' : ['boolean', false, false],
+      'features.hideBackToSignInForReset' : ['boolean', false, false],
       'features.customExpiredPassword': ['boolean', true, false],
       'features.registration': ['boolean', false, false],
       'features.consent': ['boolean', false, false],
--- a/src/RecoveryChallengeController.js
+++ b/src/RecoveryChallengeController.js
@@ -154,7 +154,11 @@ function (Okta, FormController, FormType, Enums, FooterSignout, TextBox) {
         this.add(sendEmailLink);
       }
 
-      this.add(new FooterSignout(_.extend(this.toJSON(), {linkText: Okta.loc('goback', 'login'), linkClassName: ''})));
+      if (!this.settings.get('features.hideBackToSignInForReset')) {
+        this.add(
+          new FooterSignout(_.extend(this.toJSON(), {linkText: Okta.loc('goback', 'login'), linkClassName: ''}))
+        );
+      }
     },
 
     postRender: function () {
--- a/src/util/FormController.js
+++ b/src/util/FormController.js
@@ -83,12 +83,16 @@ function (Okta, FormType, BaseLoginController, BaseLoginModel) {
       }
 
       if (this.Footer) {
-        this.footer = new this.Footer(this.toJSON());
-        this.add(this.footer);
+        this.addFooter(this.Footer);
       }
 
       this.addListeners();
       initialize.apply(this, arguments);
+    },
+
+    addFooter: function (Footer, args) {
+      this.footer = new Footer(_.extend(this.toJSON(), args || {}));
+      this.add(this.footer);
     },
 
     toJSON: function () {
--- a/test/unit/spec/ForgotPassword_spec.js
+++ b/test/unit/spec/ForgotPassword_spec.js
@@ -73,6 +73,7 @@ function (Q, Okta, OktaAuth, Util, AccountRecoveryForm, PrimaryAuthForm, Beacon,
   var setupWithoutEmail = _.partial(setup, { 'features.emailRecovery': false });
   var setupWithSmsWithoutEmail = _.partial(setup, { 'features.smsRecovery': true, 'features.emailRecovery': false });
   var setupWithCallWithoutEmail = _.partial(setup, { 'features.callRecovery': true, 'features.emailRecovery': false });
+  var setupWithHideBackLink = _.partial(setup, { 'features.hideBackToSignInForReset': true });
 
   Expect.describe('ForgotPassword', function () {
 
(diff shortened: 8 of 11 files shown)


src/ForgotPasswordController.js
```diff
@@ -52,7 +54,7 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, Util, ContactSu
       }
     },
     getTemplateData: function () {
-      return this.settings.pick('helpSupportNumber');
+      return this.settings.pick('helpSupportNumber', 'features');
```
haishengwu-okta: probably only `pick` `features.hideBackToSignInForReset`

nikhilvenkatraman-okta:
yeah, we have `back to sign-in` link in other places like Haisheng mentioned, are we not going to hide those ? Also i feel like we should not put this under a feature in the widget, since this is just a visual thing. We have put visual changes under the `features` config object in the past but we should try and avoid that if possible. Thoughts @haishengwu-okta @mauriciocastillosilva-okta ?

haishengwu-okta:
couple of ideas
- `config.hideBack`
- `settings.hideBack`

gowthamidommety-okta:
Hi @nikhilvenkatraman-okta and @haishengwu-okta,

1. These changes are to prevent users from going back to the primary auth page during the password reset flow for rdp agent. The design wiki is linked in the JIRA
2. Since we are implementing this specifically for the Password reset flow, we are not making changes for unlock account or registration. Should this be more generic and cover other flows? It may not make sense for Registration right?
3. We decided to put this under features, because it goes with the design for the existing `features.hideSignOutLinkInMFA`. I guess we could put this under config or settings if that is how we are now designing it, but it would be inconsistent?

nikhilvenkatraman-okta:
Yeah, I feel we have polluted the features object with a lot of config already and we could potentially avoid adding configs that are purely visual. Another candidate in this category is `features.showPasswordToggleOnSignInPage`. As an aside this could potentially be something to discuss this during a SIW design review meeting. Also while we are in this area we could think about re-using [FooterWithBackLink](https://github.com/okta/okta-signin-widget/blob/master/src/views/shared/FooterWithBackLink.js) if possible ?

mauriciocastillosilva-okta:
When I added `hideSignOutLinkInMFA` it went inside `features` because that made sense at the time. I think I had the config option at top level, but after some comments we decided to put it under `features`. I agree it would fit better in a different place of our config options, but Gowthami has a good point about consistency... so either we start a new config object or we stay consistent with what we have:
- new config object: probably something like  `config.links.*`? I think we need to design this in a way that we don't have a problem in the future if we add some other configs to hide other links... (and then move `features.hideSignOutLinkInMFA` eventually, with backwards compatibility).
- keep it under features, which is consistent with what we have

I like the idea of a new config object, but we haven't design what this will look like...

nikhilvenkatraman-okta:
Yeah, +1 for adding a new config object and we can do it in a different PR. But can we explore using 
 [FooterWithBackLink](https://github.com/okta/okta-signin-widget/blob/master/src/views/shared/FooterWithBackLink.js) ?

gowthamidommety-okta:
I agree that we should have a new config object and that it would be best to discuss the design first and do this as part of another PR. 
@nikhilvenkatraman-okta By reuse FooterWithBackLink, do you mean use it in place of `FooterSignout` in this PR?

nikhilvenkatraman-okta:
@gowthamidommety-okta I think we are creating a Footer inline in the controller [here](https://github.com/okta/okta-signin-widget/blob/45a239cd04b8eba4328d1322fd5c5ade71d05f93/src/ForgotPasswordController.js#L31), but we could potentially include  FooterWithBackLink and use it similar to [EnrollUserController](https://github.com/okta/okta-signin-widget/blob/f2f68141ab35ca702b856ae153daf33471678895/src/EnrollUserController.js#L17).

src/ForgotPasswordController.js
iuriipetrov-okta: Doesn't work for deep picking object.

iuriipetrov-okta:
@nikhilvenkatraman-okta @mauriciocastillosilva-okta @gowthamidommety-okta 

Could you tell me is my idea following our concept or not? And can I do it?
I suggest instead of creating a `FooterWithBackLink ` instance, add it to the View and wrap it in the FF condition inside `initialize` or `formChildren` methods, extend `FooterWithBackLink `, check the FF condition inside this component and hide it if the FF is disabled and then use this Class in `Footer` property of `FormController`.

On one hand it's simpler and cleaner way to switch the component depends of the FF, on the other hand it's a bit waste of memory because we create a View instance regardless of the FF.

nikhilvenkatraman-okta:
I am ok with that idea if it works, as we get to re-use most of the logic from `FooterWithBackLink `

gowthamidommety-okta:
I am a little hesitant to create a new view just for the password reset flow, but if it is reusing FooterWithBackLink then that is good. 
Another concern I have is not all links in this flow say `Back to sign in` some say `Sign out`, but I think that is a question we can run by UX.

src/views/shared/FooterHideableBackLink.js
```diff
@@ -0,0 +1,35 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+define(['./FooterSignout', 'okta', 'util/Enums'], function (FooterSignOut, Okta, Enums) {
+  let { Util } = Okta.internal.util;
+  let _ = Okta._;
+
+  return FooterSignOut.extend({
```
haishengwu-okta: Naming is a bit misleading, it extends `FooterSignOut` but file name is `FooterHideableBackLink.js`. 
- So what's exact purpose ?
- also looks like you'd rather update the template in parent like what you have done for `ForgotPasswordController.js`?

iuriipetrov-okta: It's been updated.

src/views/shared/FooterHideableBackLink.js
```diff
@@ -0,0 +1,52 @@
+/*!
```
gowthamidommety-okta: Maybe we should rename this as `FooterForReset`, because this is for the password reset flow? 
I'm also not very sure we should even be creating a new footer. It is not extending an existing footer, so it seems like a lot of repeated code with just an extra if check. What do you think? @haishengwu-okta @nikhilvenkatraman-okta @mauriciocastillosilva-okta if you could weigh in as well

src/ForgotPasswordController.js
```diff
@@ -52,7 +54,10 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, Util, ContactSu
       }
     },
     getTemplateData: function () {
-      return this.settings.pick('helpSupportNumber');
+      return _.extend(
+        this.settings.pick('helpSupportNumber'),
+        { hideBackToSignInForReset: !!this.settings.get('features.hideBackToSignInForReset') }
```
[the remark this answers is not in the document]
iuriipetrov-okta: Just to be sure that's a boolean, not `undefined`. Which could be if we don't have such FF in the `features` object.

src/PwdResetEmailSentController.js
```diff
@@ -50,8 +51,7 @@ function (Okta, Enums, FormController, FormType) {
               }
             })
           }),
-
-          FormType.Button({
+          (!this.settings.get('features.hideBackToSignInForReset') ? FormType.Button({
```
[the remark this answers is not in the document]
iuriipetrov-okta: Yeah, I agree.

src/ForgotPasswordController.js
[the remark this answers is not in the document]
iuriipetrov-okta: Oh, didn't see your first comment. So if we have that definition in Settings then it'll be boolean with chosen default value anyway, I see.

src/views/shared/FooterHideableBackLink.js
iuriipetrov-okta: Well, only one thing is really what we need to change from `FooterSignout` view is hiding the component if `hideBackToSignInForReset` is true. So if we can figure out how to make it easier than passing our FF to the template, then it'll be less repeatable. 

And also I believe we should move logic of click handling from the new component to `FooterSignout`. I mean this one:
image
Because, as I understand, we don't need to cancel a transaction if we don't have `settings.get('signOutLink')` property and just want to navigate back.

mauriciocastillosilva-okta: I don't think we need a new footer at all, and also this is basically the same code as we have in `FooterSignout`.
The idea of `hideBackToSignInForReset` is to hide the sing out link, and we can do that by just **not adding** the footer in the controller(s). So the controller will have the logic to know if the footer is needed or not.

src/models/Settings.js
```diff
@@ -79,6 +79,7 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
       'features.preventBrowserFromSavingOktaPassword': ['boolean', true, true],
       'features.deviceFingerprinting': ['boolean', false, false],
       'features.hideSignOutLinkInMFA' : ['boolean', false, false],
+      'features.hideBackToSignInForReset' : ['boolean', false, false],
```
nikhilvenkatraman-okta: should this be renamed to `hideBackToSignInForResetAndRecovery` ? since I see this is also used in some recovery flows ?

iuriipetrov-okta:
@vincentvoong-okta 
Could you confirm 'back link' at these screenshots?
Screen Shot 2019-03-25 at 5 08 04 PM
Screen Shot 2019-03-25 at 5 08 37 PM

src/PwdResetEmailSentController.js
```diff
@@ -14,7 +14,7 @@ define([
   'okta',
   'util/Enums',
   'util/FormController',
-  'util/FormType'
+  'util/FormType',
```
gowthamidommety-okta: Was the comma `,` added for a lint error?

src/PwdResetEmailSentController.js
```diff
@@ -39,7 +39,7 @@ function (Okta, Enums, FormController, FormType) {
       noButtonBar: true,
       attributes: { 'data-se': 'pwd-reset-email-sent' },
       formChildren: function () {
-        return [
+        let views = [
```
gowthamidommety-okta: nit: rename `views` to `children` to be consistent with other similar controllers

src/PwdResetEmailSentController.js
iuriipetrov-okta: I think it doesn't trigger any lint error.

iuriipetrov-okta:
@gowthamidommety-okta 
We have test for `RecoveryQuestionController.js`, but not for `PwdResetEmailSentController.js`; we don't have spec file for `PwdResetEmailSentController.js`.

gowthamidommety-okta:
@iuriipetrov-okta if it is out of scope for this change, please file a JIRA to add tests for PwdResetEmailSentController.js

vincentvoong-okta:
@gowthamidommety-okta @iuriipetrov-okta 
Changes look good. Lets change to "Back to Sign In"

src/ForgotPasswordController.js
```diff
@@ -52,7 +54,10 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, Util, ContactSu
       }
     },
     getTemplateData: function () {
-      return this.settings.pick('helpSupportNumber');
+      return _.extend(
```
mauriciocastillosilva-okta: no need for `_.extend` here, we could just return the object with the data we need
```
return {
  helpSupportNumber: this.settings.get('helpSupportNumber'),
  hideBackToSignInForReset: this.settings.get('features.hideBackToSignInForReset')
}
```

src/PasswordResetController.js
```diff
@@ -80,7 +80,6 @@ function (Okta, FormController, FormType, ValidationUtil, FactorUtil, FooterSign
         ];
       }
     },
-    Footer: FooterSignout,
```
mauriciocastillosilva-okta: what you are doing is correct, but it is missing something: there is some functionality that FormController.js was providing with the footer. Take a look at [back](https://github.com/okta/okta-signin-widget/blob/master/src/util/FormController.js#L100) in that file. (that method is used in [BaseLoginRouter](https://github.com/okta/okta-signin-widget/blob/master/src/util/BaseLoginRouter.js#L294))
Maybe what we can do is either

a) add a `addFooter` method to FormController and call that wherever we need
```
function addFooter(Footer, options) {
  this.footer = new Footer ...
  this.add(this.footer)
```
or 
b) just set `this.footer` before adding it to the controller.

I prefer option a, but look into what makes more sense

iuriipetrov-okta: I prefer option a as well. So does it look well now?

src/PasswordResetController.js
```diff
@@ -91,6 +90,10 @@ function (Okta, FormController, FormType, ValidationUtil, FactorUtil, FooterSign
         this.settings.processCreds(creds)
           .then(_.bind(this.model.save, this.model));
       });
+
+      if (!this.settings.get('features.hideBackToSignInForReset')) {
+        this.addFooter(FooterSignout, {linkText: Okta.loc('goback', 'login'), linkClassName: ''});
```
mauriciocastillosilva-okta: question: did we agree on changing the button text here to `Back`? This controller is for when the user is changing the password, so the user is already in the last step... just want to confirm, this could make sense to leave it as it was with "Sign Out"

gowthamidommety-okta: Yes, we thought about it and it's in the design doc. @vincentvoong-okta also agreed it would make sense to change to "Back to Sign In", which would be consistent with the rest of the flow.
However, what you're saying is true. I think at this point, the user is authenticated. So we could say "Sign Out" instead of "Back to Sign In".  I don't really have a strong opinion, because the meaning of both these labels is very similar, and they both go to the primary auth page.
Maybe we can leave this to say "Sign Out" for now.

src/RecoveryQuestionController.js
```diff
@@ -82,7 +82,7 @@ function (Okta, FormController, FormType, FooterSignout, TextBox) {
     },
     initialize: function () {
       if (!this.settings.get('features.hideBackToSignInForReset')) {
-        this.addFooter(FooterSignout, {linkText: Okta.loc('goback', 'login'), linkClassName: ''});
```
mauriciocastillosilva-okta: In this controller, `RecoveryQuestionController.js`, we are still going through the process of "authenticating". So I don't think this one needs to be reverted to "Sign out". This should be whatever was agreed with UX

iuriipetrov-okta: Well, it was `Sign out ` before this change anyway. And we agreed with UX that in those two cases the text can be either `Sign out` or `Back to Sign In`.

mauriciocastillosilva-okta: got it. Will defer to @gowthamidommety-okta on this one, the rest of the PR looks good

gowthamidommety-okta: 👍 changing it back to Sign out for now!

gowthamidommety-okta: Sorry, i did not realize this was src/RecoveryQuestionController.js. Agree with Mau, this should be Back To Sign In. src/PasswordResetController.js should stay saying Sign Out

PR #655, opened 2019-03-15
feat: Webauthn support for roaming and platform authenticators

gowthamidommety-okta:
Platform:

PlatformAuthenticatorEdge

Roaming:

RoamingAuthenticatorEdge

On browser that does not support webauthn (IE) : 

Screen Shot 2019-03-19 at 1 40 52 PM


src/EnrollWebauthnController.js
```diff
@@ -113,8 +114,14 @@ function (Okta, Errors, FormType, FormController, CryptoUtil, FidoUtil, webauthn
     },
 
     Form: {
-      title: _.partial(Okta.loc, 'enroll.u2f.title', 'login'),
-      save: _.partial(Okta.loc, 'enroll.u2f.save', 'login'),
+      title: function () {
+        return webauthn.isNewApiAvailable() ?
```
mauriciocastillosilva-okta: do we need this check? This file is specifically for `webauthn`. Before this change `webauthn` factor was showing the labels for `u2f`, but after this change isn't the intention that `webauthn` will show only `webauthn` labels? Do we still need a backup of `webauthn` working exactly as u2f?

src/VerifyWebauthnController.js
```diff
@@ -123,7 +123,9 @@ function (Okta, Errors, FormController, FormType, CryptoUtil, FidoUtil,
     Form: {
       autoSave: true,
       hasSavingState: false,
-      title: _.partial(Okta.loc, 'factor.u2f', 'login'),
+      title: function () {
+        return webauthn.isNewApiAvailable() ? Okta.loc('factor.webauthn', 'login') : Okta.loc('factor.u2f', 'login');
```
mauriciocastillosilva-okta: same question that I made in Enroll controller, do we still need `u2f` backup?

src/util/FactorUtil.js
```diff
@@ -161,7 +167,11 @@ function (Okta, TimeUtil) {
       return 'QUESTION';
     }
     if (provider === 'FIDO' && factorType === 'webauthn') {
-      if (this.settings.get('features.webauthn')) {
+      var hasWebAuthnSupport = navigator && navigator.credentials && navigator.credentials.create;
```
mauriciocastillosilva-okta: can we use `webauth.isNewApiAvailable`?

src/EnrollWebauthnController.js
gowthamidommety-okta: I had this question too and I spoke with @mpeng-okta about this. The backend still defaults to u2f as well, so I left the front-end to do the same

src/util/FactorUtil.js
gowthamidommety-okta: Yea we could.. I would have to import util/webauthn.js so I thought I would do this instead. I'll make changes

packages/@okta/i18n/package.json
```diff
@@ -79,5 +78,5 @@
     "test:unit": "mocha --ui bdd --require tasks/i18ndiff/test/support/node -R spec tasks/i18ndiff/test",
     "test:unit-report": "mocha --ui bdd --require tasks/i18ndiff/test/support/node -R mocha-junit-reporter -O mochaFile=./build2/reports/unit/mocha-junit-result.xml tasks/i18ndiff/test"
   },
-  "version": "51.0.0-beta.g9c45330"
+  "version": "51.1.0-alpha.g417fc8b"
```
mauriciocastillosilva-okta: reminder to update to beta version

src/EnrollWebauthnController.js
```diff
@@ -129,30 +100,33 @@ function (Okta, Errors, FormType, FormController, CryptoUtil, FidoUtil, webauthn
...
+                {{#if isEdge}}\
+                  <div class="webauthn-edge-text">\
+                     <p>{{{i18n code="enroll.webauthn.instructions.edge" bundle="login"}}}</p>\
+                  </div>\
+                {{/if}}\
+                <div data-se="webauthn-waiting" class="okta-waiting-spinner hide"></div>\
+              ',
+              getTemplateData: function () {
+                return {
+                  isEdge: window.navigator.userAgent.indexOf('Edge') > -1
```
mauriciocastillosilva-okta: let's add this as a function to [BrowserFeatures.js](https://github.com/okta/okta-signin-widget/blob/master/src/util/BrowserFeatures.js), similar to what we have there. (also this could simplify how to mock the tests that need the agent to be Edge)

test/unit/spec/EnrollWebauthn_spec.js
```diff
@@ -136,50 +123,68 @@ function (Okta,
...
           Expect.isVisible(test.form.enrollInstructions());
           Expect.isVisible(test.form.submitButton());
         });
       });
 
-      itp('shows a waiting spinner and devices images after submitting the form', function () {
+      itp('shows correct instructions for Edge browser', function () {
+        const userAgent = window.navigator.userAgent;
+        spyOn(webauthn, 'isNewApiAvailable').and.returnValue(true);
+        window.navigator['__defineGetter__']('userAgent', () => {
```
mauriciocastillosilva-okta: once we create a function in BrowserFeatures, this could be as simple as `spyOn(BrowserFeatures, 'isEdge').and.returnValue(true);`

shekarkoppula-okta:
DoD approved.

PR #663, opened 2019-03-28
OKTA-207909 : Fix to disable button and load spinner on click for password reset/expire

rameshbhojan-okta:


src/PasswordExpiredController.js
```diff
@@ -37,10 +37,15 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, FactorUtil, Foo
       },
       save: function () {
         return this.doTransaction(function (transaction) {
-          return transaction.changePassword({
+          this.trigger('save');
```
magizh-okta: I think this wont work because we need to trigger false only after the transaction finishes. 
Javascript is async. 

so try this. 
`
save: function() {
  this.appState.trigger('loading', true);
  return this.doTransaction(function (transaction) {
    return  transaction.changePassword({
      oldPassword: this.get('oldPassword'),
      newPassword: this.get('newPassword')
    }).fin(() => {
      this.appState.trigger('loading', false);
    });
  });
});

`


Not sure if we need the fail case like here. 
https://github.com/okta/okta-signin-widget/blob/master/src/models/PrimaryAuth.js#L130

@mauriciocastillosilva-okta  thoughts ?  Can you also confirm if my suggestion fits for this situation ?

rameshbhojan-okta: @magizh-okta - I tested my change locally and it worked fine. 
`this.trigger('save');` -> Disables the button on click
`this.appState.trigger('loading', true);` -> showed the spinner
`this.appState.trigger('removeLoading');` -> removes the spinner (without this, the spinner keeps showing even if we get a response. e.g. a validation error.

magizh-okta:
```
save: function() {
  this.appState.trigger('loading', true);
  return this.doTransaction(function (transaction) {
    return  transaction.changePassword({
      oldPassword: this.get('oldPassword'),
      newPassword: this.get('newPassword')
    }).fail(() => {
      this.appState.trigger('removeLoading');
    }).fin(() => {
      this.appState.trigger('loading', false);
    });
  });
});
```

rameshbhojan-okta: I also tried with this code, and this suggested snippet above was posing some issues and wasn't working as expected. So, I left my initial fix as as. I also have attached the recording of the fix in this [box link](https://okta.box.com/s/0tfx7rxkt6c0eh84vp5r1qlnkprjki09).

haishengwu-okta:
```

Comment (rameshbhojan-okta) [reply src/PasswordExpiredController.js]: @haishengwu-okta - I showed and discussed the issues briefly with @magizh-okta . With the code above, in case of validation errors the proper error is not displayed and we get a more generic error stating something went wrong. Also, after that if we make changes and resubmit, the flow hangs forever with the spinner spinning and button getting displayed but never hitting `fail` or `fin` and just running indefinitely. 
We briefly spent some time trying to identify the cause but couldn't get much further on identifying what went wrong. 
On the other hand, the existing fix (in this PR changes) has been working as expected without these issues as above, so I have kept the changes as is for now. Please see the [demo here](https://okta.box.com/s/0tfx7rxkt6c0eh84vp5r1qlnkprjki09). So, I was wondering if we can go ahead with this current fix, unless you have other concerns?

Comment (haishengwu-okta) [reply src/PasswordExpiredController.js]: iirc, `transaction.changePassword` is an ajax call hence it might take longer than in your local.
theoretically, the ajax call could still in flight when `loading` has been set to false. I think that's reason behind @magizh-okta 's suggestion at first place.

Is the error `something went wrong` coming from unit test troubleshooting?

Comment (rameshbhojan-okta) [reply src/PasswordExpiredController.js]: Ok, I understand. The error was not from unit test troubleshooting but instead by direct/manual testing with the changes.

Comment (mauriciocastillosilva-okta) [reply src/PasswordExpiredController.js]: ok, I was trying this out and I know what the problem is. Look at the code below
```
save: function () {
  // We trigger save so the button get disabled.
  this.trigger('save');
  this.appState.trigger('loading', true);
  return this.doTransaction(function (transaction) {
    return transaction.changePassword({
      oldPassword: this.get('oldPassword'),
      newPassword: this.get('newPassword')
    }).fail((err) => {
      this.appState.trigger('removeLoading');
      return Q.reject(err); //This is what we were missing
    }).fin(() => {
      this.appState.trigger('loading', false);
    });
  });
}
```
The problem is that by adding a `fail` to the `changePassword` promise and not returning that failure, then the [parent promise](https://github.com/okta/okta-signin-widget/blob/master/src/models/BaseLoginModel.js#L32) is not aware that there was an error, so it goes to the `then` [function](https://github.com/okta/okta-signin-widget/blob/master/src/models/BaseLoginModel.js#L33) which sets the transaction in the appState to undefined (since the promise didn't return a result). Once the transaction is undefined, the second time we call this changePassword function, it blows up because there is no transaction
By adding the line to reject the promise we bubble up the error and then everything goes back to work as usual

haishengwu-okta: @mauriciocastillosilva-okta 🥇

mauriciocastillosilva-okta: the other way to do this is instead of `return Q.reject(err);` to just throw the error `throw err;`, it has the same effect and we don't need to import Q. Either works, but might be simpler to throw the error

rameshbhojan-okta: Thanks a lot @mauriciocastillosilva-okta for identifying the issue. Updated the PR now with the recommended code! Please take a look.

PR #664, opened 2019-03-28
FIX for 213636

magizh-okta:
* If SIW is embedded in an iframe the custom factor providers (OIDC/SAML) may not load in the iframe.
* Fix here is to load the url in the top frame in the window.

OKTA-213636

Before Fix - CUSTOM factor doesn't load up in iframe

Screen Shot 2019-04-01 at 1 50 02 PM

After fix - custom OIDC factor 

siwOIDCdemos

after fix - question factor (no impact) 

siwOIDCquestion-factor

### DIFF
--- a/src/VerifyCustomFactorController.js
+++ b/src/VerifyCustomFactorController.js
@@ -14,12 +14,12 @@ define([
   'okta',
   'util/FormController',
   'views/shared/FooterSignout',
-  'util/FactorUtil'
+  'util/FactorUtil',
+  'util/Util'
 ],
-function (Okta, FormController, FooterSignout, FactorUtil) {
+function (Okta, FormController, FooterSignout, FactorUtil, Util) {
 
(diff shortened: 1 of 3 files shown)


src/util/RouterUtil.js
```diff
@@ -147,8 +147,12 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
         var targetUrl = res._links && res._links.next && res._links.next.href;
         successData.stepUp = {
           url: targetUrl,
-          finish: function () {
-            redirectFn(targetUrl);
+          finish: function (redirectTopFrame) {
+            if (redirectTopFrame) {
+              Util.redirectTopFrame(targetUrl);
```
magizh-okta: Need to know if this is okay for okta use case.

magizh-okta:
not needed anymore

PR #668, opened 2019-04-02
:fix Do not compute device fingerprint when security image is off

magizh-okta:
* When user tabs out after typing username, we should not compute device fingerprint
if sec image feature is off.

OKTA-218737

**BEfore** 
Screen Shot 2019-04-02 at 7 47 06 AM

**After** 
Screen Shot 2019-04-02 at 7 49 58 AM

secdev

### DIFF
--- a/test/unit/spec/IDPDiscovery_spec.js
+++ b/test/unit/spec/IDPDiscovery_spec.js
@@ -594,6 +594,21 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, 
     });
 
     Expect.describe('Device Fingerprint', function () {
+      itp(`is not computed if securityImage is off, deviceFingerprinting is
+        true and useDeviceFingerprintForSecurityImage is true`, function () {
+        spyOn(DeviceFingerprint, 'generateDeviceFingerprint');
+        return setup({ features: { securityImage: false, deviceFingerprinting: true,
+          useDeviceFingerprintForSecurityImage: true}})
+          .then(function (test) {
+            test.setNextResponse(resSecurityImage);
+            test.form.setUsername('testuser@clouditude.net');
+            return waitForBeaconChange(test);
+          })
+          .then(function () {
+            expect($.ajax.calls.count()).toBe(0);
+            expect(DeviceFingerprint.generateDeviceFingerprint).not.toHaveBeenCalled();
+          });
+      });
       itp(`contains fingerprint header in get security image request if deviceFingerprinting
         is true (useDeviceFingerprintForSecurityImage defaults to true)`, function () {
         spyOn(DeviceFingerprint, 'generateDeviceFingerprint').and.callFake(function () {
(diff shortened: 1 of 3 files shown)


test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -594,6 +594,25 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm,
     });
 
     Expect.describe('Device Fingerprint', function () {
+      itp(`does not compute device fingerprint if security image is off, deviceFingerprinting
+        is true and useDeviceFingerprintForSecurityImage is true)`, function () {
+        spyOn(DeviceFingerprint, 'generateDeviceFingerprint').and.callFake(function () {
```
gowthamidommety-okta: Should this just be spyOn? We don't need to define the fake function because we are not calling it right?

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -985,15 +985,33 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
...
       itp(`contains fingerprint header in get security image request if deviceFingerprinting
         is true (useDeviceFingerprintForSecurityImage defaults to true)`, function () {
         spyOn(DeviceFingerprint, 'generateDeviceFingerprint').and.callFake(function () {
           var deferred = Q.defer();
           deferred.resolve('thisIsTheDeviceFingerprint');
           return deferred.promise;
         });
-        return setup({ features: { securityImage: true, deviceFingerprinting: true,
-          useDeviceFingerprintForSecurityImage: true }})
+        return setup({ features: { securityImage: true, deviceFingerprinting: true }})
```
gowthamidommety-okta: Is useDeviceFingerprintForSecurityImage true here?

magizh-okta: Yeah it is true by default.(which is what we are testing here) 

```
contains fingerprint header in get security image request if deviceFingerprinting
        is true (useDeviceFingerprintForSecurityImage defaults to true)
```

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -594,6 +594,21 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm,
     });
 
     Expect.describe('Device Fingerprint', function () {
+      itp(`does not compute device fingerprint if security image is off, deviceFingerprinting
+        is true and useDeviceFingerprintForSecurityImage is true)`, function () {
```
mauriciocastillosilva-okta: seems that there is an extra parenthesis in the description... maybe we can shorten this to `is not computed if securityImage is off, deviceFingerprinting is true and useDeviceFingerprintForSecurityImage is true`?

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -985,15 +985,29 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
     });
 
     Expect.describe('Device Fingerprint', function () {
+      itp(`does not compute device fingerprint if security image is off, deviceFingerprinting
+        is true and useDeviceFingerprintForSecurityImage is true)`, function () {
```
mauriciocastillosilva-okta: nit: same here about the extra parenthesis that is not needed

PR #669, opened 2019-04-04
Runs e2e tests on iOS emulators (iOS 10, 11, 12)

vijetmahabaleshwar-okta:
* Adds support to run e2e tests on iOS 10, 11, 12 emulators
* Runs the tests daily as a scheduled cron job

**Open question** - How to handle cron job failures? 
Alerts can be sent to slack channel, but would be sent for all events (pr, push & cron)


scripts/travis.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# if this build was triggered as cron job, run e2e tests on iOS
```
haishengwu-okta: when the build will be triggered as cron job?

vijetmahabaleshwar-okta: Right now, I've scheduled it to run this branch as cron daily here - https://travis-ci.org/okta/okta-signin-widget/settings (Scroll down to Cron Jobs section)
If we merge this PR, we can run the master branch instead.

jmelberg-okta: Can you explain why we decided to include this coverage **only** via a cron event type?

Moving forward, I'd assume we'd like to ensure that the Widget works properly on supported Android and iOS versions. If we have failing tests for Android or iOS, I would expect to have Travis block the PR from being merged. Is this something we're gradually moving toward? Are tests on these devices simply not working today?

test/e2e/conf.js
```diff
@@ -24,14 +24,39 @@ if (process.env.TRAVIS) {
 
   // Mobile devices
   if (process.env.SAUCE_PLATFORM_NAME === 'iOS') {
```
jmelberg-okta: Since we're likely to test this on Android in the future, can we pull out the iOS configuration into a separate file? This will help make configuration verbose across platforms.

Also - lets add a comment in our iOS configuration section that links to the available [capabilities](http://appium.io/docs/en/writing-running-appium/caps/#general-capabilities), so contributors can read about the available options.

As an example, `fullReset` might be a useful option to help us reset state between e2e tests.

test/e2e/specs/npm_spec.js
```diff
@@ -36,7 +36,12 @@ describe('OIDC flows', function () {
     primaryAuth.loginToForm('{{{WIDGET_BASIC_USER_3}}}', '{{{WIDGET_BASIC_PASSWORD_3}}}');
     oktaHome.waitForPageLoad();
     Expect.toBeA11yCompliant();
-    expect(oktaHome.getLoggedInUser()).toEqual('Test');
+
+    // userName element is not visible in mobile browsers (have to click a dropdown, not a bug)
+    if (process.env.SAUCE_PLATFORM_NAME === 'iOS') {
+      return;
+    }
+    expect(oktaHome.getLoggedInUser()).toEqual('Test'); // will fail on mobile browsers
```
jmelberg-okta: As a follow-up, can we modify the `getLoggedInUser` method to click the dropdown to retrieve this value instead of terminating early from this test?

scripts/travis.sh
vijetmahabaleshwar-okta: The main reason was travis run time increased by almost 50%. The existing tests take around 15 mins. The iOS tests run on 3 device emulators (serially) and take around 15 - 18 mins. After discussing it with @robertjd and @manueltanzi-okta, we decided it's good to start this as a cron job and explore if tests can be run in parallel.
As for Android tests, there are a couple more failing tests in oidc_spec. Initially I wanted to get both iOS + Android in this PR but then I have to add more disable logic in tests. Hence decided against it. Will work on Android in the next sprint.

test/e2e/specs/npm_spec.js
vijetmahabaleshwar-okta: Actually, I just tested this. In the mobile browser, the username is not displayed even in the drop-down. It just shows the `settings` and `signout` links as shown below.
Screen Shot 2019-04-11 at 9 15 55 AM

vijetmahabaleshwar-okta: This is how the username (George) is displayed in desktop browser, which is what we look for - 

Screen Shot 2019-04-11 at 9 17 38 AM

jmelberg-okta: Gotcha - I misunderstood this comment:

```
// will fail on mobile browsers
```

Can we change it to:

```suggestion
    expect(oktaHome.getLoggedInUser()).toEqual('Test'); // This assertion will always fail because mobile browsers hide the username (expected behavior)
```

scripts/travis.sh
jmelberg-okta: Gotcha - that significantly slows down development time.

The primary reason I'm calling this out is in the past, tests that have failed via CRON jobs have been neglected. The `htmlproofer` tests within D.O.C were a great example of this. 

Another problem with cron tasks is delegating **who** is responsible for resolving the failing tests, as we'd need to track down which commit introduced the failures.

To get around this, what are your thoughts on running the Appium tests within Bacon _instead_ of Travis? When new versions of the Widget are being promoted, we'll run the Appium suite in parallel with the existing tests. That way - we can guarantee that no new versions of the Widget can be released without passing the mobile browser tests.

What do you think?

test/e2e/conf.js
vijetmahabaleshwar-okta: Do you want to have a separate config file for desktop, iOS and android? Or have a common config and split specific configs into separate files which are imported into common config?

test/e2e/specs/npm_spec.js
haishengwu-okta: The element is actually present in the DOM but not visible. Maybe we shall try to adjust the assertion like
- expect the element present and has text `Test`
- element is visible at normal browser but not mobile (smaller screen)

scripts/travis.sh
haishengwu-okta: > When new versions of the Widget are being promoted, we'll run the Appium suite in parallel with the existing tests

jmelberg-okta: > What about 'when an new commit hit master, run Appium suite'?

I thought about that originally, but then we're forced to react to the possibility of broken tests that have already been merged. Ideally we'll catch errors **before** the topic branch has been merged.

This gives us two options:
1. Run the Appium tests against all topic branches with the side-effect of doubling our CI time (~30min)
2. Run the Appium tests in Bacon to support parallelism. Similar to `bacon-topic-*` prefixes, we can run this against topic or `master`.

I'd advocate for option two, as it keeps our Travis CI time minimal. Plus, I'd assume the primary reason for running the entire browser suite (embedded + mobile) is for major/minor releases or drastic view changes. We'd just need to encourage internal devs to prefix their branches accordingly for changes that _may_ break embedded or mobile browser tests.

> and we might want to set up the testing in an way that developer has option to run those test in topic branch

Somewhat related to the first point - I'd assume we'd want this option to be available for internal contributors using our internal CI to provide some sense of a sanity check. If we wanted all external PRs to run our Appium tests (probably don't want to for security reasons), we can always keep our existing pattern of fetching an external contributor's topic branch and pushing it to a remote branch ourselves.

test/e2e/conf.js
jmelberg-okta: Ideal state would be to conditionally run our e2e tests for mobile and another for desktop. We _could_ split this into multiple `yarn` commands, but that might be overkill for now.

To make this easy to consume - what about:

```js
if (process.env.SAUCE_PLATFORM_NAME === 'iOS') {
  var appium = require('./appium-config.js');
  config.port = appium.port;
  config.multiCapabilities = appium.iOSCapabilities;
} else {
  // Desktop
  ...
}
```

Where `appium-config.js` is an exported object broken into fields that we care about.

scripts/travis.sh
vijetmahabaleshwar-okta: Running on bacon seems like a good approach.
I see that we run unit tests on bacon today for master and bacon-topic* branches. On the topic branch, does it block merges if unit tests fail?

test/e2e/specs/npm_spec.js
vijetmahabaleshwar-okta: Actually, the element doesn't seem to have the text in mobile browser. Here's the error I got on mobile browser - 
```
2) OIDC flows can login and auth in a basic flow
  Message:
    Expected '' to equal 'Test'.
  Stack:
    Error: Failed expectation
      at UserContext.<anonymous> (/home/travis/build/okta/okta-signin-widget/target/e2e/specs/npm_spec.js:39:40)
      at /home/travis/build/okta/okta-signin-widget/node_modules/jasminewd2/index.js:112:25
      at new ManagedPromise (/home/travis/build/okta/okta-signin-widget/node_modules/selenium-webdriver/lib/promise.js:1077:7)
      at ControlFlow.promise (/home/travis/build/okta/okta-signin-widget/node_modules/selenium-webdriver/lib/promise.js:2505:12)
      at schedulerExecute (/home/travis/build/okta/okta-signin-widget/node_modules/jasminewd2/index.js:95:18)
      at TaskQueue.execute_ (/home/travis/build/okta/okta-signin-widget/node_modules/selenium-webdriver/lib/promise.js:3084:14)
      at TaskQueue.executeNext_ (/home/travis/build/okta/okta-signin-widget/node_modules/selenium-webdriver/lib/promise.js:3067:27)
      at asyncRun (/home/travis/build/okta/okta-signin-widget/node_modules/selenium-webdriver/lib/promise.js:2974:25)
      at /home/travis/build/okta/okta-signin-widget/node_modules/selenium-webdriver/lib/promise.js:668:7
      at <anonymous>:null:null
      at process._tickCallback (internal/process/next_tick.js:189:7)
```

vijetmahabaleshwar-okta:
@jmelberg-okta - I have addressed all your concerns in the last PR.

* For now, I suggest we still run tests on cron job for iOS (only temporary)
* I have tested running this on bacon and it works. Only issue is some secure env vars in `.travis.yml` need to stored in aws so that we can access them from bacon script
* I will take that up in a separate PR

scripts/travis.sh
jmelberg-okta: It doesn't block merges since merges are handled by Travis. The bacon regex is primarily used internally as validation that we're not breaking any of our core sign-in flows.

As I understand the current flow, when we're making drastic changes or getting ready to release minor/major versions, we run our internal CI suite for a sanity check. There is no _hard_ requirement for this, but oftentimes will get called out in a PR review.

test/e2e/appium/ios-conf.js
```diff
@@ -0,0 +1,47 @@
+/*!
+ * Copyright (c) 2015-2016, Okta, Inc. and/or its affiliates. All rights reserved.
```
jmelberg-okta: nit: 2019-Present

PR #671, opened 2019-04-08
Fix security image loading issue for logins with "+" symbols in them

willgantt-okta:
### Description

- Since username is in the query string when we hit the endpoint to fetch the security image upon login (`/login/getImage?username=`), we were not properly retrieving images for users with a "+" in their username because the symbol was being interpreted as a space. This PR properly escapes "+" symbols in the username.
- Resolves [OKTA-210776](https://oktainc.atlassian.net/browse/OKTA-210776)

### Reviewers

@haishengwu-okta @okta/idx

### DIFF
--- a/src/models/AppState.js
+++ b/src/models/AppState.js
@@ -39,6 +39,8 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
   var securityImageUrlTpl = compile('{{baseUrl}}/login/getimage?username={{username}}');
 
   function getSecurityImage (baseUrl, username, deviceFingerprint) {
+    // Reserved characters in the username must be escaped before the query can be safely executed
+    username = encodeURIComponent(username);
     var url = securityImageUrlTpl({ baseUrl: baseUrl, username: username });
 
     // When the username is empty, we want to show the default image.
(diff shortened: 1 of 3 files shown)


src/models/AppState.js
```diff
@@ -39,6 +39,8 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
   var securityImageUrlTpl = compile('{{baseUrl}}/login/getimage?username={{username}}');
 
   function getSecurityImage (baseUrl, username, deviceFingerprint) {
+    // Any "+" symbols in the username must be escaped, since it's considered a special character in the query string
+    username = username.replace(/\+/g, '%2B');
```
mauriciocastillosilva-okta: might be easier to use `encodeURIComponent` so we cover all possible scenarios. Or is it really just `+` that can cause a problem with the username?

royalchan-okta: +1 to using a full encode

Might as well cover our bases. If a bug creates users with `&` then we open ourselves up to `&username=foobar`, not that there's much risk in it besides breaking the image.

willgantt-okta: Good catch. Just fixed it.

willgantt-okta: @royalchan-okta Encode the full query string? How would a bug introduce a `&` into the param?

PR #674, opened 2019-04-09
Push state information of hidden links

iuriipetrov-okta:
[OKTA-210961](https://oktainc.atlassian.net/browse/OKTA-210961)

### DIFF
--- a/src/views/shared/Footer.js
+++ b/src/views/shared/Footer.js
@@ -21,10 +21,11 @@ function (Okta) {
 
   return Okta.View.extend({
     template: '\
-      <a href="#" data-se="needhelp" class="link help js-help">\
+      <a href="#" data-se="needhelp" aria-expanded="false" \
+        aria-controls="help-links-container" class="link help js-help">\
       {{i18n code="needhelp" bundle="login"}}\
       </a>\
-      <ul class="help-links js-help-links">\
+      <ul class="help-links js-help-links" id="help-links-container">\
         <li>\
         <a href="#" data-se="forgot-password" class="link js-forgot-password">\
         {{i18n code="forgotpassword" bundle="login"}}\
@@ -72,7 +73,10 @@ function (Okta) {
     },
     toggleLinks: function (e) {
       e.preventDefault();
-      this.$('.js-help-links').slideToggle(200);
+
+      this.$('.js-help-links').slideToggle(200, _.bind(function () {
+        this.$('.js-help').attr('aria-expanded', this.$('.js-help-links').is(':visible'));
+      }, this));
     },
     events: {
       'click .js-help': function (e) {
--- a/test/unit/spec/IDPDiscovery_spec.js
+++ b/test/unit/spec/IDPDiscovery_spec.js
@@ -358,6 +358,19 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm, 
           expect(test.form.helpLinkHref()).toBe('https://bar.com');
         });
       });
+      itp('has helpFooter with right aria-attributes and default values', function () {
+        return setup().then(function (test) {
+          expect(test.form.helpFooter().attr('aria-expanded')).toBe('false');
+          expect(test.form.helpFooter().attr('aria-controls')).toBe('help-links-container');
+        });
+      });
+      itp('sets aria-expanded attribute correctly when clicking help', function () {
+        return setup().then(function (test) {
+          expect(test.form.helpFooter().attr('aria-expanded')).toBe('false');
+          test.form.helpFooter().click();
+          expect(test.form.helpFooter().attr('aria-expanded')).toBe('true');
+        });
+      });
       itp('has a forgot password link', function () {
         return setup().then(function (test) {
           expect(test.form.forgotPasswordLabel().trim()).toBe('Forgot password?');
--- a/test/unit/spec/PrimaryAuth_spec.js
+++ b/test/unit/spec/PrimaryAuth_spec.js
@@ -427,6 +427,19 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
           expect(cb.length).toBe(0);
         });
       });
+      itp('has helpFooter with right aria-attributes and default values', function () {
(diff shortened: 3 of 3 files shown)


iuriipetrov-okta:
@mauriciocastillosilva-okta @gowthamidommety-okta

src/views/shared/Footer.js
```diff
@@ -72,6 +76,12 @@ function (Okta) {
     },
     toggleLinks: function (e) {
       e.preventDefault();
+      var revertedVal = !(
```
mauriciocastillosilva-okta: might be easier if we use the visibility of the element... `this.$('.js-help').is(':visible')`

src/views/shared/Footer.js
```diff
@@ -72,6 +76,12 @@ function (Okta) {
...
+        this.$('.js-help').attr('aria-expanded') === true ||
+        this.$('.js-help').attr('aria-expanded') === 'true'
+      );
+
+      this.$('.js-help').attr('aria-expanded', revertedVal);
```
mauriciocastillosilva-okta: this works. But at this point the animation is about to start.. perhaps we can change the `aria-expanded` value once the toggle animation finishes? 
```
this.$('.js-help-links').slideToggle(200, function() {
  var expanded = this.$('.js-help').is(':visible');
  this.$('.js-help').attr('aria-expanded', expanded);
});
```

src/views/shared/Footer.js
iuriipetrov-okta: Because it has an animation we need to wait in order to use this approach.

src/views/shared/Footer.js
iuriipetrov-okta: If pick this, so the first question isn't relevant. So do you think it's fine to toggle the state after the animation, not before?

src/views/shared/Footer.js
```diff
@@ -21,10 +21,14 @@ function (Okta) {
 
   return Okta.View.extend({
     template: '\
-      <a href="#" data-se="needhelp" class="link help js-help">\
+      <a href="#" \
```
gowthamidommety-okta: nit: maybe just keep this on two lines. I don't think we need a line for each attribute

src/views/shared/Footer.js
gowthamidommety-okta: I think it is fine to change the state after the animation. It is only fully expanded or hidden when the animation is complete.

iuriipetrov-okta:
> Are there tests we can add for this?

I think we don't need tests here. And we also don't have `_spec` file for `Footer`.

gowthamidommety-okta:
> > Are there tests we can add for this?
> 
> I think we don't need tests here. And we also don't have `_spec` file for `Footer`.

We can add tests to ensure aria-expanded is set correctly. There are tests for the help links in PrimaryAuth_spec and IDPDiscovery

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -358,6 +358,21 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm,
...
+        return setup().then(function (test) {
+          expect(test.form.helpFooter().attr('aria-expanded')).toBe('false');
+          expect(test.form.helpFooter().attr('aria-controls')).toBe('help-links-container');
+        });
+      });
+      itp('helpFooter toggles aria-expanded correct', function () {
+        return setup().then(function (test) {
+          expect(test.form.helpFooter().attr('aria-expanded')).toBe('false');
+          test.form.helpFooter().click();
+          return tick(test);
```
gowthamidommety-okta: is this to wait for the animation to complete?

iuriipetrov-okta: Yes.

src/views/shared/Footer.js
```diff
@@ -72,7 +73,10 @@ function (Okta) {
     },
     toggleLinks: function (e) {
       e.preventDefault();
-      this.$('.js-help-links').slideToggle(200);
+
+      this.$('.js-help-links').slideToggle(200, _.bind(function () {
+        this.$('.js-help').attr('aria-expanded', this.$('.js-help-links').is(':visible'));
```
gowthamidommety-okta: Instead of using `tick()` in our tests, could we move this to named function? Then I think we could spy on it and use something like `Expect.waitForSpyCall`

test/unit/spec/IDPDiscovery_spec.js
gowthamidommety-okta: See comment below. Basically, trying to avoid using tick().

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -358,6 +358,19 @@ function (Q, OktaAuth, WidgetUtil, Okta, Util, AuthContainer, IDPDiscoveryForm,
           expect(test.form.helpLinkHref()).toBe('https://bar.com');
         });
       });
+      itp('has helpFooter with right aria-attributes and default values', function () {
+        return setup().then(function (test) {
+          expect(test.form.helpFooter().attr('aria-expanded')).toBe('false');
+          expect(test.form.helpFooter().attr('aria-controls')).toBe('help-links-container');
+        });
+      });
+      itp('helpFooter toggles aria-expanded correct', function () {
```
gowthamidommety-okta: nit: sets aria-expanded correctly when clicking help

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -427,6 +427,19 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
           expect(cb.length).toBe(0);
         });
       });
+      itp('has helpFooter with right aria-attributes and default values', function () {
+        return setup().then(function (test) {
+          expect(test.form.helpFooter().attr('aria-expanded')).toBe('false');
+          expect(test.form.helpFooter().attr('aria-controls')).toBe('help-links-container');
+        });
+      });
+      itp('helpFooter toggles aria-expanded correct', function () {
```
gowthamidommety-okta: nit: sets aria-expanded correctly when clicking help

PR #676, opened 2019-04-11
 fix: Accessibility issues for factor dropdown icon

gowthamidommety-okta:
- Adds screen reader text for factor dropdown
- Makes factor dropdown navigable via the keyboard
- Adds aria-expanded and aria-controls to the dropdown

Using MacOS accessibility voice over:

accessibility3

### DIFF
--- a/packages/@okta/courage-for-signin-widget/package.json
+++ b/packages/@okta/courage-for-signin-widget/package.json
@@ -7,7 +7,7 @@
     "build": "webpack"
   },
   "devDependencies": {
-    "@okta/courage": "4.6.0-beta.2098.gc581fa3",
+    "@okta/courage": "4.6.0-beta.2360.g478adf3",
     "webpack": "^3.5.4",
     "webpack-bundle-analyzer": "^3.0.2"
   }
--- a/src/views/mfa-verify/dropdown/FactorsDropDown.js
+++ b/src/views/mfa-verify/dropdown/FactorsDropDown.js
@@ -12,9 +12,11 @@
 
 define([
   'okta',
+  'util/FactorUtil',
+  '../../../models/Factor',
   './FactorsDropDownOptions'
 ],
-function (Okta, FactorsDropDownOptions) {
+function (Okta, FactorUtil, Factor, FactorsDropDownOptions) {
   var { _, $ } = Okta;
   var { BaseDropDown } = Okta.internal.views.components;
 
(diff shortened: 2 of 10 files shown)


packages/@okta/courage-for-signin-widget/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "webpack"
   },
   "devDependencies": {
-    "@okta/courage": "4.6.0-beta.2098.gc581fa3",
+    "@okta/courage": "4.6.0-alpha.2340.g2dc3cd2",
```
haishengwu-okta: just an reminder that don't forget to change to beta before merge.

gowthamidommety-okta: yup, will do!

src/views/mfa-verify/dropdown/FactorsDropDown.js
```diff
@@ -28,13 +29,32 @@ function (Okta, FactorsDropDownOptions) {
 
   return BaseDropDown.extend({
     className: 'bg-helper icon-button',
+    screenReaderText: function () {
+      var factors = this.options.appState.get('factors'),
```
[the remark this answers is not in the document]
gowthamidommety-okta: 👍

src/views/mfa-verify/dropdown/FactorsDropDown.js
```diff
@@ -30,18 +30,8 @@ function (Okta, FactorsDropDownOptions, Factor) {
...
-        factor = factors.findWhere(_.pick(this.options, 'provider', 'factorType'));
-        if (factor === undefined) {
-          //for factors that support cardinality and only have factorType
-          factor = factors.findWhere(_.pick(this.options, 'factorType'));
-        }
-      } else  {
-        factor = new Factor.Model(this.options.appState.get('factor'), this.toJSON());
-      }
-      factorLabel = factor.get('factorLabel');
+      var factor = FactorUtil.getCurrentFactor(_.extend(this.options, {toJSON: this.toJSON}));
```
mauriciocastillosilva-okta: didn't realize in the beginning that we would need to pass the whole JSOn of this dropdown... not sure that is the right thing to do
What about
```
var factors = this.options.appState.get('factors'), factor;
if (factors) {
  factor = FactorUtil.findFactorInFactorsArray(factors, this.options.provider, this.options.factorType);
} else  {
  factor = new Factor.Model(this.options.appState.get('factor'), this.toJSON());
}
factorLabel = factor.get('factorLabel');
```

gowthamidommety-okta: we don't need everything.. But we do need appstate, provider and factorType besides toJSON. Should I limit to those options? I like how clear it is when we just use FactorUtil.getCurrentFactor

gowthamidommety-okta: Spoke to Mau offline. Changing this

PR #677, opened 2019-04-12
fix: disables submit button when submitting change password form

mauriciocastillosilva-okta:
- Similar to [reset password](https://github.com/okta/okta-signin-widget/blob/master/src/PasswordResetController.js#L36-L45), we now disable "Change password" when the user clicks on the button
- Taking over from https://github.com/okta/okta-signin-widget/pull/663, which was intended to fix the same thing

Resolves: OKTA-207909

changepassword-disableAfterSubmit


nbarbettini:
Excellent UX improvement, thanks @mauriciocastillosilva-okta 🙂

PR #680, opened 2019-04-16
chore: upgrade handlebars to 4.0.14

mauriciocastillosilva-okta:


package.json
```diff
@@ -115,7 +115,7 @@
     "babel-runtime": "^6.26.0",
     "backbone": "1.2.1",
     "clipboard": "1.6.1",
-    "handlebars": "4.0.13",
+    "handlebars": "4.0.14",
```
swiftone: Why not 4.1.2?

mauriciocastillosilva-okta: discussed offline, we will go with 4.0.14 for now

PR #690, opened 2019-04-23
Runs e2e iOS tests on bacon (using sauce labs)

vijetmahabaleshwar-okta:
* Adds a script to run e2e tests for iOS on [bacon](https://bacon.aue1e.internal/#!/commits/okta-signin-widget)
* A new optional test suite `e2e-iOS` is now available on bacon (Click on `show all test suites` in bacon)
* Devs can run tests on iOS for their PRs using this test suite


scripts/e2e-iOS.sh
```diff
@@ -0,0 +1,71 @@
...
+export TRAVIS_BUILD_NUMBER=${TEST_SUITE_RESULT_ID}
+
+cd ${OKTA_HOME}/${REPO}
+
+# Download and start sauce connect
+curl -o ${OKTA_HOME}/${REPO}/sc-4.5.3-linux.tar.gz https://saucelabs.com/downloads/sc-4.5.3-linux.tar.gz
+tar -xzf ${OKTA_HOME}/${REPO}/sc-4.5.3-linux.tar.gz
+${OKTA_HOME}/${REPO}/sc-4.5.3-linux/bin/sc -u ${SAUCE_USERNAME} -k ${SAUCE_ACCESS_KEY} -i ${TRAVIS_JOB_NUMBER} &
+
+aws s3 --quiet --region us-east-1 cp s3://ci-secret-stash/prod/signinwidget/test_credentials ./test_credentials.yaml
```
jmelberg-okta: Just to double check - do we need to pass in a  [`--profile`](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html)?

scripts/e2e-iOS.sh
```diff
@@ -0,0 +1,71 @@
...
+export WIDGET_BASIC_USER=$(cat ./test_credentials.yaml | yq .WIDGET_BASIC_USER | tr -d '"')
+export WIDGET_BASIC_PASSWORD=$(cat ./test_credentials.yaml | yq .WIDGET_BASIC_PASSWORD | tr -d '"')
+export WIDGET_BASIC_USER_2=$(cat ./test_credentials.yaml | yq .WIDGET_BASIC_USER_2 | tr -d '"')
+export WIDGET_BASIC_PASSWORD_2=$(cat ./test_credentials.yaml | yq .WIDGET_BASIC_PASSWORD_2 | tr -d '"')
+export WIDGET_BASIC_USER_3=$(cat ./test_credentials.yaml | yq .WIDGET_BASIC_USER_3 | tr -d '"')
+export WIDGET_BASIC_PASSWORD_3=$(cat ./test_credentials.yaml | yq .WIDGET_BASIC_PASSWORD_3 | tr -d '"')
+export WIDGET_BASIC_USER_4=$(cat ./test_credentials.yaml | yq .WIDGET_BASIC_USER_4 | tr -d '"')
+export WIDGET_BASIC_PASSWORD_4=$(cat ./test_credentials.yaml | yq .WIDGET_BASIC_PASSWORD_4 | tr -d '"')
+export WIDGET_BASIC_USER_5=$(cat ./test_credentials.yaml | yq .WIDGET_BASIC_USER_5 | tr -d '"')
+export WIDGET_BASIC_PASSWORD_5=$(cat ./test_credentials.yaml | yq .WIDGET_BASIC_PASSWORD_5 | tr -d '"')
```
jmelberg-okta: super-nit: Might be easier to maintain in the long term if you loop over a list of known environment variables, and run this command instead of running the same command on multiple lines.

scripts/e2e-iOS.sh
```diff
@@ -0,0 +1,71 @@
...
+
+setup_service grunt
+
+# Install required dependencies
+npm install -g @okta/ci-update-package
```
jmelberg-okta: Please use `yarn` instead of `npm`, that way the `update_yarn_locks` method will work as expected. Otherwise, we'll output a `package-lock.json` instead of `yarn.lock`

scripts/e2e-iOS.sh
```diff
@@ -0,0 +1,71 @@
...
+    # Replace yarn artifactory with Okta's
+    sed -i "s#${YARN_REGISTRY}#${OKTA_REGISTRY}#" test/e2e/react-app/yarn.lock
+    sed -i "s#${YARN_REGISTRY}#${OKTA_REGISTRY}#" test/e2e/angular-app/yarn.lock
+}
+
+update_yarn_locks
+
+if ! npm run test:e2e; then
+  echo "e2e tests on iOS failed! Exiting..."
+  exit ${FAILURE}
```
jmelberg-okta: To ensure failures are outputted correctly, lets use `PUBLISH_TYPE_AND_RESULT_DIR_BUT_ALWAYS_FAIL`.

vijetmahabaleshwar-okta: I think we use this only when there is a result dir with results xml file. AFAIK, eng-prod asks us to use `{SUCCESS}` and `{FAILURE}` for other cases. Let me cross-check anyway.

jmelberg-okta: You're right - it looks like we're not outputting failures to a directory for our e2e tests currently. `SUCCESS` works here.

scripts/e2e-iOS.sh
vijetmahabaleshwar-okta: We don't need to. This is the format we've used in other scripts too. Working fine... 🤷‍♂

scripts/e2e-iOS.sh
```diff
@@ -0,0 +1,68 @@
...
+for WIDGET_ENV_VAR in "${WIDGET_ENV_VARS[@]}"
+do
+   export $WIDGET_ENV_VAR=$(cat ./test_credentials.yaml | yq .${WIDGET_ENV_VAR} | tr -d '"')
+done
+
+setup_service grunt
+
+# Install required dependencies
+yarn install -g @okta/ci-update-package
+yarn install -g @okta/ci-pkginfo
```
jmelberg-okta: As it currently sits, this script _should_ fail. Instead, this line should read: `yarn global add @okta/ci-update-package` and `yarn global add @okta/ci-pkginfo` respectively.

scripts/e2e-iOS.sh
```diff
@@ -0,0 +1,68 @@
...
+
+if ! npm install --no-optional --unsafe-perm; then
```
jmelberg-okta: Similarly, we need to use `yarn` here otherwise we're not locking our dependencies.

`yarn install --ignore-optional`

Also - we probably do not need the `--unsafe-perm` flag.

scripts/e2e-iOS.sh
jmelberg-okta: Since these are for publishing, we can probably remove them.

PR #691, opened 2019-04-25
Add a11y enhancements for form errors (from courage)

aarongranick-okta:
Brings in latest version of courage which contains fixes for these issues:
https://oktainc.atlassian.net/browse/OKTA-210971?filter=-1
https://oktainc.atlassian.net/browse/OKTA-210941?filter=-1
See https://github.com/okta/okta-ui/commit/a2bfd2d7b09cf1be655e055d88d6aed89df23607

Also fixes this issue
https://oktainc.atlassian.net/browse/OKTA-210937?filter=-1

Brings in latest version of i18n including new string for error icon aria-label


haishengwu-okta:
add to `login.properties` would be fine.

packages/@okta/i18n/package.json
```diff
@@ -1,33 +1,31 @@
 {
-  "_from": "@okta/i18n@54.0.0-beta.g5d8e734",
```
haishengwu-okta: doesn't seem right. how did you copy over those i18n files?

aarongranick-okta: This is the package.json from the npm module on my local hard drive

I copied the other files (.properties, .json) from the dist/ folder of the npm module on my local hard drive.

PR #692, opened 2019-04-26
chore: Upgrade okta-auth-js in okta-signin-widget 3.0

manueltanzi-okta:
### Description

- Upgrade okta-auth-js in okta-signin-widget 3.0
- Use cross-fetch auth-js version instead of jquery one

**Note:** Will go in okta-signin-widget v3.0.0.

### Reviewers

@gowthamidommety-okta @robertdamphousse-okta

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -61,12 +61,11 @@ define(['okta', './Logger', './Enums'], function (Okta, Logger, Enums) {
       return xhr;
     }
     if (!xhr.responseJSON) {
-      try {
-        xhr.responseJSON = JSON.parse(xhr.responseText);
-      } catch (parseException) {
+      if (!xhr.responseText) {
         xhr.responseJSON = { errorSummary: Okta.loc('oform.error.unexpected', 'login') };
         return xhr;
       }
+      xhr.responseJSON = xhr.responseText;
     }
     // Temporary solution to display field errors
     // Assuming there is only one field error in a response
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -95,7 +95,7 @@ var OktaSignIn = (function () {
   function OktaSignIn (options) {
     require('okta');
 
-    var OktaAuth = require('@okta/okta-auth-js/jquery');
+    var OktaAuth = require('@okta/okta-auth-js');
     var Util = require('util/Util');
     var LoginRouter = require('LoginRouter');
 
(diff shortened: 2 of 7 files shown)


src/util/Util.js
```diff
@@ -61,12 +61,11 @@ define(['okta', './Logger', './Enums'], function (Okta, Logger, Enums) {
...
     }
     if (!xhr.responseJSON) {
-      try {
-        xhr.responseJSON = JSON.parse(xhr.responseText);
-      } catch (parseException) {
+      if (!xhr.responseText) {
         xhr.responseJSON = { errorSummary: Okta.loc('oform.error.unexpected', 'login') };
         return xhr;
       }
+      xhr.responseJSON = xhr.responseText;
```
manueltanzi-okta: **Note:** The logic of how okta-auth-js returns errors changed a little bit in the new version, this is why this change. The new version of okta-auth-js return the responseText already parsed from JSON, so we don’t need to parse it again here (which would fail, since they are not anymore in JSON format). We need instead to check if that attribute is available.

src/widget/OktaSignIn.js
```diff
@@ -95,7 +95,7 @@ var OktaSignIn = (function () {
   function OktaSignIn (options) {
     require('okta');
 
-    var OktaAuth = require('@okta/okta-auth-js/jquery');
+    var OktaAuth = require('@okta/okta-auth-js');
```
manueltanzi-okta: **Note:** New version of okta-auth-js uses `cross-fetch` by default! (https://github.com/okta/okta-auth-js/pull/199), which is we want to use cross-fetch in version 3.0. So we don't need to specify `/jquery` version anymore.

(This has been already cross-browser + embedded browser tested for the widget when we made the change in okta-auth-js).

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1920,36 +1920,6 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
             expectErrorEvent(test, 429, 'API call exceeded rate limit due to too many requests.');
           });
       });
-      itp('shows an error if authClient returns with an error that is plain text', function () {
```
gowthamidommety-okta: Why are we removing this test? authClient could still return an error that is plain text right?

manueltanzi-okta: They authClient should always return the parsed object, while in this tests we were mocking that returns a string, which we don't parse anymore in the new way and which should throw this error: https://github.com/okta/okta-signin-widget/pull/692/files#diff-a2204dc4da481825a3d452cdd6c378ccR65.

But good point, I should add test for that 👍

PR #693, opened 2019-05-01
Runs e2e tests on chrome headless instead of saucelabs

vijetmahabaleshwar-okta:
* We are constantly hitting the saucelabs VM limits for our license (https://oktainc.atlassian.net/browse/OKTA-224058)
* I noticed that we are running chrome browser e2e tests on saucelabs which can be done on travis headless chrome instead
* This change removes the saucelabs dependency for travis builds

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "start:angular": "yarn --cwd test/e2e/angular-app start & wait-on http-get://localhost:4200",
     "pretest:e2e": "yarn install:react && yarn install:angular && node pretest.js",
     "test:e2e": "yarn build:release && yarn start:react && yarn start:angular && grunt test-e2e",
-    "protractor": "webdriver-manager update && protractor target/e2e/conf.js",
+    "protractor": "webdriver-manager update --versions.standalone=3.141.59 --gecko false && protractor target/e2e/conf.js",
     "build:dev": "grunt build:dev",
     "build:release": "grunt build:release",
     "build:webpack-dev": "webpack --config webpack.dev.config.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -44,7 +44,7 @@
     "start:angular": "yarn --cwd test/e2e/angular-app start & wait-on http-get://localhost:4200",
     "pretest:e2e": "yarn install:react && yarn install:angular && node pretest.js",
     "test:e2e": "yarn build:release && yarn start:react && yarn start:angular && grunt test-e2e",
-    "protractor": "webdriver-manager update && protractor target/e2e/conf.js",
+    "protractor": "webdriver-manager update --versions.standalone=3.141.59 --gecko false && protractor target/e2e/conf.js",
```
vijetmahabaleshwar-okta: Need to pin the version of selenium standalone due to an existing issue with protractor + selenium 4.x release.

test/e2e/conf.js
```diff
@@ -44,10 +43,10 @@ if (process.env.TRAVIS) {
 // WIDGET_BASIC_USER
 // WIDGET_BASIC_PASSWORD
 else {
-  const webdriverManegerConfig = require('webdriver-manager/selenium/update-config.json');
+  const webdriverManagerConfig = require('webdriver-manager/selenium/update-config.json');
```
jmelberg-okta: 👍

jmelberg-okta:
Just to confirm - we previously are **not** running cross-browser tests in Sauce, right? If so, we should reconsider this change and instead use Travis for all non-master or release candidates.

vijetmahabaleshwar-okta:
@jmelberg-okta - We were actually running chrome browser tests on saucelabs previously. But unless we want to change the tests to add browsers like ie, firefox, safari etc, there's no point in using it. Since travis chrome plugin is good enough and faster than sauce, I'm changing it.
Whether we need to test on different browsers would be a separate discussion (at which point we can go back to saucelabs)
Another option to have optional bacon test suites for safari, firefox, ie browsers (like we do in okta-core)

jmelberg-okta:
@vijetmahabaleshwar-okta - Perfect. Thats what I figured. Thanks!

PR #696, opened 2019-05-03
test:e2e: Remove npm pack / pretest. Use webpack config instead.

aarongranick-okta:
`node pretest.js` was causing modification to source files in react & angular test apps
this removes that step and the need for `npm pack`

This should speed up build time and reduce developer error when running e2e tests locally.


haishengwu-okta:
how about `angular-app`?

aarongranick-okta:
> how about `angular-app`?

I am working on this now, removing the pretest and loading the environment variables through webpack.

PR #700, opened 2019-05-05
E2E Test Environment Vars

aarongranick-okta:
centralizes use of environment variables in E2E tests. Used by:
- grunt
- protractor
- angular app
- react app

Replaces hardcoded values with new vars & "defaults": WIDGET_CLIENT_ID, and user names.

Adds a basic static web server for testing the "basic" specs

Adds README/instructions for manual E2E testing

Upgrades angular test app (to v.7) and receives all params from ENV vars (remove pretest step)
Some fixes to angular app & use latest version of OIDC.


test/e2e/README.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+You should only need to do this once.
+
+## Run specs
+
+To see protractor options simply run:
+
+```
+npx protractor
```
haishengwu-okta: curious why suggest to use `npx` over `yarn`?

test/e2e/angular-app/extra-webpack.config.js
```diff
@@ -0,0 +1,22 @@
...
+const webpack = require('webpack');
+const env = {};
+
+// List of environment variables made available to the app
+[
+  'PORT',
+  'WIDGET_TEST_SERVER',
+  'WIDGET_CLIENT_ID'
+].forEach(function (key) {
+  env[key] = JSON.stringify(process.env[key]);
```
haishengwu-okta: why need `stringify`?

test/e2e/angular-app/tslint.json
```diff
@@ -18,7 +18,6 @@
     "forin": true,
     "import-blacklist": [
       true,
-      "rxjs",
```
haishengwu-okta: curious why this got removed but not the other?

test/e2e/README.md
aarongranick-okta: This is literally running ./node_modules/.bin/protractor

`yarn protractor` would run the "protractor" script in package.jsonm

aarongranick-okta:
> curious why angular-app and react-app need `dotenv` dependency where variables are pass to those app via process.env?

All the env handling is centralized in test/e2e/env.js  This file uses dotenv and a set of defaults to build a consistent set of environment variables for each process.

React & Angular apps run in their own process and load this "env.js" file to get the same set of environment variables as grunt, etc. (even if none are passed on CLI, which is preferred)

test/e2e/angular-app/tslint.json
aarongranick-okta: This file was auto-modified by `ng upgrade`

test/e2e/angular-app/extra-webpack.config.js
aarongranick-okta: If the value is a string it must be "quoted" or we will get syntax error when webpack substitutes it in the source.

PR #703, opened 2019-05-07
fix: Fixes css priority order

manueltanzi-okta:
### Description

Introduce a bug in 3.0 branch in this PR: https://github.com/okta/okta-signin-widget/pull/650
okta-theme.css was intended to be imported after okta-sign-in.css, so, to keep the same priority order, the import needs to be moved at the end of the file.

### Reviewers

@hor-kanchan-okta @haishengwu-okta

### DIFF
--- a/assets/sass/okta-sign-in.scss
+++ b/assets/sass/okta-sign-in.scss
@@ -9,7 +9,6 @@
 @import 'container';
 @import 'modules/okta-footer';
 @import 'ie';
-@import 'okta-theme';
 
 /* Layout */
 .login-bg-image {
@@ -162,3 +161,5 @@
     background: transparent;
   }
 }
+
+@import 'okta-theme';


assets/sass/okta-sign-in.scss
```diff
@@ -162,3 +161,5 @@
     background: transparent;
   }
 }
+
```
aarongranick-okta: We are probably processing this with SASS so it may not matter. But if we switched to a lighter CSS processing system, like postCSS. Technically, all @import statements should precede all other rules: https://developer.mozilla.org/en-US/docs/Web/CSS/@import

In this case, the login rule should be @imported from a new file.

haishengwu-okta: sounds an bigger task to refactor the scss file ?

manueltanzi-okta: @aarongranick-okta Can you create a Jira if you feel this is something we should do in future? Thank you

PR #708, opened 2019-05-13
feat: Adds skip link during grace period for enrollment of required factors

gowthamidommety-okta:
Resolves: OKTA-225332

skipLinkGracePeriod


test/unit/spec/EnrollChoices_spec.js
```diff
@@ -254,6 +254,11 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, Router,
             expect(test.form.submitButtonText()).toBe('Configure factor');
           });
         });
+        itp('does not have link to skip set up during grace period when all are required', function () {
```
mauriciocastillosilva-okta: I don't think we need to say anything about `grace period` in the tests in general. At the end are showing the link only when it makes sense (one required factor enrolled and some required factors still not enrolled).
So this test can be `does not have skip setup link when all are required and none are enrolled`

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -280,6 +285,11 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, Router,
               .toHaveBeenCalledWith('signin/enroll/google/token%3Asoftware%3Atotp', { trigger: true });
           });
         });
+        itp('has link to skip set up during grace period when some are required', function () {
```
mauriciocastillosilva-okta: and this could be `has skip setup link when it is present in the response and 1 required factor is enrolled and some required are not enrolled yet`

src/views/enroll-choices/Footer.js
```diff
@@ -0,0 +1,31 @@
+/*!
+ * Copyright (c) 2017, Okta, Inc. and/or its affiliates. All rights reserved.
```
mauriciocastillosilva-okta: nit: `2019`

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -305,6 +315,21 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, Router,
             );
           });
         });
+        itp('does not have link to skip set up during grace period when all factors are optional none enrolled', function () {
```
mauriciocastillosilva-okta: similar comment for the name of these tests. Widget does not really know anything about a grace period, all we know is that the skip link is present even if some required factors are not enrolled yet

gowthamidommety-okta: 👍 removing grace period for all tests

src/EnrollChoicesController.js
```diff
@@ -200,6 +201,11 @@ function (Okta, FormController, Enums, RouterUtil, FactorList,
       }
 
       this.state.set('pageType', pageType);
+
+      var transaction = this.options.appState.get('transaction');
```
mauriciocastillosilva-okta: looking at this again, I don't think this is the best way to do this. We shouldn't be getting the `transaction` here, that is kind of abstract for when we start and manage one.
I think we should put this inside `AppState`, similar to how we have [prevLink](https://github.com/okta/okta-signin-widget/blob/master/src/models/AppState.js#L430)

gowthamidommety-okta: ahh makes sense 👍

PR #711, opened 2019-05-22
OKTA-210935 - Enhance voice sms UX for timing out

haoyuli-okta:
The solution here is to include a callout warning that will appear after 30 seconds (along with the button text change to "Redial"). This will impact the Voice Call MFA enroll and verify widgets. Additionally, we want to add this to the SMS MFA enroll and verify experiences for consistency. I have included mocks for both in a compressed file attached to this ticket.

After 30 secs, the callout text:

voice call: "Haven't received a voice call? Select Redial to try again."
SMS: "Haven't received an SMS? Select Re-send code to try again."

UI Behavior Summary:
Eg. Enrollment SMS:
Click Send Code, Send Code button disabled. After 30 secs, SMS warning widget appears, and Send Code button enabled and button text changed to Re-send code. If Re-send code button clicked, warning widget disappeared, and Re-send code button disabled, after 30 secs, SMS warning widget appears and Re-send code button enabled.

Same idea applied to Voice Enrollment, SMS Challenge, Voice Challenge. 
01 - SMS Enroll - 30 sec notif
02 - SMS Verify - 30 sec notif
01a - voice call - enroll - 30 sec notif
02b - voice call - verify - 30 sec notif
### Box Video Link
MFA_SMS_CALL_CHALLENGE_WARNING
https://okta.box.com/s/v36sg9ekqyhyzuw0jcnicbjhwagyrog5
MFA_SMS_CALL_ENROLL_WARNING
https://okta.box.com/s/uuqxegbkje5ugfl84mkf8xv44f89doks

### I18N Repo Changes
https://github.com/okta/npm/pull/1232

### Resolves

- [OKTA-210935](https://oktainc.atlassian.net/browse/OKTA-210935)


src/EnrollCallAndSmsController.js
```diff
@@ -90,6 +97,8 @@ function (Okta, FormController, Footer, PhoneTextBox, TextBox, CountryUtil, Form
         }
       },
       limitResending: function () {
+        this.set({timeoutWarning: true});
```
mauriciocastillosilva-okta: since we show the warning callout at the same time as we enable the button with `Re-send`, can we use the same var we already have `ableToResend` instead of creating a new one? Or was there any reason to create a new var?

src/views/mfa-verify/PassCodeForm.js
```diff
@@ -87,6 +102,13 @@ define(['okta', 'q', 'views/shared/TextBox'], function (Okta, Q, TextBox) {
         click: function () {
           form.clearErrors();
           this.disable();
+          if (factorType =='call' || factorType == 'sms') {
```
mauriciocastillosilva-okta: let's use `===` for comparison
and nit, space after `=` and before `'`

src/views/mfa-verify/PassCodeForm.js
```diff
@@ -87,6 +102,13 @@ define(['okta', 'q', 'views/shared/TextBox'], function (Okta, Q, TextBox) {
...
+            var warning = Okta.loc(factorType == 'call' ? 'factor.call.time.warning' : 'factor.sms.time.warning', 'login');
```
mauriciocastillosilva-okta: we should add this to the response of the function `getFormAndButtonDetails`, that is where all these labels for sms or call are

src/views/mfa-verify/PassCodeForm.js
```diff
@@ -87,6 +102,13 @@ define(['okta', 'q', 'views/shared/TextBox'], function (Okta, Q, TextBox) {
...
+            form.clearWarnings();
+            warningTimeout = Util.callAfterTimeout(_.bind(function () {
```
mauriciocastillosilva-okta: we can probably do this at the same time as we change the button text and re-enable it right? In line 124, after saving. Otherwise it could happen that we will see this callout saying `click resend` but the button won't be enabled yet

src/views/mfa-verify/PassCodeForm.js
haoyuli-okta: fixed

src/views/mfa-verify/PassCodeForm.js
haoyuli-okta: fixed

src/EnrollCallAndSmsController.js
haoyuli-okta: make sense, I can fix it. I used the variable with the intention to indicate "true" to show a warning and "false" to disable the warning, boolean values are opposite to `ableToSend,` however, I am listening to the events with a CHANGE of the value, the logic should be the same, totally agreed.

src/views/mfa-verify/PassCodeForm.js
haoyuli-okta: Actually, the time out counter starts at the time the button is clicked and appears as it greys out. If we put this at line 124, that will wait for `Q.Delay` and do another 30s wait. Since they share the same timing variable `API_RATE_LIMIT`, it would appear at the same time. Click Resend and Warning  will appear at the same time.

src/views/mfa-verify/PassCodeForm.js
```diff
@@ -68,12 +76,22 @@ define(['okta', 'q', 'views/shared/TextBox'], function (Okta, Q, TextBox) {
       return this.model.appState.get('isMfaChallenge') && this.model.get('answer');
     },
 
+    showWarning: function (msg) {
+      clearTimeout(warningTimeout);
```
mauriciocastillosilva-okta: no need to `clearTimeout` here. Actually I don't think you need the variable `warningTimeout` anymore

test/unit/helpers/dom/EnrollSmsForm.js
```diff
@@ -64,4 +64,4 @@ define(['./Form'], function (Form) {
 
   });
 
-});
```
mauriciocastillosilva-okta: let's revert the change to this file, it is not needed

src/views/mfa-verify/PassCodeForm.js
```diff
@@ -51,7 +59,7 @@ define(['okta', 'q', 'views/shared/TextBox'], function (Okta, Q, TextBox) {
         buttonClassName: '',
         formSubmit: '',
         formRetry: '',
-        formSubmitted: '',
```
mauriciocastillosilva-okta: no need to remove the comma

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1110,6 +1110,40 @@ function (Okta,
             });
           });
       });
+      it('on WARNING_TIMEOUT, after send/re-send, shows warning message for SMS after 30s idle', function () {
+        Util.speedUpPolling();
+        return setupFn().then(function (test) {
+          spyOn(LoginUtil, 'callAfterTimeout').and.callFake(function () {
+            // reducing the timeout to 0 so that test is fast.
+            return setTimeout(arguments[0], 0);
```
mauriciocastillosilva-okta: PasscodeForm doesn't use `setTimeout` anymore right? So we need to change this test and remove all the references to `setTimeout`

src/EnrollCallAndSmsController.js
```diff
@@ -274,8 +283,28 @@ function (Okta, FormController, Footer, PhoneTextBox, TextBox, CountryUtil, Form
...
+          this.add(warningTemplate, '.o-form-error-container', {options: {warning: msg}});
+        },
+        clearWarnings: function () {
+          this.$('.okta-form-infobox-warning').remove();
+        },
+
         initialize: function () {
+          this.listenTo(this.model, 'change:ableToResend', function (model, ableToResend) {
+            if (!ableToResend) {
+              warningTimeout = Util.callAfterTimeout(_.bind(function () {
```
mauriciocastillosilva-okta: I don't think we need to create a timeout here... we probably can use the logic we already have for changing the button's text. Actually I don't think we need all any change to this initialize specifically. 
Take a look at the [initialize](https://github.com/okta/okta-signin-widget/pull/711/files#diff-12bdb1aff4a9e2a67ac933eb47538706R249) function for the retry button, that is called when `ableToResend` changes values. All that is already happening every 30 secs, that a look at [limitResending](https://github.com/okta/okta-signin-widget/pull/711/files#diff-12bdb1aff4a9e2a67ac933eb47538706R98) function

test/unit/helpers/dom/EnrollSmsForm.js
haoyuli-okta: fixed

src/views/mfa-verify/PassCodeForm.js
haoyuli-okta: fixed

test/unit/spec/MfaVerify_spec.js
haoyuli-okta: fixed

src/EnrollCallAndSmsController.js
haoyuli-okta: fixed by putting it under initialize of Form

src/EnrollCallAndSmsController.js
```diff
@@ -185,9 +191,11 @@ function (Okta, FormController, Footer, PhoneTextBox, TextBox, CountryUtil, Form
       var formSubmit = Okta.loc(isCall ? 'mfa.call' : 'mfa.sendCode', 'login');
       var formRetry = Okta.loc(isCall ? 'mfa.redial' : 'mfa.resendCode', 'login');
       var formSubmitted = Okta.loc(isCall ? 'mfa.calling' : 'mfa.sent', 'login');
-
+      
+      var warning = Okta.loc(isCall ? 'factor.call.time.warning' : 'factor.sms.time.warning', 'login');
       var numberFieldClassName = isCall ? 'enroll-call-phone' : 'enroll-sms-phone';
       var buttonClassName = isCall ? 'call-request-button' : 'sms-request-button';
+      var warningTimeout;
```
mauriciocastillosilva-okta: you don't need this variable anymore, we should remove it

src/EnrollCallAndSmsController.js
```diff
@@ -274,8 +282,24 @@ function (Okta, FormController, Footer, PhoneTextBox, TextBox, CountryUtil, Form
         noButtonBar: true,
         autoSave: true,
         className: getClassName(factorType),
+
+        showWarning: function (msg) {
+          clearTimeout(warningTimeout);
```
mauriciocastillosilva-okta: same here, this is no longer needed

src/EnrollCallAndSmsController.js
```diff
@@ -17,13 +17,19 @@ define([
   'views/enroll-factors/PhoneTextBox',
   'views/shared/TextBox',
   'util/CountryUtil',
-  'util/FormType'
+  'util/FormType',
+  'util/Util'
```
mauriciocastillosilva-okta: you don't need `Util` anymore, right?

src/EnrollCallAndSmsController.js
```diff
@@ -274,8 +282,24 @@ function (Okta, FormController, Footer, PhoneTextBox, TextBox, CountryUtil, Form
...
+        },
+        clearWarnings: function () {
+          this.$('.okta-form-infobox-warning').remove();
+        },
+
         initialize: function () {
+          this.listenTo(this.model, 'change:ableToResend', function (model, ableToResend) {
+            if (ableToResend) {
+              this.showWarning(warning);
+            }
```
mauriciocastillosilva-okta: I think we need the `else` to clear the warning right? Otherwise what happens if you click Re-send or Redial? Is the warning cleared?

src/views/mfa-verify/PassCodeForm.js
```diff
@@ -10,11 +10,17 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 /* eslint complexity: [2, 7] */
-define(['okta', 'q', 'views/shared/TextBox'], function (Okta, Q, TextBox) {
+define(['okta', 'q', 'views/shared/TextBox', 'util/Util'], function (Okta, Q, TextBox, Util) {
```
mauriciocastillosilva-okta: I don't think you need `Util` anymore

src/EnrollCallAndSmsController.js
```diff
@@ -274,8 +282,24 @@ function (Okta, FormController, Footer, PhoneTextBox, TextBox, CountryUtil, Form
...
+          });
           this.listenTo(this.model, 'error errors:clear', function () {
+            this.clearWarnings();
```
mauriciocastillosilva-okta: while this probably works, we are listening to `errors` in here, not `warnings`, so I think conceptually it will be better if we don't add this line. By adding the `else` to the listener of `ableToResend` the warning should be cleared

src/EnrollCallAndSmsController.js
haoyuli-okta: fixed

src/EnrollCallAndSmsController.js
haoyuli-okta: fixed

src/views/mfa-verify/PassCodeForm.js
haoyuli-okta: you are right since we removed CallAfterTimeout, util is no longer needed

src/EnrollCallAndSmsController.js
haoyuli-okta: This is added here because warnings and errors could happen at the same time. For example, `An SMS message was recently sent. Please wait 30 seconds before trying again.` could happen at the same time with `Haven't received an SMS? To try again, click Re-send code.` I added this line to clear all the errors together

src/EnrollCallAndSmsController.js
haoyuli-okta: fixed

src/EnrollCallAndSmsController.js
haoyuli-okta: this is handled by clearErrors events, which clear both warning and errors once you clicked re-send/re-dial, I think it is necessary to put it there since we have scenarios of both warning and errors happen in the same time

test/unit/spec/EnrollCall_spec.js
```diff
@@ -585,6 +585,38 @@ function (Q, Okta, OktaAuth, LoginUtil, Util, Form, Beacon, Expect, $sandbox,
           });
       });
 
+      itp('on WARNING_TIMEOUT, after dial/redail, shows warning message for Call after 30s idle', function () {
```
mauriciocastillosilva-okta: text should be just `shows warning message to click "Redial" after 30s`, cause there is no reference to WARNING_TIMEOUT in the tests. And actually that variable is only defined in PushForm

test/unit/spec/EnrollSms_spec.js
```diff
@@ -274,6 +274,39 @@ function (Okta, OktaAuth, LoginUtil, Util, AuthContainer, Form, Beacon, Expect,
             expect(test.form.hasErrors()).toBe(true);
           });
       });
+
+      itp('on WARNING_TIMEOUT, after send/re-send, shows warning message for SMS after 30s idle', function () {
```
mauriciocastillosilva-okta: text should be just `shows warning message to click "Re-send" after 30s`

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1110,6 +1110,36 @@ function (Okta,
             });
           });
       });
+      it('on WARNING_TIMEOUT, after send/re-send, shows warning message for SMS after 30s idle', function () {
```
mauriciocastillosilva-okta: same here, text should be just `shows warning message to click "Re-send" after 30s`

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1585,6 +1615,37 @@ function (Okta,
             expect(test.form.errorBox().length).toBe(0);
           });
       });
+
+      itp('on WARNING_TIMEOUT, after dail/re-dial, shows warning message for call after 30s idle', function () {
```
mauriciocastillosilva-okta: text should be just shows warning message to click "Redial" after 30s

test/unit/spec/EnrollCall_spec.js
haoyuli-okta: fixed

test/unit/spec/EnrollSms_spec.js
haoyuli-okta: fixed

test/unit/spec/MfaVerify_spec.js
haoyuli-okta: fixed

test/unit/spec/MfaVerify_spec.js
haoyuli-okta: fixed

src/views/mfa-verify/PassCodeForm.js
```diff
@@ -68,12 +75,21 @@ define(['okta', 'q', 'views/shared/TextBox'], function (Okta, Q, TextBox) {
...
+      this.$('.okta-form-infobox-warning').remove();
+    },
+
     initialize: function () {
       var form = this;
       this.title = this.model.get('factorLabel');
 
       var factorType = this.model.get('factorType');
       var formAndButtonDetails = getFormAndButtonDetails.call(this, factorType);
+      var warningDetials = formAndButtonDetails.warning;
```
mauriciocastillosilva-okta: there is a typo here, should be `warningDetails`

PR #712, opened 2019-05-28
Adds css for accessibility to set visual focus indication

bretterer:
**Description**
Changes the UX to have a visual indicator when focus is given to a field or button.

**Changes**
 - Adds a box-shadow to the `focused-input` and `link help` classes
 - moves accessibility scss module to the end of the scss list to allow for overrides that deal with accessibility

**Resolves**
 - [OKTA-210992](https://oktainc.atlassian.net/browse/OKTA-210992)

**Photos**

Widget Focus

### DIFF
--- a/assets/sass/modules/_accessibility.scss
+++ b/assets/sass/modules/_accessibility.scss
@@ -8,3 +8,13 @@
   height: 0;
   overflow: hidden;
 }
+
+.focused-input,
+.link.help:focus {
+  box-shadow: 0 0 8px rgba(81, 203, 238, 1);
+}
+
+.o-form-button-bar.focused-input {
+  margin-bottom: 25px;
+  padding-bottom: 0px;
+}
--- a/assets/sass/okta-sign-in.scss
+++ b/assets/sass/okta-sign-in.scss
@@ -50,7 +50,6 @@
   @import 'modules/factors-dropdown';
   @import 'modules/footer';
   @import 'modules/enroll';
-  @import 'modules/accessibility';
   @import 'modules/registration';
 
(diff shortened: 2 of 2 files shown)


assets/sass/okta-sign-in.scss
```diff
@@ -73,13 +72,16 @@
     }
   }
 
+
+  @import 'modules/accessibility';
```
manueltanzi-okta: Was this moved down intentionally?

bretterer: Yes, see changes section in the PR description :)

bretterer:
@manueltanzi-okta  in theory, using the two styles that I did, it should work across all screens.  the couple that I tested, it was fine, but I am not sure how to trigger each view we have manually

alexdahl-okta:
FYI @maggielaw-okta, if you'd like to review.

assets/sass/modules/_accessibility.scss
```diff
@@ -8,3 +8,13 @@
   height: 0;
   overflow: hidden;
 }
+
+.focused-input,
```
edburyenegren-okta: Any reason not to use `:focus` on the regular input class instead of applying this via JS?

bretterer: This was already set up to use `.focused-input`.

I originally started with regular input classes with the `:focus` command, but was very un-reliable when it came to the button.

The other part of this, as the screenshots showed, the checkbox remember me has the full width box, and `:focus` on that element does not work.

maggielaw-okta:
Looks good to me, @bretterer !

PR #713, opened 2019-05-28
feat: Increase contrast on primary button

aarongranick-okta:
### Description

The primary button color is now a brighter blue color. This increases contrast and follows latest design standards.

JIRA: https://oktainc.atlassian.net/browse/OKTA-210957

### Screenshots
Screen Shot 2019-05-28 at 3 41 56 PM

### Reviewers
@manueltanzi-okta @maggielaw-okta

### DIFF
--- a/assets/sass/_variables.scss
+++ b/assets/sass/_variables.scss
@@ -33,7 +33,7 @@ $error-text-color: #e34843;
 $icon-button-bg-color: #fff;
 $button-bg-color: #fbfbfb;
 $dark-button-bg-color: #929292;
-$primary-button-bg-color: #007dc1;
+$primary-button-bg-color: #1662dd; // Action base
 $success-button-bg-color: #4cbf9c;
 $negative-button-bg-color: #c35151;
 $disabled-button-text-color: #aaa;


manueltanzi-okta:
@aarongranick-okta Code looks good! Please add a screenshot and a UX designer for review.
Also, remember that this change requires to change the _variables.scss file also in okta-core: https://github.com/okta/okta-core/blob/177facc1e166fca810f0a72878c8ffb557afdcd5/clients/loginpage/sass/_variables.scss

alexdahl-okta:
@maggielaw-okta is probably the best reviewer on the UX side.

manueltanzi-okta:
@maggielaw-okta @alexdahl-okta Will this need a change also on the color when the button is hover?

alexdahl-okta:
I'll defer to Maggie, but I would assume that the hover state also needs to pass WCAG AA.

Attempting to compute the automatic hover background color... if the automatic hover background becomes a gradient from `#1a69e8` to `#1662dd`, `#1a69e8` passes WCAG AA and should be okay.

An alternative would be to switch to Nim's action dark `#124a94` on hover.

I have no preference; I'll defer to Maggie.

edburyenegren-okta:
@alexdahl-okta Correct! `:hover` states bear no special consideration, so they need to meet the same contrast guidelines as default states. If we're swapping the button to a Nim color, we may as well use the Nim `:hover`, but anything with a AA contrast ratio will be compliant.

PR #718, opened 2019-06-03
feat: Removes tooltip in favor of explains

manueltanzi-okta:
### Description

- Removes tooltip in favor of explains
- Don't show explains when they are equal as label

### Video

explain-top-widget

### Reviewers

@haishengwu-okta @robertjd

### DIFF
--- a/src/views/primary-auth/PrimaryAuthForm.js
+++ b/src/views/primary-auth/PrimaryAuthForm.js
@@ -14,8 +14,9 @@ define([
   'okta',
   'views/shared/TextBox',
   'util/DeviceFingerprint',
-  'util/TypingUtil'
-], function (Okta, TextBox, DeviceFingerprint, TypingUtil) {
+  'util/TypingUtil',
+  'util/Util'
+], function (Okta, TextBox, DeviceFingerprint, TypingUtil, Util) {
 
   var _ = Okta._;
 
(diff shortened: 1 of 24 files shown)


manueltanzi-okta:
@nbarbettini Also take a look! As we talked, we don't show explains unless they are different from labels.

nbarbettini:
Need a review from:
* @maggielaw-okta 
* UI Guild

nbarbettini:
Looks good to the PM. 😄

src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -126,19 +125,21 @@ define([
...
-          innerTooltip: {
-            title: Okta.loc('primaryauth.password.placeholder', 'login'),
-            text: Okta.loc('primaryauth.password.tooltip', 'login')
-          }
-        }
+        type: 'password'
       };
       if (this.settings.get('features.showPasswordToggleOnSignInPage')) {
-        passwordFieldObject.params.iconDivider = true;
+        passwordFieldObject.params = {};
```
haishengwu-okta: why this change?

alexdahl-okta:
Looks great to me! I will defer to Maggie on anything.

There's no change to how a developer configures things here, correct? They just provide the same i18n strings?

(String keys like `primaryauth.username.tooltip` are now inaccurate, but I assume renaming them is a breaking change we'd like to avoid. So, no complaint from me.)

maggielaw-okta:
👍from me! Looks really good. 

Out of curiosity, will this change result in the SIW playing nice with the form fill icon button that LastPass inserts? To illustrate what I mean, I've attached a video of me clicking until I eventually find the tiny segment of hotspot that isn't deadened by the tooltip.

siwlastpass

(Cc: @edburyenegren-okta for good measure.)

Issue #544, opened 2018-09-20
Prevent Widget from stealing focus on load


nbarbettini:
Submitting this on behalf of @jpf:

### I'm submitting a
  - feature request

### Background info

> As a developer, I want to be able to prevent the Okta Sign-In Widget from stealing focus

This hack is currently required to prevent the Widget from stealing focus from another element on the page:

```js
function preventOktaSignInWidgetFromStealingFocus() {
  // Note: This requires that the "signInWidget" object has been defined already
  signInWidget.on("pageRendered", function (data) {
    setTimeout(function(){ currentlyDisplayedEditorObj.focus(); }, returnFocusToEditorDelay);
  });
}
```

aarongranick-okta:
@jpf Has this issue been resolved?

PR #720, opened 2019-06-06
feat: Adds feature for disabling auto focus for widget input fields

nickyhof:
Resolves: #544


nickyhof:
Including more context to this PR. The main reason we added this feature is because our okta signin widget is included at the bottom of our home page. When the page renders, the focus is applied to the signin form and therefore the view scrolls to the bottom which we don't want to happen.

Is there anything else you'd like me to provide to get this PR merged? Thanks!

aarongranick-okta:
@nickyhof Was this PR closed because the Signin Widget now supports your requested feature? Is there still an issue with setting focus?

deppensteiner:
features.autoFocus still doesn't work,

PR #729, opened 2019-06-19
feat: Supports cardinality with a webauthn default profile

gowthamidommety-okta:
Adds support for cardinality in the default webauthn profile
Adds tests for multiple webauthn with both the old and new api

Resolves: OKTA-230236


test/unit/spec/EnrollChoices_spec.js
```diff
@@ -689,232 +716,319 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, FactorUtil, R
...
-                  optionalFactorRow = factorRows[1];
-              //enrolled factor row should have success check and cardinality text
-              expect(test.form.factorHasSuccessCheck(null, enrolledFactorRow)).toBe(true);
-              expect(test.form.factorCardinalityText(null, enrolledFactorRow)).toBe('(2 set up)');
-              //optional factor should have button and no cardinality text
-              expect(test.form.factorButtonText(null, optionalFactorRow)).toBe('Set up another');
-              expect(test.form.factorCardinalityText(null, optionalFactorRow)).toBe('');
-            });
+          return setupMultipleFactorEnrollments({
+            factorName: factorName,
```
magizh-okta: we can use shorter notation ES6 convention here now that SIW supports it anyway,. 

factorName,
useProfiles, 
etc

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -61,7 +64,7 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, FactorUtil, R
       index:      8,
       factorType: 'web'
     },
-    'WINDOWS_HELLO': {
+    'WEBAUTHN': {
```
magizh-okta: we are not testing for windows_hello anymore ?

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -78,24 +81,34 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, FactorUtil, R
       factorType: 'assertion:oidc'
     }
   };
+  var enrolledFactor = {
```
magizh-okta: use const so we dont accidentally override in our large test files.

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -186,24 +199,38 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, FactorUtil, R
...
+      factor.status = options.status;
+      factor.enrollment = options.enrollment;
+      if (options.useProfiles) {
+        var cardinality = {
+          min: options.policyEnrollment.minimum,
+          max: options.policyEnrollment.maximum
+        };
+        factor.profiles[0]._embedded.features[0].cardinality = cardinality;
+        for (var i = 0; i < options.policyEnrollment.enrolled; i++) {
+          factor.profiles[0]._embedded.enrolledFactors.push(enrolledFactor);
```
magizh-okta: Seems like we want to reuse this function setupMultipleFactorEnrollments for other factors as well but we are hardcoding the enrolledFactor ? May be pass that as a param as well.

src/models/Factor.js
```diff
@@ -165,14 +166,39 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
...
-        deps: ['policy'],
-        fn: function (policy) {
-          if (!policy || !policy.enrollment) {
+      cardinality : {
+        deps: ['policy', 'profiles'],
+        fn: function (policy, profiles) {
+          if (profiles && profiles.length > 0) {
+            //there will only be one profile until we support multiple.
+            var profile = profiles[0];
+            var enrolled = profile._embedded.enrolledFactors.length;
```
mauriciocastillosilva-okta: is it guaranteed on the API that we will have `_embedded.enrolledFactors`? Otherwise we can check that here and return false if that is missing

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -689,232 +716,319 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, FactorUtil, R
...
+            status: 'ACTIVE',
+            enrollment: 'OPTIONAL',
+            policyEnrollment: {'enrolled': 1, 'minimum': 0, 'maximum': 3},
+            webauthnEnabled: webauthnEnabled,
+            useProfiles: useProfiles
+          }).then(function (test) {
             var factorRows = test.form.optionalFactorList();
             var factorIndex = factorRows.find(`[data-se="${factorName}"]`).index();
-            var expectedIndex = FactorUtil.getFactorSortOrder(factorEnrollList[factorName].provider, factorEnrollList[factorName].factorType) - 1;
+            var expectedIndex = FactorUtil.getFactorSortOrder.call(test.router, factorEnrollList[factorName].provider, factorEnrollList[factorName].factorType) - 1;
```
mauriciocastillosilva-okta: don't see why this changed was needed. It's good, just curious what I missed

test/unit/spec/EnrollChoices_spec.js
gowthamidommety-okta: This object is only used for factor cardinality tests. Not all factors are tested in this object - only u2f, okta verify, and now webauthn

test/unit/spec/EnrollChoices_spec.js
gowthamidommety-okta: 👍

test/unit/spec/EnrollChoices_spec.js
gowthamidommety-okta: 👍

test/unit/spec/EnrollChoices_spec.js
gowthamidommety-okta: good point 👍will change

test/unit/spec/EnrollChoices_spec.js
gowthamidommety-okta: It fails only for webauthn, because it calls `FactorUtil.getFactorNameForFactorType` and that references the `settings` object to get `features.webauthn`. So i had to set the settings object in router (in set up above) and pass it as `this` when i call this function.

src/models/Factor.js
gowthamidommety-okta: That is a good question.. @AjayThomas-okta does the API guarantee we will have `_embedded.enrolledFactors`?

AjayThomas-okta: only for factorTypes that support profiles will have it. 
others wont.
So, to begin with, only webauthn would and others like SMS, question wouldn't.
But if there's a profiles attribute for the factorType, then yes, it will have enrolledFactors always.

AjayThomas-okta: Regarding your comment: "there will only be one profile until we support multiple."
For the record, we support multiple profiles in the framework in the backend.
But, webauthn factorType creates only one profile on enabling FF.
So, unless admin uses CRUD API to create more profiles, authn response would just return one profile since that's all that's to be fetched.

cc @antonioruberto-okta @jasleennuno-okta

gowthamidommety-okta: @nandagopalseshagiri-okta can you please confirm that for this round of changes we are always assuming that webauthn has only one profile?

nandagopalseshagiri-okta: Yes, we are assuming that there is only one webauthn profile and it should be the default profile.

test/unit/spec/EnrollChoices_spec.js
magizh-okta: wouldnt it be good in here to check windows hello too ? 

https://github.com/okta/okta-signin-widget/blob/662a286b906e98a51152ee5d7bef8db2b5570ea4/test/unit/spec/EnrollChoices_spec.js#L430

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -186,24 +199,38 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, FactorUtil, R
...
-      return setup(res);
+    function setupMultipleFactorEnrollments (options) {
+      var res = options.singleFactorRes ? deepClone(options.singleFactorRes) : 
+        options.useProfiles ? deepClone(resAllFactorsProfile) : deepClone(resAllFactors);
+      var index = options.singleFactorRes ? 0 : factorEnrollList[options.factorName].index;
+      var factor = res.response._embedded.factors[index];
+      factor.status = options.status;
+      factor.enrollment = options.enrollment;
+      if (options.useProfiles) {
+        var cardinality = {
```
mauriciocastillosilva-okta: can we move this logic to a function? Something like
```
factor.profiles[0] = createProfile(options.cardinality);
```

or
```
setProfileData(factor.profiles[0], options.cardinality);
```

I'm guessing in the future we will need to add more things to test the profiles, so it might be a good idea to have that separate

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -186,24 +199,38 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, FactorUtil, R
...
+          min: options.policyEnrollment.minimum,
+          max: options.policyEnrollment.maximum
+        };
+        factor.profiles[0]._embedded.features[0].cardinality = cardinality;
+        for (var i = 0; i < options.policyEnrollment.enrolled; i++) {
+          factor.profiles[0]._embedded.enrolledFactors.push(options.enrolledFactor);
+        }
+      }
+      else {
+        factor.policy.enrollment = options.policyEnrollment;
```
mauriciocastillosilva-okta: similar to how you changed it in the src code, we can change the name from `policyEnrollment` to `cardinality`

test/unit/spec/EnrollChoices_spec.js
gowthamidommety-okta: I actually think these tests are not factor specific, so that's why some may be left out. But added WINDOWS_HELLO for good measure.

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -432,6 +432,7 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, FactorUtil, R
             expect(test.form.isFactorMinimized('SYMANTEC_VIP')).toBe(false);
             expect(test.form.isFactorMinimized('RSA_SECURID')).toBe(false);
             expect(test.form.isFactorMinimized('DUO')).toBe(false);
+            expect(test.form.isFactorMinimized('WINDOWS_HELLO')).toBe(false);
```
magizh-okta: I guess this might not work because we removed windows hello from the factorlist

gowthamidommety-okta: No, we didn't remove WINDOWS_HELLO from anywhere.

gowthamidommety-okta:
merged as part of https://github.com/okta/okta-signin-widget/pull/800

PR #733, opened 2019-06-20
feat: TOTP verify changes

magizh-okta:
* Introduces a new factor 'token:software:hotp' ( External will still be called totp).
* Verify reuses the TOTP form.

Custom_HOTP_Verify_Flow

### DIFF
--- a/src/models/Factor.js
+++ b/src/models/Factor.js
@@ -39,6 +39,7 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
           'email',
           'token',
           'token:software:totp',
+          'token:hotp',
           'token:hardware',
           'question',
           'push',
@@ -59,6 +60,7 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
           'GOOGLE',
           'YUBICO',
           'FIDO',
+          'CUSTOM',
           'GENERIC_SAML',
           'GENERIC_OIDC'
         ]
@@ -105,7 +107,7 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
       factorLabel: {
         deps: ['provider', 'factorType', 'vendorName'],
         fn: function (provider, factorType, vendorName) {
-          if (_.contains(['DEL_OATH', 'GENERIC_SAML', 'GENERIC_OIDC'], provider)) {
+          if (_.contains(['DEL_OATH', 'GENERIC_SAML', 'GENERIC_OIDC', 'CUSTOM'], provider)) {
             return vendorName;
           }
           return factorUtil.getFactorLabel.apply(this, [provider, factorType]);
--- a/src/util/FactorUtil.js
+++ b/src/util/FactorUtil.js
@@ -64,71 +64,77 @@ function (Okta, TimeUtil) {
       iconClassName: 'mfa-google-auth',
       sortOrder: 5
     },
+    'CUSTOM_HOTP': {
+      label: '',
+      description: '',
+      iconClassName: 'mfa-custom-factor',
+      sortOrder: 6
+    },
     'SMS': {
       label: 'factor.sms',
       description: Okta.loc('factor.sms.description', 'login'),
       iconClassName: 'mfa-okta-sms',
-      sortOrder: 6
+      sortOrder: 7
     },
     'CALL': {
       label: 'factor.call',
       description: Okta.loc('factor.call.description', 'login'),
       iconClassName: 'mfa-okta-call',
-      sortOrder: 7
+      sortOrder: 8
     },
     'EMAIL': {
       label: 'factor.email',
       description: '',
       iconClassName: 'mfa-okta-email',
-      sortOrder: 8
+      sortOrder: 9
     },
     'QUESTION': {
       label: 'factor.securityQuestion',
       description: Okta.loc('factor.securityQuestion.description', 'login'),
       iconClassName: 'mfa-okta-security-question',
-      sortOrder: 9
+      sortOrder: 10
     },
     'DUO': {
       label: 'factor.duo',
       description: Okta.loc('factor.duo.description', 'login'),
       iconClassName: 'mfa-duo',
-      sortOrder: 10
+      sortOrder: 11
     },
     'SYMANTEC_VIP': {
       label: 'factor.totpHard.symantecVip',
       description: Okta.loc('factor.totpHard.description', 'login'),
       iconClassName: 'mfa-symantec',
-      sortOrder: 11
+      sortOrder: 12
     },
     'RSA_SECURID': {
       label: 'factor.totpHard.rsaSecurId',
       description: Okta.loc('factor.totpHard.description', 'login'),
       iconClassName: 'mfa-rsa',
-      sortOrder: 12
+      sortOrder: 13
     },
     'ON_PREM': {
       label: '',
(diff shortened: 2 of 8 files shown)


src/models/Factor.js
```diff
@@ -58,6 +59,7 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
           'GOOGLE',
           'YUBICO',
           'FIDO',
+          'CUSTOM',
```
gowthamidommety-okta: Do we need provider? Since we are moving away from requiring provider?

src/util/FactorUtil.js
```diff
@@ -129,6 +129,12 @@ function (Okta, TimeUtil) {
       description: Okta.loc('factor.customFactor.description', 'login'),
       iconClassName: 'mfa-custom-factor',
       sortOrder: 16
+    },
+    'CUSTOM_HOTP': {
+      label: '',
```
gowthamidommety-okta: Is this empty because enroll is through api?

src/util/FactorUtil.js
```diff
@@ -139,6 +145,9 @@ function (Okta, TimeUtil) {
     if (provider === 'OKTA' && factorType === 'push') {
       return 'OKTA_VERIFY_PUSH';
     }
+    if (provider === 'CUSTOM' && factorType === 'token:software:hotp') {
```
gowthamidommety-okta: Same question as above.. Can we leave out provider check?

src/views/mfa-verify/dropdown/FactorsDropDownOptions.js
```diff
@@ -92,6 +92,17 @@ define(['okta', 'util/RouterUtil'], function (Okta, RouterUtil) {
       }
     },
 
+    'CUSTOM_HOTP': {
+      icon: 'factor-icon',
```
gowthamidommety-okta: Is there a specific icon for this yet?

test/unit/spec/MfaVerify_spec.js
```diff
@@ -3954,6 +3957,59 @@ function (Okta,
...
+        });
+      });
+
+      itp('shows the right subtitle', function () {
+        return setupFn().then(function (test) {
+          expectSubtitleToBe(test, 'Enter your Entrust passcode');
+        });
+      });
+
+      itp('calls authClient verifyFactor with correct args when submitted', function () {
```
gowthamidommety-okta: Can we add a test for when rememberDevice=true

src/util/FactorUtil.js
magizh-okta: I might update it if required during enroll.

src/models/Factor.js
magizh-okta: We would need it just so that there is a known set of values.

src/util/FactorUtil.js
magizh-okta: yeah can omit this.

src/views/mfa-verify/dropdown/FactorsDropDownOptions.js
magizh-okta: Added the same generic icon for now.

src/util/FactorUtil.js
```diff
@@ -145,7 +145,7 @@ function (Okta, TimeUtil) {
     if (provider === 'OKTA' && factorType === 'push') {
       return 'OKTA_VERIFY_PUSH';
     }
-    if (provider === 'CUSTOM' && factorType === 'token:software:hotp') {
+    if (factorType === 'token:software:hotp') {
```
gowthamidommety-okta: I think we could put this in `getFactorNameForFactorType`. Webauthn is there also

src/util/FactorUtil.js
```diff
@@ -139,6 +145,9 @@ function (Okta, TimeUtil) {
     if (provider === 'OKTA' && factorType === 'push') {
       return 'OKTA_VERIFY_PUSH';
     }
+    if (factorType === 'token:software:hotp') {
```
gowthamidommety-okta: I think we could put this in getFactorNameForFactorType. Webauthn is there also

src/util/FactorUtil.js
```diff
@@ -129,6 +129,12 @@ function (Okta, TimeUtil) {
       description: Okta.loc('factor.customFactor.description', 'login'),
       iconClassName: 'mfa-custom-factor',
       sortOrder: 16
+    },
+    'CUSTOM_HOTP': {
+      label: '',
+      description: '',
+      iconClassName: 'mfa-custom-factor',
+      sortOrder: 17
```
mauriciocastillosilva-okta: this should be a higher sort order I assume, since this is similar to other OTP methods we have. Should probably be right after Google Auth or if not after RSA, but don't think this should be at the bottom of the list since we know the strength of this factor

src/util/FactorUtil.js
mauriciocastillosilva-okta: yeah, I was going to mention the same thing. This function goes through all the provider/factorType combinations, and `getFactorNameForFactorType` looks only at factorTypes, which is what we have here

src/util/FactorUtil.js
```diff
@@ -64,71 +64,77 @@ function (Okta, TimeUtil) {
       iconClassName: 'mfa-google-auth',
       sortOrder: 5
     },
+    'CUSTOM_HOTP': {
+      label: '',
+      description: '',
+      iconClassName: 'mfa-custom-factor',
+      sortOrder: 6
```
mauriciocastillosilva-okta: just double check with UX about the order 
cc @vincentvoong-okta (take a look at the test below to see the order of all factors)

test/unit/spec/MfaVerify_spec.js
```diff
@@ -4015,6 +4018,80 @@ function (Okta,
...
+    Expect.describe('Custom HOTP Factor', function () {
+      testHOTPFactor(setupHOTP, 'testStateToken');
+    });
+
+    Expect.describe('Custom HOTP Factor in IDX', function () {
+      testHOTPFactor(setupHOTPWithIdx, '01bfpkAkRyqUZQAe3IzERUqZGOfvYhX83QYCQIDnKZ');
+    });
+
+    function testHOTPFactor (setupFn, expectedStateToken) {
+      itp('is totp', function () {
```
mauriciocastillosilva-okta: nit: `is TotpForm` (just so it doesnt read `Custom HOTP Factor is totp`, since that would be a little confusing...

src/models/Factor.js
```diff
@@ -105,7 +107,7 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
       factorLabel: {
         deps: ['provider', 'factorType', 'vendorName'],
         fn: function (provider, factorType, vendorName) {
-          if (_.contains(['DEL_OATH', 'GENERIC_SAML', 'GENERIC_OIDC'], provider)) {
+          if (_.contains(['DEL_OATH', 'GENERIC_SAML', 'GENERIC_OIDC', 'CUSTOM'], provider)) {
```
gowthamidommety-okta: Will this be changed as part of this PR?

magizh-okta: wil take care in enroll PR. Not sure if I would have to modify this.

magizh-okta: wil take care in enroll PR. Not sure if I would have to modify this.

PR #738, opened 2019-06-27
feat: support custom back link in MFA flows

mauriciocastillosilva-okta:
- new features.customBackLinkInMFA setting
- in MFA views, we can show a custom back link

Resolves: OKTA-229629

backLink-MFA

backLink-MFA-U2F

### DIFF
--- a/src/EnrollChoicesController.js
+++ b/src/EnrollChoicesController.js
@@ -18,11 +18,10 @@ define([
   './util/RouterUtil',
   'views/enroll-choices/FactorList',
   'views/enroll-choices/RequiredFactorList',
-  'views/enroll-choices/Footer',
-  'views/shared/FooterWithBackLink'
+  'views/enroll-choices/Footer'
 ],
 function (Okta, FormController, Enums, RouterUtil, FactorList,
-  RequiredFactorList, Footer, FooterWithBackLink) {
+  RequiredFactorList, Footer) {
 
   var _ = Okta._;
 
@@ -203,12 +202,13 @@ function (Okta, FormController, Enums, RouterUtil, FactorList,
 
       this.state.set('pageType', pageType);
 
-      if (this.options.appState.get('skipLink') && pageType === Enums.HAS_REQUIRED_SOME_REQUIRED_ENROLLED) {
-        this.add(new Footer(this.toJSON()));
-      } else if (this.settings.get('features.showCustomizableBackLinkInMFA')) {
-        this.add(new FooterWithBackLink(_.extend({}, this.toJSON(), {
-          backLabel: this.settings.get('customizableBackLinkInMFA.label'),
-          backFn: this.settings.get('customizableBackLinkInMFA.fn')
+      var showSkipLink = this.options.appState.get('skipLink')
+        && pageType === Enums.HAS_REQUIRED_SOME_REQUIRED_ENROLLED;
+      var showBackLink = this.settings.get('features.showCustomizableBackLinkInMFA');
+      if (showSkipLink || showBackLink) {
+        this.add(new Footer(_.extend({}, this.toJSON(), {
+          showSkipLink,
+          showBackLink
         })));
       }
     }
--- a/src/EnrollChoicesController.js
+++ b/src/EnrollChoicesController.js
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-/* eslint complexity: [2, 13] */
+/* eslint complexity: [2, 14] */
 define([
   'okta',
   './util/FormController',
@@ -18,10 +18,11 @@ define([
   './util/RouterUtil',
   'views/enroll-choices/FactorList',
   'views/enroll-choices/RequiredFactorList',
-  'views/enroll-choices/Footer'
+  'views/enroll-choices/Footer',
+  'views/shared/FooterWithBackLink'
 ],
 function (Okta, FormController, Enums, RouterUtil, FactorList,
-  RequiredFactorList, Footer) {
+  RequiredFactorList, Footer, FooterWithBackLink) {
 
   var _ = Okta._;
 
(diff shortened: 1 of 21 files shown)


src/EnrollChoicesController.js
```diff
@@ -204,6 +205,11 @@ function (Okta, FormController, Enums, RouterUtil, FactorList,
 
       if (this.options.appState.get('skipLink') && pageType === Enums.HAS_REQUIRED_SOME_REQUIRED_ENROLLED) {
         this.add(new Footer(this.toJSON()));
+      } else if (this.settings.get('features.customBackLinkInMFA')) {
+        this.add(new FooterWithBackLink(_.extend({}, this.toJSON(), {
```
gowthamidommety-okta: Will this not work when we are in grace period? What happens when in grace period and the user goes to enroll through end user settings?

src/EnrollChoicesController.js
```diff
@@ -10,18 +10,19 @@
...
   './util/RouterUtil',
   'views/enroll-choices/FactorList',
   'views/enroll-choices/RequiredFactorList',
-  'views/enroll-choices/Footer'
+  'views/enroll-choices/Footer',
+  'views/shared/FooterWithBackLink'
 ],
 function (Okta, FormController, Enums, RouterUtil, FactorList,
-  RequiredFactorList, Footer) {
+  RequiredFactorList, Footer, FooterWithBackLink) {
```
gowthamidommety-okta: If Footer also implements customBackLinkInMFA, why do we need FooterWithBackLink?

src/EnrollChoicesController.js
mauriciocastillosilva-okta: good point, this won't work with grace period... I would need to show both links at the same time... However, this new back link will be used to enroll a factor coming from factor settings, so not sure we want to show grace period in this case, cause that will mean that the user is coming from end user settings where "Set up" was clicked, and lands in a view where "skip" shows up.

Probably what makes more sense is to actually remove skip if we have a custom back link

src/EnrollChoicesController.js
mauriciocastillosilva-okta: `enroll-factors/Footer` supports custom back link, but in here we use `enroll-choices/Footer, which only supports a skip link.

src/EnrollChoicesController.js
mauriciocastillosilva-okta: Thought about this a little more. If the response comes with a skip link, we should show it. And if the widget config has a back link, we should show it too. This is most likely a scenario that won't happen, but better to support it. So I will make changes to show both if that situation happens.

mauriciocastillosilva-okta: backLink-MFA_withSkip

gowthamidommety-okta: That looks nice!

gowthamidommety-okta: Where would Skip set up go?

mauriciocastillosilva-okta: it will go to wherever the skip request takes the user to. That is outside of the control of the "Back link" that I am adding, since skip is a link coming back in the authn response

mauriciocastillosilva-okta:
seems to me that those features are independent of the customizable back link I am adding
`features.hideSignOutLinkInMFA` hides the sign out link in mfa verify flows, but that is a different thing that adding a custom back link. Also the custom back link right now is in enroll mfa flows. Once we add it to verify we can think if we need to do something about `hideSignOutLinkInMFA`, so we don't end up with 2 back links. We'll cross that line later.
`features.hideBackToSignInForReset`: this is in reset flows only, I am not changing anything there

src/EnrollChoicesController.js
```diff
@@ -202,8 +202,14 @@ function (Okta, FormController, Enums, RouterUtil, FactorList,
 
       this.state.set('pageType', pageType);
 
-      if (this.options.appState.get('skipLink') && pageType === Enums.HAS_REQUIRED_SOME_REQUIRED_ENROLLED) {
-        this.add(new Footer(this.toJSON()));
+      var showSkipLink = this.options.appState.get('skipLink')
+        && pageType === Enums.HAS_REQUIRED_SOME_REQUIRED_ENROLLED;
+      var showBackLink = this.settings.get('features.showCustomizableBackLinkInMFA');
```
magizh-okta: nit: lets start using const :)

test/unit/spec/EnrollCall_spec.js
```diff
@@ -147,6 +161,15 @@ function (Q, Okta, OktaAuth, LoginUtil, Util, Form, Beacon, Expect, $sandbox,
             Util.stopRouter();
           });
       });
+      itp('shows custom back link if features.showCustomizableBackLinkInMFA is true', function () {
```
magizh-okta: I guess have test for every factor is a nice to have. In any case we need to rebase and add/fix test for hotp factor.

src/models/Settings.js
```diff
@@ -80,6 +80,7 @@ function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
       'features.deviceFingerprinting': ['boolean', false, false],
       'features.hideSignOutLinkInMFA' : ['boolean', false, false],
       'features.hideBackToSignInForReset' : ['boolean', false, false],
+      'features.showCustomizableBackLinkInMFA' : ['boolean', false, false],
```
magizh-okta: I guess unlike other cases we can get away with using a flag and use use the object ? or you want to keep the flag for convention sake ?

magizh-okta:
Looks good over all. Lets wait for vince to get back. In general if we dont see any issue lets starting using Es6/7 goodness where ever we can.

test/unit/spec/EnrollCall_spec.js
mauriciocastillosilva-okta: I added a test for every factor... all the Enroll*_spec have a test. I will rebase and add it to HOTP as well

src/models/Settings.js
mauriciocastillosilva-okta: you mean not using `features.*` config? I prefer to keep it since that is a very simple/explicit way to turn on/off functionality in the widget

src/EnrollChoicesController.js
mauriciocastillosilva-okta: I agree we should, but I kind of like to be consistent as well, so I prefer to keep this file with the style it has today. For new files I will definitely do that

test/unit/spec/EnrollCall_spec.js
mauriciocastillosilva-okta: rebased and changed HOTP test as well.

mauriciocastillosilva-okta:
decided to not get this change in and deal with a "back to settings" outside of the siw codebase

PR #748, opened 2019-07-03
SIW support for claims provider factor

srinivasanagandla-okta:
Added support for enroll, verify of claims_provider factor.

Resolves: OKTA-236730

Please review @gowthamidommety-okta @mauriciocastillosilva-okta 

Demo video -> https://okta.box.com/s/0gd78bnikizkkop17v4je8nnvxatb4l2

@AjayThomas-okta @antonioruberto-okta FYI...

### DIFF
--- a/src/LoginRouter.js
+++ b/src/LoginRouter.js
@@ -123,6 +123,7 @@ function (BaseLoginRouter,
       'signin/verify/u2f': 'verifyU2F',
       'signin/verify/generic_saml/assertion:saml2': 'verifySAMLFactor',
       'signin/verify/generic_oidc/assertion:oidc': 'verifyOIDCFactor',
+      'signin/verify/custom/claims_provider': 'verifyClaimsFactor',
       'signin/verify/:factorType': 'verifyNoProvider',
       'signin/verify/:provider/:factorType(/:factorIndex)': 'verify',
       'signin/enroll': 'enrollChoices',
@@ -140,6 +141,7 @@ function (BaseLoginRouter,
       'signin/enroll/fido/u2f': 'enrollU2F',
       'signin/enroll/generic_saml/assertion:saml2': 'enrollSAMLFactor',
       'signin/enroll/generic_oidc/assertion:oidc': 'enrollOIDCFactor',
+      'signin/enroll/custom/claims_provider': 'enrollClaimsFactor',
       'signin/enroll/custom/token:hotp': 'enrollHotpFactor',
       'signin/enroll/:provider/:factorType': 'enrollTotpFactor',
       'signin/enroll-activate/okta/push': 'scanBarcodePushFactor',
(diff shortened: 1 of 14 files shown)


src/util/FactorUtil.js
```diff
@@ -135,6 +135,12 @@ function (Okta, TimeUtil) {
       description: Okta.loc('factor.customFactor.description', 'login'),
       iconClassName: 'mfa-custom-factor',
       sortOrder: 17
+    },
+    'CUSTOM_CLAIMS': {
+      label: '',
+      description: Okta.loc('factor.customFactor.description', 'login'),
+      iconClassName: 'mfa-custom-factor',
+      sortOrder: 18
```
gowthamidommety-okta: Should the sort order be same as GENERIC_SAML since it is going to replace saml?

src/LoginRouter.js
```diff
@@ -123,6 +123,7 @@ function (BaseLoginRouter,
       'signin/verify/u2f': 'verifyU2F',
       'signin/verify/generic_saml/assertion:saml2': 'verifySAMLFactor',
       'signin/verify/generic_oidc/assertion:oidc': 'verifyOIDCFactor',
+      'signin/verify/custom/claims_provider': 'verifyClaimsFactor',
```
gowthamidommety-okta: Will verify always have the provider? Asking because we are moving towards factors with only factorType.

test/unit/helpers/xhr/MFA_ENROLL_allFactors_OnPrem.js
```diff
@@ -249,9 +268,6 @@ define({
                 ]
              }
           }
-       },
-       "profile":{
```
gowthamidommety-okta: Do we not need profile for claims_provider?

srinivasanagandla-okta: Yeah, we don't have a strong case for the profile. So the server side is not setting it for claims_provider.

src/LoginRouter.js
srinivasanagandla-okta: Good question. As we move towards factor profiles, the provider might become irrelevant. But that's the current model and we can't remove it right away.

src/util/FactorUtil.js
srinivasanagandla-okta: Good point, I can make it to have the same order. Also we are going to remove custom_saml, custom_oidc stuff once the claims_provider goes into EA.

src/LoginRouter.js
mauriciocastillosilva-okta: and we don't support multiple factors for claims yet, correct?

src/util/FactorUtil.js
mauriciocastillosilva-okta: then maybe we should have this as `16` and move `GENERIC_SAML` and `GENERIC_OIDC` one up

srinivasanagandla-okta:
@mauriciocastillosilva-okta @gowthamidommety-okta Updated based on your feedback.

src/LoginRouter.js
srinivasanagandla-okta: @mauriciocastillosilva-okta That's right.

PR #754, opened 2019-07-10
fix: Fixes self service recovery and unlock account misalignments

manueltanzi-okta:
### Description

- Fixes self service recovery and unlock account misalignments

### Screenshots

Screen Shot 2019-07-10 at 9 51 02 AM
Screen Shot 2019-07-10 at 9 50 49 AM
Screen Shot 2019-07-10 at 9 50 37 AM
Screen Shot 2019-07-10 at 9 50 25 AM

### Reviewers

@robertdamphousse-okta @mauriciocastillosilva-okta @haishengwu-okta  @nikhilvenkatraman-okta @haoyuli-okta

### DIFF
--- a/assets/sass/modules/_forgot-password.scss
+++ b/assets/sass/modules/_forgot-password.scss
@@ -35,7 +35,7 @@
     }
   }
 
-  .o-form-input {
+  .o-form-fieldset.enroll-sms-phone {
     // Need to verify what the correct dimensions should be for this
     // to match up with other forms
     width: 178px;
--- a/src/RecoveryChallengeController.js
+++ b/src/RecoveryChallengeController.js
@@ -73,7 +73,7 @@ function (Okta, FormController, FormType, Enums, FooterSignout, TextBox) {
           FormType.Button({
             title: Okta.loc('mfa.resendCode', 'login'),
             attributes: { 'data-se': 'resend-button' },
-            className: 'button sms-request-button',
+            className: 'button sms-request-button margin-top-30',
             click: function () {
               this.model.resendCode();
             },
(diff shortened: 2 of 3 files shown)


src/RecoveryChallengeController.js
```diff
@@ -73,7 +73,7 @@ function (Okta, FormController, FormType, Enums, FooterSignout, TextBox) {
           FormType.Button({
             title: Okta.loc('mfa.resendCode', 'login'),
             attributes: { 'data-se': 'resend-button' },
-            className: 'button sms-request-button',
+            className: 'button sms-request-button margin-top-30',
```
haoyuli-okta: this also fixes phone call button i believe

assets/sass/modules/_forgot-password.scss
```diff
@@ -35,7 +35,7 @@
     }
   }
 
-  .o-form-input {
```
mauriciocastillosilva-okta: we don't need this for `.o-form-input` anymore?

assets/sass/modules/_forgot-password.scss
```diff
@@ -35,7 +35,7 @@
...
+  .o-form-fieldset.enroll-sms-phone {
```
mauriciocastillosilva-okta: what about `enroll-call-phone`? we need to fix that too

assets/sass/modules/_forgot-password.scss
manueltanzi-okta: no, we need this style in the external container only which includes also the label

assets/sass/modules/_forgot-password.scss
manueltanzi-okta: They use the same element with class `enroll-sms-phone`, is a unique controller, see the screenshots.

mauriciocastillosilva-okta: ok...  And this is under `recovery-challenge` so all good  👍

PR #765, opened 2019-07-23
Integrate introspect API 

nikhilvenkatraman-okta:
- Integrated introspect API for V1 and V2 
- Widget should make only one introspect API call for both pipelines
- The test changes are mostly just making sure we call to `Util.mockIntrospectResponse` in every place before we call `router.refreshAuthState('dummy-token')`, to be inline with the code changes.

Old Pipeline: 

Screen Shot 2019-07-23 at 3 16 45 PM

New Pipeline: Note that no screens are built in the widget on the new pipeline

Screen Shot 2019-07-23 at 3 19 33 PM

@aarongranick-okta @haishengwu-okta


src/util/Util.js
```diff
@@ -193,6 +194,21 @@ define(['okta', './Logger', './Enums'], function (Okta, Logger, Enums) {
...
 
+  Util.getRouterFromResponse = function (response) {
+    //response.version = '1.0.0';
+    var Router = response.version && response.version === '1.0.0' ? require('v2/WidgetRouter'): require('LoginRouter');
+    return Router;
+  };
+
+  Util.introspectToken = function (authClient, widgetOptions) {
+    return Model.prototype.evaluate(function () {
+      return authClient.tx.evaluate({
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: We could call `authClient.tx.evaluate` directly. I chose `evaluate` since its a method that will do an `introspect` or `evaluate` the pipeline for the stateToken based on a flag.

test/unit/helpers/xhr/v2/FACTOR_REQUIRED_EMAIL.js
```diff
@@ -2,6 +2,7 @@ define({
     "status": 200,
     "responseType": "json",
     "response": {
+    "version": "1.0.0.",
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: should be `1.0.0`

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -139,49 +142,52 @@ function (Widget, Expect, Logger, $sandbox) {
         });
       });
     });
-
-    Expect.describe('events', function () {
+    // Fix tests
+    /*Expect.describe('events', function () {
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: should be fixed, updated

src/v2/util/BaseLoginRouter.js
```diff
@@ -120,44 +113,30 @@ function (Okta, BrowserFeatures, Settings,
...
+        }
+        //set flashError
+        this.appState.set('flashError', err);
+        this.defaultAuth();
       });
 
       this.listenTo(this.appState, 'change:remediationSuccess', function (appState, trans) {
-        RouterUtil.routeAfterAuthRemediationChange(this, trans.data);
+        // set authStateRefreshed
+        this.settings.set('authStateRefreshed', true);
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: For cases when we come to the router without a stateToken, we need to be able to tell if it is because we did an introspect call and then unset the stateToken or widget was bootstrapped without a stateToken. So if there was a stateToken and we hit introspect we use `authStateRefreshed ` to route accordingly. `remediationSuccess ` would also have the response, but we would have to check for that `remediationSuccess ` object being non empty. I preferred checking for a  boolean in `authStateRefreshed`.

src/v2/util/BaseLoginRouter.js
```diff
@@ -120,44 +113,30 @@ function (Okta, BrowserFeatures, Settings,
...
+        //set authResponse
+        this.appState.setAuthResponse(trans.data);
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: yeah, `authResponse ` is not used anywhere in code right now. It was added to appState, I could remove it for now and add later if needed.

package.json
```diff
@@ -57,7 +57,7 @@
     "prepublish": "yarn build:release"
   },
   "devDependencies": {
-    "@okta/okta-auth-js": "~2.6.0",
+    "@okta/okta-auth-js": "2.7.0-alpha.ge4c8a09",
```
nikhilvenkatraman-okta: This can be updated once auth js changes are done

src/widget/OktaSignIn.js
```diff
@@ -92,10 +138,32 @@ var OktaSignIn = (function () {
     };
   }
 
+  function checkResponseVersion (response) {
+    response.version = '1.0.0';
```
aarongranick-okta: If this is temporary, please add a // TODO comment

nikhilvenkatraman-okta: Oops that will be removed. For limited EA this version we publish will be `1.0.0`. But I just learned that going forward we will be following a semantic version for this API and any version >= `1.0.0` means we are on the new pipeline. The old pipeline API will also get a version.

src/v2/util/BaseLoginRouter.js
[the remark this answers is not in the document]
haishengwu-okta: > but we would have to check for that remediationSuccess object being non empty.

src/util/RouterUtil.js
```diff
@@ -100,8 +100,13 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
     // Token has expired - no longer valid. Navigate back to primary auth.
     if (err.errorCode === ErrorCodes.INVALID_TOKEN_EXCEPTION) {
       router.appState.set('flashError', err);
-      router.controller.state.set('navigateDir', Enums.DIRECTION_BACK);
-      router.navigate('', { trigger: true });
+      if (router.controller) {
```
haishengwu-okta: do we still need this change after leveraging `introspecSuccess/Error`?

src/v2/models/AppState.js
```diff
@@ -37,15 +38,9 @@ function (Okta) {
...
-    setAuthResponse: function (res) {
-      // TODO set AppState.policy based on API response
-      // TODO set AppState.factors based on API response
-      this.set('lastAuthResponse', res);
-    },
-
     derived: {
       'remediation': {
-        deps: ['lastAuthResponse'],
+        deps: ['introspectSuccess'],
```
haishengwu-okta: the `deps` shall be `remediationSuccess`?

src/v2/util/BaseLoginRouter.js
```diff
@@ -120,44 +113,22 @@ function (Okta, BrowserFeatures, Settings,
       });
 
       this.listenTo(this.appState, 'change:remediationFailure', function (appState, err) {
-        RouterUtil.routeAfterAuthStatusChangeError(this, err);
-      });
-
-      this.listenTo(this.appState, 'change:remediationSuccess', function (appState, trans) {
```
haishengwu-okta: - maybe we need `change:remediationSuccess` for the resp changes when moving through the pipeline?
- also trivial handler for `change:introspectSuccess` to copy `introspectSuccess` into `remediationSuccess`. Doesn't seem ideal but we could add comments to explain the reason.

haishengwu-okta:
looks like `v2/RefreshAuthStateController` is not deleted.

src/util/RouterUtil.js
nikhilvenkatraman-okta: removed

src/v2/util/BaseLoginRouter.js
```diff
@@ -125,6 +125,16 @@ function (Okta, BrowserFeatures, Settings,
         this.defaultAuth();
       });
 
+      this.listenTo(this.appState, 'change:remediationSuccess', function (appState, trans) {
+        //set remediationSuccess
+        this.appState.set('remediationSuccess', trans);
```
haishengwu-okta: `change:remediationSuccess` means `remediationSuccess` had been change and why you set it again?

nikhilvenkatraman-okta: I added this assuming that when we are moving forward in the pipeline and call BaseLoginModel which does `self.appState.trigger('remediationSuccess', trans)`, we will need a listener for this change ?

haishengwu-okta: `trigger('remediationSuccess'` wont go here because this for `change:remedationSuccess`.

nikhilvenkatraman-okta: Ah, right I'll remove it for now. We have this listener in the BaseLoginController that is setting remediationSuccess.
```       
this.listenTo(this.options.appState, 'remediationSuccess', function (transaction) {
        this.options.appState.set('remediationSuccess', transaction);
      }); 
```

src/util/Util.js
```diff
@@ -193,6 +194,22 @@ define(['okta', './Logger', './Enums'], function (Okta, Logger, Enums) {
     return explain;
   };
 
+  Util.introspectToken = function (authClient, widgetOptions) {
+    var deferred = Q.defer();
+    var trans = authClient.tx.evaluate({
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: leaving it here for now.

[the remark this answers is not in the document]
haishengwu-okta: hmm, why?

src/v2/WidgetRouter.js
```diff
@@ -14,56 +14,38 @@
...
-    stateLessRouteHandlers: [
-      'defaultAuth', 'renderErrorView', 'refreshAuthState'
-    ],
-
     defaultAuth: function () {
-      var stateToken = this.settings.get('stateToken');
-      if (stateToken) {
-        //if widget bootstrapped with stateToken, make an API call to get authstate
-        this.refreshAuthState(stateToken);
+      var trans = this.appState.get('introspectSuccess') || this.appState.get('remediationSuccess');
```
haishengwu-okta: probably unnecessary given `change:introspectSuccess` had been handle at baserouter.

test/unit/spec/v2/RefreshAuthStateController_spec.js
```diff
@@ -33,30 +33,32 @@ function (Q, Okta, OktaAuth, Util, Beacon, FormView, Expect,
...
-          test.setNextResponse(resFactorRequiredEmail);
-          test.router.refreshAuthState();
           return tick(test);
         })
         .then(function () {
           expect($.ajax.calls.count()).toBe(1);
           Expect.isJsonPost($.ajax.calls.argsFor(0), {
-            //TODO update to api/v2
-            url: 'https://foo.com/api/v1/authn',
+            url: 'https://foo.com/api/v1/idx/introspect',
```
haishengwu-okta: then the description isn't accurate because it's expecting api call from refresh controller?

PR #769, opened 2019-07-25
Add mock server to playground

haishengwu-okta:
1. With such `.widgetrc`, we could run authn with mock server instead of real org. Only implement one very happy pass for this initial phase.

```
{
  "widgetOptions": {
    "baseUrl": "http://localhost:3000",
    "logo": "/img/logo_widgico.png",
    "logoText": "Windico",
    "features": {
      "router": true,
      "rememberMe": true,
      "multiOptionalFactorEnroll": true
    },
    "assets": {
      "baseUrl": "/"
    }
  }
}
```

2. Upgrade some outdated none critical dev deps


package.json
```diff
@@ -59,20 +59,21 @@
...
+    "babel-plugin-istanbul": "^5.2.0",
     "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-env": "^1.6.0",
     "backbone": "1.2.1",
     "cssnano": "^4.1.7",
     "dotenv": "^8.0.0",
+    "dyson": "^2.0.5",
     "eslint": "^5.9.0",
-    "fs-extra": "^7.0.1",
+    "fs-extra": "^8.1.0",
```
nikhilvenkatraman-okta: Is it expected that all these other packages are also getting updated?

haishengwu-okta: it's just a bonus work I did and it's independent to the mock server change.

manueltanzi-okta:
@haishengwu-okta 👏 works great! Should we add instructions somewhere? Especially for `.widgetrc` settings.

PR #770, opened 2019-07-26
Surfaces days remaining during grace period

gowthamidommety-okta:
When in grace period, the back-end will return an end date in the policy. This should be used to show the number of days remaining for a end user to enroll in a required factor.

Screen Shot 2019-07-26 at 12 08 47 PM

Screen Shot 2019-07-26 at 12 09 32 PM


src/models/AppState.js
```diff
@@ -93,6 +93,12 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
     return Okta.loc('minutes', 'login', [factorLifetimeInMinutes]);
   }
 
+  function getGracePeriodRemainingDays (gracePeriodEndDate) {
+    let endDate = new Date(gracePeriodEndDate).getTime();
```
magizh-okta: nit: both can be consts we dont reuse them.

src/EnrollChoicesController.js
```diff
@@ -47,13 +47,26 @@ function (Okta, FormController, Enums, RouterUtil, FactorList,
         case Enums.ALL_OPTIONAL_SOME_ENROLLED:
         case Enums.HAS_REQUIRED_ALL_REQUIRED_ENROLLED:
           return Okta.loc('enroll.choices.optional', 'login');
+        case Enums.HAS_REQUIRED_SOME_REQUIRED_ENROLLED:
+          const remainingDays = this.options.appState.get('gracePeriodRemainingDays');
+          return (remainingDays >= 0) ? this.getGracePeriodSubtitle(remainingDays) : this.getDefaultSubtitle();
```
magizh-okta: if gracePeriodRemainingDays is set to null when it is not defined. 

null >= 0   gives => true 

We need to update the logic here.

gowthamidommety-okta: Yup, changed this. Will push soon

src/models/AppState.js
gowthamidommety-okta: Changing this to var to be consistent with the rest of the file

src/EnrollChoicesController.js
```diff
@@ -47,13 +47,27 @@ function (Okta, FormController, Enums, RouterUtil, FactorList,
         case Enums.ALL_OPTIONAL_SOME_ENROLLED:
         case Enums.HAS_REQUIRED_ALL_REQUIRED_ENROLLED:
           return Okta.loc('enroll.choices.optional', 'login');
+        case Enums.HAS_REQUIRED_SOME_REQUIRED_ENROLLED:
+          var remainingDays = this.options.appState.get('gracePeriodRemainingDays');
+          return (Number.isInteger(remainingDays) && remainingDays >= 0) ?
```
mauriciocastillosilva-okta: I think `gracePeriodRemainingDays` should return an integer all the time, with value 0 or above. At the end, `0` means required right now... Unless we need to know that the value is not actually set, which is where returning `null` will help. But I don't think we need that

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -522,6 +545,97 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, FactorUtil, R
         });
       });
 
+      Expect.describe('Grace period', function () {
+        beforeEach(function () {
+          var today = new Date('2019-06-25T00:00:00.000Z');
+          jasmine.clock().mockDate(today);
+        });
+        Expect.describe('all factors are required', function () {
```
mauriciocastillosilva-okta: nit: `all factors are required and none enrolled`

src/EnrollChoicesController.js
mauriciocastillosilva-okta: talked offline, makes sense as is. `null` means no grace period, and then any number will be handled in `getGracePeriodSubtitle`, where `0` means we still are in grace period

test/unit/spec/EnrollChoices_spec.js
```diff
@@ -522,6 +545,97 @@ function (Okta, OktaAuth, Util, EnrollChoicesForm, Beacon, Expect, FactorUtil, R
         });
       });
 
+      Expect.describe('Grace period', function () {
+        beforeEach(function () {
```
magizh-okta: should we install/uninstall  clock so we just restrict the jasmine clock behavior to this block  ?

gowthamidommety-okta: Filed OKTA-241179 to look into this

PR #773, opened 2019-07-29
Add transformation layer for converting ion response into UI oriented result.

haishengwu-okta:


test/unit/helpers/xhr/v2/FACTOR_VERIFICATION_REQUIRED_OKTA_PUSH.js
```diff
@@ -0,0 +1,119 @@
...
+        }
+      ]
+    },
+    "factor": {
+      "type": "object",
+      "value": {
+        "factorType": "push",
+        "provider": "okta",
+        "profile": {
+          "email": "omgitstom@abbott.dev"
```
nikhilvenkatraman-okta: nit: use dummy email here

src/v2/ion/transformer.js
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  const firstLevelObjects = convertObjectType(resp);
+  const currentState = createCurrentStateObject(resp, Object.keys(firstLevelObjects));
+
+  const result = Object.assign({},
+    firstLevelObjects,
+    {
+      currentState,
+      __raw: resp,
```
nikhilvenkatraman-okta: nit: can we call this `__rawResponse`?

test/unit/spec/v2/ion/transformer_spec.js
```diff
@@ -0,0 +1,207 @@
+import transform from 'v2/ion/transformer';
+import XHRFactorRequiredEmail from '../../../helpers/xhr/v2/FACTOR_REQUIRED_EMAIL';
+import XHRFactorVerificationRequiredPush from '../../../helpers/xhr/v2/FACTOR_VERIFICATION_REQUIRED_OKTA_PUSH';
+
+describe('v2/ion/transformer', function () {
+  it('returns result when invokes with invalid resp', () => {
+    expect(transform()).toBeNull();
```
nikhilvenkatraman-okta: maybe a future item might be to return some specific error when we cant parse the response.

test/unit/spec/v2/ion/transformer_spec.js
```diff
@@ -0,0 +1,207 @@
...
+
+    expect(result.currentState['submit-factor']()).toEqual({
+      method: 'POST',
+      url: 'https://your-org.okta.com/api/v2/authn/',
+      data: {
+        stateHandle: '01OCl7uyAUC4CUqHsObI9bvFiq01cRFgbnpJQ1bz82'
+      }
+    });
+
+    expect(result.currentState['submit-factor']({foo: 'bar'})).toEqual({
```
nikhilvenkatraman-okta: Would we ever call the function like this? where we are specifying  the data that needs to be posted? Wouldn't this be available in the response ?

haishengwu-okta: sure, will  keep iterate

src/v2/ion/transformer.js
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * @returns {Object} actionObject
+ * @returns {string} actionObject.name
+ * @returns {function} actionObject.createRequest
+ */
+const createActionObj = (relObj) => {
+  const data = (relObj.value || [])
+    .filter(v => v.visible === false)
+    .map(v => { return { [v.name]: v.value }; })
+    .reduce((init, x) => Object.assign({}, init, x), {});
```
manueltanzi-okta: I think you can simplify to `.reduce((init, x) => Object.assign(init, x), {});` (might also be faster)

src/v2/ion/transformer.js
```diff
@@ -0,0 +1,185 @@
...
+};
+
+/**
+ * @returns {Object.<string, function>} actions
+ */
+const createActions = (allCreateForms) => {
+  const allActions = allCreateForms.map(createActionObj);
+
+  // turns list of actions into a mapping
+  return allActions.reduce((init, action) => {
```
manueltanzi-okta: Also here, we might be able to simplify L100-105 to:
```suggestion
  return allActions.reduce((init, action) => Object.assign(init, { [action.name]: action.createRequest }), {});
```

src/v2/ion/transformer.js
```diff
@@ -0,0 +1,185 @@
...
+
+};
+
+/**
+ *
+ * @param {AuthnResponse} resp authn v2 response
+ * @returns {AuthnResult} transformed result
+ */
+const convert = (resp) => {
+  if (!isObject(resp) || !resp) {
```
manueltanzi-okta: nit: Do we need the second statement? (I think `null` is not an object for `_`)

src/v2/ion/transformer.js
```diff
@@ -0,0 +1,185 @@
...
+/**
+ *
+ * @returns {Object} actionObject
+ * @returns {string} actionObject.name
+ * @returns {function} actionObject.createRequest
+ */
+const createActionObj = (relObj) => {
+  const data = (relObj.value || [])
+    .filter(v => v.visible === false)
+    .map(v => { return { [v.name]: v.value }; })
```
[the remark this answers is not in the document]
haishengwu-okta: it probably will become `{ {[v.name]: v.value };  } which a little harder to read.

src/v2/ion/transformer.js
haishengwu-okta: I kind of prefer not to override parameters (first one) unless it becomes issue or we have to.

src/v2/ion/transformer.js
haishengwu-okta: same as last.

PR #775, opened 2019-07-29
fix: hideSignOutLinkInMFA feature in claims provider

mauriciocastillosilva-okta:
- when hideSignOutLinkInMFA is enabled, we need to hide sign out link in claims provider factor view

Resolves: OKTA-240287


mauriciocastillosilva-okta:
I didn't add it because I think those are more for the functionality of FooterSignOut, which is already been tested in those tests you pointed out (under the section for sign out link). So wanted just to test the visibility of the link in custom factor view. Let me know if this makes sense

PR #777, opened 2019-07-30
Fixes travis script to fail fast

vijetmahabaleshwar-okta:
Resolves https://oktainc.atlassian.net/browse/OKTA-229896


scripts/travis.sh
```diff
@@ -3,9 +3,11 @@
 # if this build was triggered as cron job, run e2e tests on iOS
 if [ "${TRAVIS_EVENT_TYPE}" = "cron" ] ; then
     export SAUCE_PLATFORM_NAME="iOS";
+    set -e
     yarn build:release
     yarn test:e2e
 else
+    set -e
```
haishengwu-okta: curious what's `set -e`?

vijetmahabaleshwar-okta: It’s a `set error` command in shell script, which makes the script to exit as soon as the first thing fails.
In our case, the `yarn lint` failed, but it still continued to next line and finally `yarn test:e2e` passed. Which marked the build as success.
For us, we need to exit as soon as the first command fails.

vijetmahabaleshwar-okta: https://dougrichardson.org/2018/08/03/fail-fast-bash-scripting.html

vijetmahabaleshwar-okta: Also @haishengwu-okta, speeding up the build is definitely something we can look into, but I think that should be taken up as a separate task. For now, we should get the failing tests pass and get this PR merged?

mauriciocastillosilva-okta: I thought we needed to report at the end, so we can see all the things that failed (lint, uint test, etc)

vijetmahabaleshwar-okta: @mauriciocastillosilva-okta: We can do that as well. Need to extract the exit code after each command. And then report at the end. Do you think that’s better? There are pros and cons for both approaches

PR #779, opened 2019-07-31
Add form builder

nikhilvenkatraman-okta:
- Add form builder component 
- Render identifier first view


src/v2/util/BaseLoginRouter.js
```diff
@@ -130,12 +134,32 @@ function (Okta, BrowserFeatures, Settings,
...
 
+    handleRemediationSuccess: function (trans) {
+      // transform response
+      const ionResponse = transform(trans);
+      this.appState.set('currentState', ionResponse.currentState);
+      // set formSchema
+      const formSchema = ionResponse.currentState.remediation[0].value;
+      this.appState.set('formSchema', formSchema);
+      // set setFormSchemaInputMap
+      RemediationUtil.setFormSchemaInputMap(formSchema);
```
haishengwu-okta: sounds like another derived props and `get**` function from appState instead of maintaining state in RemediationUtil

src/v2/util/SchemaData.js
```diff
@@ -0,0 +1,24 @@
...
+const getUISchema = function (formName, factorType) {
+  if (factorType) {
+    return getSchema(formName, factorType);
+  } else {
+    return getSchema(formName);
+  }
+};
+
+module.exports = {
+  getUISchema,
```
haishengwu-okta: shall this function live in `ion/uiSchema/schema.js` as well?

PR #780, opened 2019-07-31
Fix localization bug during factor enroll flow

nikhilvenkatraman-okta:
- Fixed localization bug during factor enroll flow

Before : 

Screen Shot 2019-07-18 at 9 19 22 AM

After: 

Video : https://okta.app.box.com/file/499880679151

Factor without brandName

Screen Shot 2019-07-31 at 10 40 46 AM

Factor with brandName 

Screen Shot 2019-07-31 at 10 30 16 AM

@haishengwu-okta @mauriciocastillosilva-okta @manueltanzi-okta 

Resolves: OKTA-239313

### DIFF
--- a/src/util/FactorUtil.js
+++ b/src/util/FactorUtil.js
@@ -24,13 +24,13 @@ function (Okta, TimeUtil) {
   var factorData = {
     'OKTA_VERIFY': {
       label: 'factor.totpSoft.oktaVerify',
-      description: Okta.loc('factor.totpSoft.description', 'login'),
+      description: _.partial(Okta.loc, 'factor.totpSoft.description', 'login'),
       iconClassName: 'mfa-okta-verify',
       sortOrder: 1
     },
     'OKTA_VERIFY_PUSH': {
       label: 'factor.oktaVerifyPush',
-      description: Okta.loc('factor.push.description', 'login'),
+      description: _.partial(Okta.loc, 'factor.push.description', 'login'),
       iconClassName: 'mfa-okta-verify',
       sortOrder: 1
     },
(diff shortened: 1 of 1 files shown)


src/util/FactorUtil.js
```diff
@@ -24,69 +24,69 @@ function (Okta, TimeUtil) {
   var factorData = {
     'OKTA_VERIFY': {
       label: 'factor.totpSoft.oktaVerify',
-      description: Okta.loc('factor.totpSoft.description', 'login'),
+      description: _.partial(Okta.loc, 'factor.totpSoft.description', 'login'),
```
mauriciocastillosilva-okta: might be easier to do this the same way we do `label` one line above. We have the key here and we handle this properly in the function `getFactorLabel`.

src/util/FactorUtil.js
```diff
@@ -24,69 +24,69 @@ function (Okta, TimeUtil) {
...
-      description: Okta.loc('factor.push.description', 'login'),
+      description: _.partial(Okta.loc, 'factor.push.description', 'login'),
       iconClassName: 'mfa-okta-verify',
       sortOrder: 1
     },
     'U2F': {
       label: 'factor.u2f',
       description: (brandName) => {
         return brandName ?
-          Okta.loc('factor.u2f.description.specific', 'login', [brandName]) :
```
haishengwu-okta: is this change necessary since `description` is already a function?

nikhilvenkatraman-okta: Yeah, I think for cases with brandName we don't need the _.partial since its a function.

PR #782, opened 2019-08-01
Add initial tests for formbuilder

nikhilvenkatraman-okta:
- Add tests for formbuilder and schemaData

### DIFF
--- a/src/v2/util/FormBuilder.js
+++ b/src/v2/util/FormBuilder.js
@@ -47,34 +47,35 @@ const addInputObject = function (inputObj, inputOptions, uiSchema) {
   }
 };
 
+let formObj = {
+  inputOptions: [],
+  layout: 'o-form-theme',
+  className: 'ion-form',
+  autoSave: true,
+  noCancelButton: true,
+  title: '',
+  save: '',
+  children: [],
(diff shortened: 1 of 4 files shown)


src/v2/util/FormBuilder.js
```diff
@@ -118,6 +119,11 @@ const createInputOptions = function (appState) {
...
 };
 
+const __getFormObj = function () {
+  return formObj;
+};
+
 module.exports = {
-  createInputOptions
+  createInputOptions,
+  __getFormObj
```
haishengwu-okta: we dont like to add code specifically for testing?

nikhilvenkatraman-okta: yeah, I wanted some way to test if the formBuilder is setting up the inputs correctly. Would it help if we add an explicit comment here ?

haishengwu-okta: - Im thinking you could set `form.inputs` so that verify all the input options.
- or spy the `add` and `addInput` function, then render the form and verify the spy.

nikhilvenkatraman-okta:
tests moved to https://github.com/okta/okta-signin-widget/pull/784

PR #787, opened 2019-08-02
 fix: prevent redirect when app is in background

magizh-okta:
* Introduces a flag to check if SIW host is in background before redirecting.

    This is to overcome the issue where Chrome on android prevents app url redirects when not in foreground. This is currently happening in chrome but not on firefox on android. IOS is not an issue as safari takes care of it.

    * Ideally the check could  go in redirectUtilFn in settings model and callGlobalError/Success,
    so all paths involving redirects are handled. In order to reduce the surface area we are adding the check in success state handler for now.

    **TODO:**

    * Test on ios chrome if we support it.
    * Add more tests to cover other scenarios.


manueltanzi-okta:
@magizh-okta linter is failing 😄

PR #790, opened 2019-08-05
Restore tests for authorization_code flow (non PKCE)

aarongranick-okta:
- Also adds tests for PKCE specific behavior

### DIFF
--- a/test/unit/spec/LoginRouter_spec.js
+++ b/test/unit/spec/LoginRouter_spec.js
@@ -1014,9 +1014,9 @@ function (Okta, Q, Logger, Errors, BrowserFeatures, WidgetUtil, Bundles, config,
           expect(params.get('prompt')).toBe(options.prompt);
         }
 
-        if (responseType === 'code') {
+        if (options.code_challenge_method) {
+          expect(params.get('code_challenge_method')).toBe(options.code_challenge_method);
           expect(params.get('code_challenge')).toBeTruthy();
-          expect(params.get('code_challenge_method')).toBe('S256');
         }
 
       }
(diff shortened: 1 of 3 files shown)


test/unit/spec/LoginRouter_spec.js
```diff
@@ -1046,85 +1046,90 @@ function (Okta, Q, Logger, Errors, BrowserFeatures, WidgetUtil, Bundles, config,
             Expect.deprecated('Use "scopes" instead of "scope"');
           });
       });
-      itp('redirects instead of using an iframe if the responseType is "code"', function () {
+
+      itp('PKCE: redirects, sets the responseMode to "fragment" and sets a code_challenge', function () {
```
swiftone: I like the PKCE here, but can we flag/group/describe/something the other tests to make it clear what kind of flow they are testing so someone (me :) ) in the future doesn't recreate your discovery?

aarongranick-okta: The other is the "default" flow for the SIW, I'm not sure what it is called. For some reason I convinced myself that all these tests were not related to anything actually used in production, but were part of early PKCE experiments. I don't think anyone else is going to make that mistake, it sounds pretty crazy. We should review very carefully whenever test conditions are changed.

swiftone: Sorry to be dense, but I think I'm missing something.  All of these changes are within the OIDC test block, so why did they switch to `fragment` instead of `query`?  

Is this OIDC vs AuthN, or was this PKCE vs non-PKCE within OIDC?

swiftone: ...that said, my confusion doesn't need to hold up this PR - here you're restoring the way it was before, and within the OIDC block.

aarongranick-okta: From my understanding this flow is the standard OIDC "authorization code flow". PKCE is an add-on to this underlying flow, requiring the code challenge. The reasons for the `responseMode` are mostly historic (we are maintaining defaults we have provided previously). A developer could override this in either flow.

aarongranick-okta:
Requires latest release of okta-auth-js https://github.com/okta/okta-auth-js/pull/236

PR #799, opened 2019-08-07
Add help link to footer

nikhilvenkatraman-okta:
- Adds help link to footer with option to support custom help link set via widget config

Video : https://okta.box.com/s/rhkdqc9h3y0zp5e3crkyntrp09iy3xvf

@taylorlaubach-okta @haishengwu-okta

### DIFF
--- a/src/v2/components/Link.js
+++ b/src/v2/components/Link.js
@@ -19,8 +19,11 @@ const Link = View.extend({
   tagName: 'a',
 
   attributes () {
-    const customHelpLink = this.options.settings.get('helpLinks.help');
-    let href = customHelpLink ? customHelpLink: this.options.href || '#';
+    if (this.options.href && !this.options.customLink) {
+      // if href is defined append a baseUrl
+      this.options.href = this.options.settings.get('baseUrl') + this.options.href;
+    }
+    let href = this.options.href  || '#';
     return {
       'data-se': this.options.name,
       href: href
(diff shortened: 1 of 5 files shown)


src/v2/components/Link.js
```diff
@@ -19,8 +19,11 @@ const Link = View.extend({
   tagName: 'a',
 
   attributes () {
-    const customHelpLink = this.options.settings.get('helpLinks.help');
-    let href = customHelpLink ? customHelpLink: this.options.href || '#';
+    if (this.options.href && !this.options.customLink) {
```
haishengwu-okta: looks like you could construct the `href` (prefix base url) from `FormBuilder` so that logic in `Link.js` could be simplified as
` if this.options.href then use it otherwise '#'`.

src/v2/components/Link.js
```diff
@@ -34,17 +37,13 @@ const Link = View.extend({
...
   postRender () {
-    this.$el.click((event) => {
-      let actionName = this.options.actionName;
-      if (actionName === 'redirect') {
-        const customHelpLink = this.options.settings.get('helpLinks.help');
-        let link = customHelpLink ? customHelpLink: this.options.href;
-        this.options.appState.trigger('invokeRedirectLinkAction', link);
-      } else {
+    let actionName = this.options.actionName;
+    if (actionName !== 'redirect') {
```
haishengwu-okta: maybe do `!this.options.href` instead of a new actionName?

taylorlaubach-okta:
A couple of questions:
1. Should we be somehow maintaining context of which identifier the user entered to avoid them having to enter it again?
2. Will we ever handle flows like this from inside the widget, rather than sending the user to a completely new page?

nikhilvenkatraman-okta:
@taylorlaubach-okta 
> Should we be somehow maintaining context of which identifier the user entered to avoid them having to enter it again?

This could be solved with the `remember me` feature which we are thinking about building for EA. Currently the API does not send us any user context to do this

> Will we ever handle flows like this from inside the widget, rather than sending the user to a completely new page?

We don't have a way to support custom help content inside the widget. We could think about hosting the default okta help page in the widget but it feels like, the moment we introduce help content inside the widget, customers will want a way to customize it. 
The custom help link allows customers to have a custom help page setup that they can redirect their users to.

taylorlaubach-okta:
That makes sense. Thanks Nikhil!

nikhilvenkatraman-okta:
@taylorlaubach-okta will do

PR #802, opened 2019-08-07
UI Schema creation refactoring

haishengwu-okta:
- move ui schema creation into transformation layer


src/v2/util/FormBuilder.js
```diff
@@ -77,34 +77,10 @@ const createInputComponent = function (uiSchema, inputObj) {
...
-      break;
-    }
-  });
-};
-
 const createForm = function (remediation = {}) {
   let inputOptions = [];
   const uiSchema = uiSchemaFactory.createUISchema(remediation.name);
-  augmentUISchema(remediation.value, uiSchema);
+  uiSchema.formInputs = remediation.uiSchema;
```
haishengwu-okta: this is only good enough for now since all `formInputs` come from `formSchema` atm.

PR #804, opened 2019-08-07
feat: Adds flag to remove default tip

manueltanzi-okta:
### Description

- Adds flag to hide default tip when label is customized and tip is not

Resolves: [OKTA-240654](https://oktainc.atlassian.net/browse/OKTA-240654)

### Video

tip_fix

### Reviewers

@haishengwu-okta @alexdahl-okta


alexdahl-okta:
Looks like it works! I'm trusting that:

- When there's a custom label and a custom tip with different values, both appear.
- When there's a custom label and a custom tip with the same values, the tip does not appear.

(That's existing behavior, so we're good as long as this hasn't broken them.)

manueltanzi-okta:
@alexdahl-okta That's correct!

PR #806, opened 2019-08-08
FormBuilder Refactor

haishengwu-okta:
1. Add ViewFactory to create customized view per form name and related base classes
2. Move BaseController from util to controller folder
3. Move BaseLoginRouter from util to v2 root.


src/v2/view-builder/internals/BaseForm.js
```diff
@@ -0,0 +1,37 @@
...
+import FormInputFactory from './FormInputFactory';
+
+export default Form.extend({
+
+  layout: 'o-form-theme',
+  className: 'ion-form',
+  hasSavingState: true,
+  autoSave: false,
+  noCancelButton: true,
+  title: '',
```
nikhilvenkatraman-okta: Should we provide a default title in the BaseForm ?

haishengwu-okta: Good point. Any suggestion?

PR #825, opened 2019-08-10
SIW V2 OIE - Fix form name mapping

haishengwu-okta:


src/v2/view-builder/ViewFactory.js
```diff
@@ -5,31 +5,55 @@ import EnrollFactorPasswordView from './views/EnrollFactorPasswordView';
...
-  'enroll-factor-email': EnrollFactorEmailView,
-  'enroll-factor-password': EnrollFactorPasswordView,
-  'required-factor-email': RequiredFactorEmailView,
-  'required-factor-password': RequiredFactorPasswordView,
-  'factor-poll-verification': FactorPollVerificationView,
-  'otp': OTPView,
-  'enroll-profile': EnrollProfileView,
-  'terminal': TerminalView,
+  'identify': {
+    '_': IdentifierView,
```
nikhilvenkatraman-okta: hmmm, not sure how I feel about having a `_` for views without a factor. Is it better to just list them at the top level like  
```
'identify': IdentifierView,
'enroll-factor': {
    email: EnrollFactorEmailView,
    password: EnrollFactorPasswordView
  }

```
and we just check if config in `const config = VIEWS_MAPPING[formName]; ` is an object or something?

haishengwu-okta: - we could check against `if it's subclass of BaseView` to determine whether is nested mapping or a View.
- the reason I came up with `_` is to support a use case like, e.g.
```
for `enroll-factor`, 
if the factor is email or password, use customized view
otherwise use default view `EnrollFactorView`
```

nikhilvenkatraman-okta: yeah, I dont mind checking for `subclass of BaseView`. The other option I was going to suggest was actually `default` 😄

haishengwu-okta: I created constant `DEFAULT` to represent the default key. Let me know your thoughts?

PR #828, opened 2019-08-13
Add sign out link to footer

nikhilvenkatraman-okta:
Add sign out to all views with cancel action 
 
Screen Shot 2019-08-12 at 7 42 43 PM

Screen Shot 2019-08-12 at 7 42 48 PM

Clicking on cancel will take users to terminal state for ( LEA). Add back to sign in to `Terminal` state. Note that this flow of taking users to the terminal state on sign out still needs to discussed with PM.

Screen Shot 2019-08-12 at 8 39 12 PM

Sign out link with recovery link  
Screen Shot 2019-08-13 at 9 47 48 AM

@taylorlaubach-okta

### DIFF
--- a/assets/sass/v2/_footer.scss
+++ b/assets/sass/v2/_footer.scss
@@ -1,8 +1,7 @@
 .siw-main-footer {
   .auth-footer {
     .js-cancel {
-      display: block;
-      text-align: right;
+      float: right;
     }
   }
 }
--- a/assets/sass/v2/_footer.scss
+++ b/assets/sass/v2/_footer.scss
@@ -1,0 +1,8 @@
+.siw-main-footer {
+  .auth-footer {
+    .js-cancel {
+      display: block;
+      text-align: right;
+    }
+  }
+}
--- a/src/v2/view-builder/internals/BaseFooter.js
+++ b/src/v2/view-builder/internals/BaseFooter.js
@@ -24,7 +24,25 @@ export default View.extend({
 
   initialize () {
     const links = _.resultCtx(this, 'links', this);
+    let cancelFn = this.options.appState.get('currentState').cancel;
+    let cancelObj = {
+      'actionPath': 'cancel',
+      'label': 'Sign out',
+      'name': 'cancel',
+      'type': 'link'
+    };
+    const cancelInLinks = links.find(obj => obj.name === 'cancel');
+    const isTerminalState = this.options.appState.get('currentState').status === 'TERMINAL';
 
+    if (cancelFn && _.isFunction(cancelFn) && !cancelInLinks) {
+      //add cancel/signout link
+      links.push(cancelObj);
+    }
+
+    //remove cancel/signout link if TERMINAL state that were previously added in other states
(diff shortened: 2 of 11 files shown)


src/v2/view-builder/internals/BaseFooter.js
```diff
@@ -24,7 +24,25 @@ export default View.extend({
...
   initialize () {
     const links = _.resultCtx(this, 'links', this);
+    let cancelFn = this.options.appState.get('currentState').cancel;
+    let cancelObj = {
+      'actionPath': 'cancel',
+      'label': 'Sign out',
+      'name': 'cancel',
+      'type': 'link'
+    };
+    const cancelInLinks = links.find(obj => obj.name === 'cancel');
```
haishengwu-okta: I'd say as long as `cancelFn` is function, we could just add `cancelObject` to the `links` list.
For the case (cancel link has been added in sub class) shall be code fix in subclass instead of doing smart check in BaseFooter.

assets/sass/v2/_footer.scss
```diff
@@ -0,0 +1,8 @@
+.siw-main-footer {
+  .auth-footer {
+    .js-cancel {
+      display: block;
+      text-align: right;
```
haishengwu-okta: Cannot just be align to right which doesn't work if there are other links to the left. e.g. at required password page.

Screen Shot 2019-08-13 at 8 25 47 AM

nikhilvenkatraman-okta: yeah, it also has a display block on it, so the cancel link will actually be on its own line currently but might be better to float it right

src/v2/view-builder/internals/BaseFooter.js
nikhilvenkatraman-okta: I was thinking about that. But I prefer this since cancel is getting added in the `BaseFooter`, it can also do the job of making sure if removes it from the views that don't cancel.

haishengwu-okta: we'd only add `cancel` when the `cancelFn` exists. What use case we need to remove it even when `cancelFn` exists? at that case, can we just dont add the link?
If worry about sub class add cancel link which shall not be, it's bug in sub class and need to be fix there.

src/v2/view-builder/internals/BaseFooter.js
```diff
@@ -24,6 +24,19 @@ export default View.extend({
...
+    let cancelFn = this.options.appState.get('currentState').cancel;
+    let cancelObj = {
+      'actionPath': 'cancel',
+      'label': 'Sign out',
+      'name': 'cancel',
+      'type': 'link'
+    };
+    const isTerminalState = this.options.appState.get('currentState').status === 'TERMINAL';
+
+    if (cancelFn && _.isFunction(cancelFn) && !isTerminalState) {
```
haishengwu-okta: `cancelFn && _.isFunction(cancelFn)` can just be `_.isFunction(cancelFn)`

PR #830, opened 2019-08-13
Add polling at otp form

haishengwu-okta:
- move polling action to factor from remediation
- merge email polling and otp into one view
- fix incorrect step name


playground/mocks/api/v1/idx/data/factor-verification-email.json
```diff
@@ -48,6 +32,22 @@
         "profile": {
           "email": "o*****m@abbott.dev"
         },
+        "poll": {
```
nikhilvenkatraman-okta: 👍

src/v2/view-builder/views/OTPEmailView.js
```diff
@@ -1,14 +1,74 @@
...
+import { loc, View, createCallout, createButton, _ } from 'okta';
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
+import email from '../shared/email';
+import polling from '../shared/polling';
 
-const Body = BaseForm.extend({
-  title () {
-    return `Email link (${this.options.appState.get('factorEmail')})`;
+const SHOW_RESEND_TIMEOUT = 600;
```
nikhilvenkatraman-okta: How did we come up with this value ? Currently this was set to `const SHOW_RESEND_TIMEOUT = 60000;` In the factorPollVerificationEmailView.

src/v2/view-builder/shared/polling.js
```diff
@@ -0,0 +1,21 @@
+import { _ } from 'okta';
+
+export default {
+  addPolling () {
```
nikhilvenkatraman-okta: nit: should this be called `startPolling` ? and similarly `stopPolling` instead of `removePolling`?

src/v2/view-builder/views/OTPEmailView.js
haishengwu-okta: this shall be `60000`. modified for testing but forgot to change it back.

PR #838, opened 2019-08-15
fix: clears lastAuthResponse after a transaction is canceled

gowthamidommety-okta:
After clicking on sign out during passwordless flow, we see that lastAuthResponse is not reset and the widget still thinks it is in UNAUTHENTICATED state which causes the issue. Clearing lastAuthResponse when !res || !res.status fixes it.

Resolves: OKTA-238146

before: 
before

after:
after

### DIFF
--- a/src/util/RouterUtil.js
+++ b/src/util/RouterUtil.js
@@ -127,6 +127,7 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
   fn.routeAfterAuthStatusChange = function (router, res) {
     // Other errors are handled by the function making the authClient request
     if (!res || !res.status) {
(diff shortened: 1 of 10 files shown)


src/util/RouterUtil.js
```diff
@@ -127,6 +127,8 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
   fn.routeAfterAuthStatusChange = function (router, res) {
     // Other errors are handled by the function making the authClient request
     if (!res || !res.status) {
+      // clear lastAuthResponse when response is empty
```
mauriciocastillosilva-okta: nit: no need for the comment, it seems with the method name should be self explanatory...

PR #839, opened 2019-08-16
Add support for multiple factors: Switch factor

nikhilvenkatraman-okta:
Select factor with factor challenge

https://okta.box.com/s/631hm23l8uls7nz7ufvb6b2bi1pp5etz

Note that currently switch factor goes back to the factor list page where it will list all factors except the one you were currently on. This can potentially be changed with a fix on the API side later on to list all the factors, including the one you were currently on.

@haishengwu-okta @taylorlaubach-okta


haishengwu-okta:
Screen Shot 2019-08-16 at 2 23 55 PM

do we still need all those 5 or actually only `factor-required-options.json` ?

playground/mocks/api/v1/idx/challenge/data/factor-required-email-with-options.json
```diff
@@ -0,0 +1,102 @@
...
+  "intent": "LOGIN",
+  "remediation": {
+    "type": "array",
+    "value": [
+      {
+        "rel": [
+          "create-form"
+        ],
+        "name": "challenge-factor",
+        "href": "http://localhost:3000/api/v1/idx/challenge",
```
haishengwu-okta: curious shall this be `/api/v1/idx` or `/api/v1/idx/challenge`?

src/v2/models/AppState.js
```diff
@@ -35,21 +35,44 @@ export default Model.extend({
     factorProfile: {
       deps: ['factor'],
       fn (factor = {}) {
-        return factor.profile || {};
+        return factor && factor.profile || {};
```
haishengwu-okta: `factor = {}` in the function parameter would set `factor` to `{}` is it's undefined. why still add extra check?

src/v2/view-builder/views/RequiredFactorEmailView.js
```diff
@@ -1,15 +1,33 @@
...
   },
   email,
 ));
 
+const Footer = BaseFooter.extend({
+  links: function () {
+    var links = [];
+    const nextViewState = this.options.appState.getNextViewState();
+    // if there are 2 forms, the second form is the switch factor form
+    if (nextViewState) {
```
haishengwu-okta: shall be 
```
this.options.appState.hasRemediationForm('switch-factor');

///  which has implementation in appstate like
return this.get('remedation').value.filter(v => v.name === name).length === 1;
```

src/v2/models/AppState.js
```diff
@@ -35,21 +35,44 @@ export default Model.extend({
...
     },
-    isTerminalState: {
+    isTerminalStep: {
       deps: ['terminal'],
       fn: function (terminal) {
-        return !_.isEmpty(terminal);
+        return !_.isEmpty(terminal) && terminal.value.length;
       },
     },
+    isAuthenticateStep: {
```
haishengwu-okta: we already dispatch logic in ViewFactory base on form-name. I think we could do Class extension instead of creating derived prop base on step. I'm imaging we'll have

- SelectFactorView
- SwitchFactorView

which could extend from some BaseFactorListView

src/v2/models/AppState.js
```diff
@@ -74,7 +97,8 @@ export default Model.extend({
   setIonResponse (resp) {
     // Don't re-render view if new response is same as last.
     // Usually happening at polling and pipeline doesn't proceed to next step.
-    if (_.isEqual(resp.__rawResponse, this.get('__rawResponse'))) {
+    if (_.isEqual(resp.__rawResponse, this.get('__rawResponse'))
+        && this.get('__rawResponse').remediation.value.length === 1) {
```
haishengwu-okta: do we still need this modification?

src/v2/view-builder/ViewFactory.js
```diff
@@ -21,14 +21,17 @@ const VIEWS_MAPPING = {
   'select-factor': {
     [DEFAULT]: SelectFactorView,
   },
+  'switch-factor': {
```
haishengwu-okta: nit:
add comments regarding when is select-factor v.s when is switch-factor.

src/v2/view-builder/components/FactorOptions.js
```diff
@@ -49,8 +55,15 @@ export default ListView.extend({
   itemSelector: '.list-content',
 
   initialize: function () {
+    this.step = this.options.appState.get('currentState').step;
+    this.currentState = this.options.appState.get('currentState');
     this.listenTo(this.collection,'selectFactor', function (data) {
-      this.model.set('factorType', data);
```
haishengwu-okta: could be  `this.model.set(this.options.name, data)` if you made such change
```
--- a/src/v2/view-builder/internals/FormInputFactory.js
+++ b/src/v2/view-builder/internals/FormInputFactory.js
@@ -14,6 +14,7 @@ const createFactorTypeView = (opt) => {
   return {
     View: FactorOptions,
     options: {
+      name: opt.name,
       collection: new Collection(optionItems),
     }
   };
```

src/v2/view-builder/internals/BaseFooter.js
```diff
@@ -22,12 +22,15 @@ export default View.extend({
...
+    return isTerminalStep;
+  },
+
   initialize () {
     let links = _.resultCtx(this, 'links', this);
     const cancelFn = this.options.appState.get('currentState').cancel;
-    const isTerminalState = this.options.appState.get('isTerminalState');
-
-    if (_.isFunction(cancelFn) && !isTerminalState) {
+    if (_.isFunction(cancelFn) && !this.hideSignOutLink()) {
```
haishengwu-okta: did you confirm if `terminal` response has `cancel` then we dont need `isTerminalStep` at all?

src/v2/controllers/FormController.js
```diff
@@ -44,7 +44,16 @@ export default BaseLoginController.extend({
     this.listenTo(this.formView, 'save', this.handleFormSave);
   },
 
-  invokeAction (actionPath = '') {
+  invokeAction (options = {}) {
```
haishengwu-okta: do you still the following changes?

nikhilvenkatraman-okta: ah, good catch. This can be reverted now.

src/v2/models/AppState.js
nikhilvenkatraman-okta: This was actually failing in the HeaderBeacon where we check for factor. I can move this check in there `factor && factor.factorType`

haishengwu-okta: do you mean this piece?
```
const factor = this.options.appState.get('factor');
    let className = 'undefined-user';
    if (factor) {
      className = getFactorData(factor.factorType).iconClassName;
    }
```
we actually already have `factorType` as derived prop, could you use that instead?
probably need to update `getFactorData` a bit by return `{}` if found nothing from `factorData`

src/v2/view-builder/views/RequiredFactorEmailView.js
nikhilvenkatraman-okta: 👍

playground/mocks/api/v1/idx/challenge/data/factor-required-email-with-options.json
nikhilvenkatraman-okta: Need to confirm with API folks

src/v2/models/AppState.js
nikhilvenkatraman-okta: yeah, thats the place. I can use `factorType ` from appState

src/v2/view-builder/components/FactorOptions.js
nikhilvenkatraman-okta: 👍

src/v2/models/AppState.js
nikhilvenkatraman-okta: oops removed

src/v2/models/AppState.js
nikhilvenkatraman-okta: yeah, agree. This was a quick fix to get it working

assets/sass/v2/_footer.scss
```diff
@@ -1,7 +1,11 @@
 .siw-main-footer {
   .auth-footer {
+    .link {
+      margin-right: 10px;
```
haishengwu-okta: I wonder it might be better to align links vertically. (maybe another JIRA?)

Screen Shot 2019-08-16 at 3 57 39 PM

src/v2/models/AppState.js
```diff
@@ -41,15 +41,31 @@ export default Model.extend({
     factorType: {
       deps: ['factor'],
       fn (factor = {}) {
-        return factor.factorType;
+        return factor && factor.factorType;
```
haishengwu-okta: just `factor.factorType` shall good enough?

src/v2/models/AppState.js
```diff
@@ -41,15 +41,31 @@ export default Model.extend({
...
       },
     },
-    isTerminalState: {
+    isTerminalStep: {
```
haishengwu-okta: looks like this can be removed now.

src/v2/view-builder/views/RequiredFactorPasswordView.js
```diff
@@ -11,14 +11,26 @@ const Body = BaseForm.extend({
...
+  links: function () {
+    var links = [
+      {
+        'type': 'link',
+        'label': 'Forgot Password',
+        'name': 'forgot-password',
+        'actionPath': 'factor.recovery',
+      }
+    ];
+    // if there are 2 forms, the second form is the switch factor form
```
haishengwu-okta: comment is not accurate now. ditto in the another place.

assets/sass/v2/_footer.scss
nikhilvenkatraman-okta: Yup, agree. We are running out of space

src/v2/models/AppState.js
nikhilvenkatraman-okta: Thinking about this a bit more, we get back `select-factor` as the form name for both factor Required flow and factor enroll. Only thing different is that actual step `Authenticate` vs        
` Enroll` and we want to show different formTitle

haishengwu-okta: I thought we'll have `select-factor` and `challenge-factor` respectively?

src/v2/controllers/FormController.js
```diff
@@ -31,6 +31,14 @@ export default BaseLoginController.extend({
   postRender () {
     const currentViewState = this.options.appState.getCurrentViewState();
 
+    if (currentViewState.name === 'select-factor') {
```
haishengwu-okta: prefer to move this logic into `getCurrentViewState` and add comments regarding why override the form name.

src/v2/view-builder/components/FactorOptions.js
```diff
@@ -49,8 +51,11 @@ export default ListView.extend({
   itemSelector: '.list-content',
 
   initialize: function () {
+    this.step = this.options.appState.get('currentState').step;
```
haishengwu-okta: `this.step` isn't used.

src/v2/models/AppState.js
```diff
@@ -44,14 +44,24 @@ export default Model.extend({
...
-      deps: ['terminal'],
-      fn: function (terminal) {
-        return !_.isEmpty(terminal);
+    isAuthenticateStep: {
+      deps: ['currentState'],
+      fn: function (currentState = {}) {
+        return currentState && currentState.step === 'AUTHENTICATE';
+      },
+    },
+    isIdentifyStep: {
```
haishengwu-okta: `isIdentifyStep` isn't used. remove it?

nikhilvenkatraman-okta: It will be used to remove the sign out link on login page.

playground/mocks/api/v1/idx/challenge/data/factor-required-email-with-options.json
nikhilvenkatraman-okta: According to the postman collection its `idx/challenge` 

Screen Shot 2019-08-20 at 11 48 55 AM

taylorlaubach-okta:
@nikhilvenkatraman-okta I know I'm a little late to the game here. Looks good overall.

1. I do think that we should show all available factors (including the one they came from) after they choose switch. Otherwise, if they change their mind, they can't get back to the one they had before without picking another factor and then choosing switch again.
2. I think the beacon as you have it working is good. (We can revisit later the various states for the beacon and the lack of security image AFAIK in OIE.)
3. Can you move the buttons to the right to match how I had mocked them up for recovery?
4. Can we also update the text to match the screens from the recovery mockup? (e.g. "authentication method", "use" on the buttons)

haishengwu-okta:
@taylorlaubach-okta 
- for #2 what about hide the beacon at factor list page?

taylorlaubach-okta:
@haishengwu-okta I agree that the current implementation could be improved (especially if we are going to just show a grayed out user icon most of the time). That said, if we change it, we have to either: a. pick a new thing to show in the beacon, b. hide the beacon during some steps and show it again in others, or c. remove the beacon in other flows, too. Any of those things are a lot to figure out (and probably to implement, too), so I would leave this as is for now. I say that because I believe we are going to do larger improvements to the SIW soon, and then we can address how we are using the beacon overall.

PR #844, opened 2019-08-20
fix: add excludeCredentials support to EnrollWebauthnController.js

mpeng-okta:
Just convert excludeCredentials' public key credential object's `id` when it comes in from the backend. This fixes a bug where we would allow duplicate WebAuthn authenticator enrollments.

Resolves: OKTA-245259


haishengwu-okta:
Version bump up will be taken care of before promotion. 
no worry from developers.

PR #846, opened 2019-08-21
fix: Displays error message when verification with an idp fails

gowthamidommety-okta:
Resolves: OKTA-240089


src/models/AppState.js
```diff
@@ -172,7 +172,10 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
       // language picker, etc.
       languageCode: ['string', true],
       disableUsername: ['boolean', false, false],
-      trapMfaRequiredResponse: ['boolean', false, false]
+      trapMfaRequiredResponse: ['boolean', false, false],
+      hasFailed: ['boolean', false],
+      errorMessage: ['string', false],
```
magizh-okta: may we can add all into one object or add proper prefixes.

gowthamidommety-okta:
@magizh-okta Yea, will attach screenshots with the actual error messages now

src/models/AppState.js
magizh-okta: factorVerificationError :{
  errorMessage: ''",
  factor: ['object', false]
}


factorVerificationFailed: 
factorVerificationFailureMessage:

src/util/RouterUtil.js
```diff
@@ -218,14 +218,22 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
...
       router.navigate(url, { trigger: true });
+      router.appState.unsetFactorResultFailedProperties();
       return;
     case 'MFA_CHALLENGE':
       // Since we normally trap MFA_CHALLENGE, this will only get called on a
-      // page refresh. We need to return to MFA_REQUIRED to initialize the
+      // page refresh or when an error is returned on verification with an IdP.
+      // We need to return to MFA_REQUIRED to initialize the
       // page correctly (i.e. factors dropdown, etc)
+      var lastAuthResponse = router.appState.get('lastAuthResponse');
```
mauriciocastillosilva-okta: so with what I am suggesting, this would be something like
```
if (router.appState.get('isFactorResultFailed')) {
  router.appState.set('lastChallengedFactorData', {
    factor: router.appState.get('factor'),
    errorMessage: router.appState.get('factorResultMessage'),
  });
}
```

src/util/RouterUtil.js
```diff
@@ -218,14 +218,22 @@ function (Okta, OAuth2Util, Util, Enums, BrowserFeatures, Errors, ErrorCodes) {
     case 'FACTOR_REQUIRED':
     case 'FACTOR_CHALLENGE':
     case 'MFA_REQUIRED':
+      if (router.appState.get('hasFailed') && router.appState.get('challengeErrorFactor')) {
```
mauriciocastillosilva-okta: should we change `getDefaultFactor ` to include logic to get the last challenged factor that failed? It might be worth having that logic in the factors instead of here in the router util...

src/util/RouterUtil.js
gowthamidommety-okta: lastFailedChallengedFactorData

src/util/RouterUtil.js
gowthamidommety-okta: Discussed, not doing this

src/util/RouterUtil.js
gowthamidommety-okta: Changing to lastFailedChallengeFactorData

PR #850, opened 2019-08-22
Integrate user enroll flow

nikhilvenkatraman-okta:
- Integrates user enroll flow with API 
- https://okta.box.com/s/otj61k791xqmv8cx8uv8kl4i84s1piew
- Created OKTA-245762 to handle success after enroll is completed.

### DIFF
--- a/playground/mocks/api/v1/idx/introspect/data/identify.json
+++ b/playground/mocks/api/v1/idx/introspect/data/identify.json
@@ -26,22 +26,26 @@
               "visible": false
             }
           ]
-        }
-      ]
-    },
-    "enroll": {
-      "rel": [
-        "create-form"
-      ],
-      "name": "enroll",
-      "href": "http://localhost:3000/api/v1/idx/enroll",
-      "method": "POST",
-      "accepts": "application/vnd.okta.v1+json",
-      "value": [
+        },
         {
-          "name": "stateHandle",
-          "value": "01r2p5S9qaAjESMFuPzt7r3ZMcZZQ_vvS0Tzg56ajB",
-          "visible": false
+          "rel": [
+            "create-form"
+          ],
+          "name": "select-enroll-profile",
+          "href": "http://localhost:3000/api/v1/idx/enroll",
+          "method": "POST",
+          "accepts": "application/vnd.okta.v1+json",
+          "value": [
+            {
+              "name": "firstName",
+              "label": "FirstName"
+            },
+            {
+              "name": "stateHandle",
+              "value": "01r2p5S9qaAjESMFuPzt7r3ZMcZZQ_vvS0Tzg56ajB",
+              "visible": false
+            }
+          ]
         }
       ]
     },
--- a/src/v2/ion/uiSchemaTransformer.js
+++ b/src/v2/ion/uiSchemaTransformer.js
@@ -29,6 +29,16 @@
  * @param {IONFormField[]} remediationValue
  */
 const createUISchema = (remediationValue = []) => {
+  // For casee where field itself is a form, it has a formname and we are appending the formname to each field
+  // This is so that while making the request we can bundle these key:value pairs under the same key name
+  // For simplicity we are assuming that when field itself is a form its only one level deep
+  if (remediationValue[0] && remediationValue[0].name && remediationValue[0].form) {
+    const inputGroupName = remediationValue[0].name;
+    remediationValue = remediationValue[0].form.value.map(input => {
+      input.name = inputGroupName + '.' + input.name;
+      return input;
+    });
+  }
   return remediationValue.map(ionFormField => {
     const uiSchema = {
       type: 'text',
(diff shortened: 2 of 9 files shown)


src/v2/models/AppState.js
```diff
@@ -52,6 +52,10 @@ export default Model.extend({
     },
   },
 
+  isFormInRemediation (formName) {
+    return this.get('remediation').filter(r => r.name === formName)[0];
```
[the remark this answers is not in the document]
haishengwu-okta: looks like you didn't get latest update. you had added `hasRemediationForm`?

src/v2/ion/uiSchemaTransformer.js
```diff
@@ -57,8 +57,17 @@ const createUISchema = (remediationValue = []) => {
 const insertUISchema = (transformedResp) => {
   if (transformedResp.currentState) {
     const remediation = transformedResp.currentState.remediation || [];
-
     transformedResp.currentState.remediation = remediation.map(obj => {
+      // For casee where field itself is a form, it has a formname and we are appending the formname to each field
+      // This is so that while making the request we can bundle these key:value pairs under the same key name
+      // We are assuming that when field itself is a form its only one level deep
+      if (obj.value[0] && obj.value[0].name && obj.value[0].form) {
```
haishengwu-okta: worth unit test this logic

playground/mocks/api/v1/idx/introspect/data/identify.json
```diff
@@ -26,22 +26,26 @@
...
-          "visible": false
+          "rel": [
+            "create-form"
+          ],
+          "name": "select-enroll-profile",
+          "href": "http://localhost:3000/api/v1/idx/enroll",
+          "method": "POST",
+          "accepts": "application/vnd.okta.v1+json",
+          "value": [
+            {
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: ah, this should have the same fields as the `enroll-profile` form

src/v2/models/AppState.js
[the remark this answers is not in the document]
nikhilvenkatraman-okta: yep, looks like my branch is not updated. I was wondering the same when i added this method 😄

src/v2/ion/uiSchemaTransformer.js
haishengwu-okta: - feels like the logic shall go into `createUISchema` (before the `remediationValue.map`) so that keep `remediation.value` unchanged.
- consequently, we need to change `BaseModel.create` to use `remediation.uiSchema` instead of `remediation.value`

src/v2/models/AppState.js
```diff
@@ -53,8 +53,15 @@ export default Model.extend({
     showSignoutLink: {
       deps: ['currentState'],
       fn: function (currentState = {}) {
-        // hide signout for IDENTIFY step
-        return _.isFunction(currentState.cancel) && currentState.step !== 'IDENTIFY';
+        // hide signout for IDENTIFY and ENROLL step
+        if (_.isFunction(currentState.cancel)) {
+          let showSignout = true;
+          if (currentState.step !== 'IDENTIFY'||
+            currentState.step !== 'ENROLL') {
```
haishengwu-okta: why hide sign-out in `enroll` step?

nikhilvenkatraman-okta: sign out in enroll step does not make sense, since the user has not logged in and just hit sign up button from login screen

src/v2/ion/uiSchemaTransformer.js
```diff
@@ -29,6 +29,16 @@
  * @param {IONFormField[]} remediationValue
  */
 const createUISchema = (remediationValue = []) => {
+  // For casee where field itself is a form, it has a formname and we are appending the formname to each field
+  // This is so that while making the request we can bundle these key:value pairs under the same key name
+  // For simplicity we are assuming that when field itself is a form its only one level deep
+  if (remediationValue[0] && remediationValue[0].name && remediationValue[0].form) {
```
haishengwu-okta: why you choose index `0`? what if the 2nd item is has `form` attribute?

```
{
    "value": [
        {
            "name": "Password"
        },
        {
            "name": "userProfile",
            "form": {
                "value": [
                    {
                        "name": "lastName",
                        "label": "Last name",
                        "required": true
                    },
                ]
            }
        }
    ]
}
```

I think the we need loop through the `remediationValue` to convert each entry that has `form`. something like (psudecode)
```
newValue = _.chain(remediationValue)
  .map(v => {
    if (v.form) {
        const inputGroupName = v.name;
        return v.form.value.map(input => {
            input.name = inputGroupName + '.' + input.name;
            return input;
          });
    } else {
        return v
    }
  })
  .flatten()
  .value();
```

src/v2/models/AppState.js
```diff
@@ -53,8 +53,9 @@ export default Model.extend({
     showSignoutLink: {
       deps: ['currentState'],
       fn: function (currentState = {}) {
-        // hide signout for IDENTIFY step
-        return _.isFunction(currentState.cancel) && currentState.step !== 'IDENTIFY';
+        const invalidSignOutSteps = ['IDENTIFY', 'ENROLL', 'SUCCESS'];
+        // hide signout for IDENTIFY and ENROLL step
```
haishengwu-okta: - nit: `hide signout for IDENTIFY, ENROLL and SUCCESS step
- do you know why even `success` has `cancel` link?

nikhilvenkatraman-okta: Yeah, the API will send `cancel` in the SUCCESS state. According to @iancorcoran-okta you could still cancel without redeeming for tokens. We have asked Tom to confirm this.

src/v2/ion/uiSchemaTransformer.js
nikhilvenkatraman-okta: This was mostly because we currently only are supporting enrollment of userProfile.

PR #855, opened 2019-08-28
Add support for success state

nikhilvenkatraman-okta:


src/v2/models/AppState.js
```diff
@@ -77,7 +77,7 @@ export default Model.extend({
         Logger.warn(`Cannot find view state for form ${currentFormName}. Fall back to terminal state.`);
       }
 
-      currentViewState = this.get('terminal');
+      currentViewState = this.get('currentState').success ? this.get('currentState').success : this.get('terminal');
```
haishengwu-okta: - does `success` still exist as first citizen when `success` is a form in API?
- instead of changing formName in another function, can we do here like
```
if step === success then return 
{
  name: 'success-redirect',
}
```

haishengwu-okta: I see. the `success` is there by accident (or issue in the transformation layer). probably shall't rely on it.

nikhilvenkatraman-okta: success does not exist as a firstLevel citizen when we are trying to set `currentViewState `. Thats why i had to reach into the currentState object.

We are not really changing the formName right, we are just reading it from `resp.currentState.success.name` since there is not remediation for the success form. We can avoid checking for the step here then.

haishengwu-okta: then do you need the changes at line 101 below?

nikhilvenkatraman-okta: yes, thats where we set `currentFormName`

haishengwu-okta: `currentForm` is used in `getCurrentViewState` to get form data from remediation. since you dont use `formName` at all during the `success`, I doubt we need to change anythings in `setIonResponse`?

nikhilvenkatraman-okta: I think we can fix the responseTransformer so that we don't have to read form name from resp.currentState.success.name

src/v2/ion/responseTransformer.js
```diff
@@ -167,6 +167,10 @@ const createCurrentStateObject = (originalResp, omitKeys) => {
   if (resp.remediation) {
     remediation = resp.remediation.value.map(normalizeRemedation);
   }
+  // success does not have remediation
+  if (resp.success) {
+    remediation = [resp.success].map(normalizeRemedation);
```
haishengwu-okta: omit `success` at line 164.

src/v2/view-builder/views/SuccessView.js
```diff
@@ -0,0 +1,21 @@
...
+    return  'You will be redirected';
+  },
+  noButtonBar: true,
+  initialize (options) {
+    /* triggering invokeAction instead of saveForm here. This is because we remove 
+    firstLevelObjects when we call createCurrentStateObject in responseTransformer
+    which causes success to get omitted from currentState object, and hence we cant call 
+    this.options.appState.get('currentState')[formName] inside handleFormSave
+    */
+    options.appState.trigger('invokeAction', 'success-redirect');
```
haishengwu-okta: we can do `saveForm` now?

nikhilvenkatraman-okta: yup

src/v2/view-builder/views/SuccessView.js
```diff
@@ -0,0 +1,16 @@
+import BaseView from '../internals/BaseView';
+import BaseForm from '../internals/BaseForm';
+
+const Body = BaseForm.extend({
+  title () {
+    return  'You will be redirected';
+  },
+  noButtonBar: true,
+  initialize (options) {
+    options.appState.trigger('saveForm', options.model);
```
haishengwu-okta: call `super.initialize`.

PR #856, opened 2019-08-29
 fix: changes days remaining text to bold font

gowthamidommety-okta:
Resolves: OKTA-241034

ALL_OPTIONAL_SOME_ENROLLED:
HAS_REQUIRED_ALL_REQUIRED_ENROLLED:

Screen Shot 2019-08-29 at 9 39 26 AM

HAS_REQUIRED_SOME_REQUIRED_ENROLLED (in grace period):

Screen Shot 2019-08-28 at 5 29 40 PM

Screen Shot 2019-08-28 at 5 31 39 PM

HAS_REQUIRED_SOME_REQUIRED_ENROLLED (out of grace period):

Screen Shot 2019-08-29 at 10 07 41 AM

Default (HAS_REQUIRED_NONE_ENROLLED, ALL_OPTIONAL_NONE_ENROLLED):

Screen Shot 2019-08-29 at 9 38 34 AM

Screen Shot 2019-08-29 at 10 08 56 AM


src/EnrollChoicesController.js
```diff
@@ -36,13 +40,11 @@ function (Okta, FormController, Enums, RouterUtil, FactorList,
     Form: {
       noCancelButton: true,
 
-      title: _.partial(Okta.loc, 'enroll.choices.title', 'login'),
```
mauriciocastillosilva-okta: I don't see any special handling for the title, why do we need to change this?

src/EnrollChoicesController.js
```diff
@@ -36,13 +40,11 @@ function (Okta, FormController, Enums, RouterUtil, FactorList,
...
-
       noButtonBar: function () {
         return this.state.get('pageType') === Enums.ALL_OPTIONAL_NONE_ENROLLED;
       },
 
-      subtitle: function () {
```
mauriciocastillosilva-okta: same here, are we changing this to add a specific css class to the subtitle?

gowthamidommety-okta: The main reason i need to put this in a template is because the courage template for subtitle uses {{subtitle}} which escapes the html. I'm using triple stash {{{subtitle}}} here

src/EnrollChoicesController.js
gowthamidommety-okta: Okta.tpl takes allows us to pass an argument for which element we want to attach the template to. When i passed `$('.o-form-content')` for the subtitle element, it attached it as the first child.. So the order was subtitle -> title.. I couldn't find a way to fix this besides creating a template for title as well.. Is there a way to attach the template as a sibling instead?

src/EnrollChoicesController.js
```diff
@@ -103,6 +107,8 @@ function (Okta, FormController, Enums, RouterUtil, FactorList,
       },
 
       preRender: function () {
+        this.add(subtitleTpl({subtitle: this.getSubtitle()}), $('p.okta-form-subtitle'));
```
mauriciocastillosilva-okta: mind adding a small comment about why we do this? Something like `// form subtitle does not support unescaped strings, and we need some html elements in the subtitle for this form. So instead of a regular subtitle, we add a <span> with the text we need`

packages/@okta/i18n/dist/properties/login.properties
```diff
@@ -296,8 +296,8 @@ enroll.choices.title = Set up multifactor authentication
 enroll.choices.description = Your company requires multifactor authentication to add an additional layer of security when signing in to your Okta account
 enroll.choices.description.generic = Your company requires multifactor authentication to add an additional layer of security when signing in to your account
 enroll.choices.description.specific = Your company requires multifactor authentication to add an additional layer of security when signing in to your {0} account
-enroll.choices.description.gracePeriod = Your company recommends setting up additional factors for authentication. Set up will be required in: {0} day(s).
-enroll.choices.description.gracePeriod.oneDay = Your company recommends setting up additional factors for authentication. Set up will be required in: less than 1 day.
+enroll.choices.description.gracePeriod.bold = Your company recommends setting up additional factors for authentication. Set up will be required in: <b>{0} day(s)</b>.
+enroll.choices.description.gracePeriod.oneDay.bold = Your company recommends setting up additional factors for authentication. Set up will be required in: <b>less than a day</b>.
```
mauriciocastillosilva-okta: you might even be able to just edit the existing ones... Let's check with @santhoshbalakrishnan

gowthamidommety-okta: He replied saying it should be add and move existing strings to deleted section

PR #862, opened 2019-09-05
feat: risk number challenge:

magizh-okta:
* adds Number challenge screen in okta-verify challenge flow. 
* adds callback to update lastauthResponse when poll is happening.

The thing is there is no state change for factorResult/status. The only thing that changes is that during the poll the authn response has extra embedded info. 
https://user-images.githubusercontent.com/51595754/63372918-eab92600-c33b-11e9-8ecc-786e431db66b.png

I am using the same push form to display the number challenge view and hide the save button because the poll would still happen and things like timeout/cancel will be taken care automatically. 

Auth js change 
https://github.com/okta/okta-auth-js/pull/245

### DIFF
--- a/assets/sass/modules/_mfa-challenge-forms.scss
+++ b/assets/sass/modules/_mfa-challenge-forms.scss
@@ -67,3 +67,40 @@
     width: 100%;
   }
 }
+
+// number challenge Okta-verify push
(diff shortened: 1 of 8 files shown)


magizh-okta:
duplicate

PR #868, opened 2019-09-11
Fix playground

nikhilvenkatraman-okta:
- Update  playground folder structure to match API `/idp/idx`


nikhilvenkatraman-okta:
That got removed when i copied files over. Added it back now.

PR #872, opened 2019-09-16
Add message for unknown user

nikhilvenkatraman-okta:
- Add support to show callout when unknown user is found 
https://okta.box.com/s/7kv55iqy0xcy7plyxcgq4qgk0y9rhd80

- Move CSS to callout.scss

Screen Shot 2019-09-16 at 2 32 06 PM

### DIFF
--- a/assets/sass/v2/_all.scss
+++ b/assets/sass/v2/_all.scss
@@ -2,3 +2,4 @@
 @import './beacon';
 @import './enroll-profile';
 @import './factor-poll-verification';
+@import './callout';
--- a/assets/sass/v2/_callout.scss
+++ b/assets/sass/v2/_callout.scss
@@ -1,0 +1,16 @@
+.siw-main-view {
+  .clearfix:before,
+  .clearfix:after {
+    visibility: visible;
+    content: '';
+    height: 102%;
+  }
+  .infobox {
+    a {
+      color: #007dc1;
+    }
+    strong {
+      font-weight: bold;
+    }
+  }
+}
--- a/assets/sass/v2/_callout.scss
+++ b/assets/sass/v2/_callout.scss
@@ -7,7 +7,7 @@
   }
   .infobox {
     a {
-      color: #007dc1;
+      color: $primary-color;
     }
     strong {
       font-weight: bold;
--- a/playground/mocks/idp/idx/introspect/data/identify.json
+++ b/playground/mocks/idp/idx/introspect/data/identify.json
@@ -12,7 +12,7 @@
             "create-form"
           ],
           "name": "identify",
-          "href": "http://localhost:3000/api/v1/idx",
+          "href": "http://localhost:3000/idp/idx",
           "method": "POST",
           "accepts": "application/vnd.okta.v1+json",
           "value": [
@@ -32,7 +32,7 @@
             "create-form"
           ],
           "name": "select-enroll-profile",
-          "href": "http://localhost:3000/api/v1/idx/enroll",
+          "href": "http://localhost:3000/idp/idx/enroll",
           "method": "POST",
           "accepts": "application/vnd.okta.v1+json",
           "value": [
@@ -50,7 +50,7 @@
         "create-form"
       ],
       "name": "cancel",
-      "href": "http://localhost:3000/api/v1/idx/cancel",
+      "href": "http://localhost:3000/idp/idx/cancel",
       "method": "POST",
       "value": [
         {
--- a/playground/mocks/idp/idx/introspect/data/identify.json
+++ b/playground/mocks/idp/idx/introspect/data/identify.json
@@ -45,6 +45,10 @@
         }
       ]
     },
+      "messages": {
+      "type": "array",
+      "value": []
+    },
     "cancel": {
       "rel": [
         "create-form"
--- a/src/v2/models/AppState.js
+++ b/src/v2/models/AppState.js
@@ -108,10 +108,8 @@ export default Model.extend({
 
     this.set(resp);
 
-    // response has messages object with the same form
-    if (!_.isEmpty(resp.currentState.messages)) {
-      this.trigger('showCallout', resp.currentState.messages);
-    }
+    // broadcast idxResponseUpdated to re-render the view
+    this.trigger('idxResponseUpdated', resp);
   }
 
 });
--- a/src/v2/models/AppState.js
+++ b/src/v2/models/AppState.js
@@ -107,6 +107,11 @@ export default Model.extend({
     }
 
     this.set(resp);
+
+    // response has messages object with the same form
+    if (!_.isEmpty(resp.currentState.messages)) {
+      this.trigger('showCallout', resp.currentState.messages);
+    }
   }
 
 });
--- a/src/v2/view-builder/internals/BaseView.js
+++ b/src/v2/view-builder/internals/BaseView.js
@@ -1,4 +1,4 @@
-import { View } from 'okta';
+import { View, createCallout } from 'okta';
 import BaseForm from './BaseForm';
 import BaseModel from './BaseModel';
 import BaseHeader from './BaseHeader';
@@ -15,6 +15,7 @@ export default View.extend({
   className: 'siw-main-view',
 
   template: '<div class="siw-main-header"></div>' +
+      '<div class="siw-callout"></div>' +
       '<div class="siw-main-body"></div>' +
       '<div class="siw-main-footer"></div>',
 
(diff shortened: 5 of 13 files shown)


src/v2/view-builder/views/TerminalView.js
```diff
@@ -5,7 +5,7 @@ import BaseFooter from '../internals/BaseFooter';
 const Body = BaseForm.extend({
   title () {
     const msg = this.options.appState.get('terminal').message || {};
-    return msg.value || 'You can close this window';
+    return msg.message || 'You can close this window';
```
nikhilvenkatraman-okta: This is for an API that is being made with this PR https://github.com/okta/okta-core/pull/36839

assets/sass/v2/_callout.scss
```diff
@@ -0,0 +1,16 @@
+.siw-main-view {
+  .clearfix:before,
+  .clearfix:after {
+    visibility: visible;
+    content: '';
+    height: 102%;
+  }
+  .infobox {
+    a {
+      color: #007dc1;
```
haishengwu-okta: are you able to use an existing variable?

src/v2/view-builder/views/TerminalView.js
haishengwu-okta: can we update API mocks as well?

src/v2/view-builder/views/email/RequiredFactorEmailView.js
```diff
@@ -34,7 +34,7 @@ const ResendView = View.extend(
       // and it hide anything from `:before` node hence remove
       // `clearfix` as workaround rather change existing style
       // just to be extreme safe.
-      this.$el.find('.infobox').removeClass('clearfix');
+      //this.$el.find('.infobox').removeClass('clearfix');
```
haishengwu-okta: probably just delete the code and comment

src/v2/models/AppState.js
```diff
@@ -107,6 +107,11 @@ export default Model.extend({
     }
 
     this.set(resp);
+
+    // response has messages object with the same form
+    if (!_.isEmpty(resp.currentState.messages)) {
+      this.trigger('showCallout', resp.currentState.messages);
```
haishengwu-okta: feels like we could consolidate `change:currentFormName` and `showCallout` to be something a little more generic like `idxResponseUpdated`

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -36,6 +36,26 @@ const Footer = BaseFooter.extend({
 });
 
 export default BaseView.extend({
+  initialize () {
```
haishengwu-okta: Instead of being specific for `IdentifierView`, maybe we could pass the `messages` into View from FormController

```
this.formView = this.add(TheView, {
      options: {
        currentViewState,
        messages: this.settings.get(...),
      }
    }).last();
```

Then render the message from BaseView.

assets/sass/v2/_all.scss
```diff
@@ -2,3 +2,4 @@
 @import './beacon';
 @import './enroll-profile';
 @import './factor-poll-verification';
+@import './callout';
```
haishengwu-okta: 🥇

src/v2/view-builder/views/email/RequiredFactorEmailView.js
nikhilvenkatraman-okta: yes, will do

src/v2/view-builder/views/TerminalView.js
nikhilvenkatraman-okta: yeah, just waiting to verify with the API changes.

src/v2/models/AppState.js
nikhilvenkatraman-okta: In this case the formName actually does not change. We get back the same form with a `message` object in the response now.

haishengwu-okta: I meant we could use single event rather than `change:currentFormName` and `showCallout` to trigger FormController `render` upon API response. basically it says 'whenever we got API response and it's different from last time, re-render the view`

playground/mocks/idp/idx/introspect/data/identify.json
```diff
@@ -44,6 +44,10 @@
           ]
         }
       ]
+    },
+      "messages": {
```
haishengwu-okta: does the `identify` response actually have empty `messages` ? otherwise we shall handle the null-able in code.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -36,6 +36,25 @@ const Footer = BaseFooter.extend({
 });
 
 export default BaseView.extend({
+  initialize (options) {
+    // add callout for messages
+    this.messages = options.messages;
+    if (this.messages && this.messages.value.length) {
```
haishengwu-okta: I was thinking is to invoke showMessageCallout from BaseView. The idea is in BaseView, it renders callout whenever there are non-empty `messages`. what do you think?

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -36,6 +36,25 @@ const Footer = BaseFooter.extend({
...
+      this.showMessageCallout(options.messages.value[0].message, 'warning');
+    }
+    BaseView.prototype.initialize.call(this, arguments);
+  },
+  postRender () {
+    // If user enterted identifier is not found, API sends back a message with a link to sign up
+    // This is the click handler for that link
+    if (this.messages && this.messages.value.length) {
```
haishengwu-okta: - may not need the check because when `this.$el.find()` finds no elements, it's essentially a no operation.? What do you think?
- The comments explains is great!

playground/mocks/idp/idx/introspect/data/identify.json
nikhilvenkatraman-okta: yeah, I tried this on Peter's branch and it has the empty array. They have a JIRA to fix it post LEA

src/v2/view-builder/views/IdentifierView.js
nikhilvenkatraman-okta: I didn't want to add the listener at all if the messages object was not present.

src/v2/view-builder/views/IdentifierView.js
nikhilvenkatraman-okta: I see what you are saying now. Just moved this logic to BaseView.

src/v2/view-builder/views/IdentifierView.js
haishengwu-okta: my understanding is `this.$el.find()` will get empty array if the `js-sign-up` element is not present hence no listener is added.

src/v2/models/AppState.js
nikhilvenkatraman-okta: Although this sounds good, I feel like this has a potential to cause regression in other places. I tried this change and I broke the switch factor flow. I think we should do this refactor later, possible after LEA.

haishengwu-okta: - now I realised how bad it is without unit test/e2e test in place..
- I think you need to fix `switchForm` in `FormControlller`. If still broken, create a jira for it.

nikhilvenkatraman-okta: thats exactly why I am hesitant to make this change

src/v2/view-builder/internals/BaseView.js
```diff
@@ -34,6 +35,32 @@ export default View.extend({
       },
     });
     this.add(this.Footer, { selector : '.siw-main-footer' });
+
+    // add callout for messages
+    this.messages = options.messages;
+    if (this.messages && this.messages.value.length) {
+      this.showMessageCallout(options.messages.value[0].message, 'warning');
```
haishengwu-okta: looks like we could leverage `"class": "INFO"` from API. I guess it's fine for LEA hence could be another follow up ticket.

src/v2/view-builder/internals/BaseView.js
```diff
@@ -34,6 +35,32 @@ export default View.extend({
       },
     });
     this.add(this.Footer, { selector : '.siw-main-footer' });
+
+    // add callout for messages
+    this.messages = options.messages;
```
haishengwu-okta: nit: `this.options.messages`

src/v2/view-builder/internals/BaseView.js
```diff
@@ -34,6 +35,31 @@ export default View.extend({
...
+    if (this.messages && this.messages.value.length) {
+      this.showMessageCallout(options.messages.value[0].message, 'warning');
```
haishengwu-okta: `this.messages` was created but never used since you're referring `options.messages`.

PR #874, opened 2019-09-18
Do form post for success case

nikhilvenkatraman-okta:
Resolves: OKTA-250473

### DIFF
--- a/src/v2/view-builder/views/SuccessView.js
+++ b/src/v2/view-builder/views/SuccessView.js
@@ -1,14 +1,18 @@
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
-
+import Util from '../../../util/Util';
 const Body = BaseForm.extend({
   title () {
     return  'You will be redirected';
   },
   noButtonBar: true,
-  initialize (options) {
+  initialize () {
     BaseForm.prototype.initialize.apply(this, arguments);
-    options.appState.trigger('saveForm', options.model);
+    // TODO OKTA-250473
+    // Form post for success redirect
+    const url = this.options.appState.getCurrentViewState().href +
+      '?stateHandle=' + this.options.appState.get('currentState').stateHandle;
+    Util.redirectWithForm(url);
   },
 });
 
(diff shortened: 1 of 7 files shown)


src/v2/view-builder/views/SuccessView.js
```diff
@@ -1,14 +1,24 @@
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
+import $ from 'jquery';
```
haishengwu-okta: shall be able to import from `okta`.

src/v2/view-builder/views/SuccessView.js
```diff
@@ -1,14 +1,24 @@
...
+  initialize () {
     BaseForm.prototype.initialize.apply(this, arguments);
-    options.appState.trigger('saveForm', options.model);
+    // TODO OKTA-250473
+    // Form post for success redirect
+    const data = {
+      stateHandle: this.options.appState.get('currentState').stateHandle
+    };
+    return $.post({
+      url: '/idp/idx/login/token/redirect',
```
haishengwu-okta: hmm, sorry I forgot we didn't expose the remediation form details.. hard-code the URL is bad..
how about hack in `actionsTransformer.js`?

nikhilvenkatraman-okta: yeah, I just added this since we don't have access to the post url here. I'll try to move this elsewhere

nikhilvenkatraman-okta: How about we don't omit the `href` attribute in the responseTransformer, so that it can be accessed via currentState?

src/v2/view-builder/views/SuccessView.js
```diff
@@ -1,14 +1,18 @@
...
   },
   noButtonBar: true,
-  initialize (options) {
+  initialize () {
     BaseForm.prototype.initialize.apply(this, arguments);
-    options.appState.trigger('saveForm', options.model);
+    // TODO OKTA-250473
+    // Form post for success redirect
+    const url = this.options.appState.getCurrentViewState().href +
+    '?stateHandle=' + this.options.appState.get('currentState').stateHandle;
```
haishengwu-okta: indent is off.

PR #875, opened 2019-09-20
feat: adds okta verify number challenge

magizh-okta:
* If api responds with number challenge during okta verify challenge, UI displays number challenge.

* Updates lastAuthTransaction during polling. (needs authn changes)

https://okta.app.box.com/s/yhybly12s4jne89fsoc41rgkqxpmd2g8

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "@commitlint/cli": "^8.1.0",
     "@commitlint/config-conventional": "^8.1.0",
     "@commitlint/travis-cli": "^8.1.0",
-    "@okta/okta-auth-js": "~2.6.3",
+    "@okta/okta-auth-js": "2.9.0-alpha.g0a3ff20",
     "@sindresorhus/to-milliseconds": "^1.0.0",
     "autoprefixer": "^9.6.1",
     "axe-core": "^3.3.1",
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "@commitlint/cli": "^8.1.0",
     "@commitlint/config-conventional": "^8.1.0",
     "@commitlint/travis-cli": "^8.1.0",
-    "@okta/okta-auth-js": "2.9.0-alpha.g0a3ff20",
+    "@okta/okta-auth-js": "2.9.0",
     "@sindresorhus/to-milliseconds": "^1.0.0",
     "autoprefixer": "^9.6.1",
     "axe-core": "^3.3.1",
--- a/packages/@okta/i18n/dist/properties/login.properties
+++ b/packages/@okta/i18n/dist/properties/login.properties
@@ -484,7 +484,7 @@ oktaverify.sent = Push sent!
 oktaverify.rejected = You have chosen to reject this login.
 oktaverify.timeout = Your push notification has expired.
 oktaverify.warning = Haven't received a push notification yet? Try opening the Okta Verify App on your phone.
-oktaverify.numberchallenge.instruction = On your phone tap, <span class="challenge-number">{0}</span> on Okta Verify prompt to continue.
+oktaverify.numberchallenge.instruction = On your phone, tap <span class="challenge-number">{0}</span> on Okta Verify prompt to continue.
 oktaverify.numberchallenge.explain = This extra step helps us make sure it's really you signing in.
 
 # Username & Password
--- a/packages/@okta/i18n/dist/properties/login.properties
+++ b/packages/@okta/i18n/dist/properties/login.properties
@@ -484,6 +484,8 @@ oktaverify.sent = Push sent!
 oktaverify.rejected = You have chosen to reject this login.
 oktaverify.timeout = Your push notification has expired.
 oktaverify.warning = Haven't received a push notification yet? Try opening the Okta Verify App on your phone.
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -65,7 +65,7 @@
     "@commitlint/cli": "^8.1.0",
     "@commitlint/config-conventional": "^8.1.0",
     "@commitlint/travis-cli": "^8.1.0",
-    "@okta/okta-auth-js": "~2.6.3",
+    "@okta/okta-auth-js": "2.9.0-alpha.g0a3ff20",
```
mauriciocastillosilva-okta: reminder to update to a release version

packages/@okta/i18n/dist/properties/login.properties
```diff
@@ -484,6 +484,8 @@ oktaverify.sent = Push sent!
 oktaverify.rejected = You have chosen to reject this login.
 oktaverify.timeout = Your push notification has expired.
 oktaverify.warning = Haven't received a push notification yet? Try opening the Okta Verify App on your phone.
+oktaverify.numberchallenge.instruction = On your phone tap, <span class="challenge-number">{0}</span> on Okta Verify prompt to continue.
```
mauriciocastillosilva-okta: there seems to be an extra comma in this sentence. It should be `On your phone tap ## on Okta Verify...`, or if you need a comma, it should be after phone `On your phone, tap ## on Okta Verify...`

src/MfaVerifyController.js
```diff
@@ -123,6 +124,20 @@ function (Okta, BaseLoginController, TOTPForm, YubikeyForm, SecurityQuestionForm
         }
       }
 
+      this.listenTo(this.options.appState, 'change:isWaitingForNumberChallenge',
+        function (state, isWaitingForNumberChallenge) {
+          if (isWaitingForNumberChallenge || this.options.appState.get('lastAuthResponse').status === 'SUCCESS') {
```
mauriciocastillosilva-okta: a couple of questions on this line
1. Is this called everytime we are setting `lastAuthResponse`? or only when the value actually changes from false to true? Reason I am asking is that derived properties are calculated every time, so I want to double check when this is called (could be everytime since the property is recalculated)
2. I see that this also checks for SUCCESS, does that mean that, without number challenge, when we approve a push notification, we will hide the elements? If so, is that needed?

magizh-okta: 1) No. The function is executed only when the value of isWaitingForNumberChallenge changes. ie, if polling doesn't return number challenge the value of isWaitingForNumberChallenge remains false and the function never gets executed. 

2)  The success check will kick in only in number challenge flow and we wont hide other elements in non number challenge flow in success state. 

this is how isWaitingForNumberChallenge  changes in different state 
a) primary auth - false 
b) mfa required - false 
c) mfa challenge  - false 
d) mfa challenge with number challenge data - true
e) success - false ( this is the state that I handle here)

src/models/Factor.js
```diff
@@ -325,6 +328,7 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
                   };
                 }
                 return trans.poll(options).then(function (trans) {
+                  self.options.appState.set('lastAuthResponse', trans.data);
```
gowthamidommety-okta: Why was this added?

src/MfaVerifyController.js
```diff
@@ -22,10 +22,11 @@ define([
   'views/mfa-verify/PushForm',
   'views/mfa-verify/PasswordForm',
   'views/mfa-verify/InlineTOTPForm',
+  'views/mfa-verify/NumberChallengeView',
```
gowthamidommety-okta: Do we need this in this file?

src/views/mfa-verify/PushForm.js
```diff
@@ -110,7 +130,9 @@ define(['okta', 'util/CookieUtil', 'util/Util'], function (Okta, CookieUtil, Uti
         });
         this.trigger('save', this.model);
         warningTimeout = Util.callAfterTimeout(_.bind(function () {
-          this.showWarning(Okta.loc('oktaverify.warning', 'login'));
+          if (!this.options.appState.get('isWaitingForNumberChallenge')) {
```
gowthamidommety-okta: A UX question - do we need a similar warning for number challenge?

src/MfaVerifyController.js
magizh-okta: actually no will remove it. my bad!

src/models/Factor.js
magizh-okta: This is after the last poll -> success. The callback doesn't get  called in this case.

src/MfaVerifyController.js
mauriciocastillosilva-okta: 1. thanks
2. Got it. So this is a success that will follow showing the number challenge, not any other flow. Makes sense
👍

src/views/mfa-verify/PushForm.js
magizh-okta: discussed with Steven and decided to hold of on any custom warning messages for number challenge as current text in the UI is descriptive enough.

src/models/Factor.js
gowthamidommety-okta: Is this needed to hide the number challenge?

magizh-okta:
Taken care in path version https://github.com/okta/okta-signin-widget/pull/884/

Issue #760, opened 2019-07-18
custom authorization server not working in newly exposed authClient


BassplayerPatrick:
### I'm submitting a
  - [X] bug report
  - [ ] feature request

### Background info

I have the okta-signin-widget running in my self-hosted login-page. I have configured a custom authorization server to support login to my applications using the authorization code flow and hiding the okta dashboard from the users.
Here's my code to support the single sign-on experience using the widget:
```javascript
        this.oktaSignIn = this.okta.getWidget(config);
        console.log('##### widget initialized #####');
        this.oktaSignIn.authClient.session.get()
          .then(session => {
          console.log('Response session status = ' + session.status);
            if (session.status === 'ACTIVE') {
              this.oktaSignIn.authClient.token.getWithRedirect({
                sessionToken: session.sessionToken,
                responseType: 'code',
                scopes: ['openid', 'profile', 'email'],
                state: config.state,
                nonce: config.nonce
              });
            } else {
              console.log('##### Inactive session. Render login widget #####');
              this.oktaSignIn.renderEl({el: '#okta-login-container'});
            }
        }).catch(error => {
            console.log('Response error = ' + error);
            console.log('##### Render login widget #####');
            this.oktaSignIn.renderEl({el: '#okta-login-container'});
        });
```
config looks like this:
```javascript
      const config = {
        baseUrl: environment.oktaBaseURL,
        clientId: options.clientId,
        redirectUri: options.redirectUri,
        recoveryToken: options.recoveryToken,
        authParams: {
          display: 'page',
          responseType: options.responseType,
          issuer: environment.issuer,
          scopes: ['openid', 'profile'],
          state: options.state,
          nonce: options.nonce
        }
      };
```
The values in options are extracted from the url query-parameters
- environment.oktaBaseURL = https://< my-org >.oktapreview.com
- environment.issuer = authorizationServerId

### Expected behavior

- When a user browses to an applicatiion, he will be redirected to the login-page because he has no valid access- or id-token yet.
- The user logs in succesfully, and is redirected back to the application with an authorizationcode (from the configured custom authorization server) in the url query parameters
- The application exchanges this code for an access- and/or id-token and the user authentication process is completed
- Then the user navigates to an other application which doesn't know the user is already logged in and gets redirected to the loginpage again
- The loginpage now detects an already active (okta-) session and retrieves a new authorization code from the configured custom authorization server which is sent in the url query parameters during the redirect back to the second application
[...]

PR #882, opened 2019-09-26
fix: configure authClient using "authParams" OKTA-240327

aarongranick-okta:
See github issue for more details: https://github.com/okta/okta-signin-widget/issues/760

Customers are expecting to be able to use the `authClient` instance from the SIW to do other operations. However, this instance is incompletely configured. Currently only a minimum of parameters are passed to `OktaAuthJS` at construction and the other "authParams" are cherry picked and passed to the `authClient` as needed.

With this change, the entire `authParams` object will be passed to `OktaAuthJS`. This has no negative effect on `OktaAuthJS` or the SIW but it will allow other operations such as `getWithRedirect` to be called without requiring extra parameters. `OktaAuthJS` will generally use, by default, the configuration passed in at construction.

Also adds some test infrastructure updates/housekeeping that I found necessary for local testing:
- Adds support for `testenv` file (deprecates `.env`) - this matches convention in our other repos such as `okta-ui`
- Enhances "basic" OIDC test app (which can be run with `yarn start:app` with a textarea to edit config. This allows manual testing of scenarios using arbitrary config

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,5 @@ packages/@okta/courage-dist/*.map
 /dist
 .yalc
 .env
+testenv
 packages/@okta/i18n/dist/json/
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "test:dev": "grunt assets && karma start",
     "test:coverage": "yarn test --reporters=progress,coverage",
     "test": "yarn test:dev --single-run",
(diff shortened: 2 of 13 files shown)


aarongranick-okta:
> LGTM, but I'd love to have a better idea of the impact to users that had their own workarounds - does this break them?

I believe there will be no impact. You can still create a second instance. You can still pass parameters when you call the individual methods. Exposing the `authClient` was only added in version 3 so I think it is a fairly new feature. Before then everyone created their own authClient.

PR #888, opened 2019-09-27
fix: hide recovery password link when there is not recovery ion form

haishengwu-okta:
OKTA-252023
- https://okta.box.com/s/y4qk35m5jj0f16mc1ohmcviw9fd2f3mj


src/v2/view-builder/views/password/RequiredFactorPasswordView.js
```diff
@@ -3,6 +3,8 @@ import BaseForm from '../../internals/BaseForm';
 import BaseFooter from '../../internals/BaseFooter';
 import BaseFactorView from '../shared/BaseFactorView';
 
+const recoveryLinkAction = 'factor.recovery';
```
nikhilvenkatraman-okta: this should be `factor.recover`

PR #893, opened 2019-09-30
feat: add support for factor icons

nikhilvenkatraman-okta:
Resolves: OKTA-251871

### DIFF
--- a/src/v2/ion/uiSchemaTransformer.js
+++ b/src/v2/ion/uiSchemaTransformer.js
@@ -70,6 +70,17 @@ const createUISchema = (remediationValue = []) => {
   });
 };
 
+const addFactorType = (options, factors) => {
+  _.each(options, function (optionItem) {
+    const factorValue = optionItem.value;
+    const factor = factors.find(function (item) {
+      return (item.factorProfileId === factorValue
+        || item.factorId === factorValue);
+    });
+    optionItem.factorType = factor.factorType;
+  });
+};
+
 /**
  *
  * @param {AuthResult} transformedResp (after `actionsTransformer`)
@@ -77,8 +88,12 @@ const createUISchema = (remediationValue = []) => {
 const insertUISchema = (transformedResp) => {
   if (transformedResp.currentState) {
     const remediation = transformedResp.currentState.remediation || [];
+    const factors = transformedResp.factors && transformedResp.factors.value || [];
     transformedResp.currentState.remediation = remediation.map(obj => {
       obj.uiSchema = createUISchema(obj.value);
+      if (obj.value && obj.value[0] && obj.value[0].options) {
+        addFactorType(obj.value[0].options, factors);
+      }
       return obj;
     });
   }
--- a/src/v2/ion/uiSchemaTransformer.js
+++ b/src/v2/ion/uiSchemaTransformer.js
@@ -14,6 +14,21 @@
  * Create UI Schema into remedation object base on remediation values
  */
 import { _ } from 'okta';
+
+/**
+ * Adds factorType metadata to each select-factor option item using factors array
+ */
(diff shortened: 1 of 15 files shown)


src/v2/ion/uiSchemaTransformer.js
```diff
@@ -70,15 +70,30 @@ const createUISchema = (remediationValue = []) => {
...
+};
+
 /**
  *
  * @param {AuthResult} transformedResp (after `actionsTransformer`)
  */
 const insertUISchema = (transformedResp) => {
   if (transformedResp.currentState) {
     const remediation = transformedResp.currentState.remediation || [];
+    const factors = transformedResp.currentState.factors && transformedResp.currentState.factors.value;
```
haishengwu-okta: hmm, I thought it would come from `transformedResp.factors`?

src/v2/ion/uiSchemaTransformer.js
```diff
@@ -70,15 +70,30 @@ const createUISchema = (remediationValue = []) => {
...
  *
  * @param {AuthResult} transformedResp (after `actionsTransformer`)
  */
 const insertUISchema = (transformedResp) => {
   if (transformedResp.currentState) {
     const remediation = transformedResp.currentState.remediation || [];
+    const factors = transformedResp.factors && transformedResp.factors.value || [];
     transformedResp.currentState.remediation = remediation.map(obj => {
       obj.uiSchema = createUISchema(obj.value);
+      if (obj.value && obj.value[0] && obj.value[0].options) {
```
haishengwu-okta: `value[0]` is too specific. why not adding `factorType` (invokes addFactorType) around line 63?
```
    if (ionFormField.name === 'factorId' ||
      ionFormField.name === 'factorProfileId') {
      uiSchema.type = 'factorType';
    }
```

nikhilvenkatraman-okta: yeah, since `createUISchema` method is dealing with the remediation object to create the UI Schema I was trying to avoid sending in the factors object into that just for this.

nikhilvenkatraman-okta: I could just loop through each of the items in `obj.value`

``` 
_.each(obj.value, function (optionsObject) {
  if (optionsObject.options) {
    addFactorTypeToOptions(optionsObject.options, factors);
  }
});
  
```

haishengwu-okta: > since createUISchema method is dealing with the remediation object to create the UI Schema

It's true before but not now which says 'in order to create ui schema, we need remediation and factors`

src/v2/ion/uiSchemaTransformer.js
```diff
@@ -61,6 +77,7 @@ const createUISchema = (remediationValue = []) => {
     if (ionFormField.name === 'factorId' ||
       ionFormField.name === 'factorProfileId') {
       uiSchema.type = 'factorType';
+      addFactorTypeToOptions(ionFormField.options, factors);
```
haishengwu-okta: this is a bit tricky since looks like you change the (original) `options` in place hence it works.
I'd suggest to 
`uiSchema.options = createFactorTypeOptions(ionFormField.options, factors);`

test/unit/spec/v2/ion/uiSchemaTransformer_spec.js
```diff
@@ -44,7 +44,7 @@ describe('v2/ion/uiSchemaTransformer', function () {
     });
   });
 
-  it('converts factor enroll options', () => {
+  fit('converts factor enroll options', () => {
```
haishengwu-okta: rollback to `it`?

nikhilvenkatraman-okta: woops

PR #909, opened 2019-10-09
fix: Abort webauthn prompt when switching factors

magizh-okta:
* In some browsers the webauthn prompt is subtle and users often go back to choose another factor.
When users do this the browser prompt is still present and it is an UX issue. Also when they come back to use webauthn again it errors out.

* This fix is to kill the webauthn prompt when switching factors.

* introduced a new error to prevent displaying an error message when error is due to abort via code since factor UI was switched.

**Enroll and verify with switching works fine** 
enroll-abort
verify-abort

** verify chrome ** 
verify-abort-cc

** enroll chrome (back is triggered via js in console) user cannot switch without cancel anyway ** 
enrol-abort-cc

**BEFORE** 
recording is for u2f but it is the same issue for webauthn https://okta.app.box.com/s/xg9kp9uaw2dsjks4n2manoe0iz2spf3f

resolves: OKTA-249465


src/EnrollWebauthnController.js
```diff
@@ -88,9 +93,14 @@ function (Okta, Errors, FormType, FormController, CryptoUtil, webauthn, Footer,
               })
               .fail(function (error) {
                 self.trigger('errors:clear');
-                throw new Errors.WebAuthnError({
-                  xhr: {responseJSON: {errorSummary: error.message}}
-                });
+                // Do not display if it is abort error triggered by code when switching.
+                if (error && error.code === 20) {
```
magizh-okta: @gowthamidommety-okta @vincentvoong-okta  we can actually determine what the error message is in each case. I got it wrong previously as the console statements printed an object differently. FF as a string and chrome as object.

src/EnrollWebauthnController.js
```diff
@@ -79,7 +79,12 @@ function (Okta, Errors, FormType, FormController, CryptoUtil, webauthn, Footer,
               },
               excludeCredentials: getExcludeCredentials(activation.excludeCredentials)
             });
-            return new Q(navigator.credentials.create({publicKey: options}))
+            var webauthnAbortController = new AbortController();
```
haishengwu-okta: `AbortController` seems not compatible in older browser. would there be any regression concern?
https://www.caniuse.com/#feat=abortcontroller

src/EnrollWebauthnController.js
```diff
@@ -79,7 +79,12 @@ function (Okta, Errors, FormType, FormController, CryptoUtil, webauthn, Footer,
...
+            this.appState.set('webauthnAbortController', webauthnAbortController);
```
mauriciocastillosilva-okta: do this really belong to the `appState`? Can we instead abort in the Footer when the user clicks "Back to Factors"? It seems to me this is very specific to this factor

src/EnrollWebauthnController.js
```diff
@@ -88,9 +93,14 @@ function (Okta, Errors, FormType, FormController, CryptoUtil, webauthn, Footer,
               })
               .fail(function (error) {
                 self.trigger('errors:clear');
-                throw new Errors.WebAuthnError({
-                  xhr: {responseJSON: {errorSummary: error.message}}
-                });
+                // Do not display if it is abort error triggered by code when switching.
+                if (error && error.code === 20) {
+                  throw new Errors.WebauthnAbortError();
```
mauriciocastillosilva-okta: its not really an error, is it? Can we just not do anything here? Or in other words, just handle this error when the code is not 20?

src/EnrollChoicesController.js
```diff
@@ -87,6 +87,10 @@ function (Okta, FormController, Enums, RouterUtil, FactorList,
       },
 
       initialize: function (options) {
+        if (options.appState.get('webauthnAbortController')) {
```
mauriciocastillosilva-okta: I think this should be handled in a similar way to how we cancel polling when we switch from Okta Verify to another factor. There is a `factorSwitched` event that is fired in [FactorDropdownOptions#L31](https://github.com/okta/okta-signin-widget/blob/master/src/views/mfa-verify/dropdown/FactorsDropDownOptions.js#L31) that we listen to in [Factor#315](https://github.com/okta/okta-signin-widget/blob/master/src/models/Factor.js#L315), so we can stop polling. Sounds like we can do something similar?

src/EnrollWebauthnController.js
magizh-okta: @haishengwu-okta  older meaning IE 11 ? webauthn  itself wont work so we dont have to worry about it I guess. ie) we will safely fail before we use the abortController.

src/EnrollWebauthnController.js
magizh-okta: Since we use BaseLoginModel here there is a fail case in base login model and it expects some know error else it fails. 

we already have similar cases 
https://github.com/okta/okta-signin-widget/blob/master/src/models/BaseLoginModel.js#L37

src/EnrollChoicesController.js
magizh-okta: nice! I reused this and moved away from appstate as well.

haishengwu-okta: curious why you cannot extend `src/models/Factor.js` as Mau mentioned but have to listen to a new event when click back?

magizh-okta: @haishengwu-okta  I am reusing the event in verify flow. But during the enroll when back is clicked I don't see any event being triggered.

src/EnrollWebauthnController.js
magizh-okta: Oops taking this back! Firefox triggers a different code vs chrome :/ /cc @mauriciocastillosilva-okta  trying to fix this now

src/views/mfa-verify/dropdown/FactorsDropDownOptions.js
```diff
@@ -27,7 +27,7 @@ define(['okta', 'util/RouterUtil'], function (Okta, RouterUtil) {
     }
     var url = RouterUtil.createVerifyUrl(model.get('provider'), factorType, factorIndex);
     var self = this;
-
+
```
mauriciocastillosilva-okta: please revert this change

src/EnrollChoicesController.js
haishengwu-okta: I meant `backToFactors`. looks like enroll and challenge flow need different event?

src/EnrollWebauthnController.js
mauriciocastillosilva-okta: ah ok. Thanks for the update. Let me know what you find out, I'm curious about this

test/unit/spec/EnrollWebauthn_spec.js
```diff
@@ -88,17 +88,19 @@ function (Okta,
...
       spyOn(webauthn, 'isNewApiAvailable').and.returnValue(true);
       spyOn(navigator.credentials, 'create').and.callFake(function () {
         var deferred = Q.defer();
-        deferred.resolve({
-          response: {
-            attestationObject: CryptoUtil.strToBin(testAttestationObject),
-            clientDataJSON: CryptoUtil.strToBin(testClientData)
-          }
-        });
+        if (!doNotResolve) {
```
mauriciocastillosilva-okta: 2 comments here:
1. can we make this not a "double negative" -> if NOT doNotResolve... maybe the arg can be named something like `resolvePromise` or `successfullyResolvePromise`?
2. If it does not resolve, then we are timing out? Not sure how that is handled in the test...

test/unit/spec/VerifyWebauthn_spec.js
```diff
@@ -14,7 +14,7 @@ define([
   'sandbox',
   'util/webauthn',
   'helpers/xhr/MFA_REQUIRED_allFactors',
-  'helpers/xhr/MFA_REQUIRED_multipleFactors',
+  'helpers/xhr/MFA_REQUIRED_multipleWebauthn_question',
```
mauriciocastillosilva-okta: can we also rename the variable we use? It's just easier in the future to know what the mock response is if the name matches the actual mock response

test/unit/spec/VerifyWebauthn_spec.js
```diff
@@ -53,6 +53,14 @@ function (Okta,
     Util.mockRouterNavigate(router);
     return router;
   }
+  
+  function clickFactorInDropdown (test, factorName) {
+    if (factorName === 'WEBAUTHN') {
+      test.beacon.getOptionsLinks().eq(0).click();
+    } else {
```
mauriciocastillosilva-okta: nit: check "QUESTION", so we know it is that factor we are selecting

src/EnrollChoicesController.js
mauriciocastillosilva-okta: I think this is right, since those 2 events are different:
- `factorSwitched` is the event we have when users select a new factor in the beacon dropdown. This is only in verify flows
- `backToFactors` is the new event created in this PR, this is only in enroll flows (there is no beacon dropdown in enroll flows)

test/unit/spec/EnrollWebauthn_spec.js
magizh-okta: we will be in waiting state. More like faking the state where we see the webauthn prompt but user has not yet authenticated via touchid/u2f key

src/EnrollChoicesController.js
magizh-okta: yeah in verify we switchFactors and enroll we can go back to factorlist.

magizh-okta:
@mauriciocastillosilva-okta  I addressed comments and added some cleanup code + handled signOut case too. 

singout

src/EnrollWebauthnController.js
magizh-okta: in chrome the error code was 20 for programatic abort and 0 for user abort. But in firefox it is always 20. ( which is per spec) so I don't want to rely on this in consistent behavior.

test/unit/spec/EnrollWebauthn_spec.js
```diff
@@ -174,13 +176,34 @@ function (Okta,
           Expect.isVisible(test.form.submitButton());
         });
       });
+      
+      itp('calls aborts on appstate when switching to factor list after clicking enroll', function () {
```
mauriciocastillosilva-okta: singular abort: `calls abort on appsate`...

PR #918, opened 2019-10-11
Enable testcafe travis

haishengwu-okta:
cc @mauriciocastillosilva-okta @magizh-okta

### DIFF
--- a/webpack.playground.config.js
+++ b/webpack.playground.config.js
@@ -11,14 +11,16 @@ const WIDGET_RC = '.widgetrc';
 
 let widgetRc = {
   widgetOptions: {
-    baseUrl: 'https://{yourOktaDomain}',
+    baseUrl: 'http://localhost:3000',
     logo: '/img/logo_widgico.png',
     logoText: 'Windico',
     features: {
       router: true,
       rememberMe: true,
+      securityImage: true,
       multiOptionalFactorEnroll: true
     },
+    stateToken: 'dummy-state-token-wrc',
     // Host the assets (i.e. jsonp files) locally
     assets: {
       baseUrl: '/'
--- a/webpack.playground.config.js
+++ b/webpack.playground.config.js
@@ -17,7 +17,6 @@ let widgetRc = {
     features: {
       router: true,
       rememberMe: true,
-      securityImage: true,
       multiOptionalFactorEnroll: true
     },
     stateToken: 'dummy-state-token-wrc',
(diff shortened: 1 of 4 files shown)


webpack.playground.config.js
```diff
@@ -11,14 +11,16 @@ const WIDGET_RC = '.widgetrc';
...
 let widgetRc = {
   widgetOptions: {
-    baseUrl: 'https://{yourOktaDomain}',
+    baseUrl: 'http://localhost:3000',
     logo: '/img/logo_widgico.png',
     logoText: 'Windico',
     features: {
       router: true,
       rememberMe: true,
+      securityImage: true,
```
mauriciocastillosilva-okta: security image has not worked for me before locally... not sure if that is fixed or maybe it was just me, but if it doesn't work easily for other people, should we then leave it disabled?

nikhilvenkatraman-okta: why do we need this to be true ?  securityImage does not work for new pipeline anyways

webpack.playground.config.js
```diff
@@ -11,14 +11,16 @@ const WIDGET_RC = '.widgetrc';
 
 let widgetRc = {
   widgetOptions: {
-    baseUrl: 'https://{yourOktaDomain}',
+    baseUrl: 'http://localhost:3000',
```
magizh-okta: Right now we are having tests for idx pipeline powered by mock server right ? And if no widgetrc is provided by default playground opens up with idx flow pointing to mock server right.

haishengwu-okta: correct.

webpack.playground.config.js
haishengwu-okta: - @mauriciocastillosilva-okta it does given we mocked `http://localhost:3000/login/getimage` request
- @nikhilvenkatraman-okta the `widgetrc` is not dedicated for OIE though. but I'll remove it from the default generator now.

webpack.playground.config.js
```diff
@@ -11,14 +11,15 @@ const WIDGET_RC = '.widgetrc';
...
     logo: '/img/logo_widgico.png',
     logoText: 'Windico',
     features: {
       router: true,
```
magizh-okta: Does these features make sense in idx flows ?

haishengwu-okta: probably not. the `widgetrc` is intend for both v1/v2 and we didn't clearly call out the difference.

PR #928, opened 2019-10-23
fix: fixes security image

petermiller-okta:
- I have a video of this in action, but I can't upload to box for some reason...PM me for it if you'd like to see it.

Resolves: OKTA-240039

@okta/idx 
@haishengwu-okta

### DIFF
--- a/src/models/AppState.js
+++ b/src/models/AppState.js
@@ -39,10 +39,6 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
   var securityImageUrlTpl = compile('{{baseUrl}}/login/getimage?username={{username}}');
 
   function getSecurityImage (baseUrl, username, deviceFingerprint) {
-    // Reserved characters in the username must be escaped before the query can be safely executed
-    username = encodeURIComponent(username);
-    var url = securityImageUrlTpl({ baseUrl: baseUrl, username: username });
-
     // When the username is empty, we want to show the default image.
     if (_.isEmpty(username) || _.isUndefined(username)) {
       return Q({
@@ -50,6 +46,10 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         'securityImageDescription': UNDEFINED_USER_IMAGE_DESCRIPTION
       });
     }
+    
+    // Reserved characters in the username must be escaped before the query can be safely executed
+    username = encodeURIComponent(username);
+    var url = securityImageUrlTpl({ baseUrl: baseUrl, username: username });
 
     var data = {
       url: url,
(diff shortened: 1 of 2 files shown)


petermiller-okta:
Unfortunately I am unable to test this, but what I was observing is we are calling `getSecurityImage` an additional time with the username parameter `undefined` which was being interpreted as the string `"undefined"` in the method itself. If the username is undefined, we can just skip trying to retrieve the image. This will most likely fix the issue we are seeing.

royalchan-okta:
looks fine to me but i'd double check with someone from UI

src/models/AppState.js
```diff
@@ -126,24 +126,26 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
...
-                self.settings.callGlobalError(new Errors.UnsupportedBrowserError(
-                  Okta.loc('error.enabled.cors')
-                ));
-              }
-              else {
-                throw jqXhr;
-              }
-            })
-            .done();
+          if (username) {
```
haishengwu-okta: when `username` is empty, it will not actually make API call https://github.com/okta/okta-signin-widget/blob/master/src/models/AppState.js#L47-L51

what's exact problem here?

haishengwu-okta: I guess the problem is at https://github.com/okta/okta-signin-widget/blob/master/src/models/AppState.js#L43

nikhilvenkatraman-okta:
@petermiller-okta Were you able to verify this locally ?

petermiller-okta:
@nikhilvenkatraman-okta yes, I was able to build this locally into my dev environment and the issue went away

nikhilvenkatraman-okta:
Can we look to see if we can add a test similar to https://github.com/okta/okta-signin-widget/blob/master/test/unit/spec/PrimaryAuth_spec.js#L1437, mainly testing that we show the unknown user image if username is empty ?

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1447,6 +1447,17 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
...
       });
+      itp('undefined username does not make API call', function () {
+        return setup({ features: { securityImage: true }})
+          .then(function (test) {
+            test.setNextResponse(resSecurityImage);
+            test.form.setUsername(undefined);
+            return waitForBeaconChange(test);
+          })
+          .then(function () {
+            expect($.ajax.calls.count()).toBe(0);
```
nikhilvenkatraman-okta: In this case we show the UNDEFINED_USER security image right ? Can we also assert that the beacon gets the right classname something like `expect(test.form.securityBeacon()[0].className).toMatch('new-user')`

petermiller-okta: @nikhilvenkatraman-okta updated.

PR #929, opened 2019-10-24
fix: add terminal view class and remove terminal-state data-se

jhe-okta:
OKTA-253713

### DIFF
--- a/src/v2/view-builder/views/TerminalView.js
+++ b/src/v2/view-builder/views/TerminalView.js
@@ -3,13 +3,15 @@ import BaseForm from '../internals/BaseForm';
 import BaseFooter from '../internals/BaseFooter';
 
 const Body = BaseForm.extend({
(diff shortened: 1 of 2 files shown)


src/v2/view-builder/views/TerminalView.js
```diff
@@ -3,13 +3,14 @@ import BaseForm from '../internals/BaseForm';
...
 const Body = BaseForm.extend({
-  attributes: { 'data-se': 'terminal-state' },
-
   title () {
     const msg = this.options.appState.get('terminal').message || {};
     return msg.message || 'You can close this window';
   },
   noButtonBar: true,
+  postRender () {
+    this.$el.addClass('terminal-state');
```
haishengwu-okta: shall we invoke `super.postRender`? although nothing happens at super now but to be consistent about how to override method.

PR #932, opened 2019-10-25
feat: implement PIV / CAC card

gowthamidommety-okta:
Resolves: OKTA-269030

piccac


assets/sass/modules/_btns.scss
```diff
@@ -102,6 +102,29 @@
...
+  border-radius: $border-radius-default;
+  box-sizing: border-box;
+
+  display: block;
+  font-size: $font-size;
+  line-height: 50px;
+  margin-top: 15px;
+  text-align: center;
+
+  &:last-of-type {
```
magizh-okta: is the intention here to have a margin bottom 25px; only for the last button (be it default/piv/social) 

if so we should add it to a diff class that is common to all three. Otherwise there will be a margin bottom added for each button since they are all unique.

src/VerifyPIVController.js
```diff
@@ -0,0 +1,103 @@
...
+        this.$('.o-form-button-bar').hide();
+      },
+      
+      _stopEnrollment: function () {
+        this.$('.okta-waiting-spinner').hide();
+        this.$('.o-form-button-bar').show();
+      },
+
+      postRender: function () {
+        _.defer(_.bind(function () {
```
magizh-okta: nit we can use arrow function here instead of bind.

src/VerifyPIVController.js
```diff
@@ -0,0 +1,103 @@
...
+    Model: {
+
+      save: function () {
+        this.trigger('request');
+        var self = this,
+            data = {
+              fromURI: this.settings.get('relayState')
+            },
+            pivButton = this.settings.get('piv');
+        return Q($.post({
```
magizh-okta: This post triggers the browser to prompt for cert ? I  was expecting to see us calling some browser api somewhere.

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -208,6 +208,18 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
     return setup(settings);
   }
 
+  function setupPIV (pivAuthentication, useDefaultText) {
+    var buttonText = (useDefaultText) ? '' : 'piv test text';
+    var settings = {
+      piv: {
```
magizh-okta: Is there a case where there can be a piv config without certAuthUrl ? Should we just not have the piv config if setup(false) is called.

magizh-okta: do we have a case where users can pass piv config without the url. shouldnt setup(false) have settings without piv object ?

src/VerifyPIVController.js
gowthamidommety-okta: Yea, the server needs to ask for the cert

test/unit/spec/PrimaryAuth_spec.js
gowthamidommety-okta: certAuthUrl is required for PIV.. it wont know the endpoint to call to prompt for cert otherwise.. I want to be able to test that no certAuthUrl = no piv button, so i wrote it this way

gowthamidommety-okta: Same comment as above

assets/sass/modules/_btns.scss
gowthamidommety-okta: It is a very rare case to have a piv button with social auth or custom buttons.. But the way it would render if all three types were present is:
1. social auth buttons
2. piv
3. custom buttons

1. & 3. can have multiple buttons of that type, hence last-of-type adds some extra margin at the bottom to indicate the end that type of button. I wrote the same for piv, though there will only ever be 1 piv button.. I could add this margin-bottom after the text-align above instead. Please let me know what you think

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -591,7 +591,10 @@ registration.error.userName.notUniqueWithinOrg=An account with that email alread
 
 # PIV Auth
 piv.card=PIV Card
+piv.cac.card=Sign in with PIV / CAC card
```
mauriciocastillosilva-okta: nit: we can group the `piv.cac.*` strings together right after this block of `piv.card.*`, just to have the related strings together, instead of all mixed up

src/VerifyPIVController.js
```diff
@@ -0,0 +1,125 @@
...
+  'views/shared/FooterWithBackLink'
+],
+function (Okta, FormController, FormType, FooterWithBackLink) {
+
+  var _ = Okta._,
+      $ = Okta.$;
+  var { Util } = Okta.internal.util;
+
+  return FormController.extend({
+    className: 'mfa-verify verify-piv',
```
mauriciocastillosilva-okta: Should we have a different class name, since this is not really mfa? `piv-cac-card verify-piv`? maybe we don't even need `verify-piv`?

src/VerifyPIVController.js
```diff
@@ -0,0 +1,125 @@
...
+    Model: {
+
+      save: async function () {
+        this.trigger('request');
+        var self = this,
+            data = {
+              fromURI: this.settings.get('relayState')
+            },
+            pivButton = this.settings.get('piv');
```
mauriciocastillosilva-okta: nit `pivConfig` instead?

src/VerifyPIVController.js
```diff
@@ -0,0 +1,125 @@
...
+        }
+      },
+
+      getCert: function (certAuthUrl) {
+        return $.get({
+          url: certAuthUrl,
+          xhrFields: {
+            withCredentials: true
+          },
+          beforeSend: function () {},
```
mauriciocastillosilva-okta: do we need this empty function? Are you overriding something?

src/VerifyPIVController.js
```diff
@@ -0,0 +1,125 @@
...
+            data = {
+              fromURI: this.settings.get('relayState')
+            },
+            pivButton = this.settings.get('piv');
+
+        try {
+          await this.getCert(pivButton.certAuthUrl);
+          const res = await this.postCert(pivButton.certAuthUrl, data);
+          Util.redirect(res.redirectUrl);
+        } catch (err) {
```
mauriciocastillosilva-okta: will this `catch` block handle all errors that could happen in `getCert` and `postCert`? As in server is down, maybe  get call times out, etc

src/views/primary-auth/CustomButtons.js
```diff
@@ -34,13 +34,18 @@ define([
...
           socialIdpButtons = this.settings.get('configuredSocialIdps'),
+          pivButton = this.settings.get('piv'),
           customButtons = this.settings.get('customButtons'),
           divider = dividerTpl({text: Okta.loc('socialauth.divider.text', 'login')});
 
       _.each(socialIdpButtons, function (button) {
         children.push(this._createSocialIdpButton(button));
       }, this);
 
+      if (this.settings.get('hasPivCard')) {
```
mauriciocastillosilva-okta: wondering if PIV should be above social idps? We do the PIV sign in, instead of social idps that redirect to other page... can you double check with design or pm?

test/unit/helpers/xhr/PIV_CHALLENGE.js
```diff
@@ -0,0 +1,7 @@
+define({
```
mauriciocastillosilva-okta: should this file be named just `PIV_POST`? we are not in an mfa_challenge status or similar, are we?

src/VerifyPIVController.js
```diff
@@ -0,0 +1,125 @@
...
+      postCert: function (certAuthUrl, data) {
+        return $.post({
+          url: certAuthUrl,
+          xhrFields: {
+            withCredentials: true
+          },
+          data: JSON.stringify(data),
+          contentType: 'text/plain',
+          beforeSend: function () {},
+        }).then((result) => {
```
magizh-okta: do we need to return result explicitly ? I think it automatically does it.

src/VerifyPIVController.js
magizh-okta: It should. This test checks that I guess  
https://github.com/okta/okta-signin-widget/pull/932/files#diff-8717f20fe0dfff66217b2c9a59a60190R100

src/VerifyPIVController.js
gowthamidommety-okta: yea, im making this a simple request by not setting this header in the jquery-wrapper - https://github.com/okta/okta-ui/blob/master/packages/courage/src/util/jquery-wrapper.js#L6

src/views/primary-auth/CustomButtons.js
gowthamidommety-okta: @vincentvoong-okta do you have a preference? PIV and Social buttons together is very rare.. but just in case.

vincentvoong-okta: @mauriciocastillosilva-okta @gowthamidommety-okta 
I don't know enough about this edge case to have a very informed decision but I'd imaging if a user has the choice between PIV/CAC and social auth, PIV/CAC should have priority.

gowthamidommety-okta: thanks Vince! Will move it up

src/VerifyPIVController.js
gowthamidommety-okta: Yea verified manually that it catches errors on get as well

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -2826,6 +2838,126 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
...
+      return setupPIV(true).then(function (test){
+        expect(test.form.pivButton().text()).toEqual('piv test text');
+      });
+    });
+    itp('sets class with property passed', function () {
+      return setupPIV(true).then(function (test){
+        expect(test.form.pivButton().hasClass('piv-test-class')).toBe(true);
+      });
+    });
+    itp('navigates to piv verify view when button is clicked', function () {
```
mauriciocastillosilva-okta: nit: "navigates to piv **login flow** when button is clicked"

test/unit/spec/VerifyPIV_spec.js
```diff
@@ -0,0 +1,134 @@
+/* eslint max-params: [2, 18], max-statements: 0 */
```
mauriciocastillosilva-okta: nit: this file can be `PIVLogin_spec.js` or maybe `LoginWithPIV_spec.js`

src/VerifyPIVController.js
```diff
@@ -0,0 +1,131 @@
...
+          await this.getCert(pivConfig.certAuthUrl);
+          const res = await this.postCert(pivConfig.certAuthUrl, data);
+          Util.redirect(res.redirectUrl);
+        } catch (err) {
+          self.trigger('error', self, err);
+        }
+      },
+
+      getCert: function (certAuthUrl) {
+        return $.get({
```
haishengwu-okta: nit:
instead of using jquery, maybe we could try to use `cross-fetch` which is used by auth-js underneath? (separated PR if you have time)

gowthamidommety-okta: Spoke to Haisheng offline. Will need to move httpClient.js from v2 src to shared. Also will need to investigate behavior on different browsers (probably provide capability to override / exclude some headers). Will do this as a separate PR in Dec.

test/unit/spec/VerifyPIV_spec.js
gowthamidommety-okta: Actually would it be ok to keep this name? Just because it is in sync with VerifyPIVController.js, and follows the pattern for other spec files

src/VerifyPIVController.js
```diff
@@ -0,0 +1,131 @@
...
+        this.$('.o-form-button-bar').hide();
+      },
+      
+      _stopEnrollment: function () {
+        this.$('.okta-waiting-spinner').hide();
+        this.$('.o-form-button-bar').show();
+      },
+
+      postRender: function () {
+        _.defer(() => {
```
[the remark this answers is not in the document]
gowthamidommety-okta: I tested without it as well, and it does not make too much of a difference. I followed what is done for webauthn, and I think it is nice to have when the browser prompts for cert. Please let me know what you think.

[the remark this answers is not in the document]
haishengwu-okta: > I think it is nice to have when the browser prompts for cert
wouldn't it prompt if we just do `this.model.save()`? In general I wouldn't prefer `defer`  if it doesn't make difference.

it's not a blocking question though.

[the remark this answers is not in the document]
gowthamidommety-okta: I noticed that for some browsers, the ui doesn't render completely before the cert prompt. I will change this though, and test again on all browsers to see the effect. 👍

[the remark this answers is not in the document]
haishengwu-okta: Cool. you can deal with in another PR if you want to.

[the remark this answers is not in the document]
gowthamidommety-okta: Just got a chance to test this. It turns out that it needs that defer to correctly show / hide the spinner and button. That must be why u2f and webauthn do the same. Will leave this as is for now.

PR #933, opened 2019-10-26
fix: replace tick() with wait conditions OKTA-257628

aarongranick-okta:
Fixes some "flakey" tests by replacing uses of `tick()` with specific wait conditions
Corrects 2 instances of live network traffic being sent, because tests ended too early, while an operation was in progress. (LoginRouter)

### DIFF
--- a/test/unit/spec/LoginRouter_spec.js
+++ b/test/unit/spec/LoginRouter_spec.js
@@ -157,7 +157,9 @@ function (Okta, Q, Logger, Errors, BrowserFeatures, WidgetUtil, Bundles, config,
           test.setNextResponse(next);
           form.setAnswer('wrong');
           form.submit();
-          return tick(test);
+          var redirectFlow = settings && (settings['authParams.responseType'] !== undefined || settings['authParams.display'] === 'page');
+          var spy = redirectFlow ? test.ac.token.getWithRedirect._setLocation : test.ac.token.getWithoutPrompt;
+          return Expect.waitForSpyCall(spy, test);
         });
     }
 
@@ -1193,10 +1195,9 @@ function (Okta, Q, Logger, Errors, BrowserFeatures, WidgetUtil, Bundles, config,
                 state: OIDC_STATE
               }
             });
-            return tick();
-          })
-          .then(function () {
-            expect($sandbox.find('#' + OIDC_IFRAME_ID).length).toBe(0);
+            return Expect.wait(function () {
+              return $sandbox.find('#' + OIDC_IFRAME_ID).length === 0;
+            });
           });
       });
       itp('invokes the success function with idToken and user data when the iframe returns with data', function () {
@@ -1206,6 +1207,8 @@ function (Okta, Q, Logger, Errors, BrowserFeatures, WidgetUtil, Bundles, config,
           .then(function () {
             expect(window.addEventListener).toHaveBeenCalled();
             var args = window.addEventListener.calls.argsFor(0);
+            expect(args[0]).toBe('message');
+            // Simulate callback from an iframe
             var callback = args[1];
             callback.call(null, {
               origin: 'https://foo.com',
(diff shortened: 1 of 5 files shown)


test/unit/spec/LoginRouter_spec.js
```diff
@@ -1045,7 +1047,7 @@ function (Okta, Q, Logger, Errors, BrowserFeatures, WidgetUtil, Bundles, config,
           return Expect.waitForSpyCall(spy)
             .then(function () {
               expect(spy.calls.count()).toBe(1);
-              expectAuthorizeUrl(spy.calls.argsFor(0)[0], options);
+              options && expectAuthorizeUrl(spy.calls.argsFor(0)[0], options);
```
haishengwu-okta: curious what's been called when `options` is null given `spy.calls` is always 1.

test/unit/spec/LoginRouter_spec.js
```diff
@@ -1534,41 +1537,60 @@ function (Okta, Q, Logger, Errors, BrowserFeatures, WidgetUtil, Bundles, config,
...
             expect(form.isPrimaryAuth()).toBe(true);
             form.setUsername('testuser');
             form.setPassword('testpassword');
             form.submit();
             expect($.ajax.calls.mostRecent().args[0].headers['Accept-Language']).toBe('en');
+
+            // Wait for login success
+            test.setNextResponse(resSuccess);
+            return Expect.wait(function () {
+              return success.calls.count() > 0;
```
haishengwu-okta: is `success.calls.count()` always 1? (just trying to be a little more explicit)

aarongranick-okta: yes there will be only 1 success call

test/unit/spec/LoginRouter_spec.js
aarongranick-okta: This was left from an earlier revision where I wanted to wait for the call but not verify the parameters. I can remove it

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1446,7 +1479,11 @@ function (Okta,
           expect(test.button.trimmedText()).toEqual('Call');
           expect(test.form.answerField().val()).toEqual('123456');
           test.form.makeCall().click();
-          return tick().then(function () {
+          return Expect.wait(function () {
+            return test.form.answerField().val() === '';
```
mauriciocastillosilva-okta: nit: we could also wait for an ajax call. That way we check the field value after the call is made, and not in the wait condition. Similar comment in other tests too. the reason is that the test will fail in a `wait` condition instead of failing in an `expect`.

But this can be done later too, what you have is already much better than `tick`

test/unit/spec/MfaVerify_spec.js
```diff
@@ -1657,17 +1723,19 @@ function (Okta,
         return setupFn().then(function (test) {
           test.form.setAnswer('');
           test.form.submit();
-          return tick(test);
+          return Expect.wait(function () {
```
mauriciocastillosilva-okta: this can just be `Expect.waitForFormError(test.form, test);`

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2621,34 +2714,54 @@ function (Okta,
...
         });
         itp('displays only one error block if got an error resp on "Send email"', function () {
           var deferred = Util.mockRateLimiting();
           return setupEmail().then(function (test) {
             test.setNextResponse(resResendError);
+            $.ajax.calls.reset();
             test.form.emailSendCode().click();
-            return tick(test);
+            expect(test.form.emailSendCode().hasClass('disabled')).toBe(true);
+            return Expect.wait(function () {
```
mauriciocastillosilva-okta: you can use `Expect.waitForFormError(test.form, test);`

test/unit/spec/MfaVerify_spec.js
```diff
@@ -3389,7 +3524,9 @@ function (Okta,
                 test.setNextResponse(resInvalidTotp);
                 form.setAnswer('wrong');
                 form.inlineTOTPVerify().click();
-                return tick(form);
+                return Expect.wait(function () {
```
mauriciocastillosilva-okta: `Expect.waitForFormError(test.form, test);`

test/unit/spec/LoginRouter_spec.js
haishengwu-okta: then maybe we could do `calls.count === 1` ?

test/unit/spec/LoginRouter_spec.js
haishengwu-okta: as long as it work, let's remove it to avoid confusing.

test/unit/spec/MfaVerify_spec.js
aarongranick-okta: I'm almost certain I started with waiting for the ajax call. But I believe there was some extra async action after the ajax call which required continued waiting

PR #936, opened 2019-10-29
refactor: Updated icons/text for webauthn

magizh-okta:
* updated webauthn icon/description to guide endusers better during enroll/verify
* Updated css to accommodate long names.

* show additional 'restriction text' when browser is firefox/safari on mac. 
Screen Shot 2019-10-29 at 3 07 42 PM

Screen Shot 2019-10-30 at 11 14 15 AM

Screen Shot 2019-10-29 at 4 08 11 PM
Screen Shot 2019-10-29 at 4 08 49 PM
Screen Shot 2019-10-29 at 5 29 12 PM


assets/sass/modules/_enroll.scss
```diff
@@ -33,11 +33,19 @@
...
+  align-items: center;
 
   .enroll-factor-description {
     overflow: hidden;
   }
 
+  .enroll-factor-label {
+    display: flex;
+    justify-content: space-between;
+    min-width: 250px;
```
edburyenegren-okta: do `width: 100%` or `flex-shrink: 0` solve this? might be less brittle than the `min-width`

src/EnrollWebauthnController.js
```diff
@@ -134,32 +134,38 @@ function (Okta, Errors, FormType, FormController, CryptoUtil, webauthn, Footer,
...
+                     <p>{{{i18n code="enroll.webauthn.instructions.noSupportForBiometric" bundle="login"}}}</p>\
                   </div>\
                 {{/if}}\
                 <div data-se="webauthn-waiting" class="okta-waiting-spinner hide"></div>\
               ',
               getTemplateData: function () {
                 return {
-                  isEdge: BrowserFeatures.isEdge
+                  isEdge: BrowserFeatures.isEdge,
+                  onlySupportsSecurityKey: (BrowserFeatures.isFirefox || BrowserFeatures.isSafari) && (BrowserFeatures.isMac)
```
gowthamidommety-okta: Are there any other browser / OS (including devices) combination that might not support biometric? Would it make sense to show this note in all cases?

gowthamidommety-okta:
Another question: in some screen shots, `Security Key` has s and k capital, should we follow the same for `biometric Authenticator`?

src/EnrollWebauthnController.js
magizh-okta: windows - edge, chrome and firefox support both 
mac - only chrome support both 
android - chrome supports both I guess 
ios - no webauthn support 

I guess mac firefox/safari seems to be the most common that we would like to check for and show a message.

assets/sass/modules/_enroll.scss
magizh-okta: both doesn't help. I dont think the widget has any responsive css.

Container min-width is 400px.

magizh-okta:
@gowthamidommety-okta  you are right I updated the factor.webauthn string to be capitalized 

Screen Shot 2019-10-29 at 4 08 11 PM

assets/sass/modules/_enroll.scss
mauriciocastillosilva-okta: we have some responsive css for the white background, but not for labels or icons...

assets/sass/modules/_enroll.scss
```diff
@@ -97,11 +105,11 @@
 .enroll-factor-row-min {
   margin-bottom: 10px;
 
-  h3 {
```
mauriciocastillosilva-okta: why are we getting rid of `h3`?

packages/@okta/i18n/dist/properties/login.properties
```diff
@@ -229,8 +229,8 @@ factor.password = Password
 factor.customFactor.description = Redirect to a third party MFA provider to sign in to Okta.
 factor.customFactor.description.generic = Redirect to a third party MFA provider to sign in.
 factor.customFactor.description.specific = Redirect to a third party MFA provider to sign in to {0}.
-factor.webauthn = Security Key or Built-in Authenticator
-factor.webauthn.description = Use a security key (USB or bluetooth) or a built-in device authenticator (Windows Hello, macOS TouchID, etc.)
+factor.webauthn.v2 = Security Key or Biometric Authenticator
```
mauriciocastillosilva-okta: nit: `v2` sounds like a string for webauthn v2, as opposed to just a new string (there is now webauthn v2, I know)... maybe `factor.webauthn.biometric`?? same nit for other key changes

src/EnrollWebauthnController.js
```diff
@@ -134,32 +134,39 @@ function (Okta, Errors, FormType, FormController, CryptoUtil, webauthn, Footer,
...
                   </div>\
                 {{/if}}\
                 <div data-se="webauthn-waiting" class="okta-waiting-spinner hide"></div>\
               ',
               getTemplateData: function () {
                 return {
-                  isEdge: BrowserFeatures.isEdge
+                  isEdge: BrowserFeatures.isEdge,
+                  onlySupportsSecurityKey: (BrowserFeatures.isFirefox || BrowserFeatures.isSafari)
+                    && (BrowserFeatures.isMac)
```
mauriciocastillosilva-okta: so if it is safari and macOs, it doesn't support touch id of the macbooks?

test/unit/spec/EnrollWebauthn_spec.js
```diff
@@ -172,7 +172,20 @@ function (Okta,
           Expect.isVisible(test.form.enrollInstructions());
           Expect.isVisible(test.form.enrollEdgeInstructions());
           expect(test.form.enrollEdgeInstructions().text()).toEqual('Note: If you are enrolling a security key and' +
-          ' Windows Hello or PIN is enabled, you will need to select \'Cancel\' before continuing.');
+          ' Windows Hello or PIN is enabled, you will need to select \'Cancel\' in the prompt before continuing.');
+          Expect.isVisible(test.form.submitButton());
+        });
+      });
+
+      itp('shows a note on support restrictions for firefox and safari on mac', function () {
```
mauriciocastillosilva-okta: we should have 2 different tests, one for safari and one for firefox.

src/EnrollWebauthnController.js
magizh-okta: lol yes!

assets/sass/modules/_enroll.scss
magizh-okta: We dont need the padding any more because we are vertically  centering the text using 
align-items: center; in the parent. 

h3 here is the enroll-factor-label

mauriciocastillosilva-okta: so we don't need any padding at the top? I think that was for separating the beacon from the text... Do you mind double checking with a different factor?

PR #945, opened 2019-10-31
fix: fix enrollment for for idx

royalchan-okta:
- Add same view as Auth for Email enrollment now that states are correct.
- Backwards compat change for password enrollment matching new contract

OKTA-255628

### DIFF
--- a/playground/mocks/idp/idx/data/factor-enroll-password.json
+++ b/playground/mocks/idp/idx/data/factor-enroll-password.json
@@ -16,11 +16,11 @@
         "method": "POST",
         "value": [
           {
-            "name": "credential",
+            "name": "credentials",
             "form": {
               "value": [
                 {
-                  "name": "value",
+                  "name": "passcode",
                   "label": "Password",
                   "secret": true
                 }
--- a/src/v2/view-builder/ViewFactory.js
+++ b/src/v2/view-builder/ViewFactory.js
@@ -13,7 +13,7 @@ import EnrollFactorPasswordView from './views/password/EnrollFactorPasswordView'
 import RequiredFactorPasswordView from './views/password/RequiredFactorPasswordView';
 
 // email
-import EnrollFactorEmailView from './views/email/EnrollFactorEmailView';
+// import EnrollFactorEmailView from './views/email/EnrollFactorEmailView';
 import RequiredFactorEmailView from './views/email/RequiredFactorEmailView';
 import TerminalReturnEmailView from './views/email/TerminalReturnEmailView';
 import TerminalTransferedEmailView from './views/email/TerminalTransferedEmailView';
@@ -36,7 +36,7 @@ const VIEWS_MAPPING = {
     [DEFAULT]: EnrollProfileView,
   },
   'enroll-factor': {
-    email: EnrollFactorEmailView,
+    email: RequiredFactorEmailView, // TODO EnrollFactorEmailView is unimplemented
     password: EnrollFactorPasswordView
   },
   'challenge-factor': {
--- a/src/v2/view-builder/views/password/EnrollFactorPasswordView.js
+++ b/src/v2/view-builder/views/password/EnrollFactorPasswordView.js
@@ -1,7 +1,6 @@
 import { loc } from 'okta';
 import BaseView from '../../internals/BaseView';
 import BaseForm from '../../internals/BaseForm';
-import ValidationUtil from '../../../../util/ValidationUtil';
 import BaseFactorView from '../shared/BaseFactorView';
 
 const Body = BaseForm.extend({
@@ -41,12 +40,13 @@ export default BaseFactorView.extend({
     return ModelClass.extend({
       local,
       validate: function () {
-        return ValidationUtil.validateFieldsMatch(
-          this,
-          'credential.value',
-          'confirmPassword',
-          loc('password.error.match', 'login'),
-        );
+        if (this.get('credentials.passcode') !== this.get('confirmPassword') &&
+          this.get('credential.value') !== this.get('confirmPassword')) {
+          
+          var ret = {};
+          ret['confirmPassword'] = loc('password.error.match', 'login');
+          return ret;
+        }
       }
 
     });
--- a/playground/mocks/idp/idx/data/factor-verification-email.json
+++ b/playground/mocks/idp/idx/data/factor-verification-email.json
@@ -17,7 +17,16 @@
         "accepts": "application/vnd.okta.v1+json",
         "value": [
           {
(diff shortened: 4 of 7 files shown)


playground/mocks/idp/idx/data/factor-enroll-password.json
```diff
@@ -16,12 +16,11 @@
...
         "value": [
           {
-            "name": "credential",
+            "name": "credentials",
             "form": {
               "value": [
                 {
-                  "name": "value",
-                  "label": "Password",
+                  "name": "passcode",
```
nikhilvenkatraman-okta: I think this is breaking some of our e2e tests. https://travis-ci.org/okta/okta-signin-widget/jobs/605603470?utm_medium=notification&utm_source=github_status. You might have to update the selectors.

src/util/ValidationUtil.js
```diff
@@ -33,6 +33,17 @@ define(['okta'], function (Okta) {
     }
   };
 
+  // Validate that the password enroll matches for IdX. backwards compat, simplify later
+  fn.validatePasswordsMatchIdx = function (model) {
```
nikhilvenkatraman-okta: why are we creating another method `validatePasswordsMatchIdx`? Can we just update `validatePasswordMatch `. Also I would avoid having idx in the function name.

playground/mocks/idp/idx/data/factor-enroll-password.json
royalchan-okta: yep, will fix

src/util/ValidationUtil.js
nikhilvenkatraman-okta: Ah, i did not realize we are using a shared Util file here. If you want you can keep that method but just rename it to avoid IDX.

src/v2/view-builder/views/email/EnrollFactorEmailView.js
```diff
@@ -1,15 +1,114 @@
...
+import BaseFooter from '../../internals/BaseFooter';
+import email from '../shared/email';
+import polling from '../shared/polling';
 import BaseFactorView from '../shared/BaseFactorView';
 
-const Body = BaseForm.extend({
-  title: 'Sign in using a link sent to your email',
-  subtitle () {
-    return `Email will be sent to ${this.options.appState.get('factorProfile').email}`;
+const SHOW_RESEND_TIMEOUT = 60000;
```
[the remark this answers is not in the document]
royalchan-okta: They are for now, but I assumed they would diverge. I can just load the same view in both scenarios and leave a comment to fix it in the future if you prefer

playground/mocks/idp/idx/data/factor-enroll-password.json
```diff
@@ -16,12 +16,11 @@
         "method": "POST",
         "value": [
           {
-            "name": "credential",
+            "name": "credentials",
             "form": {
               "value": [
                 {
-                  "name": "value",
-                  "label": "Password",
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: Yes this is currently used to show the label on top of the `password` field. Screenshot below show the UI with this response where label is missing for first field.
Screen Shot 2019-10-31 at 10 29 24 AM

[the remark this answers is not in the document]
royalchan-okta: haha derp, will put it back and add one for email passcodes

src/util/ValidationUtil.js
haishengwu-okta: looks like it's only used in one place, isn't worth be a method in the util? or are we anticipating there will be more use cases?

playground/mocks/idp/idx/data/factor-enroll-password.json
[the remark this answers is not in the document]
royalchan-okta: okay, put labels consistently in okta-core

src/util/ValidationUtil.js
royalchan-okta: yeah, i agree, just moving this into the password enroll view

src/util/ValidationUtil.js
```diff
@@ -33,6 +33,17 @@ define(['okta'], function (Okta) {
     }
   };
 
+  // Validate that the password enroll matches for IdX. backwards compat, simplify later
+  fn.validatePasscodesMatch= function (model) {
```
haishengwu-okta: function is not used at all now. delete it?

src/v2/view-builder/ViewFactory.js
```diff
@@ -36,7 +36,7 @@ const VIEWS_MAPPING = {
     [DEFAULT]: EnrollProfileView,
   },
   'enroll-factor': {
-    email: EnrollFactorEmailView,
+    email: RequiredFactorEmailView, // TODO EnrollFactorEmailView is unimplemented
```
haishengwu-okta: does this mean `RequiredFactorEmailView` used for both enroll and challenge?
can we just delete `EnrollFactorEmailView` and add comments why same view for both flows.

src/v2/view-builder/views/password/EnrollFactorPasswordView.js
```diff
@@ -41,12 +40,13 @@ export default BaseFactorView.extend({
     return ModelClass.extend({
       local,
       validate: function () {
-        return ValidationUtil.validateFieldsMatch(
-          this,
-          'credential.value',
-          'confirmPassword',
-          loc('password.error.match', 'login'),
-        );
+        if (this.get('credentials.passcode') !== this.get('confirmPassword') &&
```
haishengwu-okta: nit:
looks like there is no reason to not reuse `ValidationUtil.validateFieldsMatch`

src/util/ValidationUtil.js
royalchan-okta: ah yeah, let me revert this file

src/v2/view-builder/ViewFactory.js
royalchan-okta: Maybe, I thought we might make them different at some point. Somehow we messed up the API response and always returned AUTHENTICATE during email enrollment challenges so this just magically worked before.

src/v2/view-builder/views/password/EnrollFactorPasswordView.js
royalchan-okta: does returning an OR of two calls work how I expect? It just didn't look readible

haishengwu-okta: ah, I missed it.

PR #946, opened 2019-10-31
chore: create _in translation when build assets

haishengwu-okta:
OKTA-259025


Gruntfile.js
```diff
@@ -68,6 +68,23 @@ module.exports = function (grunt) {
...
+        files: [
+          {
+            expand: true,
+            cwd: I18N_SRC,
+            src: [
+              'properties/{login,country}_id.properties'
+            ],
+            dest: I18N_SRC,
+            rename (dest, src) {
+              grunt.log.write(`copy app to target: ${dest}/${src} to ${dest}/${src.replace('_id', '_in')} \n`);
```
mauriciocastillosilva-okta: nit: `copy file to target` instead of `app`

package.json
```diff
@@ -38,7 +38,7 @@
     "lint:grunt:report": "grunt lint --checkstyle",
     "lint": "mkdir -p build2 && run-p -c lint:eslint lint:grunt lint:stylelint",
     "lint:report": "mkdir -p build2 && run-p -c lint:eslint:report lint:grunt:report lint:stylelint:report",
-    "test:dev": "yarn build:dev && karma start",
+    "test:dev": "grunt assets && karma start",
```
mauriciocastillosilva-okta: we don't need to build the app to run tests?

haishengwu-okta: no we dont. we use `build:dev` task previously because it needs those json files (`propertiesToJSON`) which is part of `build:dev` and now I moved to `assets` task.

PR #951, opened 2019-11-06
fix: fixed browser checks when displaying webauthn enroll instructions

magizh-okta:
* We do some browser checks to display some messages to guide users enrolling into webauthn.
The check partly work today because handlebars evaluates the function for us. But when we combine checks handlebars doesnt handle it.

* Also the check for safari needs to be enhanced since Chrome's UA string has safari it as well.

RESOLVES: OKTA-260148

Previously we had returned functions that was executed by handlebars.

Screen Shot 2019-11-06 at 1 14 46 PM

Chrome UA 
Screen Shot 2019-11-06 at 2 43 14 PM

**Fix**
Screen Shot 2019-11-06 at 3 04 23 PM

### DIFF
--- a/src/util/BrowserFeatures.js
+++ b/src/util/BrowserFeatures.js
@@ -58,7 +58,9 @@ define(function () {
   };
 
   fn.isSafari = function () {
-    return navigator.userAgent.toLowerCase().indexOf('safari') > -1;
+    // Chrome has safari in its useragent string so adding this extra check.
+    return navigator.userAgent.toLowerCase().indexOf('safari') > -1 &&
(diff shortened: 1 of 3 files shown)


src/util/BrowserFeatures.js
```diff
@@ -58,7 +58,9 @@ define(function () {
   };
 
   fn.isSafari = function () {
-    return navigator.userAgent.toLowerCase().indexOf('safari') > -1;
+    // Chrome has safari it its useragent string so adding this extra check.
```
gowthamidommety-okta: nit: `as its..`

PR #954, opened 2019-11-11
fix: fix the button disable style

haishengwu-okta:
OKTA-260287
demo


src/v2/controllers/FormController.js
```diff
@@ -77,13 +77,29 @@ export default BaseLoginController.extend({
...
+   * SignIn widget has its own (hacky) way to customize the button disabled state:
+   * adding `link-button-disabled` despite the name was intend only to disable
+   * `link-button`.
+   * Instead of doing decent refactor, we want to follow the convention for now.
+   *
+   * @param {boolean} disabled whether add extra disable CSS class.
+   */
+  toggleFormButtonState: function (disabled) {
+    var button = this.$el.find('.o-form-button-bar .button');
+    button.toggleClass('link-button-disabled', disabled);
```
nikhilvenkatraman-okta: was this class not getting added to this button ?

assets/sass/okta-theme.scss
```diff
@@ -52,19 +52,6 @@
     text-decoration: none;
     cursor: pointer;
   }
-
```
nikhilvenkatraman-okta: I am guessing this has been here for a while ? since we release 3.0 ?

haishengwu-okta: this was added in the OIE commit.

src/v2/controllers/FormController.js
haishengwu-okta: Courage doesn't add `link-button-disabled` for `input` buttons.

PR #958, opened 2019-11-14
chore: adds e2e tests for email factor on OIE

nikhilvenkatraman-okta:
Screen Shot 2019-11-14 at 3 09 15 PM

Resolves: OKTA-253716

### DIFF
--- a/test/testcafe/spec/ChallengeFactorEmail_OTP_spec.js
+++ b/test/testcafe/spec/ChallengeFactorEmail_OTP_spec.js
@@ -1,0 +1,56 @@
+import IdentityPageObject from '../framework/page-objects/IdentityPageObject';
+import ChallengeFactorPageObject from '../framework/page-objects/ChallengeFactorPageObject';
+import { ClientFunction, RequestMock } from 'testcafe';
+import factorRequiredEmail from '../../../playground/mocks/idp/idx/data/factor-verification-email';
+import success from '../../../playground/mocks/idp/idx/data/success';
+import invalidOTP from '../../../playground/mocks/idp/idx/data/error-email-verify';
+
+const validOTPmock = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx')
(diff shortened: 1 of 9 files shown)


test/testcafe/spec/ChallengeFactorEmail_OTP_spec.js
```diff
@@ -0,0 +1,56 @@
...
+  await identityPage.navigateToPage();
+  await identityPage.fillIdentifierField('Challenge Email');
+  await identityPage.clickNextButton();
+  return new ChallengeFactorPageObject(t);
+}
+
+test
+  .requestHooks(validOTPmock)
+  (`challenge email factor with valid OTP form has right labels`, async t => {
+    const challengeFactorPageObject = await setup(t);
```
nikhilvenkatraman-okta: One thing to note while writing these tests here is that for a long time I had the assertions for pageTitle within the same test as `challenge email factor with valid OTP`.  That added a bunch of await statements in the test which started causing flakeyness. Thats why i decided to have a test just to verify labels with few awaits.

playground/mocks/idp/idx/data/magiclink-expired.json
```diff
@@ -0,0 +1,20 @@
+{
```
haishengwu-okta: we already have a few `terminal-return-*.json`.
are you able to reconcile ?

test/testcafe/spec/TerminalReturnEmail_spec.js
```diff
@@ -1,4 +1,3 @@
-import IdentityPageObject from '../framework/page-objects/IdentityPageObject';
```
haishengwu-okta: looks like eslint doesn't catch it. mind want to follow up?

nikhilvenkatraman-okta: yeah, I just found this. Will create JIRA

PR #960, opened 2019-11-14
refactor: tweak the consent UI

haishengwu-okta:
- OKTA-242641
- https://okta.box.com/s/ij0nvc8zy0q99soxqq177tyny8k6lybx


assets/sass/modules/_consent.scss
```diff
@@ -109,16 +43,20 @@
...
 
+  .scope-list {
+    padding-bottom: $space-between-container;
+  }
+
   .consent-description {
-    padding-top: 15px;
-    padding-bottom: 25px;
-    font-size: $body-font-size;
+    padding-bottom: $space-of-last-container;
```
nikhilvenkatraman-okta: Just curious, how can we tell that this element with class of `.consent-description` should get padding-bottom of $space-of-last-container where as `.scope-list` gets a padding-bottom of $space-between-container. Would it make sense to add another class to `consent-description` something like `last-container`?

haishengwu-okta: we could go that route then we could 
- add generic class like `consent-row` to all those sections
- add css like 
    - `consent-row (not last) {padding: 25}`
    - `consent-row (last) {padding: 14}`

nikhilvenkatraman-okta: That seems ideal. Is that a small change to do?

haishengwu-okta: Addressed.

haishengwu-okta:
Screen Shot 2019-11-15 at 10 25 12 AM

venkatviswanathan-okta:
Looks good to me 🛳

PR #965, opened 2019-11-20
test: do not "Wait a tick for the tests to clean up" - OKTA-261467

aarongranick-okta:
This is the last use of `tick()` in shared code. Removing this will simplify the task of purging `tick` from individual test specs .

### DIFF
--- a/test/unit/helpers/util/Expect.js
+++ b/test/unit/helpers/util/Expect.js
@@ -24,7 +24,6 @@ define([
       };
       window.addEventListener('error', errListener);
       testFn.call(this)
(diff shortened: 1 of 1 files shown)


test/unit/helpers/util/Expect.js
```diff
@@ -24,7 +24,7 @@ define([
       };
       window.addEventListener('error', errListener);
       testFn.call(this)
-        .then(fn.tick) // Wait a tick for the tests to clean up
+        // .then(fn.tick) // Wait a tick for the tests to clean up
```
haishengwu-okta: just remove the line ?

PR #971, opened 2019-11-23
fix broken links

gabrielsroka:
dist folder is no longer here, but src is...

are these the right links?


haishengwu-okta:
@gabrielsroka yes, it's the right link. we need to update the readme. thanks for catching!

haishengwu-okta:
fixed in https://github.com/okta/okta-signin-widget/pull/973

PR #977, opened 2019-12-04
build: add device probing/polling mock and implementation w/o UX

jhe-okta:
OKTA-258151
OKTA-258152

started a server to handle request of /probe and /challenge from localhost:4000 when `yarn authenticator` is running
Screen Shot 2019-12-03 at 3 18 51 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,8 @@
     "build:webpack-release": "webpack --config webpack.release.config.js",
     "build:webpack-e2e-app": "webpack target/e2e/app/app-using-osw-entry.js target/js/app-bundle.js",
     "prestart": "yarn build:dev",
-    "prepublishOnly": "yarn build:release"
+    "prepublishOnly": "yarn build:release",
+    "mock:authenticator": "dyson playground/mocks/authenticator/ 6512"
   },
   "devDependencies": {
     "@commitlint/cli": "^8.1.0",
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -4,6 +4,7 @@
 #
 ########################################
 signout = Sign Out
+signin = Sign In
 remember = Remember me
 rememberDevice = Trust this device
 rememberDevice.timebased = Do not challenge me on this device for the next {0}
@@ -615,6 +616,8 @@ deviceTrust.sso.expire.subtitle = Unfortunately the steps required to sign in to
 deviceTrust.sso.expire.button = Sign in using Okta Mobile
 deviceTrust.universalLink.fallback.getOktaMobile.title = Get Okta Mobile
 deviceTrust.universalLink.fallback.getOktaMobile.subtitle = Go to the {0}AppStore{1}, {0}search{1} for {0}Okta Mobile{1} and tap on {0}GET{1} Okta Mobile. Once installed, sign in to Okta Mobile and follow the instructions to secure your device.
+
+# Device Challenge Probing and Polling
 
 # Consent Required
 consent.required.text = <b>{0}</b> would like to:
--- a/src/v2/view-builder/views/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/DeviceChallengePollView.js
@@ -1,0 +1,109 @@
+/* global Promise */
+import { _, $, loc } from 'okta';
+import BaseView from '../internals/BaseView';
+import BaseForm from '../internals/BaseForm';
+import Logger from '../../../util/Logger';
+
+const request = (url, method = 'GET', body = null, timeout = 1000) => {
+  return $.ajax({
+    url,
+    method,
+    timeout,
+    contentType: 'application/json',
+    data: body,
+  });
+};
+
+const Body = BaseForm.extend({
+  noButtonBar: true,
+
+  className: 'ion-form device-challenge-poll',
+
+  title: loc('signin', 'login'),
+
+  initialize () {
+    BaseForm.prototype.initialize.apply(this, arguments);
+    this.deviceChallengePollRemediation = this.options.appState.getCurrentViewState();
+    this.doChallenge();
+    this.startPolling();
+  },
+
+  remove () {
+    BaseForm.prototype.remove.apply(this, arguments);
+    this.stopPolling();
+  },
+
+  doChallenge () {
+    const deviceChallenge = this.options.appState.get('currentState')[
+      this.deviceChallengePollRemediation.relatesTo
+    ];
+    switch (deviceChallenge.challengeMethod) {
+    case 'LOOPBACK':
+      this.doLoopback(deviceChallenge.domain, deviceChallenge.ports, deviceChallenge.challengeRequest);
+      break;
+    case 'CUSTOM_URI':
+      this.doCustomURI(deviceChallenge.href);
+      break;
+    }
+  },
+
+  doLoopback (authenticatorDomainUrl = '', ports = [], challengeRequest = '') {
+    let currentPort;
+    let foundPort = false;
+
+    const getAuthenticatorUrl = (path) => {
+      return `${authenticatorDomainUrl}:${currentPort}/${path}`;
+    };
+
+    const checkPort = () => {
+      return request(getAuthenticatorUrl('probe'));
+    };
+
+    const onPortFound = () => {
+      foundPort = true;
+      return request(getAuthenticatorUrl('challenge'), 'POST', JSON.stringify({ challengeRequest }) , 3000);
+    };
+
+    const onFailure = () => {};
+
+    const doProbing = () => {
+      return checkPort()
+        .done(onPortFound)
+        .fail(onFailure);
+    };
+
+    let probeChain = Promise.resolve();
+    ports.forEach(port => {
+      probeChain = probeChain
+        .then(() => {
+          if (!foundPort) {
+            currentPort = port;
+            return doProbing();
+          }
+        })
+        .catch(() => {
+          Logger.error('Something unexpected happened during device probing.');
+        });
+    });
+  },
+
+  startPolling () {
+    const deviceChallengePollingInterval = this.deviceChallengePollRemediation.refresh;
+    if (_.isNumber(deviceChallengePollingInterval)) {
+      // TODO: how to show spinner to indicating action in progress?
+      this.polling = setInterval(() => {
+        this.options.appState.trigger('saveForm', this.model);
+      }, deviceChallengePollingInterval);
+    }
+  },
+
+  stopPolling () {
+    if (this.polling) {
+      clearInterval(this.polling);
+    }
+  },
+});
+
+export default BaseView.extend({
+  Body,
+});
--- a/test/testcafe/spec/DeviceChallengePollView_spec.js
+++ b/test/testcafe/spec/DeviceChallengePollView_spec.js
@@ -1,0 +1,61 @@
+import { RequestLogger, RequestMock, Selector } from 'testcafe';
+import DeviceChallengePollPageObject from '../framework/page-objects/DeviceChallengePollPageObject';
+import identifyWithDeviceProbingLoopback from '../../../playground/mocks/idp/idx/data/identify-with-device-probing-loopback';
+import loopbackChallengeNotReceived from '../../../playground/mocks/idp/idx/data/identify-with-device-probing-loopback-challenge-not-received';
+
+let failureCount = 0;
+
+const logger = RequestLogger(/introspect|probe|challenge/);
+
+const mock = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(identifyWithDeviceProbingLoopback)
+  .onRequestTo('http://localhost:3000/idp/idx/authenticators/poll')
+  .respond((req, res) => {
+    res.statusCode = '200';
+    if (failureCount === 2) {
+      res.setBody(loopbackChallengeNotReceived);
+    } else {
+      res.setBody(identifyWithDeviceProbingLoopback);
+    }
+  })
+  .onRequestTo('http://localhost:2000/probe')
+  .respond(null, 500, { 'access-control-allow-origin': '*' })
+  .onRequestTo('http://localhost:6511/probe')
+  .respond(null, 500, { 'access-control-allow-origin': '*' })
+  .onRequestTo('http://localhost:6512/probe')
+  .respond(null, 200, { 'access-control-allow-origin': '*' })
+  .onRequestTo('http://localhost:6512/challenge')
+  .respond(null, 200, {
+    'access-control-allow-origin': '*',
+    'access-control-allow-headers': 'Origin, X-Requested-With, Content-Type, Accept',
+    'access-control-allow-methods': 'POST, OPTIONS'
+  })
+
+fixture(`Device Challenge Polling View`)
+  .requestHooks(logger, mock)
+
+async function setup(t) {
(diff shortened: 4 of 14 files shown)


src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -0,0 +1,117 @@
...
+  },
+
+  remove () {
+    BaseForm.prototype.remove.apply(this, arguments);
+    this.stopPolling();
+  },
+
+  doChallenge () {
+    const deviceChallengeHintsKey = this.deviceChallengePollRemediation.relatesTo;
+    const deviceChallenge = deviceChallengeHintsKey &&
```
haishengwu-okta: nit:
doesn't have to check exists of `deviceChallengeHintsKey` cause `x[undefined] === undefined`

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -0,0 +1,117 @@
...
+};
+
+const Body = BaseForm.extend({
+  noButtonBar: true,
+
+  className: 'ion-form device-challenge-poll',
+
+  initialize () {
+    BaseForm.prototype.initialize.apply(this, arguments);
+    this.deviceChallengePollRemediation = this.options.appState.get('remediation').find(
```
haishengwu-okta: you could use `appState.getCurrentViewState`.

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -0,0 +1,117 @@
...
+
+  doLoopback (authenticatorDomainUrl = '', ports = [], challengeRequest = '') {
+    let currentPort;
+    let foundPort = false;
+
+    const getAuthenticatorUrl = (path) => {
+      // To use the mock server by running `yarn authenticator`
+      // return `http://localhost:4000/${path}?authenticator=${currentPort}`;
+
+      if (authenticatorDomainUrl === 'localhost:3000') {
```
haishengwu-okta: this doesn't seem ideal. 
I think we shall leave out this specific logic for local testing and find another way to make testing easier because for now
- in testcafe, you shall be able to mock response per port. port is just part of URL. (let me know if you had tried but didn't work)
- in mock server, you could just mock one port, e.g. `6512` hence given port list `[ "2000", "6511", "6512", "6513" ],`, the first 2 will be error and third port probing will succeed.

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -0,0 +1,117 @@
...
+
+    const onFailure = () => {};
+
+    const doProbing = () => {
+      return checkPort()
+        .done(onPortFound)
+        .fail(onFailure);
+    };
+
+    let probeChain = Promise.resolve();
```
haishengwu-okta: nit:
probably shall switch to jquery.deferred

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -0,0 +1,117 @@
...
+    ports.forEach(port => {
+      probeChain = probeChain
+        .then(() => {
+          if (!foundPort) {
+            currentPort = port;
+            return doProbing();
+          }
+        })
+        .catch(() => {
+          // console.log('Something unexpected happened during device probing.', 'error');
```
haishengwu-okta: use `/Util/Logger.error`

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -0,0 +1,117 @@
...
+        });
+    });
+  },
+
+  startPolling () {
+    const deviceChallengePollingInterval = this.deviceChallengePollRemediation &&
```
haishengwu-okta: `this.deviceChallengePollRemediation` must exists when renders this view.

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -0,0 +1,57 @@
...
+  return deviceChallengePollPage;
+}
+
+test
+  (`probing and polling APIs are sent and responded`, async t => {
+    await setup(t);
+    await t.expect(logger.count(record => record.response.statusCode === 200)).eql(3);
+    failureCount = 2;
+    await t.expect(logger.count(record => record.response.statusCode === 500)).eql(2);
+    const form =new Selector('.o-form').nth(0);
```
haishengwu-okta: can we leverage IdentifyPageObject to have something like
`identifyPageObject.waitForVisible`

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: I agree it is not ideal. But this is so far the only way I can make the test work. When the domain is different, even though Cross domain header is set in the response in RequestMock, the browser keeps throwing ERR_EMPTY_RESPONSE error. 
https://[a different domain than localhost]:[a different port than 3000] does not work
https://localhost:[a different port than 3000] does not work
https://localhost:3000/xxx works with cross domain header enabled

test/testcafe/spec/DeviceChallengePollView_spec.js
jhe-okta: I was thinking about it. But I don't know if this is identical to the Identify page, at least the identify input label is different (it could be just unintentional). And I think the response is just a mock, so I didn't want to assume it is the Identify page.

src/v2/view-builder/views/DeviceChallengePollView.js
haishengwu-okta: I'm guessing it's because you set ` "domain": "localhost:3000"` in the mock response.

test/testcafe/spec/DeviceChallengePollView_spec.js
jhe-okta: We just had a design review, the page is different from the regular user name page though it still have the username input. I will create a new page object when I implement the custom URL approach.

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: Why is it better?

playground/mocks/challenge.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  path: '/challenge',
+  method: 'POST',
+  render: (req, res) => {
+    res.append('Access-Control-Allow-Origin', '*');
```
haishengwu-okta: looks like dyson support cors by default which implies you dont need those `res.append...`.
could you double-check?

src/v2/view-builder/views/DeviceChallengePollView.js
haishengwu-okta: just be consistent. you're using `$.ajax` to make api request which use `$.Deferred` underneath.

test/testcafe/spec/DeviceChallengePollView_spec.js
haishengwu-okta: would you leave comments/jira in the code as well?

haishengwu-okta: missing space after `=` at 
```
form =new Selector...
```
(as a side note, we need to look into why eslint doesn't catch it)

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -616,6 +616,9 @@ deviceTrust.sso.expire.button = Sign in using Okta Mobile
 deviceTrust.universalLink.fallback.getOktaMobile.title = Get Okta Mobile
 deviceTrust.universalLink.fallback.getOktaMobile.subtitle = Go to the {0}AppStore{1}, {0}search{1} for {0}Okta Mobile{1} and tap on {0}GET{1} Okta Mobile. Once installed, sign in to Okta Mobile and follow the instructions to secure your device.
 
+# Device Challenge Probing and Polling
+deviceChallengePoll.title = Sign in
```
nikhilvenkatraman-okta: We'hv already got a few tokens in this file for `Sign in`
https://github.com/okta/okta-signin-widget/blob/a40c24fb3d1897d6537a4b0c31b9468bb413a9be/packages/%40okta/i18n/src/properties/login.properties#L477
https://github.com/okta/okta-signin-widget/blob/a40c24fb3d1897d6537a4b0c31b9468bb413a9be/packages/%40okta/i18n/src/properties/login.properties#L482
and this seems like another new key with the same value. Can we update this to be something more generic like `signin.title = Sign in`  so that we can potentially re-use this in the future or re-use the existing one ?

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -0,0 +1,53 @@
...
+    } else {
+      res.setBody(identifyWithDeviceProbingLoopback);
+    }
+  })
+  .onRequestTo('http://localhost:2000/probe')
+  .respond(null, 500, { 'access-control-allow-origin': '*' })
+  .onRequestTo('http://localhost:6511/probe')
+  .respond(null, 500, { 'access-control-allow-origin': '*' })
+  .onRequestTo('http://localhost:6512/probe')
+  .respond(null, 200, { 'access-control-allow-origin': '*' })
```
nikhilvenkatraman-okta: Just curious, how do we know the request with port 6512 onwards will give us a 200 ?

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -0,0 +1,53 @@
...
+  return deviceChallengePollPage;
+}
+
+test
+  (`probing and polling APIs are sent and responded`, async t => {
+    const deviceChallengePollPageObject = await setup(t);
+    await t.expect(deviceChallengePollPageObject.getHeader()).eql('Sign in');
+    await t.expect(logger.count(record => record.response.statusCode === 200)).eql(3);
+    failureCount = 2;
+    await t.expect(logger.count(record => record.response.statusCode === 500)).eql(2);
```
haishengwu-okta: shall we verify there is no request to port `6513` ?

playground/mocks/challenge.js
jhe-okta: You are correct.
Screen Shot 2019-12-06 at 2 30 23 PM

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: I've given it a try but not able to make it work with jquery deffered. The current implementation works consistently, and feels like we are going to avoid using jquery in other repo anyway. Are you okay with the current implementation?

test/testcafe/spec/DeviceChallengePollView_spec.js
jhe-okta: I am not sure if I follow your question. Ports will be tried one after another until one responds 200. In this mock, there are ports 2000, 6511, 6512, 6513. Port 2000 and 6512 return 500, port 6512 returns 200. Thus port 6513 won't be used. Port 6512 will continue be used to make a request to /challenge.

src/v2/view-builder/views/DeviceChallengePollView.js
haishengwu-okta: We can investigate later what's essential difference.

package.json
```diff
@@ -65,7 +65,8 @@
     "build:webpack-release": "webpack --config webpack.release.config.js",
     "build:webpack-e2e-app": "webpack target/e2e/app/app-using-osw-entry.js target/js/app-bundle.js",
     "prestart": "yarn build:dev",
-    "prepublishOnly": "yarn build:release"
+    "prepublishOnly": "yarn build:release",
+    "authenticator": "dyson playground/mocks/authenticator/ 6512"
```
haishengwu-okta: nit:
`authenticator` is not specific enough. Maybe something like `mock-authenticator`, `start-authenticator-mocks`.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -4,6 +4,7 @@
 #
 ########################################
 signout = Sign Out
+signin = Sign in
```
haishengwu-okta: looks like shall say `Sign In`

jhe-okta: Not sure. I know it is not consistent with 'Sign Out', but it is not consistent everywhere. Plus the mock says 'Sign in'.

package.json
jhe-okta: The path contains `mocks`, so I think it implies that it is the mock for authenticator. I can change it to `playground/mocks/start-authenticator`.

packages/@okta/i18n/src/properties/login.properties
haishengwu-okta: It's kind of consistent.
- `Sign in xxxx`
- or `Sign In` (when just `sign in`, first char of both word are upper case)

btw, I dont see any links to mocks but can you doule check with UX?

package.json
jhe-okta: Hmm...I guess `start-mock-authenticator` is better. I will change it.

haishengwu-okta: yeah, the path has `mocks` until people look at the code. name shall be more explicit like others `build:*`, `test:*`. having said so, `start:authenticator` seems better choice.

packages/@okta/i18n/src/properties/login.properties
nikhilvenkatraman-okta: yeah, we should bring it up to UX that we have `Sign In` and `Sign in` and we should try and be consistent

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -0,0 +1,109 @@
...
+    const checkPort = () => {
+      return request(getAuthenticatorUrl('probe'));
+    };
+
+    const onPortFound = () => {
+      foundPort = true;
+      return request(getAuthenticatorUrl('challenge'), 'POST', JSON.stringify({ challengeRequest }) , 3000);
+    };
+
+    const onFailure = () => {};
```
nikhilvenkatraman-okta: What is the use of this empty `onFailure `

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -0,0 +1,109 @@
...
+
+    const doProbing = () => {
+      return checkPort()
+        .done(onPortFound)
+        .fail(onFailure);
+    };
+
+    let probeChain = Promise.resolve();
```
nikhilvenkatraman-okta: Can you clarify why we are resolving the promise here right away ?

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: It's more like a placeholder for now until we know if we need UX to handle it.

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: It is resolved right away in order to be chained in a for loop.

packages/@okta/i18n/src/properties/login.properties
jhe-okta: The mock is in figma. I just checked, it is now updated to `Sign In`.

src/v2/view-builder/views/DeviceChallengePollView.js
nikhilvenkatraman-okta: should we add a comment then?

PR #978, opened 2019-12-06
fix: supports password factor explain

mauriciocastillosilva-okta:
- Adding explain to password factor field (similar to password in primary auth)
- explain will only be rendered if it is a custom i18n property (no default value)

Resolves: OKTA-250443


mauriciocastillosilva-okta:
Work was done in https://github.com/okta/okta-core/pull/38718

PR #987, opened 2019-12-12
test: fix remove tick in the test (credplat)

haishengwu-okta:
OKTA-266329

### DIFF
--- a/test/unit/spec/EnrollCall_spec.js
+++ b/test/unit/spec/EnrollCall_spec.js
@@ -1,4 +1,4 @@
-/* eslint max-params: [2, 18] */
+/* eslint max-params: 0 */
 define([
   'q',
   'okta',
@@ -21,14 +21,12 @@ define([
   'helpers/xhr/MFA_ENROLL_ACTIVATE_invalid_phone',
   'LoginRouter'
 ],
-/* eslint max-params: [2, 20] */
 function (Q, Okta, OktaAuth, LoginUtil, Util, Form, Beacon, Expect, $sandbox,
   resAllFactors, resAllFactorsIdx, resExistingPhone, resFactorEnrollExistingPhone, resFactorEnrollCall, resEnrollSuccess, resFactorEnrollActivateCall, resEnrollError, resActivateError,
   resEnrollInvalidPhoneError, Router) {
 
   var { _, $ } = Okta;
   var itp = Expect.itp;
-  var tick = Expect.tick;
 
   Expect.describe('EnrollCall', function () {
 
@@ -46,7 +44,7 @@ function (Q, Okta, OktaAuth, LoginUtil, Util, Form, Beacon, Expect, $sandbox,
       router.on('afterError', afterRenderHandler);
       Util.registerRouter(router);
       Util.mockRouterNavigate(router, startRouter);
-      return tick()
+      return Q()
         .then(function () {
           setNextResponse(resp || resAllFactors);
           return Util.mockIntrospectResponse(router, resp || resAllFactors);
@@ -66,11 +64,20 @@ function (Q, Okta, OktaAuth, LoginUtil, Util, Form, Beacon, Expect, $sandbox,
             setNextResponse: setNextResponse,
             afterRenderHandler: afterRenderHandler
           });
+        })
+        .then(test => {
+          spyOn(test.router.controller, 'trapAuthResponse').and.callThrough();
+          return test;
         });
     }
 
     function setupWithFactorEnroll () {
       return setup(resAllFactorsIdx);
+    }
+
+    function waitForEnrollActivateSuccess (test) {
+      test.router.controller.trapAuthResponse.calls.reset();
+      return Expect.waitForSpyCall(test.router.controller.trapAuthResponse, test);
     }
 
     function enterPhone (test, countryCode, phoneNumber, phoneExtension) {
@@ -83,7 +90,7 @@ function (Q, Okta, OktaAuth, LoginUtil, Util, Form, Beacon, Expect, $sandbox,
       test.setNextResponse(res);
       enterPhone(test, countryCode, phoneNumber, phoneExtension);
       test.form.sendCodeButton().click();
-      return tick(test);
+      return test;
     }
 
     function sendCodeOnEnter (test, res, countryCode, phoneNumber, phoneExtension) {
@@ -95,7 +102,7 @@ function (Q, Okta, OktaAuth, LoginUtil, Util, Form, Beacon, Expect, $sandbox,
       var keyup = $.Event('keyup');
       keyup.which = 13;
       test.form.phoneNumberField().trigger(keyup);
-      return tick(test);
+      return test;
     }
 
     function setupAndSendCode (res, countryCode, phoneNumber, phoneExtension) {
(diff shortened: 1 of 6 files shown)


test/unit/spec/EnrollQuestions_spec.js
```diff
@@ -228,7 +227,9 @@ function (Q, Okta, OktaAuth, Util, Form, Beacon, Expect, Router, RouterUtil, Bro
         spyOn(RouterUtil, 'isDocumentVisible').and.callFake(function () {
           return true;
         });
-        return tick();
+        // document.removeEventListener happens to be last call in the flow
+        // before invoke global.successFn.
+        return Expect.waitForSpyCall(document.removeEventListener);
```
aarongranick-okta: Would it work to wait for the global.successFn?

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -1772,10 +1738,9 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
           return Expect.waitForSpyCall(test.successSpy, test);
         })
           .then(function (test) {
+            // it depends client to invoke `callGlobalSuccess`
+            // hence form is still disabling.
```
aarongranick-okta: would it make sense to wait for `callGlobalSuccess` ?

haishengwu-okta: it does wait for global success function at line above
`return Expect.waitForSpyCall(test.successSpy, test);`
My comment is a little misleading. I'm trying to explain why the form is still disabled even when `callGlobalSuccess` is invoked.

test/unit/spec/EnrollQuestions_spec.js
haishengwu-okta: ah, good point!

test/unit/spec/PrimaryAuth_spec.js
aarongranick-okta: OK, I see. Yes, I was confused by the comment. Perhaps the meaning can be clarified?

haishengwu-okta: yeah, updated comments.

test/unit/spec/EnrollSms_spec.js
```diff
@@ -153,12 +186,14 @@ function (Okta, OktaAuth, LoginUtil, Util, AuthContainer, Form, Beacon, Expect,
...
-          test.form.backLink().click();
-          return Expect.waitForEnrollChoices();
-        })
+        return sendValidCodeFn()
+          .then(function (test) {
+            $.ajax.calls.reset();
+            test.setNextResponse(resAllFactors);
+            test.form.backLink().click();
+            return Expect.waitForEnrollChoices();
+            // return Expect.waitForSpyCall($.ajax);
```
aarongranick-okta: remove commented code?

test/unit/spec/EnrollSms_spec.js
```diff
@@ -262,59 +298,68 @@ function (Okta, OktaAuth, LoginUtil, Util, AuthContainer, Form, Beacon, Expect,
...
+            return test;
           })
+          .then(waitForEnrollActivateSuccess)
           .then(function (test){
             // Re-send after 30s wil show the warning again
             expectResendButton(test);
             expect(test.form.warningMessage()).toContain(
               'Haven\'t received an SMS? To try again, click Re-send code.');
-          });     
+            jasmine.clock().uninstall();
```
aarongranick-okta: could cause interference with other tests?

test/unit/spec/EnrollCall_spec.js
```diff
@@ -46,7 +44,7 @@ function (Q, Okta, OktaAuth, LoginUtil, Util, Form, Beacon, Expect, $sandbox,
       router.on('afterError', afterRenderHandler);
       Util.registerRouter(router);
       Util.mockRouterNavigate(router, startRouter);
-      return tick()
+      return Q()
```
mauriciocastillosilva-okta: I wonder if could we just return what is returned in the next `then` block? instead of `Q`
```
setNextResponse(resp || resAllFactors);
return Util.mockIntrospectResponse(router, resp || resAllFactors);
```

test/unit/spec/EnrollQuestions_spec.js
```diff
@@ -25,23 +25,24 @@ function (Q, Okta, OktaAuth, Util, Form, Beacon, Expect, Router, RouterUtil, Bro
...
       baseUrl: baseUrl,
       authClient: authClient,
+      globalSuccessFn: successSpy,
       'features.router': startRouter
     });
     router.on('afterError', afterErrorHandler);
     Util.registerRouter(router);
     Util.mockRouterNavigate(router, startRouter);
-    return tick()
+    return Q()
```
mauriciocastillosilva-okta: same question as in the other test, if we can return the logic from the `then` block below

test/unit/spec/EnrollSms_spec.js
```diff
@@ -262,59 +298,68 @@ function (Okta, OktaAuth, LoginUtil, Util, AuthContainer, Form, Beacon, Expect,
...
-        return setup().then(function (test) {
-          test.setNextResponse(resFactorEnrollActivateSms);
-          enterCode(test, 'US', '4151111111');
-          test.form.sendCodeButton().click();
-          return tick(test);
-        })
+      itp('shows warning message to click "Re-send" after 30s', function () {
+        Util.speedUpDelay();
+        return sendCodeFn(successRes, 'US', '4151111111')
+          .then(waitForEnrollActivateSuccess)
```
mauriciocastillosilva-okta: isn't the `sendCodeFn` already waiting for enrollActivateSuccess? Similar to what you are doing in Call tests, the changes to the sendCode functions have an `Expect.watifForEnrollActivateSuccess` now... or maybe this is a different sendCodeFn?

test/unit/spec/EnrollSms_spec.js
haishengwu-okta: shall not but let me remove it since I didn't use clock and forgot to detele this line.

test/unit/spec/EnrollSms_spec.js
haishengwu-okta: Not really (too much sharing cause confusing sometime)
`sendCodeFn` for `testEnrollPhoneNumber` is either `setupAndSendCode` or `setupAndSendCodeIdx` which only does `setup` and `sendCode`.
`setupAndSendValidCode` does set up `setup`, `sendCode` and `waitForSuccess`

This test may deserve further refactors to make it more clear but I gave up at this PR to reduce the scope.

test/unit/spec/EnrollQuestions_spec.js
haishengwu-okta: turns out I cannot because the whole chain is like
```
setup() {
  return Util.mockIntrospectResponse.then(..).then(..)
}
.then(..)
.then(* clear up tests; done(); */)                  <-- from Expect.runTest
.done(done)                                          <-- from Expect.runTest
```

When `Util.mockIntrospectResponse` is resolved, it will trigger `done(done)` when ends test hence cause many failures. Not sure if I did some other stupid thing but this was what I found last time.

test/unit/spec/EnrollCall_spec.js
haishengwu-okta: relied in another similar comment.

test/unit/spec/EnrollQuestions_spec.js
mauriciocastillosilva-okta: oh.... I see, a little more complicated then... sigh... thanks for trying...
Maybe it needs to load the page first, and then it can do the introspect call... so this `Q` could be a `Expect.waitForPrimaryAuth` or something like that, since we can't really skip this step. We need to load the initial page, then introspect, which refreshes auth-state, and then go on... 

Anyways, if you want to try it, go for it, but it is good as is, already better than `tick`

haishengwu-okta: Sounds good. Will give an try next time I visit this file ;)

PR #988, opened 2019-12-12
test: remove tick from test (device fingerprint)

haishengwu-okta:
OKTA-266331

### DIFF
--- a/test/unit/spec/DeviceFingerprint_spec.js
+++ b/test/unit/spec/DeviceFingerprint_spec.js
@@ -119,11 +119,7 @@ function (Expect, $sandbox, DeviceFingerprint) {
         return false;
       });
       mockIFrameMessages(true);
-
-      DeviceFingerprint.generateDeviceFingerprint(baseUrl, $sandbox);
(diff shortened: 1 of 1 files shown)


test/unit/spec/DeviceFingerprint_spec.js
```diff
@@ -102,23 +108,30 @@ function (Expect, $sandbox, DeviceFingerprint) {
...
-        .fail(function (reason) {
-          done.fail('Fingerprint promise should not have been rejected. ' +  reason);
-        });
-      Expect.tick().then(function () {
-        // give it time to check if promise resolves or rejects.
+
+      DeviceFingerprint.generateDeviceFingerprint(baseUrl, $sandbox);
+      expect(DeviceFingerprint.isMessageFromCorrectSource).not.toHaveBeenCalled();
+
+      window.addEventListener('message', () => {
```
aarongranick-okta: Is this running in jest or karma? If karma, I think we'll need to remove this event listener to prevent interference with other tests

PR #991, opened 2019-12-18
chore: remove tick() from LoginRouter_spec OKTA-266328

aarongranick-okta:


test/unit/spec/LoginRouter_spec.js
```diff
@@ -902,12 +900,13 @@ function (Okta, Q, Logger, Errors, BrowserFeatures, WidgetUtil, Bundles, config,
           Util.mockSDKCookie(test.ac);
           test.setNextResponse([resMfaChallengeDuo, resMfa]);
           test.router.navigate('signin/verify/duo/web', { trigger: true });
-          return Expect.waitForMfaVerify();
```
mauriciocastillosilva-okta: curious why this didn't work... the view that will load should be MFAVerify...

aarongranick-okta: It does work for the test, but sometimes an ajax call would occur after the test had completed, causing instability in other tests. If you like I can put the original wait condition back and then add the wait for 3 ajax calls at the end of the test.

PR #998, opened 2020-01-03
build: add custom URI approach for signing in device

jhe-okta:
OKTA-258116

- the flow is
  1. try loopback server approach first (where you see the sign in and spinner)
  2. shows the sign in page with okta verify link when loopback server approach fails (no response)
  3. clicking on the okta verify link takes user to the launch UI where Okta Verify app will be launched as well (you don't see it from the demo below as there's no Okta Verify app installed, but you should see the launch-okta-verify POST api call)
  4. clicking on the 'Go back' link takes user back to step ii 
- the download link is expected to not to be functional at the moment for BETA
authenticator-siw-custom-uri
Screen Shot 2020-01-03 at 10 59 06 AM


santoshmale-okta:
@okta/device-platform

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -104,6 +122,27 @@ const Body = BaseForm.extend({
   },
 });
 
+const Footer = BaseFooter.extend({
+  links () {
+    let links = [];
+    const deviceChallenge = this.options.appState.get(
```
haishengwu-okta: Could be the same logic as line 36?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -34,19 +33,34 @@ const Body = BaseForm.extend({
...
+      `);
+      this.customURI = deviceChallenge.href;
+      this.doCustomURI();
       break;
     }
   },
 
+  postRender () {
+    BaseForm.prototype.postRender.apply(this, arguments);
+    this.$('#launch-ov').on('click', _.bind(this.doCustomURI, this));
```
haishengwu-okta: use native `bind` instead of `_.bind`

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -87,6 +101,10 @@ const Body = BaseForm.extend({
     });
   },
 
+  doCustomURI () {
+    return request(this.customURI, 'POST', null, 3000);
```
haishengwu-okta: there are a couple of places that timeout is `3000`. can you add comments about why choose 3000?
if they have commonality, extra to an const variable.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -1,11 +1,32 @@
...
+import { View, loc, createButton } from 'okta';
 import BaseView from '../internals//BaseView';
 import BaseForm from '../internals//BaseForm';
 import BaseFooter from '../internals//BaseFooter';
 
 const Body = BaseForm.extend({
+
   title: loc('primaryauth.title'),
   save: loc('oform.next', 'login'),
+  postRender () {
```
haishengwu-okta: nit:
- probably could overload `initialize`
- otherwise invoke super.postRender first.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -1,11 +1,32 @@
-import { loc } from 'okta';
+import { View, loc, createButton } from 'okta';
 import BaseView from '../internals//BaseView';
 import BaseForm from '../internals//BaseForm';
 import BaseFooter from '../internals//BaseFooter';
```
haishengwu-okta: the double `//` was probably typo at some points. could you help to fix it?

assets/sass/v2/_identify.scss
```diff
@@ -0,0 +1,75 @@
+.sign-in-with-device-option {
+  .okta-verify-container {
+    .link-button {
```
haishengwu-okta: why one-off the link-button? it seems odyssey design and seems not compatible with other elements?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -34,19 +33,34 @@ const Body = BaseForm.extend({
...
       this.doLoopback(deviceChallenge.domain, deviceChallenge.ports, deviceChallenge.challengeRequest);
       break;
     case 'CUSTOM_URI':
-      this.doCustomURI(deviceChallenge.href);
+      this.title = 'Verify account access';
+      this.subtitle = 'Launching Okta Verify...';
+      this.add(`
+        If nothing prompts from the browser, 
+        <a href="#" id="launch-ov" class="link">click here</a> to launch Okta Verify, 
+        or <a href="#" id="download-ov" class="link">download and run Okta Verify</a>.
```
haishengwu-okta: is there a plan to implement the `download-ov` link?

assets/sass/v2/_identify.scss
nikhilvenkatraman-okta: Same question. This is the only button that will be as per odyssey design ?

jhe-okta:
> text for "Launching Okta Verify" should be #2F3F4A and semibold. (font size 16pts)
> Text below should also appear as #2F3F4A.
> Links should also be underlined all the time, not only on hover.

@nbories-okta , the title (Verify account access) has the font size of 14px. The subtitle (Launching Okta Verify...) will be bigger that the title if we set it to 16pt. Its current size is 13px by default.

assets/sass/v2/_device-challenge.scss
```diff
@@ -0,0 +1,32 @@
...
+      margin: 0 auto;
+
+      .spinner {
+        height: 38px;
+        margin-top: 40px;
+        background: url('../img/ui/indicators/loader@1x.gif') no-repeat center;
+      }
+
+      .link {
+        color: #007dc1;
```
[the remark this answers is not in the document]
jhe-okta: I don't see other v2 scss files are using var. And some of the colors here are from Odyssey, which is not available yet in SIW. I can replace the one available currently with var. For the ones that are not available, do you want me to create in-file var?

assets/sass/v2/_device-challenge.scss
```diff
@@ -0,0 +1,32 @@
...
+      }
+    }
+  }
+
+  .siw-main-footer {
+    .auth-footer {
+      .js-sign-in-options {
```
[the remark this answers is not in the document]
jhe-okta: Otherwise it is only shown as grey text. Screen Shot 2020-01-06 at 12 40 01 PM

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: I can override BaseView.initialize to make deviceChallenge available for both Body and Footer.

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: The custom timeout is only needed in loopback approach. In custom URI, we actually don't need to customize the timeout. I will remove the timeout from ajax call for custom URI approach.

assets/sass/v2/_identify.scss
jhe-okta: This is the only button in SIW v2 with Odyssey design as far as I can tell. @nbories-okta ^

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: We will not have the implementation for BETA. We will document this so the customers do not expect the download link to be working.

haishengwu-okta: Would be it slightly better to remove the content from the page at all?

assets/sass/v2/_device-challenge.scss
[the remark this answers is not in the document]
haishengwu-okta: That's the style across the board. I don't have strong opinion which style to use as long as UX decided but am looking consisency across all pages.

assets/sass/v2/_device-challenge.scss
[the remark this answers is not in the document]
haishengwu-okta: ah, that's bad practice. :(
we could either refer `/sass/common/shared/helpers/_variables.scss` or create a `v2/variable.scss` and refer/redefine colors

nbories-okta:
> > text for "Launching Okta Verify" should be #2F3F4A and semibold. (font size 16pts)
> > Text below should also appear as #2F3F4A.
> > Links should also be underlined all the time, not only on hover.
> 
> @nbories-okta , the title (Verify account access) has the font size of 14px. The subtitle (Launching Okta Verify...) will be bigger that the title if we set it to 16pt. Its current size is 13px by default. The link in SIW, as far as I can see, only has underline when hover on.

Ok, i am fine with keeping the font size @ 13pts in proxima nova. Let's make it semibold though rather than regular. Fine with following what currently exists in the SIW regarding the links so when we do change that pattern it gets applied without issues.

assets/sass/v2/_device-challenge.scss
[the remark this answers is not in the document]
jhe-okta: @nbories-okta ^

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: I don't have strong opinion, but based on the slack channel discussion as I remember, this is the text the team agreed on.

assets/sass/v2/_device-challenge.scss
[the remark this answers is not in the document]
jhe-okta: @haishengwu-okta the 'sign up' link in the identify footer also have custom css to make it blue. 
Screen Shot 2020-01-06 at 3 34 36 PM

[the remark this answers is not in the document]
haishengwu-okta: yes, that's really a one-off ! 
but we have other `go back` link at other pages. this makes thing more inconsistency.

src/v2/view-builder/views/DeviceChallengePollView.js
haishengwu-okta: oh, I see. then nvm for now.

jhe-okta:
Screen Shot 2020-01-06 at 5 08 51 PM

assets/sass/v2/_device-challenge.scss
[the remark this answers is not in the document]
nikhilvenkatraman-okta: I think it should probably be ok to change the color for `sign up` button link also to `$primary-color`.

src/v2/view-builder/views/DeviceChallengePollView.js
nikhilvenkatraman-okta: I second that, maybe we should think about removing the `download-ov` link. Otherwise this is this just going to be a link without any action right ?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -2,25 +2,27 @@
...
-  return $.ajax({
+  let ajaxOpt = {
     url,
     method,
-    timeout,
     contentType: 'application/json',
     data: body,
-  });
+  };
+  if (typeof timeout === 'number') {
```
nikhilvenkatraman-okta: why do we need this check ?

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: @nbories-okta ^ would you like to reword the sentence so that we don't mention of downloading the OV?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -34,19 +45,39 @@ const Body = BaseForm.extend({
...
+      this.add(`
+        If nothing prompts from the browser, make sure you have Okta Verify installed.
+      `);
+      this.customURI = deviceChallenge.href;
+      this.doCustomURI();
       break;
     }
   },
 
+  // Re-enable for EA
```
haishengwu-okta: just curious why even the launch-ov link is post BETA?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -2,25 +2,36 @@
...
 import Logger from '../../../util/Logger';
 
-const request = (url, method = 'GET', body = null, timeout = 1000) => {
-  return $.ajax({
+const request = (opts) => {
+  let url;
+  let method = 'GET';
+  let body = null;
+  let timeout = null;
+  if (typeof opts === 'string') {
```
haishengwu-okta: This is over design IMO. the parameter could be either Object or multiple variables all the time.

haishengwu-okta:
also, don't need to squash and force push.

jhe-okta:
@haishengwu-okta @nikhilvenkatraman-okta I don't have strong opinion on how the request function should look. The reason of the change is to have timeout null as default so that ajax call does not override the timeout value.

There's also test for `DeviceChallengePollView` to make sure view switch at the end is correct. I can follow up with the same test for `DeviceChallengePollViewOktaVerify` as there's timing issue currently.

I don't see any critical/blocking comments for the PR. Please let me know your final thought ASAP. I will create a topic branch off 3.6 so I can have a weekly release. We need to start integration test ASAP. Thanks.

src/v2/view-builder/views/DeviceChallengePollView.js
nikhilvenkatraman-okta: When is opts just a string that is equal to the api url ? For those cases can we just pass in an object instead `{ url : 'some url' }` and avoid this check ?

test/testcafe/spec/DeviceChallengePollViewOktaVerify_spec.js
```diff
@@ -0,0 +1,40 @@
...
+  await t.expect(deviceChallengePollPageObject.getHeader()).eql('Verify account access');
+  await t.expect(deviceChallengePollPageObject.getSubtitle()).eql('Launching Okta Verify...');
+  await t.expect(deviceChallengePollPageObject.getContent())
+    .eql('If nothing prompts from the browser, click here to launch Okta Verify, or make sure Okta Verify is installed.');
+  await t.expect(deviceChallengePollPageObject.getFooterLink().innerText).eql('Back to Sign In');
+  await t.expect(deviceChallengePollPageObject.getFooterLink().getAttribute('href')).eql('http://localhost:3000');
+});
+
+test(`should have the correct links`, async t => {
+  const deviceChallengePollPageObject = await setup(t);
```
nikhilvenkatraman-okta: I once saw that having too many of these await statements in the test caused flakyness on travis CI, but the test passed locally. Maybe try and break the expectations into multiple tests to see if that resolves the problem.

assets/sass/v2/_device-challenge.scss
```diff
@@ -0,0 +1,27 @@
+.siw-main-view {
+  .device-challenge-poll {
+    .o-form-content {
+      .okta-form-subtitle {
+        font-weight: 600;
+      }
+    }
+    .o-form-fieldset-container {
+      width: 88%;
```
haishengwu-okta: curious why would need to style the container?

playground/mocks/authenticator/challenge.js
```diff
@@ -1,5 +1,5 @@
 // To use this mock,
-// in terminal, run `yarn authenticator` to start the authenticator mock server
+// in terminal, run `yarn mock:authenticator` to start the authenticator mock server
```
haishengwu-okta: nit:
maybe remove the comment from those mock files and add to `playground/mocks/authenticator/README.md`

src/v2/view-builder/views/DeviceChallengePollView.js
haishengwu-okta: jquery would not use any undefined `timeout` value hence this check it's redundancy if your intent to avoid overwrite `timeout` to `undefined`

assets/sass/v2/_device-challenge.scss
jhe-okta: to make the text container narrower and text to show in more lines

test/testcafe/spec/DeviceChallengePollViewOktaVerify_spec.js
jhe-okta: This is fixed now. It was working locally before as I have the mock server running. Added mock response now in the test to fix the issue.

assets/sass/v2/_device-challenge.scss
jhe-okta: To make the text container narrower so that text is displayed in more lines.

PR #999, opened 2020-01-06
feat: minify node module entry, add no-polyfill bundle - OKTA-264040

aarongranick-okta:
- minifies node module entry. This will remove the console warning "The Okta Sign-In Widget is running in development mode." when importing/requiring the node module
- add instructions to README for using source maps with webpack
- For CDN users, build another bundle without polyfill (README should be updated after release to publicize this feature)

Related issue: https://oktainc.atlassian.net/browse/OKTA-264040

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,9 @@ After installing:
     // Use OktaSignIn
     var signIn = new OktaSignIn(/* configOptions */);
     ```
+
+    Source maps are provided as an external .map file. If you are using [Webpack](https://webpack.github.io/), these can be loaded using the [source-map-loader](https://github.com/webpack-contrib/source-map-loader) plugin.
+
     **Note:** If you use [Browserify](http://browserify.org/) to bundle your app, you'll need to use the `--noparse` option:
 
     ```bash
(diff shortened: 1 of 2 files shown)


haishengwu-okta:
Curious if you take some investigation on those 2 options mentioned at 
https://github.com/okta/okta-signin-widget/pull/614 and what's the outcome ?

aarongranick-okta:
> Curious if you take some investigation on those 2 options mentioned at
> #614 and what's the outcome ?

Yes I did look through those options. I think the main issue is that the user would like to use the SIW without a bundled polyfill, so they can provide their own. The other options such as 'idempotent-babel-polyfill', will require the user to use the SAME polyfill. Since the babel-polyfill is deprecated as of babel 7.4 (https://babeljs.io/docs/en/babel-polyfill), I do not want to force the user to use any SPECIFIC polyfill.  So the new bundle without polyfill should satisfy CDN users.

The issue with bundling themselves with webpack is that we are providing the unminified (DEV) code and there is no (simple) way for the user to bundle it with prod settings. They consider this a security risk (although there is none). Setting this entry to prod settings will remove the DEV warning and minify the code, so it should satisfy node module users.  Adding the source-map-loader will make the issue of minified code irrelevant to the developer experience.

aarongranick-okta:
> * It seems to need some rethink of #353 as I'm not a fan of delivery multiple bundles.

I added agenda topic for SIW review tomorrow, if we have time we can discuss it there.

PR #1003, opened 2020-01-09
build: add custom URI approach for signing in device

jhe-okta:
OKTA-258116

move the PR from https://github.com/okta/okta-signin-widget/pull/998 after rebase off branch 3.6.


assets/sass/v2/_device-challenge.scss
```diff
@@ -0,0 +1,27 @@
+.siw-main-view {
+  .device-challenge-poll {
+    .o-form-content {
+      .okta-form-subtitle {
+        font-weight: 600;
+      }
+    }
+    .o-form-fieldset-container {
```
haishengwu-okta: this is just not ideal.
either get ride of it if we don't have to or override on specific class for the view if possible.

PR #1005, opened 2020-01-09
fix: fixes bug where widget keeps auto sending magic link emails

nikhilvenkatraman-okta:
Resolves: OKTA-268169

- Backend is now going to do this optimization of auto sending the magic link email after factor selection.
- Removing the code from the widget.
- Also renamed the action link for resend.
- Note that this will be merged next week since we want to merge both backend and front end changes together.
- PR on release branch https://github.com/okta/okta-signin-widget/pull/1006


haishengwu-okta:
in favor of https://github.com/okta/okta-signin-widget/pull/1006

PR #1006, opened 2020-01-09
fix: fixes bug where widget keeps auto sending magic link emails

nikhilvenkatraman-okta:
Resolves: OKTA-268169

- Backend is now going to do this optimization of auto sending the magic link email after factor selection.
- Removing the code from the widget.
- Also renamed the action link for resend.
- **Note that this will be merged next week since** we want to merge both backend and front end changes together.

### DIFF
--- a/test/testcafe/spec/ChallengeFactorEmail_MagicLink_spec.js
+++ b/test/testcafe/spec/ChallengeFactorEmail_MagicLink_spec.js
@@ -1,8 +1,9 @@
 import IdentityPageObject from '../framework/page-objects/IdentityPageObject';
 import ChallengeFactorPageObject from '../framework/page-objects/ChallengeFactorPageObject';
-import { RequestMock } from 'testcafe';
+import { RequestMock, RequestLogger } from 'testcafe';
 import magicLinkReturnTab from '../../../playground/mocks/idp/idx/data/terminal-return-email';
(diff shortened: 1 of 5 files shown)


haishengwu-okta:
could you add some tests?

nikhilvenkatraman-okta:
yeah, I was thinking about that. Don't think we have anything specific for the email auto sending bit. I will look into writing something for the resend email flow.

test/testcafe/spec/ChallengeFactorEmail_MagicLink_spec.js
```diff
@@ -38,3 +49,25 @@ test
...
 
+  test
+  .requestHooks(logger, resendEmailMock, magicLinkEmailSentMock)
+  (`challenge email factor with magic link sent renders and has resend link`, async t => {
+    const challengeFactorPageObject = await setup(t);
+    await t.expect(challengeFactorPageObject.resendEmailView().getStyleProperty('display')).eql('none');
+    // wait for resend button to appear
+    await t.wait(65000);
+    // making sure we keep polling while we wait for the resend view to appear
+    await t.expect(logger.count(
```
haishengwu-okta: nit: 
- add assertion that logger has total count 17.
- explain a bit why 16 poll (Math.floor(65000/4000))

PR #1008, opened 2020-01-12
fix: adds a generic error message for piv

gowthamidommety-okta:
- adds a generic error message for piv when the error is empty

Resolves: OKTA-254793

Screen Shot 2020-01-12 at 4 11 48 PM

### DIFF
--- a/test/unit/spec/VerifyPIV_spec.js
+++ b/test/unit/spec/VerifyPIV_spec.js
@@ -21,7 +21,7 @@ function (Q, Okta, OktaAuth, Util, PivForm, PrimaryAuthForm, Beacon, Expect,
   var { _ } = Okta;
   var itp = Expect.itp;
 
-  function setup (hasError) {
+  function setup (errorResponse, responseTextOnly) {
     var setNextResponse = Util.mockAjax();
     var baseUrl = 'https://foo.com';
     var authClient = new OktaAuth({url: baseUrl});
@@ -43,7 +43,7 @@ function (Q, Okta, OktaAuth, Util, PivForm, PrimaryAuthForm, Beacon, Expect,
     Util.registerRouter(router);
     Util.mockRouterNavigate(router);
     spyOn(SharedUtil, 'redirect');
-    setNextResponse(hasError ? [resError] : [resGet, resPost]);
+    setNextResponse(errorResponse ? [errorResponse] : [resGet, resPost], responseTextOnly);
     router.verifyPIV();
     return Expect.waitForVerifyPIV({
       router: router,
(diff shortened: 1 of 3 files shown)


test/unit/spec/VerifyPIV_spec.js
```diff
@@ -127,6 +132,44 @@ function (Q, Okta, OktaAuth, Util, PivForm, PrimaryAuthForm, Beacon, Expect,
...
+      itp('shows generic error message for empty error response', function () {
+        var res = deepClone(resError);
+        res.response = {};
+        return setup(res).then(function (test) {
+          return Expect.waitForFormError(test.form, test);
+        }).then(function (test) {
+          expect(test.form.hasErrors()).toBe(true);
+          expect(test.form.errorBox()).toHaveLength(1);
+          expect(test.form.errorMessage())
+            .toEqual('We found some errors. Please review the form and make corrections.');
```
mauriciocastillosilva-okta: shouldn't this be the new error message you added? `There was an error signing in. Click the button below to try again.`? response json is empty and there is no response text.

test/unit/spec/VerifyPIV_spec.js
```diff
@@ -127,6 +132,44 @@ function (Q, Okta, OktaAuth, Util, PivForm, PrimaryAuthForm, Beacon, Expect,
...
+        var res = deepClone(resError);
+        res.responseType = 'text';
+        res.response = '';
+        return setup(res).then(function (test) {
+          return Expect.waitForFormError(test.form, test);
+        }).then(function (test) {
+          expect(test.form.hasErrors()).toBe(true);
+          expect(test.form.errorBox()).toHaveLength(1);
+          expect(test.form.errorMessage())
+            .toEqual('We found some errors. Please review the form and make corrections.');
```
mauriciocastillosilva-okta: same question here. 
Or if this is correct and I am missing something, then we need another test for that new error message

test/unit/spec/VerifyPIV_spec.js
```diff
@@ -97,8 +101,9 @@ function (Q, Okta, OktaAuth, Util, PivForm, PrimaryAuthForm, Beacon, Expect,
       });
     });
     Expect.describe('Error', function () {
+      var pivError = deepClone(resError);
```
mauriciocastillosilva-okta: nit: you probably don't need this, you can pass `resError` in the tests that do not change anything in that object

gowthamidommety-okta: Because of this logic - https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/mocks/Util.js#L104-L107 I've been having issues with these tests because it modifies the response directly.. One way around this was to do the above

mauriciocastillosilva-okta: not sure I follow why, maybe we can talk offline. Doesn't hurt to clone the response here, I just want to know whats going on

mauriciocastillosilva-okta: talked offline, thanks for explaining what's going on.
Util changes the original object to add a `responseText`, so after the first run we wouldn't be able to test a response that has no `responseText`.

PR #1012, opened 2020-01-15
fix: fixes redundant authn api call

nikhilvenkatraman-okta:
Resolves: OKTA-271959

Video 
https://okta.box.com/s/dgrqn0nrb90o05un0fk15t1qv6ptpxpm

@haishengwu-okta

### DIFF
--- a/src/RefreshAuthStateController.js
+++ b/src/RefreshAuthStateController.js
@@ -27,22 +27,20 @@ define(['q', 'okta', 'util/FormController'], function (Q, Okta, FormController) 
       var token = this.options.token;
       this.model.startTransaction(function (authClient) {
         appState.trigger('loading', true);
-        if (token && (token === this.settings.get('stateToken'))) {
-          // widget bootstrapped with statetoken via settings
-          // check if stateToken in url is same as settings
-          //unset stateToken to prevent the baseloginrouter controller from calling it again on render
+        if (this.options.appState.get('introspectSuccess') ||
+            this.options.appState.get('introspectError')) {
+          // Also handles the case when we hit introspect and are on the oldpipeline
           this.settings.unset('stateToken');
           var trans = this.options.appState.get('introspectSuccess');
           var transError = this.options.appState.get('introspectError');
-          if (trans || transError) {
-            if (trans && trans.data) {
-              return Q.resolve(trans);
-            } else {
-              return Q.reject(transError);
-            }
+          if (trans && trans.data) {
+            return Q.resolve(trans);
+          } else {
+            return Q.reject(transError);
           }
         } else if (token) {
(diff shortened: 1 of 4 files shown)


nikhilvenkatraman-okta:
It will hit https://github.com/okta/okta-signin-widget/blob/master/src/util/BaseLoginRouter.js#L144 where it checks that stateToken is not set, so it wont load the refreshAuthStateController but instead it loads defaultAuth route.

haishengwu-okta:
stateToken may be used for bootstrap widget to MFA page directly. would that be broken?

nikhilvenkatraman-okta:
ah, good catch. Looks like it is breaking flows where we bootstrap stateToken on oldPipeline.

src/RefreshAuthStateController.js
```diff
@@ -27,10 +27,12 @@ define(['q', 'okta', 'util/FormController'], function (Q, Okta, FormController)
       var token = this.options.token;
       this.model.startTransaction(function (authClient) {
         appState.trigger('loading', true);
-        if (token && (token === this.settings.get('stateToken'))) {
-          // widget bootstrapped with statetoken via settings
-          // check if stateToken in url is same as settings
-          //unset stateToken to prevent the baseloginrouter controller from calling it again on render
+        if (token && (token === this.settings.get('stateToken') ||
+          this.options.appState.get('introspectSuccess'))) {
```
haishengwu-okta: wondering if `this.options.appState.get('introspectSuccess')` alone is sufficient ?

nikhilvenkatraman-okta: yeah, I was hesitant to remove the other check which was added to fix a mobile flows, that hits the widget with a stateToken in the url.

haishengwu-okta:
notes: this is the last PR to fix the mobile flow.

src/RefreshAuthStateController.js
nikhilvenkatraman-okta: We will have to check for `this.options.appState.get('introspectSuccess') || this.options.appState.get('introspectError')` otherwise in case of an error during introspect, it'll make another request to authn API.

test/unit/spec/v2/OktaSignInWidget_spec.js
```diff
@@ -38,11 +38,14 @@ Expect.describe('OktaSignIn initialization', function () {
...
     });
+    spyOn(signIn.authClient.tx, 'resume').and.callFake(function () {
+      return options.response;
+    });
     spyOn($, 'ajax');
     signIn.renderEl({ el: $sandbox });
     return Q({});
   }
-  Expect.describe('Introspects token on load', function () {
+  Expect.describe('Calls Introspect API on load for new pipelione', function () {
```
ruthadrineda-okta: typo 🤓

PR #1013, opened 2020-01-15
build: add App SSO Extension authenticator approach

jhe-okta:
OKTA-270655

Once SIW receive the response with apple sso extension authenticator method, post the form right away to proceed.

authenticator-siw-app-sso-extension

### DIFF
--- a/playground/mocks/idp/idx/sso-extension.js
+++ b/playground/mocks/idp/idx/sso-extension.js
@@ -1,5 +1,5 @@
 const templateHelper = require('../../../config/templateHelper');
 
 module.exports = templateHelper({
-  path: '/idp/idx/authenticators/sso_extension/transactions/123/verify',
+  path: '/idp/idx/authenticators/sso_extension/transactions/:transactionId/verify',
 });
--- a/playground/mocks/idp/idx/sso-extension.js
+++ b/playground/mocks/idp/idx/sso-extension.js
@@ -1,0 +1,5 @@
+const templateHelper = require('../../../config/templateHelper');
+
+module.exports = templateHelper({
+  path: '/idp/idx/authenticators/sso_extension/transactions/123/verify',
+});
--- a/src/v2/view-builder/views/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/DeviceChallengePollView.js
@@ -30,6 +30,10 @@ const Body = BaseForm.extend({
     this.stopPolling();
   },
 
+  render () {
+    return BaseForm.prototype.render.apply(this, arguments);
+  },
+
   doChallenge () {
     const deviceChallenge = this.options.appState.get(
       this.deviceChallengePollRemediation.relatesTo
--- a/src/v2/view-builder/views/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/DeviceChallengePollView.js
@@ -30,10 +30,6 @@ const Body = BaseForm.extend({
     this.stopPolling();
   },
 
-  render () {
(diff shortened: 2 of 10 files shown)


src/v2/view-builder/views/SSOExtensionView.js
```diff
@@ -0,0 +1,21 @@
...
+const Body = BaseForm.extend({
+  noButtonBar: true,
+
+  className: 'ion-form sso-extension-challenge',
+
+  title: loc('signin', 'login'),
+
+  initialize () {
+    BaseForm.prototype.initialize.apply(this, arguments);
+    this.add('<div class="spinner"></div>');
```
nikhilvenkatraman-okta: How does this spinner get removed after the API request is done? 

Additionally I think currently the spinner is pretty basic, but we should move this out to some util file and just call a `showSpinner` method from any view that wants to show a spinner. In the future when we have more stuff built into the spinner it might be helpful. We should also have a scss file for the all the styles specific to the spinner. If we cant tackle it now, maybe create a JIRA for it ?

jhe-okta: It's not just the spinner get removed, once the form submit responds we shows a new view accordingly and the old view is removed. It's just a random example here that the identity is responded thus Identity view is displayed after. JIRA created: https://oktainc.atlassian.net/browse/OKTA-272203.

playground/mocks/idp/idx/sso-extension.js
```diff
@@ -0,0 +1,5 @@
+const templateHelper = require('../../../config/templateHelper');
+
+module.exports = templateHelper({
+  path: '/idp/idx/authenticators/sso_extension/transactions/123/verify',
```
haishengwu-okta: we shall be able to write path like `.../sso_extension/transactions/:transactionId/verify` instead of hard code id.

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -30,6 +30,10 @@ const Body = BaseForm.extend({
     this.stopPolling();
   },
 
+  render () {
```
haishengwu-okta: why need this override?

test/testcafe/spec/SSOExtensionView_spec.js
```diff
@@ -0,0 +1,32 @@
...
+  const ssoExtensionPage = new BasePageObject(t);
+  await ssoExtensionPage.navigateToPage();
+  await t.expect(logger.count(
+    record => record.response.statusCode === 200 &&
+    record.request.url.match(/introspect/)
+  )).eql(1);
+  await t.expect(logger.count(
+    record => record.response.statusCode === 200 &&
+    record.request.method === 'post' &&
+    record.request.url.match(/verify/)
```
haishengwu-okta: - a little more explicit like match `authenticators/sso_extension/transactions/123/verify`?
- or we could change the assertion to something like

```
expect logger's size is 2
expect logger[0] status code is 200 
expect logger[0] url is xxxxx
// same for logger[1]

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: No need. Forgot to remove the debugging code :p

src/v2/view-builder/views/SSOExtensionView.js
nikhilvenkatraman-okta: ah, gotcha. Yeah, it might be best to have control over when the spinner is getting removed.

PR #1014, opened 2020-01-16
fix: add challenge request body for authenticator loopback server

jhe-okta:
OKTA-272360

Screen Shot 2020-01-16 at 9 35 56 AM

### DIFF
--- a/src/v2/view-builder/views/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/DeviceChallengePollView.js
@@ -79,7 +79,7 @@ const Body = BaseForm.extend({
       return request({
         url: getAuthenticatorUrl('challenge'),
         method: 'POST',
-        body: JSON.stringify({ challengeRequest }),
+        data: JSON.stringify({ challengeRequest }),
         timeout: 3000 // authenticator should respond within 3000ms for challenge request
       });
     };
--- a/test/testcafe/spec/DeviceChallengePollView_spec.js
+++ b/test/testcafe/spec/DeviceChallengePollView_spec.js
@@ -5,7 +5,7 @@ import loopbackChallengeNotReceived from '../../../playground/mocks/idp/idx/data
 
 let failureCount = 0;
 
-const logger = RequestLogger(/introspect|probe|challenge/);
+const logger = RequestLogger(/introspect|probe|challenge/, { logRequestBody: true, stringifyRequestBody: true });
 
 const mock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/introspect')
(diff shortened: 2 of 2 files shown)


test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -49,6 +49,11 @@ test
       record => record.response.statusCode === 200 &&
       record.request.url.match(/introspect|6512/)
     )).eql(3);
+    await t.expect(logger.count(
+      record => record.response.statusCode === 200 &&
+      record.request.url.match(/challenge/) &&
+      record.request.body.match(/challengeRequest":"eyJraWQiOiI1/)
```
haishengwu-okta: are we able to run `equal` instead of match ?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -79,7 +79,7 @@ const Body = BaseForm.extend({
       return request({
         url: getAuthenticatorUrl('challenge'),
         method: 'POST',
-        body: JSON.stringify({ challengeRequest }),
+        data: JSON.stringify({ challengeRequest }),
```
haishengwu-okta: 🤦‍♂

test/testcafe/spec/DeviceChallengePollView_spec.js
jhe-okta: The challengeRequest is very long.

haishengwu-okta: what if you don't enable `stringifyRequestBody`, would `request.body` be an simple object  ?
```
{ challengRequest: xxxx }
```

jhe-okta: It will be a [Buffer](https://nodejs.org/api/buffer.html).

src/v2/view-builder/views/DeviceChallengePollView.js
nikhilvenkatraman-okta: guess this is needed because we are using jquery to make the post ?

jhe-okta: The wrong key was used before, thus no data was sent in the body.

nikhilvenkatraman-okta: yeah, I wonder if its because we are using jquery.ajax. I see that we set `body` in [httpClient](https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/httpClient.js#L28) that internally uses [cross-fetch](https://github.com/lquixada/cross-fetch). Additional [fetch API ](https://github.github.io/fetch/) doc

PR #1015, opened 2020-01-16
bump 3.6.2

haishengwu-okta:


commitlint.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   extends: ['@commitlint/config-conventional'],
   rules: {
     'references-empty': [2, 'never'],
+    'header-max-length': [2, 'always', 80]
```
nikhilvenkatraman-okta: Why are we just adding this change with this version bump ?

haishengwu-okta: Other commits merged to 3.6 got long commit message header. (when merged via github, it wont run commitlint.)
I cannot even push the bump up topic branch from my local cause of those commits violate commitlint rule `max-header-length`.
I could fix `3.6` and force pushed from my local then run version revive. I figured this is hacky. So I fixed the commitlint rule.

nikhilvenkatraman-okta: gotcha, thanks for the explanation 👍

PR #1017, opened 2020-01-17
feat: add support for email enrollment

haishengwu-okta:
- OKTA-258585 (UX design doc is linked in the jira)
- Demo: https://okta.box.com/s/jkje8krtd6wp7kuwn74xwl4xiemvn5qp


package.json
```diff
@@ -80,7 +80,7 @@
     "axios": "^0.19.0",
     "babel-core": "6",
     "babel-loader": "7",
-    "babel-plugin-istanbul": "^5.2.0",
+    "babel-plugin-istanbul": "4.1.6",
```
haishengwu-okta: Apparently `test:coverage` has been broken for while since the incompatibility on bable
- siw using babel 6
- plugin-istanbul 5 is using babel 7

Hence rollback to old version.

yarn.lock
```diff
@@ -27,7 +27,7 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/generator@^7.4.0", "@babel/generator@^7.4.4":
+"@babel/generator@^7.4.4":
```
gowthamidommety-okta: were these changes also made because of broken test:coverage? Should we break this out into a separate PR?

haishengwu-okta: I think so. sounds good to break into a separate PR.

haishengwu-okta: https://github.com/okta/okta-signin-widget/pull/1018

haishengwu-okta:
@mauriciocastillosilva-okta 
- Yes it's the UX we like to have (for both email enroll and verification) See the UX design doc linked in the JIRA
- Good question on the `send again`. I did ask the question in the design wiki but no response yet.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -330,6 +331,15 @@ enroll.password.setup = Select a password
 enroll.sms.setup = Receive a code via SMS to authenticate
 enroll.sms.try_again = The number you entered seems invalid. If the number is correct, please try again.
 
+# Enroll Email
+enroll.email.title = Set up Email Authentication
+enroll.email.send.description = Send a verification code to your registered email.
+enroll.email.save = Send me the code
+enroll.activate.email.description.1 = A verification code was sent to
```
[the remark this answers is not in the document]
haishengwu-okta: does that imply we'll need to use `{{{` to avoid escape the DOM element `span` ?

playground/config/responseConfig.js
```diff
@@ -1,78 +1,96 @@
...
+  '/idp/idx/authenticators/sso_extension/transactions/123/verify': [
+    'identify', // as a signal of success
+  ],
+};
+
+module.exports = {
+  // TODO: some testcafe test rely on setting mocks to idx
+  // which is apparently bad. ideally, testcafe test shall pass even
+  // no mocks has been config.
+  mocks: idx,
```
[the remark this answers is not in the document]
haishengwu-okta: - the mock is only intend for manual testing hence could be set to whatever we want to
- however as what the TODO says, there are some testcafe tests (for idx scenarios) are implicitly rely on the mock setup. we definitely shall address it soon.

haishengwu-okta:
@mauriciocastillosilva-okta 
just got new design regarding the `send again` and I'm thinking to do some refactor separately. how do you think?

packages/@okta/i18n/src/properties/login.properties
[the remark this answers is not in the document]
mauriciocastillosilva-okta: yeah, unfortunately that's the side effect

src/EnrollActivateEmailController.js
```diff
@@ -0,0 +1,115 @@
...
+      formChildren: [
+        // message
+        FormType.View({
+          View: Okta.View.extend({
+            className: 'enroll-activate-email-content',
+            attributes: {
+              'data-se': 'enroll-activate-email-content',
+            },
+
+            // Why use `{{{` for the first description?
```
mauriciocastillosilva-okta: nit: `for description?` (there are no longer description 1 and 2, it's only just once description now

test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_email.js
```diff
@@ -0,0 +1,69 @@
+define({
+  "status": 200,
+  "responseType": "json",
+  "response": {
```
mauriciocastillosilva-okta: not for this PR, but maybe something to think about: could we import `playground/mocks/api/v1/authn/data/enroll-activate-email.json` here instead of writing the json response again? Would that be a good idea or are there some side effects?

test/unit/spec/EnrollEmail_spec.js
```diff
@@ -0,0 +1,180 @@
...
+      });
+      var successSpy = jasmine.createSpy('successSpy');
+      var router = new Router({
+        el: $sandbox,
+        baseUrl: baseUrl,
+        authClient: authClient,
+        globalSuccessFn: successSpy,
+        'features.router': true
+      });
+      // router.on('afterError', afterErrorHandler);
```
mauriciocastillosilva-okta: should we remove this line?

test/unit/spec/EnrollEmail_spec.js
```diff
@@ -0,0 +1,180 @@
...
+        });
+    }
+
+    itp('display start enroll email form and is able to send code to email', function () {
+      return setup(xhrEnrollEmail)
+        .then(function (test) {
+          expect(test.form.titleText()).toBe('Set up Email Authentication');
+          expect(test.form.enrollEmailContent()).toBe('Send a verification code to your registered email.');
+          expect(test.form.submitButtonText()).toBe('Send me the code');
+          return test;
```
mauriciocastillosilva-okta: do we really need this `return`? Can we get away deleting lines 70-72? Asking since there are no `waitFor`'s here

haishengwu-okta: we could. I was trying to separate assertions: assert form elements v.s. ajax call.

mauriciocastillosilva-okta: either works. I was just curious if there was some technical reason.

test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_email.js
haishengwu-okta: Well, it's good idea and I had thought about it but didn't find time to do some investigation because I'm not sure
- whether I could import the JSON directly via AMD module syntax? It used to need some static loader like `json!/path/to/file`. 
- Or would webpack json loader plugin be helpful?
- otherwise, we may have to convert AMD to ES6 import style.

test/unit/spec/EnrollEmail_spec.js
haishengwu-okta: no technical reason at all.

test/unit/spec/EnrollEmail_spec.js
```diff
@@ -0,0 +1,182 @@
...
+        .then(function () {
+          router.refreshAuthState('dummy-token');
+          return Expect.waitForEnrollChoices();
+        })
+        .then(function () {
+          router.enrollEmail();
+          return Expect.waitForEnrollEmail({
+            router: router,
+            authContainer: new AuthContainer($sandbox),
+            beacon: new Beacon($sandbox),
```
gowthamidommety-okta: I think the only test missing now is `displays the correct factorBeacon` similar to other factors _spec files.

haishengwu-okta: 🤦‍♂

PR #1018, opened 2020-01-21
test: fix test:coverage script error

haishengwu-okta:
OKTA-273340

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "axios": "^0.19.0",
     "babel-core": "6",
     "babel-loader": "7",
-    "babel-plugin-istanbul": "^5.2.0",
+    "babel-plugin-istanbul": "4.1.6",
     "babel-plugin-transform-runtime": "^6.23.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -80,7 +80,7 @@
     "axios": "^0.19.0",
     "babel-core": "6",
     "babel-loader": "7",
-    "babel-plugin-istanbul": "^5.2.0",
+    "babel-plugin-istanbul": "4.1.6",
```
swiftone: Since this is a version downgrade, we should probably note what the issue is so we can know when/if we can return to current versions.  

Putting that note here sounds sufficient.  What is the error this fixes and why was a downgrade needed?

nikhilvenkatraman-okta: From what I saw in the PR linked, this is just an incompatibility with babel where widget is using babel 6 but `plugin-istanbul version 5` is using babel 7.

haishengwu-okta: yeah, it's exact what Nikhil mentioned.

PR #1022, opened 2020-01-28
chore: upgrade duo sdk 

magizh-okta:
* Upgrading duo web sdk. Not much changes
    https://github.com/duosecurity/duo_web_sdk/commits/master/index.js

* Moved duo and typingdna to packages

duo

**Typing pattern is still sent**

Screen Shot 2020-02-04 at 4 41 18 PM

RESOLVES: OKTA-223082

### DIFF
--- a/packages/TypingDnaRecorder-JavaScript/typingdna.js
+++ b/packages/TypingDnaRecorder-JavaScript/typingdna.js
@@ -1,3 +1,9 @@
+/*
+ * Checking in as there is no npm module to consume. 
+ * Current version is 2.15.
+ * Repo - https://github.com/TypingDNA/TypingDnaRecorder-JavaScript
+ */
+
 /**
  * TypingDNA - Typing Biometrics JavaScript API
  * https://api.typingdna.com/scripts/typingdna.js
--- a/packages/duo_web_sdk/index.js
+++ b/packages/duo_web_sdk/index.js
@@ -1,6 +1,18 @@
 /**
+ * 
+ * This file has been modified to override doPostBack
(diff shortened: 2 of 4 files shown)


src/vendor/Duo-Web-v2.6.js
```diff
@@ -1,436 +0,0 @@
...
-    //     if (typeof submitCallback === "function") {
-    //         submitCallback.call(null, form);
-    //     } else {
-    //         form.submit();
-    //     }
-    // }
-
-    // OKTA: Overriding their doPostBack logic because we want to submit this
-    // through ajax, not a postBack
-    function doPostBack(response) {
```
magizh-okta: @haishengwu-okta  I guess we need to have it checked in for this reason.

haishengwu-okta: 🤦‍♂ 
Sure. Glad you find the issue this time!

I'm thinking another idea is to create a Duo wrapper to override the post back function
```
define(['duo_web_sdk'], function (Duo) {
  const function oktaDoPostBack = ...
  return Object.assign(Duo, {_doPostBack: oktaDoPostBack });
});
```
I think it shall work and how do you think?
The key point is to make future upgrade a little easier and correct.

magizh-okta: I tired ur suggestion but I guess we would have to comment out and modify the src file because the doPOstBAck is called within the lib with its internal name and not _doPostback


https://github.com/duosecurity/duo_web_sdk/blob/master/index.js#L408

PR #1026, opened 2020-01-30
fix: update error message for failed login

mauriciocastillosilva-okta:
Resolves: OKTA-275377

SIW - Unable to Sign in

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -44,7 +44,7 @@ error.username.required = Please enter a username
 error.password.required = Please enter a password
 
 # ERROR MESSAGE OVERRIDES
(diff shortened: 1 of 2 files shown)


cniesen:
Why was this change made?  "Unable to sign in" sounds extremely vague.  I as a user would think Okta is down or some other non user fixable issue.  Not something that tells me that my password is invalid.

Also would like to note the importance for this change to be mentioned in the release notes.  Error E0000004 is one of the most common error users will encounter so this change is a major impact on user experience.

robertjd:
Hi @cniesen, thanks for the comments.  E0000004 is overloaded with a few cases, it's also generically defined to not leak account state information to attackers.

If you prefer a different message, you can define the translation strings for `errors.E0000004` in your widget configuration.

PR #226, opened 2017-05-16
:seedling: Adds Email as a factor

gowthamidommety-okta:
Enables users to use Email as a factor for authentication

Resolves: OKTA-124988

PR #1029, opened 2020-02-03
Enhance the email verification pages

haishengwu-okta:
several demos
- https://okta.box.com/s/1vi04qrpj0dk6pgokf8u610t0lwarjf9
- https://okta.box.com/s/ckxq0ru884yz9cgj1lwk3v2acraqqy54
- https://okta.box.com/s/sih5gsdwhdqtve3ymp28jp4p88qyvtr2

links to initial email factor implementation
- https://github.com/okta/okta-core/pull/20836
- https://github.com/okta/okta-signin-widget/pull/226


assets/sass/modules/_email-factor.scss
```diff
@@ -0,0 +1,12 @@
+@mixin email-factor {
+  .resend-email-infobox {
+    margin-bottom: 15px;
```
santhoshbalakrishnan: Are we already using standard Odyssey spacing here or not yet?

playground/mocks/api/v1/authn/factors/verify/index.js
```diff
@@ -0,0 +1,5 @@
+const templateHelper = require('../../../../../../config/templateHelper');
```
santhoshbalakrishnan: Use Import?

src/MfaVerifyController.js
```diff
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-/* eslint complexity: [2, 21] max-statements: [2, 25] max-params: [2, 12]*/
+/* eslint complexity: [2, 21] max-statements: [2, 25] max-params: 0 */
 define([
```
santhoshbalakrishnan: Guess if we move to imports, we can just get rid of this?

assets/sass/modules/_email-factor.scss
haishengwu-okta: no we don't

playground/mocks/api/v1/authn/factors/verify/index.js
haishengwu-okta: oh, I see why you confused now. files under `playground/mocks` are intent for nodejs dyson mock server hence I choose the default CMD without extra ES setup. we were using node-8.
I think node-12 would probably be able to support import and will give an shot.
https://oktainc.atlassian.net/browse/OKTA-276371

src/MfaVerifyController.js
haishengwu-okta: correct.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -280,6 +280,17 @@ recovery.sms = SMS
 recovery.call = Voice Call
 recovery.smsOrCall = SMS or Voice Call
 
+# Email enrollment and verification
+send.email.code.save = Send me the code
```
mauriciocastillosilva-okta: nit: it is easier to identify strings if they have some common prefix, maybe something as simple as `email.*` or `email.factor.*`

src/views/mfa-verify/SendEmailAndVerifyCodeForm.js
```diff
@@ -0,0 +1,120 @@
...
+              this.render();
+            });
+        }
+      }
+    }),
+
+    initialize: function () {
+      Okta.Form.prototype.initialize.apply(this, arguments);
+
+      this.state.set('isEmailChallengeStatus', false);
```
mauriciocastillosilva-okta: can you use [isMfaChallenge](https://github.com/okta/okta-signin-widget/blob/master/src/models/AppState.js#L246) instead of adding this new `isEmailChallengeStatus`?

src/views/mfa-verify/SendEmailAndVerifyCodeForm.js
```diff
@@ -0,0 +1,120 @@
...
+            label: false,
+            'label-top': true,
+            placeholder: this.options.appState.get('rememberDeviceLabel'),
+            className: 'margin-btm-0',
+            name: 'rememberDevice',
+            type: 'checkbox'
+          });
+        }
+      } else {
+        this.add(Okta.View.extend({
```
mauriciocastillosilva-okta: I think this could be added initially to the form (not in `postRender` but in `initialize` or somewhere before render) so the Form always has an input. And then when you click "send" you already have in that event handler a call to `removeChildren`, so that can call can add the "code verification" inputs.
Or another idea could be to use `showWhen` in the inputs, and changing based on the appState `isMfaChallenge`...
I think what it is a little odd for me is that all inputs are added in postRender

src/views/ResendEmailView.js
```diff
@@ -0,0 +1,54 @@
+/*!
+ * Copyright (c) 2015-2019, Okta, Inc. and/or its affiliates. All rights reserved.
```
mauriciocastillosilva-okta: nice reusable view 👍 
nit: `2015-2020`?

test/unit/spec/MfaVerifyEmail_spec.js
```diff
@@ -0,0 +1,299 @@
...
+      });
+    });
+    itp('shows send email page', function () {
+      return setupEmail().then(function (test) {
+        expect(test.form.titleText()).toBe('Verify with Email Authentication');
+        expect(test.form.el('mfa-send-email-content').html()).toBe(
+          'Send a verification code to ' +
+            '<span class="mask-email">a...1@clouditude.net</span>.'
+        );
+        expect(test.form.submitButtonText()).toBe('Send me the code');
```
mauriciocastillosilva-okta: nit: can we check we don't show the answer input?
```
expect(test.form.answerField().length).toBe(0);
```

test/unit/spec/MfaVerifyEmail_spec.js
```diff
@@ -0,0 +1,299 @@
...
+          Expect.isJsonPost($.ajax.calls.argsFor(0), {
+            url: 'https://foo.com/api/v1/authn/factors/emailhp9NXcoXu8z2wN0g3/verify?rememberDevice=true',
+            data: {
+              passCode: '456123',
+              stateToken: 'testStateToken'
+            }
+          });
+        });
+    });
+    itp('calls verifyFactor', function () {
```
mauriciocastillosilva-okta: nit: `calls verify endpoint when form is submitted with a passcode`

src/views/mfa-verify/SendEmailAndVerifyCodeForm.js
haishengwu-okta: ah, that seems cool!

src/views/mfa-verify/SendEmailAndVerifyCodeForm.js
haishengwu-okta: - I think `showWhen` by default only works for Inputs but not normal Views.
- I can try to move the views creation to `initialize`

src/models/Factor.js
```diff
@@ -255,6 +255,14 @@ function (Okta, Q, factorUtil, Util, Errors, BaseLoginModel) {
       // we don't need passcode for email with magic link flow
       return !(this.options.appState.get('isIdxStateToken') && this.get('factorType') === 'email');
     },
+    resend: function () {
```
gowthamidommety-okta: Why was resend added here?

haishengwu-okta: Because email enroll and email verification are using same `ResendEmailView.js` to render resend view, which trigger resend by invoking `model.resend`.
However email enroll and email verification are using different Model, EnrollActivateEmailController.Model v.s. Factor.js hence need to update the Factor.js

src/views/mfa-verify/SendEmailAndVerifyCodeForm.js
```diff
@@ -0,0 +1,114 @@
...
+    attributes: {
+      'data-se': 'factor-email',
+    },
+    save: function () {
+      return this.options.appState.get('isMfaChallenge')
+        ? Okta.loc('mfa.challenge.verify', 'login')
+        : Okta.loc('email.button.send', 'login');
+    },
+
+    events: Object.assign({}, Okta.Form.prototype.events, {
```
mauriciocastillosilva-okta: curious, did `events: { submit: .... }` not work for some reason? Or is there anything in parent Form that we need here?

test/unit/spec/EnrollEmail_spec.js
```diff
@@ -15,9 +15,11 @@ define([
   'helpers/xhr/MFA_ENROLL_email',
   'helpers/xhr/MFA_ENROLL_ACTIVATE_email',
   'helpers/xhr/SUCCESS',
-], function (Q, Okta, OktaAuth, LoginUtil, Util, AuthContainer,
+  'helpers/xhr/SMS_RESEND_error',
```
mauriciocastillosilva-okta: nit: we could rename `SMS_RESEND_error` to `MFA_RESEND_error`

test/unit/spec/MfaVerifyEmail_spec.js
```diff
@@ -0,0 +1,300 @@
...
+      router.refreshAuthState('dummy-token');
+      return Expect.waitForMfaVerify()
+        .then(function () {
+          var factors = router.appState.get('factors'),
+              selectedFactor = factors.findWhere({ factorType: 'email' });
+          router.verify(
+            selectedFactor.get('provider'),
+            selectedFactor.get('factorType'));
+          return Expect.wait(() => {
+            return $sandbox.find('.mfa-verify-email').length;
```
mauriciocastillosilva-okta: I think we can add this to [helpers/util/Expect.js](https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/util/Expect.js#L302), so this wait will be just something like `return Expect.waitForVerifyEmail(test);`

src/views/mfa-verify/SendEmailAndVerifyCodeForm.js
haishengwu-okta: I think it would work at this case although it removes [other events](https://github.com/okta/okta-ui/blob/master/packages/courage/src/views/forms/BaseForm.js#L65) defined in the base form.
In general when I do override, I tend to be explicit and keep whatever else comes from Parent.

playground/mocks/api/v1/authn/factors/verify/index.js
haishengwu-okta: fyi: https://github.com/TypeStrong/ts-node/issues/922#issuecomment-571734880

PR #1030, opened 2020-02-04
feat: implements loopback server failfast for authenticator

yannongli-okta:
[OKTA-273979](https://oktainc.atlassian.net/browse/OKTA-273979)

demo: https://okta.box.com/s/5q2r72s1puoytrorobrmb4754gjnaa12

### DIFF
--- a/playground/config/responseConfig.js
+++ b/playground/config/responseConfig.js
@@ -55,13 +55,24 @@ const windowAuthnLoopback = {
   '/idp/idx/authenticators/poll': [
     'identify-with-device-probing-loopback', // 2
     'identify-with-device-probing-loopback', // 3
-    'identify-with-device-probing-loopback-challenge-not-received', // 4
-    'identify-with-device-launch-authenticator', // 6
-    'identify', // 7: as a signal of success
+    'identify', // 4: as a signal of success
+  ],
+};
+
+//Windows authenticator with loopback server failfast
+const windowAuthnLoopbackFailfast = {
+  '/idp/idx/introspect': [
+    'identify-with-device-probing-loopback',
+  ],
+  '/idp/idx/authenticators/poll': [
+    'identify',
   ],
   '/idp/idx/authenticators/okta-verify/launch': [
-    'identify-with-device-launch-authenticator', // 5
-  ]
+    'identify-with-device-launch-authenticator',
+  ],
+  '/idp/idx/authenticators/poll/cancel': [
+    'identify-with-device-probing-loopback-challenge-not-received',
+  ],
 };
 
 // Windows/Android authenticator with custom URI
--- a/src/v2/view-builder/views/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/DeviceChallengePollView.js
@@ -62,6 +62,7 @@ const Body = BaseForm.extend({
   doLoopback (authenticatorDomainUrl = '', ports = [], challengeRequest = '') {
     let currentPort;
     let foundPort = false;
+    let countFailedPorts = 0;
 
     const getAuthenticatorUrl = (path) => {
       return `${authenticatorDomainUrl}:${currentPort}/${path}`;
@@ -102,7 +103,12 @@ const Body = BaseForm.extend({
           }
         })
         .catch(() => {
-          Logger.error('Something unexpected happened during device probing.');
+          countFailedPorts++;
+          Logger.error(`Authenticator is not listening on port ${currentPort}.`);
+          if (countFailedPorts === ports.length) {
+            Logger.error('No available ports. Loopback server failed and polling is cancelled.');
+            this.options.appState.trigger('invokeAction', 'authenticatorChallenge.cancel-polling');
+          }
         });
     });
   },
--- a/playground/mocks/idp/idx/authenticators-poll-cancel.js
+++ b/playground/mocks/idp/idx/authenticators-poll-cancel.js
@@ -1,0 +1,5 @@
+const templateHelper = require('../../../config/templateHelper');
+
+module.exports = templateHelper({
+  path: '/idp/idx/authenticators/poll/cancel',
+});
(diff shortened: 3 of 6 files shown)


src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -102,7 +103,13 @@ const Body = BaseForm.extend({
           }
         })
         .catch(() => {
-          Logger.error('Something unexpected happened during device probing.');
```
haishengwu-okta: I'm wondering if this would work slightly better?
- logger error detail and increase failed port count in `catch` callback
- cancel polling in `finally` callback

test/testcafe/spec/DeviceChallengePollViewLoopbackFailfast_spec.js
```diff
@@ -0,0 +1,68 @@
...
+import DeviceChallengePollPageObject from '../framework/page-objects/DeviceChallengePollPageObject';
+import IdentityPageObject from '../framework/page-objects/IdentityPageObject';
+import identifyWithDeviceProbingLoopback from '../../../playground/mocks/idp/idx/data/identify-with-device-probing-loopback';
+import loopbackChallengeNotReceived from '../../../playground/mocks/idp/idx/data/identify-with-device-probing-loopback-challenge-not-received';
+
+let failureCount = 0;
+
+const logger = RequestLogger(/introspect|probe|challenge/, { logRequestBody: true, stringifyRequestBody: true });
+
+const mock = RequestMock()
```
haishengwu-okta: - do you need to mock `authenticate-polling` endpoint and assert it has been hit?
- feels like we could add extra tests to `DeviceChallengePollView_spec` instead of creating a new test file.

playground/config/responseConfig.js
```diff
@@ -64,6 +64,22 @@ const windowAuthnLoopback = {
   ]
 };
 
+//Windows authenticator with loopback server failfast
+const windowAuthnLoopbackFailfast = {
```
jhe-okta: With the failfast mock to cover the case of loopback server -> custom URI -> success (identify as a signal in this example), the original loopback server mock should be updated to loopback server -> success (identify). Like:
```
// Windows authenticator with loopback server (make sure to start authenticator mock server from a separate terminal using yarn mock:authenticator)
const windowAuthnLoopback = {
  '/idp/idx/introspect': [
    'identify-with-device-probing-loopback', // 1 (response order)
  ],
  '/idp/idx/authenticators/poll': [
    'identify-with-device-probing-loopback', // 2
    'identify-with-device-probing-loopback', // 3
    'identify', // 4: as a signal of success
  ]
};
```

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -102,7 +103,13 @@ const Body = BaseForm.extend({
...
+          countProbedPorts++;
+          if (countProbedPorts === ports.length) {
+            Logger.error('No available ports. Loopback server failed and polling is cancelled.');
+            this.options.appState.trigger('invokeAction', 'authenticatorChallenge.cancel-polling');
+          } else {
+            Logger.error('Something unexpected happened during device probing.');
```
jhe-okta: Can you change the log message to something like `Authenticator is not listening on port ${currentPort}.`?

test/testcafe/spec/DeviceChallengePollViewLoopbackFailfast_spec.js
```diff
@@ -0,0 +1,68 @@
+import { RequestLogger, RequestMock } from 'testcafe';
```
jhe-okta: I think the tests should be updated in DeviceChallengePollViewOktaVerify_spec.js. And also, the DeviceChallengePollView_spec.js should be updated as well to reflect the case of loopback server login successfully.

test/testcafe/spec/DeviceChallengePollViewLoopbackFailfast_spec.js
```diff
@@ -0,0 +1,68 @@
...
+          res.setBody(loopbackChallengeNotReceived);
+      } else {
+          res.setBody(identifyWithDeviceProbingLoopback);
+      }
+  })
+  .onRequestTo('http://localhost:2000/probe')
+  .respond(null, 500, { 'access-control-allow-origin': '*' })
+  .onRequestTo('http://localhost:6511/probe')
+  .respond(null, 500, { 'access-control-allow-origin': '*' })
+  .onRequestTo('http://localhost:6512/probe')
```
jhe-okta: For failfast approach, all the authenticator request should respond with error.

src/v2/view-builder/views/DeviceChallengePollView.js
yannongli-okta: @haishengwu-okta I tried and noticed some issues regarding you advice:
1. Since we iterate through several ports, this finally block must be outside of `.forEach`
2. If 1 is true, it's working good with the failfast scenario. But for successful connection with any port, there will still be a `/poll/cancel` api call.
3. Maybe we can refactor the logic to follow the advice, but i think that's out of scope on this PR and is kind of risky.
@jhe-okta Can you also take look here?

test/testcafe/spec/DeviceChallengePollViewLoopbackFailfast_spec.js
yannongli-okta: Sorry. It's a copy and paste error. I copied the wrong file after fork the repo.

test/testcafe/spec/DeviceChallengePollViewLoopbackFailfast_spec.js
yannongli-okta: New test is added to `DeviceChallengePollViewOktaVerify_spec.js`

PR #1033, opened 2020-02-05
feat: replaces usage of introspect, save & actions transformer to use idx-js

nikhilvenkatraman-okta:
First round of change to integrate idx-js library in the widget

- Added idx-js as a package temporarily by copying bundle file in the widget repo
- Replaced introspect and all xhr post calls to use idx-js helper methods (idx.start, idx.proceed)
- Removed usage of actionsTransformer 
- Updated actions names like `factor-recover`, `factor-resend` to match what is being sent inside `actions` object  within idx-js.

Resolves: OKTA-275026

@magizh-okta @brettritter-okta @jhe-okta


src/v2/controllers/FormController.js
```diff
@@ -71,21 +88,25 @@ export default Controller.extend({
...
       .then(resp => {
-        this.options.appState.trigger('remediationSuccess', resp.response);
+        this.options.appState.set('idx', resp);
+        this.options.appState.trigger('remediationSuccess', resp.rawIdxState);
       })
-      .catch(error => {
-        model.trigger('error', model, {'responseJSON': error}, true);
-        this.toggleFormButtonState(false);
+      .catch(errorTransaction => {
+        errorTransaction.then(error => {
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: This is fixed now. idx-js will return a value instead of a promise.

src/v2/controllers/FormController.js
```diff
@@ -71,21 +88,25 @@ export default Controller.extend({
 
   handleFormSave (model) {
     const formName = model.get('formName');
-    const actionFn = this.options.appState.get('currentState')[formName];
-    if (!_.isFunction(actionFn)) {
+    const idx = this.options.appState.get('idx');
+    const actionFn = idx['neededToProceed'][formName];
+    if (!actionFn) {
       model.trigger('error', `Cannot find http action for "${formName}".`);
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: Ideally we should never encounter this error.

src/v2/models/AppState.js
```diff
@@ -87,7 +87,11 @@ export default Model.extend({
 
     let currentViewState;
     if (!_.isEmpty(this.get('remediation'))) {
-      currentViewState = this.get('remediation').filter(r => r.name === currentFormName)[0];
+      currentViewState = this.get('remediation').filter(r => {
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: I updated this while I was debugging something. I can revert this.

src/v2/models/AppState.js
```diff
@@ -102,18 +106,23 @@ export default Model.extend({
...
     // `currentFormName` is default to first form of remediation object or nothing.
     resp.currentFormName = null;
 
-    if (!_.isEmpty(resp.currentState.remediation)) {
-      resp.currentFormName = resp.currentState.remediation[0].name;
+    if (idx.neededToProceed && idx.rawIdxState.remediation) {
+      resp.currentFormName = idx.rawIdxState.remediation.value[0].name;
+    }
+
+    if (idx.rawIdxState.success) {
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: This was previously coming inside `resp.currentState.remediation[0].name;`which got updated on line 120 where we are now looking at `idx.rawIdxState.remediation`. But success is not a remediation and hence it comes outside.

src/widget/OktaSignIn.js
```diff
@@ -30,13 +30,15 @@ var OktaSignIn = (function () {
...
-          .then(_.bind(function (response) {
+        Util.introspectToken(widgetOptions)
+          .then(_.bind(function (data) {
+            var response = data.rawIdxState;
             var isNewPipeline = checkResponseVersion(response);
             if (isNewPipeline) {
               Router = require('v2/WidgetRouter');
             }
             router = bootstrapRouter.call(this, Router, authClient, widgetOptions, renderOptions, successFn, errorFn);
+            router.appState.set('idx', data);
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: I am going to leave this as is, since I am going to be revisiting this logic soon for when I do the introspect API changes. I'll refactor it then.

src/v2/models/AppState.js
```diff
@@ -102,18 +106,23 @@ export default Model.extend({
   },
 
   setIonResponse (resp) {
+    const idx = this.get('idx');
     // Don't re-render view if new response is same as last.
     // Usually happening at polling and pipeline doesn't proceed to next step.
     // expiresAt will be different for each response, hence compare objects without that property
-    if (_.isEqual(_.omit(resp.__rawResponse, 'expiresAt'), _.omit(this.get('__rawResponse'), 'expiresAt'))) {
+    if (_.isEqual(_.omit(idx.rawIdxState, 'expiresAt'), _.omit(this.get('__rawResponse'), 'expiresAt'))) {
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: I think it does a deep comparison https://underscorejs.org/#isEqual. The `expiresAt` attribute does come at the top level in the API response.

src/util/Util.js
```diff
@@ -198,15 +199,15 @@ define(['q', 'okta', './Logger', './Enums'], function (Q, Okta, Logger, Enums) {
...
-  Util.introspectToken = function (authClient, widgetOptions) {
+  Util.introspectToken = function (widgetOptions) {
+    const domain = widgetOptions.baseUrl;
+    const stateHandle = widgetOptions.stateToken;
+    var trans = start({ domain, stateHandle });
     var deferred = Q.defer();
-    var trans = authClient.tx.introspect({
-      stateToken: widgetOptions.stateToken
-    });
     if (Q.isPromiseAlike(trans)) {
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: This is still true, because otherwise the promise wont get resolved or rejected. Looking at [this](https://github.com/kriskowal/q/wiki/API-Reference#qispromisealikevalue) doc, looks like it just checks whether the given value is a promise.

src/v2/BaseLoginRouter.js
```diff
@@ -114,7 +113,6 @@ export default Router.extend({
     // transform response
     const ionResponse = _.compose(
       uiSchemaTransformer,
-      actionsTransformer,
```
jhe-okta: I don't have the context of what idx.js does. Is the `actionsTransformer` now part of idx.js?

src/v2/controllers/FormController.js
```diff
@@ -48,17 +48,34 @@ export default Controller.extend({
   },
 
   invokeAction (actionPath = '') {
-    const actionFn = this.options.appState.getActionByPath(actionPath);
+    const idx = this.options.appState.get('idx');
+    if (idx['neededToProceed'][actionPath]) {
+      idx.proceed(actionPath, {}).then((resp) => {
```
jhe-okta: nit: 
```
idx.proceed(actionPath, {})
  .then(resp => {
```

src/v2/controllers/FormController.js
```diff
@@ -71,17 +88,18 @@ export default Controller.extend({
 
   handleFormSave (model) {
     const formName = model.get('formName');
-    const actionFn = this.options.appState.get('currentState')[formName];
-    if (!_.isFunction(actionFn)) {
+    const idx = this.options.appState.get('idx');
+    if (!idx['neededToProceed'][formName]) {
```
jhe-okta: Why don't we need both `idx['neededToProceed'][formName]` and `_.isFunction(actionFn)`?

src/v2/controllers/FormController.js
jhe-okta: Why is there no data passed in?

src/v2/models/AppState.js
```diff
@@ -102,18 +102,23 @@ export default Model.extend({
...
```
jhe-okta: 👍

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -107,7 +107,7 @@ const Body = BaseForm.extend({
           Logger.error(`Authenticator is not listening on port ${currentPort}.`);
           if (countFailedPorts === ports.length) {
             Logger.error('No available ports. Loopback server failed and polling is cancelled.');
-            this.options.appState.trigger('invokeAction', 'authenticatorChallenge.cancel-polling');
+            this.options.appState.trigger('invokeAction', 'authenticatorChallenge-cancel');
```
jhe-okta: Before this change, this action path is composed with `authenticatorChallenge` and the name value within the `cancel`. Does that mean we won't look at the name anymore with the change?

src/v2/BaseLoginRouter.js
nikhilvenkatraman-okta: yeah, we are now relying on the data that idx-js is giving us.

src/v2/controllers/FormController.js
nikhilvenkatraman-okta: Previously we just called `actionFn ` now we are doing `idx.proceed(actionPath, {})`. This is slightly different than calling idx.proceed in the `handleFormSave ` since this is not a remediation action and one of the out of bound actions like `cancel` `factor-resend` `factor-poll` etc hence there is no data other than the stateHandle which idx-js knows to send.

src/v2/controllers/FormController.js
nikhilvenkatraman-okta: Previously `actionFn ` pointed to `this.options.appState.get('currentState')[formName]` and now it points to `idx['neededToProceed'][formName]`. I have just removed the need to hold `idx['neededToProceed'][formName]` in a const and used it directly in the if block as per [this](https://github.com/okta/okta-signin-widget/pull/1033#discussion_r383384294) request.

src/v2/view-builder/views/DeviceChallengePollView.js
nikhilvenkatraman-okta: We should just look at the names of the actions that idx-js provides within the object and trigger events with those names accordingly.

packages/@okta/idx-js/idx.js
```diff
@@ -0,0 +1,2 @@
...t.name,rt(t)]})))),o=ot(t);return{remediations:e,context:o.context,actions:o.actions}},at=function(t){var r,n=it(t),e=n.remediations,i=n.context,c=n.actions,u={};m()(r=y()(e)).call(r,(function(t){var r=h()(t,2),n=r[0],e=r[1];u[n]=e.neededParams}));return{proceed:function(t){var r,n=arguments;return o.a.async((function(o){for(;;)switch(o.prev=o.next){case 0:if(r=n.length>1&&void 0!==n[1]?n[1]:{},e[t]){o.next=3;break}return o.abrupt("return",a.a.reject("Unknown remediation choice: [".concat(t,"]")));case 3:return o.abrupt("return",e[t](r));case 4:case"end":return o.stop()}}))},neededToProceed:u,actions:c,context:i,rawIdxState:t}};r.default={start:function(t){var r,n,e,i;return o.a.async((function(c){for(;;)switch(c.prev=c.next){case 0:if(r=t.domain,n=t.stateHandle){c.next=3;break}return c.abrupt("return",a.a.reject({error:"stateHandle is required"}));case 3:if(r){c.next=5;break}return c.abrupt("return",a.a.reject({error:"domain is required"}));case 5:return c.next=7,o.a.awrap(p({domain:r,stateHandle:n}).catch((function(t){return a.a.reject({error:"introspect call failed",details:t})})));case 7:return e=c.sent,i=at(e),c.abrupt("return",i);case 10:case"end":return c.stop()}}))}}}])}));
```
magizh-okta: may be add the commit id of  version we are consuming.

nikhilvenkatraman-okta: yes, I have a [PR](https://github.com/okta/okta-idx-js/pull/8) for a small idx-js fix. once that is merged I'll include that commit as a reference.

src/v2/controllers/FormController.js
jhe-okta: Interesting. So we are making assumption that only stateHandle is sent even though it can be a POST request? Maybe the IDX response can be improved, but even currently, in one of the device authenticator flow, we need to make POST request with challengeRequest data. We are not using the invokeAction to do so now, but it is because we need to use timeout option in the request (this should be fixable).

https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/idp/idx/data/identify-with-device-probing-loopback.json#L60

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: Make sense. But I wonder if the name within is redundant.

src/v2/controllers/FormController.js
jhe-okta: But never mind, the request is not one of those 'rel-form', so it is probably not following the convention.

webpack.test.config.js
```diff
@@ -7,9 +7,15 @@ var testConfig    = commonConfig('main-tests.js');
 var rootDir       = path.resolve(__dirname);
 var RemoveStrictPlugin = require( 'remove-strict-webpack-plugin' );
 var plugins = createPlugins({ isProduction: false });
+var webpack = require('webpack');
```
nikhilvenkatraman-okta: https://oktainc.atlassian.net/browse/OKTA-277852
cc: @aarongranick-okta

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -229,15 +231,65 @@ function (Widget, Expect, Logger, Util, $sandbox, idxResponse, Okta, Q, idx, Ide
...
+        return Q({
+          data: introspectResponse.response
+        });
+      });
+      signIn.renderEl({ el: $sandbox });
+      return Expect.wait(() => {
+        return ($('.primary-auth').length === 1);
+      });
+    }
+    Expect.describe('Introspects token and loads primary auth view for old pipeline', function () {
```
magizh-okta: can we add a test for error case ?

nikhilvenkatraman-okta: fixed

PR #1034, opened 2020-02-06
Fix testcafe test debt and update its eslint rule

haishengwu-okta:
- Refactoring existing testcafe test to be self contained instead of depending on the mock server config
- Change redirect URI to `localhost:3000/app/UserHome` instead of httpbin
- Update `.eslintrc` for test files

### DIFF
--- a/test/testcafe/spec/ChallengeFactorEmail_MagicLink_spec.js
+++ b/test/testcafe/spec/ChallengeFactorEmail_MagicLink_spec.js
@@ -1,4 +1,3 @@
-import IdentityPageObject from '../framework/page-objects/IdentityPageObject';
 import ChallengeFactorPageObject from '../framework/page-objects/ChallengeFactorPageObject';
 import { RequestMock, RequestLogger } from 'testcafe';
 import magicLinkReturnTab from '../../../playground/mocks/idp/idx/data/terminal-return-email';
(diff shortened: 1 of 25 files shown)


test/testcafe/spec/ChallengeFactorEmail_MagicLink_spec.js
```diff
@@ -1,57 +1,52 @@
...
-  .requestHooks(magicLinkExpiredMock)
-  (`challenge email factor with expired magic link`, async t => {
+  .requestHooks(magicLinkExpiredMock)(`challenge email factor with expired magic link`, async t => {
     const challengeFactorPageObject = await setup(t);
     const pageTitle = challengeFactorPageObject.getPageTitle();
     await t.expect(pageTitle).eql('This email link has expired. To resend it, return to the screen where you requested it.');
   });
 
-  test
-  .requestHooks(logger, resendEmailMock, magicLinkEmailSentMock)
```
jhe-okta: Is resendEmailMock not needed anymore?

test/testcafe/spec/ChallengeFactorPassword_spec.js
```diff
@@ -1,31 +1,29 @@
...
-  const identityPage = new IdentityPageObject(t);
-  await identityPage.navigateToPage();
-  await identityPage.fillIdentifierField('Challenge Password');
-  await identityPage.clickNextButton();
-  return new ChallengeFactorPageObject(t);
+  const challengeFactorPage = new ChallengeFactorPageObject(t);
+  await challengeFactorPage.navigateToPage();
+  return challengeFactorPage;
 }
-const getPageUrl = ClientFunction(() => window.location.href);
```
jhe-okta: 👍

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -1,4 +1,4 @@
-import { RequestLogger, RequestMock, Selector } from 'testcafe';
+import { RequestLogger, RequestMock } from 'testcafe';
```
jhe-okta: Why does this PR merge directly to master instead of 3.7? This fix is on 3.7 I believe.

test/testcafe/spec/ChallengeFactorEmail_MagicLink_spec.js
```diff
@@ -1,57 +1,52 @@
...
 import ChallengeFactorPageObject from '../framework/page-objects/ChallengeFactorPageObject';
 import { RequestMock, RequestLogger } from 'testcafe';
 import magicLinkReturnTab from '../../../playground/mocks/idp/idx/data/terminal-return-email';
 import magicLinkExpired from '../../../playground/mocks/idp/idx/data/terminal-return-expired-email';
 import magicLinkEmailSent from '../../../playground/mocks/idp/idx/data/factor-verification-email';
 
 const magicLinkReturnTabMock = RequestMock()
-  .onRequestTo('http://localhost:3000/idp/idx')
-  .respond(magicLinkReturnTab)
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
```
jhe-okta: Is this change of mocking API in test to make testcafe independent of mock server?

test/testcafe/spec/ChallengeFactorEmail_MagicLink_spec.js
haishengwu-okta: I merged resendEmailMock and magicLinkEmailSentMock.

test/testcafe/spec/ChallengeFactorEmail_MagicLink_spec.js
haishengwu-okta: Yes. the idea is test shall only depend on playground server but not mock server config (which is subject to change on manual testing use cases by developers)

test/testcafe/spec/DeviceChallengePollView_spec.js
haishengwu-okta: this is more refactor work and doesn't worth to go with patch.
we can deal with merge conflict (if any) when back port patch changes to master.

nikhilvenkatraman-okta: I think this is just test clean up which does not need to be patched immediately by putting it to `3.7`. PRs merged to master will go out at the end of the month when we release a minor version.

jhe-okta: Make sense. Thanks for the clarification!

test/testcafe/spec/ChallengeFactorPassword_spec.js
```diff
@@ -1,31 +1,29 @@
...
+import SuccessPageObject from '../framework/page-objects/SuccessPageObject';
 import ChallengeFactorPageObject from '../framework/page-objects/ChallengeFactorPageObject';
-import { ClientFunction, RequestMock } from 'testcafe';
+import { RequestMock } from 'testcafe';
 import factorRequiredPassword from '../../../playground/mocks/idp/idx/data/factor-verification-password';
 import success from '../../../playground/mocks/idp/idx/data/success';
 
 const mock = RequestMock()
-  .onRequestTo('http://localhost:3000/idp/idx')
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
```
nikhilvenkatraman-okta: Will these tests now directly start from  the factorRequiredPassword screen ? instead of starting from the identifier screen ?

haishengwu-okta: correct.

nikhilvenkatraman-okta: Hmmm is that a concern? Since we are not testing the e2e flow going from the identifier page to the select-factor/challenge page ? Ultimately its an end-2-end test which should ideally test then entire flow.

haishengwu-okta: good question. I dont have strong opinion either way. Here are what I'm thinking
- Given our implementation (the spiritual of IDX), the logic from Identify to Page B and Page B to Page C is same.
- Not all flow starts with Identify (e.g. device authenticator) which already cause a bit inconsistency.
- Unless the page needs specific data/state passed from Identify, we could start with the page directly. (e.g. widget could bootstrap to MFA page even today.) So the pattern for each test is to just test the page like DOM structure and click behavior of button/links (e.g. if the click would trigger page switch, shall also verify the change)

happy to chat more.

PR #1042, opened 2020-02-13
feat: replace XHR post to Apple Okta Verify approach with window loca…

jhe-okta:
…tion change

RESOLVES: OKTA-277722

As of iOS 13, XHR request cannot be intercepted by SSO extension. Form POST does not work for all the apps. The workaround is to change window location. The new location is from the remediation href. We are still working on finalizing the remediation object change. So the rel, method and value may be changed later on.

I keep the spinner in the UI and delay the window location change for half second because I observe an intermittent blank UI or a partially rendered SIW otherwise. 

authenticator-siw-app-sso-extension-window-location


haishengwu-okta:
> Form POST does not work for all the apps

Do you why and how it failed some apps?

playground/mocks/idp/idx/sso-extension.js
```diff
@@ -1,5 +1,7 @@
 const templateHelper = require('../../../config/templateHelper');
 
 module.exports = templateHelper({
-  path: '/idp/idx/authenticators/sso_extension/transactions/:transactionId/verify',
+  path: '/idp/idx/authenticators/sso_extension/transactions/123/verify',
```
haishengwu-okta: do you really need to change the `path` to have a constant transaction id?

playground/mocks/idp/idx/sso-extension.js
```diff
@@ -1,5 +1,7 @@
...
+  method: 'GET',
+  template: '<html>Verifying the device...the login flow will be resumed afterwardss</html>'
```
nikhilvenkatraman-okta: nit: `afterwardss`

src/v2/view-builder/views/SSOExtensionView.js
```diff
@@ -12,7 +13,7 @@ const Body = BaseForm.extend({
   initialize () {
     BaseForm.prototype.initialize.apply(this, arguments);
     this.add('<div class="spinner"></div>');
-    this.options.appState.trigger('saveForm', this.model);
+    setTimeout(() => Util.redirectWithFormGet(this.options.currentViewState.href), 500);
```
nikhilvenkatraman-okta: How did we arrive at this `500` milsec wait ? Can we add a comment for this ?

nikhilvenkatraman-okta:
The functionality does seem similar to the success use case. Where we look at the form name `success-redirect` thats inside the success object. https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/ViewFactory.js#L69.  It almost feels like we could call that object `redirect` ? or something more generic in the API so that it can be used for both success and `device-apple-sso-extension` use cases. Curious to hear @royalchan-okta 's thoughts.

royalchan-okta:
+1 to @haishengwu-okta 's question, what's the issue with form post? This will not work well under app-embed scenarios since the widget is not okta hosted.

nbories-okta:
Per our conversation, I would recommend we don't add our own delay simply to have the UI render and show the spinner. A user already has so much to go through across this experience that adding our own delay simply for visual experience doesn't seem needed. Without the delay the SIW UI will barely have time to render (beacon won't even show) that i think we should use the static text and not use the spinner for the macOS and iOS end user experience.

The SIW will evolve to not even show the beacon in the future, so i would recommend we use the redirect success screen @jinghu-okta showed me. The only feedback on that screen though would be to update the copy to say "You are being redirected"

screen_shot_2020-02-14_at_10 48 09_am


cc: @vincentvoong-okta some feedback on the SIW success redirect screen copy.

jhe-okta:
- updated the UI to use the SuccessView as the intermittent view before window location change
- set stateHandle in cookie (server will delete it later on)

authenticator-siw-app-sso-extension-window-location-success

src/v2/view-builder/views/SuccessView.js
```diff
@@ -8,10 +8,13 @@ const Body = BaseForm.extend({
   noButtonBar: true,
   initialize () {
     BaseForm.prototype.initialize.apply(this, arguments);
+    const currentViewState = this.options.appState.getCurrentViewState();
+    if (currentViewState.name === 'device-apple-sso-extension') {
+      document.cookie = `stateHandle=${this.options.appState.get('currentState').stateHandle};path=/`;
```
haishengwu-okta: - would be better to keep `SSOExtensionView.js` given this specific logic.
- please add comment regarding why the sso extension flow is so special.

jhe-okta: @haishengwu-okta , `SSOExtensionView.js` is back, but not extended from SuccessView as SuccessView.Body is not accessible.

src/v2/view-builder/views/SSOExtensionView.js
```diff
@@ -1,18 +1,21 @@
...
+// - XHR request is not interceptable
+// - form post is interceptable, but some app (Outlook) closes the app after
+// We may have a different approach/UX for EA
 const Body = BaseForm.extend({
   noButtonBar: true,
 
-  className: 'ion-form sso-extension-challenge',
-
-  title: loc('signin', 'login'),
+  title: 'You are being redirected',
```
haishengwu-okta: nit:
looks like it's recommendation from UX. would you mind to update `SuccessView.title` as well? maybe extra to login.properties.

src/v2/view-builder/views/SSOExtensionView.js
```diff
@@ -1,18 +1,21 @@
...
 
   initialize () {
     BaseForm.prototype.initialize.apply(this, arguments);
-    this.add('<div class="spinner"></div>');
-    this.options.appState.trigger('saveForm', this.model);
+    document.cookie = `stateHandle=${this.options.appState.get('currentState').stateHandle};path=/`;
```
haishengwu-okta: Just curious do we need to set expiration date for this cookie?

test/testcafe/spec/SSOExtensionView_spec.js
```diff
@@ -1,32 +1,39 @@
...
+7KYXmWXzP00bA8jmcGLb7i9bFwhjw9OBCgdNcqxKXMLmWQA0JZritRDR6u0ZcEjykca-eUCJtG5ISQOONs_lUBGL3Ezz6QsfWtW16\
+E9QJAVEwf06gULnbw5n6wpiAiDqa4HlkKP6K5-v1Y0XQ`;
 
 const mock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/introspect')
   .respond(identifyWithAppleSSOExtension)
-  .onRequestTo('http://localhost:3000/idp/idx/authenticators/sso_extension/transactions/123/verify')
-  .respond(success)
+  .onRequestTo(verifyUrl)
+  .respond('<html><h1>Sign in verified</h1></html>')
```
haishengwu-okta: nit: change to a bit more specific like `sign in verified at ssoextensionview_spec`

PR #1043, opened 2020-02-14
fix: use GET method to launch Okta Verify

jhe-okta:
RESOLVES: OKTA-277738

Screen Shot 2020-02-14 at 7 50 02 AM

### DIFF
--- a/src/v2/view-builder/views/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/DeviceChallengePollView.js
@@ -116,7 +116,6 @@ const Body = BaseForm.extend({
   doCustomURI () {
     return request({
       url: this.customURI,
-      method: 'POST',
     });
   },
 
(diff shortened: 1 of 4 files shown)


src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -116,7 +116,6 @@ const Body = BaseForm.extend({
   doCustomURI () {
     return request({
       url: this.customURI,
-      method: 'POST',
```
haishengwu-okta: shall we add `method` to the API response in order to make UI dumb ?
https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/idp/idx/data/identify-with-device-launch-authenticator.json#L70

jhe-okta: Yea, that's mentioned in some discussion. But no decision yet. I can follow up later.

haishengwu-okta: could you cut a JIRA and add to the code as comment?

jhe-okta: https://oktainc.atlassian.net/browse/OKTA-278014

Issue #1047, opened 2020-02-17
Feature Request: Allow overriding SaaS variables by adding a !default flag to existing variables


stueccles:
### I'm submitting a
  - [ ] bug report
  - [X] feature request

### Background info

I'm opening an issue on the same subject as  https://github.com/okta/okta-signin-widget/pull/608. 

Generally an SaaS library allows for overriding variables for custom theming by using the !default flag. The Okta live widget https://developer.okta.com/live-widget/ indicates this should be possible 

### Expected behavior

Expect variables set before importing Okta SaaS to be used in generating CSS.

```
$primary-button-bg-color: aqua;

@import "node_modules/@okta/okta-signin-widget/dist/sass/okta-sign-in.scss";

@debug $primary-button-bg-color;  // aqua

```

### What went wrong?

Variables are overriden by Okta defaults

```
$primary-button-bg-color: aqua;

@import "node_modules/@okta/okta-signin-widget/dist/sass/okta-sign-in.scss";

@debug $primary-button-bg-color;  // #007dc1
```

### Steps to reproduce

see above 

### Your environment

  - Okta Sign-In Widget Version: 3.6.2
  - Browser: All
  - OS: All
  - Language: JavaScript

swiftone:
Internal ref: OKTA-280944

MrLyfing:
Any update regarding this enhancement? That'd be highly appreciated to have this in place to ease custom theming with sass. 
As @stueccles mentioned, this issue was requested more than a year ago https://github.com/okta/okta-signin-widget/pull/608 which also includes a pull request but there were no update from the Okta team since.

gwvt:
Another request here. It should be possible to customize SASS for the sign-in widget.

PR #608, opened 2018-12-10
Add !default to variables

ryanrishi:
This allows users to set SCSS variables.

Without setting `!default`:
```scss
$primary-button-bg-color: aqua;

@import "okta-sign-in";
@import "okta-theme";

@debug $primary-button-bg-color;  // #007dc1
```

With setting `!default` on variables:
```scss
$primary-button-bg-color: aqua;

@import "okta-sign-in";
@import "okta-theme";

@debug $primary-button-bg-color;  // aqua
```

PR #1048, opened 2020-02-17
feat: adds saas variable theming

stueccles:
by adding !default to saas variables can import saas with overriden variables for building css

Resolves: #1047


haishengwu-okta:
notes: Duplicates with https://github.com/okta/okta-signin-widget/pull/608

PR #1050, opened 2020-02-19
fix: increase /probe timeout to 3000 ms for Android

yannongli-okta:
Resolves: OKTA-278180

Demo: https://okta.box.com/s/qipbtx3i6yr7cwrehh68ul9zwvms1vpx

Demo displayed timeouts for different platforms. Note that the console message in above video is just for testing purpose and is not included in this PR.

@jhe-okta @nikhilvenkatraman-okta @haishengwu-okta

### DIFF
--- a/src/v2/view-builder/views/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/DeviceChallengePollView.js
@@ -4,6 +4,7 @@ import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
 import BaseFooter from '../internals//BaseFooter';
 import Logger from '../../../util/Logger';
+import DeviceFingerprint from '../../../util/DeviceFingerprint';
 
 const request = (opts) => {
   const ajaxOptions = Object.assign({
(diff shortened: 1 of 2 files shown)


src/v2/util/UserAgentUtil.js
```diff
@@ -0,0 +1,9 @@
+import DeviceFingerprint from '../../util/DeviceFingerprint';
+
+const isAndroid = function () {
```
jhe-okta: Might as well add this to DeviceFingerprint

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -71,7 +72,9 @@ const Body = BaseForm.extend({
     const checkPort = () => {
       return request({
         url: getAuthenticatorUrl('probe'),
-        timeout: 1000 // if authenticator & its probing endpoint exist, it should respond within 1000ms
+        // if authenticator & its probing endpoint exist, it should respond within 1000ms
```
jhe-okta: IMO, the comment should gives context of what the change is for and any pending TODOs. In our case, port checking in Android is slower because that SSL handshake is needed, without increasing timeout, request may fail sometimes. While other platform (Windows) uses http only at the moment, response time is within 1000 millseconds. So once Windows switches to use https, we will face the same issue. At the same time, we need to investigate and see how to improve the response time.

src/v2/util/UserAgentUtil.js
nikhilvenkatraman-okta: Or just move that `getUserAgent` method into this file.

yannongli-okta: I prefer to add it to `DeviceFingerprint` since `getUserAgent` is necessary in `DeviceFingerprint` and moving it is quite a big change. 
But adding a simple android user agent check involves much smaller changes. I think we don't even need to add test to `DeviceFingerprint_spec.js`
@nikhilvenkatraman-okta @jhe-okta

PR #1053, opened 2020-02-19
fix: stop polling in device challenge view

jhe-okta:
once polling responds with error, also, remove spinner

OKTA-278071

authenticator-siw-poll-fail-stop

### DIFF
--- a/src/v2/view-builder/views/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/DeviceChallengePollView.js
@@ -1,10 +1,11 @@
 /* global Promise */
-import { _, $, loc } from 'okta';
+import { $, loc } from 'okta';
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
 import BaseFooter from '../internals//BaseFooter';
 import Logger from '../../../util/Logger';
 import DeviceFingerprint from '../../../util/DeviceFingerprint';
+import polling from './shared/polling';
 
 const request = (opts) => {
   const ajaxOptions = Object.assign({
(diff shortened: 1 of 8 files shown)


src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -13,127 +14,122 @@ const request = (opts) => {
...
+      this.startDevicePolling();
+    },
+
+    onPollingFail () {
+      this.$('.spinner').hide();
+      this.stopPolling();
+    },
+
+    remove () {
+      BaseForm.prototype.remove.apply(this, arguments);
```
nikhilvenkatraman-okta: when does this get called? When device polling succeeds and you render a new screen?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -13,127 +14,122 @@ const request = (opts) => {
...
+        foundPort = true;
+        return request({
+          url: getAuthenticatorUrl('challenge'),
+          method: 'POST',
+          data: JSON.stringify({ challengeRequest }),
+          timeout: 3000 // authenticator should respond within 3000ms for challenge request
+        });
+      };
+
+      const onFailure = () => {};
```
nikhilvenkatraman-okta: why do we have this empty onFailure method. Can we at least log something use in here ?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -13,127 +14,122 @@ const request = (opts) => {
...
+
+    initialize () {
+      BaseForm.prototype.initialize.apply(this, arguments);
+      this.listenTo(this.model, 'error', this.onPollingFail);
+      this.deviceChallengePollRemediation = this.options.appState.getCurrentViewState();
+      this.doChallenge();
+      this.startDevicePolling();
+    },
+
+    onPollingFail () {
```
nikhilvenkatraman-okta: 👍

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: Yes.
This actually needs improvement as I talked to Haisheng before. Currently, SIW is being too aggressively in rerendering the view, it simply compares the before and after IDX response object, as long as there's any diff, model triggers change event thus view rerender. This does not work well at least in polling view. Sometimes the only difference in responses between pollings is only the timestamp, we don't want the view to refresh in that case.

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: I can add some log, but practically it is never going to get there.

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -13,127 +14,124 @@ const request = (opts) => {
...
+        return request({
+          url: getAuthenticatorUrl('challenge'),
+          method: 'POST',
+          data: JSON.stringify({ challengeRequest }),
+          timeout: 3000 // authenticator should respond within 3000ms for challenge request
+        });
+      };
+
+      const onFailure = () => {
+        Logger.error(`Something unexpected happens on checking port ${currentPort}.`);
```
nikhilvenkatraman-okta: nit: `Something unexpected happened while we were checking port ${currentPort}`

Issue #742, opened 2019-07-01
Login fields do not indicate that they are required


Matthewundy:
:information_source: **If you have a question**, please post it on the [Okta Developer Forum](https://devforum.okta.com) instead. Issues in this repository are reserved for bug reports and feature requests.

### I'm submitting a
  - [x] bug report
  - [ ] feature request

### Background info

Login fields should have required attribute to support screen readers

### Expected behavior

![image](https://user-images.githubusercontent.com/12891920/60465747-c6996c80-9c49-11e9-87da-17ad889353cd.png)

![image](https://user-images.githubusercontent.com/12891920/60465804-dca72d00-9c49-11e9-8120-7bd4ac393c67.png)

### What went wrong?

Please provide log or error messages if applicable.

### Steps to reproduce

If the current behavior is a bug, please provide the steps to reproduce and a minimal demo if possible (you can use this template as a starting point: https://jsfiddle.net/m2gyLq4v/).

### Your environment

  - Okta Sign-In Widget Version: Latest
  - Browser: Any 
  - OS: Any
  - Language: Any

PR #1057, opened 2020-02-20
feat: add "required" property to appropriate input fields - OKTA-237727

shuowu-okta:
feat: Support `required` attribute for a11y

Support `required` attribute for userName and password fields in PrimaryAuthForm

Resolves: #742

### DIFF
--- a/src/views/primary-auth/PrimaryAuthForm.js
+++ b/src/views/primary-auth/PrimaryAuthForm.js
@@ -191,6 +191,22 @@ define([
       if (this.settings.get('features.trackTypingPattern')) {
         TypingUtil.track('okta-signin-username');
       }
-    }
+    },
+
+    // TODO: support a11y attrs in Courage - OKTA-279025
+    render: function () {
+      Okta.Form.prototype.render.call(this);
+
+      this.$el.attr('novalidate', 'novalidate');
+
+      [
+        this.getUsernameField().inputId, 
+        this.getPasswordField().inputId
+      ].forEach(inputId => {
+        const $el = this.$el.find(`#${inputId}`);
+        $el.prop('required', true);    
+      });
+    },
+
   });
 });
--- a/src/views/primary-auth/PrimaryAuthForm.js
+++ b/src/views/primary-auth/PrimaryAuthForm.js
@@ -23,6 +23,7 @@ define([
   return Okta.Form.extend({
     className: 'primary-auth-form',
     noCancelButton: true,
+    attributes: { 'novalidate': 'novalidate' },
     save: function () {
       if (this.settings.get('features.passwordlessAuth')) {
         return Okta.loc('oform.next', 'login');
@@ -119,7 +120,11 @@ define([
         input: TextBox,
         inputId: 'okta-signin-username',
         type: 'text',
-        disabled: this.options.appState.get('disableUsername')
+        disabled: this.options.appState.get('disableUsername'),
+        // TODO: support a11y attrs in Courage - OKTA-279025
+        render: function () {
+          this.$(`input[name=${this.options.name}]`).prop('required', true);
+        }
       };
 
       return userNameFieldObject;
@@ -145,7 +150,11 @@ define([
         name: 'password',
         inputId: 'okta-signin-password',
         validateOnlyIfDirty: true,
-        type: 'password'
+        type: 'password',
+        // TODO: support a11y attrs in Courage - OKTA-279025
+        render: function () {
+          this.$(`input[name=${this.options.name}]`).prop('required', true);
+        }
       };
       if (this.settings.get('features.showPasswordToggleOnSignInPage')) {
         passwordFieldObject.params = {};
@@ -191,21 +200,6 @@ define([
       if (this.settings.get('features.trackTypingPattern')) {
         TypingUtil.track('okta-signin-username');
       }
-    },
-
-    // TODO: support a11y attrs in Courage - OKTA-279025
-    render: function () {
-      Okta.Form.prototype.render.call(this);
-
-      this.$el.attr('novalidate', 'novalidate');
-
-      [
-        this.getUsernameField().inputId, 
-        this.getPasswordField().inputId
-      ].forEach(inputId => {
-        const $el = this.$el.find(`#${inputId}`);
-        $el.prop('required', true);    
-      });
     },
 
   });
--- a/test/unit/spec/PrimaryAuth_spec.js
+++ b/test/unit/spec/PrimaryAuth_spec.js
@@ -463,6 +463,7 @@ function (Q, OktaAuth, LoginUtil, Okta, Util, AuthContainer, PrimaryAuthForm, Be
           expect(username.length).toBe(1);
           expect(username.attr('type')).toEqual('text');
           expect(username.attr('id')).toEqual('okta-signin-username');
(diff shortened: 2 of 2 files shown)


src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -191,6 +191,22 @@ define([
...
+    // TODO: support a11y attrs in Courage - OKTA-279025
+    render: function () {
+      Okta.Form.prototype.render.call(this);
+
+      this.$el.attr('novalidate', 'novalidate');
+
+      [
+        this.getUsernameField().inputId, 
+        this.getPasswordField().inputId
+      ].forEach(inputId => {
```
aarongranick-okta: This approach (enumerating some inputs and finding them by ID) will work. However, you could also encapsulate this logic within the input definition by putting a `render()` method directly on the input defined in `getUsernameField` and `getPasswordField`. See https://github.com/okta/okta-signin-widget/blob/master/src/EnrollCallAndSmsController.js#L209 for an example. Doing it this way you won't need to find the input by ID, `this.$` will already be scoped to the individual input. This would help keep the logic for the input all together in one place.

src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -191,6 +191,22 @@ define([
...
         TypingUtil.track('okta-signin-username');
       }
-    }
+    },
+
+    // TODO: support a11y attrs in Courage - OKTA-279025
+    render: function () {
+      Okta.Form.prototype.render.call(this);
+
+      this.$el.attr('novalidate', 'novalidate');
```
aarongranick-okta: You can define attributes for the view by adding an `attributes` property to the Form definition: https://backbonejs.org/#View-attributes Here's an example: https://github.com/okta/okta-signin-widget/blob/master/src/AccountUnlockedController.js#L37

src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -191,6 +191,22 @@ define([
...
+
+      [
+        this.getUsernameField().inputId, 
+        this.getPasswordField().inputId
+      ].forEach(inputId => {
+        const $el = this.$el.find(`#${inputId}`);
+        $el.prop('required', true);    
+      });
+    },
```
aarongranick-okta: If attributes are added to Form definition and render/.prop logic is moved to Input definition, this render() method can be removed.

src/views/primary-auth/PrimaryAuthForm.js
shuowu-okta: done

src/views/primary-auth/PrimaryAuthForm.js
shuowu-okta: done

src/views/primary-auth/PrimaryAuthForm.js
shuowu-okta: done

src/views/primary-auth/PrimaryAuthForm.js
```diff
@@ -119,7 +120,11 @@ define([
         input: TextBox,
         inputId: 'okta-signin-username',
         type: 'text',
-        disabled: this.options.appState.get('disableUsername')
+        disabled: this.options.appState.get('disableUsername'),
+        // TODO: support a11y attrs in Courage - OKTA-279025
+        render: function () {
+          this.$(`input[name=${this.options.name}]`).prop('required', true);
```
nikhilvenkatraman-okta: Do we know if the `required` attribute is supported by all browsers ? SIW gets initialized inside embedded browsers in some cases.

mauriciocastillosilva-okta: Some comments:
- isn't the name always `username`? As in, we probably don't need to get it from `this.options.name`... or even `this.$el.prop('required', true);` could work
- should this be done in `postRender`? 
- Can we wait until OKTA-279025 is done? Instead of doing a one-off here, we could wait for courage to support this... it might take more time, so I'm fine if we don't want to wait, but wanted to ask just in case

shuowu-okta: > Do we know if the `required` attribute is supported by all browsers ? SIW gets initialized inside embedded browsers in some cases.

@nikhilvenkatraman-okta  As `novalidate` attribute is added to the form, validations are still depend on our own logic. If `required` is not supported, it will just be ignored by browser silently.

shuowu-okta: > * isn't the name always `username`? As in, we probably don't need to get it from `this.options.name`... or even `this.$el.prop('required', true);` could work
Looks like the input actually is a `div` container wraps `input`, so `this.$el.prop('required', true);` will only apply `required` to the container `div`. 
Per use the ref from `this.options.name`, it can free us from potential attr change, even by myself I don't really think it may happen :)

> * should this be done in `postRender`?
As I mentioned in the question above, the model refer to the input container instead of the input element, I still think a post render is needed here.

> * Can we wait until OKTA-279025 is done? Instead of doing a one-off here, we could wait for courage to support this... it might take more time, so I'm fine if we don't want to wait, but wanted to ask just in case
Good question, probably my project manager @chrismorris-okta can help answer this question better. IMO, as the changes are already ready, the engineer effort of waiting for `courage` support would not differ too much. But roll it out can make devs access to this a11y feature earlier.

shuowu-okta:
> We have some other forms where we have a username: forgot password and unlock account are the ones that come to mind. Maybe we can create followup jiras to fix that if we don't want to fix it in this PR

Internal follow up ticket: OKTA-279025

src/views/primary-auth/PrimaryAuthForm.js
mauriciocastillosilva-okta: sounds good about 1 and 3
about the `postRender`, I meant if this should be done in the `postRender` method instead of overriding `render`

shuowu-okta: > sounds good about 1 and 3
> about the `postRender`, I meant if this should be done in the `postRender` method instead of overriding `render`

Good catch! Didn't realize `postRender` hook been provided.

shuowu-okta: @mauriciocastillosilva-okta After debugging `postRender` is not available for inputs in `PrimaryAuthForm`. Let's keep using `render` here.

mauriciocastillosilva-okta: it should be supported. The input we are using is a `Textbox` which we already override in [views/shared/Textbox](https://github.com/okta/okta-signin-widget/blob/master/src/views/shared/TextBox.js) and we are using a `postRender` function there... I can take a look later today see if I can figure this out.

mauriciocastillosilva-okta: talked offline. This works fine: `render` here is called inside the `postRender` of the input, which is confusing.
We will create a follow up jira to change this later, either when OKTA-279025 is done (courage support for `required`) or when we more inputs to support `required`.

shuowu-okta: Ticket to handle render method: https://oktainc.atlassian.net/browse/OKTA-280596

PR #1075, opened 2020-03-04
fix: fixes text on reg complete screen

nikhilvenkatraman-okta:
Resolves: OKTA-269534

- Updated registration complete text to not show email value, as the user just entered it on the previous screen and it breaks when they try to refresh the page.
- Adjusted button to be a link as per UX feeedback

Before: 
Screen Shot 2020-03-03 at 11 15 07 AM

After: 

Screen Shot 2020-03-04 at 2 36 02 PM

@taylorlaubach-okta 
@magizh-okta 
@iancorcoran-okta


nikhilvenkatraman-okta:
The tests were checking for string contains, hence did not fail. I updated the tests and added an expect for the new text.

PR #1077, opened 2020-03-05
test: applitools integration for E2E tests

vijetmahabaleshwar-okta:
resolves: OKTA-278402


jmelberg-okta:
This is a feature that we intend to add support for in the future. Today, basic testing is captured within our core service test suites, so we're going to close this PR as "won't do".

PR #1080, opened 2020-03-06
chore: updates for okta-auth-js@3.0 - OKTA-276557

aarongranick-okta:
- Brings in [changes from AuthjS 3.0](https://github.com/okta/okta-auth-js/blob/master/CHANGELOG.md#300)
- PKCE is true by default
- PKCE will use responseMode=query by default
- jQuery removed from unit tests
- Q removed from AuthJS, replaced with standard promise
- showSignInToGetTokens changed:
  - getAccessToken is now `true` by default
  - `authorizationServerId` option has been removed from To specify a custom authorization server, use `authParams.issuer`. The issuer should be specified as a full URI, not just the server ID.
- minimum node version is 10

### DIFF
--- a/src/util/BaseLoginRouter.js
+++ b/src/util/BaseLoginRouter.js
@@ -10,11 +10,12 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-/* eslint max-params: [2, 15], max-statements: [2, 18] */
+/* eslint max-params: [2, 16], max-statements: [2, 18] */
 // BaseLoginRouter contains the more complicated router logic - rendering/
 // transition, etc. Most router changes should happen in LoginRouter (which is
 // responsible for adding new routes)
 define([
+  'q',
   'okta',
   './BrowserFeatures',
   'models/Settings',
@@ -31,7 +32,7 @@ define([
   'util/Bundles',
   'util/Logger'
 ],
-function (Okta, BrowserFeatures, Settings,
+function (Q, Okta, BrowserFeatures, Settings,
   Header, SecurityBeacon, AuthContainer, AppState, ColorsUtil, RouterUtil, Animations,
   Errors, Util, Enums, Bundles, Logger) {
 
@@ -225,7 +226,7 @@ function (Okta, BrowserFeatures, Settings,
       // First run fetchInitialData, in case the next controller needs data
       // before it's initial render. This will leave the current page in a
       // loading state.
-      this.controller.fetchInitialData()
+      return Q(this.controller.fetchInitialData())
         .then(_.bind(function () {
 
           // Beacon transition occurs in parallel to page swap
@@ -278,8 +279,7 @@ function (Okta, BrowserFeatures, Settings,
             oldController.remove();
             oldController.$el.remove();
           }
-        })
-        .done();
+        });
 
     },
 
--- a/src/util/BaseLoginRouter.js
+++ b/src/util/BaseLoginRouter.js
@@ -15,7 +15,6 @@
 // transition, etc. Most router changes should happen in LoginRouter (which is
 // responsible for adding new routes)
 define([
-  'q',
   'okta',
   './BrowserFeatures',
   'models/Settings',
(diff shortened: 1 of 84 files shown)


src/util/BaseLoginRouter.js
```diff
@@ -31,7 +32,7 @@ define([
   'util/Bundles',
   'util/Logger'
 ],
-function (Okta, BrowserFeatures, Settings,
+function (Q, Okta, BrowserFeatures, Settings,
```
nikhilvenkatraman-okta: Curious why this change is preferred over what we had before.

src/util/OAuth2Util.js
```diff
@@ -48,54 +48,50 @@ define(['okta', './Enums', './Errors', './Util'], function (Okta, Enums, Errors,
 
     var authClient = settings.getAuthClient(),
         options = settings.toJSON({ verbose: true }),
-        oauthParams = {},
-        extraOptions = {};
+        getTokenOptions = {};
```
nikhilvenkatraman-okta: nit: is it better to call this just tokenOptions? `getTokenOptions ` sounds more like a fn name.

test/e2e/templates/done.tpl
```diff
@@ -2,27 +2,42 @@
 // OIDC Redirect Flow - this is the page that is redirected to with
 // tokens in the parameters
 
+// PKCE cannot be enabled because the test app is a "web" type. OKTA-246000 
+var pkce = false;
```
nikhilvenkatraman-okta: Is this why we are setting `pkce: false` to a bunch of tests in the widget ?

test/unit/helpers/mocks/Util.js
```diff
@@ -59,75 +69,105 @@ function (Okta, Q, Duo, keys, wellKnown, wellKnownSharedResource) {
   };
 
   fn.mockAjax = function (responses) {
+    jasmine.Ajax.install();
```
nikhilvenkatraman-okta: 👏

test/unit/spec/ConsentRequired_spec.js
```diff
@@ -155,14 +155,14 @@ function (Okta, OktaAuth, LoginUtil, Util, ConsentRequiredForm, Expect, Router,
...
+          Util.resetAjaxRequests();
           test.setNextResponse(resSuccess);
           test.form.consentButton().click();
-          return Expect.waitForSpyCall($.ajax);
+          return Expect.waitForAjaxRequest();
         })
           .then(function () {
-            expect($.ajax.calls.count()).toBe(1);
-            Expect.isJsonPost($.ajax.calls.argsFor(0), {
+            expect(Util.numAjaxRequests()).toBe(1);
```
nikhilvenkatraman-okta: 👍

src/util/OAuth2Util.js
aarongranick-okta: Or maybe `authParams`? We use this name in other places to refer to the config passed to the authClient. The options we are building here correspond with this documentation: https://github.com/okta/okta-auth-js#authorize-options Maybe `authOptions`?

src/util/BaseLoginRouter.js
aarongranick-okta: I don't remember exactly. My guess is it has to do with unhandled promise rejection? I'll change it back and see what effect it has on the tests.

test/e2e/templates/done.tpl
aarongranick-okta: In the E2E tests doing OIDC flow, yes. They were built using implicit flow. I have a [task](https://oktainc.atlassian.net/browse/OKTA-246000) to add a SPA client ID to the test config. I would probably want to add this to the 4.0 branch before we merge to master.

src/util/OAuth2Util.js
nikhilvenkatraman-okta: either works

src/util/BaseLoginRouter.js
aarongranick-okta: There is a mixed usage of Q and standard promise. I am adding another commit which changes `fail` to `catch` in these cases so we can accept either.

src/views/shared/Header.js
```diff
@@ -139,7 +139,7 @@ function (Okta, Animations, LoadingBeacon) {
             removeBeacon(self);
             addBeacon(self, NextBeacon, selector, options);
           })
-          .done();
+          .done(); // TODO: can this be removed if fadeOut returns standard ES6 Promise?
```
nikhilvenkatraman-okta: Should we create a JIRA to refactor this later i.e use standard ES6 Promise ?

jhe-okta:
I will leave it to @nikhilvenkatraman-okta since he has the context.

PR #1101, opened 2020-03-20
fix: prevent additional polling request in OV enrollment

magizh-okta:
* When enrolling OV with push the polling is reinitiated again when user
    navigates from 'scan QR code' to 'send link via sms' and back to 'scan QR code'.

    * We are not navigating by invoking trrans.prev rather starting a new one with resume which doesn't cancel the ongoing poll. 

   Fixing this by invoking prev and then calling factor.enroll when user wants to get back to QR code view. 

RESOLVES: OKTA-285071

**OV enroll with push (polling occurs only every 6 seconds) - even when user switches back and forth.**
https://okta.box.com/s/ff2ttflt9zvuca0qmev4xnznb0my9dp3

back to QR code after submitting wrong number sms 
 https://okta.box.com/s/9ol00xxrj993wwpp0srfm28isw52v1g8

**OV enroll without push - just totp (no polling)** 

https://okta.box.com/s/mbtivljm6wce9q2627vvr10z96qjbxxt

### DIFF
--- a/src/views/enroll-factors/ManualSetupPushFooter.js
+++ b/src/views/enroll-factors/ManualSetupPushFooter.js
@@ -39,28 +39,17 @@ define(['okta', 'util/RouterUtil'], function (Okta, RouterUtil) {
         e.preventDefault();
         var goToFactor = _.partial(goToFactorActivation, this.options.appState);
         this.options.appState.unset('factorActivationType');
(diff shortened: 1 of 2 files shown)


magizh-okta:
Created a new one as travis had issues.

PR #1110, opened 2020-03-25
fix: prevent additional polling request in OV enrollment

magizh-okta:
* When enrolling OV with push the polling is reinitiated again when user
navigates from 'scan QR code' to 'send link via sms' and back to 'scan QR code'.

* We are not navigating by invoking trrans.prev rather starting a new one with resume which doesn't cancel the ongoing poll. 
Fixing this by invoking prev and then calling factor.enroll when user wants to get back to QR code view.

**Commit 2**
Cancel poll additinally when we navigate form scan QR code as well. 

RESOLVES: OKTA-285071

**Latest - with two poll cancel **
https://okta.box.com/s/aoilye03jzp6llxvvqxni2i5r0fa7cj6

OV enroll with push (polling occurs only every 6 seconds) - even when user switches back and forth.
https://okta.box.com/s/ff2ttflt9zvuca0qmev4xnznb0my9dp3

back to QR code after submitting wrong number sms
https://okta.box.com/s/9ol00xxrj993wwpp0srfm28isw52v1g8

OV enroll without push - just totp (no polling)

https://okta.box.com/s/mbtivljm6wce9q2627vvr10z96qjbxxt


magizh-okta:
@mauriciocastillosilva-okta  stopping the poll when moving away from Barcode view is an option and can help in a few cases. But we still need this fix in the case where the user 
navigated from QR code -> submits sms info ->   poll begins in new screen -> user clicks back -> User is now in manual setup screen but polling continues -> user clicks 'scan barcode' 

I am continuing to look into stopping the poll when we navigate from QR code view to handle on case where user navigates to sms and doesnt do anything.

mauriciocastillosilva-okta:
ah yeah, I meant in addition to what you have
But we already spoke offline :)

src/views/enroll-factors/BarcodeView.js
```diff
@@ -44,9 +44,25 @@ define([
     events: {
       'click [data-type="manual-setup"]': function (e) {
         e.preventDefault();
-        var url = RouterUtil.createActivateFactorUrl(this.model.get('__provider__'),
+        const url = RouterUtil.createActivateFactorUrl(this.model.get('__provider__'),
           this.model.get('__factorType__'), 'manual');
-        this.options.appState.trigger('navigate', url);
+        if (this.model.get('__factorType__') === 'push') {
```
mauriciocastillosilva-okta: for future, wonder if this same problem happens with other factors... just food for thought, nothing to change or add to this PR

magizh-okta: here this is to check for push or totp. I  dont think there is polling in other factors,

test/unit/spec/EnrollTotpController_spec.js
```diff
@@ -257,11 +258,10 @@ function (Okta, OktaAuth, LoginUtil, Util, DeviceTypeForm, BarcodeForm,
 
         // Start the enrollment
         test.form.selectDeviceType('APPLE');
+        // Submit would trigger enroll activate call and 1 poll.
         test.form.submit();
 
-        return tick(test)    // 1: submit enrollFactor
-          .then(function () { return tick(test); }) // 2: submit enrollFactor poll
-          .then(function () { return tick(test); }); // Final response tick
+        return tick(test); // To trigger next poll.
```
magizh-okta: Refactored this to be accurate now. The activate and first poll happens without any delay.

src/views/enroll-factors/BarcodeView.js
mauriciocastillosilva-okta: right, so does the same problem happen for totp? If so, we could probably file a jira for later...

magizh-okta:
@mauriciocastillosilva-okta  totp wont have the same prob because there is no polling in it. https://github.com/okta/okta-signin-widget/blob/master/src/BarcodeTotpController.js#L55

src/views/enroll-factors/BarcodeView.js
mpeng-okta: I believe we don't have the same problem for totp because we need only verify by passing in a totp in the browser client and sending that request.

We only have to poll for OVP because the request doesn't come from the browser client. Instead, it's from a separate mobile client, but we still need to advance the browser view, so we have to poll the server to check whether it received that out-of-band verification.

src/views/enroll-factors/BarcodeView.js
```diff
@@ -44,9 +44,25 @@ define([
...
+          // cancel the poll and navigate to manual setup.
+          this.model.doTransaction(function (transaction) {
+            return transaction.prev()
+              .then(function (trans) {
+                var factor = _.findWhere(trans.factors, {
```
rchild-okta: nit: It'd be useful for the future to add a comment here explaining why we're doing this - reading the code, I don't think it'd be obvious why we're in the push state, cancel it, and then go back to it.

Issue #633, opened 2019-02-21
Document the logoText option


nbarbettini:
The Widget allows you to set the alt text of the logo via `config.logoText`, but this isn't documented in the README.

aarongranick-okta:
Internal ref: OKTA-284243

PR #1117, opened 2020-04-02
chore: add logoText config to readme - OKTA-284243 (#633)

shuowu-okta:
## Description: 
Add `logoText` in `Baisc config options` section in ReadMe file.

### Issue:
- [OKTA-284243](https://oktainc.atlassian.net/browse/OKTA-284243)

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,7 +468,7 @@ var signIn = new OktaSignIn(config);
     logo: 'https://acme.com/img/logo.png'
     ```
 
-- **logoText:** Text for `alt` attribute of the logo image
+- **logoText:** Text for `alt` attribute of the logo image, logo text will only show up when logo image is not avaiable
 
     ```javascript
     // Text to describe the logo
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -467,6 +468,13 @@ var signIn = new OktaSignIn(config);
     logo: 'https://acme.com/img/logo.png'
     ```
 
+- **logoText:** Text for `alt` attribute of the logo image
```
pradeepdewda-okta, line 471: this is only when image is not displayed. Make it clear in the description or comments.

shuowu-okta:
> Wondering if the [header](https://github.com/okta/okta-signin-widget/blob/master/src/views/shared/Header.js#L82) should have a default alt logotext if no alt text is provided? What we have still seems to be better than having an empty alt text. No alt text at all is bad according to [this](https://davidwalsh.name/accessibility-tip-empty-alt-attributes) article.

Agree. How about create another ticket to track the issue, not sure if different translations needed.

PR #1123, opened 2020-04-07
feat: cancel transaction when Apple SSO Extension is not installed

jhe-okta:
## Description:

Add support to cancel transaction when Apple SSO Extension is not installed.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

siw-cancel-transaction

### Reviewers:
@aarongranick-okta @nikhilvenkatraman-okta 

### Issue:

- [OKTA-288101](https://oktainc.atlassian.net/browse/OKTA-288101)

### DIFF
--- a/playground/mocks/idp/idx/data/identify-with-no-sso-extension.json
+++ b/playground/mocks/idp/idx/data/identify-with-no-sso-extension.json
@@ -1,0 +1,61 @@
+{
+  "stateHandle": "02IAickZV0j0TL0JaJJUxe4GnX7YHTeZY9pCYe-pBZ",
+  "version": "1.0.0",
+  "expiresAt": "2020-15T17:28:11.000Z",
+  "step": "IDENTIFY",
+  "intent": "LOGIN",
+  "remediation": {
+      "type": "array",
+      "value": [
+          {
+                "rel": ["create-form"],
+                "name": "cancel-transaction",
+                "href": "http://localhost:3000/idp/idx/authenticators/sso_extension/transactions/456/verify/cancel",
+                "method": "POST",
+                "accepts": "application/ion+json;okta-version=1",
+                "produces": "application/ion+json;okta-version=1",
+                "value": [
+                    {
+                       "name": "stateHandle",
+                       "value": "02IAickZV0j0TL0JaJJUxe4GnX7YHTeZY9pCYe-pBZ",
+                       "required": true,
+                       "visible": false
+                    }      
+                ]      
+          }
+      ]
+  },
+  "cancel": {
+      "rel": ["create-form"],
+      "name": "cancel",
+      "href": "http://localhost:3000/idp/idx/cancel",
+      "method": "POST",
+      "accepts": "application/ion+json;okta-version=1",
+      "produces": "application/ion+json;okta-version=1",
+      "value": [
+          {
+              "name": "stateHandle",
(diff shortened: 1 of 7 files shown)


playground/mocks/idp/idx/data/identify-with-no-sso-extension.json
```diff
@@ -0,0 +1,61 @@
...
+  "version": "1.0.0",
+  "expiresAt": "2020-15T17:28:11.000Z",
+  "step": "IDENTIFY",
+  "intent": "LOGIN",
+  "remediation": {
+      "type": "array",
+      "value": [
+          {
+                "rel": ["create-form"],
+                "name": "cancel-transaction",
```
nikhilvenkatraman-okta, line 12: what is the difference between this form with name `cancel-transaction` and the `cancel` action that you get in the response ?

royalchan-okta: this form name should actually be more specific, since it's sort of globally unique, something like `cancel-sso-extension`. Would need a backend change as well.

src/v2/view-builder/ViewFactory.js
```diff
@@ -36,6 +36,9 @@ const VIEWS_MAPPING = {
   'device-apple-sso-extension': {
     [DEFAULT]: SSOExtensionView,
   },
+  'cancel-transaction': {
+    [DEFAULT]: SSOExtensionView,
```
pradeepdewda-okta, line 40: can we do something better here, where different keys have same values ?? Can we reuse existing ?

pradeepdewda-okta: for instance `device-apple-sso-extension` and `cancel-transaction` use same View.

playground/mocks/idp/idx/data/identify-with-no-sso-extension.json
yannongli-okta: The original idea is to make it to cancel authenticator transaction. So maybe we can rename it as cancel-authenticator-transaction

santoshmale-okta: @royalchan-okta my reasoning behind naming it as `cancel-transaction` is to keep it generic so that it could be used for any future use cases where we want to cancel transactions like canceling any future authenticator/identification transactions etc. Do you think `transaction` is kind of an internal detail that SIW need not have to know and is that the reason to make it specific?

royalchan-okta: Hm, that's actually a good point, we could retrofit this form to anything by changing the url. Alright, I'm convinced we can leave it this way for now as long as there isn't issues with the SIW views

jhe-okta: Thanks for the feedback. Will keep the remediation name as it is.

src/v2/view-builder/ViewFactory.js
jhe-okta: @pradeepdewda-okta the API design has been reviewed and remediations are named differently on purpose. They are both for SSO Extension approach in different step of the flow. They do have the same appearance and need the same trigger from the view level thus share the same view. But what they actually do is different, the idx actions may have different method/data.

pradeepdewda-okta: I was trying to suggest what the code is doing at line 89 in this file to get the step view. If it's not feasible then it's fine.

PR #1138, opened 2020-04-13
chore: remove JSONP logic OKTA-279865

aarongranick-okta:
## Description:

Replaces JSONP with JSON. This is part of the CSP effort. Since this is a breaking change, it will go into the 4.0 release.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-290150](https://oktainc.atlassian.net/browse/OKTA-290150)


aarongranick-okta:
@haishengwu-okta Correct, JSON will be loaded from CDN. We  have confirmed that necessary CORS headers are set in production

aarongranick-okta:
@haishengwu-okta I have confirmed that the cross-domain XHR does work if using fetch. The issue appears to be with the extra header being added in courage: https://github.com/okta/okta-ui/blob/master/packages/courage/src/util/jquery-wrapper.js#L6

We could switch to use fetch here: https://github.com/okta/okta-signin-widget/pull/1138/files#diff-fd0c42bcb3e6f1c94a622775027b5836R147 
-Or- add logic in courage to exclude this header if the XSRF is empty.

aarongranick-okta:
@haishengwu-okta Now using fetch instead of jQuery for JSON files. This should solve issue with the XSRF header

PR #1148, opened 2020-04-22
feat: add support for okta verify forced upgrade

magizh-okta:
* During OV verififcation if the OV app is outdated the SIW receives a
response with factorResult: 'REJECTED' and factorResultMessage 'OKTA_VERIFY_UPGRADE_REQUIRED'
we display a error messaging prompting the user to upgrade the app.

RESOLVES: OKTA-292351

**Error message when rejected due to app upgrade requirement**

**Ignore the name in the header (test iphone)** 

Screen Shot 2020-04-23 at 2 48 47 PM

Screen Shot 2020-04-23 at 2 47 04 PM

**Explicit User reject in normal scenario**

Screen Shot 2020-04-22 at 9 38 43 AM

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-292351](https://oktainc.atlassian.net/browse/OKTA-292351)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -483,6 +483,7 @@ oktaverify.send = Send Push
 oktaverify.resend = Re-send Push
 oktaverify.sent = Push sent!
 oktaverify.rejected = You have chosen to reject this login.
+oktaverify.rejected.upgradeRequired = Verification failed because your Okta Verify version is no longer supported. To sign in, please update Okta Verify, then try again.
```
pradeepdewda-okta, line 486: would it be helpful to show app versions to user ?

src/views/mfa-verify/PushForm.js
```diff
@@ -50,6 +50,15 @@ function (Okta, CookieUtil, Util, NumberChallengeView) {
         }
       );
 
+      this.listenTo(this.options.appState, 'change:isMfaRejectedDueToOutdatedApp',
+        function (state, isMfaRejectedDueToOutdatedApp) {
```
pradeepdewda-okta, line 54: it would be clean to create a separate new handler for this.

magizh-okta: You mean not use a  inline function ?

packages/@okta/i18n/src/properties/login.properties
magizh-okta: @stevennguyen-okta @ManuMalhotra-okta  thoughts ?

src/views/mfa-verify/PushForm.js
pradeepdewda-okta: yeah.

packages/@okta/i18n/src/properties/login.properties
stevennguyen-okta: Interesting idea, I discussed this with Aakash and we think that it wouldn't add much because the user would just have to go update anyway, and knowing what version number wouldn't change anything since presumably the update would fix whatever the reason is for updating.

src/views/mfa-verify/PushForm.js
magizh-okta: @pradeepdewda-okta  not sure if we are adding much value here since all other functions are inline and we are not unit testing this function per say(SIW unite tests are more flow based). 

Also naming these functions are tricky.

packages/@okta/i18n/src/properties/login.properties
mauriciocastillosilva-okta: nit: @stevennguyen-okta should we explicitly mention they need to update the app? Not sure if users might think of something like "admin needs to update my Okta Verify version" as opposed to "I need to go to the app store and update my app". e.g.
```
Your Okta Verify app is no longer supported. To sign in, please update Okta Verify, then try again.
```

test/unit/helpers/xhr/MFA_CHALLENGE_push_rejected_app_upgrade_needed.js
```diff
@@ -0,0 +1,77 @@
+define({
+    "status": 200,
+    "responseType": "json",
+    "response": {
+      "stateToken": "testStateToken",
+      "expiresAt": "2015-10-22T02:23:28.893Z",
+      "status": "MFA_CHALLENGE",
+      "factorResult": "REJECTED",
+      "factorResultMessage": "OKTA_VERIFY_UPGRADE_REQUIRED",
```
mauriciocastillosilva-okta, line 9: if this is the only difference with `MFA_CHALLENGE_push_rejected` file, maybe instead of adding a new file we can clone the object in the test and add this? e.g.
```
const resRejectedPushAppUpgrade = JSON.parse(JSON.stringify(resRejectedPush));
resRejectedPushAppUpgrade.repsonse.factorResultMessage = 'OKTA_VERIFY_UPGRADE_REQUIRED';
```

src/models/AppState.js
```diff
@@ -260,7 +260,16 @@ function (Okta, Q, Factor, BrowserFeatures, Errors) {
         // user clicks 'deny' on his phone.
         deps: ['lastAuthResponse'],
         fn: function (res) {
-          return res.factorResult === 'REJECTED';
+          return res.factorResult === 'REJECTED' && !res.factorResultMessage;
+        }
+      },
+      'isMfaRejectedDueToOutdatedApp': {
```
mauriciocastillosilva-okta, line 266: here's another idea: keep `isMfaRejectedByUser` (maybe change it to `isMfaRejected`) and add a new derived property to know if the app is outdated
```
isOktaVerifyAppOutdated: {
  deps: ['lastAuthResponse'],
  fn: function (res) {
    return res.factorResultMessage === 'OKTA_VERIFY_UPGRADE_REQUIRED';
  }
}
```

That way in the Form we listen to changes for `change:isMfaRejectedByUser` and we just handle the error message different, as opposed to adding a new listener. The logic for both listeners is basically the same, that is why I suggest this

With that said, what we have is good as well, it is just an idea

packages/@okta/i18n/src/properties/login.properties
stevennguyen-okta: Good idea, we originally had a title on the infobox that kinda helped but decided not to hack the infobox so that message was lost. I've updated the mocks and the message should now mention "please update Okta Verify on {appStoreName}". So if you're on iOS it'd be "on the App Store" and Android would be "on Google Play"

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2975,6 +2977,56 @@ function (Okta,
                   });
               });
             });
+            itp('on REJECTED requirering app upgrade, re-enables submit, displays an error, and allows resending', function () {
```
gowthamidommety-okta, line 2980: nit: `requiring`

src/models/AppState.js
magizh-okta: I like this idea too but this introduces additional checks in the PushForm increasing the complexity a bit. Leaving it as it is.

src/views/mfa-verify/PushForm.js
```diff
@@ -50,6 +50,19 @@ function (Okta, CookieUtil, Util, NumberChallengeView) {
...
+          this.setSubmitState(isMfaRejectedDueToOutdatedApp);
+          if (isMfaRejectedDueToOutdatedApp) {
+            if (this.options.appState.get('factor').profile.platform === 'IOS') {
+              this.showError(Okta.loc('oktaverify.rejected.upgradeRequired.ios', 'login'));
```
pradeepdewda-okta, line 58: i wonder if we can just pass the argument to i18n property, rather than using two different property variables for the same sentence.

magizh-okta: I thought so initially but there is a subtle difference. 

'on the App Store' and 'on Google Play'

pradeepdewda-okta: you can pass `the App Store` and `Google Play` right ?

magizh-okta: 'the' in other languages might not be 'the' since it unlike 'App Store'

pradeepdewda-okta: Interesting. Didn't think that way.

src/models/AppState.js
mauriciocastillosilva-okta: ok, I think it will probably make it simpler, but if we go with what we have today, the other change that is happening is that `isMfaRejectedByUser` is now checking for an empty `factorResultMessage`. Do we know if that is true for all rejected flows?
cause the new `isMfaRejectedDueToOutdatedApp` checks for an specific message, so if there is a flow that comes with a different messages, we won't be handling that anymore. Let's just double check that

magizh-okta: Good point Mau. I checked and isMfaRejectedByUser is true when factorResult is 'REJECTED' and factorResultMessage is empty. My check here works. 

But like you said I guess better to not bank on that I will make the check explicit.

src/views/mfa-verify/PushForm.js
```diff
@@ -146,6 +139,26 @@ function (Okta, CookieUtil, Util, NumberChallengeView) {
...
+    handleRejectStateChange: function (state, isMfaRejected) {
+      if (isMfaRejected) {
+        this.setSubmitState(isMfaRejected);
+        this.setRejectedErrorMessage();
+      }
+    },
+    setRejectedErrorMessage: function () {
+      // If rejection is due to outdated app, show error message per platform
+      // else show user rejected message.
+      if (this.options.appState.get('lastAuthResponse').factorResultMessage
```
magizh-okta, line 152: Not making this a derived prop since this is not used in many places.

src/views/mfa-verify/PushForm.js
```diff
@@ -41,27 +41,7 @@ function (Okta, CookieUtil, Util, NumberChallengeView) {
...
-          if (isMfaRejectedDueToOutdatedApp) {
-            if (this.options.appState.get('factor').profile.platform === 'IOS') {
-              this.showError(Okta.loc('oktaverify.rejected.upgradeRequired.ios', 'login'));
-            } else {
-              this.showError(Okta.loc('oktaverify.rejected.upgradeRequired.android', 'login'));
-            }
-          }
-        }
-      );
+      this.listenTo(this.options.appState, 'change:isMfaRejected', this.handleRejectStateChange);
```
pradeepdewda-okta, line 44: this is much better now..

src/views/mfa-verify/PushForm.js
mauriciocastillosilva-okta: nit: I would still vote to add the derived property so if we need it in the future it is already there. Otherwise if someone else adds this usage might not get replaced...
This is a nit, so no prob at all if we keep it as is

magizh-okta:
@mauriciocastillosilva-okta  I see an issue with too many derived props. Will discuss this in UI office hours tomo. Merging this to go ahead with some testing.\

PR #1154, opened 2020-04-28
feat: make styles consistent with okta loginpage OKTA-294104

aarongranick-okta:
## Description:

Copies variable values from https://github.com/okta/okta-core/blob/master/clients/loginpage/sass/_variables.scss so that styling is consistent between Okta loginpage and custom widget page / customer hosted widget

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Original style:
Screen Shot 2020-04-28 at 1 07 16 PM

Updated style:
Screen Shot 2020-04-28 at 1 11 58 PM

### Issue:

- [OKTA-294104](https://oktainc.atlassian.net/browse/OKTA-294104)

### DIFF
--- a/assets/sass/_variables.scss
+++ b/assets/sass/_variables.scss
@@ -1,30 +1,30 @@
 // Page Properties
-$fonts: 'montserrat', Arial, Helvetica, sans-serif;
+$fonts: 'proxima nova', 'montserrat', Arial, Helvetica, sans-serif;
 $default-border-radius: 3px;
 $container-width: 400px;
 $container-min-width: 300px;
 $num-tabs: 5;
 
 // Font Size
-$font-size: 13px;
-$font-size-small: 12px;
-$font-size-header: 14px;
-$font-size-footer: 11px;
-$font-size-default: 14px;
+$font-size: 14px;
+$font-size-small: 13px;
+$font-size-header: 15px;
+$font-size-footer: 12px;
+$font-size-default: 15px;
 
 // Font Weight
 $font-weight-body: normal;
 $font-weight-labels: 600;
-$font-weight-header: bold;
+$font-weight-header: 600;
 
 // Theme Color
 $primary-bg-color: #fff;
 $secondary-bg-color: #f9f9f9;
 
 // Text Color
-$header-text-color: #2f3f4a; // gray 900
-$dark-text-color: #2f3f4a; // gray 900
-$medium-text-color: #5c6971; // gray 700
+$header-text-color: #5e5e5e;
+$dark-text-color: #5e5e5e;
+$medium-text-color: #777;
 $light-text-color: #a7a7a7;
 $link-text-color: #0074b3;
 $placeholder-text-color: #aaa;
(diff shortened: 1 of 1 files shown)


assets/sass/_variables.scss
```diff
@@ -1,30 +1,30 @@
...
 $font-weight-body: normal;
 $font-weight-labels: 600;
-$font-weight-header: bold;
+$font-weight-header: 600;
 
 // Theme Color
 $primary-bg-color: #fff;
 $secondary-bg-color: #f9f9f9;
 
 // Text Color
```
haishengwu-okta, line 24: - Looked at history of this file, `header-text-color`, `primary-button-bg-color` and `google-icon-bg-color` are recent changes (within a year). shall we keep them instead of override by loginpage variables?

assets/sass/_variables.scss
```diff
@@ -1,30 +1,30 @@
...
 $container-min-width: 300px;
 $num-tabs: 5;
 
 // Font Size
-$font-size: 13px;
-$font-size-small: 12px;
-$font-size-header: 14px;
-$font-size-footer: 11px;
-$font-size-default: 14px;
+$font-size: 14px;
```
haishengwu-okta, line 9: @edburyenegren-okta just want to know what your thoughts to make this change. Does the look-n-feel work for you or who we shall double check?

aarongranick-okta:
> shall we add those changes to 4.0 ?

I think that makes sense. Although the changes are very subtle, some users may be unpleasantly surprised to see changes in style outside of a major version bump.

assets/sass/_variables.scss
aarongranick-okta: That probably makes sense. I could be wrong but I don't think we show the google signin buttons on our standard login page. I think you need to enable a custom signin page in order to login with IDP. Maybe @edburyenegren-okta can confirm?

haishengwu-okta: yeah, 3rd-party IdP is for custom signin page as today.

aarongranick-okta: OK, I have reverted the google icon styles

haishengwu-okta: what about other 2 change set as I mentioned earlier.

aarongranick-okta: Thanks, didn't catch that earlier. I've reverted those two changes as well. This may cause very a slight change in appearance on the Okta loginpage when we remove those variables in okta-core. As long as we are using the correct/intentional value I think it is good that we make it consistent.

assets/sass/_variables.scss
edburyenegren-okta: I defer to @vincentvoong-okta and @alexdahl-okta here until SIW starts using Odyssey.

At that point, I assume SIW and Okta Core will (separately) treat that as their baseline - ensuring the same styles in both places without SIW relying on Okta Core for code.

aarongranick-okta:
@haishengwu-okta @mauriciocastillosilva-okta @magizh-okta Had a meeting with the design team, they were in agreement that the variables from okta-core seem more correct/up to date/have been verified/seen by everyone. I've adjusted all the variables to match okta-core, except the google buttons which do not appear on our standard login page.

PR #1165, opened 2020-05-05
chore: updates Smart Card IdP documentation in the SIW

gowthamidommety-okta:
Resolves: OKTA-295394

## Description:

Updates X509 documentation in the SIW

Screen Shot 2020-06-29 at 2 03 48 PM

Screen Shot 2020-06-29 at 2 03 26 PM

Note clicking on the EA Feature link takes us to https://developer.okta.com/docs/reference/releases-at-okta/#early-access-ea

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-295394](https://oktainc.atlassian.net/browse/OKTA-295394)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ You can learn more on the [Okta + JavaScript][lang-landing] page in our document
   - [Registration](#registration)
   - [IdP Discovery](#idp-discovery)
   - [OpenID Connect](#openid-connect)
+  - [Smart Card IdP](#smart-card-idp)
   - [Bootstrapping from a recovery token](#bootstrapping-from-a-recovery-token)
   - [Feature flags](#feature-flags)
 - [Events](#events)
@@ -1119,6 +1120,23 @@ Options for the [OpenID Connect](http://developer.okta.com/docs/api/resources/oi
     ```javascript
     authParams: {
       authScheme: 'OAUTH2'
+    }
+    ```
+
+### Smart Card IdP
+
+**:information_source: EA feature:** The Smart Card IdP feature is currently an [EA feature](https://developer.okta.com/docs/api/getting_started/releases-at-okta#early-access-ea). This feature is only available for okta-hosted widget implementations.
+
+Settings for authentication with a Smart Card `X509` type IdP.
+
+- `text` *(optional)* - Label for the Smart Card IdP button. By default, this value will be "Sign in with PIV / CAC card".
+
+- `className` *(optional)* - Class that can be added to the Smart Card IdP button.
+
+    ```javascript
+    piv: {
+      text: 'Authenticate with a Smart Card',
+      className: 'custom-style'
     }
     ```
 


README.md
```diff
@@ -1120,6 +1121,27 @@ Options for the [OpenID Connect](http://developer.okta.com/docs/api/resources/oi
...
+
+- `className` - Optional class that can be added to the Smart Card IdP button.
+
+- `isCustomDomain` - Optional boolean that is passed to the certAuthUrl endpoint to indicate whether the request comes from a custom domain.
+
+    ```javascript
+    piv: {
+      text: 'Authenticate with a Smart Card',
+      className: 'custom-style',
+      certAuthUrl: 'https://${yourOktaDomain}.okta.com/login/cert',
```
abdullahchougle-okta, line 1140: Shouldn't this be `https://${yourOktaDomain}.okta.com/api/internal/v1/authn/cert`? `https://${yourOktaDomain}.okta.com/login/cert` will lead to the JSP instead.

@nikolaystoilov-okta will it be a problem if we reference an internal API here?

README.md
```diff
@@ -1120,6 +1121,27 @@ Options for the [OpenID Connect](http://developer.okta.com/docs/api/resources/oi
     }
     ```
 
+### Smart Card IdP
+
+Options for authentication with a Smart Card `X509` type IdP.
```
mauriciocastillosilva-okta, line 1126: nit: "**Settings** for authentication with a Smart Card `X509` type IdP."

README.md
```diff
@@ -1120,6 +1121,27 @@ Options for the [OpenID Connect](http://developer.okta.com/docs/api/resources/oi
...
+
+- `certAuthUrl` - Required url endpoint that requests the browser for a certificate.
```
mauriciocastillosilva-okta, line 1128: we can probably omit "Required" and just add "Optional" in the ones we need. This is similar to other configs we have. 
nit: we should uppercase URL
```
`certAuthUrl` - URL endpoint that requests the browser for a certificate.
`text` *(optional)* - 
`className ` *(optional)* - 
etc
```
same comment for the other optional params

README.md
```diff
@@ -1120,6 +1121,27 @@ Options for the [OpenID Connect](http://developer.okta.com/docs/api/resources/oi
...
+
+- `text` - Optional string that can be set as the Smart Card IdP button text. By default, this value will be "Sign in with PIV / CAC card".
```
mauriciocastillosilva-okta, line 1130: what about
```
- `text` *(optional)* - Label for the Smart Card IdP button. By default, this value will be "Sign in with PIV / CAC card".
```

README.md
gowthamidommety-okta: Yea, this was just an example.. I wasn't sure if we could reference an internal api here. Will wait for Nikolay's input

mauriciocastillosilva-okta: +1 on leaving it as is. This is just an example and we shouldn't make any reference to internal APIs in our doc

abdullahchougle-okta: ok, but it's a confusing example since `/login/cert` won't ever work...

mauriciocastillosilva-okta: its a good point... but we don't have a public endpoint that will work, correct? We can make it more obvious that this is just an example with something like `/your/cert/validation/endpoint`?

abdullahchougle-okta: That might be an OK middle ground.

I assume that the customer will either need to contact Okta or see how the default SIW works to figure out the real URL.

haishengwu-okta:
looks like `certAuthUrl` is required? https://github.com/okta/okta-signin-widget/blob/master/src/models/Settings.js#L293

magizh-okta:
@gowthamidommety-okta  just double checking if it works with custom signon/url pages where okta still hosts. If not I dont see a point in documenting since it cannot be used by customers anyway.

gowthamidommety-okta:
> looks like `certAuthUrl` is required? https://github.com/okta/okta-signin-widget/blob/master/src/models/Settings.js#L293

@haishengwu-okta We added a comment here, noting that this feature would only be available to okta hosted SIW. In those cases, the certAuthUrl is passed by default from okta.

gowthamidommety-okta:
> @gowthamidommety-okta just double checking if it works with custom signon/url pages where okta still hosts. If not I dont see a point in documenting since it cannot be used by customers anyway.

@magizh-okta @abdullahchougle-okta For a SIW that okta hosts (okta default hosted / okta custom hosted) , the PIV configuration will work by default. This documentation is only so customers can know how to change the piv button text and add a custom piv button className if needed.

gowthamidommety-okta:
Cannot merge this in. Created a new PR https://github.com/okta/okta-signin-widget/pull/1280

PR #1181, opened 2020-05-13
chore: fix/update retirejs OKTA-228855

aarongranick-okta:
## Description:

Changes the path for `retirejs` from `src` to `target`. We are using webpack config to include specific versions of certain modules. Scanning the `target` folder will inspect the built webpack bundles for security vulnerabilities.

Also update the .requirejsignore.json to ignore the patched version of jQuery from courage we are using.

## PR Checklist

### Reviewers:
@nikhilvenkatraman-okta 
@haishengwu-okta 
@swiftone  

### Issue:

- [OKTA-228855](https://oktainc.atlassian.net/browse/OKTA-228855)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "clean": "rimraf target && rimraf dist && rimraf build2",
     "generate-phone-codes": "node buildtools generate-phone-codes",
     "generate-config": "node buildtools generate-language-config",
-    "retirejs": "retire --jspath src --package",
+    "retirejs": "retire --jspath target/js --package",
     "lint:eslint": "eslint .",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "clean": "rimraf target && rimraf dist && rimraf build2",
     "generate-phone-codes": "node buildtools generate-phone-codes",
     "generate-config": "node buildtools generate-language-config",
-    "retirejs": "retire --jspath src --package",
+    "retirejs": "retire --jspath target --package",
```
haishengwu-okta, line 31: could it be more specific to `target/js` ?

aarongranick-okta: Yes, that could save a few milliseconds

Issue #1130, opened 2020-04-08
Password Complexity rules are not displayed during the password expiration warning


epierce:
### I'm submitting a
  - [ ] bug report
  - [x] feature request

### Background info
When a user's password is expiring soon, they are given the chance to change their password, but the password complexity rules are not shown. The user is only given the password rules once their password has actually expired.

### Expected behavior
The user should be able to see the complexity rules whenever they are prompted to change their password.

swiftone:
Internal ref: OKTA-289869

tmackay-cenet:
I would like to see the rules also form part of the client side form validation - rather than submitting a known bad password and relying on a 403 error.

tmackay-cenet:
Bump. This is still causing considerable headaches for our users with the resulting confusion and account lockouts. Still I think input validation and red exclamation marks would be required to get the point across.

tmackay-cenet:
This seems to have made it into production with the 'showPasswordRequirementsAsHtmlList' option. Tested on v5.4.1 with password expiry warning (not yet expired).

PR #1185, opened 2020-05-18
feat: adds password rules for expiring passwords

swiftone:
Currently, the password complexity rule are only shown when a user's password has expired. No rules are displayed during a "password is expiring soon" warning. The feature flag alwayDisplayPasswordRules (defaults to "false") replaces the password warning subtitle with the complexity rules.

Resolves: OKTA-289869

## PR Checklist

- [] Have you verified the basic functionality for this change?

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-289869](https://oktainc.atlassian.net/browse/OKTA-289869)

### DIFF
--- a/src/PasswordExpiredController.js
+++ b/src/PasswordExpiredController.js
@@ -67,7 +67,8 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, FactorUtil, Uti
         }
       },
       subtitle: function () {
-        if (this.options.appState.get('isPwdExpiringSoon')) {
+        if (this.options.appState.get('isPwdExpiringSoon') &&
+           !this.settings.get('features.alwaysDisplayPasswordRules')) {
           return this.settings.get('brandName') ?
             Okta.loc('password.expiring.subtitle.specific', 'login', [this.settings.get('brandName')]) :
             Okta.loc('password.expiring.subtitle.generic', 'login');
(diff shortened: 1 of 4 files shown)


swiftone:
See issue #1130 for problem description

anipendakur-okta:
@swiftone , would it be possible for you to attach a screenshot of the change pls?

src/PasswordExpiredController.js
```diff
@@ -67,7 +67,8 @@ function (Okta, FormController, Enums, FormType, ValidationUtil, FactorUtil, Uti
         }
       },
       subtitle: function () {
-        if (this.options.appState.get('isPwdExpiringSoon')) {
```
haishengwu-okta, line 70: instead of adding `feature`, shall we change the logic to be
```
IF policy exists THEN
  display password rule
ELSE IF isPwdExpiringSoon THEN
  display generic subtitle
ELSE
  do nothing. (or maybe display generic password expired subtitle)
```

epierce: That was my first thought, but I figured that some deployers must want the current behavior since this hadn't been requested up until now, so I left that as the default and added a feature flag.

nikhilvenkatraman-okta: Isn't the request that we want to show the password complexity rules even when `isPwdExpiringSoon` is true ?

Might be good to add comments to go with the different FF's in code.

PR #1190, opened 2020-05-21
feat: add support for Apple IdP

haishengwu-okta:
## Description:

- support Apple IdP as social idp.
- integration would be done later when I got a apple account. Being able to redirect to okta then apple gives confident that widget is doing the right thing.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

- https://okta.box.com/s/bqdhw9p6htafxr7v41mmocuzf7ezyisi

### Issue:

- [OKTA-294133](https://oktainc.atlassian.net/browse/OKTA-294133)


assets/sass/_variables.scss
```diff
@@ -53,6 +53,9 @@ $facebook-border-color: #375685;
 $google-icon-bg-color: #fff;
 $google-label-bg-color: $button-bg-color;
 $google-border-color: darken($button-bg-color, 22%);
+$apple-icon-bg-color: #fff;
+$apple-label-bg-color: $button-bg-color;
+$apple-border-color: darken($button-bg-color, 22%);
```
jmelberg-okta, line 58: We should follow Apple's [design requirements](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple/overview/buttons/) for web **Sign in with Apple** buttons.

Looking at the video - we might want to consider making the button background black to avoid adding a one-off border.

haishengwu-okta: we're certainly not strictly following apple guide line.. :(
looks like our current design align with `White with Outline` section.
@ZoeMonosson-okta ^

jmelberg-okta: Right - and I think it is ok that we're not matching 100%.

From what I've found, users who are accustomed to the Apple ecosystem tend to interact with familiar branding. As a frequent Sign-in/up With Apple user myself, I'd be concerned with a button that doesn't quite fit what I'm used to. Likewise, seeing a Sign-in with Facebook button that isn't blue would likely detour users.

Again, just my two cents. 😄

haishengwu-okta: Certainly a gap and we'd like to close as soon as we could.

PR #1193, opened 2020-05-27
feat: remove requirements from error message if description is shown

santhoshbalakrishnan:
## Description:

- This is a follow up to showing the password requirements as a HTML list.
- This is a specific case where don't want to repeat the requirements again in the error message, since this is already shown in the description.
- Limiting this to places where the feature is on and if language is Japanese, since the requirements message was misleading for some combinations of the requirements. The description as bullet-points itself should give an indication of the requirements.
- We cannot check for error code this in this case, as the error code is shared between requirements not met message, common password message, etc. So error summary is the only differentiating factor.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

FF on:

Screen Shot 2020-05-27 at 4 05 50 PM

FF off:

Screen Shot 2020-05-27 at 4 06 13 PM

### Issue:

- [OKTA-300832](https://oktainc.atlassian.net/browse/OKTA-300832)

### DIFF
--- a/src/util/FactorUtil.js
+++ b/src/util/FactorUtil.js
@@ -346,6 +346,13 @@ function (Okta, TimeUtil) {
     return localizedQuestion.indexOf('L10N_ERROR') < 0 ? localizedQuestion : questionObj.questionText;
   };
 
+  fn.removeRequirementsFromError = function (responseJSON, policy) {
+    var passwordRequirementsAsString = this.getPasswordComplexityDescription(policy);
+    responseJSON.errorCauses[0].errorSummary = responseJSON.errorCauses[0].errorSummary
+        .replace(passwordRequirementsAsString, '').trim();
+    return responseJSON;
+  };
+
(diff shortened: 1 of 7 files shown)


src/util/FactorUtil.js
```diff
@@ -346,6 +346,13 @@ function (Okta, TimeUtil) {
     return localizedQuestion.indexOf('L10N_ERROR') < 0 ? localizedQuestion : questionObj.questionText;
   };
 
+  fn.removeRequirementsFromError = function (responseJSON, policy) {
+    var passwordRequirementsAsString = this.getPasswordComplexityDescription(policy);
+    responseJSON.errorCauses[0].errorSummary = responseJSON.errorCauses[0].errorSummary
```
haishengwu-okta, line 351: better to safe check `errorCauses` is Array and `errorCause[0].errorSummary` is string.

PR #1198, opened 2020-05-30
test: adding support for ie + edge in saucelabs

vijetmahabaleshwar-okta:
resolves okta-298610

## Description:

* Use saucelabs to add tests for IE and Edge

* There have been bugs introduced in trex due to widget not loading on IE/Edge

* Adding these tests will catch them before they’re deployed to staging

* Only runs the basic_spec.js test for now. Some tests fail on IE. This will at least give us confidence that widget loads on IE/Edge

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-302703](https://oktainc.atlassian.net/browse/OKTA-302703)

### DIFF
--- a/test/e2e/conf.js
+++ b/test/e2e/conf.js
@@ -16,7 +16,12 @@ require('./env').config();
 var config = {
   framework: 'jasmine2',
   specs: ['specs/*.js'],
-  restartBrowserBetweenTests: false
+  restartBrowserBetweenTests: false,
+  onPrepare: function () {
+    return browser.getProcessedConfig().then(data => {
+      global.browserName = data.capabilities.browserName;
+    });
+  }
 };
 
 // Travis sauceLabs tests
@@ -25,8 +30,6 @@ if (process.env.TRAVIS) {
     // Mobile emulators on sauce labs
     config.sauceUser = process.env.SAUCE_USERNAME;
     config.sauceKey = process.env.SAUCE_ACCESS_KEY;
-    // Default port for Appium
-    config.port = 4723;
   } else {
     // Desktop browser
     config.capabilities = {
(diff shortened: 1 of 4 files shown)


haishengwu-okta:
@vijetmahabaleshwar-okta curious where can I find the test result? I see nothing from saucelab that I can login from my enduser dashboard.

vijetmahabaleshwar-okta:
@haishengwu-okta - The link to saucelabs run will be available in the travis log at the end.
Unfortunately some of the tests are still failing (unrelated to the IE bug), which I have to investigate.
Here's one such run - http://saucelabs.com/jobs/31937417a14342f5a3bb22be5921c61a
Also, I think we should discuss during the widget review meeting on whether we want to run IE tests for each PR.
We also have optional crossbrowser test suites on bacon. We should run them for every widget version change in the monolith. We should have a test runbook for widget upgrade I believe.

haishengwu-okta:
can we rebase to `4.1` ?

test/e2e/specs/basic_spec.js
```diff
@@ -72,6 +72,12 @@ describe('Basic flows', function () {
   });
 
   it('has the style from config.colors', function () {
+    /*global browserName*/
+    // In IE, primaryButton.getCssValue('background') is empty. Expectation on line 96 fails
```
haishengwu-okta, line 76: do u know why ?

test/e2e/conf.js
```diff
@@ -38,14 +41,36 @@ if (process.env.TRAVIS) {
...
+      'build': process.env.TRAVIS_BUILD_NUMBER
+    },
+    {
+      'browserName': 'MicrosoftEdge',
+      'browserVersion': 'latest',
+      'platformName': 'Windows 10',
+      'tunnel-identifier': process.env.TRAVIS_JOB_NUMBER,
+      'build': process.env.TRAVIS_BUILD_NUMBER
+    }];
+    config.exclude = ['specs/angular_spec.js', 'specs/dev_spec.js', 'specs/npm_spec.js',
```
haishengwu-okta, line 71: why exclude those tests?

test/e2e/specs/basic_spec.js
vijetmahabaleshwar-okta: I haven't been able to figure out why. But it consistently fails. 
I can create a JIRA for it and add it in the comment here.

test/e2e/conf.js
vijetmahabaleshwar-okta: Some of the tests simply don't pass on IE 11. For example, getting the browser logs throws an error in IE, probably because it's not supported? I have to still dig deep.
```
dev_spec.js - expectToFindLogMessage
1) Dev Mode flows log a console message when tokens are not parsed from the URL after the Widget is rendered
Error: Failed expectation
        at /Users/vijetm/okta/okta-signin-widget/target/e2e/specs/dev_spec.js:116:21
```
I thought it's better to have some coverage first (that gives us confidence that the widget loads on IE) and then figure out why all these tests are failing on IE.
What do you suggest?

test/e2e/specs/basic_spec.js
vijetmahabaleshwar-okta: Update: I see this thread which explains why this doesn't work on IE - https://stackoverflow.com/questions/11503709/style-attribute-issue-with-internet-explorer

test/e2e/conf.js
haishengwu-okta: sounds good to me. better to include all that are able to pass today?

vijetmahabaleshwar-okta: I can do that, but in most of the specs, out of 5 of them, 1 or 2 pass. 
That'll add a lot of `if browser == ie` statements. I don't think that's clean. This one spec has the most tests passing. So, I'll add this for now. Will create a JIRA to track other failures.

PR #1202, opened 2020-06-01
chore: refactoring how to handle the authenticator api response

haishengwu-okta:
## PR Checklist

- [ ] Have you verified the basic functionality for this change?

- [ ] Added unit tests?

- [ ] Added e2e tests

- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?

### Issue:

- [OKTA-300796](https://oktainc.atlassian.net/browse/OKTA-300796)

PR #1203, opened 2020-06-02
feat: adding support for remember me checkbox

anipendakur-okta:
## Description:

support for remember me checkbox

OKTA-297792

## PR Checklist

- [X] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
image

### Reviewers:
@haishengwu-okta @nikhilvenkatraman-okta 

### Issue:

- [OKTA-297792](https://oktainc.atlassian.net/browse/OKTA-297792)


src/v2/ion/uiSchemaTransformer.js
```diff
@@ -134,6 +134,19 @@ const createUISchema = (remediationValue = [], factors = [], authenticators = []
...
       type: 'text',
     };
+
+    if (ionFormField.name === 'rememberMe') { 
+      // For Remember Me checkbox, we need the label only on the right side of it.
+      uiSchema.placeholder = ionFormField.label;
+      uiSchema.label = false;
+      // Separating prop type for Backbone.Model
+      // from html input type
+      uiSchema.modelType = ionFormField.type;
```
anipendakur-okta, line 144: Note for @haishengwu-okta : I know we discussed this to be a generic field for uiSchema level, however it seems like weird values were getting passed `ionField.type` can be `set` or `object` etc (discovered when i was writing tests), hence kept it to only for rememberMe checkbox.

haishengwu-okta: good catch!
which field has type `set` ?

haishengwu-okta: now Im thinking we can have generic function call `convertIonTypeToCourageModelType`? thoughts?

anipendakur-okta: Yep, it makes sense to have it for every field. Also, makes sense to always keep HTML type separated from Courage Model type.

anipendakur-okta: Here are the two objects that will have `modelType` preset to `set`:
https://github.com/okta/okta-signin-widget/blob/52209eaa88564525dda86ca3bc9eb6b3f031e2c9/test/unit/spec/v2/ion/uiSchemaTransformer_spec.js#L356

https://github.com/okta/okta-signin-widget/blob/52209eaa88564525dda86ca3bc9eb6b3f031e2c9/test/unit/spec/v2/ion/uiSchemaTransformer_spec.js#L576

anipendakur-okta: But having weird values like `set` makes it hard to have a generic function like that. what do you think? I have also noticed `object` for type. It will be complex to handle such scenarios.

haishengwu-okta: `object` make sense. What you have is good for now.

test/unit/spec/v2/ion/uiSchemaTransformer_spec.js
```diff
@@ -10,6 +10,178 @@ import XHRAuthenticatorEnrollOptions  from '../../../../../playground/mocks/idp/
 
 
 describe('v2/ion/uiSchemaTransformer', function () {
+  it('converts identify remidiation response', () => {
+    const rawIdentifyResponse = identifyResponse;
+    const transformedResponse  = {
```
haishengwu-okta, line 15: check my PR https://github.com/okta/okta-signin-widget/pull/1202, I did refactoring the way at constructing `transformedResponse` and assertion.

playground/mocks/idp/idx/data/identify.json
```diff
@@ -20,6 +20,11 @@
                   "name":"identifier",
                   "label":"Username"
               },
+              {
+               "name":"rememberMe",
```
haishengwu-okta, line 24: just double check, Is `rememberMe` field always in the API?

anipendakur-okta: I believe so ( https://github.com/okta/okta-core/pull/42373 ), @jadsiblini-okta can confirm it certainly.

src/v2/ion/uiSchemaTransformer.js
nikhilvenkatraman-okta: can we not check for `ionFormField.type==='boolean'` instead of `ionFormField.name === 'rememberMe'`?

anipendakur-okta: We want to target `rememberMe` and hence not checking against `type`. A few reasons - 
* A `boolean` may not always translate to checkbox (dropdown)
* In the case of rememberMe, we don't want label but we want placeholder to have the value from label. (to have the text display only on the right side of the checkbox).

pradeepdewda-okta:
@anipendakur-okta i see a very slight vertical alignment issue between checkbox and remember me label.

playground/mocks/idp/idx/data/identify.json
pradeepdewda-okta: @jadsiblini-okta you mean defaults to `"visible" : "false"` ?

anipendakur-okta: I believe @jadsiblini-okta is talking about the POST response from UI from identify step. If `rememberMe` is not sent, it will be considered `false` gracefully.

jadsiblini-okta: Right

src/v2/ion/uiSchemaTransformer.js
nikhilvenkatraman-okta: > A boolean may not always translate to checkbox (could be dropdown in UI - depends on the implementation)
I don't see how a dropdown component would fit a boolean datatype. A dropdown will have many items in it which more suite an array datatype.

> In the case of rememberMe, we don't want label but we want placeholder to have the value from label. (to have the text display only on the right side of the checkbox).

I would assume that we want all our checkboxes to look the same. Curious why do we think we don't want this same behavior of getting value from label for any other checkbox.

anipendakur-okta:
Fixed CSS - new screenshot - 
image

src/v2/ion/uiSchemaTransformer.js
haishengwu-okta: I was suggesting to check on `name` cause it feels to me that we might want deal with `boolean` type differently per use case.
Given your comments, maybe we could just check on type `boolean` for now is a good idea.
sorry @anipendakur-okta for back and force.

src/v2/ion/uiSchemaTransformer.js
```diff
@@ -134,6 +134,19 @@ const createUISchema = (remediationValue = [], factors = [], authenticators = []
...
+      // uiSchema type is the html input type desired.
+      uiSchema.type = 'checkbox';
+      uiSchema.required = false;
```
haishengwu-okta, line 147: when `required` is missing, default to `false`. would be better to ask to change API by adding `required: false`

assets/sass/common/shared/o-forms/_o-form.scss
```diff
@@ -441,7 +441,6 @@
 %custom-checkbox-label {
   display: block;
   position: relative;
-  line-height: 16px;
```
haishengwu-okta, line 444: Looks like even okta-ui/courage has this setting. Can we double check with design team before making changes?

src/v2/ion/uiSchemaTransformer.js
anipendakur-okta: it doesn't appear to be false by default as error gets thrown.

test/unit/spec/v2/ion/uiSchemaTransformer_spec.js
```diff
@@ -1755,4 +1755,140 @@ describe('v2/ion/uiSchemaTransformer', function () {
...
+        }
+      ],
+      'neededToProceed': identifyResponse.remediation.value,
+      'actions': {},
+      'context': identifyResponse.context,
+      'rawIdxState': rawIdentifyResponse
+    };
+
+    // Overriding to a smaller string locally for easy testing.
+    transformedResponse.context.value[0].value = 'asdf97987asdfasdf';
```
haishengwu-okta, line 1794: dont think it's good idea to make special case for testing..

test/unit/spec/v2/ion/uiSchemaTransformer_spec.js
```diff
@@ -1755,4 +1755,140 @@ describe('v2/ion/uiSchemaTransformer', function () {
     });
   });
 
+  it('converts identify remidiation response', () => {
+    const rawIdentifyResponse = identifyResponse;
+    const transformedResponse  = {
+      'remediations': [
```
haishengwu-okta, line 1761: `transformedResponse` doesn't have `remediations`

test/unit/spec/v2/ion/uiSchemaTransformer_spec.js
```diff
@@ -1755,4 +1755,140 @@ describe('v2/ion/uiSchemaTransformer', function () {
...
+          'method':'post',
+          'value':[
+    
+          ],
+          'uiSchema':[
+    
+          ]
+        }
+      ],
+      'neededToProceed':[
```
haishengwu-okta, line 1846: identifyResponse.remediation.value

test/unit/spec/v2/ion/uiSchemaTransformer_spec.js
```diff
@@ -1755,4 +1755,140 @@ describe('v2/ion/uiSchemaTransformer', function () {
...
+        {
+          'name': 'select-enroll-profile',
+          'href': 'http://localhost:3000/idp/idx/enroll',
+          'method': 'post',
+          'value': []
+        }
+      ],
+      'neededToProceed': identifyResponse.remediation.value,
+      'actions': {},
+      'context': identifyResponse.context,
```
haishengwu-okta, line 1789: `context: _.omit(identifyResponse, 'remediation', `cancel`)`

test/unit/spec/v2/ion/uiSchemaTransformer_spec.js
```diff
@@ -1755,4 +1755,140 @@ describe('v2/ion/uiSchemaTransformer', function () {
...
+          'rel':[ 'create-form' ],
+          'name':'select-enroll-profile',
+          'href':'http://localhost:3000/idp/idx/enroll',
+          'method':'POST',
+          'accepts':'application/vnd.okta.v1+json',
+          'value':[],
+        }
+      ],
+      'actions':{},
+      'context':{
```
haishengwu-okta, line 1875: `context: _.omit(identifyResponse, 'remediation', `cancel`)`

src/v2/ion/uiSchemaTransformer.js
anipendakur-okta: makes sense to do `uiSchema.required = ionFormField.required || false;` (it is not `false` by default right now)

test/unit/spec/v2/ion/uiSchemaTransformer_spec.js
anipendakur-okta: `identifyResponse` has a lot more fields, hence plucked the `context` out manually.

src/v2/ion/uiSchemaTransformer.js
nikhilvenkatraman-okta: yeah, I think we can check for boolean now given that this is the only use case for it. We can always update our logic later.

I also feel like since we have asked the API to include `type` specifically for the remember me checkbox, we should make use of it to see how it fits our needs on the front-end. If it does not meet our requirements, then we should suggest changes.

assets/sass/common/shared/o-forms/_o-form.scss
anipendakur-okta: The effective change is `line-height: 18.2px` after removing this. It seems like this code was added years ago and it probably went unnoticed. I have removed it since it fixes the misalignment issue, will change it if @pradeepdewda-okta thinks otherwise..

Since the checkbox gets created from backend directly in my case, i'll have to add the class somewhere in the uiSchema before it gets created. It makes sense to me to have a generic fix for all checkboxes in SIW in this case.

pradeepdewda-okta: @taylorlaubach-okta what are your thoughts ?

anipendakur-okta: Changed the styling to target only v2 - remember me checkbox.

haishengwu-okta: we wouldn't like one-off in general. If what Pradeep found is an issue and ux team agreed, we could fix the infobox common style as well as in okta-ui (could be separated ticket)

taylorlaubach-okta: @pradeepdewda-okta Thanks for looping me in! I agree with you--the old alignment looks bad and the fix looks correct.

I'm okay with fixing the issue anywhere it exists. Just want to make sure that we don't break correct alignment anywhere. (For example, if some checkboxes need the `line-height: 16px` style in order to be correctly aligned and removing it breaks that for them.)

pradeepdewda-okta: Thanks @taylorlaubach-okta.!! @anipendakur-okta it's your call to fix it in this PR or create a new one and provide the fix there. Any case it should not hold you to merge this PR.

anipendakur-okta: It makes sense to address it as a separate JIRA since this involves testing all radios and checkboxes on sign-in-widget. I opened a new JIRA to address it - https://oktainc.atlassian.net/browse/OKTA-303139

assets/sass/v2/_identify.scss
```diff
@@ -40,4 +40,4 @@
       color: $border-color-default;
     }
   }
-}
+}
```
pradeepdewda-okta, line 43: nit:

anipendakur-okta: updated :)

src/v2/ion/uiSchemaTransformer.js
```diff
@@ -173,45 +173,79 @@ const createUISchema = (transformedResp, remediationForm) => {
     })
     .flatten()
     .value();
+
+  const getCheckboxUiSchema = (ionFormField, uiSchema) => {
```
anipendakur-okta, line 177: @haishengwu-okta , looks like there are new lint rules (only 15 statements allowed in a function or an arrow function), hence extracted them to their own granular functions.

haishengwu-okta: that's because `  /* eslint complexity: [2, 14] */` at line 157.
my intention was let it grow and find chance to refactor later.
I'm fine with 
- either your changes but your `get***` function shall return pure object instead change `options` in place. otherwise it's unconventional.
- or update the eslint override

anipendakur-okta: Updated with some ES6 sugar. Nevertheless, needed to update the complexity as well.

src/v2/ion/uiSchemaTransformer.js
```diff
@@ -154,7 +154,7 @@ const createAuthenticatorOptions = (options = [], authenticators = []) => {
  * @param {IONForm} remeditationForm
  */
 const createUISchema = (transformedResp, remediationForm) => {
-  /* eslint complexity: [2, 14] */
+  /* eslint complexity: [2, 15] */
```
haishengwu-okta, line 157: Thanks for refactor!

assets/sass/common/shared/o-forms/_o-form.scss
taylorlaubach-okta: @anipendakur-okta Thanks for doing that! Let me know if you need anything from me.

PR #1206, opened 2020-06-02
chore: wip

haishengwu-okta:
okta-123

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:


playground/config/responseConfig.js
```diff
@@ -4,11 +4,7 @@ const idx = {
   // ===== IDX
 
   '/idp/idx/introspect': [
-    // 'authenticator-select-verify-options',
```
pradeepdewda-okta, line 7: don't remove these, its helpful to quickly switch views when using localhost.

src/v2/models/AppState.js
```diff
@@ -101,55 +103,50 @@ export default Model.extend({
...
     // expiresAt will be different for each response, hence compare objects without that property
-    if (_.isEqual(_.omit(idxResp.rawIdxState, 'expiresAt'), _.omit(this.get('__previousResponse'), 'expiresAt'))) {
+    const previousRawState = this.has('idx') ? this.get('idx').rawIdxState : null;
+    if (_.isEqual(_.omit(transformedResponse.idx.rawIdxState, 'expiresAt'), _.omit(previousRawState, 'expiresAt') )) {
       return;
     }
 
     // `currentFormName` is default to first form of remediation object or nothing.
-    idxResp.currentFormName = null;
+    transformedResponse.currentFormName = null;
```
pradeepdewda-okta, line 116: this would going to trigger `render` and later on one of the if condition can also again trigger `render` because now we listen for `change:currentFormName` event ? can we suppress this double event trigger ? or is it required ?

haishengwu-okta: why `transformedResponse.currentFormName = null` will trigger `render` event? 
It doesn't set to Model yet.

haishengwu-okta:
@pradeepdewda-okta updated description by add wiki page. No UI changes are expected but more like additive and refactoring.

src/v2/models/AppState.js
pradeepdewda-okta: ah okay make sense.

src/v2/controllers/FormController.js
```diff
@@ -123,8 +119,7 @@ export default Controller.extend({
   },
 
   updateAppStateWithNewIdx: function (idxResp) {
```
pradeepdewda-okta, line 121: do we even need this func now ? i mean what's the usefulness.

src/v2/controllers/FormController.js
```diff
@@ -51,9 +50,8 @@ export default Controller.extend({
   invokeAction (actionPath = '') {
     const idx = this.options.appState.get('idx');
     if (idx['neededToProceed'].find(item => item.name === actionPath)) {
```
pradeepdewda-okta, line 52: I think this `if` and next `if` at line 64 can be part of one if..elseif..else clause.. What do you think?

src/v2/models/AppState.js
```diff
@@ -24,27 +27,26 @@ export default Model.extend({
...
   },
 
   derived: {
     authenticatorProfile: {
       // While we're moving toward `authenticator` platform, but still
       // need to support `factor` for certain period.
       // Could remove `factor` after it's deprecated completely.
-      deps: ['currentAuthenticator', 'authenticatorEnrollment', 'factor'],
-      fn (currentAuthenticator = {}, authenticatorEnrollment = {}, factor = {}) {
+      deps: ['currentAuthenticator', 'enrollmentAuthenticator', 'factor'],
```
pradeepdewda-okta, line 37: instead of `enrollmentAuthenticator` use `enrollAuthenticator`.. `enrollment` is a verb..

src/v2/models/AppState.js
```diff
@@ -101,55 +103,50 @@ export default Model.extend({
...
     this.clear({silent: true});
     // set new app state properties
-    this.set(idxResp);
-    // reset idx
-    this.set('idx', idxResp);
-    this.set('__previousResponse', idxResp.rawIdxState);
-
-    // broadcast idxResponseUpdated to re-render the view
-    this.trigger('idxResponseUpdated', idxResp);
+    this.set(transformedResponse);
```
pradeepdewda-okta, line 149: this is pretty neat. just one `set` reduced from three..

src/v2/models/AppState.js
haishengwu-okta: `enrollmentAuthenticator` is from api contractor. might be able to clean up when we see the final JSON ( when integrating with api )

src/v2/ion/uiSchemaTransformer.js
```diff
@@ -154,19 +154,26 @@ const createAuthenticatorOptions = (options = [], authenticators = []) => {
...
     .map(v => {
+      // `v.form` is probably not right structure but `v.value.form`
+      // TODO: clean up after API stablization.
       if (v.form) {
         const inputGroupName = v.name;
         return v.form.value.map(input => {
           return Object.assign({}, input, { name: inputGroupName + '.' + input.name });
         });
+      } else if (v.value && v.value.form) {
+        const inputGroupName = v.name;
```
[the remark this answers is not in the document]
haishengwu-okta: Could be. but as the TODO above, we may able to remove the first `if` eventually. It's clear to just keep separated.

src/v2/controllers/FormController.js
haishengwu-okta: Could be. I guess it's about preference on code structure. ;)

However, I think it's a bit hacky (possible confusing) to invoke remediation Form in this function . This is probably quick work around today.
Initially, we would expect `invokeAction` only used to invoke form outside remediation and use `switchForm` to render other remediation forms.
Wonder we shall enhance `switchForm` function to see
- if the form has no ui schema, then just submit it.
- otherwise, render that form

thoughts ?
also @nikhilvenkatraman-okta

src/v2/controllers/FormController.js
haishengwu-okta: yes and we probably shall reuse this function instead of `trigger remediationSuccess` directly.

pradeepdewda-okta: I feel the func name is not relevant and trigger probably won't need to be encapsulated in a function until and unless we are doing more inside that function.

haishengwu-okta: Yeah, I'll rename the function name.

src/v2/controllers/FormController.js
```diff
@@ -40,7 +40,6 @@ export default Controller.extend({
     this.formView = this.add(TheView, {
       options: {
         currentViewState,
-        messages: this.options.appState.get('messages'),
```
pradeepdewda-okta, line 44: removed ?

haishengwu-okta: yes, https://github.com/okta/okta-signin-widget/pull/1112/

src/v2/ion/uiSchemaTransformer.js
```diff
@@ -154,19 +154,26 @@ const createAuthenticatorOptions = (options = [], authenticators = []) => {
...
   const remediationValue = _.chain(remediationForm.value || [])
     .map(v => {
+      // `v.form` is probably not right structure but `v.value.form`
+      // TODO: clean up after API stablization.
       if (v.form) {
         const inputGroupName = v.name;
         return v.form.value.map(input => {
           return Object.assign({}, input, { name: inputGroupName + '.' + input.name });
         });
+      } else if (v.value && v.value.form) {
```
anipendakur-okta, line 172: Instead of a check like this, can we not make sure of this in response transformer?

Also, once we start supporting ES2020, this would be a good candidate for optional chaining.

src/v2/controllers/FormController.js
```diff
@@ -51,10 +49,8 @@ export default Controller.extend({
   invokeAction (actionPath = '') {
     const idx = this.options.appState.get('idx');
     if (idx['neededToProceed'].find(item => item.name === actionPath)) {
-      idx.proceed(actionPath, {}).then((resp) => {
-        this.options.appState.set('idx', resp);
-        this.options.appState.trigger('remediationSuccess', resp.rawIdxState);
-      })
+      idx.proceed(actionPath, {})
```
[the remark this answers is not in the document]
haishengwu-okta: I'm all for async-await! (probably do it later given this PR already has lots of changes) Thanks for catching it!

src/v2/ion/uiSchemaTransformer.js
haishengwu-okta: > can we not make sure of this in response transformer?

src/v2/controllers/FormController.js
nikhilvenkatraman-okta: yeah, I think ideally this function would only deal with invoking the action functions that are inside `idx['actions']`, but we have some cases where the action is actually inside `idx['neededToProceed']`

IIRC see we could have 3 scenarios

- submit a form by calling handleFormSave with some model params to be posted 
- invokeAction where the action is within idx['actions'] and we can just call it 
- submit a form without any specific model params (which today is happening by calling invokeAction with form name)

We could move the last scenario into a separate method to keep them separate?  That would mean we need to update the places where we are calling invokeAction with a formName to instead call this new method.

src/v2/ion/uiSchemaTransformer.js
anipendakur-okta: I was referencing to       
// `v.form` is probably not right structure but `v.value.form`
Since it looks like a either-or scenario, can we make this decision in responseTransformer and pick either one ?

src/v2/controllers/FormController.js
haishengwu-okta: for #3, I'm thinking we could refactor the `switchForm` function and replace `invokeAction` with `switchForm` event.

src/v2/ion/uiSchemaTransformer.js
haishengwu-okta: Possibly. Will leave it for later when we got more real data from API integration testing. My hope is to just remove another `if` block.

anipendakur-okta:
Looks good to me. 📦

src/v2/BaseLoginRouter.js
```diff
@@ -86,7 +86,7 @@ export default Router.extend({
     // TODO: OKTA-244631 How to suface up the CORS error in IDX?
     // Since in new pipeline, it invokes introspect API first
     // hence no way to call GlobalError when CORS failure.
-    this.listenTo(this.appState, 'change:remediationFailure', function (appState, err) {
+    this.listenTo(this.appState, 'change:introspectError', function (appState, err) {
```
nikhilvenkatraman-okta, line 89: Is this because we are handling remediation failures in the form controller via `updateAppStateWithNewIdx ` ?

src/v2/ion/uiSchemaTransformer.js
```diff
@@ -178,12 +185,18 @@ const createUISchema = (transformedResp, remediationForm) => {
...
+
     if (ionFormField.secret === true) {
       uiSchema.type = 'password';
       uiSchema.params = {
         showPasswordToggle: true,
       };
     }
+
+    if (Array.isArray(ionFormField.options)) {
+      uiSchema.type = 'radio';
```
nikhilvenkatraman-okta, line 197: curious what is the `type` value that the backend sends in this case?

src/widget/OktaSignIn.js
```diff
@@ -37,13 +37,8 @@ var OktaSignIn = (function () {
       }
       if (widgetOptions.stateToken) {
         Util.introspectToken(authClient, widgetOptions)
-          .then(_.bind(function (data) {
-            var response = data;
+          .then(_.bind(function (response) {
             router = bootstrapRouter.call(this, Router, authClient, widgetOptions, renderOptions, successFn, errorFn);
-            if (isNewPipeline) {
-              response = data.rawIdxState;
-              router.appState.set('idx', data);
```
nikhilvenkatraman-okta, line 45: Are we no longer setting idx on the appstate ?

src/v2/ion/uiSchemaTransformer.js
haishengwu-okta: The API design doesn't specify `type`. I bet it's `string` given how we send back the `value`

src/widget/OktaSignIn.js
haishengwu-okta: We do. See the change to `responseTransformer`

src/v2/BaseLoginRouter.js
haishengwu-okta: Not really. This probably never worked 
- I didn't find where fire `remediationFailure` (even at oie initial commit) 🤦 
- as v2 router didn't take any `introspectError` b/c when `introspectError`, we fall back to v1. That's why `TODO OKTA-244631`.

haishengwu-okta:
@pradeepdewda-okta thanks for checking. run through bacon and get 🟢

src/v2/ion/uiSchemaTransformer.js
nikhilvenkatraman-okta: I see. Guess it would be ideal if  the API returned a type here though ? Since we are picking just one option with a radio a string might be ok.

PR #1207, opened 2020-06-02
fix: css fix for resend verification code callout

anipendakur-okta:
## Description:
fix: css fix for resend verification code callout
OKTA-297792

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Before:
image

After:
image

### Issue:

- [OKTA-297792](https://oktainc.atlassian.net/browse/OKTA-297792)


anipendakur-okta:
Note: I couldn't grab screenshots from IE because of https://oktainc.atlassian.net/browse/OKTA-302041 , I still see the same error when I use saucelabs tunnel.

anipendakur-okta:
On a second thought, it makes more sense to have left margin than right, changed the fix. (if the link falls down to the next page, it won't have a weird margin to the left now.)

German screenshot - 
image

assets/sass/modules/_email-factor.scss
```diff
@@ -1,12 +1,14 @@
 @mixin email-factor {
   .resend-email-infobox {
     margin-bottom: 15px;
+    .infobox-warning p span {
```
haishengwu-okta, line 4: shall we apply changes to all infobox in general (within oie scope)

anipendakur-okta: Not really, the mark-up (the span and the adjacent link) is bound to this particular resend-email-infobox, hence it should be targeted.

PR #1215, opened 2020-06-05
feat: enroll webauthn via idx

magizh-okta:
* enroll webauthn via IDX.
* Fix description in factor select view.

RESOLVES: OKTA-301556
** select factor  ** 

Screen Shot 2020-06-09 at 8 39 42 AM

**Browser doesnt support webauthn** 
Screen Shot 2020-06-05 at 8 35 11 AM

**User verification Required**
Screen Shot 2020-06-05 at 8 48 48 AM

**on edge - windows** 
Screen Shot 2020-06-05 at 9 28 17 AM

**User verification discouraged**
Screen Shot 2020-06-05 at 8 54 19 AM

**Webauthn prompt canceled by user**
Screen Shot 2020-06-05 at 8 55 09 AM

**Success enroll**
enroll

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-301556](https://oktainc.atlassian.net/browse/OKTA-301556)

### DIFF
--- a/playground/mocks/data/idp/idx/authenticator-select-enroll-options.json
+++ b/playground/mocks/data/idp/idx/authenticator-select-enroll-options.json
@@ -118,7 +118,7 @@
       {
         "label": "Security Key or Biometric Authenticator (FIDO2)",
         "value": {
-          "type": "webauthn",
+          "type": "security_key",
           "id": "aidtheidkwh282hv8g3"
         }
       },
--- a/src/v2/util/FactorUtil.js
+++ b/src/v2/util/FactorUtil.js
@@ -25,7 +25,7 @@ const factorData = {
   },
 
   'phone': {
-    label: loc('factor.phone', 'login'),
+    label: loc('oie.phone', 'login'),
     description: '',
     // TODO: need new icon for `mfa-okta-phone`
     iconClassName: 'mfa-okta-call',
--- a/src/v2/util/FactorUtil.js
+++ b/src/v2/util/FactorUtil.js
@@ -31,9 +31,16 @@ const factorData = {
     iconClassName: 'mfa-okta-call',
   },
 
+  // Will get rid of this after fully implementing verify. OKTA-301557
   'webauthn': {
-    label: loc('factor.webauthn', 'login'),
-    description: '',
+    label: loc('oie.webauthn', 'login'),
+    description: loc('oie.webauthn.description', 'login'),
+    iconClassName: 'mfa-webauthn',
+  },
+
+  'security_key': {
+    label: loc('oie.webauthn', 'login'),
+    description: loc('oie.webauthn.description', 'login'),
     iconClassName: 'mfa-webauthn',
   },
 };
--- a/src/v2/view-builder/internals/FormInputFactory.js
+++ b/src/v2/view-builder/internals/FormInputFactory.js
@@ -19,7 +19,7 @@ const createFactorSelectView = (opt) => {
 const createAuthenticatorSelectView = (opt) => {
   var optionItems = (opt.options || [])
     .map(opt => {
-      return Object.assign({}, FactorUtil.getFactorData(opt.authenticatorType), opt);
+      return Object.assign({}, opt, FactorUtil.getFactorData(opt.authenticatorType));
     });
   return {
     View: FactorOptions,
--- a/src/v2/view-builder/views/webauthn/EnrollWebauthnView.js
+++ b/src/v2/view-builder/views/webauthn/EnrollWebauthnView.js
@@ -1,0 +1,126 @@
+import { _, loc, createCallout, createButton } from 'okta';
+import BaseForm from '../../internals/BaseForm';
+import BaseFactorView from '../shared/BaseFactorView';
+import webauthn from '../../../../util/webauthn';
+import BrowserFeatures from 'util/BrowserFeatures';
+import CryptoUtil from '../../../../util/CryptoUtil';
+
+function getExcludeCredentials (credentials = []) {
+  return credentials.map((credential) => {
+    return {
+      type: 'public-key',
+      id: CryptoUtil.strToBin(credential.id),
+    };
+  });
+}
+
+const Body = BaseForm.extend({
+  title: loc('oie.enroll.webauthn.title', 'login'),
+  className: 'oie-webauthn',
+  modelEvents: {
+    'error': '_stopEnrollment',
+  },
+  getUISchema () {
+    const schema = [];
+    const activationData = this.options.appState.get('currentAuthenticator').contextualData.activationData;
+    // Returning custom array so no input fields are displayed for webauthn
+    if (webauthn.isNewApiAvailable()) {
+      schema.push({
+        View: `
+          <p class="idx-webauthn-enroll-text">{{i18n code="oie.enroll.webauthn.instructions" bundle="login"}}</p>`,
+      });
+
+      if (BrowserFeatures.isEdge()) {
+        schema.push({
+          View: `
+            <p class="idx-webauthn-enroll-text-edge">
+              {{i18n code="oie.enroll.webauthn.instructions.edge" bundle="login"}}
+            </p>`,
+        });
+      }
+
+      if (activationData.authenticatorSelection.userVerification === 'required') {
+        schema.push({
+          View: createCallout({
+            className: 'uv-required-callout',
+            size: 'slim',
+            type: 'warning',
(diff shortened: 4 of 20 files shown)


playground/mocks/data/idp/idx/authenticator-select-verify-options.json
```diff
@@ -63,6 +63,29 @@
                   }
                 }
               },
+              {
```
magizh-okta, line 66: my bad will remove.

src/v2/view-builder/internals/FormInputFactory.js
```diff
@@ -19,7 +19,7 @@ const createFactorSelectView = (opt) => {
 const createAuthenticatorSelectView = (opt) => {
   var optionItems = (opt.options || [])
     .map(opt => {
-      return Object.assign({}, FactorUtil.getFactorData(opt.authenticatorType), opt);
+      return Object.assign({}, opt, FactorUtil.getFactorData(opt.authenticatorType));
```
magizh-okta, line 22: changing the order to override label in options.

test/testcafe/framework/page-objects/EnrollWebauthnAuthenticatorPageObject.js
```diff
@@ -0,0 +1,15 @@
+import BasePageObject from './BasePageObject';
+
+export default class EnrollPasswordPageObject extends BasePageObject {
```
haishengwu-okta, line 3: `class EnrollWebauthnAuthenticatorPageObject`

test/testcafe/spec/EnrollWebauthnAuthenticator_spec.js
```diff
@@ -0,0 +1,28 @@
...
+fixture(`Enroll Webauthn Authenticator`)
+  .requestHooks(mock);
+
+async function setup(t) {
+  const enrollWebauthnPage = new EnrollWebauthnAuthenticatorPageObject(t);
+  await enrollWebauthnPage.navigateToPage();
+  return enrollWebauthnPage;
+}
+
+// This is the only test we can perfrom in testcafe as webauthn api is not available in the version of chrome/chromeheadless
```
haishengwu-okta, line 21: just curious do u know why? at least your local, testcafe using my current latest chrome. (wonder if there are some control parameter that testcafe send to webauthn)

src/v2/view-builder/views/webauthn/EnrollWebauthnView.js
```diff
@@ -0,0 +1,126 @@
...
+  _stopEnrollment: function () {
+    this.$('.okta-waiting-spinner').hide();
+    this.$('.webauthn-setup').show();
+  },
+});
+
+export default BaseFactorView.extend({
+  Body,
+  postRender () {
+    this.$el.find('.o-form-button-bar [type="submit"]').remove();
```
haishengwu-okta, line 124: better to call `super.postRender` even though it might be empty.

src/v2/view-builder/views/webauthn/EnrollWebauthnView.js
```diff
@@ -0,0 +1,126 @@
...
+  className: 'oie-webauthn',
+  modelEvents: {
+    'error': '_stopEnrollment',
+  },
+  getUISchema () {
+    const schema = [];
+    const activationData = this.options.appState.get('currentAuthenticator').contextualData.activationData;
+    // Returning custom array so no input fields are displayed for webauthn
+    if (webauthn.isNewApiAvailable()) {
+      schema.push({
```
haishengwu-okta, line 28: nit:
instead of pushing multiple Views, you could create one View to contain all those following Views and that Single View to `schema`. Feels clearer but up to you.

src/v2/view-builder/views/webauthn/EnrollWebauthnView.js
```diff
@@ -0,0 +1,126 @@
...
+        excludeCredentials: getExcludeCredentials(activationData.excludeCredentials)
+      });
+      this.webauthnAbortController = new AbortController();
+      navigator.credentials.create({
+        publicKey: options,
+        signal: this.webauthnAbortController.signal
+      }).then((newCredential) => {
+        this.model.set({
+          credentials : {
+            clientData: CryptoUtil.binToStr(newCredential.response.clientDataJSON),
```
haishengwu-okta, line 97: Just for curiosity, does it make sense to change name in API so that we'd be able avoid hard code names.
- `clientData -> clientDataJSON`
- `assestation -> attestationObject`

test/testcafe/spec/EnrollWebauthnAuthenticator_spec.js
magizh-okta: yeah seems like that. Even if thats the case I am testing all the stuff in my unit test since I have to mock some functions like isWebauthn available or is Edge etc.

src/v2/view-builder/views/webauthn/EnrollWebauthnView.js
magizh-okta: Might be a bit late. @mpeng-okta ^

mpeng-okta: @haishengwu-okta I don't quite follow what is meant by "avoid hard code names". Would you mind elaborating -- are these not also hard-coded? Thanks!

haishengwu-okta: > "avoid hard code names"

The UI has to know map `response .clientDataJSON` to `clientData` that API expects.
If API sends UI name as `clientDataJSON`, the UI can do  `newCredential.response[ionFIeldName]` without worrying what's the actual ionFieldName.

src/v2/util/FactorUtil.js
```diff
@@ -31,9 +31,16 @@ const factorData = {
     iconClassName: 'mfa-okta-call',
   },
 
+  // Will get rid of this after fully implementing verify. OKTA-301557
   'webauthn': {
```
gowthamidommety-okta, line 35: I don't fully understand why we are keeping this.. Was this added for LEA?

playground/mocks/data/idp/idx/authenticator-select-enroll-options.json
```diff
@@ -118,7 +118,7 @@
       {
         "label": "Security Key or Biometric Authenticator (FIDO2)",
         "value": {
-          "type": "webauthn",
+          "type": "security_key",
```
gowthamidommety-okta, line 121: Did we decide the type would be 'security_key'?

mpeng-okta: Yeah, type will be security_key per https://github.com/okta/okta-core/blob/a12c55b0a748b3366dd94aa082e7d697bb7dcf2a/server/factors.interface/src/main/java/com/okta/services/factors/iface/types/OrgAuthenticatorType.java#L28 and https://github.com/okta/okta-core/blob/a12c55b0a748b3366dd94aa082e7d697bb7dcf2a/server/core/src/main/java/com/saasure/db/dto/core/entity/authenticators/securitykey/SecurityKeyAuthenticator.java#L23-L26

(Security Key is the authenticator type in which we'll have the WebAuthn method)

src/v2/util/FactorUtil.js
mpeng-okta: It was added not for LEA but for Oktane demo, I think?

src/v2/view-builder/views/webauthn/EnrollWebauthnView.js
mpeng-okta: Actually, I'm not sure about the name change, because the `clientDataJSON` is as named, but we're not actually sending `clientDataJSON` here -- we're sending an object that's the base64-encoded version of `clientDataJSON`, and calling it `clientData`.

I think I would therefore prefer to name it specifically as we're doing here, since it's not actually the JSON received from the WebAuthn API. Let me know if that makes sense to you as well.

haishengwu-okta: oh, I see. Thanks!

magizh-okta: It is not only that the clientData is and assestation is wrapped under credentials too. so at the very least SIW needs to put them into credentials. and that needs to be hardcoded.

PR #1226, opened 2020-06-09
test: adds support to update chromedriver dynamically based on chrome version

vijetmahabaleshwar-okta:
## Description:
* Every time travis upgrades chrome, we need to manually update our scripts to use the corresponding chromedriver
* This PR makes this process dynamic, wherein we parse the chromedriver downloads page to get the corresponding chromedriver version to use in `webdriver-manager update` command

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-299731](https://oktainc.atlassian.net/browse/OKTA-299731)


scripts/update-se-drivers.js
```diff
@@ -0,0 +1,65 @@
...
+
+got(chromeDriverDownloadUrl).then(async(response) => {
+  os = getOS();
+  console.log(`Operating System - ${os}`);
+
+  if (os === 'MacOS') {
+    chromeVersionString = execSync(`/Applications/Google\\ Chrome.app/Contents/MacOS/Google\\ Chrome --version`) + '';
+    // Example output of chromeVersionString - Google Chrome 83.0.4103.97
+    // Split this string by spaces to get the version number
+    chromeVersion = chromeVersionString.split(" ")[2];
```
haishengwu-okta, line 31: nit:
given the version number has only digits and dot, would be easier to use regex?

scripts/update-se-drivers.js
```diff
@@ -0,0 +1,65 @@
...
+    os = "MacOS";
+  } else if (os == "win32" || os == "win64") {
+    os = "Windows";
+  } else if (os == "linux") {
+    os = "Linux";
+  }
+  return os;
+}
+
+got(chromeDriverDownloadUrl).then(async(response) => {
```
haishengwu-okta, line 23: we already have `axios`. are we able to use it?

scripts/update-se-drivers.js
```diff
@@ -0,0 +1,65 @@
...
+    chromeVersion = execSync(`google-chrome --product-version`) + '';
+  }
+
+  console.log(`Chrome Version - ${chromeVersion}`);
+
+  const $ = cheerio.load(response.body);
+
+  // Chromedriver download URLs are of the form https://chromedriver.storage.googleapis.com/index.html?path=83.0.4103.39/
+  // Filter through all href of this format, depending on the chrome version installed and get the chrome driver version from that
+  $('a').filter(isChromeDriverDownloadUrl).each((i, link) => {
```
haishengwu-okta, line 42: how do you think if we do
- extra major chrome version
- fetch for `base-url/LATEST_RELEASE_83` to extra chrome version
- fetch for  `base-url/chrome-version-from-step-2`

scripts/update-se-drivers.js
vijetmahabaleshwar-okta: Updated to use `axios`

scripts/update-se-drivers.js
vijetmahabaleshwar-okta: regex makes sense. Updated.

scripts/update-se-drivers.js
vijetmahabaleshwar-okta: I'm not sure I understand. 
I'm trying to get the chromedriver version corresponding to the chrome binary version.
We can sync up to discuss this on slack.

haishengwu-okta: the way you're implementing now is 
- search through page https://chromedriver.chromium.org/downloads to find chromedriver version matched major chrome version.

what I am saying, given chrome major version, we could directly
- fetch file like : `https://chromedriver.storage.googleapis.com/LATEST_RELEASE_83` (83 is a major chrome version)
- the response of that url is just the chrome driver version.

vijetmahabaleshwar-okta: Now, I get it. Thanks!
This makes the script much more cleaner and simpler.
Have updated it. Take a look.

PR #1236, opened 2020-06-15
fix: Refactor footer css 

nikhilvenkatraman-okta:
Resolves: OKTA-298330

## Description:
- Refactor footer css to match designs 

## PR Checklist

- [x] Added e2e tests

### Screenshot/Video:

Screen Shot 2020-06-15 at 3 43 22 PM

Screen Shot 2020-06-15 at 3 43 32 PM

### Issue:

- [OKTA-298330](https://oktainc.atlassian.net/browse/OKTA-298330)

### DIFF
--- a/assets/sass/v2/_footer.scss
+++ b/assets/sass/v2/_footer.scss
@@ -1,13 +1,9 @@
 .siw-main-footer {
   .auth-footer {
+    display: flex;
+    flex-direction: column;
     .link {
       padding: 0.3rem 0;
-      display: block;
-    }
-    .js-cancel {
-      float: right;
-      padding: 0;
-      margin-top: 0.3rem;
     }
   }
 }
--- a/assets/sass/v2/_footer.scss
+++ b/assets/sass/v2/_footer.scss
@@ -1,11 +1,13 @@
 .siw-main-footer {
   .auth-footer {
     .link {
-      margin-right: 10px;
+      padding: 0.3rem 0;
+      display: block;
     }
     .js-cancel {
       float: right;
-      padding-bottom: 20px;
+      padding: 0;
+      margin-top: 0.3rem;
     }
   }
 }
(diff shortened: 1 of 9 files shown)


assets/sass/v2/_footer.scss
```diff
@@ -2,10 +2,14 @@
   .auth-footer {
     .link {
       margin-right: 10px;
+      padding: 8px 0;
```
anipendakur-okta, line 5: can we use rem values instead of px, helps in scaling on a browser-to-browser basis

src/v2/view-builder/internals/BaseFooter.js
```diff
@@ -25,15 +25,15 @@ export default View.extend({
   initialize () {
     let links = _.resultCtx(this, 'links', this);
     if (this.options.appState.get('showSignoutLink')) {
-      //add cancel/signout link
-      links = links.concat([
+      //add cancel/signout link as the first item of links since its floated right
+      links.unshift(
```
nikhilvenkatraman-okta, line 29: Screen Shot 2020-06-15 at 3 43 32 PM

This was done to make sure sign out link floats to the right.

anipendakur-okta:
LGTM

src/v2/view-builder/internals/BaseFooter.js
haishengwu-okta: Would like to chat with Vince to see if we can move `sign out` to left if you have not done so.

nikhilvenkatraman-okta: I did speak to him about this. It seems like for now `sign out` is the only link that they think will be on the right. Other actionable links will be on the left.
@vincentvoong-okta what are your thought on moving sign out link as well to the left ?

vincentvoong-okta: @nikhilvenkatraman-okta Moving `Sign out` to the left side makes sense and should be below the other options. A note - it should only appear in the appropriate context (e.g. Sign out should only be an option if the user is already signed in. The example above shows Sign out but still asks for the users password)

nikhilvenkatraman-okta: cool, I can update this to be left aligned with sign out as the last option.

Screen Shot 2020-06-16 at 11 50 17 AM

@vincentvoong-okta in most places we show the sign out in the right context. Here the user has entered identifier in the first step before getting to this screen, hence we show the sign out.

PR #1237, opened 2020-06-15
feat: support idp config

haishengwu-okta:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

hw-authenticator-social-idp

### Issue:

- https://oktainc.atlassian.net/browse/OKTA-300033


src/v2/BaseLoginRouter.js
```diff
@@ -109,7 +109,7 @@ export default Router.extend({
     // transform response
     const ionResponse = _.compose(
```
pradeepdewda-okta, line 110: why we need to use compose, bind and pass idxResponse as argument and make this complex.

I feel this kind of complexities can be avoided by simplifying it.

haishengwu-okta: I feel it's easier to understand personally. 

Otherwise it would be (guess it's more tedious when we have 3 transformers hence we have to make 3 temporary variables.

```
const tranformedResp = responseTranformer(this.settings, idpResp);
const tranformedRespWithUiSchema = uiSchemaTransform(tranformedResp);
this.appState.setIonResponse(tranformedRespWithUiSchema);
```

I wouldn't like

```
const tranformedRespWithUiSchema = uiSchemaTransform(responseTranformer(this.settings, idpResp));
this.appState.setIonResponse(tranformedRespWithUiSchema);
```

pradeepdewda-okta: I feel it's much easier to understand and not have to rely on maintaining the sequencing that is needed for compose. what if someone changes the sequence in _.compose ?

src/v2/ion/responseTransformer.js
```diff
@@ -75,28 +75,66 @@ const getRemediationValues = (idx) => {
...
+        name: idpConfig.text,
+      };
+      if (idpConfig.className) {
+        idp.className = idpConfig.className;
+      }
+      return {
+        name: 'redirect-idp',
+        type: idpConfig.type,
+        idp,
+        href: `${baseUrl}/sso/idps/${idpConfig.id}?stateToken=${stateHandle}`
```
pradeepdewda-okta, line 106: change this href to get a value from a variable

src/v2/BaseLoginRouter.js
haishengwu-okta: > what if someone changes the sequence in _.compose

Wouldn't we worry about people change sequence in others cases?

pradeepdewda-okta: other's cases means ?

haishengwu-okta: @pradeepdewda-okta meaning if we write code in other ways like I posted earlier, the sequence still matters.

src/models/Settings.js
```diff
@@ -19,22 +19,17 @@ define([
...
   'util/Logger',
+  'util/IDP',
   'config/config.json'
-],
-function (Okta, Q, Errors, BrowserFeatures, Util, Logger, config) {
+], function (Okta, Q, Errors, BrowserFeatures, Util, Logger, IDP, config) {
 
   var SharedUtil = Okta.internal.util.Util;
-
+  var _ = Okta._;
```
anipendakur-okta, line 27: Hi, just curious as to why we are not using ES6 (let/const) through out the file. Even if this file is not transpiled looks like IE11 supports let / const (as long as we don't use a for loop in which case the behavior could be different when not transpiled).

pradeepdewda-okta: this part of old code.

haishengwu-okta: there is no reason to not use `let/const` but trying to be consistent in the old code.

PR #1240, opened 2020-06-17
fix: password enroll

anipendakur-okta:
## Description:
Integration for password enroll and tweaking of labels and footer sign out link

## PR Checklist

- [ ] Added unit tests?

- [X] Added e2e tests

### Screenshot/Video:
**Note**:

- I took the screenshots with my local SIW pointing to an IDX enabled org.
- Also, I have updated the title to `Set up password` in code (not shown in screenshots).

image
image

### Reviewers:
@haishengwu-okta @nikhilvenkatraman-okta 

### Issue:

- [OKTA-296702](https://oktainc.atlassian.net/browse/OKTA-296702)


src/v2/models/AppState.js
```diff
@@ -54,10 +54,10 @@ export default Model.extend({
     showSignoutLink: {
       deps: ['idx'],
       fn: function (idx = {}) {
-        const invalidSignOutSteps = ['IDENTIFY', 'ENROLL', 'SUCCESS'];
-        // hide signout for IDENTIFY, ENROLL & SUCCESS step
+        const invalidSignOutSteps = ['IDENTIFY', 'LOGIN', 'ENROLL', 'SUCCESS'];
```
anipendakur-okta, line 57: Note: The `intent` here (seen through debugger) seems to be `LOGIN` and not `ENROLL`. However, for `LOGIN` we don't need to show sign out link as well?

src/v2/models/AppState.js
```diff
@@ -54,10 +54,10 @@ export default Model.extend({
...
+        // Hide signout for IDENTIFY, LOGIN, ENROLL & SUCCESS intents
         return idx.actions
-          && _.isFunction(idx.actions.cancel) && !invalidSignOutSteps.includes(idx.context.step);
+          && _.isFunction(idx.actions.cancel) && !invalidSignOutSteps.includes(idx.context.intent);
```
anipendakur-okta, line 60: Seems like there's no `step` but `intent` instead. Please let me know if I'm missing something here.

haishengwu-okta: I don't think. `step` has been removed. `intent` isn't sufficient given `signout` is also more ux concern.
e.g. `identify` has cancel but I doubt we want to show `signout` link.
I'm thinking eventually we have to rely on remediation form name.
any thoughts @nikhilvenkatraman-okta

playground/mocks/config/responseConfig.js
```diff
@@ -4,7 +4,8 @@ const idx = {
   // ===== IDX
 
   '/idp/idx/introspect': [
-    'authenticator-select-verify-options',
+    'factor-enroll-password',
```
haishengwu-okta, line 7: I'd like to use `authenticator-enroll-password`

playground/mocks/data/idp/idx/authenticator-enroll-password.json
```diff
@@ -20,7 +20,7 @@
               "value": [
                 {
                   "name": "passcode",
-                  "label": "Password",
+                  "label": "Enter password",
```
anipendakur-okta, line 23: Opened up a ticket to fix this - https://oktainc.atlassian.net/browse/OKTA-305923

src/v2/view-builder/ViewFactory.js
```diff
@@ -60,7 +60,7 @@ const VIEWS_MAPPING = {
   },
   'enroll-factor': {
     email: RequiredFactorEmailView, // TODO EnrollFactorEmailView is unimplemented
-    password: EnrollFactorPasswordView,
+    password: EnrollAuthenticatorPasswordView,
```
haishengwu-okta, line 63: do not change anything `factor` related. just keep there and we will clean up once for all (probably post-m1)

src/v2/models/AppState.js
nikhilvenkatraman-okta: yes ,we can no longer use `step` and I dont think `intent` is right either. We need to change this logic to use the form names instead. Should we handle that in a different ticket and keep this the same here for now ?

anipendakur-okta: It makes sense to open another ticket - [OKTA-305970](https://oktainc.atlassian.net/browse/OKTA-305970) - keeping this as is in this PR. We can use the new ticket for verifying all the screens as well.

test/testcafe/spec/EnrollFactorPassword_spec.js
```diff
@@ -1,20 +1,20 @@
 import { RequestMock } from 'testcafe';
-import FactorEnrollPasswordPageObject from '../framework/page-objects/FactorEnrollPasswordPageObject';
+import EnrollPasswordPageObject from '../framework/page-objects/EnrollPasswordPageObject';
 import SuccessPageObject from '../framework/page-objects/SuccessPageObject';
-import xhrFactorEnrollPassword from '../../../playground/mocks/data/idp/idx/factor-enroll-password';
+import xhrAuthenticatorEnrollPassword from '../../../playground/mocks/data/idp/idx/authenticator-enroll-password';
```
anipendakur-okta, line 4: @haishengwu-okta is this change ok? that means we're moving away from `factor-enroll-password` , or I can create another test file for `authenticator-enroll-password`

src/v2/view-builder/views/password/EnrollFactorPasswordView.js
```diff
@@ -4,19 +4,19 @@ import BaseForm from '../../internals/BaseForm';
 import BaseFactorView from '../shared/BaseFactorView';
```
haishengwu-okta, line 4: don't you need to copy `EnrollFactorPasswordView.js` into `EnrollAuthenticatorPasswordView.js` and make changes to it?

anipendakur-okta: I am probably missing context on why we need to copy. `EnrollFactorPasswordView` does all the work we need for Password Enroll.

haishengwu-okta: `EnrollFactorPasswordView` is to support enroll password for LEA. what if there are some selenium tests in monolith that verify this page? your change **might** breaking selenium tests. Sure, you can fix selenium test when upgrade widget in monolith. Wouldn't better we save this effort later? In other words, would be better that your changes are backward compatible.
Copy into `EnrollAuthenticatorPasswordView` (also update ViewFactor) would be just additive work and no harm.

test/testcafe/framework/page-objects/EnrollAuthenticatorPhoneObject.js
```diff
@@ -5,7 +5,7 @@ const VOICE_RADIO_SELECTOR = 'input[value="voice"]';
 const PHONE_NUMBER_SELECTOR = '.phone-authenticator-enroll__phone';
 const PHONE_NUMBER_EXTENSION_SELECTOR = '.phone-authenticator-enroll__phone-ext';
 const phoneFieldName = 'authenticator\\.phoneNumber';
-export default class EnrollPasswordPageObject extends BasePageObject {
+export default class EnrollPhonePageObject extends BasePageObject {
```
haishengwu-okta, line 8: nit:
just make sure class name is same as file name.

PR #1243, opened 2020-06-18
feat: add api version support for idx-js

nikhilvenkatraman-okta:
Resolves: OKTA-302538

## Description:

- Start sending `version` while calling idx.start 

## PR Checklist

- [x] Added unit tests?

### Screenshot/Video:

https://okta.box.com/s/aofeip2kwuryesw7db5wgqihqj4czloh

### Issue:

- [OKTA-302538](https://oktainc.atlassian.net/browse/OKTA-302538)


src/widget/OktaSignIn.js
```diff
@@ -36,18 +36,19 @@ var OktaSignIn = (function () {
         Router = require('v2/WidgetRouter');
       }
       if (widgetOptions.stateToken) {
+        router = bootstrapRouter.call(this, Router, authClient, widgetOptions, renderOptions, successFn, errorFn);
+        if (!widgetOptions.apiVersion) {
+          widgetOptions.apiVersion = router.settings.get('apiVersion');
```
haishengwu-okta, line 41: `widgetOptions` already has `apiVersion`.
`router.settings` is actually initialized from `widgetOptions`

nikhilvenkatraman-okta: If we missed specifying the apiVersion , then we want to get the default `1.0.0` from settings and I am setting widgetOptions.apiVersion so that we can call idx-js with `1.0.0`

haishengwu-okta: oh I see.. 🤦

nikhilvenkatraman-okta: I did this to avoid breaking the local widget for everyone when they pull this change in as they might not have apiVersion set locally 😃

src/util/BaseLoginRouter.js
```diff
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-/* eslint max-params: [2, 16], max-statements: [2, 18] */
```
haishengwu-okta, line 13: nit:
revert changes.

src/v2/controllers/FormController.js
```diff
@@ -32,6 +32,9 @@ export default Controller.extend({
 
   postRender () {
     const currentViewState = this.options.appState.getCurrentViewState();
+    if (!currentViewState) {
```
haishengwu-okta, line 35: curious when we run into this?
when error happened, we wouldn't render this form, do we?

src/v2/view-builder/ViewFactory.js
```diff
@@ -98,6 +98,9 @@ const VIEWS_MAPPING = {
     [DEFAULT]: TerminalView,
     'email': TerminalReturnEmailView,
   },
+  'terminal-widget-error': {
```
haishengwu-okta, line 101: looks like can be reverted.

src/widget/OktaSignIn.js
haishengwu-okta, line 41: doesn't change `widgetOptions` that user passed in place?
if yes, we'd better create a new object.

src/v2/controllers/FormController.js
nikhilvenkatraman-okta: Yeah, it does try to render this form, I think its because BaseLoginRouter calls this.defaultAuth(). This was another reason why I was forcing it to render a terminal view in my previous implementation.

haishengwu-okta: maybe remove `this.defaultAuthn`? feel like it makes sense for v1 (render primary auth) but not v2 ?

nikhilvenkatraman-okta: just removing that is not enough, as the BaseController still calls render method which then ends up in postRender within FormController

Issue #1038, opened 2020-02-10
country / language assets cannot be found


BassplayerPatrick:
using the latest update of the okta-signin-widget, version 3.7.1, some warnings appear in the console about some language and country related jsonp files that cannot be found.

for instance: the following should be able for download, but doesn't exist:

https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/3.7.1/labels/jsonp/country_de.jsonp?callback=jsonp_country

https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/3.7.1/labels/jsonp/login_nl_NL.jsonp?callback=jsonp_login

Please fix ASAP

haishengwu-okta:
@BassplayerPatrick 
You're suppose to use link like `https://global.oktacdn.com/okta-signin-widget/3.1.0/js/okta-sign-in.min.js`. Please check https://github.com/okta/okta-signin-widget#using-the-okta-cdn

BassplayerPatrick:
Incorrect. I'm not downloading the widget, the widget downloads the assets,
which are unavailable. It's not customizable code.

Op ma 10 feb. 2020 17:38 schreef Haisheng Wu <notifications@github.com>:

haishengwu-okta:
@BassplayerPatrick How do you load/initialize widget in your page?

BassplayerPatrick:
I initialize the widget like in the examples

To be clear. Up until version 3.5.0 this was not an issue. We updated to
3.7.1 without any other changes and encountered this problem.

Op ma 10 feb. 2020 20:07 schreef Haisheng Wu <notifications@github.com>:

haishengwu-okta:
@BassplayerPatrick 
- Can you show us the code how you load the widget (not how to initialize it)
    - I bet you loads widget from domain `https://ok1static.oktacdn.com/`
- Your issue will be resolved by itself after several weeks (depends on our release schedule) or you load widget from `https://global.oktacdn.com`. (read my first response please)

BassplayerPatrick:
I'm not loading the widget dynamically. I have the package locally. If the
url got changed , then the widget has to be updated with that. I shouldn't
be bothered with it.

Op di 11 feb. 2020 02:15 schreef Haisheng Wu <notifications@github.com>:

haishengwu-okta:
@BassplayerPatrick sorry about that. I think I found the root cause now
[...]

PR #1251, opened 2020-06-19
fix: use global CDN for language assets OKTA-234964

aarongranick-okta:
## Description:

Changes default baseURL for CDN assets to the global CDN. This fixes an issue where language assets are not available on patch/minor versions that have not been deployed to okta1 cdn.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Reviewers:
@haishengwu-okta 

### Issue:

- [OKTA-234964](https://oktainc.atlassian.net/browse/OKTA-234964)


aarongranick-okta:
> The change itself looks good! But this is the first time I'm seeing a global.oktacdn.com. Curious about the deploy cadence to that.

This is where we deploy to when we do the widget release. The other CDN (oktaX) is managed by monolith release. All static assets used by the web site will be on monolith CDN (such as loginpage bundle). I believe the baseUrl is also set on Okta pages to match monolith CDN. So this change will mainly affect people who are self-hosting the widget or loading it from the CDN as described here:

https://github.com/okta/okta-signin-widget#using-the-okta-cdn

aarongranick-okta:
Will resolve issue: https://github.com/okta/okta-signin-widget/issues/1038

santhoshbalakrishnan:
Ah cool! Thanks for the explanation. Nice change!

Issue #408, opened 2018-03-14
Unsafe eval


roberttenbrincke:
This widget requires the 'unsafe-eval' CSP header to be added (or the use of a nonce-based approach) in order to work correctly. Is it possible to make the widget operate without resorting to eval()?

Further investigation suggests this is caused by Handlebars.

nbarbettini:
It's not possible to avoid eval() right now. This is due to how the widget uses Handlebars internally (as you noticed).

We're investigating how to remove this requirement. It 's a longer-term project, but it is on our radar.

roberttenbrincke:
Handlebars offers recompilation. Would that be a viable route?

XaxisJosh:
any updates on this?

swiftone:
@XaxisJosh - Thanks for the followup.  No updates at the moment (it remains on our radar) but the issue is getting more attention (see also https://github.com/okta/okta-signin-widget/issues/902 ).

If this issue accumulates enough +1s and/or our business unit sees enough demand, this will rise in priority.

omgitstom:
https://github.com/wycats/handlebars.js/blob/master/FAQ.md#why-doesnt-this-work-with-content-security-policy-restrictions

haishengwu-okta:
https://github.com/wycats/handlebars.js/issues/1443

bretterer:
internal ref: OKTA-156161

CPedrini:
Any update?

aarongranick-okta:
@CPedrini The changes have been implemented internally and we are currently testing it. We expect to release a CSP-compliant widget soon.

zustellr:
@aarongranick-okta do you have more information on when "soon" is expected to be? Thanks!

aarongranick-okta:
@zustellr @CPedrini @afalahi @XaxisJosh @roberttenbrincke This has been released. Try updating widget to the latest version, let us know how it goes!

matthew-borland:
FYI I was just now using version 4.5.2 of the Okta signin widget, and I appeared to still have this problem.  I will take a further look at this.

matthew-borland:
OK, I verified CSP without unsafe-eval appears to work with this version; sorry for the confusion!

swiftone:
Closing issue as confirmed fixed

melcor76:
> OK, I verified CSP without unsafe-eval appears to work with this version; sorry for the confusion!

Hi, how did you solve your issue after it appeared to not be working?
[...]

PR #1257, opened 2020-06-24
fix: widget CSP issues

aarongranick-okta:
## Description:

Is compliant with strict CSP policies. CSP policy is enforced in playground and in karma tests. Karma tests also enforce that all templates are precompiled.

Will resolve https://github.com/okta/okta-signin-widget/issues/408

Includes alpha version of Courage built from this branch: https://github.com/okta/okta-ui/pull/2127

The files being added under `packages/@okta/babel-plugin-handlebars-inline-precompile/` are copied from okta-ui/courage by the `courage-for-signin-widget` build script

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Reviewers:
@haishengwu-okta 
@rajaguruduraisamy-okta 
@jmelberg-okta 

### Issue:

- [OKTA-286452](https://oktainc.atlassian.net/browse/OKTA-286452)


packages/@okta/courage-for-signin-widget/scripts/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -x
+
+# Copy handlebars precompile babel plugin
+mkdir -p ../babel-plugin-handlebars-inline-precompile
+cp -rf ./node_modules/@okta/babel-plugin-handlebars-inline-precompile/* ../babel-plugin-handlebars-inline-precompile/
```
haishengwu-okta, line 5: does the webpack copy plugin support copy dir?

src/EnrollCallAndSmsController.js
```diff
@@ -24,11 +27,13 @@ define([
 
   var _ = Okta._;
   var { Keys } = Okta.internal.util;
-  var warningTemplate = '<div class="okta-form-infobox-warning infobox infobox-warning login-timeout-warning">\
+  var warningTemplate = Okta.View.extend({
+    template: hbs('<div class="okta-form-infobox-warning infobox infobox-warning login-timeout-warning">\
```
haishengwu-okta, line 31: nit:
shall we follow to avoid any DOM changes
```
View.extend({
  className: 'okta-form-infobox-warning....',
  template: hbs('<span...></span><p>..</p>')
})

File: src/util/Bundles.js [hunk 3]
```diff
@@ -141,10 +138,7 @@ define([
     // zh-CN -> zh_CN
     languageCode = language.replace('-', '_');
 
-    path = assets.rewrite(bundlePathTpl({
-      bundle: bundle,
-      languageCode: languageCode
-    }));
+    path = assets.rewrite(`/labels/json/${bundle}_${languageCode}.json`);
```
haishengwu-okta, line 141: shall it run through `encodeURI` ?

src/v2/view-builder/internals/BaseView.js
```diff
@@ -14,9 +15,12 @@ export default View.extend({
 
   className: 'siw-main-view',
 
-  template: '<div class="siw-main-header"></div>' +
-      '<div class="siw-main-body"></div>' +
-      '<div class="siw-main-footer"></div>',
+  template: hbs`
+    <div class="siw-main-header"></div>
+    <div class="siw-callout"></div>
```
haishengwu-okta, line 20: why add `siw-callout` ?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -143,9 +146,13 @@ const Body = BaseForm.extend(Object.assign(
 
     doCustomURI () {
       this.ulDom && this.ulDom.remove();
-      this.ulDom = this.add(`
-        <iframe src="${this.customURI}" id="custom-uri-container" style="display:none;"></iframe>
-      `).last();
+      this.ulDom = this.add(View.extend({
+        template: hbs`<iframe src="{{{customURI}}}" id="custom-uri-container" style="display:none;"></iframe>`
```
haishengwu-okta, line 150: just use the pure string as before? or does the `template` have to be handlerbar template now?

test/testcafe/spec/EnrollFactorPassword_spec.js
```diff
@@ -31,16 +31,16 @@ test(`should have both password and confirmPassword fields and both are required
   // fields are required
   await enrollPasswordPage.clickNextButton();
   await enrollPasswordPage.waitForErrorBox();
-  await t.expect(enrollPasswordPage.getPasswordError()).eql('This field cannot be left blank');
-  await t.expect(enrollPasswordPage.getConfirmPasswordError()).eql('This field cannot be left blank');
+  await t.expect(enrollPasswordPage.getPasswordError()).contains('This field cannot be left blank');
```
haishengwu-okta, line 34: hmm, why cannot use `eql`?

test/unit/main.js
```diff
@@ -1,3 +1,5 @@
+import './karma/karma-enforce-precompile';
```
haishengwu-okta, line 1: you have added `customHeaders` in karma.conf. why still need this?

aarongranick-okta: The custom headers will detect things like `eval` which we will not have as long as we are using `handlebars/runtime`. If we accidentally switch to full handlebars, the CSP will start failing again.

This check in karma detects a template that has not been precompiled. This would not throw a CSP error, instead the template would not be properly rendered. Depending on how the test is written we may not detect that or it may be confusing to the developer as to why the test is failing.

test/testcafe/spec/EnrollFactorPassword_spec.js
aarongranick-okta: The precompiled templates sometimes have an extra whitespace in them (if the template in the sourcefile has line breaks and empty spaces). Most of these are stripped, but there are some cases where an extra space will stay in. (Input label is one of these cases)

src/v2/view-builder/views/DeviceChallengePollView.js
aarongranick-okta: At this point in time, we are allowing `add()` with a string. As long as it doesn't have handlebars in it, there is technically no problem with that. Before I added the karma enforce precompile check, I was converting some of these cases to be Views with templates because I had a warning for using `add()` with a bare string. In this case we can probably go back to pure string.

src/v2/view-builder/internals/BaseView.js
aarongranick-okta: Good catch. I think this is probably a mistake from doing rebases/merges over the last couple months. I'll fix it.

test/testcafe/spec/EnrollFactorPassword_spec.js
haishengwu-okta: `getPasswordError` tries to get inner text of following `p` element.
Wouldn't expect extra line break ? 
so maybe `trim` whilespace so that we can still use `eql` ?

```
<p id="input-container-error6" class="okta-form-input-error o-form-input-error o-form-explain" role="alert"><span class="icon icon-16 error-16-small" role="img" aria-label="Error"></span>This field cannot be left blank</p>
```

packages/@okta/courage-for-signin-widget/scripts/build.sh
aarongranick-okta: I was running into an issue using webpack to do the copy. This babel plugin is used BY the webpack at a very early phase, so the plugin needs to be in place before we start compiling any source files. There may be a way to get it to work. If this is important, I can spend some more time on it.

test/unit/main.js
haishengwu-okta: Then shall the changes made to framework itself (`courage-for-signin-widget/src/`)?

src/EnrollCallAndSmsController.js
aarongranick-okta: I'm not exactly sure how that works, this is related to how courage is implemented?

src/EnrollYubikeyController.js
```diff
@@ -50,7 +51,9 @@ function (Okta, FormType, FormController, Footer, TextBox) {
       className: 'enroll-yubikey',
       formChildren: [
         FormType.View({
-          View: '<div class="yubikey-demo" data-type="yubikey-example"></div>'
+          View: Okta.View.extend({
+            template: hbs`<div class="yubikey-demo" data-type="yubikey-example"></div>`
```
aarongranick-okta, line 55: note: may not need to be a handlebars template. I will check this.

src/util/Bundles.js
aarongranick-okta: It probably won't hurt. I'll give it a try.

test/testcafe/spec/EnrollFactorPassword_spec.js
aarongranick-okta: I'll take another look at this

packages/@okta/courage-for-signin-widget/scripts/build.sh
haishengwu-okta: The problem is using wrapper build script is it doesn't support `webpack` parameters (although there is probably a quick and may not matter much for this package)
Would be good to avoid if possible (separate PR is fine).

src/EnrollCallAndSmsController.js
haishengwu-okta: yeah.
`View` will add another parent (`div` by default) around the `template`

test/testcafe/spec/EnrollFactorPassword_spec.js
haishengwu-okta: can be separate PR if it's not quick fix.

webpack.test.config.js
```diff
@@ -17,10 +17,10 @@ plugins.push(new webpack.DefinePlugin({
   SDK_VERSION: JSON.stringify(SDK_VERSION)
 }));
 
-testConfig.module.rules.push({
-  test: /\.js$/,
-  use: ['source-map-loader'],
-  enforce: 'pre'
+testConfig.module.rules.push({
```
haishengwu-okta, line 20: remove extra space

packages/@okta/courage-for-signin-widget/scripts/build.sh
aarongranick-okta: I see, you'd like to do something like `yarn build --watch` (instead of `npx webpack --watch`)?

test/unit/main.js
aarongranick-okta: I did originally have this check in CourageForSigninWidget.js, but I decided to move it to karma because I didn't want a runtime penalty of looking for `{{` in every template string, and I also want to be careful about causing a runtime error visible to users. Seeing a UI with `{{` in it is a lower priority bug than having an exception which breaks the whole login flow. Of course the karma check only works if templates are rendered within unit tests.  Another option would be an eslint  plugin to detect strings with `{{` that are not wrapped in `hbs`.

packages/@okta/courage-for-signin-widget/scripts/build.sh
haishengwu-okta: yeah and other parameters for `webpack` command.
it might not be big deal for this module but always trying to avoid build wrapper if possible.

test/unit/main.js
haishengwu-okta: `eslint plugin` sounds better!

test/testcafe/spec/EnrollFactorPassword_spec.js
aarongranick-okta: I investigated, this is already fixed/tightened up on the courage side, so this must have been from an earlier iteration. I'll put it back to `eql`

eslint-local-rules.js
```diff
@@ -0,0 +1,5 @@
+var noBareTemplates = require('./packages/@okta/eslint-plugin-okta-ui/lib/rules/no-bare-templates');
+
+module.exports = {
+  'no-bare-templates': noBareTemplates
```
haishengwu-okta, line 4: 👍

packages/@okta/courage-for-signin-widget/scripts/build.sh
aarongranick-okta: I moved this to webpack copy and removed the build script

test/unit/main.js
aarongranick-okta: I updated the eslint plugin on the courage side and am copying it over with the build from courage-for-signin-widget.

packages/@okta/courage-for-signin-widget/package.json
```diff
@@ -4,11 +4,12 @@
...
   "main": "src/CourageForSigninWidget.js",
   "scripts": {
-    "build": "scripts/build.sh"
+    "build": "webpack"
   },
   "devDependencies": {
-    "@okta/babel-plugin-handlebars-inline-precompile": "3.1.0-alpha.3586.gd86894f",
-    "@okta/courage": "4.6.0-alpha.3586.gd86894f",
+    "@okta/babel-plugin-handlebars-inline-precompile": "3.1.0-alpha.3660.g8faa6ac",
+    "@okta/eslint-plugin-okta-ui": "0.1.0-alpha.3660.g8faa6ac",
```
haishengwu-okta, line 11: what's `@okta/eslint-plugin-okta-ui` for?

aarongranick-okta: Thats where the `no-bare-templates` eslint rule is coming from. The same rule is run on courage

packages/@okta/courage-for-signin-widget/package.json
```diff
@@ -4,11 +4,12 @@
   "description": "generates a Courage version for Sign-In Widget",
   "main": "src/CourageForSigninWidget.js",
   "scripts": {
-    "build": "scripts/build.sh"
+    "build": "webpack"
   },
   "devDependencies": {
-    "@okta/babel-plugin-handlebars-inline-precompile": "3.1.0-alpha.3586.gd86894f",
-    "@okta/courage": "4.6.0-alpha.3586.gd86894f",
+    "@okta/babel-plugin-handlebars-inline-precompile": "3.1.0-alpha.3660.g8faa6ac",
```
haishengwu-okta, line 10: This module `packages/@okta/courage-for-signin-widget/` getting more responsibility rather than just pulling Courage.
Some further refactoring seems necessary. (addressed separately) 
One idea:
- rename to more general module name like `okta-internal-modules`
- add script per module like
    - `build-courage`
    - `build-handlebar-inline-precompile`
    - `build-eslint-rule`
    - `build` compose all above (and apparently can be done in parallel)

aarongranick-okta:
@haishengwu-okta Created new JIRA OKTA-311926 for `courage-for-signin-widget` refactor

packages/@okta/babel-plugin-handlebars-inline-precompile/handlebars/patch-precompile.js
```diff
@@ -0,0 +1,17 @@
+// Monkey patch handlebars to remove extra whitespace from templates before they are precompiled
+module.exports = function (Handlebars) {
+  const precompile = Handlebars.precompile;
+  Handlebars.precompile = function (template) {
+    // Trim extra whitespace from template
+    template = template
+      .replace(/\s+</g, '<') // all spaces before opening tag
+      .replace(/>\s+/g, '>') // all spaces after closing tag
+      .replace(/}}\s+{{/g, '}}{{') // all spaces betwen mustaches
```
jmelberg-okta, line 9: Is this a safe operation? In admin we have many templates that rely on chaining one value to another with a space in between. Looks like the closest occurrence here (uses a slightly different format) is in [FactorsDropDownOptions.js](https://github.com/okta/okta-signin-widget/blob/d294c668f1a2fffb5078bd2f9ef72f39cb400d2b/src/views/mfa-verify/dropdown/FactorsDropDownOptions.js#L21).

If we omit this substitution, does anything break?

packages/@okta/courage-for-signin-widget/webpack.config.js
```diff
@@ -6,10 +6,12 @@ const CopyWebpackPlugin = require('copy-webpack-plugin');
 const PACKAGE_JSON = require('./package.json');
 
 const EMPTY = resolve(__dirname, 'src/empty');
-const SHARED_JS = resolve(__dirname, 'node_modules/@okta/courage/src');
-const COURAGE_DIST = resolve(__dirname, 'node_modules/@okta/courage/dist');
-const PUBLISH_DIR = resolve(__dirname, '../courage-dist');
-const I18N_DIR = resolve(__dirname, '../i18n');
+const NODE_MODULES_SRC = resolve(__dirname, 'node_modules/@okta');
+const NODE_MODULES_DEST = resolve(__dirname, '..');
```
jmelberg-okta, line 10: super-nit: Switch the order of these vars to gradually increase:

```suggestion
const NODE_MODULES = resolve(__dirname, '..');
const OKTA_NODE_MODULES = resolve(NODE_MODULES_DEST, '@okta');
```

packages/@okta/eslint-plugin-okta-ui/lib/rules/no-bare-templates.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+  return false;
+}
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Detect use of bare (unwrapped) handlebars template strings',
+      category: 'CSP Issues',
+      recommended: true,
```
jmelberg-okta, line 33: Would also be nice to provide logic for `--fix` since the change is minimal.

packages/@okta/babel-plugin-handlebars-inline-precompile/handlebars/patch-precompile.js
aarongranick-okta: This file is copied from `okta-ui` by the `courage-for-signin-widget` build. There is a separate PR for these changes over here: https://github.com/okta/okta-ui/pull/2127

This prevents extra (unintentional) whitespace in the precompiled templates. Without this, the bundle size is larger and also many tests would need to be changed to account for extra whitespace.

packages/@okta/eslint-plugin-okta-ui/lib/rules/no-bare-templates.js
aarongranick-okta: Good suggestion. This file is also copied from `okta-ui`, the original source is in a separate PR: https://github.com/okta/okta-ui/pull/2127

aarongranick-okta:
Changes can be tested in monolith using SIW version `4.2.0-alpha.gb8d39a3` https://bacon-go.aue1e.saasure.net/#!/commits/okta-signin-widget/branch/bacon-topic-ag-siw-csp

packages/@okta/babel-plugin-handlebars-inline-precompile/handlebars/patch-precompile.js
jmelberg-okta: Both the changes proposed in [okta/okta-ui#2127](https://github.com/okta/okta-ui/pull/2127) and here have this potential side-effect.

In `okta-ui`, I'd assume we'd need to remove this line to make sure chained template values that expect spaces are met.

Examples:
- [PendingRequestsView.js](https://github.com/okta/okta-ui/blob/master/packages/admin/appinstance/src/assignments/views/PendingRequestsView.js#L72)
- [DevicesList.js](https://github.com/okta/okta-ui/blob/master/packages/admin/devices/src/views/DevicesList.js#L53)
- [UserPicker.js](https://github.com/okta/okta-ui/blob/master/packages/admin/shared/src/views/ud/mappings/components/UserPicker.js#L30)
- [SearchUtil](https://github.com/okta/okta-ui/blob/master/packages/admin/syslog/src/util/SearchUtil.js#L95)

How much extra of a bundle size does the removal of this space save? Refactoring our template logic to include spaces at the `*.properties` or passed param level (i.e `{ space: '  '}`) doesn't seem like the best option.

cc: @rajaguruduraisamy-okta

patrickmcdougle-okta: 👋🏻  Hi everyone, coming to you from the future as someone who has been tasked with updating SearchUtil.js to use precompiled templates and I just want to say that @jmelberg-okta was absolutely right this change is problematic there and the only way to proceed is to replace handlebars with es6 template literals.

PR #1258, opened 2020-06-24
feat: add forgot password link at identify page

haishengwu-okta:
## Description:

- add `forgot password` link to identify page when `currentAuthenticatorEnrollment-recover` is available.
- refactorings
    - remove `httpClient` which has been deprecated when integrates to idxjs.
    - rename `shared/BaseFactorView` to `components/BaseAuthenticatorView`
    - creates `AuthenticatorVerifyFooter.js` for all challenge authenticator views.
    - rename `AuthenticatorUtil` to `LinksUtil` to be more explicit.
    - delete `BaseFactorTerminalView.js` which adds complexity.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

forgot-password-at-identify-page

### Issue:

- OKTA-305911


src/v2/view-builder/components/AuthenticatorVerifyFooter.js
```diff
@@ -0,0 +1,8 @@
+import BaseFooter from '../internals/BaseFooter';
```
nikhilvenkatraman-okta, line 1: Why is this called AuthenticatorVerifyFooter? and not something more generic?

haishengwu-okta: Aren't only page for authenticator verify having `use something else` link?
Otherwise we could rename it to more generic.

nikhilvenkatraman-okta: I see ok, that makes sense

PR #1268, opened 2020-06-27
chore: add playground readme

haishengwu-okta:
OKTA-309015

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

### DIFF
--- a/playground/README.md
+++ b/playground/README.md
@@ -29,8 +29,8 @@ The mock server is implemented using [dyson.js](https://www.npmjs.com/package/dy
 - `spec-okta-api` has mocks for Okta APIs, e.g. `/api/v1/*`, `/idp/idx/*`, etc
 - `spec-device-authenticator` is for mocking 3rd-party device authenticator application.
 
-Most of cases, the response is from a JSON file, it would be good to have abstraction to avoid tedious copy/parse. Take a look at `config/templateHelper` regardling how to handle the common use in general fashion.
-Therefore in order to create mock for a URL which response is from a JSON file, e.g. `/idp/idx/foo`, you have to do two things:
+In most cases, the response is coming from a JSON file. It would be good to have abstractions to avoid tedious copy/paste. Take a look at config/templateHelper to get a sense of how to handle the common use cases in a generic fashion.
+In order to create mock, i.e have a url /idp/idx/foo, respond with JSON stored in my-foo.json, you need to do two things:
 
 1. Adds URL to `idx` object in `spec-okta-api/idp/idx/index.js`
 2. Specify JSON file name in `idx` object in `config/responseConfig` like
@@ -41,8 +41,8 @@ Therefore in order to create mock for a URL which response is from a JSON file, 
 ],
 ```
 
-Why the value is a List instead of a single String? Well, this is the poor man's solution to dynamic responses.
-It means you got different API response depends on the number of times you hit the URL.
+You may wonder why the value is a List instead of a single String? Well, this is the poor man's solution for providing dynamic responses.
+It means that you get different API response every time you hit the URL.
 
 ```
 index = (number of times - 1) `mod` (size of mock data array)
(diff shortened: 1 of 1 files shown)


playground/README.md
```diff
@@ -0,0 +1,66 @@
...
+├── spec-device-authenticator
+└── spec-okta-api
+```
+
+The mock server is implemented using [dyson.js](https://www.npmjs.com/package/dyson). Basically, you need to create a *mapping* in terms of what URI shall respond what response and those *mappings* are all living in `spec-*` folders.
+
+- `spec-okta-api` has mocks for Okta APIs, e.g. `/api/v1/*`, `/idp/idx/*`, etc
+- `spec-device-authenticator` is for mocking 3rd-party device authenticator application.
+
+Most of cases, the response is from a JSON file, it would be good to have abstraction to avoid tedious copy/parse. Take a look at `config/templateHelper` regardling how to handle the common use in general fashion.
```
nikhilvenkatraman-okta, line 32: few typos/corrections: 
In most cases, the response is coming from a JSON file. It would be good to have abstractions to avoid tedious copy/paste. Take a look at `config/templateHelper` to get a sense of how to handle the common use cases in a generic fashion.

playground/README.md
```diff
@@ -0,0 +1,66 @@
...
+Therefore in order to create mock for a URL which response is from a JSON file, e.g. `/idp/idx/foo`, you have to do two things:
```
nikhilvenkatraman-okta, line 33: Should we update this to say something like...

In order to create mock, i.e have a url `/idp/idx/foo`,  respond with JSON stored in `my-foo.json`, you need to do two things:

playground/README.md
```diff
@@ -0,0 +1,66 @@
...
+1. Adds URL to `idx` object in `spec-okta-api/idp/idx/index.js`
+2. Specify JSON file name in `idx` object in `config/responseConfig` like
+
+```
+'idp/idx/foo': [
+  'my-foo.json',
+],
+```
+
+Why the value is a List instead of a single String? Well, this is the poor man's solution to dynamic responses.
```
nikhilvenkatraman-okta, line 44: Why the value a list instead of a single string you ask? Well, this is a poor man's solution for providing dynamic responses. It just means that you get different a API response every time you hit the URL.

playground/README.md
```diff
@@ -0,0 +1,66 @@
...
+It means you got different API response depends on the number of times you hit the URL.
+
+```
+index = (number of times - 1) `mod` (size of mock data array)
+apiResponse = dataArray[index]
+```
+
+## Testcafe tests, playground, and mock server
+
+Testcafe tests (`/test/testcafe/spec`) are using playground application to run the tests. Does it also using mock server since mock server is started along with playground application? Apparently, it won't be ideal if it does because
```
nikhilvenkatraman-okta, line 54: Testcafe tests located under `/test/testcafe/spec` use the playground application to run the tests.
However, they do not use the mock server even though the mock server is started along with playground application.  This is not ideal for the following reasons

- each test expects different sets of API mocks.
- unless the test can change mock server config during a test, leveraging the mock server won't work.

playground/README.md
```diff
@@ -0,0 +1,66 @@
...
+- each test expects different sets of API mocks.
+- unless the test can change mock server config during testing, leverage mock server won't work.
+
+Luckily, testcafe has its own to [intercept http request and mock response](https://devexpress.github.io/testcafe/documentation/guides/advanced-guides/intercept-http-requests.html).
```
nikhilvenkatraman-okta, line 58: Luckily, testcafe has its own API to intercept http requests and mock responses (https://devexpress.github.io/testcafe/documentation/guides/advanced-guides/intercept-http-requests.html).

playground/README.md
```diff
@@ -0,0 +1,68 @@
...
+The configurations have been done in two folds
+
+- [widgetrc](https://github.com/okta/okta-signin-widget#the-widgetrc-config-file)
+- `./main.js`
+
+What `main.js` does is initialize sign-in widget using settings from `widgetrc` plus simple success handler.
+
+## Mock server
+
+While in `widgetrc` you can point `baseUrl` to any Okta tenants, it can also be set to mock server, which is started at `http://localhost:3000` when starting playground application.
```
nikhilvenkatraman-okta, line 14: While starting the playground application, the `baseUrl` in `widgetrc` can be used to point to any Okta tenant. Similarly it can also be set to point to the mock server, which runs at `http://localhost:3000`

playground/README.md
```diff
@@ -0,0 +1,68 @@
+# SignIn Widget Playground
+
+## The playground App
```
nikhilvenkatraman-okta, line 3: ## The playground App

In a nutshell, the playground application demonstrates a customer hosted sign-in widget use-case, which loads sign-in widget assets and bootstrap widget via configurations.

The configuration consists of

- [widgetrc](https://github.com/okta/okta-signin-widget#the-widgetrc-config-file)
- `./main.js`

`main.js` initializes the sign-in widget using settings from `widgetrc` along with a simple success handler.

PR #1269, opened 2020-06-30
feat: auto redirect to idp iff one idp

haishengwu-okta:
## Description:

- convert redirect-idp to success-redirect iff one redirect-idp in remediation
- also fix issue at override primary color using `colors.brand`

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

- https://okta.box.com/s/obr0srzc52rbmvsxjb1es3zr2pcgufs5

### Issue:

- https://oktainc.atlassian.net/browse/OKTA-309364


src/v2/BaseLoginRouter.js
```diff
@@ -116,6 +116,14 @@ export default Router.extend({
   },
 
   render: function (Controller, options = {}) {
+
+    // Since we have a wrapper view, render our wrapper and use its content
+    // element as our new el.
+    // Note: Render it here because we know dom is ready at this point
+    if (!this.header.rendered()) {
```
nikhilvenkatraman-okta, line 123: Is this to fix the colors.brand issue ?

haishengwu-okta: yes

src/v2/ion/responseTransformer.js
```diff
@@ -120,6 +120,26 @@ const injectIdPConfigButtonToRemediation = (settings, idxResp) => {
...
+ * widget has to automatically redirect following link from `redirect-idp` form.
+ *
+ * The idea now is to reuse `success-redirect` thus converts `redirect-idp` to `success-redirect` form.
+ */
+const convertRedirectIdPToSuccessRedirectIffOneIdp = (result) => {
+  if (Array.isArray(result.remediations)) {
+    const redirectIdpRemediations = result.remediations.filter(idp => idp.name === RemediationForms.REDIRECT_IDP);
+    if (redirectIdpRemediations.length === 1 && result.remediations.length === 1) {
+      const successRedirect = {
+        name: RemediationForms.SUCCESS_REDIRECT,
```
nikhilvenkatraman-okta, line 134: Although this seems ok to, I am wondering what the ideal approach here would be. If there is only one idp I am guessing that we are not going to even render any idp buttons, which means we are not making use of the idp object in the API. Should the API send back a success redirect form in this case with the href as the idp redirect url, because thats essentially what we are doing here on the client side.

haishengwu-okta: the feedback from backend is backend believe it's ux concern about auto redirect. 
other people could implement as display idp button regardless how many  redirect-idp remediation.

My thinking if reuse `success-redirect` doesn't quite make sense, we can coin a new remediation like `auto-redirect-idp` which only meant for widget.

nikhilvenkatraman-okta: I am not against falling back to `success-redirect`.  Re-using it seems better than creating `auto-redirect-idp`.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -14,24 +14,33 @@ const Body = BaseForm.extend({
...
+    const hasUISchemas = this.getUISchema().length > 0;
+
+    // add external idps buttons
     const idpButtons = createIdpButtons(this.options.appState.get('remediations'));
     if (Array.isArray(idpButtons) && idpButtons.length) {
       this.add(signInWithIdps, {
         selector: '.o-form-button-bar',
         options: {
           idpButtons,
+          addSeparateLine: hasUISchemas,
```
nikhilvenkatraman-okta, line 37: should we call this `hasMultipleIdps` ?

test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js
```diff
@@ -11,13 +13,23 @@ const mockWithoutIdentify = RequestMock()
...
   .respond(identifyWithIdpsNoIdentify);
 
-fixture(`Identity With IDPs Form`);
+const mockOnlyOneIdp = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(identifyOnlyOneIdp)
+  .onRequestTo('http://localhost:3000/sso/idps/facebook-idp-id-123?stateToken=inRUXNhsc6Evt7GAb8DPAA')
+  .respond(async (req, res) => {
+    // Delay response for 2 seconds
+    await new Promise((r) => setTimeout(r, 2000));
```
nikhilvenkatraman-okta, line 22: is this simulating the auto re-direct ?

haishengwu-okta: yeah but main goal is to slow down the redirect url so that can assert on SuccessRedirect view.

src/v2/view-builder/views/IdentifierView.js
haishengwu-okta: hmm.. not sure if I read correctly but the separate line is not determined by number of IDPs but whether there are ui schema for identify view.?

nikhilvenkatraman-okta: gotcha !

PR #1270, opened 2020-07-01
Handle error message nested in `options` list for SQ enrollment

haishengwu-okta:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-309274](https://oktainc.atlassian.net/browse/OKTA-309274)


src/v2/ion/IonResponseHelper.js
```diff
@@ -1,47 +1,70 @@
...
 
-/*
-returns errors
-errors = [
-  {property : fieldName1, errorSummary: [errorMessage1]},
-  {property : fieldName2, errorSummary: [errorMessage1]}
-  {property : fieldName3, errorSummary: [errorMessage1, errorMessage2]}
-]
-*/
+const convertErrorMessageToErrorSummary = (maybeFormName, remediationValues = []) => {
```
pradeepdewda-okta, line 3: why this name `maybeFormName` ?

test/unit/spec/v2/ion/IonResponseHelper_spec.js
```diff
@@ -0,0 +1,294 @@
...
+                        'message': 'bar error'
+                      }
+                    ]
+                  }
+                },
+                {
+                  label: 'Password',
+                  name: 'password',
+                  messages: {
+                    value: [
```
pradeepdewda-okta, line 137: can you verify by adding one more test that checks when a field has multiple errors. like password field can contain multiple errors.

src/v2/ion/IonResponseHelper.js
haishengwu-okta: cause this function has been used at several places but the formName is null at one case.
yeah, now I see it's confusing. The param name shall be `formName` but just it's type is `Maybe String`.

PR #1273, opened 2020-07-01
feat: password expiry support

anipendakur-okta:
## PR Checklist

- [ ] Added unit tests?

- [X] Added e2e tests

### Screenshot/Video:
image

### Reviewers:
@haishengwu-okta 

### Issue:

- [OKTA-309398](https://oktainc.atlassian.net/browse/OKTA-309398)


test/testcafe/spec/ExpiredAuthenticatorPasswordView_spec.js
```diff
@@ -0,0 +1,90 @@
...
+
+fixture(`Authenticator Expired Password`);
+
+async function setup(t) {
+  const expiredPasswordPage = new AuthenticatorExpiredPasswordPageObject(t);
+  await expiredPasswordPage.navigateToPage();
+  return expiredPasswordPage;
+}
+
+test
```
santhoshbalakrishnan, line 28: Should we add more tests to cover all policy cases?

src/v2/view-builder/views/password/ExpiredAuthenticatorPasswordView.js
```diff
@@ -0,0 +1,22 @@
+import { loc } from 'okta';
+import EnrollAuthenticatorPasswordView from './EnrollAuthenticatorPasswordView';
+
+const Body = EnrollAuthenticatorPasswordView.prototype.Body.extend({
```
santhoshbalakrishnan, line 4: Curious, why are we extending this way?

src/v2/view-builder/views/password/ExpiredAuthenticatorPasswordView.js
```diff
@@ -0,0 +1,22 @@
...
+    return loc('oie.password.expired.primaryButton', 'login');
+  },
+
+  getPasswordPolicy () { 
+    return this.options.appState.get('recoveryFactor').settings;
+  },
+
+});
+
+export default EnrollAuthenticatorPasswordView.extend({ Body });
```
santhoshbalakrishnan, line 22: Not for this PR, but can we move away from default exports? The imports would be much more defined that way.

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
```diff
@@ -11,6 +12,37 @@ const Body = BaseForm.extend({
...
+  },
+
+  displayPasswordPolicy (policy) {
+    if (policy) {
+      let listHtml = '';
+      const rulesList = getPasswordComplexityDescriptionForHtmlList( policy );
+      rulesList.forEach(rule => listHtml += `<li>${rule}</li>`);
+  
+      this.add(
+        `<section class="password-authenticator__rules">
```
santhoshbalakrishnan, line 28: Where are the CSS naming standards coming from?

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
```diff
@@ -11,6 +12,37 @@ const Body = BaseForm.extend({
...
+          <div class="password-authenticator__rules-heading">
+            ${loc('password.complexity.requirements.header', 'login')}
+          </div>
+          <ul class="password-authenticator__rules-list">${listHtml}</ul>
+        </section>`,
+        {
+          prepend: true,
+          selector: '.o-form-fieldset-container',
```
santhoshbalakrishnan, line 36: Instead of prepend, can we add in a container and then add this list to the container?

assets/sass/v2/_password.scss
```diff
@@ -0,0 +1,12 @@
+.password-authenticator {
+  &__rules-heading {
+    margin-bottom: 1rem;
+  }
+
+  &__rules-list {
+    margin-bottom: 1rem;
+    li {
```
santhoshbalakrishnan, line 8: Highly prefer not using direct html elements in CSS. Consider moving to a class name.

assets/sass/v2/_password.scss
```diff
@@ -0,0 +1,12 @@
+.password-authenticator {
+  &__rules-heading {
+    margin-bottom: 1rem;
```
santhoshbalakrishnan, line 3: Maybe not for this PR, but are we using Odyssey standards for v2? If so, can we use standard spacing units from Odyssey? If not, might be worth it to consolidate the spacing units in one file and then re-use. Eventually, we can convert from that one place to design systems when we move.

anipendakur-okta: This has nothing to do with Odyssey, it's just BEM naming convention - http://getbem.com/introduction/ , I believe Odyssey has borrowed the standard. It gives a clean way of writing css class names - helps in improving css code when used with sass/scss.

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
anipendakur-okta: I'm not sure I understand, can you give me an example pls? I don't know what value it will add though.. I'm just using what courage offers.

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
anipendakur-okta: It's BEM convention - http://getbem.com/introduction/

test/testcafe/spec/ExpiredAuthenticatorPasswordView_spec.js
anipendakur-okta: This addresses all possible policy cases for expired password as indicated by the mock. Do you know if I'm missing any? 

Also, when I address password will expire soon and admin initiated reset scenarios, the password requirements for those will be covered then

src/v2/view-builder/views/password/ExpiredAuthenticatorPasswordView.js
anipendakur-okta: Sure, are you suggesting a named export? That way we're imposing a strict constraint on import.

src/v2/view-builder/views/password/ExpiredAuthenticatorPasswordView.js
anipendakur-okta: Discussed this with @haishengwu-okta - here's what we thought - 

There are at least 3 forms - admin init password reset (TBD), password will expire soon (TBD) and password expired (this PR) that have very similar structure. `EnrollAuthenticatorPasswordView` can be extended with minimal changes to achieve these.

assets/sass/v2/_password.scss
anipendakur-okta: Also, it's being used in different places in SIW (I probably would be a major contributor to that so far :) )

src/v2/view-builder/views/password/ExpiredAuthenticatorPasswordView.js
anipendakur-okta: I think we will need to refactor it at some point in order to support tree-shaking. Some things like individual exports will also need to be done -
`
  export function abc () {};
  export function def () {};
`
vs
`
  export {
    abc,
    def
  };
`

assets/sass/v2/_password.scss
haishengwu-okta: @santhoshbalakrishnan correct me wrong, the naming convention in Odyssey is using `-` and `--` but not `__`

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
haishengwu-okta: @santhoshbalakrishnan this entire section for password requirement already inside a container `<section class="password-authenticator__rules">`

iirc, in v1, this list was added to `subtitle` container. does it work here?

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
```diff
@@ -11,6 +12,37 @@ const Body = BaseForm.extend({
...
+        }
+      );
+    }
+  },
+
+  getPasswordPolicy () {
```
haishengwu-okta, line 42: add comment to explain why `null`

src/v2/view-builder/views/password/ExpiredAuthenticatorPasswordView.js
haishengwu-okta: curious what's the drawback of `export default` here?

test/testcafe/framework/page-objects/AuthenticatorExpiredPasswordPageObject.js
```diff
@@ -0,0 +1,13 @@
+import FactorEnrollPasswordPageObject from './FactorEnrollPasswordPageObject';
+
+const requirementsSelector = '.password-authenticator__rules';
+
+export default class AuthenticatorExpiredPasswordPageObject extends FactorEnrollPasswordPageObject {
+  constructor (t) {
+    super(t);
+  }
+
+  getRequirements () {
```
haishengwu-okta, line 10: shall this method live in `FactorEnrollPasswordPageObject` ?

test/testcafe/spec/ExpiredAuthenticatorPasswordView_spec.js
haishengwu-okta: > cover all policy cases

Good point and I'm leaning toward add unit test for `v1/FactorUtil/getPasswordComplexityDescriptionForHtmlList`

src/v2/view-builder/views/password/ExpiredAuthenticatorPasswordView.js
santhoshbalakrishnan: In short: 

**Named Exports**

- Can export multiple values
-  MUST use the exported name when importing

**Default Exports**

- Export a single value
- Can use any name when importing

So this kinda enforces that you use the same name, thus allowing for more stricter defined approach. The one concern might be clarity, in that we will use the direct imported name of the method like `importedMethod()` rather than `importedModule.method()`, but with the power of IDE these days, that is hardly a problem.

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
santhoshbalakrishnan: @anipendakur-okta Here's an example:

```
template: '<div class="callout-container"></div>'
...
this.add(callout, '.callout-container');
```

This way, the callout renders within the callout container.

@haishengwu-okta Unfortunately, subtitle might not work here since, we subtitles only take in a string and not a view. So we had to explicitly add it as a child.

assets/sass/v2/_password.scss
santhoshbalakrishnan: ^ Exactly what @haishengwu-okta said. We already have a naming convention(even though not documented) of using `-` and `--` in class names. We can consolidate and use a single version.

But, I'm glad that you care and have chosen to follow some good naming conventions for CSS :)

src/v2/view-builder/views/password/ExpiredAuthenticatorPasswordView.js
santhoshbalakrishnan: That makes sense! Thanks!

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
haishengwu-okta: where will `<div class="callout-container"></div>` be added ?

the end result looks similar
either
```
<div class="callout-container">
 ...// policy list
</div>
```

or 

```
<section class="password-authenticator__rules">
... // policy list
</section>
```

And such entire container will be added to somewhere in the form.

santhoshbalakrishnan: Ah cool, misunderstood earlier. Looks good.

test/testcafe/framework/page-objects/AuthenticatorExpiredPasswordPageObject.js
anipendakur-okta: I think it makes sense for it to be in `FactorEnrollPasswordPageObject` so that other views can use it too. Does it still make sense to leave `AuthenticatorExpiredPasswordPageObject` as it without any methods in it for future? Or shall I just switch over to using `FactorEnrollPasswordPageObject` ?

haishengwu-okta: use `FactorEnrollPasswordPageObject` and maybe add comment about why it can be reused by other cases like ExpiredPassword.

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
```diff
@@ -11,6 +12,37 @@ const Body = BaseForm.extend({
...
+      let listHtml = '';
+      const rulesList = getPasswordComplexityDescriptionForHtmlList( policy );
+      rulesList.forEach(rule => listHtml += `<li>${rule}</li>`);
+  
+      this.add(
+        `<section class="password-authenticator__rules">
+          <div class="password-authenticator__heading--small">
+            ${loc('password.complexity.requirements.header', 'login')}
+          </div>
+          <ul class="password-authenticator__list">${listHtml}</ul>
```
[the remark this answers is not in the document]
anipendakur-okta: I can't use that since this is not a View with a `template`

test/testcafe/spec/ExpiredAuthenticatorPasswordView_spec.js
anipendakur-okta: Opened this - https://oktainc.atlassian.net/browse/OKTA-310415, will address separately

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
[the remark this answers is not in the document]
haishengwu-okta: Turns out it's bad practice to add just `string` which will be turned into template by jquery template. https://github.com/okta/okta-ui/blob/master/packages/courage/src/framework/View.js#L319
What about turn into a view?

```
this.add(okta.View.extend({
  tagName: 'section',
  className: ...,
  template: '...',
}));

```

test/testcafe/framework/page-objects/AuthenticatorExpiredPasswordPageObject.js
anipendakur-okta: Added comments

src/v2/view-builder/ViewFactory.js
```diff
@@ -87,7 +88,9 @@ const VIEWS_MAPPING = {
     phone: null,
     'security_question': EnrollAuthenticatorSecurityQuestion
   },
-
+  [RemediationForms.RE_ENROLL_AUTHENTICATOR]: {
+    password: ExpiredAuthenticatorPasswordView,
```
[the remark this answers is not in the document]
anipendakur-okta: Updated with a comment

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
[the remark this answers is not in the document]
anipendakur-okta: oh! ok, i'll try that.

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
anipendakur-okta: added comment

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
anipendakur-okta: Updated with new class - `password-authenticator--rules`

assets/sass/v2/_password.scss
anipendakur-okta: Updated

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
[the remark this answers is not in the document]
anipendakur-okta: Updated

assets/sass/v2/_password.scss
```diff
@@ -0,0 +1,11 @@
+.password-authenticator {
+  &--heading {
+    margin-bottom: 1rem;
```
santhoshbalakrishnan, line 3: BTW the original thread got converted to a discussion on class names. I was asking about the value of `1rem`.

> Maybe not for this PR, but are we using Odyssey standards for v2? If so, can we use standard spacing units from Odyssey? If not, might be worth it to consolidate the spacing units in one file and then re-use. Eventually, we can convert from that one place to design systems when we move.

So, the idea is that we create variables like `$base-spacing: 1rem;` and re-use that everywhere rather than having different spacing units in different places.

anipendakur-okta: Sure, I think it makes sense to address it as a overall refactor and possibly fetch odyssey constants + add more if necessary.. We will need to discuss on what path we take and work on it as a separate PR.

santhoshbalakrishnan: Yeah, that sounds good to me! Might be a good widget meeting topic as well.

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
```diff
@@ -11,6 +12,42 @@ const Body = BaseForm.extend({
...
+          selector: '.o-form-fieldset-container',
+        }
+      );
+    }
+  },
+
+  getPasswordPolicy () {
+    // This will be overridden by password expired and password will expire soon
+    // scenarios since the policies could be different for those.
+    return null;
```
santhoshbalakrishnan, line 48: Instead of this, you can throw an error.

https://stackoverflow.com/questions/29480569/does-ecmascript-6-have-a-convention-for-abstract-classes

```
throw new TypeError("Must override method");
```

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
```diff
@@ -55,4 +92,4 @@ export default BaseAuthenticatorView.extend({
       }
     });
   }
-});
+});
```
santhoshbalakrishnan, line 95: nit: newline

test/testcafe/framework/page-objects/FactorEnrollPasswordPageObject.js
```diff
@@ -2,7 +2,16 @@ import BasePageObject from './BasePageObject';
 
 const passwordFieldName = 'credentials\\.passcode';
 const confirmPasswordFieldName = 'confirmPassword';
-
+const requirementsSelector = '.password-authenticator--rules';
```
santhoshbalakrishnan, line 5: Can we use `data-se` attribute here to fetch the elements and not rely on class names?

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
anipendakur-okta: This will be changed when I work on password requirements for enrollment, I'll address it then

test/testcafe/framework/page-objects/FactorEnrollPasswordPageObject.js
anipendakur-okta: Will address it when I work on admin initiated password reset / password will expire soon.

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
haishengwu-okta: it is not strictly an abstract function which doesn't have default implementation.
here `getPasswordPolicy` has default implementation (return null) because EnrollAuthenticatorPassword doesn't support password requirement yet.

if we throw error, we have to `try/catch` where this function got called. feels complicated use case.

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
anipendakur-okta: Fixed in https://github.com/okta/okta-signin-widget/pull/1281

test/testcafe/framework/page-objects/FactorEnrollPasswordPageObject.js
anipendakur-okta: Fixed in https://github.com/okta/okta-signin-widget/pull/1281

PR #1274, opened 2020-07-01
feat: clean up terminal state for user registration

nikhilvenkatraman-okta:
Resolves: OKTA-304745

## Description:

- Add terminal state for user registration flow
- Created https://oktainc.atlassian.net/browse/OKTA-310251 as a follow up

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:

Screen Shot 2020-07-01 at 3 42 24 PM

### Issue:

- [OKTA-304745](https://oktainc.atlassian.net/browse/OKTA-304745)

### DIFF
--- a/src/v2/view-builder/internals/BaseForm.js
+++ b/src/v2/view-builder/internals/BaseForm.js
@@ -64,10 +64,12 @@ export default Form.extend({
 
   showMessages () {
     // render messages as text
-    const messagesObj = this.options.appState.get('messages');
-    if (messagesObj && messagesObj.value.length && messagesObj.value[0].class === 'INFO') {
-      const content = messagesObj.value[0].message;
-      this.add(`<span class="terminal-content">${content}</span>`, '.o-form-content');
+    const messagesObjs = this.options.appState.get('messages');
+    if (messagesObjs && messagesObjs.value.length) {
+      const content = messagesObjs.value.map((messagesObj) => {
+        return messagesObj.message;
+      });
+      this.add(`<span class="ion-messages-container">${content.join(' ')}</span>`, '.o-form-error-container');
     }
   },
 });
--- a/src/v2/view-builder/internals/BaseForm.js
+++ b/src/v2/view-builder/internals/BaseForm.js
@@ -1,4 +1,4 @@
-import { Form, loc, createCallout } from 'okta';
+import { Form, loc } from 'okta';
 import FormInputFactory from './FormInputFactory';
 
 export default Form.extend({
@@ -18,7 +18,7 @@ export default Form.extend({
     const inputOptions = uiSchemas.map(FormInputFactory.create);
 
     //should be used before adding any other input components
-    this.addCallouts();
+    this.showMessages();
 
     inputOptions.forEach(input => {
       this.addInputOrView(input);
@@ -62,14 +62,12 @@ export default Form.extend({
     }
   },
 
-  addCallouts () {
-    const warningMsgs = this.options.appState.get('messages');
-    if (warningMsgs && warningMsgs.value.length) {
-      const messageCallout = createCallout({
-        content: warningMsgs.value[0].message,
-        type: 'warning',
-      });
-      this.add(messageCallout, '.o-form-error-container');
+  showMessages () {
+    // render messages as text
+    const messagesObj = this.options.appState.get('messages');
+    if (messagesObj && messagesObj.value.length && messagesObj.value[0].class === 'INFO') {
+      const content = messagesObj.value[0].message;
+      this.add(`<span class="terminal-content">${content}</span>`, '.o-form-content');
     }
   },
 });
--- a/src/v2/view-builder/views/IdentifierView.js
+++ b/src/v2/view-builder/views/IdentifierView.js
@@ -44,9 +44,13 @@ const Body = BaseForm.extend({
     }
   },
   showMessages () {
-    // render messages with class INFO as warning callout
+    /**
+     * Renders a warning callout for unknown user flow
+     * Note: Anytime we get back `messages` object along with identify view
+     * we would render it as a warning callout
+     * */
     const messagesObj = this.options.appState.get('messages');
-    if (messagesObj && messagesObj.value.length && messagesObj.value[0].class === 'INFO'
+    if (messagesObj && messagesObj.value.length
         && this.options.appState.get('currentFormName') === 'identify') {
       const content = messagesObj.value[0].message;
       const messageCallout = createCallout({
--- a/src/v2/view-builder/views/IdentifierView.js
+++ b/src/v2/view-builder/views/IdentifierView.js
@@ -1,4 +1,4 @@
-import { loc } from 'okta';
+import { loc, createCallout } from 'okta';
 import { FORMS as RemediationForms } from '../../ion/RemediationConstants';
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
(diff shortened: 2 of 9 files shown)


src/v2/view-builder/internals/BaseForm.js
```diff
@@ -62,14 +62,12 @@ export default Form.extend({
...
-        content: warningMsgs.value[0].message,
-        type: 'warning',
-      });
-      this.add(messageCallout, '.o-form-error-container');
+  showMessages () {
+    // render messages as text
+    const messagesObj = this.options.appState.get('messages');
+    if (messagesObj && messagesObj.value.length && messagesObj.value[0].class === 'INFO') {
+      const content = messagesObj.value[0].message;
+      this.add(`<span class="terminal-content">${content}</span>`, '.o-form-content');
```
pradeepdewda-okta, line 70: probably change this to `.o-form-terminal-message-container`

pradeepdewda-okta: by the way this can still live inside `o-form-error-container` right ?

src/v2/view-builder/internals/BaseForm.js
```diff
@@ -62,14 +62,12 @@ export default Form.extend({
...
-    if (warningMsgs && warningMsgs.value.length) {
-      const messageCallout = createCallout({
-        content: warningMsgs.value[0].message,
-        type: 'warning',
-      });
-      this.add(messageCallout, '.o-form-error-container');
+  showMessages () {
+    // render messages as text
+    const messagesObj = this.options.appState.get('messages');
+    if (messagesObj && messagesObj.value.length && messagesObj.value[0].class === 'INFO') {
```
haishengwu-okta, line 68: - probably don't have to check `class` here?
- why only display first message instead of listing all given this is generic handler?

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -42,7 +42,20 @@ const Body = BaseForm.extend({
     if (!hasUISchemas) {
       this.$el.find('.button-primary').hide();
     }
-  }
+  },
+  showMessages () {
+    // render messages with class INFO as warning callout
```
haishengwu-okta, line 47: nit:
maybe elaborate a little more like
- `as warning callout for identify view`
- this logic is intended for handling `unknow-user` case at identify view but the consequence is any `messages` along with `identify` view would render as callout regardless it's only meant for `unknown-user` case or not. don't have good answer now but let's put comments here.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -42,7 +42,20 @@ const Body = BaseForm.extend({
...
+    const messagesObj = this.options.appState.get('messages');
+    if (messagesObj && messagesObj.value.length && messagesObj.value[0].class === 'INFO'
+        && this.options.appState.get('currentFormName') === 'identify') {
+      const content = messagesObj.value[0].message;
```
haishengwu-okta, line 51: ditto why only display first message?

src/v2/view-builder/internals/BaseForm.js
haishengwu-okta: the class name probably shall not have `terminal` keyword which implies it's meant for `TerminalView` but this is just a generic handler for displaying messages.
maybe `o-form-ion-messages-container` ?

nikhilvenkatraman-okta: We can also remove `o-form` from the className should we just call it `ion- messages-container`? 

I moved it out of `o-form-error-container` because messages are not just errors and it need not be within the `o-form-error-container`.

src/v2/view-builder/internals/BaseForm.js
nikhilvenkatraman-okta: I added class check to be sure we dont render error as a text, but no other reason. 
I didnt think we would have a case for multiple messages. Should we just append and generate one single string and render it ?

src/v2/view-builder/internals/BaseForm.js
haishengwu-okta: I'm guessing this container will be add before form header because of using select `.o-form-content` ?

src/v2/view-builder/internals/BaseForm.js
haishengwu-okta: Given this is generic handler, we need to account for ERROR or no `class` type.
The easiest way is probably list all messages per line (with error renders in error callout)
Concat non-error into one string is also fine.

Can create JIRA and tackle separately for `error` type.

Maybe we really don't have to care `type` now given `ERROR` is probably handled in form error handler (when none 200 response). 
So probably just 
- either list the messages
- or concat into one string

src/v2/view-builder/internals/BaseForm.js
nikhilvenkatraman-okta: Screen Shot 2020-07-01 at 9 53 08 PM

haishengwu-okta: that would work well only for terminal view or view that has no inputs (ui schema)
kind of defeat the purpose of being generic. 😅

src/v2/view-builder/internals/BaseForm.js
nikhilvenkatraman-okta: Sounds good, I can concat into one string. I also think error case wont hit here because its going to be a non 200 response which we handle elsewhere.

src/v2/view-builder/internals/BaseForm.js
nikhilvenkatraman-okta: btw @vincentvoong-okta confirmed that we want to remove the heading title from terminal view, so I will remove that in my next commit.

nikhilvenkatraman-okta: I put it back under `o-form-error-container` for now.

src/v2/view-builder/internals/BaseForm.js
```diff
@@ -62,14 +62,14 @@ export default Form.extend({
...
-        type: 'warning',
+  showMessages () {
+    // render messages as text
+    const messagesObjs = this.options.appState.get('messages');
+    if (messagesObjs && messagesObjs.value.length) {
+      const content = messagesObjs.value.map((messagesObj) => {
+        return messagesObj.message;
       });
-      this.add(messageCallout, '.o-form-error-container');
+      this.add(`<span class="ion-messages-container">${content.join(' ')}</span>`, '.o-form-error-container');
```
pradeepdewda-okta, line 72: i would not include `ion` in class name, its implementation detail.

haishengwu-okta: but `messages-container` is just too general?

pradeepdewda-okta: `terminal-messages` ?

src/v2/view-builder/views/TerminalView.js
```diff
@@ -15,20 +14,8 @@ const Body = BaseForm.extend({
...
-      {
-        'type': 'link',
-        'label': loc('backToSignin', 'login'),
-        'name': 'back',
-        'href': '/'
-      }
-    ];
-  }
-});
+// TODO add cancel link to the footer if cancel is present in the API
```
haishengwu-okta, line 17: JIRA it?

src/v2/view-builder/internals/BaseForm.js
haishengwu-okta: `terminal` would imply it's for `TerminalView` but this is generic container for any messages.

pradeepdewda-okta: any other name but definitely don't include `ion`

src/v2/view-builder/views/IdentifierView.js
pradeepdewda-okta: this was because we don't want to show all the messages if there are many.. it will look ugly. Taylor confirmed that we show just one message.

src/v2/view-builder/internals/BaseForm.js
nikhilvenkatraman-okta: yeah, I was trying to avoid `terminal` in the classname and I also felt `messages-container` is too generic. For now I think I am going to leave it as is we can update it later if we come up with something better. 

TBH I don't actually like the name `messages` that the API is sending. I would have chosen a different name for it 🤷‍♂️ and chose a corresponding classname.

src/v2/view-builder/views/TerminalView.js
nikhilvenkatraman-okta: removed this

src/v2/view-builder/views/IdentifierView.js
haishengwu-okta: sure. let's add comments regarding the decision.

PR #1282, opened 2020-07-04
feat: admin init password reset

anipendakur-okta:
## PR Checklist

- [ ] Added unit tests?

- [X] Added e2e tests

### Screenshot/Video:
image

### Reviewers:
@haishengwu-okta 

### Issue:

- [OKTA-299307](https://oktainc.atlassian.net/browse/OKTA-299307)


src/v2/view-builder/views/password/ResetAuthenticatorPasswordView.js
```diff
@@ -0,0 +1,22 @@
...
+
+  title () {
+    return loc('password.reset.title.generic', 'login');
+  },
+
+  save () {
+    return loc('password.reset', 'login');
+  },
+
+  getPasswordPolicy () {
```
haishengwu-okta, line 16: just thought and not necessarily make changes now.

Feels like the default implementation of `getPasswordPolicy` in `EnrollAuthenticatorPassword` can be

```
return this.options.appState.get('currentAuthenticator').settings;
```

as it covers most of case. one override reads `currentFactor` is probably not correct for M1 and will be addressed soon.

anipendakur-okta: Absolutely! I was going to address it in my next ticket (https://oktainc.atlassian.net/browse/OKTA-308791) and wanted to talk to you about it before I do, I will use `currentAuthenticator` in `EnrollAuthenticatorPassword`.

Another question - I am currently using `recoveryFactor` for `password expired scenario`, does it make sense to change it to `recoveryAuthenticator` or do I need to use both since `recoveryAuthenticator` is only present when `AUTHENTICATOR_PLATFORM` is enabled. Ref: https://okta.slack.com/archives/CCA77QVL2/p1593714897203000

anipendakur-okta: Although, even if it is the same, I'd rather override it here and have the same code. Because if logic to get password policy in `EnrollAuthenticatorPassword` changes, it will break `ResetAuthenticatorPassword` if not overridden, any thoughts?

haishengwu-okta: > I am currently using recoveryFactor for password expired scenario

It's probably going to be `recoveryAuthenticator`.

> Although, even if it is the same, I'd rather override it here and have the same code

I cannot tell if it's possible that logic in `EnrollAuthenticatorPassword` changed but not others. (a.k.a `EnrollAuthenticatorPassword` doesn't read `recoveryAuthenticator` but other child class.).
It feels to me unlikely happen. But I'm fine to keep the override in sub class now.

haishengwu-okta: oh I see it now. it reads `currentAuthenticator` during EnrollPassword but `recoveryAuthenticator` at other cases.

PR #1286, opened 2020-07-06
Password Requirements in Enrollment

anipendakur-okta:
## Description:
I have also fixed Password Requirements in `Password Expired` remediation to use `recoveryAuthenticator` instead of `recoveryFactor`

## PR Checklist

- [ ] Added unit tests?

- [X] Added e2e tests

### Screenshot/Video:
image

### Reviewers:
@haishengwu-okta 

### Issue:

- [OKTA-308791](https://oktainc.atlassian.net/browse/OKTA-308791)

### DIFF
--- a/src/v2/view-builder/views/password/ReEnrollAuthenticatorPasswordView.js
+++ b/src/v2/view-builder/views/password/ReEnrollAuthenticatorPasswordView.js
@@ -14,7 +14,7 @@ const Body = EnrollAuthenticatorPasswordView.prototype.Body.extend({
   },
 
   getPasswordPolicy () { 
-    return this.options.appState.get('recoveryFactor').settings;
+    return this.options.appState.get('recoveryAuthenticator').settings;
   },
 
 });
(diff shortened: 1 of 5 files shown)


src/v2/view-builder/views/password/ReEnrollAuthenticatorPasswordView.js
```diff
@@ -14,7 +14,7 @@ const Body = EnrollAuthenticatorPasswordView.prototype.Body.extend({
   },
 
   getPasswordPolicy () { 
-    return this.options.appState.get('recoveryFactor').settings;
+    return this.options.appState.get('recoveryAuthenticator').settings;
```
anipendakur-okta, line 17: Do we need to use it something like this ?
`return this.options.appState.get('recoveryAuthenticator').settings || this.options.appState.get('recoveryFactor').settings;`

Because if `AUTHENTICATOR_PLATFORM` is ON, it will be `recoveryAuthenticator` and if it is `OFF`, it will be `recoveryFactor`? Or should we just ignore factors?

Reference: https://okta.slack.com/archives/CCA77QVL2/p1593714897203000

haishengwu-okta: when `AUTHENTICATOR_PLATFORM` is OFF, it shall hint `ReEnrollFactorPasswordView` which doesn't need exists.

PR #1288, opened 2020-07-07
Add i18n override for label coming from API

haishengwu-okta:
## Description:

1. Find all Label (for normal field, or options) or Placeholder (for checkbox) base on their path. e.g.
    - `identify.identifer` <-- user name at identify page
    - `select-authenticator-enroll.authenticator.email` <-- authenticator label of email at `select-authenticator-enroll` page
    - so on and so far
2. define a mapping between the path and i18n key.
3. not all path need i18n override. e.g. `challenge-authenticator.security_question.credentials.questionKey` which is security question during verify step.
4. if we can find the override (path <-> i18n key), override label using the i18n value.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- OKTA-299837


src/v2/ion/uiSchemaLabelTransformer.js
```diff
@@ -0,0 +1,142 @@
...
+        return;
+      }
+      let i18nPathOption;
+      if (o.authenticatorType) {
+        i18nPathOption = `${i18nPath}.${o.authenticatorType}`;
+      } else if (o.value !== undefined) { // value could be string or number or undefined.
+        i18nPathOption = `${i18nPath}.${o.value}`;
+      }
+      Logger.info('\t 4: ', i18nPathOption);
+      overrideLabelIfI18NPathExists(i18nPathOption, o.label);
```
anipendakur-okta: `nit:` Do we need such a long name - `overrideLabelIfI18NPathExists`, how about something simpler - `overrideI18NLabel`. Whether if it exists or not can be left out to implementation details of the function.

anipendakur-okta: My thinking is that we can override only if it exists...

haishengwu-okta: > My thinking is that we can override only if it exists...

but the code will be slight different.

```
overrideI18NLabel = (uiSchema, key, i18nPath) => {
  if getI18nValue(i18nPath) then
    uiSchema[key] = getI18nValue(i18nPath);
}
...
overrideI18NLabel(uiSchema, 'label', i18nPath);

```

Looks like implementation is simpler.
I kind of leaning towards explicit function name at this case: `overrideLabelIfI18NKeyExists`

src/v2/ion/ion-type-handlers/object.js
```diff
@@ -0,0 +1,79 @@
...
+ * Example of the option like
+ * @param {AuthenticatorOption[]} options
+ * @param {( AuthenticatorEnrollment[] || Authenticator[] )} authenticators
+ */
+const createAuthenticatorOptions = (options = []) => {
+  return options.map(option => {
+    const value = option.value && option.value.form && option.value.form.value || [];
+
+    // Each authenticator option has list of ION field.
+    // Currently we are only support merely selecting one of options
```
pradeepdewda-okta, line 11: should be `Currently we support only selecting one option`

src/v2/ion/ion-type-handlers/object.js
```diff
@@ -0,0 +1,79 @@
...
+    // rather than pop up another page to collection extra data
```
pradeepdewda-okta, line 12: collection -> `collect extra data`

src/v2/ion/ion-type-handlers/object.js
```diff
@@ -0,0 +1,79 @@
...
+    // Thus only surface up fields that are `required: true; mutable: false`
+    // which implies it already has `value`.
+    const valueObject = value
+      .filter(v => v.required === true && v.mutable === false)
+      .reduce((init, v) => {
+        return Object.assign(init, { [v.name]: v.value });
+      }, {});
+
+    return {
+      label: option.label,
```
pradeepdewda-okta, line 25: can option.label be null ? or undefined ?

src/v2/ion/ion-type-handlers/object.js
```diff
@@ -0,0 +1,79 @@
...
+      value: valueObject,
+      authenticatorType: option.relatesTo && option.relatesTo.type,
```
pradeepdewda-okta, line 27: same here for `relatesTo` ?

src/v2/ion/ion-type-handlers/object.js
```diff
@@ -0,0 +1,79 @@
...
+};
+
+const getAuthenticatorsVerifyUiSchema = ({ options }) => {
+  return {
+    type: 'authenticatorVerifySelect',
+    options: createAuthenticatorOptions(options),
+  };
+};
+
+const createUiSchemaForObject = (ionFormField, remediationForm, transformedResp, createUISchema) => {
```
pradeepdewda-okta, line 46: some function description in comment would be more useful here..

src/v2/ion/ion-type-handlers/object.js
```diff
@@ -0,0 +1,79 @@
+/**
```
pradeepdewda-okta, line 1: why this file named object.js ?

src/v2/ion/ion-type-handlers/string.js
```diff
@@ -0,0 +1,84 @@
+/* eslint max-depth: [2, 3] */
```
pradeepdewda-okta, line 1: why name this file string.js ?

src/v2/ion/uiSchemaLabelTransformer.js
```diff
@@ -0,0 +1,143 @@
+/*!
+ * Copyright (c) 2019, Okta, Inc. and/or its affiliates. All rights reserved.
```
pradeepdewda-okta, line 2: is this comment necessary in this file ?

src/v2/ion/uiSchemaLabelTransformer.js
```diff
@@ -0,0 +1,143 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { _, loc, $ } from 'okta';
+import Bundles from 'util/Bundles';
+import Logger from 'util/Logger';
+
+const SECURITY_QUESTION_PREFIX = 'enroll-authenticator.security_question.credentials.questionKey.';
+
+const I18N_OVERRIDE_MAPPINGS = {
```
pradeepdewda-okta, line 19: should we put this mappings in a separate file and import here for more clarity?

pradeepdewda-okta, line 19: can you also put a comment that this would be used only till we have i18n keys in API itself..

src/v2/ion/uiSchemaLabelTransformer.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  if (i18nKey && !Bundles.login[i18nKey]) {
+    Logger.warn(`expect i18n key ${i18nKey} for ${i18nPath} but not found in 'login' bundle.`);
+  }
+
+  return i18nKey;
+};
+
+const overrideLabelIfI18NPathExists = (i18nPath, value) => {
```
pradeepdewda-okta, line 69: i would change the name of this function to something general.. because it returns `value` as well when i18nkey not exists that means not overriding ?

src/v2/ion/uiSchemaLabelTransformer.js
```diff
@@ -0,0 +1,143 @@
...
+  } else {
+    return value;
+  }
+};
+
+const updateLabelForUiSchema = (remediation, uiSchema) => {
+  if (uiSchema.mutable === false) {
+    return;
+  }
+  Logger.info('i18n label transformer');
```
pradeepdewda-okta, line 82: Do we need to expose all this logger messages in console ?

src/v2/ion/uiSchemaTransformer.js
```diff
@@ -231,10 +77,13 @@ const createUISchema = (transformedResp, remediationForm) => {
...
  */
 const insertUISchema = (transformedResp) => {
-
   if (transformedResp) {
     transformedResp.remediations = transformedResp.remediations.map(obj => {
       obj.uiSchema = createUISchema(transformedResp, obj);
+
+      const logs = _.pick(obj, 'value', 'uiSchema', 'name', 'relatesTo');
+      logs.value = logs.value && logs.value.filter(v => v.name !== 'stateHandle');
+      // console.log(JSON.stringify(logs, null, 2));
```
pradeepdewda-okta, line 86: remove ?

src/v2/ion/ion-type-handlers/object.js
haishengwu-okta: might be but didn't see use case atm.
what else we can do if `label` is null ?

src/v2/ion/ion-type-handlers/object.js
haishengwu-okta: when `option.relatesTo` is null, `authenticatorType` is null and what else we can do?

src/v2/ion/ion-type-handlers/object.js
haishengwu-okta: per ion type.

src/v2/ion/ion-type-handlers/string.js
haishengwu-okta: ditto.

src/v2/ion/uiSchemaLabelTransformer.js
haishengwu-okta: I think we shall add comments to all JS files.

src/v2/ion/uiSchemaLabelTransformer.js
haishengwu-okta: `I18N_OVERRIDE_MAPPINGS` is meant to be used in this file only hence feels make more sense to leave within this file for now.

src/v2/ion/uiSchemaLabelTransformer.js
haishengwu-okta: yeah, Ani made similar comment above and I'll tweak the function implementation accordingly.

src/v2/ion/ion-type-handlers/object.js
pradeepdewda-okta: to give some default value or empty string ?

src/v2/ion/uiSchemaLabelTransformer.js
haishengwu-okta: It only shows up on `dev` bundle (`yarn build:dev`) so probably fine ?

src/v2/ion/uiSchemaLabelTransformer.js
pradeepdewda-okta: Do you expect it can grow larger overtime ?

src/v2/ion/ion-type-handlers/object.js
haishengwu-okta: - empty string is same as null for courage.
- feels not necessary to have default value for now.

src/v2/ion/uiSchemaLabelTransformer.js
haishengwu-okta: - Could be and until then move it another file if more readable
- What likely to happen is another refactor when backend send i18n key and we have to move it another file then.

src/v2/ion/ion-type-handlers/string.js
pradeepdewda-okta: sounds like this file will contains functions to operate on strings or would have some string helper functions.

But what it does is, It handles ION string type to convert into meaningful form input types..

haishengwu-okta: > sounds like this file will contains functions to operate on strings or would have some string helper functions.

Could be unless you pay attention to the directory name.
I was debating myself whether shall make file name more wording even though the directory name made the hints.

pradeepdewda-okta: basically what i meant was the output is not a string but an object (uiSchema).. string.js, object.js are too general and sounds lower level implementations. Name suggest that we are overriding functioning of string or object here.. But actually it doesn't do that.

haishengwu-okta: I meant the directory name could(expect) give you hints. but anyway I'll try to come up another name.

test/testcafe/spec/Identify_spec.js
```diff
@@ -88,7 +88,7 @@ test.requestHooks(identifyMock)(`should have correct display texts`, async t =>
   await t.expect(identityPageTitle).eql('Sign In');
 
   const rememberMeText = identityPage.getRememberMeText();
-  await t.expect(rememberMeText).eql('Remember Me');
+  await t.expect(rememberMeText).eql('Remember me');
```
magizh-okta, line 91: Have u checked this with UX ?

test/unit/spec/v2/ion/uiSchemaLabelTransformer_spec.js
```diff
@@ -0,0 +1,827 @@
...
+          ]
+        }
+      ]
+    });
+  });
+
+  it('converts label for select-authenticator-authenticate', () => {
+    const resp = {
+      remediations: [
+        {
```
magizh-okta, line 138: For this the strings in factorUtil will still take the priority right ?

test/testcafe/spec/Identify_spec.js
haishengwu-okta: I think it's probably fine given it's same as v1.

test/unit/spec/v2/ion/uiSchemaLabelTransformer_spec.js
haishengwu-okta: yeah, I'm thinking to remove the label override in FactorUtil (in another PR). what do you think?

magizh-okta: Might be tricky. OV needs some special things. Lets discuss.

haishengwu-okta: yeah, once we need to support multi authenticator in select authenticator page, the override doesn't work correctly.

src/v2/ion/IonResponseHelper.js
```diff
@@ -1,3 +1,15 @@
+/*!
+ * Copyright (c) 2019, Okta, Inc. and/or its affiliates. All rights reserved.
```
nikhilvenkatraman-okta, line 2: bit: update to 2020

src/v2/ion/uiSchemaLabelTransformer.js
```diff
@@ -0,0 +1,172 @@
...
+  Logger.info('\t remediationName: ', remediation.name);
+  Logger.info('\t uiSchema: ', JSON.stringify(uiSchema));
+
+  const authenticatorType = remediation.relatesTo && remediation.relatesTo.value.type
+    ? `.${remediation.relatesTo.value.type}`
+    : '';
+  const i18nPrefix = `${remediation.name}${authenticatorType}.`;
+  const i18nPath = `${i18nPrefix}${uiSchema.name}`;
```
nikhilvenkatraman-okta, line 119: is `i18nPath` what we expect the API to eventually return ?

haishengwu-okta: no. we need to revisit this mapping once API return i18n key. see comment at very top.

src/v2/ion/uiSchemaLabelTransformer.js
```diff
@@ -0,0 +1,172 @@
...
+
+  if (uiSchema.type === 'checkbox' && uiSchema.placeholder) {
```
nikhilvenkatraman-okta, line 121: can we add some comments for these different checks ?

haishengwu-okta: I was thinking add comment but ended up with unit test which is probably bettor to elaborate use cases ?

src/v2/ion/uiSchemaLabelTransformer.js
```diff
@@ -0,0 +1,172 @@
...
+///////////////////////////////////////////////////////////////////////////////
+
+import { _, loc, $ } from 'okta';
+import Bundles from 'util/Bundles';
+import Logger from 'util/Logger';
+
+const SECURITY_QUESTION_PREFIX = 'enroll-authenticator.security_question.credentials.questionKey.';
+
+const I18N_OVERRIDE_MAPPINGS = {
+  'identify.identifier': 'primaryauth.username.placeholder',
```
nikhilvenkatraman-okta, line 39: This mapping seems like it allows us to keep our existing set of new keys that we created for M1 under the `oie` name space i.e 
```
{
i18npath1 : oie.key1,
i18npath2 : oie.key2
}
```
What is our plan to for when we want to make existing customer overrides work?  Would we need to map those keys to the same i18n path ?

src/v2/ion/uiSchemaLabelTransformer.js
```diff
@@ -0,0 +1,172 @@
...
+  'identify.credentials.passcode': 'primaryauth.password.placeholder',
```
nikhilvenkatraman-okta, line 40: why does this use the old v1 key ? vs some others that use the new oie key

src/v2/ion/uiSchemaLabelTransformer.js
haishengwu-okta: My current thinking is to just re-use v1 keys as much as possible.

src/v2/ion/uiSchemaLabelTransformer.js
haishengwu-okta: My current thinking is to just re-use v1 keys as much as possible.

src/v2/ion/uiSchemaLabelTransformer.js
nikhilvenkatraman-okta: just throwing it out there 

Maybe we can have the i18n path map to an array of keys which can have both M1 and V1 keys. This way we can try to handle the override case for v1 customers.

```
'identify.identifier': ['oie.username','primaryauth.username.placeholder']
```

haishengwu-okta: The way we can support backward compatible for V1 customer only on those label that from API.
We still lacking the ability of backward compatible for those new keys used directly in v2 views. 
Maybe need to use this mechanism anywhere need backward compatible?

PR #1290, opened 2020-07-07
refactor: surface up runtime error when adding view

haishengwu-okta:
okta-123

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

### DIFF
--- a/src/v2/controllers/FormController.js
+++ b/src/v2/controllers/FormController.js
@@ -40,11 +40,18 @@ export default Controller.extend({
       currentViewState.name,
       this.options.appState.get('authenticatorType'),
     );
-    this.formView = this.add(TheView, {
-      options: {
-        currentViewState,
-      }
-    }).last();
+    try {
+      this.formView = this.add(TheView, {
+        options: {
+          currentViewState,
+        }
+      }).last();
+    } catch (error) {
+      // This is the place where runtime error (NPE) happens at most of time.
+      // It has been swallowed by Q.js hence add try/catch to surface up errors.
+      this.options.settings.callGlobalError(error);
+      return;
+    }
 
     this.listenTo(this.formView, 'save', this.handleFormSave);
   },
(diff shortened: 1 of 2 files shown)


src/v2/controllers/FormController.js
```diff
@@ -40,11 +40,18 @@ export default Controller.extend({
...
+        }
+      }).last();
+    } catch (error) {
+      // This is the place where runtime error (NPE) happens at most of time.
+      // It has been swallowed by Q.js hence add try/catch to surface up errors.
+      this.options.settings.callGlobalError(error);
+      return;
+    }
 
     this.listenTo(this.formView, 'save', this.handleFormSave);
```
anipendakur-okta, line 56: What if we moved this to line 49 inside `try` and remove the `return` in catch block?

haishengwu-okta: that also works. and also don't really like add `return`
but feels like it's slightly better that being more explicit at which line cause trouble? :D

anipendakur-okta: Agreed, I'm good either way. Also, if we move listenTo statement inside, it's a free catch block for anything that can possibly go wrong in it :) Probably personal preference plays a big role here..

haishengwu-okta: also It's subject to change cause there may be better place to add `try/catch`

src/v2/controllers/FormController.js
```diff
@@ -40,11 +40,18 @@ export default Controller.extend({
...
+    try {
+      this.formView = this.add(TheView, {
+        options: {
+          currentViewState,
+        }
+      }).last();
+    } catch (error) {
+      // This is the place where runtime error (NPE) happens at most of time.
+      // It has been swallowed by Q.js hence add try/catch to surface up errors.
+      this.options.settings.callGlobalError(error);
```
pradeepdewda-okta, line 52: where is this fn callGlobalError defined ?

haishengwu-okta: `src/models/Settings.js`

PR #1291, opened 2020-07-08
feat: support Okta verify in select authenticator

magizh-okta:
* Added OV support for verify flow.

This is for m1. Eventually if we need to show multiple OV enrollments with additional text we need api support for that. 

    RESOLVES: OKTA-311498

Screen Shot 2020-07-08 at 9 28 42 AM

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-311498](https://oktainc.atlassian.net/browse/OKTA-311498)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -769,6 +769,8 @@ oie.verify.webauthn.uv.required.instructions = Biometric verification or a PIN i
 oie.security.question.enroll.title = Set up security question
 oie.security.question.challenge.title = Verify with your Security Question
 
+## ODA/ Okta verify
+oie.app = Okta Verify
 
 ## Select factor/authenticator enrollment
 oie.select.authenticators.enroll.title = Set up Authenticators
--- a/src/v2/view-builder/utils/FactorUtil.js
+++ b/src/v2/view-builder/utils/FactorUtil.js
@@ -49,6 +49,12 @@ const factorData = {
     description: loc('oie.webauthn.description', 'login'),
     iconClassName: 'mfa-webauthn',
   },
+
+  'app': {
+    label: loc('oie.app', 'login'),
+    description: '',
+    iconClassName: 'mfa-okta-verify',
+  },
 };
 
 const getFactorData = function (factorName) {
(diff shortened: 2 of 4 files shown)


nbories-okta:
Looks good to me. Will let @stevennguyen-okta approve UI once he's reviewed it!

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -769,6 +769,8 @@ oie.verify.webauthn.uv.required.instructions = Biometric verification or a PIN i
 oie.security.question.enroll.title = Set up security question
 oie.security.question.challenge.title = Verify with your Security Question
 
+## ODA/ Okta verify
+oie.app = Okta Verify
```
gowthamidommety-okta, line 773: No existing string we can use?

src/v2/view-builder/utils/FactorUtil.js
```diff
@@ -49,6 +49,12 @@ const factorData = {
     description: loc('oie.webauthn.description', 'login'),
     iconClassName: 'mfa-webauthn',
   },
+
+  'app': {
+    label: loc('oie.app', 'login'),
```
gowthamidommety-okta, line 54: should the label not be the displayName from https://github.com/okta/okta-signin-widget/pull/1291/files#diff-7b4f6eaf0714ce098b2871fa2b55df38R232

magizh-okta: This is good for m1. For m2 we need that. It is still undecided if the remediation label would have it or we need to get form authenticatorEnrollment.

packages/@okta/i18n/src/properties/login.properties
magizh-okta: Technically this is new for m1 since signed nonce uses it. Other names are 


factor.totpSoft.oktaVerify = Okta Verify
AND 
factor.oktaVerifyPush = Okta Verify

I guess for m2 we are going to pick the name for api anyway. U have any suggestions ? 

/cc @haishengwu-okta

gowthamidommety-okta: If this will change for m2, then i guess this is fine as is

haishengwu-okta: Leaning toward using `oie.*` for this case given `factor.*` doesn't make sense for OIE.
Also why `oie.app` instead of `oie.okta.verify` ?

magizh-okta: Checked with Wils and he was fine with this since it is a net new flow. OV (fastpass + totp + push)

PR #1294, opened 2020-07-09
feat: adds support i18n override for 'messages`

haishengwu-okta:
## Description:

- refactoring some use case to reuse v1 keys
- adds support i18n override for `messages`

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- OKTA-311407

### DIFF
--- a/src/CustomPasswordExpiredController.js
+++ b/src/CustomPasswordExpiredController.js
@@ -30,7 +30,7 @@ function (Okta, FormController, FormType, Footer) {
         var expiringSoon = this.options.appState.get('isPwdExpiringSoon'),
             numDays = this.options.appState.get('passwordExpireDays');
         if (expiringSoon && numDays > 0) {
-          return Okta.loc('oie.password.expiry.warning.title', 'login', [numDays]);
+          return Okta.loc('password.expiring.title', 'login', [numDays]);
         }
         else if (expiringSoon && numDays === 0) {
           return Okta.loc('password.expiring.today', 'login');
--- a/src/CustomPasswordExpiredController.js
+++ b/src/CustomPasswordExpiredController.js
@@ -30,7 +30,7 @@ function (Okta, FormController, FormType, Footer) {
         var expiringSoon = this.options.appState.get('isPwdExpiringSoon'),
             numDays = this.options.appState.get('passwordExpireDays');
         if (expiringSoon && numDays > 0) {
-          return Okta.loc('password.expiring.title', 'login', [numDays]);
+          return Okta.loc('oie.password.expiry.warning.title', 'login', [numDays]);
         }
         else if (expiringSoon && numDays === 0) {
           return Okta.loc('password.expiring.today', 'login');
--- a/src/EnrollActivateEmailController.js
+++ b/src/EnrollActivateEmailController.js
@@ -46,7 +46,7 @@ define([
       title: _.partial(Okta.loc, 'email.enroll.title', 'login'),
       noButtonBar: false,
       autoSave: true,
-      save: _.partial(Okta.loc, 'oform.verify', 'login'),
+      save: _.partial(Okta.loc, 'oform.verify', 'login'), // TODO: deprecated by mfa.challenge.verify
       formChildren: [
         // message
         FormType.View({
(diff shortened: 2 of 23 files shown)


src/CustomPasswordExpiredController.js
```diff
@@ -30,7 +30,7 @@ function (Okta, FormController, FormType, Footer) {
         var expiringSoon = this.options.appState.get('isPwdExpiringSoon'),
             numDays = this.options.appState.get('passwordExpireDays');
         if (expiringSoon && numDays > 0) {
-          return Okta.loc('password.expiring.title', 'login', [numDays]);
+          return Okta.loc('oie.password.expiry.warning.title', 'login', [numDays]);
```
nikhilvenkatraman-okta, line 33: Should we revert this change ? because it looks like `oie.password.expiry.warning.title` is removed from login.properties file but I still see `password.expiring.title`.

src/EnrollActivateEmailController.js
```diff
@@ -46,7 +46,7 @@ define([
       title: _.partial(Okta.loc, 'email.enroll.title', 'login'),
       noButtonBar: false,
       autoSave: true,
-      save: _.partial(Okta.loc, 'oform.verify', 'login'),
+      save: _.partial(Okta.loc, 'oform.verify', 'login'), // TODO: deprecated by mfa.challenge.verify
```
nikhilvenkatraman-okta, line 49: what is stopping us from replacing `oform.verify` with `mfa.challenge.verify` in this PR ?

src/v2/ion/uiSchemaLabelTransformer.js
```diff
@@ -158,15 +163,42 @@ const updateLabelForUiSchema = (remediation, uiSchema) => {
...
+  if (transformedResp.messages && Array.isArray(transformedResp.messages.value)) {
+    transformedResp.messages.value.forEach(message => {
+      message.message = getMessage(message);
+    });
+  }
   return transformedResp;
 };
 
-module.exports = uiSchemaLabelTransformer;
+export default uiSchemaLabelTransformer;
```
nikhilvenkatraman-okta, line 202: should we just stick to named exports here instead of mixing both default and named exports?

src/EnrollActivateEmailController.js
haishengwu-okta: In theory, if customer overrides `oform.verify` but not `mfa.challenge.verify` will be broken.
But sounds like very very rare?

src/CustomPasswordExpiredController.js
haishengwu-okta: 🤦  I meant to use `password.expiring.title`

src/v2/ion/uiSchemaLabelTransformer.js
haishengwu-okta: I have a mixing feeling. I used to like to stick with one style of exporting. But looks like when people starts to use ES6 import/export, it's likely get mixed because of the flexibility.

I establish mind set like
- `export default` creates the default behavior of this module
- `export` creates extra utilities.

Another reason is the usage other 2 `transformer` is using default import so I want to keep consistent.
`import xxxTransformer from './xxxTransformer'`
hence ended up using `export {getMessages}`

src/v2/ion/uiSchemaLabelTransformer.js
```diff
@@ -158,15 +163,42 @@ const updateLabelForUiSchema = (remediation, uiSchema) => {
...
+
+export { getMessage };
```
pradeepdewda-okta, line 204: let's put this inside a package and export the package instead.

src/v2/ion/uiSchemaLabelTransformer.js
haishengwu-okta: @nikhilvenkatraman-okta @pradeepdewda-okta how about this?
```
export { uiSchemaLabelTransformer as default, getMessage };

```

src/EnrollActivateEmailController.js
nikhilvenkatraman-okta: yeah, I think that is rare.

src/v2/ion/uiSchemaLabelTransformer.js
nikhilvenkatraman-okta: I don't mind that. Seems descriptive enough.

src/EnrollActivateEmailController.js
haishengwu-okta: I'll make change to master in separate PR.

PR #1296, opened 2020-07-10
Fixes back link for authenticate and enroll flows 

nikhilvenkatraman-okta:
## Description:

- Re-selecting an authenticator after hitting back from challenge or enroll page was broken as the new response was same as previous response

- Also fixes small CSS alignment issue

## PR Checklist

- [x] Added e2e tests

### Reviewers:

@haishengwu-okta @gowthamidommety-okta 

### Issue:

- [OKTA-312137](https://oktainc.atlassian.net/browse/OKTA-312137)

- [OKTA-311814](https://oktainc.atlassian.net/browse/OKTA-311814)


test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
```diff
@@ -72,6 +72,21 @@ test.requestHooks(mockEnrollAuthenticatorPassword)(`should navigate to password
...
+
+  selectFactorPage.selectFactorByIndex(0);
+  const enrollPasswordPage = new FactorEnrollPasswordPageObject(t);
+  await t.expect(enrollPasswordPage.passwordFieldExists()).eql(true);
+  await t.expect(enrollPasswordPage.confirmPasswordFieldExists()).eql(true);
+  await enrollPasswordPage.clickGoBackLink();
+  // re-select password
+  selectFactorPage.selectFactorByIndex(0);
+  await t.expect(enrollPasswordPage.passwordFieldExists()).eql(true);
+  await t.expect(enrollPasswordPage.confirmPasswordFieldExists()).eql(true);
```
haishengwu-okta, line 87: verify Requests loggers. In this case would expect
- one request for render authenticator list
- one request for render password challenge
- one request for render password challenge again

src/v2/models/AppState.js
```diff
@@ -103,12 +103,34 @@ export default Model.extend({
...
+       * and re-select the same authenticator for challenge. In this case also new response will be identical
+       * to the old response.
+       */
+      reRender = true;
+    }
+    return reRender;
+  },
+
+  setIonResponse (transformedResponse) {
+    if (!this.reRenderView(transformedResponse)){
```
haishengwu-okta, line 133: rename to `shallReRenderView` ?

nikhilvenkatraman-okta: how about `shouldReRenderView` ?

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -103,6 +103,20 @@ test.requestHooks(mockChallengePassword)(`should navigate to password challenge
...
 });
 
+test.requestHooks(mockChallengePassword)(`select password challenge page and hit switch authenticator and re-select password`, async t => {
+  const selectFactorPage = await setup(t);
+  await t.expect(selectFactorPage.getFormTitle()).eql('Verify it\'s you with an authenticator');
+
+  selectFactorPage.selectFactorByIndex(0);
+  const challengeFactorPage = new ChallengeFactorPageObject(t);
+  await t.expect(challengeFactorPage.getPageTitle()).eql('Sign in using your password');
+  await challengeFactorPage.clickSwitchAuthenticatorButton();
```
gowthamidommety-okta, line 113: after this, should we add a check to see if we are actually at the select factor page?

nikhilvenkatraman-okta:
@gowthamidommety-okta 

We have some login in appState to keep track of currentResponse and previousResponse which determines when we should reRender the view.

For most cases we update the previousResponse with the currentResponse. In this case since  hitting back from challenge/enroll does not trigger a call previousResponse is not updated and hence it doesn't trigger re-render.

Here is the flow 

1.  When we are on the `select-authenticator` page and we select an authenticator we make a `/challenge` or `/enroll` request depending on the flow.

2. At this point `previousResponse` is different (identify response ) compared to the new `challenge/enroll` response and hence it renders the challenge/enroll page.
 
3. Now from the challenge/enroll page we hit back to `select-authenticator` page. Note that previousResponse is not updated yet.
 
4. Now we re-select an authenticator to enroll or challenge, we make a similar  `/challenge` or `/enroll` call which comes back with a new challenge/enroll response which at this point is same as the previousResponse hence prevents the re-render.

test/testcafe/framework/page-objects/ChallengeFactorPageObject.js
```diff
@@ -27,6 +27,10 @@ export default class ChallengeFactorPageObject extends BasePageObject {
     return Selector(SWITCH_AUTHENTICATOR_SELECTOR).textContent;
   }
 
+  clickSwitchAuthenticatorButton() {
```
gowthamidommety-okta, line 30: nit: do we need to add this here and in FactorEnrollPasswordPageObject.js? It's already in BaseFormObject.js right?

assets/sass/v2/_forms.scss
```diff
@@ -8,4 +8,11 @@
     margin: 25px 0 10px;
     text-align: left;
   }
+
+  .siw-main-body {
+    .ion-messages-container {
```
haishengwu-okta, line 13: this will break the message when it renders in callout .

test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
```diff
@@ -72,6 +79,30 @@ test.requestHooks(mockEnrollAuthenticatorPassword)(`should navigate to password
...
+
+  await t.expect(requestLogger.count(() => true)).eql(3);
+  const req1 = requestLogger.requests[0].request;
+  await t.expect(req1.url).eql('http://localhost:3000/idp/idx/introspect');
+  const req2 = requestLogger.requests[1].request;
+  await t.expect(req2.url).eql('http://localhost:3000/idp/idx/credential/enroll');
+  const req3 = requestLogger.requests[2].request;
+  await t.expect(req3.url).eql('http://localhost:3000/idp/idx/credential/enroll');
```
haishengwu-okta, line 103: also request body ?

test/testcafe/framework/page-objects/ChallengeFactorPageObject.js
nikhilvenkatraman-okta: I think I can move these to BasePageObject so that I dont need to add it here and on FactorEnrollPasswordPageObject

assets/sass/v2/_forms.scss
nikhilvenkatraman-okta: why is that ?

nikhilvenkatraman-okta: callout is rendered from the view and it should not even get this `ion-messages-container` className iirc

haishengwu-okta: the callout render at IdentifyView for unknown user messages.

PR #1308, opened 2020-07-15
feat: add support for custom footer links

nikhilvenkatraman-okta:
Resolves: OKTA-312538

## Description:

 add support for custom footer links to idx widget 

## PR Checklist

- [x] Added unit tests?

### Screenshot/Video:

https://okta.box.com/s/r22ij8exarvo46lva93pfmhm9n5m24ti

### Issue:

- [OKTA-312538](https://oktainc.atlassian.net/browse/OKTA-312538)


src/v2/view-builder/utils/LinksUtil.js
```diff
@@ -30,32 +30,53 @@ const getSwitchAuthenticatorLink = (appState) => {
...
   const forgotPasswordLink = {
     'type': 'link',
     'label': loc('forgotpassword', 'login'),
     'name': 'forgot-password',
   };
 
+  const customForgotPasswordHref = settings.get('helpLinks.forgotPassword');
+
   // at identify page when only Org Authenticator Password may be available
   if (appState.getActionByPath(ORG_PASSWORD_RECOVERY_LINK)) {
```
haishengwu-okta, line 43: I think it can be
```
if custom forgot password link
  return {}
else if (ORG_PASSWORD_RECOVERY_LINK) 
  return ..
else if (ENROLLED_PASSWORD_RECOVERY_LINK)
  return ..
```

src/v2/view-builder/utils/LinksUtil.js
```diff
@@ -70,8 +91,29 @@ const goBackLink = (appState) => {
...
+const getSignOutLink  = (settings) => {
+  if (settings && settings.get('signOutLink')) {
+    return [
+      {
+        'label': loc('signout', 'login'),
+        'name': 'cancel',
+        'href': settings.get('signOutLink')
+      },
+    ];
+  } else {  return [
```
haishengwu-okta, line 103: move `return [` to next line.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -90,11 +90,22 @@ const Footer = BaseFooter.extend({
       });
     }
 
-    const forgotPasswordLink = getForgotPasswordLink(this.options.appState);
+    const forgotPasswordLink = getForgotPasswordLink(this.options.appState, this.options.settings);
+
+    const customHelpLinks = [];
+    if (this.options.settings.get('helpLinks.custom')) {
```
haishengwu-okta, line 96: nit:

```
if Array.isArray(helplinks.custom)

```

test/unit/spec/v2/view-builder/views/Identify_spec.js
```diff
@@ -0,0 +1,88 @@
+import IdentifierView from 'v2/view-builder/views/IdentifierView';
+import AppState from 'v2/models/AppState';
+import $sandbox from 'sandbox';
+import Settings from 'models/Settings';
+import IdentifyWithPasswordResponse from '../../../../../../playground/mocks/data/idp/idx/identify-with-password.json';
+
+describe('Identifier View shows custom links configured via config on load', function () {
+  beforeEach(function () {
+    this.init = (response, formName, View) => {
+      const currentViewState = {
```
haishengwu-okta, line 10: Just a bit tedious to manually create all config.
How about
- use `v2/MockUtil` to mock api response
- add another helper to create `appState` (run through transformers, and invoke `setIonResponse`. Basically mimic what BaseLoginRouter does)
- Then init `FormController`

nikhilvenkatraman-okta: I didnt really see a pattern for this, so I followed what we have in EnrollWebauthn_spec. Let me see if using `v2/MockUtil` is any easier.

haishengwu-okta: Let's see if it's easier and we can refactor `EnrollWebauthn_spec` later on. cc @magizh-okta ^

src/v2/view-builder/utils/LinksUtil.js
```diff
@@ -30,32 +30,53 @@ const getSwitchAuthenticatorLink = (appState) => {
   return [];
 };
 
-const getForgotPasswordLink = (appState) => {
+const getForgotPasswordLink = (appState, settings) => {
```
pradeepdewda-okta, line 33: I think we can go advance here and just create one function that can give us all the links, we don't want to create a new function every time we add a new link.. What do you say ? We can pass different parameters to get different links ?

src/v2/view-builder/internals/BaseFooter.js
```diff
@@ -38,16 +39,11 @@ export default View.extend({
       links = links.filter(l => $.isPlainObject(l));
     }
 
+    const settings = this.options.settings;
```
pradeepdewda-okta, line 42: if we are not using this new variable at many places then i would keep `this.options.settings` and not create a new variable

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -90,11 +90,22 @@ const Footer = BaseFooter.extend({
...
+        customHelpLink.name = 'custom';
+        customHelpLink.label = customHelpLink.text;
+        customHelpLinks.push(customHelpLink);
+      });
+    }
 
     return forgotPasswordLink
       .concat(signupLink)
-      .concat(helpLink);
+      .concat(helpLink)
```
pradeepdewda-okta, line 107: if we write a shared functions or a class, i think that can give us all the required links for the particular view and concat as well.

src/v2/view-builder/utils/LinksUtil.js
haishengwu-okta: That would be mega function. what's the benefit?

pradeepdewda-okta: I don't think it would be a mega function if you closely look at what we are doing in those functions. Isn't it ?

haishengwu-okta: I thought you meant merge 
```
getSwitchAuthenticatorLink,
getForgotPasswordLink,
goBackLink,
getSignOutLink
```
into one function that be able to generate all links base on parameter and it could be expanded over the course so I think it would be mega function.

pradeepdewda-okta: For more clarification, I intend to say that instead of exposing all these function we can expose just one. The LinksUtil can still keep these functions but those would be private. The single function that we expose can do the work of identifying required links for current view based on parameters.

haishengwu-okta: What would the extra parameter look like ?

test/unit/spec/v2/view-builder/views/Identify_spec.js
nikhilvenkatraman-okta: going to delete this test. Refactored to use e2e tests

src/v2/view-builder/utils/LinksUtil.js
nikhilvenkatraman-okta: That single function still needs to be able to accept params that are needed to generate all the different types of links, even though you might just be trying to get `ForgotPasswordLink`. I am  not sure what that extra layer of abstraction adds here TBH when that top level function is internally going to call another method.  We might need different pieces of information to generate different types of links (appState, settings etc) and I feel its easier if have individual functions to handle those.

pradeepdewda-okta: this article shows the closest what i intend to achieve.

https://davidwalsh.name/javascript-functions

haishengwu-okta: Feels a little different. the example show how to use [higher order functions](https://en.wikipedia.org/wiki/Higher-order_function) but I thought what you meant is kind of Factory method. Similar to the ViewFactory which creates View per parameters. 
```
enum LinkType = Password | SignOut | ...

function createLink (appState, settings, LinkType) {
  case LinkType of
    Password        -> getPasswordLink(...),
    Signout            -> getSignOutLink(...),
    ...
}
```

It's not necessary a bad idea but just not sure if it's worth to introduce this abstract atm. :D

test/testcafe/spec/WidgetCustomization_spec.js
```diff
@@ -0,0 +1,65 @@
...
+  const selectAuthenticatorPageObject = new SelectFactorPageObject(t);
+  await selectAuthenticatorPageObject.navigateToPage();
+  return selectAuthenticatorPageObject;
+}
+
+test.requestHooks(identifyMock)(`should show custom footer links`, async t => {
+  const identityPage = await setup(t);
+  await rerenderWidget();
+  await t.expect(identityPage.getCustomForgotPasswordLink()).eql('https://okta.okta.com/signin/forgot-password');
+  await t.expect(identityPage.getCustomHelpLink()).eql('https://google.com');
```
haishengwu-okta, line 55: also verify link label?

nikhilvenkatraman-okta: yup will add

PR #1310, opened 2020-07-16
chore: clean up terminal view implemenation

haishengwu-okta:
## Description:

- clean up mock by delete `terminal` objects
- move `terminal` remediation creation into `responseTransformer`
- update tests

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- okta-310251


src/v2/models/AppState.js
```diff
@@ -87,17 +87,18 @@ export default Model.extend({
...
-    if (!currentViewState) {
-      if (currentFormName) {
-        Logger.warn(`Cannot find view state for form ${currentFormName}. Fall back to terminal state.`);
-      }
+    // didn't expect `remediations` is empty. See `setIonResponse`.
+    const currentViewState = this.get('remediations').filter(r => r.name === currentFormName)[0];
 
-      currentViewState = this.get('terminal');
+    if (!currentViewState ) {
+      Logger.error('Panic!!');
```
pradeepdewda-okta, line 98: this!! :)

playground/mocks/config/responseConfig.js
```diff
@@ -53,12 +53,11 @@ const idx = {
     // 'identify-with-universal-link',
     // 'select-factor-authenticate',
     // 'select-factor-for-password-recovery',
-    // 'success',
+    'success',
```
nikhilvenkatraman-okta, line 56: We should just always  default here to `identify` response. Every time I pull the latest and run `yarn start` I see a different screen that loads up 😆 .

src/v2/ion/RemediationConstants.js
```diff
@@ -36,6 +36,10 @@ const FORMS = {
   DEVICE_APPLE_SSO_EXTENSION: 'device-apple-sso-extension',
   CANCEL_TRANSACTION: 'cancel-transaction',
   LAUNCH_AUTHENTICATOR: 'launch-authenticator',
+
+  // 'terminal` is not ION Form name but only coined in widget
```
nikhilvenkatraman-okta, line 40: 👍

src/v2/view-builder/views/TerminalView.js
```diff
@@ -3,11 +3,6 @@ import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
 
 const Body = BaseForm.extend({
-  title () {
-    // dont show title for terminal view
-    const msg = this.options.appState.get('terminal').message || {};
```
nikhilvenkatraman-okta, line 8: Doesn't this break LEA ? Is the assumption here that API will never send `terminal`? From what I remember LEA API will still send terminal ?

playground/mocks/config/responseConfig.js
haishengwu-okta: 😅

src/v2/view-builder/views/TerminalView.js
haishengwu-okta: > From what I remember LEA API will still send terminal.

I don't think so after James work. @pradeepdewda-okta correct me wrong.

nikhilvenkatraman-okta: ok I checked my old conversation with James and looks like we removed this from LEA as well. So we should be good!

PR #1305, opened 2020-07-14
feat: add support for user verification

jhe-okta:
## Description:
Supports loopback server, custom URI, universal link and credential SSO extension approach for user verification. 
The API format is different from device probing step. The UX mock is the same https://oktawiki.atlassian.net/wiki/spaces/UX/pages/1170385761/Okta+Verify+M1.

## PR Checklist

- [x] Have you verified the basic functionality for this change?

- [x] Added unit tests?

- [ ] Added e2e tests

- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?

### Screenshot/Video:
loopback
uv-loopback

custom URI
uv-customURI

universal link
uv-ul

credential SSO extension
Screen Shot 2020-07-14 at 2 48 24 PM

### Reviewers:
@nbories-okta @stevennguyen-okta @nikhilvenkatraman-okta @magizh-okta 

### Issue:

- [OKTA-309152](https://oktainc.atlassian.net/browse/OKTA-309152)
- [OKTA-310954](https://oktainc.atlassian.net/browse/OKTA-310954)
- [OKTA-312545](https://oktainc.atlassian.net/browse/OKTA-312545)

PR #1312, opened 2020-07-16
feat: add support for user verification

jhe-okta:
## Description:
Supports loopback server, custom URI, universal link and credential SSO extension approach for user verification.
The API format is different from device probing step. The UX mock is the same https://oktawiki.atlassian.net/wiki/spaces/UX/pages/1170385761/Okta+Verify+M1.

The PR was partially reviewed https://github.com/okta/okta-signin-widget/pull/1305

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
loopback
uv-loopback

custom URI
uv-customURI

universal link
Screen Shot 2020-07-16 at 7 26 49 AM

credential SSO extension
Screen Shot 2020-07-14 at 2 48 24 PM

### Reviewers:
@nbories-okta @stevennguyen-okta @nikhilvenkatraman-okta @magizh-okta 

### Issue:

- [OKTA-309152](https://oktainc.atlassian.net/browse/OKTA-309152)
- [OKTA-310954](https://oktainc.atlassian.net/browse/OKTA-310954)
- [OKTA-312545](https://oktainc.atlassian.net/browse/OKTA-312545)

### DIFF
--- a/playground/mocks/data/idp/idx/identify-with-user-verification-universal-link.json
+++ b/playground/mocks/data/idp/idx/identify-with-user-verification-universal-link.json
@@ -1,0 +1,170 @@
+{
+  "stateHandle": "02im-3M2f6UXHgNfS7Ns7C85EKHzGaKw0u1CC4p9_r",
+  "version": "1.0.0",
+  "expiresAt": "2020-01-13T21:14:37.000Z",
+  "intent": "LOGIN",
+  "remediation": {
+    "type": "array",
+    "value": [
+      {
+        "rel": [
+          "create-form"
+        ],
+        "relatesTo": [
+          "$.currentAuthenticatorEnrollment"
+        ],
+        "name": "challenge-authenticator", 
+        "href": "http://localhost:3000/idp/idx/authenticators/poll",  
+        "method": "POST",
+        "accepts": "application/vnd.okta.v1+json",
+        "refresh": 4000,
+        "value": [
+          {
+            "name": "stateHandle",
+            "required": true,
+            "value": "022P5Fd8jBy3b77XEdFCqnjz__5wQxksRfrAS4z6wP",
+            "visible": false,
+            "mutable": false
+          }
+        ]
+      },
+      {
+        "rel": [
+          "create-form"
+        ],
+        "name": "select-authenticator-authenticate",
+        "href": "http://localhost:3000/idp/idx/challenge",
+        "method": "POST",
+        "accepts": "application/vnd.okta.v1+json",
+        "value": [
+          {
+            "name": "authenticator",
+            "type": "object",
+            "options": [
+              {
+                "label": "Okta Verify",
+                "relatesTo": "$.authenticatorEnrollments.value[0]",
+                "form": {
+                  "value": [
+                    {
+                      "name": "id",
+                      "value": "auttheidkwh282hv8g3",
+                      "required": true,
+                      "mutable": false,
+                      "visible": false
+                    },
+                    {
+                      "name": "methodType",
+                      "value": "signed_nonce",
+                      "required": true,
+                      "mutable": false,
+                      "visible": false
+                    }
+                  ]
+                }
+              },
+              {
+                "label": "Okta Password",
+                "relatesTo": "$.authenticatorEnrollments.value[1]",
+                "form": {
+                  "value": [
+                    {
+                      "name": "id",
+                      "value": "auttmbseAWnMPtLe20g3",
+                      "required": true,
+                      "mutable": false,
+                      "visible": false
+                    },
+                    {
+                      "name": "methodType",
+                      "value": "password",
+                      "required": true,
+                      "mutable": false,
+                      "visible": false
+                    }
+                  ]
+                }
+              }
+            ]
+          },
+          {
+            "name": "stateHandle",
+            "required": true,
+            "value": "02im-3M2f6UXHgNfS7Ns7C85EKHzGaKw0u1CC4p9_r",
+            "visible": false,
+            "mutable": false
+          }
+        ]
+      }
+    ]
+  },
+   "currentAuthenticatorEnrollment": {
+      "type": "object",
+      "value": {
+          "profile": {
+            "name": "Okta Verify"
+          },
+          "type": "app", 
+          "id": "aen1mz5J4cuNoaR3l0g4",
+          "cancel": {
+              "rel": [
+                  "create-form"
+              ],
+              "name": "cancel-polling",
+              "href": "http://localhost:3000/idp/idx/authenticators/poll/cancel",
+              "method": "POST",
+              "accepts": "application/vnd.okta.v1+json",
+              "value": [
+                  {
+                      "name": "stateHandle",
+                      "required": true,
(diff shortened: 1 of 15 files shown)


jhe-okta:
@santoshmale-okta simply changing the remediation name does not seem to be working right away. It seems like some other value is missing or the format need to be adjusted in order to distinguish with other types.

test/testcafe/spec/DeviceChallengePollViewUserVerification_spec.js
```diff
@@ -0,0 +1,172 @@
...
+    await t.expect(loopbackSuccessLogger.count(
+      record => record.response.statusCode === 200 &&
+      record.request.url.match(/introspect|6512/)
+    )).eql(3);
+    await t.expect(loopbackSuccessLogger.count(
+      record => record.response.statusCode === 200 &&
+      record.request.url.match(/challenge/) &&
+      record.request.body.match(/challengeRequest":"eyJraWQiOiJW/)
+    )).eql(1);
+    probeSuccess = true;
```
magizh-okta, line 102: do this in 108 as it makes more sense.

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -48,36 +54,54 @@ const Body = BaseForm.extend(Object.assign(
...
-        this.add(createButton({
-          className: 'ul-button button button-wide button-primary',
-          title: loc('universalLink.button', 'login'),
-          click () {
-            // only window.location.href can open universal link in iOS/MacOS
-            // other methods won't do, ex, AJAX get or form get (Util.redirectWithFormGet)
-            Util.redirect(deviceChallenge.href);
-          }
-        }));
+        if (this.currentRemediationName === RemediationForms.USER_VERIFICATION_CHALLENGE_POLL) {
```
haishengwu-okta, line 73: In general, I dont remember share view between different ion form if you have to compare form in the view.
so why not create two views instead ?

test/testcafe/spec/DeviceSSOExtensionView_spec.js
```diff
@@ -68,6 +75,19 @@ test
   await t.expect(identityPage.getIdentifierValue()).eql('Test Identifier');
 });
 
+test
+.requestHooks(logger, uvCredentialSSOExtensionMock)(`with credential SSO Extension approach during user verification, opens the verify URL`, async t => {
+  const ssoExtensionPage = new BasePageObject(t);
```
haishengwu-okta, line 80: this test looks like only verify the identify page without more context.
would you expect something else happen? like something in between widget receive `identifyUserVerificationWithCredentialSSOExtension` and render `identify` ?

jhe-okta:
> The Screen with the custom URI or Universal links should have the Okta Verify icon in the beacon. We shouldn't be showing the blank user icon there. Also I believe the beacon is being removed from the initial enter ur identifier screens etc.. (We might need to double check with @nikhilvenkatraman-okta regarding that) So we shouldn't have a beacon on the signing in using Okta Fastpass, or the you are being redirected sceen. Maybe this hasn't been worked on yet. I'd like to double check for this though.

@nbories-okta most the idx SIW screens still have the beacon. Yes, I can replace the user icon with the Okta Verify icon. Do you only want to replace in the user verification step or device probe step as well?

jhe-okta:
> as per our slack conversation. For the challenge/verify authenticator screens (in this case all OV prompts) should have the OV logo in the beacon.
> 
> cc: @stevennguyen-okta

@nbories-okta @stevennguyen-okta I've updated all the challenge/verify UIs for device probe & user verification steps to use okta verify beacon.

Screen Shot 2020-07-22 at 3 51 58 PM
Screen Shot 2020-07-23 at 7 30 30 AM
Screen Shot 2020-07-23 at 7 43 42 AM

test/testcafe/spec/DeviceSSOExtensionView_spec.js
jhe-okta: It simply need to verify if entering the UI launches the sso extension link right away and if it results in the right page as it is mocked. So there's nothing in between we need to verify.

assets/sass/v2/_identify.scss
```diff
@@ -1,3 +1,7 @@
+.sign-in-with-device-option.okta-verify-container.button:focus {
+  border-color: #b3b3b3;
```
haishengwu-okta, line 2: any css variable to use ? or why customize border color

playground/mocks/data/idp/idx/identify-with-user-verification-universal-link.json
```diff
@@ -0,0 +1,170 @@
+{
```
haishengwu-okta, line 1: shall we rename to `challenge-authenticator-okta-verify-xxx.json` for those mocks that has `challenge-authenticator` as first remediation?

src/v2/view-builder/ViewFactory.js
```diff
@@ -120,6 +121,7 @@ const VIEWS_MAPPING = {
     'security_key': ChallengeWebauthnView,
     'security_question': ChallengeAuthenticatorSecurityQuestion,
     phone: ChallengeAuthenticatorPhoneView,
+    app: UserVerificationDeviceChallengePollView,
```
haishengwu-okta, line 124: - how about rename to `ChallengeAuthenticatorOktaVerifyView` ?
- curious the diff between `DeviceChallengePollView` and `UserVerificationDeviceChallengePollView`.  the implementation looks quite similar. what are use cases or we'll deprecate former?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -172,6 +175,11 @@ const Footer = BaseFooter.extend({
 });
 
 export default BaseView.extend({
+  Header: BaseHeader.extend({
```
haishengwu-okta, line 178: Can you use `BaseAuthenticatorView.js` which already defines beacon header.

src/v2/view-builder/views/UserVerificationDeviceChallengePollView.js
```diff
@@ -0,0 +1,159 @@
...
+      };
+
+      const doProbing = () => {
+        return checkPort()
+          // TODO: can we use standard ES6 promise methods, then/catch?
+          .done(onPortFound)
+          .fail(onFailure);
+      };
+
+      let probeChain = Promise.resolve();
```
haishengwu-okta, line 126: Since we're using jquery.ajax to make request, would you consider to switch to `jquery.Deferred`? just to be consistent.

test/testcafe/spec/DeviceSSOExtensionView_spec.js
haishengwu-okta: there is no evidence in the test that the UI `launches the sso extension link`. are there any request log we can inspect?
the test code basically says `navigate to a page, which will make call to introspect and render identify screen` which is identical to what `identify` flow does.

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: The API response for this view is not for a factor, using BaseAuthenticatorView won't generate the right class for beacon.

PR #1319, opened 2020-07-17
fix: fix the password expiry warning page title

haishengwu-okta:
- handle variants of expireToDays

OKTA-313618

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

### DIFF
--- a/src/models/Settings.js
+++ b/src/models/Settings.js
@@ -36,7 +36,6 @@ define([
 
   return Okta.Model.extend({
 
-    flat: true,
     authClient: undefined,
 
     local: {
@@ -104,12 +103,15 @@ define([
       },
 
       // OAUTH2
+      'clientId': 'string',
+      'redirectUri': 'string',
+      'oAuthTimeout': ['number', false],
+
       'authScheme': ['string', false, 'OAUTH2'],
       'authParams.display': {
         type: 'string',
         values: ['none', 'popup', 'page']
       },
-
       // Note: It shouldn't be necessary to override/pass in this property -
       // it will be set correctly depending on what the value of display is
       // and whether we are using Okta or a social IDP.
@@ -117,28 +119,23 @@ define([
         type: 'string',
         values: ['query', 'fragment', 'form_post', 'okta_post_message']
       },
-
       // Can either be a string or an array, i.e.
       // - Single value: 'id_token', 'token', or 'code'
       // - Multiple values: ['id_token', 'token']
       'authParams.responseType': ['any', false, 'id_token'],
       'authParams.scopes': ['array', false],
-
       'authParams.issuer': ['string', false],
       'authParams.authorizeUrl': ['string', false],
       'authParams.state': ['string', false],
       'authParams.nonce': ['string', false],
 
-      'policyId': 'string',
-      'clientId': 'string',
-      'redirectUri': 'string',
+      // External IdPs
       'idps': ['array', false, []],
       'idpDisplay': {
         type: 'string',
         values: ['PRIMARY', 'SECONDARY'],
         value: 'SECONDARY'
       },
-      'oAuthTimeout': ['number', false],
 
       // HELP LINKS
       'helpLinks.help': 'string',
@@ -150,6 +147,7 @@ define([
       'customButtons': ['array', false, []],
 
       //Registration
+      'policyId': 'string',
       'registration.click': 'function',
       'registration.parseSchema': 'function',
       'registration.preSubmit': 'function',
--- a/src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
+++ b/src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
@@ -15,7 +15,7 @@ const Body = BaseForm.extend({
 
   initialize () {
     BaseForm.prototype.initialize.apply(this, arguments);
-    const policy = this.getPasswordPolicy();
+    const policy = this.getPasswordPolicySettings();
     this.displayPasswordPolicy(policy);
   },
 
@@ -45,7 +45,7 @@ const Body = BaseForm.extend({
     }
   },
 
-  getPasswordPolicy () {
+  getPasswordPolicySettings () {
     // This will be overridden by following scenario since the policies could be different for those.
     // - password reset (`ReEnrollAuthenticatorPasswordView.js`)
     //
--- a/src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
+++ b/src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
@@ -46,9 +46,11 @@ const Body = BaseForm.extend({
   },
 
   getPasswordPolicy () {
-    // This will be overridden by password expired and password will expire soon
-    // scenarios since the policies could be different for those.
-    return this.options.appState.get('currentAuthenticator').settings;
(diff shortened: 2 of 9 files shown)


src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
```diff
@@ -5,17 +5,21 @@ import EnrollAuthenticatorPasswordView from './EnrollAuthenticatorPasswordView';
...
+    } else {
+      return loc('password.expiring.soon', 'login');
+    }
   },
 
   save () {
     return loc('password.expired.submit', 'login');
   },
-
-  getPasswordPolicy () {
```
anipendakur-okta, line 16: The reason why I left it there is because if in `EnrollAuthenticatorPaswordView` the way we retrieve password policy changes, it will break here..

src/v2/view-builder/views/password/ResetAuthenticatorPasswordView.js
```diff
@@ -13,10 +13,6 @@ const Body = EnrollAuthenticatorPasswordView.prototype.Body.extend({
     return loc('password.reset', 'login');
   },
 
-  getPasswordPolicy () {
```
anipendakur-okta, line 16: The reason why I left it there is because if in `EnrollAuthenticatorPaswordView` the way we retrieve password policy changes, it will break here..

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
```diff
@@ -46,9 +46,11 @@ const Body = BaseForm.extend({
   },
 
   getPasswordPolicy () {
-    // This will be overridden by password expired and password will expire soon
-    // scenarios since the policies could be different for those.
-    return this.options.appState.get('currentAuthenticator').settings;
+    // This will be overridden by following scenario since the policies could be different for those.
+    // - password reset (`ReEnrollAuthenticatorPasswordView.js`)
+    //
+    const relatesToObject = this.options.currentViewState.relatesTo;
```
anipendakur-okta, line 52: `nit:` what do you think about this?
```
const { value: passwordPolicy } = this.options.currentViewState.relatesTo;
return passwordPolicy && passwordPolicy.settings;
```
Improves readability.

src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
haishengwu-okta: I hope they'll stay in sync in terms of how to retrieve policy for Enroll/Reset password.
If changes happen to Enroll but not reset, let it break which remind us to chat with backend to see why the inconsistency. Otherwise it's a bit confusing that seems like we're anticipating there will be inconsistency happening.?

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
```diff
@@ -87,11 +89,13 @@ export default BaseAuthenticatorView.extend({
       local,
       validate () {
         if (this.get('credentials.passcode') !== this.get('confirmPassword') &&
-          this.get('credential.value') !== this.get('confirmPassword')) {
+            this.get('credential.value') !== this.get('confirmPassword')) {
```
anipendakur-okta, line 92: `nit:` how about this? This just avoids an extra errors var:
```
return (this.get('credentials.passcode') !== this.get('confirmPassword')
  && this.get('credential.value') !== this.get('confirmPassword')
  ? { 'confirmPassword': loc('password.error.match', 'login') }
  : null );
```

anipendakur-okta: or we can just avoid the `else` and `return null;` after if conditional.

src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
```diff
@@ -5,17 +5,21 @@ import EnrollAuthenticatorPasswordView from './EnrollAuthenticatorPasswordView';
 const Body = EnrollAuthenticatorPasswordView.prototype.Body.extend({
   className: 'password-authenticator',
   title () {
-    const daysToExpiry = this.options.appState.get('currentAuthenticator').settings.daysToExpiry;
-    return `${loc('password.expiring.title', 'login', [ daysToExpiry ])}`;
+    const passwordPolicy = this.getPasswordPolicy() || {};
```
anipendakur-okta, line 8: on a second thought it makes sense to rename the function to `getPasswordPolicySettings`

src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
```diff
@@ -5,17 +5,21 @@ import EnrollAuthenticatorPasswordView from './EnrollAuthenticatorPasswordView';
...
+    const daysToExpiry = passwordPolicy.daysToExpiry;
+
+    if (daysToExpiry > 0) {
```
anipendakur-okta, line 11: `nit:` This might be a personal preference, I'd rather have two if conditionals with return inside followed by a return outside instead of a if-else ladder.

anipendakur-okta:
Just a bunch of nits , looks good to ship !

src/models/Settings.js
```diff
@@ -104,41 +103,39 @@ define([
       },
 
       // OAUTH2
+      'clientId': 'string',
```
anipendakur-okta, line 106: we can discuss online, but why are we changing this file?

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
haishengwu-okta: doesn't seem improving readability to me though. 
I read as: `relatesTo has a property 'value' and rename it to passwordPolicy. return 'passwordPolicy.settings' as long as 'passwordPolicy' not null`

v.s other way (I guess I get used this already)

`return this.options.currentViewState.relatesTo.value.settings` safely (aka as long as it exists.)

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
haishengwu-okta: > or we can just avoid the else and return null; after if conditional.

that works. I'm not consistent at this style but usually I like explicit `if/else`

src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
haishengwu-okta: yeah, I usually prefer `if/elseif` unless it getting more complicated. 3 if-else clauses seems ok.

src/models/Settings.js
haishengwu-okta: I just move around properties definitions to sections they suppose to be. feels a bit easier to read.

src/models/Settings.js
```diff
@@ -36,7 +36,6 @@ define([
 
   return Okta.Model.extend({
 
-    flat: true,
```
haishengwu-okta, line 39: default value of `flat` is true in `Okta.Model`

PR #1323, opened 2020-07-20
feat: adds css classnames to body for parity with v1

nikhilvenkatraman-okta:
Resolves: OKTA-312533

## Description:

- adds css classnames to body for parity with v1

- Class names to be added for parity with V1
https://github.com/okta/okta-signin-widget/blob/master/assets/sass/okta-theme.scss#L910

``` 
    .primary-auth

    .password-expired
    .password-reset
    .password-reset-email-sent
    .forgot-password

    .recovery-challenge
    .recovery-loading
    .recovery-question

    .enroll-choices

    .enroll-sms
    .enroll-call
    .enroll-question

    .mfa-verify
    .mfa-verify-question
    .mfa-verify-passcode  used for email, call, sms in V1
    .mfa-verify-password

```

- adds css classnames for M1 based on form-name + authenticator 

## PR Checklist

### Issue:

- [OKTA-312533](https://oktainc.atlassian.net/browse/OKTA-312533)

### DIFF
--- a/src/v2/ion/ViewClassNamesFactory.js
+++ b/src/v2/ion/ViewClassNamesFactory.js
@@ -1,0 +1,83 @@
+/* eslint complexity: [2, 25] */
+import {FORMS} from './RemediationConstants';
+
+const FORMNAME_CLASSNAME_MAPPINGS = {
+  [FORMS.IDENTIFY]: {
+    [FORMS.IDENTIFY]: ['primary-auth'],
+    'password': ['primary-auth']
+  },
+  [FORMS.ENROLL_PROFILE]: {
+    [FORMS.ENROLL_PROFILE]: ['registration'],
+  },
+  [FORMS.CHALLENGE_AUTHENTICATOR]: {
+    email: ['mfa-verify-passcode'],
+    password: ['mfa-verify-password'],
+    sms: ['mfa-verify-passcode'],
+    voice: ['mfa-verify-passcode'],
+    'security_question': ['mfa-verify-question']
+  },
+  [FORMS.ENROLL_AUTHENTICATOR]: {
+    email: ['enroll-email'],
+    password: ['enroll-password'],
+    sms: ['enroll-sms'],
+    voice: ['enroll-call'],
+    'security_question': ['enroll-question'],
+  },
+
+  [FORMS.SELECT_AUTHENTICATOR_ENROLL]: {
+    'select-authenticator-enroll': ['enroll-choices']
+  },
+  [FORMS.SELECT_AUTHENTICATOR_AUTHENTICATE]: {
+    'password': ['forgot-password']
+  },
+  [FORMS.REENROLL_AUTHENTICATOR]: {
+    'password':['password-expired']
+  },
+
+  [FORMS.RESET_AUTHENTICATOR]: {
+    'password':['forgot-password']
+  },
+};
+
+const getV1ClassName = (formName, type, isPasswordRecoveryFlow) => {
+  if (isPasswordRecoveryFlow && formName === 'identify') {
+    // if password reset flow from identifier page with recoveryAuthenticator add forgot-password class
+    return ['forgot-password'];
+  } else {
+    return FORMNAME_CLASSNAME_MAPPINGS[formName][type];
+  }
+};
+
+const getClassNameMapping = (formName, authenticatorType, methodType) => {
+  let type = formName;
+  let v2ClassName = formName;
+  let classNames = [];
+
+  if (authenticatorType === 'phone') {
+    /**
+      Both sms and call have same type phone
+      currentAuthenticatorEnrollment is during verify and currentAuthenticator during enroll flows
+    **/
+    type = `${methodType}`;
+  }
+  else if (authenticatorType) {
+    type = `${authenticatorType}`;
+  }
+
+  if (type !== formName) {
+    // If we have a type which is authenticatorType/methodType use that to generate a V2 className
+    v2ClassName = formName + '--' + type;
+  }
+  classNames.push(v2ClassName);
+  //do a lookup for any V1 classNames and concat
+  if (FORMNAME_CLASSNAME_MAPPINGS[formName] && FORMNAME_CLASSNAME_MAPPINGS[formName][type]) {
+    classNames = classNames.concat(getV1ClassName(formName, type));
+  }
+  return classNames;
+};
+
+export {
+  getClassNameMapping,
+  getV1ClassName
+};
+
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -4,6 +4,7 @@ const idx = {
   // ===== IDX
 
   '/idp/idx/introspect': [
+    'identify',
     // 'error-internal-server-error',
     // 'authenticator-enroll-password',
     // 'authenticator-enroll-phone',
@@ -12,14 +13,14 @@ const idx = {
     // 'authenticator-enroll-data-phone-voice',
     // 'error-internal-server-error',
     // 'authenticator-enroll-security-question',
-    'authenticator-enroll-select-authenticator',
+    // 'authenticator-enroll-select-authenticator',
     // 'authenticator-enroll-select-authenticator-with-skip',
     // 'authenticator-enroll-webauthn',
     // 'authenticator-verification-data-phone-sms-then-voice',
     // 'authenticator-verification-data-phone-voice-only',
     // 'authenticator-verification-data-phone-voice-then-sms',
     // 'authenticator-verification-email',
-    'authenticator-verification-password',
+    // 'authenticator-verification-password',
     // 'authenticator-verification-phone-sms',
     // 'authenticator-verification-phone-voice',
     // 'authenticator-verification-security-question',
--- a/src/v2/models/AppState.js
+++ b/src/v2/models/AppState.js
@@ -52,6 +52,14 @@ export default Model.extend({
           || '';
       },
     },
+    authenticatorMethodType: {
+      deps: ['currentAuthenticator', 'currentAuthenticatorEnrollment',],
+      fn (currentAuthenticator = {}, currentAuthenticatorEnrollment = {}) {
+        return currentAuthenticator.methods && currentAuthenticator.methods[0].type
+          || currentAuthenticatorEnrollment.methods && currentAuthenticatorEnrollment.methods[0].type
+          || '';
+      },
+    },
     showSignoutLink: {
       deps: ['idx', 'currentFormName'],
       fn: function (idx = {}, currentFormName) {
@@ -60,6 +68,12 @@ export default Model.extend({
           && !FORMS_WITHOUT_SIGNOUT.includes(currentFormName);
       },
     },
+    isPasswordRecovery: {
+      deps: ['recoveryAuthenticator'],
+      fn: function (recoveryAuthenticator = {}) {
+        return recoveryAuthenticator && recoveryAuthenticator.type === 'password';
+      }
+    }
   },
 
   hasRemediationObject (formName) {
@@ -148,6 +162,8 @@ export default Model.extend({
 
     // clear appState before setting new values
     this.clear({silent: true});
+    // clear cache for derived props.
+    this.trigger('cache:clear');
 
     // set new app state properties
     this.set(transformedResponse);
--- a/src/v2/view-builder/internals/BaseView.js
+++ b/src/v2/view-builder/internals/BaseView.js
@@ -4,6 +4,7 @@ import BaseModel from './BaseModel';
 import BaseHeader from './BaseHeader';
 import BaseFooter from './BaseFooter';
 import hbs from 'handlebars-inline-precompile';
+import {getClassNameMapping} from '../../ion/ViewClassNamesFactory';
 
 export default View.extend({
 
@@ -13,7 +14,16 @@ export default View.extend({
 
   Footer: BaseFooter,
 
-  className: 'siw-main-view',
+  className () {
+    let classNames = ['siw-main-view'];
+    const appState = this.options.appState;
+
+    const formName = appState.get('currentFormName');
+    const authenticatorType = appState.get('authenticatorType');
+    const methodType = appState.get('authenticatorMethodType');
+    classNames = classNames.concat(getClassNameMapping(formName, authenticatorType, methodType));
+    return classNames.join(' ');
+  },
 
   template: hbs`
     <div class="siw-main-header"></div>
--- a/src/v2/view-builder/views/SelectAuthenticatorVerifyView.js
+++ b/src/v2/view-builder/views/SelectAuthenticatorVerifyView.js
@@ -16,8 +16,7 @@ const Body = BaseForm.extend({
     return loc('oie.select.authenticators.verify.subtitle', 'login');
   },
   isPasswordRecoveryFlow () {
-    const recoveryAuthenticator = this.options.appState.get('recoveryAuthenticator');
-    return recoveryAuthenticator && recoveryAuthenticator.type === 'password';
+    return this.options.appState.get('isPasswordRecovery');
   },
   noButtonBar: true,
 });
--- a/test/unit/spec/v2/ion/ViewClassNamesFactory_spec.js
+++ b/test/unit/spec/v2/ion/ViewClassNamesFactory_spec.js
@@ -1,0 +1,117 @@
+import { getClassNameMapping, getV1ClassName } from 'v2/ion/ViewClassNamesFactory';
+
+describe('Adds the right classname', function () {
+
+  it('for identify form with password', function () {
+    const result = getClassNameMapping('identify', 'password', null);
+    expect(result).toEqual(['identify--password', 'primary-auth']);
+  });
+
+  it('for identify form', function () {
+    const result = getClassNameMapping('identify', null , null);
+    expect(result).toEqual(['identify', 'primary-auth']);
+  });
+
+  it('for select-authenticator-authenticate form', function () {
+    const result = getClassNameMapping('select-authenticator-authenticate', null , null);
+    expect(result).toEqual(['select-authenticator-authenticate']);
+  });
+  it('for select-authenticator-authenticate form with password ', function () {
+    const result = getClassNameMapping('select-authenticator-authenticate', 'password' , null);
+    expect(result).toEqual(['select-authenticator-authenticate--password', 'forgot-password']);
+  });
+
+  it('for select-authenticator-enroll form', function () {
+    const result = getClassNameMapping('select-authenticator-enroll', null , null);
+    expect(result).toEqual(['select-authenticator-enroll', 'enroll-choices']);
+  });
+
+  it('for challenge-authenticator form with email ', function () {
+    const result = getClassNameMapping('challenge-authenticator', 'email' , null);
+    expect(result).toEqual(['challenge-authenticator--email', 'mfa-verify-passcode']);
+  });
+
+  it('for challenge-authenticator form with security_question ', function () {
+    const result = getClassNameMapping('challenge-authenticator', 'security_question' , null);
+    expect(result).toEqual(['challenge-authenticator--security_question', 'mfa-verify-question']);
+  });
+
+  it('for enroll-authenticator form with security_question ', function () {
+    const result = getClassNameMapping('enroll-authenticator', 'security_question' , null);
+    expect(result).toEqual(['enroll-authenticator--security_question', 'enroll-question']);
+  });
+
+  it('for enroll-authenticator form with email ', function () {
+    const result = getClassNameMapping('enroll-authenticator', 'email' , null);
+    expect(result).toEqual(['enroll-authenticator--email', 'enroll-email']);
+  });
+
+  it('for enroll-authenticator form with password ', function () {
+    const result = getClassNameMapping('enroll-authenticator', 'password' , null);
+    expect(result).toEqual(['enroll-authenticator--password', 'enroll-password']);
+  });
+
+  it('for enroll-authenticator form with sms ', function () {
+    const result = getClassNameMapping('enroll-authenticator', 'sms' , null);
+    expect(result).toEqual(['enroll-authenticator--sms', 'enroll-sms']);
+  });
+
+  it('for enroll-authenticator form with voice ', function () {
+    const result = getClassNameMapping('enroll-authenticator', 'voice' , null);
+    expect(result).toEqual(['enroll-authenticator--voice', 'enroll-call']);
+  });
+
+
+  it('for reenroll-authenticator form with voice ', function () {
+    const result = getClassNameMapping('reenroll-authenticator', 'password' , null);
+    expect(result).toEqual(['reenroll-authenticator--password', 'password-expired']);
+  });
+
+  it('for reset-authenticator form with voice ', function () {
+    const result = getClassNameMapping('reset-authenticator', 'password' , null);
+    expect(result).toEqual(['reset-authenticator--password', 'forgot-password']);
+  });
+
+  it('for success form', function () {
+    const result = getClassNameMapping('success-redirect', null , null);
+    expect(result).toEqual(['success-redirect']);
+  });
+
+  it('for terminal form ', function () {
+    const result = getClassNameMapping('terminal', null , null);
+    expect(result).toEqual(['terminal']);
+  });
+
+  it('for authenticator-verification-data form wih email method sms', function () {
+    const result = getClassNameMapping('authenticator-verification-data', 'phone' , 'sms');
+    expect(result).toEqual(['authenticator-verification-data--sms']);
+  });
+
+  it('for authenticator-verification-data form wih email method voice', function () {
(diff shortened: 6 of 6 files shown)


src/v2/ion/RemediationConstants.js
```diff
@@ -42,6 +42,68 @@ const FORMS = {
...
+
+const getClassNameMapping = (appState) => {
+  const formName = appState.get('currentFormName');
+  let authenticatorType = appState.get('authenticatorType');
+  const recoveryAuthenticator = appState.get('recoveryAuthenticator');
+  let isPasswordRecoveryFlow = recoveryAuthenticator && recoveryAuthenticator.type === 'password';
+
+  if (authenticatorType === 'phone') {
+    // both sms and call have same type
+    authenticatorType = appState.get('currentAuthenticatorEnrollment') &&
```
haishengwu-okta, line 92: is this methodType? would prefer more explicitly like

```
if authenticator type is phone then
  key += `-${methodType}`
else if authenticatorType exists
  key += `-${authenticatorType}`

File: src/v2/ion/RemediationConstants.js [hunk 2]
```diff
@@ -42,6 +42,68 @@ const FORMS = {
...
 
+const FORMNAME_CLASSNAME_MAPPING = {
+  'identify': ['identify', 'primary-auth'],
+  'identify-password': ['identify', 'primary-auth'],
+  'select-identify': ['select-identify'],
+  'enroll-profile': ['enroll-profile', 'registration'],
+
+  // 1st step of the challenge flow for sms/call
+  'authenticator-verification-data-sms': ['authenticator-verification-data-sms'],
+  'authenticator-verification-data-phone': ['authenticator-verification-data-phone'],
```
haishengwu-okta, line 53: feels the redundancy is not necessary since it can be created on the flay. (the `key` in the `getClassNameMapping` function)

src/v2/ion/RemediationConstants.js
```diff
@@ -42,6 +42,68 @@ const FORMS = {
   TERMINAL: 'terminal',
 };
 
+const FORMNAME_CLASSNAME_MAPPING = {
```
haishengwu-okta, line 45: Move to a dedicated file like ViewFactory. 
(maybe call it ViewClassNamesFactory ?)

src/v2/ion/RemediationConstants.js
```diff
@@ -42,6 +42,68 @@ const FORMS = {
...
+  'identify': ['identify', 'primary-auth'],
```
haishengwu-okta, line 46: nit:
prefer to use same structure of the view factory like 
```
{
  [form-name]: {
   [default]: [],
   [phone]: ...
   ...
  }
}
```

src/v2/ion/RemediationConstants.js
```diff
@@ -42,6 +42,68 @@ const FORMS = {
...
+  'redirect-idp': ['redirect-idp'],
+
+  'cancel-transaction': ['cancel-transaction'],
+};
+
+const getClassNameMapping = (appState) => {
+  const formName = appState.get('currentFormName');
+  let authenticatorType = appState.get('authenticatorType');
+  const recoveryAuthenticator = appState.get('recoveryAuthenticator');
+  let isPasswordRecoveryFlow = recoveryAuthenticator && recoveryAuthenticator.type === 'password';
```
haishengwu-okta, line 88: looks like we could move this logic to appState: `isPasswordRecovery`? 
(either derived properties or function, leaning toward later)

src/v2/ion/RemediationConstants.js
```diff
@@ -42,6 +42,68 @@ const FORMS = {
...
+
+  if (authenticatorType === 'phone') {
+    // both sms and call have same type
+    authenticatorType = appState.get('currentAuthenticatorEnrollment') &&
+    appState.get('currentAuthenticatorEnrollment').methods[0].type;
+  }
+  let key = `${formName}`;
+  if (authenticatorType) {
+    key += `-${authenticatorType}`;
```
haishengwu-okta, line 97: I wonder if using `--` is slightly better to distinguish form name and authenticator type?
e.g. `challenge-authenticator--password`

@anipendakur-okta any thoughts?

anipendakur-okta: Yes, but it would require more refactor and testing than what is needed for this PR? I'm good either doing as a part of this PR or have a backlog task..

src/v2/ion/RemediationConstants.js
nikhilvenkatraman-okta: right, I think this can be removed.

src/v2/ion/RemediationConstants.js
nikhilvenkatraman-okta: yeah, authenticatorType is phone for sms and call so I had to look into the method[0].type

src/v2/ion/RemediationConstants.js
haishengwu-okta: I meant only for new class format `form name + authenticator type`.
V1 class names will still be whatever format it was.

anipendakur-okta:
LGTM, after @haishengwu-okta 's comments

src/v2/ion/RemediationConstants.js
```diff
@@ -42,6 +42,68 @@ const FORMS = {
...
+  }
+
+  if (isPasswordRecoveryFlow) {
+    // password reset flow from identifier page with recoveryAuthenticator add forgot-password class
+    FORMNAME_CLASSNAME_MAPPING[key] = ['identify', 'forgot-password'];
+  }
+  return FORMNAME_CLASSNAME_MAPPING[key];
```
gowthamidommety-okta, line 104: should we return `[]` if `FORMNAME_CLASSNAME_MAPPING[key]` is undefined?

src/v2/ion/RemediationConstants.js
```diff
@@ -42,6 +42,68 @@ const FORMS = {
...
+    appState.get('currentAuthenticatorEnrollment').methods[0].type;
+  }
+  let key = `${formName}`;
+  if (authenticatorType) {
+    key += `-${authenticatorType}`;
+  }
+
+  if (isPasswordRecoveryFlow) {
+    // password reset flow from identifier page with recoveryAuthenticator add forgot-password class
+    FORMNAME_CLASSNAME_MAPPING[key] = ['identify', 'forgot-password'];
```
gowthamidommety-okta, line 102: For my information, what `key`s would this override in Password Recovery Flow?

nikhilvenkatraman-okta: We have 2 ways to start password recover in M1 

- From Identifier screen when we have both username and password shown on identifier page
- Or from password authenticator challenge screen 

the first flow is similar to the V1 password reset flow, hence we are adding the V1 class name to that screen.

src/v2/ion/RemediationConstants.js
nikhilvenkatraman-okta: Doing this change is actually exposing a big we have where the derived properties are not getting updated correctly. When we clear the appState it still keeps the values for the derived properties intact causing randomness.

src/v2/ion/RemediationConstants.js
nikhilvenkatraman-okta: I think we may have another recovery flow where we get `recoveryAuthenticator ` in the response 

Screen Shot 2020-07-21 at 2 01 26 PM

package.json
```diff
@@ -18,11 +18,6 @@
     "> 0.1%",
     "not ie <= 9"
   ],
-  "husky": {
```
haishengwu-okta, line 21: don't commit this. 🤦

nikhilvenkatraman-okta: yup, removed

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
...
+    'password':['reset-authenticator', 'forgot-password']
+  },
+};
+
+const getClassNameMapping = (appState) => {
+  const formName = appState.get('currentFormName');
+  const authenticator = appState.get('currentAuthenticatorEnrollment') || appState.get('currentAuthenticator');
+  let authenticatorType = authenticator && authenticator.type;
+  const methodType = authenticator && authenticator.methods && authenticator.methods[0].type;
+  const isPasswordRecoveryFlow = appState.get('recoveryAuthenticator') &&
```
haishengwu-okta, line 56: replaced with `isPasswordRecovery` ?

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
...
+  },
+  [FORMS.RESET_AUTHENTICATOR]: {
+    'password':['reset-authenticator', 'forgot-password']
+  },
+};
+
+const getClassNameMapping = (appState) => {
+  const formName = appState.get('currentFormName');
+  const authenticator = appState.get('currentAuthenticatorEnrollment') || appState.get('currentAuthenticator');
+  let authenticatorType = authenticator && authenticator.type;
```
haishengwu-okta, line 54: `appState.get('authenticatorType')`

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
...
+  const methodType = authenticator && authenticator.methods && authenticator.methods[0].type;
```
haishengwu-okta, line 55: nit: maybe also create derived prop in appstate as well?

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
...
+const FORMNAME_CLASSNAME_MAPPINGS = {
+  [FORMS.IDENTIFY]: {
+    'identify': ['identify', 'primary-auth'],
+    'password': ['identify', 'primary-auth']
+  },
+  [FORMS.ENROLL_PROFILE]: {
+    'enroll-profile': ['enroll-profile', 'registration'],
+  },
+  [FORMS.CHALLENGE_AUTHENTICATOR]: {
+    email: ['challenge-authenticator-email', 'mfa-verify-passcode'],
```
haishengwu-okta, line 13: I thought we gonna remove the redundancy like `challenge-authenticator-email` ?

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
...
+    password: ['challenge-authenticator-password', 'mfa-verify-password'],
+    sms: ['challenge-authenticator-voice', 'mfa-verify-passcode'],
```
haishengwu-okta, line 15: aren't there `mfa-verify-sms`, `mfa-verify-voice` ?

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
...
+    voice: ['challenge-authenticator-voice', 'mfa-verify-passcode'],
+    'security_question': ['challenge-authenticator-security_question', 'mfa-verify-question']
+  },
+  [FORMS.ENROLL_AUTHENTICATOR]: {
+    email: ['enroll-authenticator-email', 'enroll-email'],
+    password: ['enroll-authenticator-password', 'enroll-password'],
+    sms: ['enroll-authenticator-sms', 'enroll-sms'],
+    voice: ['challenge-authenticator-voice', 'mfa-verify-passcode'],
```
haishengwu-okta, line 23: is it `enroll-voice` ?

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
...
+    'security_question': ['enroll-authenticator-security_question', 'enroll-question'],
+  },
+  [FORMS.AUTHENTICATOR_ENROLLMENT_DATA]: {
```
haishengwu-okta, line 26: don't define it if there are no override.

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
+/* eslint complexity: [2, 25] */
+import {FORMS} from './RemediationConstants';
+
+const FORMNAME_CLASSNAME_MAPPINGS = {
+  [FORMS.IDENTIFY]: {
+    'identify': ['identify', 'primary-auth'],
+    'password': ['identify', 'primary-auth']
+  },
+  [FORMS.ENROLL_PROFILE]: {
+    'enroll-profile': ['enroll-profile', 'registration'],
```
haishengwu-okta, line 10: use `_` as default ?

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
+/* eslint complexity: [2, 25] */
+import {FORMS} from './RemediationConstants';
+
+const FORMNAME_CLASSNAME_MAPPINGS = {
+  [FORMS.IDENTIFY]: {
+    'identify': ['identify', 'primary-auth'],
```
haishengwu-okta, line 6: default to `_`?

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
...
+    'password': ['identify', 'primary-auth']
```
haishengwu-okta, line 7: looks like this is not used.

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: I didnt see any references for these. I saw that we use `mfa-verify-passcode` class in  https://github.com/okta/okta-signin-widget/blob/master/src/views/mfa-verify/PassCodeForm.js#L71

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: I tried that approach, it required me to define a formName entry for every form. This is an edge case for when you have more than one class name and dont have a factor, only for those cases we use the key as formname

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: What is the redundancy here ? For the `challenge-authenticator form` with `email` authenticator  we are adding a `challenge-authenticator-email` class. 

We could generate this dynamically and only keep the override className here but I think its easier to know what classes are getting added if they are all in the same place.

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: These cases still have an `authenticatorType` hence it goes into the if block and tries to evaluate `FORMNAME_CLASSNAME_MAPPINGS[formName][type]`

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: Referencing the derived property from appState actually exposing a bug we have where the derived properties are not getting updated correctly when the ION response changes. When we clear the appState, it still keeps the values for the derived properties intact causing randomness.

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: Referencing the derived property from appState actually exposing a bug we have where the derived properties are not getting updated correctly when the ION response changes. When we clear the appState, it still keeps the values for the derived properties intact causing randomness.

src/v2/ion/ViewClassNamesFactory.js
haishengwu-okta: I see it's kind of similar. basically need to have `let type = '_'` in the following method.

src/v2/ion/ViewClassNamesFactory.js
haishengwu-okta: interesting. sounds like inconsistency in v1...

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: This is so that we can add the `primary-auth` classname for when we have password authenticator in response on the identifier screen i.e identifier + password on login screen use case.

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: - FORMNAME_CLASSNAME_MAPPINGS does not currently have an entry for every formName example success, terminal etc. 
- For those cases that are not here we are assuming that there is no type so we just return [formName] as the class.
- If we use type as `_`,  we would need to add an entry here for that formName with the type as `_` ?

src/v2/ion/ViewClassNamesFactory.js
haishengwu-okta: because we suppose to generate `challenge-authenticator-email` instead of hard code. 
At lease `${FORMS.CHALLENGE_AUTHENTICATOR}-email`?

Readability is good point. I feel it's fine as long as we can make clear about rule that how addition css class are generated.

src/v2/ion/ViewClassNamesFactory.js
haishengwu-okta: - I probably meant base on the another comment that remove V2 classes from this mapping.
- Also, shall those class be `authenticator-enrollment-data-sms`, `authenticator-enrollment-data-voice` ?

src/v2/ion/ViewClassNamesFactory.js
haishengwu-okta: that sounds big trouble. 
when we render the view, AppState **shall** been refreshed. (last two line in `AppState#setIonResponse`)
Otherwise we need to figure out why/how the bug happens.

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
...
+    **/
+    type = `${methodType}`;
+  }
+  else if (authenticatorType) {
+    type = `${authenticatorType}`;
+  }
+
+  if (isPasswordRecoveryFlow) {
+    // password reset flow from identifier page with recoveryAuthenticator add forgot-password class
+    FORMNAME_CLASSNAME_MAPPINGS[formName][type] = ['identify', 'forgot-password'];
```
haishengwu-okta, line 73: what will get added for the `SELECT_AUTHENTICATOR_AUTHENTICATE` from when password recovery flow?
will it get `identify` which seem not make sense.
or this definition doesn't add `forgot-password` either.

```
[FORMS.SELECT_AUTHENTICATOR_AUTHENTICATE]: {
    'password': ['select-authenticator-authenticate']
  },
```

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+  if (type) {
+    return FORMNAME_CLASSNAME_MAPPINGS[formName][type];
+  } else {
+    /**
+      If there is no type but you have more than one class to be applied
+      add an entry for FORMNAME_CLASSNAME_MAPPINGS[formName]
+    **/
+    return FORMNAME_CLASSNAME_MAPPINGS[formName] || [formName];
```
haishengwu-okta, line 82: shall this be `FORMNAME_CLASSNAME_MAPPINGS[formName][formName]` ?

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,88 @@
...
+    /**
+      Both sms and call have same type phone
+      currentAuthenticatorEnrollment is during verify and currentAuthenticator during enroll flows
+    **/
+    type = `${methodType}`;
+  }
+  else if (authenticatorType) {
+    type = `${authenticatorType}`;
+  }
+
```
haishengwu-okta, line 70: Wonder if 70-82 can be

```suggestion
cssNames = either FORMNAME_CLASSNAME_MAPPINGS[formName][type] or FORMNAME_CLASSNAME_MAPPINGS[formName][formName];

if is password recovery then
  cssNames.push('forgot-password')
endif

```

haishengwu-okta: can be simplified to if use `_` as default type.
```
let type = '_';
if authenticator exists then
  type = authenticatorType;

if authenticator == phone then
  type = methodType;

cssNames = FORMNAME_CLASSNAME_MAPPINGS[formName][type];

if is password recovery then
  cssNames.push('forgot-password')
endif
```

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: I have a feeling its similar to what @magizh-okta had once brought up related to derived properties https://okta.slack.com/archives/C7KV3NLE8/p1588011981249400

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: I want sure if this password recovery flow with `SELECT_AUTHENTICATOR_AUTHENTICATE ` is the equivalent of V1 password recovery flow. Did we even have an equivalent for this in V1 for this ? This is why I added 

```
[FORMS.SELECT_AUTHENTICATOR_AUTHENTICATE]: {
    'password': ['select-authenticator-authenticate']
  }
```
I can add `forgot-password` to that list of classnames if that makes sense

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: yeah I did think about that but I left it as is since part `${FORMS.CHALLENGE_AUTHENTICATOR}-email` is still hardcoded. But I don't feel too strongly about it. I can update it to use the formName for now

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: 🤦  yup will update

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: Right, we would still eventually return` FORMNAME_CLASSNAME_MAPPINGS[formName][type]` ?

For forms like `success`, `terminal` and a few others without type, the type would evaluate to `_` which would then mean that we need to add an entry 

```
  [FORMS.SUCCESS]: {
    '_': ['success']
  },
```
I am trying to avoid adding an entry for each form and doing `FORMNAME_CLASSNAME_MAPPINGS[formName] || [formName];` which would return ['success'] if it doesn't find an entry for a form.

src/v2/ion/ViewClassNamesFactory.js
haishengwu-okta: It's slightly different. Magizh was listening to a derived prop changes and in the handler he expects all derived props got updated. however it's not the case yet (courage bug) because derived prop update (plus fire change event) is happening in sequence.
At your case, when we hit `className` function where invokes `getClassNameMapping`, all derived props shall have been updated. Otherwise it's another bug or something I wasn't aware.

src/v2/ion/ViewClassNamesFactory.js
haishengwu-okta: yes, if we omit V2 form that has no corresponding to V1,  it could be
```
v2ClassName = formName;
if type != '_' 
  v2ClassName = formName + '--' + type;

v1Classes = [];
if FORMNAME_CLASSNAME_MAPPINGS[formName]
  v1Classes = FORMNAME_CLASSNAME_MAPPINGS[formName][type];

cssNames = v1Classes.concat(v2ClassName);
```

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: clearing the app state cache fixed this issue this.trigger('cache:clear');

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: IMO this feels like someone has to look at this code and figure out what classes are being added and my goal was to to make it so that someone could just look at the mapping and see that the classnames are being applied with an exception that if an entry does not exist we will add the formName as the class name as default.

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,84 @@
+/* eslint complexity: [2, 25] */
+import {FORMS} from './RemediationConstants';
+
+const FORMNAME_CLASSNAME_MAPPINGS = {
+  [FORMS.IDENTIFY]: {
+    'identify': ['primary-auth'],
```
haishengwu-okta, line 6: identify --> `FORMS.IDENTIFY`

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -0,0 +1,84 @@
...
+    'password': ['primary-auth']
+  },
+  [FORMS.ENROLL_PROFILE]: {
+    'enroll-profile': ['registration'],
```
haishengwu-okta, line 10: enroll-profile --> `FORMS.ENROLL_PROFILE`

PR #1326, opened 2020-07-21
feat: trigger 'action' error into existing form

haishengwu-okta:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

- https://okta.box.com/s/ck8sia4aw9x41hurxtedw9epnjgqh5mu

### Issue:

- okta-314251


src/v2/controllers/FormController.js
```diff
@@ -52,7 +52,11 @@ export default Controller.extend({
       return;
     }
 
-    this.listenTo(this.formView, 'save', this.handleFormSave);
```
haishengwu-okta, line 55: code clean up since `save` was triggered through `appState#saveForm` event.

src/v2/controllers/FormController.js
```diff
@@ -109,18 +115,21 @@ export default Controller.extend({
...
   showFormErrors (model, error) {
     //check if error format is an ION response by looking for version attribute. To handle both types of responses.
     if(error.version) {
       const convertedErrors = IonResponseHelper.convertFormErrors(error);
-      model.trigger('error', model, convertedErrors, convertedErrors.responseJSON.errorCauses.length ? false : true);
+      const showBanner = convertedErrors.responseJSON.errorCauses.length ? false : true;
+      model.trigger('error', model, convertedErrors, showBanner);
     } else {
-      model.trigger('error', model, {'responseJSON': error}, true);
+      model.trigger('error', model, {responseJSON: error}, true);
```
nikhilvenkatraman-okta, line 131: Is this mostly for LEA or do we have other places where we get back a non ION error response  ?

haishengwu-okta: - I'm actually not sure before but wonder if @pradeepdewda-okta aware?
- now there are couple of places where error resp is constructed in this Controller.

nikhilvenkatraman-okta: I see that we are calling showFormErrors above https://github.com/okta/okta-signin-widget/pull/1326/files#diff-cc2782304bc3071a149a654082447a69R96-R99 without setting the version. That might be another use case?

src/v2/controllers/FormController.js
```diff
@@ -109,18 +115,21 @@ export default Controller.extend({
         } else {
           this.showFormErrors(model, error);
         }
+      })
+      .finally(() => {
+        this.toggleFormButtonState(false);
```
magizh-okta, line 120: Some forms might not have any button in form button bar. I guess we can check for button's existence and then toggle here. Or Jquery handles it ? 

button && button.toggleClass('link-button-disabled', disabled);

https://github.com/okta/okta-signin-widget/blob/master/src/v2/controllers/FormController.js#L141

playground/mocks/data/idp/idx/error-forgot-password.json
```diff
@@ -0,0 +1,16 @@
+{
+  "stateHandle": "01OCl7uyAUC4CUqHsObI9bvFiq01cRFgbnpJQ1bz82",
+  "version": "1.0.0",
+  "expiresAt": "2020-05-18T20:42:23.000Z",
+
+  "intent": "LOGIN",
+  "messages": {
```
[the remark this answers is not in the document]
haishengwu-okta: Usually they can back messages along with remediations. I'll update this mock.

src/v2/controllers/FormController.js
haishengwu-okta: That's fine. jQuery handles the edge case.

src/v2/controllers/FormController.js
haishengwu-okta: this is kind of new case where I'm trying to reuse the `showFormError` method.
I think it was because `showFormErrors` only handle error from `idx.process` may not return API error response but manually crafted error.

playground/mocks/data/idp/idx/error-forgot-password.json
```diff
@@ -0,0 +1,16 @@
...
+    "type": "array",
+    "value": [
+      {
+        "message": "Reset passwordis not allowed at this time. Please contact support for assistance.",
```
pradeepdewda-okta, line 11: typo : `passwordis`

src/v2/controllers/FormController.js
pradeepdewda-okta: It's a fall back for LEA cases. I don't know of any other places.

pradeepdewda-okta, line 131: needs to be 'responseJSON' ?

src/v2/controllers/FormController.js
```diff
@@ -71,32 +75,34 @@ export default Controller.extend({
...
-    this.options.appState.set('currentFormName', formName);
-  },
-
   handleFormSave (model) {
     const formName = model.get('formName');
+
     const idx = this.options.appState.get('idx');
     if (!idx['neededToProceed'].find(item => item.name === formName)) {
-      model.trigger('error', model, { errorSummary: `Cannot find http action for "${formName}".`});
+      var errorResp = {
```
pradeepdewda-okta, line 96: use let or const..

src/v2/controllers/FormController.js
```diff
@@ -71,32 +75,34 @@ export default Controller.extend({
...
       actionFn()
         .then(this.handleIdxSuccess.bind(this))
         .catch(error => {
-          throw error;
+          this.showFormErrors(this.formView.model, error);
         });
     } else {
-      throw `Invalid action selected: ${actionPath}`;
+      const errorResp = {
+        errorSummary: `Invalid action selected: ${actionPath}`,
```
pradeepdewda-okta, line 85: does it make sense to show actionPath in the error to enduser ? I don't think it's necessary.

src/v2/controllers/FormController.js
haishengwu-okta: why? I believe doesn't matter.

src/v2/controllers/FormController.js
haishengwu-okta: I think this is more for handling edge case where it shall never happens to enduser. If it does, surface detail might help troubleshooting?
Alternatively we can do log.

pradeepdewda-okta: I think keeping just the sentence should be sufficient for enduser.. `Invalid action selected. Please <do something>`... and we can log the action path detail in console

src/v2/view-builder/internals/BaseView.js
```diff
@@ -43,6 +43,7 @@ export default View.extend({
     const IonModel = this.createModelClass(
       this.options.currentViewState,
       optionUiSchemaConfig);
+
     const model = new IonModel ({
```
anipendakur-okta, line 47: How about change this to 
```this.model = new IonModel ({```
And remove line #55, then change the rest of occurrences of `model` to `this.model` ?

Comment (haishengwu-okta) [reply src/v2/view-builder/internals/BaseView.js:47]: There are many places need to update to use `this.model` hence I'm being a little lazy to just make minimal changes. :D

File: test/testcafe/framework/page-objects/BasePageObject.js [hunk 9]
```diff
@@ -34,6 +34,10 @@ export default class BasePageObject {
     return Selector(selector).innerText;
   }
 
+  getIonMessages () {
```
haishengwu-okta, line 37: move around methods

test/testcafe/framework/page-objects/ChallengeFactorPageObject.js
```diff
@@ -27,14 +25,6 @@ export default class ChallengeFactorPageObject extends BasePageObject {
     return Selector(SWITCH_AUTHENTICATOR_SELECTOR).textContent;
   }
 
-  forgotPasswordExists() {
```
haishengwu-okta, line 30: Moved to ChallengePasswordPageObject

src/v2/controllers/FormController.js
```diff
@@ -109,18 +111,29 @@ export default Controller.extend({
...
-    //check if error format is an ION response by looking for version attribute. To handle both types of responses.
-    if(error.version) {
+    this.formView.clearFormError();
+    if (!error) {
+      error = 'FormController - unknown error found';
+      this.options.settings.callGlobalError(error);
+    }
+
+    if(IonResponseHelper.isIonErrorResponse(error)) {
+      // check if error format is an ION response by looking for version attribute.
```
pradeepdewda-okta, line 128: you can put this comment in IonResponseHelper

src/v2/view-builder/internals/BaseView.js
```diff
@@ -77,4 +79,11 @@ export default View.extend({
     return BaseModel.create(currentViewState, optionUiSchemaConfig);
   },
 
+  clearFormError () {
```
pradeepdewda-okta, line 82: Why we are overriding this method? Can't the courage handling of clearing error is not sufficient ?

haishengwu-okta: This is not override but invoking `clearErrors` on BaseForm.
BaseForm does auto clear error on form submit. For errors from other places (click link at footer) are also being render as form error now hence we have to manually clear it up. (Imaging user click a link multiple times where all requests are error)

pradeepdewda-okta: is it possible to trigger this method only when there are such unbound errors ? Rather then calling this method on every show error.

haishengwu-okta: How so. could you elaborate?

BTW, the if condition here is to make this function as no-operation when there is not existing error.

pradeepdewda-okta: Calling this function on line 121, give me impression that it clears all form errors. But it clears only pre-existing errors which can't be automatically cleared.

Also, i was in the impression that showFormErrors only gets called on form submit.

Can you attach a video so that i can understand better?

pradeepdewda-okta: Can you add a video that shows this scenario ?

pradeepdewda-okta: Just wanted to see if there is a way we can avoid doing this.

haishengwu-okta: - what's different when you say `all form errors` v.s `pre-existing errors` ?
- `showFormErrors` is a general error handler now
- you can reproduce the issue easily if you can start playground on this topic branch. (remove this call at ` line 121` and click forgot password twice). It might be easier than a video.

haishengwu-okta: Ideally Courage Base Form can handle `model.trigger('clearError')` event.
We could add this event to `v2/BaseForm` but I figure calling function is easier to understand entire flow.

pradeepdewda-okta: If we already do `model.trigger('error')`, we can trigger `clearError` as well. I am insisting on this because in one of my PR you suggested to not create functions that does the similar thing in courage. So i wanted to stick to that philosophy, because it made sense :) if `model.trigger('clearError') can do the trick then we should settle for that

haishengwu-okta: Sounds good. thanks for catching my inconsistency!

playground/mocks/data/idp/idx/error-forgot-password.json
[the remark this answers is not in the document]
magizh-okta: No remediation here ?

PR #1336, opened 2020-07-23
fix: do not show sign out link in v2 when config is set

mauriciocastillosilva-okta:
## Description:

- if features.hideSignOutLinkInMFA is true, don't show cancel/signout link for verify identity forms

Resolves: OKTA-311941

## PR Checklist

- [x] Added unit tests?
- [ ] Added e2e tests

### Screenshot/Video:
hideSignOutInMFA-SIW

### Issue:

- [OKTA-311941](https://oktainc.atlassian.net/browse/OKTA-311941)


src/v2/view-builder/internals/BaseFooter.js
```diff
@@ -39,7 +39,10 @@ export default View.extend({
       links = links.filter(l => $.isPlainObject(l));
     }
 
-    if (this.options.appState.get('showSignoutLink')) {
+    if (this.options.settings.get('features.hideSignOutLinkInMFA') &&
+        this.options.appState.get('isVerifyIdentityForm')) {
+      // do not add cancel/signout link if the config is on and it is a verify identity form
+    } else if (this.options.appState.get('showSignoutLink')) {
```
gowthamidommety-okta, line 45: why can't we just check for !features.hideSignOutLinkInMFA here?

test/unit/spec/v2/view-builder/internals/BaseFooter_spec.js
```diff
@@ -125,5 +131,30 @@ describe('v2/view-builder/internals/BaseFooter', function () {
     ]);
   });
 
+  it('does not add signout link when features.hideSignOutLinkInMFA is true, even if `showSignoutLink` is true', () => {
```
haishengwu-okta, line 134: do we need a test case for `features.hideSignOutLinkInMFA` is true?

src/v2/view-builder/internals/BaseFooter.js
mauriciocastillosilva-okta: because the config is for MFA only, it is not for all flows. So I don't want to completely hide it from all forms if the feature is enabled, I only want to hide it for the MFA forms.

test/unit/spec/v2/view-builder/internals/BaseFooter_spec.js
mauriciocastillosilva-okta: oh sigh... yes, actually that is the one we need and that is what I meant to do... fixing it

src/v2/view-builder/internals/BaseFooter.js
nikhilvenkatraman-okta: If i understand this correct, we wont send `features.hideSignOutLinkInMFA` from the server every-time? We will only send it as true for MFA forms ie 
```
FORMS.SELECT_AUTHENTICATOR_AUTHENTICATE
FORMS.CHALLENGE_AUTHENTICATOR
```

mauriciocastillosilva-okta: okta hosted widget will always send that config, but it will be `false` in all flows EXCEPT enrolling an authenticator from end user settings page.

nikhilvenkatraman-okta: 2 questions 

- If its going to be true for enrolling an authenticator from end user settings page should these be ? 

```
FORMS.SELECT_AUTHENTICATOR_ENROLL,
FORMS.ENROLL_AUTHENTICATOR
```

- Secondly in that case can we just check if  `features.hideSignOutLinkInMFA` is true ? As the backend knows to set this  to true for only these flow

mauriciocastillosilva-okta: no, `ENROLL_AUTHENTICATOR` is already on the list of `FORMS_WITHOUT_SIGNOUT`, and that is an enroll form.
the config `hideSignOutLinkInMFA` only affects challenge forms, that is why I need to add `CHALLENGE_AUTHENTICATOR`

Not sure I understand your second point. We were discussing this in chat, since this config is only meant for challenge/verify forms, I cannot just check for that alone, I need to also be sure it is a form that is affected by that config

haishengwu-okta: another opinion is change `showSignoutLink ` to be a function instead of derive props ?
```
showSignoutLink(hideSignOutLinkInMFA) {
 ...
}
```

src/v2/ion/RemediationConstants.js
```diff
@@ -61,8 +61,14 @@ const FORMS_WITH_STATIC_BACK_LINK = [
   FORMS.SELECT_AUTHENTICATOR_AUTHENTICATE
 ];
 
+const FORMS_FOR_IDENTITY_VERIFICATION = [
```
haishengwu-okta, line 64: why not just say `FORMS_FOR_VERIFICATION `

src/v2/view-builder/internals/BaseFooter.js
nikhilvenkatraman-okta: yeah, that way we can keep all the checks in one place

mauriciocastillosilva-okta: done

src/v2/models/AppState.js
```diff
@@ -137,6 +123,18 @@ export default Model.extend({
     return reRender;
   },
 
+  showSignoutLinkInCurrentForm (hideSignOutLinkInMFA) {
```
haishengwu-okta, line 126: - maybe add comments to elaborate use cases, why we need complicated logic.?
- rename to some thing like `isXXX` or `shallXXX` (e.g. shallShowSignoutLink) ?

test/unit/spec/v2/view-builder/internals/BaseFooter_spec.js
```diff
@@ -2,15 +2,25 @@ import BaseFooter from 'v2/view-builder/internals/BaseFooter';
...
 describe('v2/view-builder/internals/BaseFooter', function () {
 
   let testContext;
 
-  const renderFooter = (links) => {
+  const renderFooter = (links, currentFormName) => {
     const FooFooter = BaseFooter.extend({
       links,
     });
+    spyOn(AppState.prototype, 'get').and.callFake(name => {
```
haishengwu-okta, line 15: nit:
I'd prefer to unit test BaseFooter by mocking `showSignoutLinkInCurrentForm` function. Plus unit test `showSignoutLinkInCurrentForm` itself. and add testcafe for _integration_ testing.

mauriciocastillosilva-okta: there are currently no unit tests for AppState, but I'll create one for this then. I won't cover the rest of the AppState model though, that would have to be done later

haishengwu-okta: Yeah, didn't mean to create unit test for other method of AppState but only the new method.

PR #1343, opened 2020-07-25
feat: add support for email enroll flow

haishengwu-okta:
## Description:

- Add support for enroll email authenticator.
- Refactor class structure for email enroll and verify to code reusability.
- Add testcafe tests and refactor testcafe naming convention.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

- https://okta.box.com/s/n9yicopur1yvzl8wbx07voy123d7fyqe

### Issue:

- OKTA-316078


src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -0,0 +1,86 @@
+import { loc, View, createCallout, _ } from 'okta';
+import BaseForm from '../../internals/BaseForm';
+import email from '../shared/email';
+import polling from '../shared/polling';
+import BaseAuthenticatorView from '../../components/BaseAuthenticatorView';
+import { SHOW_RESEND_TIMEOUT } from '../../utils/Constants';
+
+const ResendView = View.extend(
```
anipendakur-okta, line 8: Do we have a ticket to extract this into its own component? We have a few places now to be refactored I believe.

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -0,0 +1,86 @@
...
+      this.showCalloutWithDelay();
+    },
+
+    postRender () {
+      this.showCalloutWithDelay();
+    },
+
+    showCalloutWithDelay () {
+      this.showMeTimeout = _.delay(() => {
+        this.$el.removeClass('hide');
```
anipendakur-okta, line 38: This needs to be changed to `this.el.classList.remove('hide')` which makes it inline with line 26-28.

src/v2/view-builder/views/email/ChallengeAuthenticatorEmailView.js
```diff
@@ -1,107 +1,49 @@
...
-      const sendText = loc('oie.email.verify.sentText', 'login');
-      const enterCodeText = loc('email.mfa.email.sent.description.emailCodeText', 'login');
+  template: hbs`
+    {{i18n code="oie.email.verify.sentText" bundle="login"}}&nbsp;
+    <span class="strong">{{email}}.</span>
+    &nbsp;{{i18n code="email.mfa.email.sent.description.emailCodeText" bundle="login"}}
+  `,
+
+  getTemplateData () {
+    return _.pick(this.options, 'email');
```
anipendakur-okta, line 21: importing `_` just for this seems like an overkill. I think it would be better to just do ` return this.options && this.options.email;` and remove the `_` import.

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
haishengwu-okta: https://oktainc.atlassian.net/browse/OKTA-317315

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
haishengwu-okta: Shall we use more jQuery instead of native DOM API so that we dont need to worry about cross browser at all?
If our end goal was to deprecate jQuery (which definitely not short term goal), we could do it in one (or many) dedicated rounds ?

nikhilvenkatraman-okta: yeah, we should pick one style for a file. Maybe avoid using jquery if possible for this entire file?

anipendakur-okta: * I think it's better to move away from jQuery & and `_` as much as possible. In the long term, we'll be able to remove the heavy dependencies (hopefully 🤞 ). Most of the functionality should be achievable using ES6. And once we have support for `optional chaining` and `nullish coalescing` we can simplify existential checks too.
* At the least, we shouldn't be mixing jQuery and JS in the same file to avoid confusion..

haishengwu-okta: - Removing jQuery could hardly be a goal because of backbone depends on it. thus I usually prefer to use jQuery so I dont have to worry my code is cross browser compatible (search for caniuse or MDN), even though there are less and less API gaps between browsers. Like I mentioned earlier, if we want to clean up jQuery, I hope it can be done by some transformation tool (e.g. if there is usage of `hasClass`, convert to `el.classList.contain`)
- Same story for underscore but it only provides utilizes hence we have better chance to avoid it.

- Mixing is always bad practice.

PR #1346, opened 2020-07-27
feat: add support for brandName

nikhilvenkatraman-okta:
Resolves: OKTA-312535

## Description:

**TLDR;** 

If a brandName  is configured via a widget config, then on some of the screens we will adjust the title to show with the brandName in context.

**For example on the Password Reset screen**

without brandName = Reset your password
with brandName = Reset your Spaghett Inc password

IDX widget screens where we need to add support for `brandName` for parity with V1

```
Enroll Choices: 
enroll.choices.description.generic = Your company requires multifactor authentication to add an additional layer of security when signing in to your account
enroll.choices.description.specific = Your company requires multifactor authentication to add an additional layer of security when signing in to your {0} account

Password Reset: 
password.reset.title.generic = Reset your password
password.reset.title.specific = Reset your {0} password

Password Expired: 
password.expired.title.generic = Your password has expired
password.expired.title.specific = Your {0} password has expired

Password Expiring:
password.expiring.subtitle.generic = When password expires you will be locked out of your account.
password.expiring.subtitle.specific = When password expires you will be locked out of your {0} account.

```

**V1 screens that support brandName but don't exist in M1 yet**

```
Windows Hello : (Not built in M1)
factor.windowsHello.signin.description.generic = Sign in using Windows Hello. factor.windowsHello.signin.description.specific = Sign in to {0} using Windows Hello.

Verify Windows Hello: (Not built in M1)
verify.windowsHello.subtitle.signingIn.generic = Signing in...
verify.windowsHello.subtitle.signingIn.specific = Signing in to {0}...

U2F : (Not built in M1)
factor.u2f.description.generic = Use a Universal 2nd Factor (U2F) security key to sign in.
factor.u2f.description.specific = Use a Universal 2nd Factor (U2F) security key to sign in to {0}.

Custom Factor: (Not built in M1)
factor.customFactor.description.generic = Redirect to a third party MFA provider to sign in.
factor.customFactor.description.specific = Redirect to a third party MFA provider to sign in to {0}.

Enroll TOTP: (Not built in M1)
enroll.totp.manualSetupInstructions.generic = To set up manually enter your Account username and then input the following in the Secret Key Field
enroll.totp.manualSetupInstructions.specific = To set up manually enter your {0} Account username and then input the following in the Secret Key Field

Enroll TOTP Shared Secret: (Not built in M1)
enroll.totp.sharedSecretInstructions.generic = Enter your Account username and enter the following in the Secret Key Field
enroll.totp.sharedSecretInstructions.specific = Enter your {0} Account username and enter the following in the Secret Key Field

```

## PR Checklist

- [x] Added e2e tests

### Screenshot/Video:

Screen Shot 2020-07-27 at 4 18 51 PM

Screen Shot 2020-07-27 at 4 18 57 PM

Screen Shot 2020-07-27 at 4 19 02 PM

Screen Shot 2020-07-27 at 4 19 09 PM

### Issue:

- [OKTA-312535](https://oktainc.atlassian.net/browse/OKTA-312535)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -761,6 +761,7 @@ oie.okta_verify.authenticator.description = Verify with a code or receive a push
 ## Select authenticator enrollment
 oie.select.authenticators.enroll.title = Set up Authenticators
 oie.select.authenticators.enroll.subtitle = Set up authenticators to ensure that only you have access to your account.
+oie.select.authenticators.enroll.subtitle.custom = Set up authenticators for {0} to ensure that only you have access to your account.
 oie.setup.required = Set up required
 oie.setup.optional = Set up optional
 oie.optional.authenticator.button.title = Finish
--- a/src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
+++ b/src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
@@ -4,6 +4,11 @@ import EnrollAuthenticatorPasswordView from './EnrollAuthenticatorPasswordView';
 
 const Body = EnrollAuthenticatorPasswordView.prototype.Body.extend({
   className: 'password-authenticator',
+  subtitle () {
+    if (this.options.settings.get('brandName')) {
+      return loc('password.expiring.subtitle.specific', 'login', [this.options.settings.get('brandName')]);
+    }
+  },
   title () {
     const passwordPolicy = this.getPasswordPolicySettings() || {};
     const daysToExpiry = passwordPolicy.daysToExpiry;
@@ -20,19 +25,13 @@ const Body = EnrollAuthenticatorPasswordView.prototype.Body.extend({
   save () {
     return loc('password.expired.submit', 'login');
   },
-
   showMessages () {
-    // render messages as text
-    const messagesObjs = this.options.appState.get('messages');
-    if (messagesObjs && messagesObjs.value.length) {
-      let content = messagesObjs.value.map((messagesObj) => {
-        return messagesObj.message;
-      });
-      content = this.options.settings.get('brandName')?
-        [loc('password.expiring.subtitle.specific', 'login', [this.options.settings.get('brandName')])]:
-        content;
-      this.add(`<div class="ion-messages-container">${content.join(' ')}</div>`, '.o-form-error-container');
+    // if brandName is configured and messages is present, render as subtitle with brandName in context
+    if (this.options.settings.get('brandName')) {
+      return null;
     }
+    // else if brandName is not set, render messages object sent from server as text
+    EnrollAuthenticatorPasswordView.prototype.Body.prototype.showMessages.apply(this, arguments);
   },
 });
 
--- a/src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
+++ b/src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
@@ -20,6 +20,20 @@ const Body = EnrollAuthenticatorPasswordView.prototype.Body.extend({
   save () {
     return loc('password.expired.submit', 'login');
   },
+
+  showMessages () {
+    // render messages as text
+    const messagesObjs = this.options.appState.get('messages');
+    if (messagesObjs && messagesObjs.value.length) {
+      let content = messagesObjs.value.map((messagesObj) => {
+        return messagesObj.message;
+      });
+      content = this.options.settings.get('brandName')?
+        [loc('password.expiring.subtitle.specific', 'login', [this.options.settings.get('brandName')])]:
+        content;
+      this.add(`<div class="ion-messages-container">${content.join(' ')}</div>`, '.o-form-error-container');
+    }
+  },
 });
 
 const Footer = BaseFooter.extend({
--- a/src/v2/view-builder/views/SelectAuthenticatorEnrollView.js
+++ b/src/v2/view-builder/views/SelectAuthenticatorEnrollView.js
@@ -8,7 +8,10 @@ const Body = BaseForm.extend({
     return loc('oie.select.authenticators.enroll.title', 'login');
   },
   subtitle: function () {
-    return loc('oie.select.authenticators.enroll.subtitle', 'login');
+    const subtitle = this.options.settings.get('brandName') ?
+      loc('oie.select.authenticators.enroll.subtitle.custom', 'login', [this.options.settings.get('brandName')]):
+      loc('oie.select.authenticators.enroll.subtitle', 'login');
+    return subtitle;
   },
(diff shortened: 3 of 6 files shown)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -760,6 +760,7 @@ oie.okta_verify.authenticator.description = Verify with a code or receive a push
 ## Select authenticator enrollment
 oie.select.authenticators.enroll.title = Set up Authenticators
 oie.select.authenticators.enroll.subtitle = Set up authenticators to ensure that only you have access to your account.
+oie.select.authenticators.enroll.subtitle.custom = Set up authenticators for {0} to ensure that only you have access to your account.
```
anipendakur-okta, line 763: Looks like `oie.select.authenticators.enroll.subtitle` is a new key in M1 too. How about changing it to `oie.select.authenticators.enroll.subtitle.generic` and the other one to `oie.select.authenticators.enroll.subtitle.specific` ?

src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
```diff
@@ -20,6 +20,20 @@ const Body = EnrollAuthenticatorPasswordView.prototype.Body.extend({
...
+    if (messagesObjs && messagesObjs.value.length) {
+      let content = messagesObjs.value.map((messagesObj) => {
+        return messagesObj.message;
+      });
+      content = this.options.settings.get('brandName')?
+        [loc('password.expiring.subtitle.specific', 'login', [this.options.settings.get('brandName')])]:
+        content;
+      this.add(`<div class="ion-messages-container">${content.join(' ')}</div>`, '.o-form-error-container');
+    }
+  },
```
anipendakur-okta, line 36: Hi @nikhilvenkatraman-okta , looks like I may have missed this while working on the ReEnroll view.
- Shouldn't this be in a `subtitle()` function or something like that?
- Trying to understand how generic this logic is? Like should we put it in `EnrollAuthenticatorPasswordView` instead?
- Also the message seems to be something else here under Password will expire soon section -https://oktawiki.atlassian.net/wiki/spaces/UX/pages/737092192/Sign+In+Widget+Design+-+IdX#Password 
I can't remember if there was a decision made over slack, anyhow, I don't like the message in the mock as it's too detailed, that's just my opinion though.

anipendakur-okta:
Do we need specific `title` s for `ReEnrollAuthenticatorWarningPasswordView` - https://github.com/okta/okta-signin-widget/pull/1346/files#diff-4fc64ffbdec8f6a2003219c0883dd4eeR7-R18

packages/@okta/i18n/src/properties/login.properties
anipendakur-okta: This would make it inline with other changes.

src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
haishengwu-okta: would rather 
- either 
```
subtitle() { 
  if brandName then return loc('password.expiring.subtitle.specific', 'login', ...);
  else return ....
},
showMessages() {
  return null;
}
```
- or

```
subtitle() { 
  if brandName then return loc('password.expiring.subtitle.specific', 'login', ...);
},
showMessages() {
  if brandName then return null;
}
```

packages/@okta/i18n/src/properties/login.properties
nikhilvenkatraman-okta: I wasn't a fan of that naming convention TBH. I prefer to use `oie.select.authenticators.enroll.subtitle` and `oie.select.authenticators.enroll.subtitle.custom` as it tells us that its used when we have a custom title. But I am happy to update it to keep it consistent with the other keys.

src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js
nikhilvenkatraman-okta: @anipendakur-okta @haishengwu-okta I saw that on this screen the message ` When your password expires you will be locked out of your Okta account.` is coming from the server within the `messages`object which was being rendered by the `showMessages` function we have in `BaseForm`.  I was keeping it consistent with what we did in other places by overriding the `showMessages` at the view layer and add the customization logic.

I can render this as a subtitle  but that would mean we would extract `messages` object within the `subtitle` function and return null from the `showMessages ` function, just so that we can render this as a subtitle. Visually there is no difference. In all other places we render `messages` inside the ion-message-container but here we would render as a subtitle.

I am curious why this one-off is better  ?

nikhilvenkatraman-okta: when we have a `brandName` we are just displaying a static text with the brandName as a parameter. In that case a subtitle makes sense. When we don't have a brandName we can override the showMessages and render text within the ion-messages-container. This would need us to check for brandName
```
  subtitle () {
    if (this.options.settings.get('brandName')) {
      return loc('password.expiring.subtitle.specific', 'login', [this.options.settings.get('brandName')]);
    }
  },

  showMessages () {
    if (this.options.settings.get('brandName')) {
      return null;
    }
    // render messages as text
    if (messagesObjs && messagesObjs.value.length) {
      let content = messagesObjs.value.map((messagesObj) => {
        return messagesObj.message;
      });
      this.add(`<div class="ion-messages-container">${content.join(' ')}</div>`, '.o-form-error-container');
    }
  },

```
I prefer this as this seems to be clear and in a way consistent with what we are doing in other places. Let me know what you guys think.

haishengwu-okta: I guess you would do in `initialize` function
```
initialize() {
  super.initialize();
  if brand name exists {
    this.subtitle = () => { return loc(...); }
    this.showMessages = () => {}; // literally no-op
  }
}
```

anipendakur-okta: I think I like @nikhilvenkatraman-okta 's solution better for a few reasons. 
* We are using functions that probably stick to `prototype` or whatever `View.extend({})` does, that way we're following courage guidelines.
* `this.subtitle = () => {return loc();}` will force the function on the instance level which seems a bit hacky to me.

haishengwu-okta: the problem is `showMessages` has duplicated code form its parent unless we can do
```
showMessages() {
  if brand name then return null
  else super.showMessages();
}
```

nikhilvenkatraman-okta: updated to call `showMessages ` from BaseForm

PR #1359, opened 2020-07-30
optional chaining refactor

anipendakur-okta:
## Description:
Using optional chaining as it's in stage 4 of wc39 recommendation. (Included in babel 7 already). I have verified that it gets traslated in `okta-sign-in.js` as per Babel REPL.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Reviewers:
@haishengwu-okta 

### Issue:

- [OKTA-313084](https://oktainc.atlassian.net/browse/OKTA-313084)

### DIFF
--- a/src/v2/BaseLoginRouter.js
+++ b/src/v2/BaseLoginRouter.js
@@ -93,7 +93,7 @@ export default Router.extend({
   },
 
   handleIdxResponseFailure (error = {}) {
-    if (error && error.details && error.details.stateHandle) {
+    if (error?.details?.stateHandle) {
       // 1. loosely check whether is IDX error response
       // see idx for details: https://github.com/okta/okta-idx-js/blob/master/src/index.js
 
--- a/src/v2/ion/i18nTransformer.js
+++ b/src/v2/ion/i18nTransformer.js
@@ -120,7 +120,7 @@ const updateLabelForUiSchema = (remediation, uiSchema) => {
   Logger.info('\t remediationName: ', remediation.name);
   Logger.info('\t uiSchema: ', JSON.stringify(uiSchema));
 
-  const authenticatorType = remediation.relatesTo && remediation.relatesTo.value.type
+  const authenticatorType = remediation.relatesTo?.value.type
     ? `.${remediation.relatesTo.value.type}`
     : '';
   const i18nPrefix = `${remediation.name}${authenticatorType}.`;
@@ -180,7 +180,7 @@ const updateLabelForUiSchema = (remediation, uiSchema) => {
  * @param {Message} message
  */
 const getMessage = (message) => {
-  if (message.i18n && message.i18n.key) {
+  if (message.i18n?.key) {
     const i18nKey = message.i18n.key;
     if (Bundles.login[i18nKey]) {
       Logger.info(`Override messages using i18n key ${i18nKey}`);
@@ -207,7 +207,7 @@ const uiSchemaLabelTransformer = (transformedResp) => {
   // 2. See `IonResponseHelper.js` where handle `messages` object when none 200 response.
   // 3. Handling `messages` in remediation forms on 200 response is not considered yet.
   //    Is that possible?
-  if (transformedResp.messages && Array.isArray(transformedResp.messages.value)) {
+  if (Array.isArray(transformedResp.messages?.value)) {
     transformedResp.messages.value.forEach(message => {
       message.message = getMessage(message);
     });
--- a/src/v2/ion/i18nTransformer.js
+++ b/src/v2/ion/i18nTransformer.js
@@ -120,7 +120,7 @@ const updateLabelForUiSchema = (remediation, uiSchema) => {
   Logger.info('\t remediationName: ', remediation.name);
   Logger.info('\t uiSchema: ', JSON.stringify(uiSchema));
 
-  const authenticatorType = remediation.relatesTo?.value.type
+  const authenticatorType = remediation.relatesTo?.value?.type
     ? `.${remediation.relatesTo.value.type}`
     : '';
   const i18nPrefix = `${remediation.name}${authenticatorType}.`;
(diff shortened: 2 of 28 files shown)


src/v2/view-builder/views/TerminalView.js
```diff
@@ -17,7 +17,7 @@ const Body = BaseForm.extend({
       messagesObjs.value
         .forEach(messagesObj => {
           const msg = messagesObj.message;
-          if (messagesObj.class && messagesObj.class === 'ERROR') {
+          if (messagesObj.class === 'ERROR') {
```
anipendakur-okta, line 20: @haishengwu-okta - not sure what this was supposed to be. `(messagesObj.class && messagesObj.class === 'ERROR')` is same as `(messagesObj.class === 'ERROR')`

src/v2/ion/i18nTransformer.js
```diff
@@ -120,7 +120,7 @@ const updateLabelForUiSchema = (remediation, uiSchema) => {
   Logger.info('\t remediationName: ', remediation.name);
   Logger.info('\t uiSchema: ', JSON.stringify(uiSchema));
 
-  const authenticatorType = remediation.relatesTo && remediation.relatesTo.value.type
+  const authenticatorType = remediation.relatesTo?.value.type
```
haishengwu-okta, line 123: I guess it doesn't hurt to do `?.value?.type`

src/v2/view-builder/views/TerminalView.js
pradeepdewda-okta: @anipendakur-okta i think it checks if `class` as property exists or not.. In some api responses we might not have `class` property

src/v2/BaseLoginRouter.js
```diff
@@ -93,7 +93,7 @@ export default Router.extend({
   },
 
   handleIdxResponseFailure (error = {}) {
-    if (error && error.details && error.details.stateHandle) {
+    if (error?.details?.stateHandle) {
```
pradeepdewda-okta, line 96: is it going to be supported in all browsers ?

pradeepdewda-okta:
@anipendakur-okta @haishengwu-okta this is nice. thanks for doing this Ani.

Only concern, If we overuse it at the places where some objects must be defined otherwise we won't be able to catch errors ( for example if remediation must be defined then we should never use `remediation?.anything`

src/v2/ion/i18nTransformer.js
anipendakur-okta: It won't hurt, but we're unnecessarily adding extra logic / bytes into the compiled code. We should be prudent & use this only when necessary IMHO.

anipendakur-okta:
> @anipendakur-okta @haishengwu-okta this is nice. thanks for doing this Ani.
> 
> Only concern, If we overuse it at the places where some objects must be defined otherwise we won't be able to catch errors ( for example if remediation must be defined then we should never use `remediation?.anything` )

I totally agree. I chose to translate the logic of what is already out there and didn't introduce any blind checks. We shouldn't be blindly using it as it will also introduce extra logic and hence extra bytes into compiled code.

src/v2/view-builder/views/TerminalView.js
anipendakur-okta: If class didn't exist, `=== 'ERROR'` wouldn't be true hence would result in false. An existential check before it seems unnecessary.

src/v2/ion/i18nTransformer.js
pradeepdewda-okta: We need to make some guidance on when and when not to use it. probably introduce a lint to support these constraints

src/v2/view-builder/views/TerminalView.js
pradeepdewda-okta: wouldn't it throw error `can't access class of undefiend`

anipendakur-okta: `messageObj` is an array element, so it would only fail if we had something like `[undefined]` which would not happen in this case. The change is still ok if we had something like `[{}]` (no `class` in the object).

src/v2/ion/i18nTransformer.js
haishengwu-okta: Basically we'd assume when `relatesTo` exists it must have `relatesTo.value`. Ideally we shall check whether `relatesTo.value` exists before looking into `relatesTo.value.type` (no assumption on API response otherwise runtime error)
Hence I think the guidance would be also use `?.` op when looking for object properties.

src/v2/BaseLoginRouter.js
haishengwu-okta: It transpiled to JavaScript that any browser would understand. kind of same as we write `a && a.b && a.b.c` but in slim fashion.
(I guess it's because we didn't set tranpile target to be only browsers that already support chain operator)

anipendakur-okta: Yes, it's dictated by our browserslist config in package.json - 
```
"browserslist": [
  "> 0.1%",
  "not ie <= 9"
],
```
By changing the above config and transpiling again resulted in a considerable size difference in `okta-sign-in.js`

haishengwu-okta: @anipendakur-okta curious what you changed and the size difference?

anipendakur-okta: Hi @haishengwu-okta I meant I changed the browserslist config in package.json to this and noticed a change -
```
  "browserslist": [
    "last 5 chrome version"
  ]
```
image

pradeepdewda-okta:
Just one comment do you think we can introduce a linter to prevent overuse of this operator ?

anipendakur-okta:
> Just one comment do you think we can introduce a linter to prevent overuse of this operator ?

I couldn't find any lint rule, don't believe there's any that exist to constrain the number of these that can be used in a statement. I did search quite a bit and didn't come across any.

> Only concern, If we overuse it at the places where some objects must be defined otherwise we won't be able to catch errors ( for example if remediation must be defined then we should never use remediation?.anything )

This anyway can not be solved by a linter, it's more of a logical decision left to the developer.

PR #1378, opened 2020-08-15
feat: implement resend push for okta verify on idx

gowthamidommety-okta:
Resolves: OKTA-322821

## Description:

Uses new remediation form `resend-push` to challenge for push

resendPush

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-322821](https://oktainc.atlassian.net/browse/OKTA-322821)

### DIFF
--- a/playground/mocks/data/idp/idx/authenticator-verification-okta-verify-reject-push.json
+++ b/playground/mocks/data/idp/idx/authenticator-verification-okta-verify-reject-push.json
@@ -8,6 +8,10 @@
     "value": [
       {
         "message": "You have chosen to reject this login.",
+        "i18n": {
+          "key": "api.authn.poll.error.push_rejected",
+          "params": []
+        },
         "class": "ERROR"
       }
     ]
@@ -19,15 +23,38 @@
         "rel":[
           "create-form"
         ],
-        "name":"challenge-poll",
+        "name":"resend-push",
         "relatesTo":[
           "$.currentAuthenticator"
         ],
-        "href":"http://localhost:3000/idp/idx/challenge/poll",
+        "href":"http://localhost:3000/idp/idx/challenge",
         "method":"POST",
         "accepts":"application/vnd.okta.v1+json",
-        "refresh": 4000,
-        "value":[
+        "value": [
+          {
+            "name": "authenticator",
+            "type": "object",
+            "label": "Okta Verify",
+            "value": {
+              "form": {
+                "value": [
+                  {
+                    "name": "id",
+                    "value": "auteq0lLiL9o1cYoN0g4",
+                    "required": true,
+                    "mutable": false
+                  },
+                  {
+                    "name": "methodType",
+                    "value" : "push",
+                    "label" :  "PUSH",
+                    "required": true,
+                    "mutable": false
+                  }
+                ]
+              }
+            }
+          },
           {
             "name": "stateHandle",
             "required": true,
@@ -51,7 +78,7 @@
             "type": "object",
             "options": [
               {
-                "label": "Okta Verify",
+                "label": "",
                 "value": {
                   "form": {
                     "value": [
@@ -64,7 +91,7 @@
                       },
                       {
                         "name": "methodType",
-                        "required": true,
+                        "required": false,
                         "mutable": false,
                         "visible": false,
                         "type": "string",
@@ -126,13 +153,11 @@
   "currentAuthenticator":{
     "type":"object",
     "value": {
-      "profile": {
-        "name": "Okta Verify"
-      },
       "type": "app", 
       "id": "auttheidkwh282hv8g3",
+      "displayName": "",
       "methods": [
-        { "methodType": "push" }
+        { "type": "push" }
       ]
     }
   },
@@ -142,16 +167,16 @@
       {
         "type": "app",
         "id": "auteq0lLiL9o1cYoN0g4",
-        "displayName": "Okta Verify",
+        "displayName": "",
         "methods": [
           {
-            "value": "signed_nonce"
+            "type": "signed_nonce"
           },
           {
-            "value": "push" 
+            "type": "push"
           },
           {
-            "value": "totp" 
+            "type": "totp"
           }
         ]
       },
@@ -161,7 +186,7 @@
         "displayName": "Okta Password",
         "methods": [
           {
-            "value": "password"
+            "type": "password"
           }
         ]
       }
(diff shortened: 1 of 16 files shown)


src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
```diff
@@ -0,0 +1,45 @@
...
+
+    save () {
+      return loc('oie.okta_verify.push.send', 'login');
+    },
+
+    showMessages () {
+      // override showMessages to display error in cases like reject push
+      // or timeout. Borrowed this logic from TerminalView.
+      const messagesObjs = this.options.appState.get('messages');
+      if (messagesObjs && Array.isArray(messagesObjs.value)) {
```
magizh-okta, line 22: @gowthamidommety-okta  @haishengwu-okta  feels like this is something that should go in baseForm itself. What do you guys think ?

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
```diff
@@ -0,0 +1,45 @@
...
+        this.add('<div class="ion-messages-container"></div>', '.o-form-error-container');
+
+        messagesObjs.value.forEach(messagesObj => {
+          const msg = messagesObj.message;
+          if (messagesObj.class && messagesObj.class === 'ERROR') {
```
anipendakur-okta, line 27: We use `optional chaining` now in SIW. This seems like a good candidate for it -
```
if (messagesObj?.class === 'ERROR') {
```
Same concept can be used in line 22 and line 29 (probably) here.

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
haishengwu-okta: hmm, @gowthamidommety-okta did you get 200 response with `messages` when ResendPush ?

playground/mocks/data/idp/idx/authenticator-verification-okta-verify-reject-push.json
```diff
@@ -19,15 +23,38 @@
         "rel":[
           "create-form"
         ],
-        "name":"challenge-poll",
+        "name":"resend-push",
```
akhtarshiekh-okta, line 26: do you only `resend-push` remediation in `poll` when user has rejected the notificaiton?

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
gowthamidommety-okta: Yes, it's a success response, similar to the flow in v1. That's why I'm overriding it here.. BaseForm currently shows all form messages as text even if type is ERROR.

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
gowthamidommety-okta: You're right! took this from terminalView.. will change here 👍

playground/mocks/data/idp/idx/authenticator-verification-okta-verify-reject-push.json
gowthamidommety-okta: reject or timeout

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
haishengwu-okta: mock server will respond 4xx when mock file start with `error-*`.

haishengwu-okta: It may make more sense to make `showMessages` in BaseForm more generic.
Don't remember we have specific reason why only show message as text in BaseForm.
cc @pradeepdewda-okta @nikhilvenkatraman-okta

pradeepdewda-okta: @haishengwu-okta i loosely recall we had a long discussion about this. Earlier showMessages in the baseForm had the logic to show errors and warnings. But then @nikhilvenkatraman-okta had some different use case and then we decided to overwrite this method for the views where they needs to show these something differently. Don't remember why..  this was the PR https://github.com/okta/okta-signin-widget/pull/1274

playground/mocks/data/idp/idx/authenticator-verification-okta-verify-reject-push.json
gowthamidommety-okta: @ManuMalhotra-okta can confirm, but this may be for any errors. Because on any error the user should have the option to resend push

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
```diff
@@ -0,0 +1,45 @@
...
+    showMessages () {
+      // override showMessages to display error in cases like reject push
+      // or timeout. Borrowed this logic from TerminalView.
+      const messagesObjs = this.options.appState.get('messages');
+      if (messagesObjs && Array.isArray(messagesObjs.value)) {
+        this.add('<div class="ion-messages-container"></div>', '.o-form-error-container');
+
+        messagesObjs.value.forEach(messagesObj => {
+          const msg = messagesObj.message;
+          if (messagesObj?.class === 'ERROR') {
```
anipendakur-okta, line 27: `nit:` On a second thought, is it possible for `messagesObjs.value` to have falsy value like this `[{class: 'ERROR', message: 'error msg'}, undefined]` ? If it's empty (`[]`), it's taken care of by `forEach`. 
If the above condition holds good, this can indeed by simplified as -
```
const {message: msg, class} = messagesObj;

// class can be undefined if not present, but it is taken care of by ===
if (class === 'ERROR') {
.
.
```
This is just a `nit`, I'm good if it's not done as well.

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
magizh-okta: Seems like the reason why @nikhilvenkatraman-okta  had to override is to reduce the number of messages displayed. This is not he case here. We just can enhance the showMessages in base form. @gowthamidommety-okta  can we create a separate jira for this?

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
gowthamidommety-okta: 👍  Will handle this in the JIRA that takes care of messages on baseform

PR #1399, opened 2020-08-27
Update CookieUtil.js for SameSite attribute

i-e-b:
## Description:
Set an explicit `samesite` property to prevent warnings in Chrome and Edge.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

**Note: this page is not available without login**

### Screenshot/Video:
*N/A*


swiftone:
@i-e-b - Thanks for the report.  Actually, all the cookies set by the widget already have the samesite property correctly set.  The warnings come about because of a call where we (attempt) to send the cookies unnecessarily to the backend.  

For now the warnings post no functional nor security issue, but we plan on removing the cause anyway (in fact, the ticket has recently been assigned, though it still has some items in front of it)

We have an issue tracking this here: https://github.com/okta/okta-auth-js/issues/347

swiftone:
@i-e-b - Sorry, I just checked your PR and see you've found a spot we missed.  Thanks!

PR #1430, opened 2020-09-16
feat: redirect to error page in mfa only flow

gowthamidommety-okta:
- adds new feature to handle UX for mfa only flows
- also changes an existing session expired error to say "sign in" for consistency

Resolves: OKTA-329312

## Description:

Screen Shot 2020-09-15 at 7 51 58 PM

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-329312](https://oktainc.atlassian.net/browse/OKTA-329312)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -42,7 +42,7 @@ error.unsupported.browser = Unsupported browser
 error.unsupported.cors = Unsupported browser - missing CORS support
 error.unsupported.localStorage = Unsupported browser - missing localStorage support
 error.enabled.cors = There was an error sending the request - have you enabled CORS?
-error.expired.session = Your session has expired. Please try to log in again.
+error.expired.session = Your session has expired. Please try to sign in again.
```
gowthamidommety-okta, line 45: Please let me know what you think of this change. Guidance from UX is to always say `sign in` not `log in`. This breaks some selenium tests. Will create another PR to fix those.

src/util/RouterUtil.js
```diff
@@ -107,8 +107,13 @@ fn.routeAfterAuthStatusChangeError = function (router, err) {
   // Token has expired - no longer valid. Navigate back to primary auth.
   if (err.errorCode === ErrorCodes.INVALID_TOKEN_EXCEPTION) {
     router.appState.set('flashError', err);
-    router.controller.state.set('navigateDir', Enums.DIRECTION_BACK);
-    router.navigate('', { trigger: true });
+    if (router.settings.get('features.isMfaOnlyFlow')) {
+      router.controller.state.set('navigateDir', Enums.DIRECTION_BACK);
```
mauriciocastillosilva-okta, line 111: this line is common for both, so we can probably move it outside of the `if` block

packages/@okta/i18n/src/properties/login.properties
mauriciocastillosilva-okta: This change is good, so we should keep it

With that said, you're adding a new controller just for mfa invalid session error, so this error might not be the best. The user can't really go and sign in again since we landed in a dead-end error page without links. Maybe we should have a different error message? Something generic like
_"Error validating MFA. Please try again"_ (not sign in specifically, but whatever it is they were trying)
What do you think?

src/LoginRouter.js
```diff
@@ -126,6 +127,7 @@ export default BaseLoginRouter.extend({
     'signin/refresh-auth-state(/:token)': 'refreshAuthState',
     'signin/register': 'register',
     'signin/register-complete': 'registerComplete',
+    'signin/mfa-invalid-session': 'mfaInvalidSession',
```
mauriciocastillosilva-okta, line 130: This is good, but what do you think if we make this more generic? Something like `signin/error` so this route can be reused if we ever need to. The controller is just the most basic controller that shows errors, nothing else... 
Thoughts? We can also cross this bridge once we need to (if we ever need to)

gowthamidommety-okta: That actually makes sense, and it was something i was thinking about.. I'll make this change

packages/@okta/i18n/src/properties/login.properties
gowthamidommety-okta: I think it's ok to keep the same message. mfa is also part of signing in. They would need to close the window / refresh to restart the flow. what do you think of that experience?

haishengwu-okta (approved):
LGTM after address Mau's comments

packages/@okta/i18n/src/properties/login.properties
magizh-okta: @gowthamidommety-okta  when ever we update the text we need a new key.

src/MfaOnlyInvalidSessionController.js
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import FormController from 'util/FormController';
+export default FormController.extend({
+  className: 'mfa-invalid-session',
+  Model: function () {},
```
magizh-okta, line 16: Can we just pass the backbone. model itself.

src/util/RouterUtil.js
```diff
@@ -107,8 +107,13 @@ fn.routeAfterAuthStatusChangeError = function (router, err) {
   // Token has expired - no longer valid. Navigate back to primary auth.
   if (err.errorCode === ErrorCodes.INVALID_TOKEN_EXCEPTION) {
     router.appState.set('flashError', err);
-    router.controller.state.set('navigateDir', Enums.DIRECTION_BACK);
-    router.navigate('', { trigger: true });
+    if (router.settings.get('features.isMfaOnlyFlow')) {
```
magizh-okta, line 110: nit : just keep the navigate inside the if since we do 
      router.controller.state.set('navigateDir', Enums.DIRECTION_BACK);

 in both

packages/@okta/i18n/src/properties/login.properties
magizh-okta: Also this text doesnt make sense in MFA only flow. What can the user do at this point just close the modal and retry ?

mauriciocastillosilva-okta: why do we need a new key?

And I still would think it is better to have a different error for mfa only flows... can we double check with UX. If they are good then I will say no more

gowthamidommety-okta: I already checked with @vincentvoong-okta but let me check again.

vincentvoong-okta: @gowthamidommety-okta @mauriciocastillosilva-okta 
Initially I thought we had to reuse an existing error message. It sounds like this proposal is going to be a new message - I think that is fine so long as this message only appears in an MFA dead-end state. I recommend "Unable to authenticate at this time. Try again.".

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -43,6 +43,7 @@ error.unsupported.cors = Unsupported browser - missing CORS support
 error.unsupported.localStorage = Unsupported browser - missing localStorage support
 error.enabled.cors = There was an error sending the request - have you enabled CORS?
 error.expired.session = Your session has expired. Please try to log in again.
+error.mfa.only.expired.session = Unable to authenticate at this time. Try again.
```
magizh-okta, line 46: Please run this by UX to see if they can come up with something better. 
'Try again.'   how ?  there is no button not text on what are the next steps.

magizh-okta (commented):
LGTM except the text.

packages/@okta/i18n/src/properties/login.properties
gowthamidommety-okta: How is the hard part. In RDP we close the window, in end user settings, we need to go back. There is no sure way of telling a user how to try again. Maybe we should just say `Unable to authenticate at this time.`? @mauriciocastillosilva-okta @vincentvoong-okta thoughts?

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -43,6 +43,7 @@ error.unsupported.cors = Unsupported browser - missing CORS support
 error.unsupported.localStorage = Unsupported browser - missing localStorage support
 error.enabled.cors = There was an error sending the request - have you enabled CORS?
 error.expired.session = Your session has expired. Please try to log in again.
```
mauriciocastillosilva-okta, line 45: I guess this was lost in the previous comment, I'm 100% with changing this to `Please try to sign in again.`. Also good if you prefer to do it in another jira/PR since that involves selenium tests as well

packages/@okta/i18n/src/properties/login.properties
mauriciocastillosilva-okta: yeah, it is hard to come up with something accurate that can help, since it really is a dead end error.

We will show this because of an InvalidToken error, which just literally means that: the token is no longer valid. In the regular sign in flow, the user can go back and start over, so we can show a next step to try to start login flow again. However with `mfaOnlyFlow` the user needs a new token to start over, and we don't know **how** they can get that token. Leaving RDP aside, since for SIW purposes we shouldn't think of a specific usage of the widget, I think the error should just tell the user what happened. Maybe just `Your token is invalid.`, or if we follow the same pattern as the regular flows, this will be something like `Unable to validate your credentials at this time.`. If we try to be clever about next steps, then I think what Gowthami has is the best guess we can have, since we don't exactly know where they come from. 

With that said, I think @vincentvoong-okta should make the call

mauriciocastillosilva-okta (approved):
Just confirm the error message with UX, rest of the changes look good 🐑

packages/@okta/i18n/src/properties/login.properties
gowthamidommety-okta: yes, sorry. I thought i would leave this as is for now.. but maybe it is better to change this to `sign in` in this PR itself.

magizh-okta (approved):

PR #1451, opened 2020-10-07
Fix for terminal view in case of expired email link

VadymYaremchuk-okta:
## Description:

Show **Back to signin** link for expired email link terminal view

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://oktainc.atlassian.net/secure/attachment/374712/ExpiredEmailLinkBackToSigninDemo.mov

### Reviewers:

@bellatumaneng-okta 
@haishengwu-okta 

### Issue:

- [OKTA-318058](https://oktainc.atlassian.net/browse/OKTA-318058)

### DIFF
--- a/src/v2/models/AppState.js
+++ b/src/v2/models/AppState.js
@@ -13,7 +13,7 @@
 import { _, Model } from 'okta';
 import Logger from 'util/Logger';
 import { FORMS_WITHOUT_SIGNOUT, FORMS_WITH_STATIC_BACK_LINK,
-  FORMS_FOR_VERIFICATION } from '../ion/RemediationConstants';
+  FORMS_FOR_VERIFICATION, FORMS } from '../ion/RemediationConstants';
 
 /**
  * Keep track of stateMachine with this special model. Similar to `src/models/AppState.js`
@@ -151,6 +151,12 @@ export default Model.extend({
     return !hideSignOutConfigOverride
       && _.isFunction(idxActions?.cancel)
       && !FORMS_WITHOUT_SIGNOUT.includes(currentFormName);
+  },
+
+  containsMessageWithI18nKey (key) {
+    const messagesObjs = this.get('messages');
+    return messagesObjs && Array.isArray(messagesObjs.value)
+      && messagesObjs.value.some(messagesObj => messagesObj.i18n?.key === key);
   },
 
   setIonResponse (transformedResponse) {
--- a/src/v2/view-builder/utils/LinksUtil.js
+++ b/src/v2/view-builder/utils/LinksUtil.js
@@ -113,9 +113,22 @@ const getSignOutLink = (settings) => {
   }
 };
 
+const getBackToSignInLink = (settings) => {
+  let baseUrl = settings?.get('baseUrl') ? settings.get('baseUrl') : window.location.hostname;
+  return [
+    {
+      'type': 'link',
+      'label': loc('backToSignin', 'login'),
+      'name': 'go-back',
+      'href': baseUrl + '/login/login.htm'
+    },
+  ];
+};
+
 export {
   getSwitchAuthenticatorLink,
   getForgotPasswordLink,
   goBackLink,
-  getSignOutLink
+  getSignOutLink,
+  getBackToSignInLink
 };
--- a/src/v2/view-builder/views/TerminalView.js
+++ b/src/v2/view-builder/views/TerminalView.js
@@ -1,12 +1,24 @@
-import { createCallout } from 'okta';
+import { createCallout, loc } from 'okta';
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
+import BaseFooter from '../internals/BaseFooter';
+import {getBackToSignInLink} from '../utils/LinksUtil';
+
+const RETURN_LINK_EXPIRED_KEY = 'idx.return.link.expired';
 
 const Body = BaseForm.extend({
   noButtonBar: true,
+
   postRender () {
     BaseForm.prototype.postRender.apply(this, arguments);
     this.$el.addClass('terminal-state');
+  },
+
+  title () {
+    if (this.options.appState.containsMessageWithI18nKey(RETURN_LINK_EXPIRED_KEY)) {
+      return loc('oie.email.return.link.expired.title', 'login');
+    }
+    return BaseForm.prototype.title.apply(this, arguments);
   },
 
   showMessages () {
@@ -17,7 +29,7 @@ const Body = BaseForm.extend({
       messagesObjs.value
         .forEach(messagesObj => {
           const msg = messagesObj.message;
-          if (messagesObj.class === 'ERROR') {
+          if (messagesObj.class === 'ERROR' || messagesObj.i18n?.key === RETURN_LINK_EXPIRED_KEY) {
             this.add(createCallout({
               content: msg,
               type: 'error',
@@ -32,6 +44,15 @@ const Body = BaseForm.extend({
 
 });
 
+const Footer = BaseFooter.extend({
+  links: function () {
+    if (this.options.appState.containsMessageWithI18nKey(RETURN_LINK_EXPIRED_KEY)) {
+      return getBackToSignInLink(this.options.settings);
+    }
+  }
+});
+
 export default BaseView.extend({
-  Body
+  Body,
+  Footer
 });
--- a/test/testcafe/spec/ChallengeFactorEmailMagicLink_spec.js
+++ b/test/testcafe/spec/ChallengeFactorEmailMagicLink_spec.js
@@ -43,6 +43,8 @@ test
     await setup(t);
     const terminalPageObject = await new TerminalPageObject(t);
     await t.expect(terminalPageObject.getMessages()).eql('Please return to the original tab.');
+    await t.expect(await terminalPageObject.goBackLinkExists()).notOk();
+    await t.expect(terminalPageObject.getFormTitle()).eql('Authenticate');
   });
 
 test
@@ -50,13 +52,18 @@ test
     await setup(t);
     const terminalPageObject = await new TerminalPageObject(t);
     await t.expect(terminalPageObject.getMessages()).eql('Flow continued in a new tab.');
+    await t.expect(await terminalPageObject.goBackLinkExists()).notOk();
+    await t.expect(terminalPageObject.getFormTitle()).eql('Authenticate');
   });
 
 test
   .requestHooks(magicLinkExpiredMock)('challenge email factor with expired magic link', async t => {
     await setup(t);
     const terminalPageObject = await new TerminalPageObject(t);
-    await t.expect(terminalPageObject.getMessages()).eql('This email link has expired. To resend it, return to the screen where you requested it.');
+    await t.expect(terminalPageObject.getErrorMessages().isError()).eql(true);
+    await t.expect(terminalPageObject.getErrorMessages().getTextContent()).eql('This email link has expired. To resend it, return to the screen where you requested it.');
+    await t.expect(await terminalPageObject.goBackLinkExists()).ok();
+    await t.expect(terminalPageObject.getFormTitle()).eql('Verify with your email');
   });
 
 test
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -741,6 +741,7 @@ oie.email.authenticator.description = Verify with a link or code sent to your em
 oie.email.mfa.title = Verify with your email
 oie.email.verify.sentText = An email was sent to
 oie.email.enroll.subtitle = Please check your email and enter the code below.
+oie.email.return.link.expired.title = Verify with your email
 
 ## Webauthn
 oie.authenticator.webauthn.label = Security Key or Biometric Authenticator
--- a/test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
+++ b/test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
@@ -191,5 +191,8 @@ test
   .requestHooks(magicLinkExpiredMock)('challenge email factor with expired magic link', async t => {
     await setup(t);
     const terminalPageObject = new TerminalPageObject(t);
-    await t.expect(terminalPageObject.getMessages()).eql('This email link has expired. To resend it, return to the screen where you requested it.');
+    await t.expect(terminalPageObject.getErrorMessages().isError()).eql(true);
+    await t.expect(terminalPageObject.getErrorMessages().getTextContent()).eql('This email link has expired. To resend it, return to the screen where you requested it.');
+    await t.expect(await terminalPageObject.goBackLinkExists()).ok();
+    await t.expect(terminalPageObject.getFormTitle()).eql('Verify with your email');
   });


src/v2/view-builder/views/TerminalView.js
```diff
@@ -9,6 +11,16 @@ const Body = BaseForm.extend({
     this.$el.addClass('terminal-state');
   },
 
+  title () {
+    const messagesObjs = this.options.appState.get('messages');
+    if (messagesObjs && Array.isArray(messagesObjs.value)) {
+      if (messagesObjs.value.some(messagesObj => messagesObj.i18n?.key === 'idx.return.link.expired')) {
+        return 'Verify with your email';
```
haishengwu-okta, line 18: create an i18n key instead of hard code English.

src/v2/view-builder/views/TerminalView.js
```diff
@@ -17,7 +29,7 @@ const Body = BaseForm.extend({
       messagesObjs.value
         .forEach(messagesObj => {
           const msg = messagesObj.message;
-          if (messagesObj.class === 'ERROR') {
+          if (messagesObj.class === 'ERROR' || messagesObj.i18n?.key === 'idx.return.link.expired') {
```
haishengwu-okta, line 32: @pradeepdewda-okta I think we shall make the error handling more generic like

```
loop all message
  if is error then add error callout
  if is info then add infor callout
  etc..
```

src/v2/view-builder/utils/LinksUtil.js
```diff
@@ -113,9 +113,26 @@ const getSignOutLink = (settings) => {
   }
 };
 
+const getDefaultBackSignInLink = (settings, appState) => {
+  const messagesObjs = appState?.get('messages');
+  if (messagesObjs && Array.isArray(messagesObjs.value)
+    && messagesObjs.value.some(messagesObj => messagesObj.i18n?.key === 'idx.return.link.expired')) {
```
anipendakur-okta, line 119: Instead of using this check, i'd recommend using a separate remediation object for this. Perhaps work with backend on it? Any thoughts @haishengwu-okta ?

anipendakur-okta:
Also, if you're aiming for this week's patch, please rebase it to `4.5`.

src/v2/view-builder/views/TerminalView.js
VadymYaremchuk-okta: with is the translation process? or we can keep only english for now?

src/v2/view-builder/views/TerminalView.js
pradeepdewda-okta: @haishengwu-okta I had that in mind, i will create a ticket for this.

src/v2/view-builder/utils/LinksUtil.js
```diff
@@ -113,9 +113,26 @@ const getSignOutLink = (settings) => {
   }
 };
 
+const getDefaultBackSignInLink = (settings, appState) => {
```
pradeepdewda-okta, line 116: we should rename this just to `geBackToSignInLink` ?

src/v2/view-builder/utils/LinksUtil.js
```diff
@@ -113,9 +113,26 @@ const getSignOutLink = (settings) => {
...
+  const messagesObjs = appState?.get('messages');
+  if (messagesObjs && Array.isArray(messagesObjs.value)
+    && messagesObjs.value.some(messagesObj => messagesObj.i18n?.key === 'idx.return.link.expired')) {
+    return [
+      {
+        'type': 'link',
+        'label': loc('goback', 'login'),
```
pradeepdewda-okta, line 123: should we use `backToSignin` instead of `goback` ? Because no need to have `Sign In` in uppecases

src/v2/view-builder/utils/LinksUtil.js
haishengwu-okta: It's actually good idea but not sure if it might be overkill to have separate remediation object for this.
Worth to check with IdX team though @VadymYaremchuk-okta

src/v2/view-builder/views/TerminalView.js
VadymYaremchuk-okta: seems I can reuse exist one
```
oie.email.mfa.title = Verify with your email
```

haishengwu-okta: hm.. `oie.email.mfa.title` is intent for email factor page.
why you need to rename the title?

jmelberg-okta: @VadymYaremchuk-okta - When you add a new entry to `login.properties`, it'll be picked up by our i18n tooling to request translations through our l10n vendor. Most of this is automated - so the action for devs is pretty straightforward.

Also, instead of reusing an existing property key - I'd suggest creating a new key that better matches with this view. For example `oie.return.link.expired.title`.

jmelberg-okta: Also - is this text settled? I'm not sure an enduser will understand what we're asking them to verify in this flow.

src/v2/view-builder/utils/LinksUtil.js
VadymYaremchuk-okta: According to our decision ```We are going to revisit the use cases for Embedded Deployment Models will be explored and speced later``` we may keep the fix simple for now and then update it with remediation object later if we need to.

Btw, are we expecting by design to have remediation object in terminal state? Since I supposed that remediation can be present only while we are not in terminal state.

src/v2/view-builder/views/TerminalView.js
VadymYaremchuk-okta: Sure, can create new i18n key if needed.

The title I took according the screenshot Vince provided me with https://oktawiki.atlassian.net/wiki/spaces/eng/pages/1281525702/OIE-AppEmbed-+POD+Deployment+Models+Wiki?preview=/1281525702/1345359732/image%20(7).png

Before the title was `Authenticate` (default one). I can keep the old one.

The error message itself is controlled by backend. 

What is the process of i18n translation?

jmelberg-okta: Full process is described [here](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/741293778/i18n+dev+process+From+Oct+2019+onwards).

VadymYaremchuk-okta: seems to simplify this we can keep the currently exist title `Authenticate`.
@vincentvoong-okta @wdawson can you please confirm that it is OK to keep the exist `Authenticate` title for terminal view in case of expired email link used?

wdawson: @VadymYaremchuk-okta can you include a screenshot for the page you're talking about? I want to make sure I understand the context.

VadymYaremchuk-okta: @wdawson 
screen with the issue:
Screen Shot 2020-10-07 at 4 07 01 PM
screen with the fix based on UI mock:
Screen Shot 2020-10-07 at 4 07 47 PM

VadymYaremchuk-okta: UI mock received from Vince:
image (7)

wdawson: Thanks, sounds good.

src/v2/view-builder/views/TerminalView.js
```diff
@@ -1,14 +1,32 @@
-import { createCallout } from 'okta';
+import { createCallout, loc } from 'okta';
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
+import BaseFooter from '../internals/BaseFooter';
+import {getBackToSignInLink} from '../utils/LinksUtil';
 
 const Body = BaseForm.extend({
   noButtonBar: true,
+  isTerminalForExpiredLink: false,
```
haishengwu-okta, line 9: probably better to have a helper function (in separated file) which take `messagesObjs` and figure out whether it's a email link expired therefore can be reused even when creating the go back link.

src/v2/view-builder/views/TerminalView.js
```diff
@@ -32,6 +50,13 @@ const Body = BaseForm.extend({
 
 });
 
+const Footer = BaseFooter.extend({
+  links: function () {
+    return getBackToSignInLink(this.options.settings, this.options.appState);
```
haishengwu-okta, line 55: prefer
```
if is email expired link error
  return getBackToSignInLink(this.options.settings);
else return [];

File: src/v2/view-builder/views/TerminalView.js (hunk 6 above)
Comment (pradeepdewda-okta) [inline src/v2/view-builder/views/TerminalView.js:9]: TerminalView may contain different other links as well in future.. So instead of having a specific boolean here we can change this into an object that can contain any type of links.
`terminalLinks : {}

and later initialize accordingly
this.terminalLinks['expiredLinks'] = <some statement>

and use it
if(this.terminalLInks['expireLinks'])
`

File: test/testcafe/spec/ChallengeFactorEmailMagicLink_spec.js [hunk 8]
```diff
@@ -43,20 +43,27 @@ test
     await setup(t);
     const terminalPageObject = await new TerminalPageObject(t);
     await t.expect(terminalPageObject.getMessages()).eql('Please return to the original tab.');
+    await t.expect(await terminalPageObject.goBackLinkExists()).notOk();
```
haishengwu-okta, line 46: `ChallengeFactorEmailMagicLink_spec.js` is deprecated (test with `Factor` name are legacy) . are you seeing error because of your change?

src/v2/view-builder/views/TerminalView.js
```diff
@@ -1,14 +1,32 @@
...
+import BaseFooter from '../internals/BaseFooter';
+import {getBackToSignInLink} from '../utils/LinksUtil';
 
 const Body = BaseForm.extend({
   noButtonBar: true,
+  isTerminalForExpiredLink: false,
+
+  initialize () {
+    const messagesObjs = this.options.appState.get('messages');
+    this.isTerminalForExpiredLink = messagesObjs && Array.isArray(messagesObjs.value)
```
pradeepdewda-okta, line 13: change this check to a function that can return a boolean

```
toDisplayLink(key) {
 return messagesObjs && Array.isArray(messagesObjs.value) &&
  && messagesObjs.value.some(messagesObj => messagesObj.i18n?.key === 'idx.return.link.expired');
}

File: src/v2/view-builder/views/TerminalView.js [hunk 10]
```diff
@@ -1,14 +1,32 @@
...
+      && messagesObjs.value.some(messagesObj => messagesObj.i18n?.key === 'idx.return.link.expired');
+    BaseForm.prototype.initialize.apply(this, arguments);
```
pradeepdewda-okta, line 15: We always call baseform Initialize method first and then override other stuff.. If you are unable to do that then you have to find some other way to do it.

src/v2/view-builder/views/TerminalView.js
VadymYaremchuk-okta: that was a trade of between calling function (computing every time) versus compute once and keep it in variable.

test/testcafe/spec/ChallengeFactorEmailMagicLink_spec.js
VadymYaremchuk-okta: bacon is green https://bacon-go.aue1e.saasure.net/#!/commits/okta-signin-widget/sha/2b215450a475c9060ebae4e56b0fd92ab343a7ed, but wasn't when I have bug in my change.

src/v2/view-builder/views/TerminalView.js
VadymYaremchuk-okta: I will not need this since the advices from above is to use method instead of `isTerminalForExpiredLink ` variable.

test/testcafe/spec/ChallengeFactorEmailMagicLink_spec.js
```diff
@@ -43,20 +43,27 @@ test
...
 
 test
   .requestHooks(magicLinkExpiredMock)('challenge email factor with expired magic link', async t => {
     await setup(t);
     const terminalPageObject = await new TerminalPageObject(t);
-    await t.expect(terminalPageObject.getMessages()).eql('This email link has expired. To resend it, return to the screen where you requested it.');
+    await t.expect(terminalPageObject.getErrorMessages().isError()).eql(true);
+    await t.expect(terminalPageObject.getErrorMessages().getTextContent()).eql('This email link has expired. To resend it, return to the screen where you requested it.');
+    await t.expect(await terminalPageObject.goBackLinkExists()).ok();
+    await t.expect(terminalPageObject.getFormTitle()).eql('Verify with your email');
```
jmelberg-okta, line 66: Can we also add a test to validate when `isTerminalForExpiredLink` isn't provided this falls back to the `BaseForm` title?

test/testcafe/spec/ChallengeFactorEmailMagicLink_spec.js
haishengwu-okta: I didn't follow.

test/testcafe/spec/ChallengeFactorEmailMagicLink_spec.js
VadymYaremchuk-okta: we already doing it with `await t.expect(terminalPageObject.getFormTitle()).eql('Authenticate');` in other tests from this class that doesn't expect to have expired link error present.

test/testcafe/spec/ChallengeFactorEmailMagicLink_spec.js
VadymYaremchuk-okta: I am not sure why you asked about errors and what errors did you mean - I don't see any errors with the bacon recent build.
Please be specific about the class/place if I need to move tests to another place.

src/v2/view-builder/views/TerminalView.js
VadymYaremchuk-okta: we can add those later once have it, currently added single method that checks the state

test/testcafe/spec/ChallengeFactorEmailMagicLink_spec.js
VadymYaremchuk-okta: for example this build https://bacon-go.aue1e.saasure.net/#!/commits/okta-signin-widget/sha/a4f274a5538ae2d5f8c97f8b579576125d929863 was failing because of the bugs in tests including `ChallengeFactorEmailMagicLink_spec.js`

src/v2/view-builder/views/TerminalView.js
```diff
@@ -1,14 +1,24 @@
...
   postRender () {
     BaseForm.prototype.postRender.apply(this, arguments);
     this.$el.addClass('terminal-state');
   },
 
+  title () {
+    if (this.options.appState.isExpiredEmailLinkTerminalState()) {
+      return loc('oie.email.return.link.expired.title', 'login');
+    }
+    return BaseForm.prototype.title();
```
haishengwu-okta, line 19: `BaseForm.prototype.title.apply(this, arguments);`

src/v2/view-builder/views/TerminalView.js
haishengwu-okta: reuse the `this.options.appState.isExpiredEmailLinkTerminalState`;
you could declare a variable at the beginning of this method (`showMessages`) if you want to avoid duplicated compute.

src/v2/models/AppState.js
```diff
@@ -153,6 +153,12 @@ export default Model.extend({
       && !FORMS_WITHOUT_SIGNOUT.includes(currentFormName);
   },
 
+  isExpiredEmailLinkTerminalState () {
```
pradeepdewda-okta, line 156: if we can pass a `key` as argument then we can make this function generalized. I won't prefer a function that does only one thing. Please use an argument here.

src/v2/view-builder/views/TerminalView.js
pradeepdewda-okta: Let's fix it in this PR.

src/v2/view-builder/views/TerminalView.js
VadymYaremchuk-okta: will change, but I'm curious why thats better to use?

haishengwu-okta: `BaseForm.prototype.title` may mess up `this` context.

haishengwu-okta (approved):

src/v2/view-builder/views/TerminalView.js
pradeepdewda-okta: Basically calling the base method we are ignoring the current context of this view, If we had some options that needed to be propagated they won't go if we don't apply the context..

pradeepdewda-okta (approved):
@VadymYaremchuk-okta thanks for making these changes.. 

looks good. 💯

PR #1454, opened 2020-10-08
Adding translations for locked out error

bibinmichael:
## Description:
Updating message for lock account lock out scenario in Nike.net.


robertjd:
Hello @bibinmichael , modifying this file here would change these strings for all companies that are using the widget, so I don't think we can merge this.  Are you maintaining your own fork of the widget for your needs?

PR #1461, opened 2020-10-15
fix: allow i18nKey to be passed to the custom button title for i18n

shawyu-okta:
## Description:

* Allows for configuring each `customButtons` button with an additional option, `i18nKey` instead of `title`
* The custom `i18nKey` can be set using the `i18n` config option to pass translations for each button
* If both a `title` and `i18nKey` are specified, `title` is used

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Using a minimal configuration:
```js
module.exports = {
  baseUrl: 'https://dev1.okta1.com',
  i18n: {
    en: {
      'my.custom.i18n.key': 'hello world',
    },
  },
  customButtons: [{
    i18nKey: 'my.custom.i18n.key',
  }],
};
```

image

### Issue:

- [OKTA-329120](https://oktainc.atlassian.net/browse/OKTA-329120)

### DIFF
--- a/README.md
+++ b/README.md
@@ -783,9 +783,26 @@ customButtons: [{
     window.location.href = 'http://www.example.com';
   }
 }]
+
+// An example that adds a custom button with a localized title underneath the login form on the primary auth page
+i18n: {
+  en: {
+    'customButton.title': 'Custom Button Title',
+  },
+},
+customButtons: [{
+  i18nKey: 'customButton.title',
+  className: 'btn-customAuth',
+  click: function() {
+    // clicking on the button navigates to another page
+    window.location.href = 'http://www.example.com';
+  }
+}]
 ```
 
-- **customButtons.title** - String that is set as the button text
+- **customButtons.title** - String that is set as the button text (set only one of `title` OR `i18nKey`)
+
+- **customButtons.i18nKey** - Custom translation key for button text specified in `i18n` config option (set only one of `title` OR `i18nKey`)
 
 - **customButtons.className** - Optional class that can be added to the button
 
--- a/src/views/primary-auth/CustomButtons.js
+++ b/src/views/primary-auth/CustomButtons.js
@@ -110,7 +110,9 @@ export default View.extend({
         'data-se': options.dataAttr,
       },
       className: options.className + ' default-custom-button',
-      title: options.title,
+      title: function () {
+        return options.title || loc(options.i18nKey);
+      },
       click: options.click,
     });
   },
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -785,7 +785,9 @@ customButtons: [{
 }]
 ```
 
-- **customButtons.title** - String that is set as the button text
+- **customButtons.title** - String that is set as the button text (set only one of `title` OR `i18nKey`)
```
haishengwu-okta, line 788: update sample above for using i18nKey

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -3083,6 +3083,24 @@ Expect.describe('Additional Auth Button', function () {
       expect(test.form.additionalAuthButton().hasClass('new-class')).toBe(true);
     });
   });
+  itp('displays custom button translation', function () {
```
haishengwu-okta, line 3086: another test case when both title or i18nKey is passed, i18nKey will be ignored.

src/views/primary-auth/CustomButtons.js
```diff
@@ -110,7 +110,9 @@ export default View.extend({
         'data-se': options.dataAttr,
       },
       className: options.className + ' default-custom-button',
-      title: options.title,
+      title: function() {
```
pradeepdewda-okta, line 113: can you also check if it works as expected with idp_discovery flow

haishengwu-okta (commented):
Looks great! thanks1

src/views/primary-auth/CustomButtons.js
shawyu-okta: I'm new to this widget -- do you mean a test in the same `PrimaryAuth_spec` file that has `settings` that includes both `customButtons` and also `idps` config?

README.md
shawyu-okta: I've added a separate example without `title` but including `i18nKey` to avoid a confusing configuration where both are included.

src/views/primary-auth/CustomButtons.js
shawyu-okta: @pradeepdewda-okta I've added the same tests in the `IDPDiscovery_spec` test file

README.md
```diff
@@ -783,9 +783,21 @@ customButtons: [{
     window.location.href = 'http://www.example.com';
   }
 }]
+
+// An example that adds a custom button with a localized title underneath the login form on the primary auth page
+customButtons: [{
```
haishengwu-okta, line 788: what about this? so that developer know they must specify i18n

```
i18n: {
  en: {..}
},
customButtons: [
  {...}
]

```

haishengwu-okta (commented):
one comment on doc

pradeepdewda-okta (approved):

README.md
shawyu-okta: Good idea, I've added the `i18n` field in the README for this example

haishengwu-okta (approved):

PR #1467, opened 2020-10-21
fix: adds logic to translate duplicate email error on registration

### LINKED ISSUE #1062: Config self-existing registration property does affect error text for existing email
When I try to modify either of the following config properties before rendering the widget:

config.i18n.en["registration.error.userName.notUniqueWithinOrg"]
config.i18n.en["errors.E0000114"]

In order to modify the warning when self-service registering user enters an existing email, the error message is not affected and always states "A user with this Email already exists":

image

It appears that text is only affected by modifying the self-service username field label in the self-service properties set by an administrator. So the error could say for example: "A user with this [editable text] already exists".

https://github.com/okta/okta-signin-widget/blob/9dd46661a7d05e88507671cdba0a757b962c8315/packages/%40okta/i18n/src/properties/login.properties#L607

shawyu-okta:
## Description:

* Adds error handling logic to handle cases where the user's email address already exists for an organization during registration and localizes it for the widget
* Allows the above error message to be overridden using `i18n` configuration
* Closes #1062.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Using this `.widgetrc.js` config:
```js
module.exports = {
  baseUrl: 'http://dev1.okta1.com:1802',
  logoText: 'Okta',
  features: {
    registration: true,
  },
  i18n: {
    en: {
      'registration.error.userName.notUniqueWithinOrg': 'custom i18n',
    }
  }
};
```
image

If no custom i18n key is added for this error (default behavior):
```js
module.exports = {
  baseUrl: 'http://dev1.okta1.com:1802',
  logoText: 'Okta',
  features: {
    registration: true,
  },
};
```
image

### Issue:

- [OKTA-280970](https://oktainc.atlassian.net/browse/OKTA-280970)

### DIFF
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -198,6 +198,19 @@ export default BaseLoginController.extend({
         noCancelButton: true,
         title: loc('registration.form.title', 'login'),
         save: loc('registration.form.submit', 'login'),
+        parseErrorMessage: function (resp) {
+          const hasErrorCauses = resp.errorCauses && resp.errorCauses.length;
+
+          if (hasErrorCauses) {
+            const isDuplicateEmailCause = resp.errorCode === 'E0000001'
+              && resp.errorCauses[0].reason === 'UNIQUE_CONSTRAINT';
+
+            if (isDuplicateEmailCause) {
+              resp.errorCauses[0].errorSummary = loc('registration.error.userName.notUniqueWithinOrg', 'login');
+            }
+          }
+          return resp;
+        },
       });
       const form = new RegistrationControllerForm(self.toJSON());
 
--- a/test/unit/helpers/xhr/ERROR_notUnique.js
+++ b/test/unit/helpers/xhr/ERROR_notUnique.js
@@ -1,0 +1,18 @@
+export default {
+  status: 400,
+  responseType: 'jrd+json',
+  response: {
+    errorCode: 'E0000001',
+    errorSummary: 'Api validation failed: null',
+    errorLink: 'E0000001',
+    errorId: 'oaeXbQp2gytRhijPgA-Rew56g',
+    errorCauses: [
+      {
+        domain: 'registration request',
+        errorSummary: 'A user with this Email already exists',
+        locationType: 'body',
+        reason: 'UNIQUE_CONSTRAINT',
+      },
+    ],
+  },
+};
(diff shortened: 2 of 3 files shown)


src/RegistrationController.js
```diff
@@ -198,6 +198,19 @@ export default BaseLoginController.extend({
...
         save: loc('registration.form.submit', 'login'),
+        parseErrorMessage: function (resp) {
+          const hasErrorCauses = resp.errorCauses && resp.errorCauses.length;
+
+          if (hasErrorCauses) {
+            const isDuplicateEmailCause = resp.errorCode === 'E0000001'
+              && resp.errorCauses[0].reason === 'UNIQUE_CONSTRAINT';
+
+            if (isDuplicateEmailCause) {
+              resp.errorCauses[0].errorSummary = loc('registration.error.userName.notUniqueWithinOrg', 'login');
```
shawyu-okta, line 209: One use-case we need to cover is allowing developers to override this message. I think using the `login` bundle may prevent that, though. Is there a better way to allow this?

shawyu-okta: I wanted to use the existing translations that exist on the widget but perhaps we need to move this outside.

haishengwu-okta: Why would you think using `login` bundle would prevent this. It's been used all over the place.

shawyu-okta: I have the impression that the login bundle cannot be overridden through the user configuration, is that not the case?

haishengwu-okta: no. it meant to be override currently by the `i18n` config.

haishengwu-okta (approved):
Add test & 🚀

shawyu-okta:
FYI @swiftone this was merged into `4.5`

c-droege:
Hello, 
the issue is already closed for a while but we're facing a similiar problem. When we customize the translation for:

`registration.error.userName.notUniqueWithinOrg`

to be something like 'Unfortunately we can't fulfill your request' to hide the information for an attacker that an account with this email already exists we receive an error:

`L10N_ERROR[registration.error.userName.notUniqueWithinOrg]`

The parameter {0} seems to be necessary for the message but we don't want to leak the information that the email already exists.

We're using the widget in Version 5.16.0

Hopefully someone has an idea.

PR #1468, opened 2020-10-22
feat: add screen for required sign in with Okta Verify authenticator

jhe-okta:
## Description:
- when device probing is required, show I 22 and I 33
- when it is not, show I 23 and I 32
- update custom URI screen to I 34
- note: the Okta Verify icon is not added to the button in this PR, will address it separately

UX: https://www.figma.com/file/VEFAgD40zwFS9kTJkrem3cg4/Okta-Verify-Library-(WIP)?node-id=4099%3A980

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
when device probing is required:
smart-probe

when device probing is not required:
smart-probe-not-required

### Reviewers:
@stevennguyen-okta @haishengwu-okta @santoshmale-okta 

### Issue:

- [OKTA-338135](https://oktainc.atlassian.net/browse/OKTA-338135)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -513,7 +513,8 @@ primaryauth.password.tooltip = Password
 primaryauth.submit = Sign In
 primaryauth.newUser.tooltip=This is the first time you are connecting to {0} from this browser
 primaryauth.newUser.tooltip.close=Close
-oktaVerify.button = Sign in with Okta Verify
+oktaVerify.description = To access this resource, your organization requires you to sign in using your device.
+oktaVerify.button = Sign in using Okta Verify on this device
 
 # IDP Discovery
 idpDiscovery.email.placeholder = Email
@@ -678,6 +679,8 @@ signin.with.fastpass = Signing in using Okta FastPass
 customUri.title = Verify account access
 customUri.subtitle = Launching Okta Verify...
 customUri.content = <div class="skinny-content"> If nothing prompts from the browser, <a href="#" id="launch-ov" class="link">click here</a> to launch Okta Verify, or make sure Okta Verify is installed.</div>
+customUri.required.content.p1 = Please click&nbsp;<span class="strong">Open oktaverify.app</span>&nbsp;if you see the system dialog.
+customUri.required.content.p2 = If nothing prompts from the browser,&nbsp;<a href="#" id="launch-ov" class="link">click here</a>&nbsp;to launch Okta Verify, or&nbsp;<a href="{0}" target="_blank" id="download-ov" class="link">download & run Okta Verify</a>.
 universalLink.title = Verify your sign in
 universalLink.content = To continue, you\'ll need to use the <span class="highlight-text">Okta Verify</span> app to confirm your sign in.
 universalLink.userVerification.title = Extra verification needed
--- a/src/v2/view-builder/views/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/DeviceChallengePollView.js
@@ -61,10 +61,22 @@ const Body = BaseForm.extend(Object.assign(
         this.doLoopback(deviceChallenge.domain, deviceChallenge.ports, deviceChallenge.challengeRequest);
         break;
       case 'CUSTOM_URI':
-        this.title = loc('customUri.title', 'login');
-        this.subtitle = loc('customUri.subtitle', 'login');
+        this.title = loc('oktaVerify.button', 'login');
         this.add(View.extend({
-          template: hbs`{{{i18n code="customUri.content" bundle="login"}}}`
+          className: 'skinny-content',
+          template: hbs`
+            <p>
+              {{{i18n code="customUri.required.content.p1" bundle="login"}}}
+            </p>
+            <p>
+              {{{i18n code="customUri.required.content.p2" bundle="login" arguments="downloadOVLink"}}}
+            </p>
+          `,
+          getTemplateData () {
+            return {
+              downloadOVLink: deviceChallenge.downloadHref
+            };
+          },
         }));
         this.customURI = deviceChallenge.href;
         this.doCustomURI();
@@ -167,7 +179,7 @@ const Footer = BaseFooter.extend({
         {
           name: 'sign-in-options',
           type: 'link',
-          label: loc('goback', 'login'),
+          label: loc('oie.go.back', 'login'),
           href: this.settings.get('baseUrl')
         }
       ];
--- a/src/v2/view-builder/views/IdentifierView.js
+++ b/src/v2/view-builder/views/IdentifierView.js
@@ -2,12 +2,12 @@ import { loc, createCallout } from 'okta';
 import { FORMS as RemediationForms } from '../../ion/RemediationConstants';
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
-import BaseFooter from '../internals/BaseFooter';
+import IdentifierFooter from '../components/IdentifierFooter';
 import signInWithIdps from './signin/SignInWithIdps';
 import customButtonsView from './signin/CustomButtons';
 import signInWithDeviceOption from './signin/SignInWithDeviceOption';
 import { createIdpButtons, createCustomButtons } from '../internals/FormInputFactory';
-import { getForgotPasswordLink } from '../utils/LinksUtil';
+
 
 const Body = BaseForm.extend({
 
@@ -21,7 +21,7 @@ const Body = BaseForm.extend({
 
     // Launch Device Authenticator
     if (this.options.appState.hasRemediationObject(RemediationForms.LAUNCH_AUTHENTICATOR)) {
-      this.add(signInWithDeviceOption, '.o-form-fieldset-container', false, true);
+      this.add(signInWithDeviceOption, '.o-form-fieldset-container', false, true, { isRequired: false });
     }
 
     // This IdentifierView has been reused for the case when there is no `identify` remediation form
@@ -74,56 +74,9 @@ const Body = BaseForm.extend({
   },
 });
 
-const Footer = BaseFooter.extend({
-  links () {
-    let helpLinkHref;
-    if (this.options.settings.get('helpLinks.help')) {
-      helpLinkHref = this.options.settings.get('helpLinks.help');
-    } else {
-      const baseUrl = this.options.settings.get('baseUrl');
-      helpLinkHref = baseUrl + '/help/login';
-    }
-
-    const helpLink = [
-      {
-        'name': 'help',
-        'label': loc('help', 'login'),
-        'href': helpLinkHref,
-      },
-    ];
-
-    const signupLink = [];
-    if (this.options.appState.hasRemediationObject(RemediationForms.SELECT_ENROLL_PROFILE)) {
-      signupLink.push({
-        'type': 'link',
-        'label': loc('signup', 'login'),
-        'name': 'enroll',
-        'actionPath': RemediationForms.SELECT_ENROLL_PROFILE,
-      });
-    }
-
-    const forgotPasswordLink = getForgotPasswordLink(this.options.appState, this.options.settings);
-
-    const customHelpLinks = [];
-    if (this.options.settings.get('helpLinks.custom')) {
-      //add custom helpLinks
-      this.options.settings.get('helpLinks.custom').forEach(customHelpLink => {
-        customHelpLink.name = 'custom';
-        customHelpLink.label = customHelpLink.text;
-        customHelpLinks.push(customHelpLink);
-      });
-    }
-
-    return forgotPasswordLink
-      .concat(signupLink)
-      .concat(helpLink)
-      .concat(customHelpLinks);
-  }
-});
-
 export default BaseView.extend({
   Body,
-  Footer,
+  Footer: IdentifierFooter,
 
   postRender () {
     BaseView.prototype.postRender.apply(this, arguments);
--- a/src/v2/view-builder/views/SignInDeviceView.js
+++ b/src/v2/view-builder/views/SignInDeviceView.js
@@ -1,0 +1,31 @@
+import { loc } from 'okta';
+import BaseView from '../internals/BaseView';
+import BaseForm from '../internals/BaseForm';
+import IdentifierFooter from '../components/IdentifierFooter';
+import SignInWithDeviceOption from './signin/SignInWithDeviceOption';
+
+const Body = BaseForm.extend({
+
+  title () {
+    return loc('primaryauth.title', 'login');
+  },
+
+  noButtonBar: true,
+
+  initialize () {
+    BaseForm.prototype.initialize.apply(this, arguments);
+    this.add(
+      SignInWithDeviceOption,
+      { selector: '.o-form-fieldset-container',
+        bubble: false,
+        prepend: true,
+        options: { isRequired: true }
+      }
+    );
+  },
+});
+
+export default BaseView.extend({
+  Body,
+  Footer: IdentifierFooter,
+});
--- a/src/v2/view-builder/views/signin/SignInWithDeviceOption.js
+++ b/src/v2/view-builder/views/signin/SignInWithDeviceOption.js
@@ -4,8 +4,18 @@ import hbs from 'handlebars-inline-precompile';
 export default View.extend({
   className: 'sign-in-with-device-option',
   template: hbs`
-    <div class="okta-verify-container"></div>
-    <div class="separation-line"><span>OR</span></div>
+    <div class="okta-verify-container">
+    {{#if signInWithDeviceIsRequired}}
+      <div class="signin-with-ov-description">
+        {{i18n code="oktaVerify.description" bundle="login"}}
+      </div>
+    {{/if}}
+    </div>
+    {{#unless signInWithDeviceIsRequired}}
+    <div class="separation-line">
+      <span>{{i18n code="authbutton.divider.text" bundle="login"}}</span>
+    </div>
+    {{/unless}}
   `,
   initialize () {
     const appState = this.options.appState;
@@ -13,8 +23,18 @@ export default View.extend({
       className: 'button',
       title: loc('oktaVerify.button', 'login'),
       click () {
-        appState.trigger('invokeAction', 'launch-authenticator');
+        if (this.options.isRequired) {
+          appState.trigger('saveForm', this.model);
+        } else {
+          appState.trigger('invokeAction', 'launch-authenticator');
+        }
       }
     }), '.okta-verify-container');
+  },
+
+  getTemplateData () {
+    return {
+      signInWithDeviceIsRequired: !!this.options.isRequired,
+    };
   }
 });
--- a/test/testcafe/framework/page-objects/SignInDevicePageObject.js
+++ b/test/testcafe/framework/page-objects/SignInDevicePageObject.js
@@ -1,0 +1,36 @@
+import { Selector } from 'testcafe';
+import BasePageObject from './BasePageObject';
+
+export default class SignInDeviceViewPageObject extends BasePageObject {
+  constructor(t) {
+    super(t);
+    this.t = t;
+    this.beacon = new Selector('.beacon-container');
+    this.body = new Selector('.launch-authenticator');
+    this.footer = new Selector('.auth-footer');
+  }
+
+  getBeaconClass() {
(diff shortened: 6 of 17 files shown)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -678,6 +679,7 @@ signin.with.fastpass = Signing in using Okta FastPass
 customUri.title = Verify account access
 customUri.subtitle = Launching Okta Verify...
 customUri.content = <div class="skinny-content"> If nothing prompts from the browser, <a href="#" id="launch-ov" class="link">click here</a> to launch Okta Verify, or make sure Okta Verify is installed.</div>
+customUri.required.content = <div class="skinny-content">Please click&nbsp;<span class="strong">Open oktaverify.app</span>&nbsp;if you see the system dialog.</div><div class="skinny-content">If nothing prompts from the browser,&nbsp;<a href="#" id="launch-ov" class="link">click here</a>&nbsp;to launch Okta Verify, or&nbsp;<a href="{0}" target="_blank" id="download-ov" class="link">download & run Okta Verify</a>.</div>
```
pradeepdewda-okta, line 682: I have never seen this much of html inside a properties file. I don' think it's ideal. We should just keep the text and move out html to js file.

stevennguyen-okta:
What is the difference between `Sign out` and `Back to sign in` on the custom URI screen? It would be cool if the `sign out` wasn't there because the user hasn't tried to sign in yet at that point (such as entering their identifer)

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -61,10 +61,14 @@ const Body = BaseForm.extend(Object.assign(
         this.doLoopback(deviceChallenge.domain, deviceChallenge.ports, deviceChallenge.challengeRequest);
         break;
       case 'CUSTOM_URI':
-        this.title = loc('customUri.title', 'login');
-        this.subtitle = loc('customUri.subtitle', 'login');
+        this.title = loc('oktaVerify.button', 'login');
         this.add(View.extend({
-          template: hbs`{{{i18n code="customUri.content" bundle="login"}}}`
+          template: hbs`{{{i18n code="customUri.required.content" bundle="login" arguments="downloadOVLink"}}}`,
```
pradeepdewda-okta, line 66: put properties file html here..

jhe-okta:
> What is the difference between `Sign out` and `Back to sign in` on the custom URI screen? It would be cool if the `sign out` wasn't there because the user hasn't tried to sign in yet at that point (such as entering their identifer)

The `Sign out` link comes for free when there's a cancel object in the API response. The custom URI screen is shared between when device probe is required and not. I guess as a MFA, a session can be there if user has done primary auth and silent probe failed?

jhe-okta:
@stevennguyen-okta I changed the link in custom uri from `Back to sign in` to `Go back` per your request.
Screen Shot 2020-10-22 at 11 58 11 AM

stevennguyen-okta (approved):
Can you add the line break/paragraph spacing between the first sentence of custom URI and the second? It seems to have disappeared in your latest screenshot. Otherwise rest of it seems good to me

jhe-okta:
> Can you add the line break/paragraph spacing between the first sentence of custom URI and the second? It seems to have disappeared in your latest screenshot. Otherwise rest of it seems good to me

Oh yea, my bad. 
Screen Shot 2020-10-22 at 12 38 42 PM

src/v2/view-builder/views/SignInDeviceView.js
```diff
@@ -0,0 +1,80 @@
...
+      { selector: '.o-form-fieldset-container',
+        bubble: false,
+        prepend: true,
+        options: { isRequired: true, form: this }
+      }
+    );
+  },
+});
+
+const Footer = BaseFooter.extend({
```
haishengwu-okta, line 30: seems identical to `IdentifyView.BaseFooter` ?

src/v2/view-builder/views/signin/SignInWithDeviceOption.js
```diff
@@ -4,17 +4,37 @@ import hbs from 'handlebars-inline-precompile';
...
   `,
   initialize () {
     const appState = this.options.appState;
     this.add(createButton({
       className: 'button',
       title: loc('oktaVerify.button', 'login'),
       click () {
-        appState.trigger('invokeAction', 'launch-authenticator');
+        if (this.options.isRequired) {
+          this.options.form.trigger('save', this.options.form.model);
```
haishengwu-okta, line 25: Could be

```
    this.options.appState.trigger('saveForm', model);
```

and avoid to pass `form`

src/v2/view-builder/views/signin/SignInWithDeviceOption.js
```diff
@@ -4,17 +4,37 @@ import hbs from 'handlebars-inline-precompile';
...
+          appState.trigger('invokeAction', 'launch-authenticator');
+        }
       }
     }), '.okta-verify-container');
+  },
+
+  getTemplateData () {
+    const signInWithDeviceIsRequired = !!this.options.isRequired;
+    return {
+      signInWithDeviceIsRequired: signInWithDeviceIsRequired,
```
haishengwu-okta, line 36: you can use `if/else` in template instead of creating 2 vars.

test/testcafe/framework/page-objects/SignInDevicePageObject.js
```diff
@@ -0,0 +1,36 @@
...
+export default class SignInDeviceViewPageObject extends BasePageObject {
+  constructor(t) {
+    super(t);
+    this.t = t;
+    this.beacon = new Selector('.beacon-container');
+    this.body = new Selector('.launch-authenticator');
+    this.footer = new Selector('.auth-footer');
+  }
+
+  getBeaconClass() {
```
haishengwu-okta, line 13: Do you need to verify Beacon in test as well?

src/v2/view-builder/views/SignInDeviceView.js
jhe-okta: I will move the footer to shared/IdentifierFooter.js

src/v2/view-builder/views/signin/SignInWithDeviceOption.js
jhe-okta: I can make the change. But form is available in options, I don't need to pass in for this view specifically.

src/v2/view-builder/views/signin/SignInWithDeviceOption.js
jhe-okta: I can make the change. Was doing it to avoid weird closing `</div>` placement in the template.

test/testcafe/framework/page-objects/SignInDevicePageObject.js
jhe-okta: No, not really. I will remove the method.

pradeepdewda-okta:
> @stevennguyen-okta I changed the link in custom uri from `Back to sign in` to `Go back` per your request.
> Screen Shot 2020-10-22 at 11 58 11 AM

Do we really need Sign Out link here ? As the screen suggests to  `Sign In Using Okta Verify`, it would create confusion for enduser to have sign out link.. Can we hide or remove it ?

santoshmale-okta:
> Do we really need Sign Out link here ? As the screen suggests to Sign In Using Okta Verify, it would create confusion for enduser to have sign out link.. Can we hide or remove it ?

We have a lot of screens in the end user sign in flow and i see that almost all of them show the "Sign Out" link. For example, "Factor Selection Screen". If it is really a UX concern, should be handled broadly and not as a one off thing?

src/v2/view-builder/views/shared/IdentifierFooter.js
```diff
@@ -0,0 +1,51 @@
+import { loc } from 'okta';
+import BaseFooter from '../../internals/BaseFooter';
+import { FORMS as RemediationForms } from '../../../ion/RemediationConstants';
+import { getForgotPasswordLink } from '../../utils/LinksUtil';
+
+export default BaseFooter.extend({
```
pradeepdewda-okta, line 6: Need a spec file for this ?

jhe-okta:
> > @stevennguyen-okta I changed the link in custom uri from `Back to sign in` to `Go back` per your request.
> > Screen Shot 2020-10-22 at 11 58 11 AM
> 
> Do we really need Sign Out link here ? As the screen suggests to `Sign In Using Okta Verify`, it would create confusion for enduser to have sign out link.. Can we hide or remove it ?

As I explained to Steven, the `Sign out` link is there if cancel object exists in the API. This UI is shared in 2 different cases: 1. when device probe is required 2. when device probe is not. There's no session in case #1 but there could be session in case #2. In this UI, we cannot tell from the API which case it is for. We rely on the API to not to return the cancel object if sign out is not needed. Hiding the link is not ideal for case #2. I think we should follow with an API fix if QA found problem with it when testing.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -2,12 +2,12 @@ import { loc, createCallout } from 'okta';
 import { FORMS as RemediationForms } from '../../ion/RemediationConstants';
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
-import BaseFooter from '../internals/BaseFooter';
+import IdentifierFooter from './shared/IdentifierFooter';
```
haishengwu-okta, line 5: move to `component/IdentifierViewFooter`

stevennguyen-okta:
> > Do we really need Sign Out link here ? As the screen suggests to Sign In Using Okta Verify, it would create confusion for enduser to have sign out link.. Can we hide or remove it ?
> 
> We have a lot of screens in the end user sign in flow and i see that almost all of them show the "Sign Out" link. For example, "Factor Selection Screen". If it is really a UX concern, should be handled broadly and not as a one off thing?

For most of the screens like in the factor selection, the SignOut button made sense because they would have already provided their identifier. But yea my original question to see if we can get rid of the SIgn Out is for the same reason mentioned above, where it seems misleading because the user hasn't signed in yet

pradeepdewda-okta:
> > Do we really need Sign Out link here ? As the screen suggests to Sign In Using Okta Verify, it would create confusion for enduser to have sign out link.. Can we hide or remove it ?
> 
> We have a lot of screens in the end user sign in flow and i see that almost all of them show the "Sign Out" link. For example, "Factor Selection Screen". If it is really a UX concern, should be handled broadly and not as a one off thing?

Factor selection is a secondary factor, is this a secondary factor ?

src/v2/view-builder/views/signin/SignInWithDeviceOption.js
haishengwu-okta: looks like. it's form the view. 
```
{ isRequired: .., form: this }
``

File: test/testcafe/framework/page-objects/SignInDevicePageObject.js (hunk 6 above)
Comment (haishengwu-okta) [reply test/testcafe/framework/page-objects/SignInDevicePageObject.js]: Looks like the view does have beacon? I thought you would need to add assertion

File: src/v2/view-builder/views/shared/IdentifierFooter.js (hunk 7 above)
Comment (haishengwu-okta) [reply src/v2/view-builder/views/shared/IdentifierFooter.js:6]: Im leading toward to be covered when testing first class Views instead of individual spec for this shared component.

**Comment by santoshmale-okta** [comment]
> For most of the screens like in the factor selection, the SignOut button made sense because they would have already provided their identifier. But yea my original question to see if we can get rid of the SIgn Out is for the same reason mentioned above, where it seems misleading because the user hasn't signed in yet

In more than half of the ODA flows, the end user doesnt provide a user identifier at all. They silent land on the factor selection screen or directly get access.

> Factor selection is a secondary factor, is this a secondary factor ?

Why is factor selection a secondary factor screen? I directly land on the factor selection screen if i am entering username. For the ODA flows where silent probe succeeds, i dont even enter the user name. I just try to get to the app and i directly land on factor selection screen.

Also, yes, this could be a secondary factor screen as well in some cases as Jintai mentioned

File: src/v2/view-builder/views/signin/SignInWithDeviceOption.js [hunk 9]
```diff
@@ -4,17 +4,37 @@ import hbs from 'handlebars-inline-precompile';
 export default View.extend({
   className: 'sign-in-with-device-option',
   template: hbs`
-    <div class="okta-verify-container"></div>
-    <div class="separation-line"><span>OR</span></div>
+    <div class="okta-verify-container">
```
haishengwu-okta, line 7: would it be like this?
```

<div class="okta-verify-container">
     {{#if signInWithDeviceIsRequired}}
       <div class="signin-with-ov-description">
         {{i18n code="oktaVerify.description" bundle="login"}}
       </div>
     {{else}}
     <div class="separation-line">
       <span>{{i18n code="authbutton.divider.text" bundle="login"}}</span>
     </div>
     {{/if}}
</div>
```

src/v2/view-builder/views/signin/SignInWithDeviceOption.js
```diff
@@ -4,17 +4,37 @@ import hbs from 'handlebars-inline-precompile';
...
   `,
   initialize () {
     const appState = this.options.appState;
     this.add(createButton({
       className: 'button',
       title: loc('oktaVerify.button', 'login'),
       click () {
-        appState.trigger('invokeAction', 'launch-authenticator');
+        if (this.options.isRequired) {
+          this.options.appState.trigger('saveForm', this.model);
```
haishengwu-okta, line 27: `appState.trigger..`.
to be consistent with line 29 below.

src/v2/view-builder/views/SignInDeviceView.js
```diff
@@ -0,0 +1,31 @@
...
+  noButtonBar: true,
+
+  initialize () {
+    BaseForm.prototype.initialize.apply(this, arguments);
+    this.add(
+      SignInWithDeviceOption,
+      { selector: '.o-form-fieldset-container',
+        bubble: false,
+        prepend: true,
+        options: { isRequired: true, form: this }
```
haishengwu-okta, line 22: remove `form: this`

src/v2/view-builder/views/signin/SignInWithDeviceOption.js
jhe-okta: that's not the current DOM structure. div `.okta-verify-container` and div `.separation-line` are in parallel

pradeepdewda-okta:
> I directly land on the factor selection

from where ? Would it be possible to figure out if a session is in the progress ? so that we would know when to show or hide Sign Out ?

src/v2/view-builder/views/signin/SignInWithDeviceOption.js
haishengwu-okta: oh..missed that.
when `signInWithDeviceIsRequired` is false, you got empty `okta-verify-container` container, why not just get rid of completely.

santoshmale-okta:
> from where ? Would it be possible to figure out if a session is in the progress ? so that we would know when to show or hide Sign Out ?
Which scenario are you talking about? In all IDX flows where ODA is not involved the user enters username and they get the factor selection screen. For some ODA flows user just does an SP-init flow, the browser redirects to Okta, SIW just talks to Okta Verify (shows spinner for end user) and the SIW just displays the factor selection screen.

In fact i have a jira to make the Sign Out button "work" in this case. https://oktainc.atlassian.net/browse/OKTA-329366 which the PM approved to be fixed before EA :| I will be more than happy if we make the decision to get rid of sign out button everywhere in the sign in flow. :) But, you might have to talk to PM. I wouldn't block this PR on that. Even if this needs to be fixed, is this a server side fix or the Okta SIW?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -61,10 +61,21 @@ const Body = BaseForm.extend(Object.assign(
...
-        this.title = loc('customUri.title', 'login');
-        this.subtitle = loc('customUri.subtitle', 'login');
+        this.title = loc('oktaVerify.button', 'login');
         this.add(View.extend({
-          template: hbs`{{{i18n code="customUri.content" bundle="login"}}}`
+          template: hbs`
+            <div class="skinny-content">
+              {{{i18n code="customUri.required.content.p1" bundle="login"}}}
+            </div>
+            <div class="skinny-content">
```
pradeepdewda-okta, line 70: we can remove the extra `div` with same class and instead use <p> tags ?

jhe-okta: This is to put some spacing between the two divs.

pradeepdewda-okta:
> > from where ? Would it be possible to figure out if a session is in the progress ? so that we would know when to show or hide Sign Out ?
> 
> Which scenario are you talking about? In all IDX flows where ODA is not involved the user enters username and they get the factor selection screen. For some ODA flows user just does an SP-init flow, the browser redirects to Okta, SIW just talks to Okta Verify (shows spinner for end user) and the SIW just displays the factor selection screen.
> 
> In fact i have a jira to make the Sign Out button "work" in this case. https://oktainc.atlassian.net/browse/OKTA-329366 which the PM approved to be fixed before EA :| I will be more than happy if we make the decision to get rid of sign out button everywhere in the sign in flow. :) But, you might have to talk to PM. I wouldn't block this PR on that. Even if this needs to be fixed, is this a server side fix or the Okta SIW?

Talked to Jintai off line. I am good with this 👍  , If API returns `cancel` object SIW shows it otherwise not.

src/v2/view-builder/views/DeviceChallengePollView.js
pradeepdewda-okta: yes, the <p> tags will does it for you.

src/v2/view-builder/views/signin/SignInWithDeviceOption.js
jhe-okta: No, when it is false, another button is added from line 22.

haishengwu-okta: I see. try https://handlebarsjs.com/guide/builtin-helpers.html#unless . probably can simplify a bit.

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: `p` does not add the spacing for you. With `div`, no new style needs to be added. Other than making the UI more semantic, what's the significance of using `p` here?

pradeepdewda-okta: I think it's good enough reason to use p, proper semantics. If they were different classes i would have not asked to change it.

use p for paragraphs, div for containers.

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -61,10 +61,21 @@ const Body = BaseForm.extend(Object.assign(
         this.doLoopback(deviceChallenge.domain, deviceChallenge.ports, deviceChallenge.challengeRequest);
         break;
       case 'CUSTOM_URI':
-        this.title = loc('customUri.title', 'login');
-        this.subtitle = loc('customUri.subtitle', 'login');
+        this.title = loc('oktaVerify.button', 'login');
         this.add(View.extend({
-          template: hbs`{{{i18n code="customUri.content" bundle="login"}}}`
+          template: hbs`
+            <p class="skinny-content">
```
pradeepdewda-okta, line 67: Sorry if i was not clear.. I meant something like this..
```
<div class="skinny-content">
 <p> para 1 </p>
 <p> para 2 </p>
</div>
```

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -61,10 +61,22 @@ const Body = BaseForm.extend(Object.assign(
         this.doLoopback(deviceChallenge.domain, deviceChallenge.ports, deviceChallenge.challengeRequest);
         break;
       case 'CUSTOM_URI':
-        this.title = loc('customUri.title', 'login');
-        this.subtitle = loc('customUri.subtitle', 'login');
+        this.title = loc('oktaVerify.button', 'login');
         this.add(View.extend({
-          template: hbs`{{{i18n code="customUri.content" bundle="login"}}}`
+          className: 'skinny-content',
```
pradeepdewda-okta, line 66: much better..

pradeepdewda-okta (approved):
thanks for making these changes.!! 💯

haishengwu-okta (approved):

PR #1471, opened 2020-10-22
chore: bundle dependencies, remove externals

aarongranick-okta:
OKTA-320626

## Description:

Removes webpack "externals" for node module. This will bundle all required dependencies and will not require any additional modules to be installed alongside it. This will avoid errors when external modules are installed with incompatible versions or customizations.

This also enables `yarn link` to work locally.

Although "externals" is not a documented feature, this is considered a breaking change and should be included with the 5.0.0 release

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-320626](https://oktainc.atlassian.net/browse/OKTA-320626)

### DIFF
--- a/webpack.release.config.js
+++ b/webpack.release.config.js
@@ -19,27 +19,9 @@ var plugins = require('./buildtools/webpack/plugins');
 var useRuntime = require('./buildtools/webpack/runtime');
 var usePolyfill = require('./buildtools/webpack/polyfill');
 
-// 1. entryConfig (node module main entry. minified, no polyfill, external dependencies)
+// 1. entryConfig (node module main entry. minified, no polyfill)
 var entryConfig = config('okta-sign-in.entry.js');
 entryConfig.output.filename = 'okta-sign-in.entry.js';
-entryConfig.externals = {
-  // TODO: remove handlebars external OKTA-320626
-  'handlebars/runtime': {
-    'commonjs': 'handlebars/dist/handlebars.runtime',
-    'commonjs2': 'handlebars/dist/handlebars.runtime',
-    'amd': 'handlebars.runtime',
-    'root': 'handlebars'
-  },
-  'handlebars': {
-    'commonjs': 'handlebars/dist/handlebars.runtime',
-    'commonjs2': 'handlebars/dist/handlebars.runtime',
-    'amd': 'handlebars.runtime',
-    'root': 'handlebars'
-  },
-  'q': true,
-  'u2f-api-polyfill': true,
-  'underscore': true
-};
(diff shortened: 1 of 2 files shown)


webpack.release.config.js
```diff
@@ -19,27 +19,9 @@ var plugins = require('./buildtools/webpack/plugins');
 var useRuntime = require('./buildtools/webpack/runtime');
 var usePolyfill = require('./buildtools/webpack/polyfill');
 
-// 1. entryConfig (node module main entry. minified, no polyfill, external dependencies)
+// 1. entryConfig (node module main entry. minified, no polyfill)
 var entryConfig = config('okta-sign-in.entry.js');
 entryConfig.output.filename = 'okta-sign-in.entry.js';
-entryConfig.externals = {
```
haishengwu-okta, line 25: does this imply `entry.js` will bundle all those dependencies (handlebar, underscore etc)?

aarongranick-okta: yes

aarongranick-okta:
Screen Shot 2020-10-22 at 3 41 02 PM

pradeepdewda-okta:
@aarongranick-okta should we test it out on okta-core with alpha version to see if bacon is happy there too ?

aarongranick-okta:
> @aarongranick-okta should we test it out on okta-core with alpha version to see if bacon is happy there too ?

On the monolith we load widget from the CDN, so it would not be using the `.entry.js` file but rather the `.min.js` file which is already bundling dependencies.

This change only affects users of the NPM module

pradeepdewda-okta (approved):

webpack.release.config.js
haishengwu-okta: - why would do that.? and `entry.js` becomes same to `noPolyfillConfig`

aarongranick-okta: More information about why we want this change is here: https://oktainc.atlassian.net/browse/OKTA-320626

I think we probably can remove the no-polyfill entry and suggest people use entry.js from CDN for the no-polyfill option.
Moving deps to devDeps would reduce install time for the SIW, I'm OK with doing that. Sometimes people leave deps that are bundled in webpack for other reasons (like determining whether a security vuln is relevant or not). I'm not sure if we have a clear standard on that.

haishengwu-okta: Im not sure either and doesn't seem matter atm. most vuln concern we got so far is jQuery. also we shall try to keep 3rd-party as latest as possible

haishengwu-okta (approved):

PR #1473, opened 2020-10-23
docs: updates documentation of `renderEl` to be more comprehensive

### LINKED ISSUE #424: Improve documentation of function OktaSignIn.renderEl
The documentation of the function `OktaSignIn.renderEl` should be more precise and complete regards what possible values for `type` and `status` a developer can expect in the `success` callback. See https://github.com/okta/okta-signin-widget#rendereloptions-success-error

E.g. the `type` could be `SESSION_SSO`, but this isn't documented anywhere.

Even though there is a code snippet that shows a possible implementation of the `success` callback, this code snippet is not ideal, because it contains some dead code. The lines about _OIDC response_ and _signIn.tokenManager_ are never executed, because there are `return` statements before.

shawyu-okta:
## Description:

**PR to `4.5` branch**

* Updates documentation of `renderEl` and `success` callback parameters.
* Improves documentation of possible states that can be handled by the success handler

Closes #424.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-320763](https://oktainc.atlassian.net/browse/OKTA-320763)


README.md
```diff
@@ -15,44 +15,52 @@ You can learn more on the [Okta + JavaScript][lang-landing] page in our document
 
 <!-- TOC depthFrom:2 depthTo:3 -->
 
-- [Getting started](#getting-started)
```
shawyu-okta, line 18: It looks like there's an auto-TOC generator that updated this. It seems to have picked up some new links and deleted some orphaned ones.

README.md
```diff
@@ -277,62 +287,137 @@ signIn.renderEl(
...
         console.log('Target resource url: ' + res.stepUp.url);
         res.stepUp.finish();
         return;
-      } else {
-        // If the widget is not configured for OIDC, the response will contain
-        // user metadata and a sessionToken that can be converted to an Okta
-        // session cookie:
+      }
+
+      if (res.next) {
```
shawyu-okta, line 346: One of the confusing parts in the documentation before was that users did not have any understanding of when `next()` would be available and when `session.setCookieAndRedirect` would be available. I'm trying to differentiate here by checking for `next` as a property. Would love feedback here and on the documentation I've included for this block.

README.md
```diff
@@ -277,62 +287,137 @@ signIn.renderEl(
...
+  },
+
+  function error(err) {
+    // This function is invoked with errors the widget cannot recover from:
+    // Known errors: CONFIG_ERROR, UNSUPPORTED_BROWSER_ERROR
+  }
+);
+```
+
+**OIDC Configuration:**
```
shawyu-okta, line 385: I've divided the example code into two examples, one for non-OIDC configuration and one for OIDC-configurations, so that both examples are comprised of valid code that makes sense, rather than multiple examples in one.

README.md
```diff
@@ -277,62 +287,137 @@ signIn.renderEl(
...
 
       // If the widget is configured for OIDC with multiple responseTypes, the
       // response will be an array of tokens:
       // i.e. authParams.responseType = ['id_token', 'token']
-      signIn.tokenManager.add('my_id_token', res[0]);
-      signIn.tokenManager.add('my_access_token', res[1]);
+
+      // Example handling of multiple responseTypes instead:
+      // signIn.tokenManager.add('my_id_token', res[0]);
+      // signIn.tokenManager.add('my_access_token', res[1]);
```
shawyu-okta, line 412: Here's one way I'm trying to pack two scenarios into one example while still presenting technically valid code. The user can see both examples and uncomment/comment the relevant code example they will need.

README.md
```diff
@@ -277,62 +287,137 @@ signIn.renderEl(
...
+    if (res.status === 'ACTIVATION_EMAIL_SENT') {
+      // Any followup action you want to take
+      // Properties available on `res`:
+      // - status (string)
+      // - username (string)
+      return;
+    }
+
+    // The user has completed the registration flow
+    if (res.status === 'REGISTRATION_COMPLETE') {
```
shawyu-okta, line 317: We previously did not document `REGISTRATION_COMPLETE` as a possible status. I could use some assistance on better documenting the scenario that results in this block being called beyond what I've written so far.

README.md
```diff
@@ -277,62 +287,137 @@ signIn.renderEl(
...
       // Any followup action you want to take
+      // Properties available on `res`:
+      // - status (string)
+      // - username (string)
+      return;
+    }
+
+    // The user has started the activation flow, and is on the confirmation
+    // screen letting them know that an email is on the way.
+    if (res.status === 'ACTIVATION_EMAIL_SENT') {
```
shawyu-okta, line 308: We previously did not document `ACTIVATION_EMAIL_SENT ` as a possible status. I could use some assistance on better documenting the scenario that results in this block being called beyond what I've written so far.

shawyu-okta:
I've tried to do my best to impute the behavior and properties available on the `response` object based on digging through the code and API documentation, but someone with deeper knowledge getting eyes on the documentation I've written would be really helpful.

README.md
shawyu-okta: Highly recommend viewing this PR with `&w=1` to see actual TOC changes.

README.md
aarongranick-okta: I don't think `signIn.tokenManager` actually exists. In the next version, we will be removing/deprecating using `signIn.authClient` as well. It's not the widget's responsibility to store tokens, and it never reads them. We can leave this part a little bit vague I think, or provide links to full samples which show how to put it together with the other SDKs

README.md
aarongranick-okta: Do you know what generates the TOC? I'm making similar changes

README.md
shawyu-okta: I can leave both commented out as hypothetical SDKs that might exist and let the user write the implementation that works best for them, knowing that `res` or `res[0], res[1]` exist for them to consume.

README.md
aarongranick-okta: I think this happens when using self-service registration. The admin can configure either SMS or email activation I think.

README.md
aarongranick-okta: My guess is this is when self-service registration is enabled but neither SMS or EMAIL is required for activation.

README.md
```diff
@@ -277,62 +287,137 @@ signIn.renderEl(
...
+
+      if (res.next) {
+        // If the response contains a redirect URL, the response object will contain a
+        // function that redirects the user when called.
+        // Properties available on `res`:
+        // - status (string)
+        // - user (object)
+        // - type (string)
+        // - next (function)
         console.log(res.user);
```
anipendakur-okta, line 354: `nit:` Just curious as to why we need this `console.log` in readme? Since it doesn't add any value, can this be cleaned up?

Also, if this is going to be cleaned up, perhaps we can clean up other places in readme as well since this PR is changing readme.

README.md
aarongranick-okta: `setCookieAndRedirect` is available at the end of the flow when the widget is not configured for OIDC. It is used by customers in some federated / IDP scenarios. Use of this function has been identified as an "anti-pattern" in most cases because it uses an internal object (sessionToken) and allows an arbitrary redirect. It will be great to document how it works, but we should also try to nudge users towards using standard OIDC flows

README.md
shawyu-okta: Yeah, I was wondering about the `console.log` too. My best guess is that it was to help show what properties were available to the user in each scenario, but if we like the comment block documentation then I think removing the log statements as duplicative makes sense. WDYT?

README.md
aarongranick-okta: It would be great to have a little bit of text explaining why you would use OIDC (you probably should) vs. non-OIDC (probably not, unless your scenario is exotic in some way). Maybe there are some documents we can link to here.

README.md
shawyu-okta: Looks like it is a VSCode extension I just installed for Markdown called "Markdown All in One" (a feature I may be disabling, at least for `updateOnSave`)

README.md
anipendakur-okta: Imho, you're right, we don't need console.log. The updates you are doing indeed are helpful and are more elegant way of documentation. Hence `console.log` s can be removed. I'll let @haishengwu-okta and @aarongranick-okta chime in as well..

shawyu-okta: I'll make this change based on no objections so far

README.md
shawyu-okta: @aarongranick-okta I found this on the Dev Docs site, it seems to be a good place to link to: https://developer.okta.com/docs/guides/build-sso-integration/openidconnect/overview/#choosing-a-protocol

shawyu-okta: I can put the OIDC example first and lead with some text nudging the user to use OIDC for new apps and link this document.

README.md
```diff
@@ -263,6 +271,49 @@ Renders the widget to the DOM, and passes control back to your app through succe
...
+
+  function success(res) {
+    // Handling of other statuses in `res.status` is the same as the non-OIDC configuration
+
+    // The user has successfully completed the authentication flow
+    if (res.status === 'SUCCESS') {
+
+      // If the widget is configured for OIDC with a single responseType, the
+      // response will be the token.
+      // i.e. authParams.responseType = 'id_token':
```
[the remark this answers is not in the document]
aarongranick-okta: https://developer.okta.com/code/javascript/okta_sign-in_widget/#sign-in-and-display-user-s-email

README.md
```diff
@@ -263,6 +271,49 @@ Renders the widget to the DOM, and passes control back to your app through succe
...
+  },
+
+  function error(err) {
+    // Same as non-OIDC error handling function
+  }
+);
+```
+
+**Non-OIDC Configuration:**
+
```
aarongranick-okta, line 316: Can we provide any text here on WHY a user would choose a non-oidc configuration? What is the use-case?

aarongranick-okta: I think this **might be** when using the customized widget that is Okta hosted

README.md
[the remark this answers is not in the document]
aarongranick-okta: Yes, there is no difference in the result when setting it as a string or as an array. That was changed in auth-js 3.0: https://github.com/okta/okta-auth-js#from-2x-to-3x

haishengwu-okta (approved):

README.md
shawyu-okta: I swapped the spots of OIDC and non-OIDC but I have added a sentence with an example of when non-OIDC response might be handled (hosted SIW to log into Okta dashboard)

README.md
```diff
@@ -264,6 +272,10 @@ Renders the widget to the DOM, and passes control back to your app through succe
 - `success` *(optional)* - Function that is called when the user has completed an authentication flow. If an [OpenID Connect redirect flow](#openid-connect) is used, this function can be omitted.
 - `error` *(optional)* - Function that is called when the widget has been initialized with invalid config options, or has entered a state it cannot recover from. If omitted, a default function is used to output errors to the console.
 
+**OIDC Configuration:**
+
+This configuration is used for OIDC integrations. New integrations are recommended to use OIDC as it uses a more lightweight REST-based protocol and has more widespread usage.
```
shawyu-okta, line 277: I pulled back the link I had here. It was related to OIDC/OAuth vs SAML protocol. I couldn't find anything better on the docs site that would help a user pick OIDC other than most docs use examples of OIDC integrations.

README.md
```diff
@@ -264,6 +272,10 @@ Renders the widget to the DOM, and passes control back to your app through succe
 - `success` *(optional)* - Function that is called when the user has completed an authentication flow. If an [OpenID Connect redirect flow](#openid-connect) is used, this function can be omitted.
 - `error` *(optional)* - Function that is called when the widget has been initialized with invalid config options, or has entered a state it cannot recover from. If omitted, a default function is used to output errors to the console.
 
+**OIDC Configuration:**
```
haishengwu-okta, line 275: Will devs get same (similar) possible `status` regardless OIDC or Non-OIDC config?

shawyu-okta: I believe the `status` outcomes are the same, regardless of either config. I've left comments in the Non-OIDC config to the effect: https://github.com/okta/okta-signin-widget/pull/1473/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R364

README.md
```diff
@@ -261,82 +271,56 @@ Renders the widget to the DOM, and passes control back to your app through succe
...
 
-      // OIDC response
+- `user` *(optional\<object\>)* - Only present when widget is in a non-OIDC configuration, `status` is `SUCCESS`, and `type` is `SESSION_STEP_UP`.
 
-      // If the widget is configured for OIDC with a single responseType, the
-      // response will be the token.
-      // i.e. authParams.responseType = 'id_token':
-      console.log(res.claims);
-      signIn.tokenManager.add('my_id_token', res);
+- `stepUp` *(optional\<function\>)* - Only present when widget is in a non-OIDC configuration, `status` is `SUCCESS`, and `type` is `SESSION_STEP_UP`. `res.stepUp.finish()` call redirect the user to the URL at `res.stepUp.url`.
```
haishengwu-okta, line 315: what is `res.stepUp.url`. is it related to any config arguments when init SignInWidget ?

README.md
```diff
@@ -261,82 +271,56 @@ Renders the widget to the DOM, and passes control back to your app through succe
...
 
-      // If the widget is configured for OIDC with multiple responseTypes, the
-      // response will be an array of tokens:
-      // i.e. authParams.responseType = ['id_token', 'token']
-      signIn.tokenManager.add('my_id_token', res[0]);
-      signIn.tokenManager.add('my_access_token', res[1]);
+- `next` *(optional\<function\>)* - May be present when widget is in a non-OIDC configuration, `status` is `SUCCESS`, and the response contains a redirect URL. Calling this function will redirect the user.
```
haishengwu-okta, line 317: > May be present

Is that possible to more specific when it's available ?

README.md
```diff
@@ -261,82 +271,56 @@ Renders the widget to the DOM, and passes control back to your app through succe
...
 
+#### success
+
+Function that handles non-error events, including the completion of a successful authentication flow by a user. Also handles other flows such as reset password and self-registration flows.
+
+_OIDC vs Non-OIDC Configurations:_
+
+New integrations are recommended to use OIDC as it uses a more lightweight REST-based protocol and has more widespread usage. Less common is the need for non-OIDC integrations in self-hosted applications, however one example of this usage is when using the Okta-hosted sign-in widget to redirect a user to the Okta Dashboard after authentication.
+
+**`res` parameter:**
```
haishengwu-okta, line 285: The style for `h4` (`####`) and bold looks very close and may confuse people `res` is at same layer as `success`.
Would be more proper to change to
`/res parameter/` 
the goal is to make more clear that we're talking about the parameter of `success` callback.

README.md
```diff
@@ -261,82 +271,56 @@ Renders the widget to the DOM, and passes control back to your app through succe
...
-    // screen letting them know that an email is on the way.
-    if (res.status === 'UNLOCK_ACCOUNT_EMAIL_SENT') {
-      // Any followup action you want to take
-      return;
-    }
+- `status` *(string)* - Always present. One of: `FORGOT_PASSWORD_EMAIL_SENT`, `UNLOCK_ACCOUNT_EMAIL_SENT`, `ACTIVATION_EMAIL_SENT`, `REGISTRATION_COMPLETE`, or `SUCCESS`
 
-    // The user has successfully completed the authentication flow
-    if (res.status === 'SUCCESS') {
+- `username` *(optional\<string\>)* - Only present when `status` is one of `FORGOT_PASSWORD_EMAIL_SENT`, `UNLOCK_ACCOUNT_EMAIL_SENT`, `ACTIVATION_EMAIL_SENT`, or `REGISTRATION_COMPLETE`.
```
haishengwu-okta, line 305: a bit inconsistency:
`optional\<string\>` vs `Optional\<String\>`

No preference but just make it consistent.

shawyu-okta: Ah yeah, I'll go with lowercase ones since the previous docs have it that way.

README.md
shawyu-okta: I think it's the URL that gets returned from authentication, so it wouldn't be related to the config, however I'm not very familiar with how Step-Up session auth works.

README.md
shawyu-okta: I don't know when we return a redirect URL in the auth response. I looked in the code and we check for certain response fields and if they exist we add this function and set it to redirect. The logic is here in `RouterUtil`: https://github.com/okta/okta-signin-widget/blob/master/src/util/RouterUtil.js#L163

haishengwu-okta (approved):
rebase to 5.0 & 🚀

PR #1474, opened 2020-10-23
5.0 API changes

aarongranick-okta:
## Description:

Adds some breaking changes for 5.0 release. See changes in `MIGRATING.md` for full details. 

- `showSignInAndGetTokens` now returns a Promise and does **not redirect**
- new method `showSignInAndRedirect` which will do the redirect flow

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

### Issue:

- [OKTA-335945](https://oktainc.atlassian.net/browse/OKTA-335945)


MIGRATING.md
```diff
@@ -26,6 +26,32 @@ The existing `.widgetrc` file used to configure the Widget has been **removed**.
...
+## `scope` option has been removed
+
+This option has been deprecated with a warning since version 3 and has now been removed. You should use `scopes` (array) instead.
+
+### showSignInToGetTokens no longer redirects
+
+`showSignInToGetTokens` now returns a Promise that resolves to an object containing tokens. This avoids a round-trip redirect to Okta and provides a faster, simpler, and more seamless experience for both developers and end-users. A `redirectUri` must still be configured, but it is no longer necessary to provide any callback logic when using this method. If your application requires a redirect for some reason, such as server-side processing of an authorization code, you should use [showSignInAndRedirect](https://github.com/okta/okta-signin-widget#showsigninandredirect) instead.
+
+### renderEl now returns a Promise
+
```
swiftone, line 52: If we have `showSignInToGetTokens` to get the tokens directly, and `showSignInAndRedirect` to handle a redirect, we should say when they should make use of either promise OR the callbacks.

aarongranick-okta: They are equivalent, I don't think there's a particular reason to use one or the other except for syntax. Personally, I would use the Promise. At some point, we might deprecate the callbacks. Overall, I don't think there's many good reasons to use `renderEl` at all. If there are other common use cases we aren't covering with `signInToGetTokens` and `signInAndRedirect` we should add those as top-level functions.

MIGRATING.md
```diff
@@ -26,6 +26,32 @@ The existing `.widgetrc` file used to configure the Widget has been **removed**.
 - }
 ```
 
+## Migrating from 4.x to 5.x
+
+## authClient has been removed
+
+The internal instance of [@okta/okta-auth-js](https://github.com/okta/okta-auth-js) is no longer being exposed. If you need to work with the `okta-auth-js` library, you should add a dependency to [@okta/okta-auth-js](https://github.com/okta/okta-auth-js) and create an instance in your application.
```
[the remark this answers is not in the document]
aarongranick-okta: I don't think we are planning on accepting an instance of authClient from the outside. The internal instance will be customized in certain ways to avoid side-effects. For example, we will forcibly disable autorenew and other "background tasks" to avoid concurrency issues. In the longer term, we intend to break auth-js into smaller pieces and have only the relevant portions used by the SIW.

We could add a bit of an example of sharing config between SIW and auth-js however.

MIGRATING.md
```diff
@@ -26,6 +26,32 @@ The existing `.widgetrc` file used to configure the Widget has been **removed**.
...
+
+This option has been deprecated with a warning since version 3 and has now been removed. You should use `scopes` (array) instead.
+
+### showSignInToGetTokens no longer redirects
+
+`showSignInToGetTokens` now returns a Promise that resolves to an object containing tokens. This avoids a round-trip redirect to Okta and provides a faster, simpler, and more seamless experience for both developers and end-users. A `redirectUri` must still be configured, but it is no longer necessary to provide any callback logic when using this method. If your application requires a redirect for some reason, such as server-side processing of an authorization code, you should use [showSignInAndRedirect](https://github.com/okta/okta-signin-widget#showsigninandredirect) instead.
+
+### renderEl now returns a Promise
+
+The `success` and `error` callbacks are still supported, but you may prefer to use the Promise syntax.
```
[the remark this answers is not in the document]
aarongranick-okta: I added it for `showSignInToGetTokens` but I do think it gives better syntax, such as:

```
  try {
    await signIn.renderEl();
  } catch (e) {
    // handle error
  }
```

or

```
 signIn.renderEl().then(res => {
   // do something
 });

```

As implemented, it does not break the existing syntax in any way, it is just a new feature. Other methods which we may add that internally call `renderEl` will benefit from being able to use promise chains.

There are some weaknesses of the function callback pattern, such as difficulty catching errors that may be thrown from your success or error callbacks. The promise chain can handle this well with an additional `catch` or `finally` block.

MIGRATING.md
swiftone: > I don't think there's many good reasons to use renderEl at all.

That's the key part I'm looking to add.  If we don't recommend it, we should say that.  If there are specific circumstances where it's necessary, we should describe those, and say "otherwise, prefer to use `signInXXX`"

MIGRATING.md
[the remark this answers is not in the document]
haishengwu-okta: Make sense to customized the internal instance but why remove the exposure? It seems convenient if people need to play with some underlining function from auth-js for some cases. (I assume that are probably reasons to expose auth client at first place?)
Otherwise this seems very breaking changes. Developers need to create another auth-js instance which seems redundant.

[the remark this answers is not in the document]
swiftone: We already have problems with people using an external AuthJS that ends up conflicting with an internal AuthJS - rather than giving devs more ways to complicate matters, we want to reduce them.

AuthJS itself needs to slim down - see the issues we have with the widget when it isn't even using  AuthJS with OIE.  

Why create more ways to have problems?

src/models/Settings.js
```diff
@@ -215,17 +194,16 @@ export default Model.extend({
       },
     },
     oauth2Enabled: {
-      deps: ['clientId', 'authScheme', 'authParams.responseType'],
-      fn: function (clientId, authScheme, responseType) {
+      deps: ['clientId', 'authScheme'],
+      fn: function (clientId, authScheme) {
```
haishengwu-okta, line 198: looks like it can be
```
return (!!clientId) && authScheme.toLowerCase() === 'oauth2';

```

src/util/OAuth2Util.js
```diff
@@ -16,14 +16,21 @@ import Errors from './Errors';
...
   } else {
-    return type === params.responseType;
+    return type === responseType;
   }
 }
 
+util.isAuthorizationCodeFlow = function (settings) {
+  const authClient = settings.getAuthClient();
+  const pkce = authClient.options.pkce;
+  const responseType = settings.get('authParams.responseType');
```
haishengwu-okta, line 30: It's controversial that `authParams.responseType` has been removed from `props` definition but still being used.

MIGRATING.md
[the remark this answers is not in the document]
aarongranick-okta: @haishengwu-okta It is "convenient" to have a pre-created auth-js instance, but it has caused a lot of confusion and problems:
- the version which is bundled in SIW is almost never the current version, and if there are bugs, the user cannot install a newer version (because it is bundled in the webpack). It's not even clear what version you are using. Using your own version from NPM gives you control to upgrade it independently of the SIW
- authJS supports a lot of configuration options, passing these all in under `authParams` leads to confusion as to what SIW does and what authJS does. One center of confusion is the TokenManager, which SIW does not (and should not) use.
- we have added more features in authJS that involve it having background logic, such as token auto-renew. Having this logic running in the SIW could cause concurrency issues so we want to forcibly disable these features in the instance used by the SIW. This means we will ignore certain configuration options even if they are passed in authParams. So the instance of auth-js inside SIW is more limited than a standalone instance. We don't really want to document or explain this other than to say the internal instance is internal and we reserve the right to change it as we see fit in the future.
- with IDX all of these things come into play. There is an internal instance that is configured in a certain way for the SIW. We cannot guarantee that you could use this internal instance in the same way as a separate instance. We cannot even guarantee that the internal instance IS auth-js, it may in fact be an instance of idx-js.

src/util/OAuth2Util.js
aarongranick-okta: The `authParams` section is still supported, as it is used to configure the "internal" instance. However, the backbone props definitions potentially conflict with the actual configuration definition, which is defined here: https://github.com/okta/okta-auth-js/blob/master/lib/types/OktaAuthOptions.ts

This object is passed in blindly to the constructor of OktaAuth. The reason I am reading responseType here from widget settings rather than the authClient.options is because `responseType` is not saved when passed on the OktaAuth constructor, it is a `getToken` option: https://github.com/okta/okta-auth-js/blob/master/lib/types/api.ts#L77

I think **maybe** this is a bug in AuthJS, we should take `responseType` in the constructor, but also allow it to be overridden on a call to `getToken`. It would need to be added to the list here: https://github.com/okta/okta-auth-js/blob/master/lib/browser/browser.ts#L129

aarongranick-okta: https://oktainc.atlassian.net/browse/OKTA-340725 I am going to try to get this in before the 4.1 release which is planned for later this week. Then we can update SIW to 4.1 and it will not need to check `authParams.responseType` from settings.

MIGRATING.md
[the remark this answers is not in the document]
aarongranick-okta: Moving these changes to a separate PR: https://github.com/okta/okta-signin-widget/pull/1480 we can continue conversation there

aarongranick-okta:
Screen Shot 2020-10-28 at 11 40 01 PM

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -18,6 +18,7 @@ const itp = Expect.itp;
 Expect.describe('OktaSignIn initialization', function () {
   let signIn;
 
+  /* eslint jasmine/no-global-setup:0 */
```
haishengwu-okta, line 21: what is this change?

Gruntfile.js
```diff
@@ -177,6 +177,7 @@ module.exports = function (grunt) {
                 devLayout = grunt.file.read('./test/e2e/layouts/cdn-dev.tpl', {encoding: 'utf8'}),
                 indexLayout = grunt.file.read('./test/e2e/layouts/index.tpl', {encoding: 'utf8'}),
                 npmLayout = grunt.file.read('./test/e2e/layouts/npm.tpl', {encoding: 'utf8'}),
+                sharedFunctions = grunt.file.read('./test/e2e/partials/shared-functions.js', {encoding: 'utf8'}),
```
haishengwu-okta, line 180: why not just copy to `target/js` and loads from template directly?

test/e2e/partials/shared-functions.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable no-unused-vars */
+function addMessageToPage (id, msg) {
```
haishengwu-okta, line 2: nitpick: put under namespace like `window.widgetTestHelper = { ... }`

haishengwu-okta (commented):
Nice work!

Gruntfile.js
aarongranick-okta: With a `<script>` tag? That could work. In that case I would probably want a different folder name than `partials` if it's not being used in the templating system.

test/unit/spec/OktaSignIn_spec.js
aarongranick-okta: I really don't know why I was getting lint errors about the `beforeEach`, I think it doesn't recognize `Expect.describe` but not sure why it is effecting me, maybe version of eslint.

haishengwu-okta: Only the top level we need to use `Expect.describe` otherwise `describe` is good enough.
Does it fail bacon/travis as well (without this overwrite)

Gruntfile.js
haishengwu-okta: yeah, since this is just shared utilities, we just just load from base tpl.

test/unit/spec/OktaSignIn_spec.js
aarongranick-okta: yes I was seeing the same failure on Bacon.

aarongranick-okta: https://bacon-go.aue1e.saasure.net/#!/commits/okta-signin-widget/sha/cad8dc85197e71ce3cee5a0d6a2e115818ed4ad6

haishengwu-okta (approved):

PR #1491, opened 2020-11-06
feat: does not display phone number or email when not in the response

gowthamidommety-okta:
Resolves: OKTA-343133

## Description:

Phone:

Screen Shot 2020-11-06 at 11 36 53 AM

Screen Shot 2020-11-06 at 11 37 05 AM

Screen Shot 2020-11-06 at 11 37 14 AM

Screen Shot 2020-11-06 at 11 36 35 AM

Screen Shot 2020-11-06 at 11 36 43 AM

Phone number is now inline with message:

Screen Shot 2020-11-06 at 1 41 11 PM

Email

Screen Shot 2020-11-06 at 12 27 00 PM

Email subtitle has changed for m1:

Screen Shot 2020-11-06 at 12 28 01 PM

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-343133](https://oktainc.atlassian.net/browse/OKTA-343133)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -742,7 +742,8 @@ oie.phone.alternate.title = your phone
 ## Email
 oie.email.authenticator.description = Verify with a link or code sent to your email
 oie.email.mfa.title = Verify with your email
-oie.email.verify.sentText = An email was sent to
+oie.email.verify.sentText = Check <$1>{0}</$1> for a verification message. Click the verification button in your email or enter the code below to continue.
+oie.email.verify.alternate.sentText = Check your email for a verification message. Click the verification button in your email or enter the code below to continue.
 oie.email.enroll.subtitle = Please check your email and enter the code below.
 oie.email.return.link.expired.title = Verify with your email
 
--- a/playground/mocks/data/idp/idx/authenticator-verification-data-phone-voice-then-sms-no-number.json
+++ b/playground/mocks/data/idp/idx/authenticator-verification-data-phone-voice-then-sms-no-number.json
@@ -1,0 +1,324 @@
+{
+  "stateHandle":"eyJ6aXAiOiJERUYiLCJhbGlhcyI6ImVuY3J5cHRpb25rZXkiLCJ2ZXIiOiIxIiwib2lkIjoiMDBvdjRlUHM0eXFjT21TcDAwZzMiLCJlbmMiOiJBMjU2R0NNIiwiYWxnIjoiZGlyIn0..4BQNyMFP1RQojM0f.dIMxajkel5Sin9G9c8XjGYwcDH6q6rK2dWcZlDovCDQa3SmShn6Z_XgtuW8UpDtdBZYjLTTBcBmM6z_vdx7hhxZiDeAO6G-WRAs56bP-66O4l8aHok4zemRGRw1LLwviSF6LaqZ0ROL4l-wouvUd1nA2qBUABM-JYI6yBlUNy3biAGsUimKkohqhzYtKI99RBcXKHIeKhleVkKeLIBDUib64K1lAUaxXk-aIHuAp5_iAmdAvLXq3JBnBSduQTcrVtZuBwmcnCuDJD8vidINt1WSaayxguLSB4LbdY1CPoY1Le58c1-RWOMUeQwVMtWbdKDF2Lj-bLnPma62HFCtJuRmbzeUdFHUg9qwQ5wuaCy57dMKoymnVclkktcZ_ECC6DL2TrSpiJSMEBmzI-vYb91jxs5-Kov5IJSfMdCypqCBn49Zkthc-Fa3iLq7smqFMVZP2MH346YwCDtT8JJ6AhILFRzR3s-l9or8jlRkKtsqQ58yIPZiKGjGwTAP0RUSVQJDmOEhBo94uOH75YC8e5se8Iu8jm0BF8SHNeZwEvPSApGb6Bz_zKLC4p8YVq_9GWV-OLX1PByv4Af-Njkctcp8I-EW-b175SNKR4B66K_zd5x5CmH_yr3dsOp3mv6RgjKHToesUxjKXwToeFJnuP1HpcL133OdIbYQtk3T1CeEl-U37jBpTqm95ovu7qYTIuo4SKN7-z6MybO-T3ZBGfVav7y5FNpNIepNAWUFPaqR2flY9wJGtYT113NvTkLt3TuGLhomjGHWMYw-41lfifFldAcj3cc6nvu7GHSruz8Xf3qGOHA-v1wUaRXonFYfyWHvDtx5tXb7T8z29_U84w4gvyjN6GmCDld_EGhljRp9Lj9lgVqbR59QTdW_q9n5hTzAndgFaFqxqvJcx8A-8sbfD16vd0CAdFMMP2IGu5uSonwXYhoCtXAZ2Q-Y3gQ_hf6umqBXml9EZrkIFS4dx0JBldwlH-7hUn8bOweNEKYbiR-J8n2PpIFlZCxtbtjvf6S0ge2ijplynoHTd887z23dsefl3jiijEOThaNaz7CveLf5uLhg7-dsOK2bHPZvB43d40N1d7ZRm_ASeHF0DpLi6VKwNI-Hj5K8GHFsEiiGjWjjGyJxuXpzv2B0kDT3VElPbDZ8c9Tl2lnXhqPQuET6eVitgx-8DPyvBhM_Wpbkt4h3EGHbAAQBcKkZ8UvPy1_EPbGwDjX7_Uqw_EiQ3aR74C7Nzu6URdTI_RTFDMxZ-IC9woGXHeN8UESpHbOKskyd-Tr3KbDasvMgCYmgcSh574Xx7__kj-o311zQRu8sVbQCGFSmaaRkvqEXN1LcOo-atXD76gRe-_PARMcVEk3xjI1u-4cJbJDNQb7ji57NULeMY8Wbe7iEeLcAKgyHN3e3q0vOYXQpWOzatVWKnI1s2nS54tREp0HQ.-9f3flqUepYMc0GXwlQ7nA",
+  "version":"1.0.0",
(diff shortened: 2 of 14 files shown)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -742,7 +742,8 @@ oie.phone.alternate.title = your phone
 ## Email
 oie.email.authenticator.description = Verify with a link or code sent to your email
 oie.email.mfa.title = Verify with your email
-oie.email.verify.sentText = An email was sent to
+oie.email.verify.sentText = Check <$1>{0}</$1> for a verification message. Click the verification button in your email or enter the code below to continue.
+oie.email.verify.alternate.sentText = Check your email for a verification message. Click the verification button in your email or enter the code below to continue.
```
anipendakur-okta, line 746: Two questions -
* Do we also need to change the page title to `Verify with my email` to conform to the screenshot on this ticket - https://oktainc.atlassian.net/browse/OKTA-343133
* Does it make sense to change the message to `Check your email for a message with your code. Enter the code below.` (as per the screenshot in the ticket OKTA-343133). That way, when OKTA-343133 is done, there will only be an intermediate screen that conveys the user to either click on the link or enter the code on the next screen after clicking on a link.

anipendakur-okta:
Could you please add a screenshots of the changes for reference here..

packages/@okta/i18n/src/properties/login.properties
gowthamidommety-okta: Asking @vincentvoong-okta about 1.
About 2: The new JIRA to improve email magic link UX will be done post EA. So I think it's better to change the text for `sentText` until then right?

gowthamidommety-okta:
> Could you please add a screenshots of the changes for reference here..

Done! I wasn't expecting a review so soon 😄

packages/@okta/i18n/src/properties/login.properties
gowthamidommety-okta: 1. @vincentvoong-okta said it should be `Verify with your email`

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -33,10 +33,10 @@ const Body = BaseForm.extend(
       const sendText = ( this.model.get('primaryMode') === 'sms' )
         ? loc('oie.phone.verify.sms.sendText', 'login')
         : loc('oie.phone.verify.call.sendText', 'login');
+      const strongClass = this.model.get('phoneNumber') !== loc('oie.phone.alternate.title', 'login') ? 'strong' : '';
```
pradeepdewda-okta, line 36: instead of empty string i would add a class that doesn't make it strong.

gowthamidommety-okta: I couldn't find a class like that in courage.. What i see is that either the strong style is applied through the class
```
strong {
  font-weight: 600;
}
```
Or the class is not there which results in the font-weight being the same as the rest of the surrounding text

playground/mocks/data/idp/idx/authenticator-verification-data-phone-voice-then-sms-no-number.json
```diff
@@ -0,0 +1,324 @@
...
+            "visible":false,
+            "mutable":false
+          }
+        ]
+      },
+      "profile": {},
```
mauriciocastillosilva-okta, line 232: is this the only difference in all of the mock data? Same question for this and the other test files we are adding...
I'm asking cause it might be simpler to just clone the data in the test file and make profile empty, instead of adding new files that are basically the same...

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
pradeepdewda-okta: or we add the class to the dom only when it's needed, otherwise default is always no bold.

playground/mocks/data/idp/idx/authenticator-verification-select-authenticator-no-number.json
```diff
@@ -0,0 +1,290 @@
...
+        "authenticatorId": "aidtheidkwh282hv8g3"
+      },
+      {
+        "displayName": "FIDO2 (WebAuthn)",
+        "type": "security_key",
+        "id": "autwa6eD9o02iBbtv0g2",
+        "authenticatorId": "fwftheidkwh282hv8g3"
+      },
+      {
+        "displayName": "Okta Email",
```
mauriciocastillosilva-okta, line 225: email does not come with empty `profile`?

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -33,10 +33,10 @@ const Body = BaseForm.extend(
       const sendText = ( this.model.get('primaryMode') === 'sms' )
```
mauriciocastillosilva-okta, line 33: we will need to change this one in the future. Can you create a jira for the backlog?
Instead of having one text with "Send code to" and another with "your phone", we should have
"Send code to {0}" and "Send code to your phone"

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
mauriciocastillosilva-okta: I think this is good, but can we rename the variable name? Maybe something like `extraCssClasses`?

mauriciocastillosilva-okta (commented):
minor comments, looks good overall

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
pradeepdewda-okta: @mauriciocastillosilva-okta why adding an empty string is good? leaving DOM like this
`<span class>`

playground/mocks/data/idp/idx/authenticator-verification-select-authenticator-no-number.json
gowthamidommety-okta: For the authenticator selection screen there doesn't seem to be an email that we show. There can be multiple phone authenticators so in that case, the api returns phone numbers for each. cc @anipendakur-okta please confirm that my understanding is right

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
gowthamidommety-okta: let me take another look at this

playground/mocks/data/idp/idx/authenticator-verification-data-phone-voice-then-sms-no-number.json
gowthamidommety-okta: Yea, that is the only difference. I'm using it in multiple places though - tests and responseConfig mock data. That's why I created a file instead of clone and modify

playground/mocks/data/idp/idx/authenticator-verification-select-authenticator-no-number.json
gowthamidommety-okta: Seems like that logic is here - https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/utils/AuthenticatorUtil.js#L26

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
mauriciocastillosilva-okta: didn't see the other message before, but yes, my vote will be for not adding an empty class attribute if we don't need to. Agree 100% on that
I think that is what Gowthami implemented in the latest changes

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -33,10 +33,13 @@ const Body = BaseForm.extend(
...
         ? loc('oie.phone.verify.sms.sendText', 'login')
         : loc('oie.phone.verify.call.sendText', 'login');
-      const strongClass = this.model.get('phoneNumber') !== loc('oie.phone.alternate.title', 'login') ? 'strong' : '';
+      const extraCssClasses =
+        this.model.get('phoneNumber') !== loc('oie.phone.alternate.title', 'login') ?
+          'strong' : '';
       // Courage doesn't support HTML, hence creating a subtitle here.
-      this.add(`<div class="okta-form-subtitle" data-se="o-form-explain">${sendText}
-        <span class='${strongClass}'>${this.model.escape('phoneNumber')}</span></div>`);
+      this.add(`<div class="okta-form-subtitle" data-se="o-form-explain">${sendText} \
```
mauriciocastillosilva-okta, line 40: do you need backslash at the end `\`?

playground/mocks/data/idp/idx/authenticator-verification-select-authenticator-no-number.json
mauriciocastillosilva-okta: you're right, our [other mock data](https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/data/idp/idx/authenticator-verification-select-authenticator.json#L265-L275) also doesn't have `profile`...

mauriciocastillosilva-okta (approved):
just one question, rest looks good.

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
gowthamidommety-okta: I think i can remove it.. It doesn't look bad, it just has a new line after every line. I'll remove it.

playground/mocks/data/idp/idx/authenticator-verification-select-authenticator-no-number.json
anipendakur-okta: > There can be multiple phone authenticators so in that case, the api returns phone numbers for each.

That is correct.

pradeepdewda-okta (approved):

PR #1500, opened 2020-11-11
feat: implement oda/mdm remediation for Okta Verify not installed case

yannongli-okta:
## Description:

This is implementing a special use case in oda/mdm remediation flow. When Okta Verify is not installed and user clicks on the button, it will be first redirected to login.okta.com and then redirected back to /authenticators/ov-not-installed. This endpoint will return a new SIW along with a proxyIdxResponse value in SIW config which contains the device enrollment object values we need.

The way we implement it is: if SIW detects such a proxyIdxResponse in config, it will still go to V2 router but won't make an introspect call by default. This is to make it a more generic solution in case we have similar use case to pass some customized config from server side.

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:

https://okta.box.com/s/c6d7fit1wzxsxq475rhfg4t33nhl92y1

### Reviewers:

@haishengwu-okta @pradeepdewda-okta @jhe-okta 

### Issue:

- [OKTA-342925](https://oktainc.atlassian.net/browse/OKTA-342925)

### DIFF
--- a/src/models/Settings.js
+++ b/src/models/Settings.js
@@ -54,6 +54,10 @@ export default Model.extend({
     // IDX API VERSION
     apiVersion: ['string', true, '1.0.0'],
 
+    // attribute to hold proxy (fake) idx response
+    // to render static pages without initiating idx pipeline
+    proxyIdxResponse: ['object', false],
+
     // FEATURES
     'features.router': ['boolean', true, false],
     'features.securityImage': ['boolean', true, false],
--- a/src/v2/BaseLoginRouter.js
+++ b/src/v2/BaseLoginRouter.js
@@ -13,6 +13,7 @@
 // BaseLoginRouter contains the more complicated router logic - rendering/
 // transition, etc. Most router changes should happen in LoginRouter (which is
 // responsible for adding new routes)
+/* global Promise */
 import { _, $, Backbone, Router, loc } from 'okta';
 import Settings from 'models/Settings';
 import Bundles from 'util/Bundles';
@@ -34,6 +35,14 @@ const introspectStateToken = (settings) => {
   const stateHandle = settings.get('stateToken');
   const version = settings.get('apiVersion');
   return idx.start({ domain, stateHandle, version });
+};
+
+const handleProxyIdxResponse = (settings) => {
+  return Promise.resolve({
+    rawIdxState: settings.get('proxyIdxResponse'),
+    context: settings.get('proxyIdxResponse'),
+    neededToProceed: [],
+  });
 };
 
 export default Router.extend({
@@ -150,10 +159,13 @@ export default Router.extend({
     // introspect stateToken when widget is bootstrap with state token
     // and remove it from `settings` afterwards as IDX response always has
     // state token (which will be set into AppState)
-    if (this.settings.get('stateToken')) {
-      return introspectStateToken(this.settings)
+    if (this.settings.get('stateToken') || this.settings.get('proxyIdxResponse')) {
+      const idxRespPromise = this.settings.get('proxyIdxResponse') ?
+        handleProxyIdxResponse(this.settings) : introspectStateToken(this.settings);
+      return idxRespPromise
         .then(idxResp => {
           this.settings.unset('stateToken');
+          this.settings.unset('proxyIdxResponse');
           this.appState.trigger('remediationSuccess', idxResp);
           this.render(Controller, options);
         })
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -153,7 +153,7 @@ var OktaSignIn = (function () {
     var authClient = createAuthClient(authParams);
 
     var Router;
-    if (options.stateToken && !Util.isV1StateToken(options.stateToken)) {
+    if ((options.stateToken || options.proxyIdxResponse) && !Util.isV1StateToken(options.stateToken)) {
       Router = V2Router;
     } else {
       Router = V1Router;
--- a/test/unit/spec/OktaSignIn_spec.js
+++ b/test/unit/spec/OktaSignIn_spec.js
@@ -355,6 +355,30 @@ Expect.describe('OktaSignIn v2 bootstrap', function () {
     return signIn.renderEl({ el: $sandbox });
   }
 
+  function setupProxyIdxResponse (options) {
+    signIn = new Widget(
+      Object.assign(
+        {
+          baseUrl: url,
+          proxyIdxResponse: {
+            deviceEnrollment: {
+              type: 'object',
+              value: {
+                name: options.enrollmentType,
+                platform: 'IOS',
+                enrollmentLink: 'https://sampleEnrollmentlink.com',
+                vendor: 'Airwatch',
+                signInUrl: 'https://idx.okta1.com'
+              }
+            }
+          }
+        },
+        options || {}
+      )
+    );
+    return signIn.renderEl({ el: $sandbox });
+  }
+
   Expect.describe('Introspects token and loads Identifier view for new pipeline', function () {
     itp('calls introspect API on page load using idx-js as client', function () {
       const form = new IdentifierForm($sandbox);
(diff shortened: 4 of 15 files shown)


src/widget/OktaSignIn.js
```diff
@@ -153,7 +153,7 @@ var OktaSignIn = (function () {
     var authClient = createAuthClient(authParams);
 
     var Router;
-    if (options.stateToken && !Util.isV1StateToken(options.stateToken)) {
+    if ((options.stateToken && !Util.isV1StateToken(options.stateToken)) || options.proxyIdxResponse) {
```
pradeepdewda-okta, line 156: I would change this to 
```
(options.stateToken || options.proxyIdxResponse) && !Util.isV1StateToken(options.stateToken
```

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -12,6 +12,7 @@ import Q from 'q';
 import $sandbox from 'sandbox';
 import Logger from 'util/Logger';
 import Widget from 'widget/OktaSignIn';
+import DeviceEnrollmentTerminalViewForm from '../helpers/dom/v2/DeviceEnrollmentTerminalViewForm';
```
pradeepdewda-okta, line 15: We don't need to test this in here.. Just add a test that can confirm that if proxyIdxResponse exists in the config as parameter then it bootstrap V2 router or not.. that's it.!

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -386,4 +411,45 @@ Expect.describe('OktaSignIn v2 bootstrap', function () {
       });
     });
   });
-});
+
+  Expect.describe('Directly consumes proxyIdxResponse from widget config and do not introspect state token', function () {
```
pradeepdewda-okta, line 415: we cam get rid of these tests.

src/v2/BaseLoginRouter.js
```diff
@@ -53,6 +62,7 @@ export default Router.extend({
 
     this.settings = new Settings(_.omit(options, 'el', 'authClient'), { parse: true });
     this.settings.setAuthClient(options.authClient);
+    this.settings.set('proxyIdxResponse', options.proxyIdxResponse);
```
pradeepdewda-okta, line 65: i think adding a local property in src/models/Settings.js would make more sense.. instead of setting this here. @haishengwu-okta can confirm once.

pradeepdewda-okta (commented):
looks good to me.

src/v2/BaseLoginRouter.js
haishengwu-okta, line 65: you don't have to. line 63 shall take care it.

src/widget/OktaSignIn.js
haishengwu-okta: I'd prefer what Yannong did. :)

pradeepdewda-okta: In that case if it has v1stateToken & contains `proxyIdxResponse` then it will still go to V2 pipeline..  right ?

src/v2/BaseLoginRouter.js
yannongli-okta: @haishengwu-okta The reason i added this is to get a nested `proxyIdxResponse` object directly. Otherwise it's flattened into each specific property like `proxyIdxResponse.deviceEnrollment.name` or `proxyIdxResponse.deviceEnrollment.platform`. I will need to re-construct the same `proxyIdxResponse` object which is redundant. Do you think there is a better way to handle this?

yannongli-okta: also ^^ @pradeepdewda-okta for the reason i explicitly add it in settings.

haishengwu-okta: like @pradeepdewda-okta mentioned, you can add definition to `models/Settings.js`
```
local: {
    ...,
    proxyIdxResponse: 'object',
    ...,
}
```

src/widget/OktaSignIn.js
haishengwu-okta: I think so.

pradeepdewda-okta: I feel if there is v1StateToken then it should only go to V1 router ? and just ignore `proxyIdxResponse` param..

src/models/Settings.js
```diff
@@ -54,6 +54,10 @@ export default Model.extend({
     // IDX API VERSION
     apiVersion: ['string', true, '1.0.0'],
 
+    // if ProxyIdxResponse is present in SIW config
```
pradeepdewda-okta, line 57: `attribute to hold proxy (fake) idx responses to render static pages without initiating idx pipeline`

src/widget/OktaSignIn.js
yannongli-okta: I think either way will work, just how we want to handle v1/v2 state token use cases when proxyIdxResponse is present.

src/v2/BaseLoginRouter.js
```diff
@@ -13,6 +13,7 @@
 // BaseLoginRouter contains the more complicated router logic - rendering/
 // transition, etc. Most router changes should happen in LoginRouter (which is
 // responsible for adding new routes)
+/* global Promise */
```
pradeepdewda-okta, line 16: we need that ?

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -355,6 +355,30 @@ Expect.describe('OktaSignIn v2 bootstrap', function () {
     return signIn.renderEl({ el: $sandbox });
   }
 
+  function setupProxyIdxResponseForDeviceEnrollment (options) {
```
pradeepdewda-okta, line 358: rename `setupProxyIdxResponseForDeviceEnrollment` to `setupProxyIdxResponse`

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -386,4 +410,23 @@ Expect.describe('OktaSignIn v2 bootstrap', function () {
       });
     });
   });
-});
+
+  Expect.describe('Directly consumes proxyIdxResponse from widget config and do not introspect state token', function () {
+    itp('Gets device enrollment values from proxyIdxResponse and render mdm device enrollment terminal view', function () {
```
pradeepdewda-okta, line 415: it's not relevant here to mention anything about `device enrollment`.. just say `gets proxy idx response and render terminal view`

src/widget/OktaSignIn.js
pradeepdewda-okta: since there is no concept of `proxyIdxResponse` for V1.. I would assume it's always for v2 pipeline only.. @haishengwu-okta what do you think ?

src/v2/BaseLoginRouter.js
yannongli-okta: Yeah this is required to use built-in Promise. Otherwise i can import Q.

test/unit/spec/OktaSignIn_spec.js
haishengwu-okta: I guess it's fine to leave a simple sanity check here ?

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -355,6 +355,30 @@ Expect.describe('OktaSignIn v2 bootstrap', function () {
     return signIn.renderEl({ el: $sandbox });
   }
 
+  function setupProxyIdxResponse (options) {
```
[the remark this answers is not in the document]
yannongli-okta: yeah that will work. It's just i would like to avoid unnecessary setup for this use case. And also, since we won't actually make introspect call, `setupIntrospect` doesn't seems to be a appropriate name.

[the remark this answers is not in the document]
haishengwu-okta: hmm, good point.

src/v2/BaseLoginRouter.js
pradeepdewda-okta: I mean this comment.. Without that `Promise` won't be available ?

src/widget/OktaSignIn.js
yannongli-okta: already update this part based on @pradeepdewda-okta 's advice

haishengwu-okta:
in favor of https://github.com/okta/okta-signin-widget/pull/1505

PR #1505, opened 2020-11-15
feat: implement oda/mdm remediation for Okta Verify not installed case

yannongli-okta:
## Description:

This is implementing a special use case in oda/mdm remediation flow. When Okta Verify is not installed and user clicks on the button, it will be first redirected to login.okta.com and then redirected back to /authenticators/ov-not-installed. This endpoint will return a new SIW along with a proxyIdxResponse value in SIW config which contains the device enrollment object values we need.

The way we implement it is: if SIW detects such a proxyIdxResponse in config, it will still go to V2 router but won't make an introspect call by default. This is to make it a more generic solution in case we have similar use case to pass some customized config from server side.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
https://okta.box.com/s/c6d7fit1wzxsxq475rhfg4t33nhl92y1

### Issue:

- [OKTA-342925](https://oktainc.atlassian.net/browse/OKTA-342925)

### DIFF
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -153,7 +153,7 @@ var OktaSignIn = (function () {
     var authClient = createAuthClient(authParams);
 
     var Router;
-    if (options.stateToken && !Util.isV1StateToken(options.stateToken)) {
+    if ((options.stateToken && !Util.isV1StateToken(options.stateToken)) || options.proxyIdxResponse) {
       Router = V2Router;
     } else {
       Router = V1Router;
--- a/src/models/Settings.js
+++ b/src/models/Settings.js
@@ -54,6 +54,10 @@ export default Model.extend({
     // IDX API VERSION
     apiVersion: ['string', true, '1.0.0'],
 
+    // attribute to hold proxy (fake) idx response
+    // to render static pages without initiating idx pipeline
+    proxyIdxResponse: ['object', false],
+
     // FEATURES
     'features.router': ['boolean', true, false],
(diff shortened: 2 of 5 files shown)


yannongli-okta:
This is a duplicate PR of https://github.com/okta/okta-signin-widget/pull/1500 That one was rebased on wrong branch and had bad commit history. So creating this one for actual merging.

src/widget/OktaSignIn.js
```diff
@@ -153,7 +153,7 @@ var OktaSignIn = (function () {
     var authClient = createAuthClient(authParams);
 
     var Router;
-    if (options.stateToken && !Util.isV1StateToken(options.stateToken)) {
+    if ((options.stateToken || options.proxyIdxResponse) && !Util.isV1StateToken(options.stateToken)) {
```
haishengwu-okta, line 156: hmm, I thought we want

haishengwu-okta, line 156: I thought we want to use your initial logic: when `proxyIdxResponse` is present, we shall go to v2 regardless state token.

yannongli-okta: @haishengwu-okta Oh I saw you gave a ship on this discussion https://github.com/okta/okta-signin-widget/pull/1500#discussion_r521740565 so i thought you agreed with what @pradeepdewda-okta suggested?

haishengwu-okta (approved):

jhe-okta (approved):

PR #1514, opened 2020-11-18
test: adds data-se attribute on buttons for authenticators select list

gowthamidommety-okta:
Resolves: OKTA-347191

## Description:

Enroll:

enrollAuthenticators

Verify:

verifyAuthenticators

Verify OV:

verifyAuthenticatorsOV

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-347191](https://oktainc.atlassian.net/browse/OKTA-347191)

### DIFF
--- a/src/v2/view-builder/utils/AuthenticatorUtil.js
+++ b/src/v2/view-builder/utils/AuthenticatorUtil.js
@@ -14,6 +14,15 @@ import FactorUtil from '../../../util/FactorUtil';
 
 const { getPasswordComplexityDescription, getPasswordComplexityDescriptionForHtmlList } = FactorUtil;
 
+const getButtonDataSeAttr = function (authenticator) {
+  if (authenticator.authenticatorType) {
+    const method = authenticator.value?.methodType ?
+      '-' + authenticator.value?.methodType : '';
+    return authenticator.authenticatorType + method;
+  }
+  return '';
+};
+
 /* eslint complexity: [2, 19] */
 const getAuthenticatorData = function (authenticator, isVerifyAuthenticator) {
   const authenticatorType = authenticator.authenticatorType || authenticator.factorType;
@@ -26,6 +35,7 @@ const getAuthenticatorData = function (authenticator, isVerifyAuthenticator) {
         ? ''
         : loc('oie.email.authenticator.description', 'login'),
       iconClassName: 'mfa-okta-email',
+      buttonDataSeAttr: getButtonDataSeAttr(authenticator),
     });
     break;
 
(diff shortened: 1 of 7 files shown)


mauriciocastillosilva-okta (approved):

test/unit/spec/v2/view-builder/internals/FormInputFactory_spec.js
```diff
@@ -38,13 +38,15 @@ describe('v2/view-builder/internals/FormInputFactory', function () {
         factorType: 'password',
         iconClassName: 'mfa-okta-password',
         description: 'Choose a password for your account',
+        buttonDataSeAttr: '',
```
mauriciocastillosilva-okta, line 41: why is this empty?

src/v2/view-builder/utils/AuthenticatorUtil.js
```diff
@@ -14,6 +14,13 @@ import FactorUtil from '../../../util/FactorUtil';
 
 const { getPasswordComplexityDescription, getPasswordComplexityDescriptionForHtmlList } = FactorUtil;
 
+const getButtonDataSeAttr = function (authenticator) {
+  const method = authenticator.value?.methodType ?
+    '-' + authenticator.value?.methodType : '';
+  const authenticatorType = authenticator.authenticatorType || '';
```
mauriciocastillosilva-okta, line 20: is it possible that an authenticator does not have a type?

test/unit/spec/v2/view-builder/internals/FormInputFactory_spec.js
gowthamidommety-okta: This test is for the old LEA implementation that has `factorType` instead of `authenticatorType`. This implementation also uses the same `getAuthenticatorData` in `AuthenticatorUtil` - https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/internals/FormInputFactory.js#L12

src/v2/view-builder/utils/AuthenticatorUtil.js
gowthamidommety-okta: Same answer as the other - Because `getAuthenticatorData` is used for both LEA factorTypes and IDX Authenticators, i'm returning `''` in case the object does not have authenticatorType. Otherwise the LEA tests will set `'undefined'` instead of `''`. I think this is something we could clean up when we remove LEA code. Similar code here - https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/utils/AuthenticatorUtil.js#L19

mauriciocastillosilva-okta: Got it. One suggestion though, should we rewrite this something like
```
if (authenticator.authenticatorType) {
  const method = authenticator.value?.methodType ? '-' + authenticator.value?.methodType : '';
  return authenticator.authenticatorType + method;
}
return ''; // or 'unknown-factor' if we want a string in the data-se attribute
```

gowthamidommety-okta: Will change to `''` for now 👍

mauriciocastillosilva-okta (approved):
looks good 🐑

PR #1525, opened 2020-11-23
fix: terminalview beacon for email scenarios

anipendakur-okta:
## Description:
Please refer to https://github.com/okta/okta-signin-widget/pull/1523 for screenshots

## PR Checklist

- [ ] Have you verified the basic functionality for this change?

- [ ] Added unit tests?

- [ ] Added e2e tests

- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?

### Reviewers:
@thomasmaslen-okta @haishengwu-okta 

### Issue:

- [OKTA-312021](https://oktainc.atlassian.net/browse/OKTA-312021)

PR #1527, opened 2020-11-23
fix: fixing beacon for email terminal views

anipendakur-okta:
## Description:
Ref https://github.com/okta/okta-signin-widget/pull/1525/files

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Before - 
image

After -
image

### Issue:

- [OKTA-312021](https://oktainc.atlassian.net/browse/OKTA-312021)

### DIFF
--- a/src/v2/view-builder/views/TerminalView.js
+++ b/src/v2/view-builder/views/TerminalView.js
@@ -2,9 +2,30 @@ import { createCallout, loc } from 'okta';
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
 import BaseFooter from '../internals/BaseFooter';
-import {getBackToSignInLink} from '../utils/LinksUtil';
+import BaseHeader from '../internals/BaseHeader';
+import HeaderBeacon from '../components/HeaderBeacon';
+import { getBackToSignInLink } from '../utils/LinksUtil';
+import { getIconClassNameForBeacon } from '../utils/AuthenticatorUtil';
 
 const RETURN_LINK_EXPIRED_KEY = 'idx.return.link.expired';
+
+const EMAIL_AUTHENTICATOR_TERMINAL_KEYS = [
+  'idx.transferred.to.new.tab',
+  'idx.return.to.original.tab',
+  RETURN_LINK_EXPIRED_KEY,
+  'idx.return.stale',
+  'idx.return.error'
+];
+
+const EMAIL_AUTHENTICATOR_TYPE = 'email';
+
+const HeaderBeaconTerminal = HeaderBeacon.extend({
+  getBeaconClassName: function () {
+    return this.options.appState.containsMessageWithI18nKey(EMAIL_AUTHENTICATOR_TERMINAL_KEYS)
+      ? getIconClassNameForBeacon(EMAIL_AUTHENTICATOR_TYPE)
+      : HeaderBeacon.prototype.getBeaconClassName.apply(this, arguments);
+  }
+});
 
 const Body = BaseForm.extend({
   noButtonBar: true,
@@ -53,6 +74,9 @@ const Footer = BaseFooter.extend({
 });
 
 export default BaseView.extend({
+  Header: BaseHeader.extend({
+    HeaderBeacon: HeaderBeaconTerminal,
+  }),
   Body,
   Footer
 });
--- a/src/v2/view-builder/views/TerminalView.js
+++ b/src/v2/view-builder/views/TerminalView.js
@@ -14,7 +14,8 @@ const EMAIL_AUTHENTICATOR_TERMINAL_KEYS = [
   'idx.return.to.original.tab',
   RETURN_LINK_EXPIRED_KEY,
   'idx.return.stale',
-  'idx.return.error'
+  'idx.return.error',
+  'idx.email.verification.required'
 ];
 
 const EMAIL_AUTHENTICATOR_TYPE = 'email';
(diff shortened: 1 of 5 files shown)


src/v2/view-builder/views/TerminalView.js
```diff
@@ -2,10 +2,32 @@ import { createCallout, loc } from 'okta';
...
 
 const RETURN_LINK_EXPIRED_KEY = 'idx.return.link.expired';
 
+const EMAIL_AUTHENTICATOR_TERMINAL_KEYS = [
+  'idx.transferred.to.new.tab',
+  'idx.return.to.original.tab',
+  RETURN_LINK_EXPIRED_KEY,
+  'idx.return.stale',
+  'idx.return.error',
+  'idx.email.verification.required'
```
thomasmaslen-okta, line 18: For comments / caveats see https://github.com/okta/okta-signin-widget/pull/1525/files#r528879396 -- and a sanity check from @haishengwu-okta would be a good thing -- but yes, in general I'm in favour of including this `idx.email.verification.required` key.

thomasmaslen-okta (approved):

test/testcafe/spec/TerminalView_spec.js
```diff
@@ -0,0 +1,47 @@
+import { RequestMock } from 'testcafe';
```
gowthamidommety-okta, line 1: Should we name this test file something specific to email? `EmailTerminalView`?

src/v2/view-builder/views/TerminalView.js
```diff
@@ -2,10 +2,32 @@ import { createCallout, loc } from 'okta';
...
+];
+
+const EMAIL_AUTHENTICATOR_TYPE = 'email';
+
+const HeaderBeaconTerminal = HeaderBeacon.extend({
```
gowthamidommety-okta, line 23: Can you attach some screenshots of what this looks like with and without email terminal response?

test/testcafe/spec/GenericErrorHandling_spec.js
```diff
@@ -39,4 +39,5 @@ test.requestHooks(securityAccessDeniedMock)('should be able display error when r
   const terminalPage = await setup(t);
   await terminalPage.waitForErrorBox();
   await t.expect(terminalPage.getErrorMessages().getTextContent()).eql('You do not have permission to perform the requested action.');
+  await t.expect(terminalPage.getBeaconClass()).notContains('mfa-okta-email');
```
gowthamidommety-okta, line 42: is there a specific class that it does have?

test/testcafe/spec/TerminalView_spec.js
anipendakur-okta: The idea is to expand it to other type of authenticators' terminalviews by checking the i18n keys (and thus expand the same test file) OR cleaning up this code if we decide to have a proper remediation response that indicates what this terminalview corresponds to..

anipendakur-okta: Having separate test files seems like an overkill at this point..

test/testcafe/spec/GenericErrorHandling_spec.js
anipendakur-okta: It has these common classes `okta-sign-in-beacon-border auth-beacon-border`, however the intent here is that the error view does not have email specific class added by the method `AuthenticatorUtil. getIconClassNameForBeacon`, hence checking for `notContains`

src/v2/view-builder/views/TerminalView.js
anipendakur-okta: I've attached before and after screenshots..

vincentvoong-okta (changes requested):
In this terminal state since it is part of the Email verify flow we should use the same title for email verify. `Verify with Email Authentication`

A nit, the line that goes across the beacon is slightly off-center being too high.

thomasmaslen-okta:
> In this terminal state since it is part of the Email verify flow we should use the same title for email verify. `Verify with Email Authentication`

@vincentvoong-okta just checking to make sure whether we're all talking about the same thing when we say "verify":

This PR pertains to (setting the "beacon" icon for) 6 terminal states:

1. 1 of those states is for the `idx.email.verification.required` message.  This one occurs during IDX Activation when we need to verify the user's Email address -- so yes, this one does pertain to verification.
2. the other 5 states are used by the Email Authenticator (in email-with-link flows).  In these cases the user's Email address has already been verified (by the IDX Activation above), and instead the Email authenticator is being used either for authentication or for self-service password recovery.

When you say "the Email verify flow" I expect you're referring to 1. and _not_ to 2., but let me know if I'm misinterpreting.

anipendakur-okta:
> In this terminal state since it is part of the Email verify flow we should use the same title for email verify. Verify with Email Authentication
> 
> A nit, the line that goes across the beacon is slightly off-center being too high.

Hi @vincentvoong-okta , this PR is only for the beacon icon change for terminal views corresponding to the different email termination states as explained by Thomas above. For the title & the CSS change, would you mind opening up a ticket please? Keeping the timeline in mind (with the holidays), I'd like to keep the scope of the PR to what the ticket originally was meant for. I will address the title and css change in a separate PR post EA. Please let me know if any questions/concerns. Thanks!

test/testcafe/spec/TerminalView_spec.js
gowthamidommety-okta: ok, im fine either way. Just had to ask because the tests for safe mode that use this terminal view, are in a separate safe mode test spec.

src/v2/view-builder/views/TerminalView.js
```diff
@@ -2,10 +2,32 @@ import { createCallout, loc } from 'okta';
...
+  getBeaconClassName: function () {
+    return this.options.appState.containsMessageWithI18nKey(EMAIL_AUTHENTICATOR_TERMINAL_KEYS)
```
haishengwu-okta, line 25: nit:
prefer
```
  const classNames = super.getBeaconClassName();
  return containsMessageWithI18nKey() ? getIconClassNameForBeacon() : classNames;
```

PR #1529, opened 2020-11-23
feat: implements changes to support safe mode during authenticator enrollment

gowthamidommety-okta:
## Description:

safeMode

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-345300](https://oktainc.atlassian.net/browse/OKTA-345300)

### DIFF
--- a/playground/mocks/data/idp/idx/safe-mode-credential-enrollment-intent.json
+++ b/playground/mocks/data/idp/idx/safe-mode-credential-enrollment-intent.json
@@ -1,0 +1,22 @@
+{
+  "stateHandle": "02lXCU8zPR9jbdnipnWiPcyzdwQELBBW9K2oRm1mmt",
+  "version": "1.0.0",
+  "expiresAt": "2020-11-23T01:34:19.000Z",
+  "intent": "CREDENTIAL_ENROLLMENT",
+  "remediation": {
+    "type": "array",
+    "value": []
+  },
+  "messages": {
+    "type": "array",
+    "value": [
+      {
+        "message": "Set up is temporarily unavailable due to server maintenance. Click 'Back to Settings' to continue.",
+        "i18n": {
+          "key": "idx.error.server.safe.mode.cred.enroll"
+        },
+        "class": "ERROR"
+      }
+    ]
+  }
+}
--- a/playground/mocks/data/idp/idx/safe-mode-optional-enrollment.json
+++ b/playground/mocks/data/idp/idx/safe-mode-optional-enrollment.json
@@ -1,0 +1,41 @@
+{
+  "stateHandle": "02-20rHvS_j-Wq-odoJpMO33YcBwaL50217Nm9hysZ",
+  "version": "1.0.0",
+  "expiresAt": "2020-11-23T01:37:08.000Z",
+  "intent": "LOGIN",
+  "remediation": {
+    "type": "array",
+    "value": [
+      {
+        "rel": [
+          "create-form"
+        ],
+        "name": "skip",
+        "href": "http://orgw5iedi28k0g3.okta1.com:1802/idp/idx/skip",
+        "method": "POST",
+        "value": [
+          {
+            "name": "stateHandle",
+            "required": true,
+            "value": "02-20rHvS_j-Wq-odoJpMO33YcBwaL50217Nm9hysZ",
+            "visible": false,
+            "mutable": false
+          }
+        ],
+        "accepts": "application/ion+json; okta-version=1.0.0"
+      }
+    ]
+  },
+  "messages": {
+    "type": "array",
+    "value": [
+      {
+        "message": "Set up is temporarily unavailable due to server maintenance. Click 'Skip set up' to continue.",
+        "i18n": {
+          "key": "idx.error.server.safe.mode.assurance.satisfied"
+        },
+        "class": "ERROR"
+      }
+    ]
+  }
+}
--- a/src/v2/ion/responseTransformer.js
+++ b/src/v2/ion/responseTransformer.js
@@ -59,8 +59,10 @@ const getFirstLevelObjects = (resp) => {
 
 const getRemediationValues = (idx) => {
   const remediationValues = [];
-  if (_.isEmpty(idx.neededToProceed)) {
-    // no remediation but only success
+  const hasSkipRemediationOnly =
+    idx.neededToProceed.length === 1 && idx.neededToProceed[0].name === 'skip';
+  if (_.isEmpty(idx.neededToProceed) || hasSkipRemediationOnly) {
+    // no remediation or only skip remediation with success
     if (idx.context.success) {
       remediationValues.push({
         name: idx.context.success.name,
@@ -68,7 +70,7 @@ const getRemediationValues = (idx) => {
         value: []
       });
     } else if (idx.context.messages) {
-      // no remediation but only messages
+      // no remediation or only skip remediation with messages
       remediationValues.push({
         name: RemediationForms.TERMINAL,
         // Using `value` is unnecessary as `messages` will be display via `TerminalView.showMessages`,
@@ -78,7 +80,7 @@ const getRemediationValues = (idx) => {
         value: [],
       });
     } else if (idx.context.deviceEnrollment) {
-      // no remediation or messages for device enrollment state
+      // no remediation or only skip remediation with messages for device enrollment state
       // and the state is meant to be terminal state with different UI than the regular terminal view
       remediationValues.push({
         name: RemediationForms.DEVICE_ENROLLMENT_TERMINAL,
--- a/src/v2/view-builder/views/TerminalView.js
+++ b/src/v2/view-builder/views/TerminalView.js
@@ -2,9 +2,10 @@ import { createCallout, loc } from 'okta';
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
 import BaseFooter from '../internals/BaseFooter';
-import {getBackToSignInLink} from '../utils/LinksUtil';
+import { getBackToSignInLink, getSkipSetupLink } from '../utils/LinksUtil';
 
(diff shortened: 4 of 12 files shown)


src/v2/view-builder/views/TerminalView.js
```diff
@@ -49,6 +53,9 @@ const Footer = BaseFooter.extend({
     if (this.options.appState.containsMessageWithI18nKey(RETURN_LINK_EXPIRED_KEY)) {
       return getBackToSignInLink(this.options.settings);
     }
+    if (this.options.appState.containsMessageStartingWithI18nKey(SAFE_MODE_KEY)) {
```
magizh-okta, line 56: Can we add a derived variable in appstate to check this? 

Like 

'isSafeMode' 
and check if the remediation is just skip over there ?

magizh-okta (commented):
Looking good. Just once comment

playground/mocks/data/idp/idx/safe-mode-optional-enrollment.json
```diff
@@ -0,0 +1,41 @@
...
+        ],
+        "accepts": "application/ion+json; okta-version=1.0.0"
+      }
+    ]
+  },
+  "messages": {
+    "type": "array",
+    "value": [
+      {
+        "message": "Set up is temporarily unavailable due to server maintenance. Click 'Skip set up' to continue.",
```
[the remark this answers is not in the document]
anishashetty-okta: yes it is. Its a part of api.properties

src/v2/ion/responseTransformer.js
```diff
@@ -59,7 +59,9 @@ const getFirstLevelObjects = (resp) => {
 
 const getRemediationValues = (idx) => {
   const remediationValues = [];
-  if (_.isEmpty(idx.neededToProceed)) {
+  const hasSkipRemediationOnly =
```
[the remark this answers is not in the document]
anishashetty-okta: +1

anishashetty-okta (approved):
lgtm!

src/v2/ion/responseTransformer.js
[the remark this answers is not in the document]
mauriciocastillosilva-okta: yes, we need that comment too, and also we need to update the comments inside the `if` block, since all those assume `neededToProceed` was empty, but that is no longer the case.
Or another way to do this is to have another `if` for when we only have `skip`, instead of mixing them with the cases when we have an empty `neededToProceed`

playground/mocks/data/idp/idx/safe-mode-enduser-enrollment.json
```diff
@@ -0,0 +1,22 @@
+{
```
mauriciocastillosilva-okta, line 1: file name should be something like `safe-mode-credential-enrollment-intent.json` or so. SIW doesn't really care much about end user settings page, it care about what comes in the response. 
And well, this is similar to a required enrollment, in the sense that we have no way to move forward, but instead we need to cancel. This response does not have a cancel though... so another suggestion can be `safe-mode-credential-enrollment-no-cancel-no-skip.json`

src/v2/view-builder/views/TerminalView.js
```diff
@@ -2,9 +2,10 @@ import { createCallout, loc } from 'okta';
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
 import BaseFooter from '../internals/BaseFooter';
-import {getBackToSignInLink} from '../utils/LinksUtil';
+import { getBackToSignInLink, getSkipSetupLink } from '../utils/LinksUtil';
 
 const RETURN_LINK_EXPIRED_KEY = 'idx.return.link.expired';
+const SAFE_MODE_KEY = 'idx.error.server.safe.mode';
```
mauriciocastillosilva-okta, line 8: we should add some comments in the API code to know this should be the prefix for all error messages for safe mode
@anishashetty-okta

test/testcafe/spec/SafeMode_spec.js
```diff
@@ -0,0 +1,59 @@
...
+
+test.requestHooks(safeModeRequiredMock)('should display correct error and sign out link when server is in safe mode during required enrollment', async t => {
+  const terminalPage = await setup(t);
+  await terminalPage.waitForErrorBox();
+  await t.expect(terminalPage.getErrorMessages().getTextContent()).eql('Set up is temporarily unavailable due to server maintenance.Try again later.');
+  await t.expect(await terminalPage.skipSetUpLinkExists()).notOk();
+  await t.expect(await terminalPage.signoutLinkExists()).ok();
+});
+
+test.requestHooks(safeModeEnduserMock)('should display correct error and no links when server is in safe mode during enduser enrollment', async t => {
```
mauriciocastillosilva-okta, line 53: same comment here about end user settings enrollment, we should say instead `when server is in safe mode during credential enrollment INTENT` or something along those lines

src/v2/view-builder/views/TerminalView.js
magizh-okta: If we rely on the fact that just skip remediation -> safeMode then we dont have to worry about this.

anishashetty-okta: Ack. will update my PR

playground/mocks/data/idp/idx/safe-mode-enduser-enrollment.json
```diff
@@ -0,0 +1,22 @@
...
+  "intent": "CREDENTIAL_ENROLLMENT",
+  "remediation": {
+    "type": "array",
+    "value": []
+  },
+  "messages": {
+    "type": "array",
+    "value": [
+      {
+        "message": "Set up is temporarily unavailable due to server maintenance. Click 'Back to Settings' to continue.",
```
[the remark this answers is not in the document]
anishashetty-okta: I think @vincentvoong-okta 's idea was wrt to the end user settings page. you can click back to settings to proceed to the settings page. Click 'Back to Settings' to continue: sounds okay to me.
@vincentvoong-okta @thanh-hanguyen-okta ^^

src/v2/ion/responseTransformer.js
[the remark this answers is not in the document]
gowthamidommety-okta: Makes sense.. Let me look at this again

src/v2/view-builder/views/TerminalView.js
gowthamidommety-okta: We only know if it is in safe mode based on the error message key. There may be a skip remediation, or it may be empty

src/v2/view-builder/views/TerminalView.js
gowthamidommety-okta: @magizh-okta we may or may not have skip remediation.. our only signal is the error message key

vincentvoong-okta (approved):
LGTM

src/v2/ion/responseTransformer.js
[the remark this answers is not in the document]
gowthamidommety-okta: I think it makes sense where it is. We currently don't have a form for the skip remediation, it shows a default view. We need to check the presence of messages anyway to get to the terminal view. This is why i think saying if either remediation is empty (neededToProceed is empty) or skip is the only remediation, we should handle it the same. Will update comments in the code. cc @haishengwu-okta

playground/mocks/data/idp/idx/safe-mode-enduser-enrollment.json
gowthamidommety-okta: Yea, you have a good point. I named this to test the flow when coming from end user with no cancel and no skip links. So probably `safe-mode-credential-enrollment-no-cancel-no-skip.json` is a good option

gowthamidommety-okta: Talked to Mau offline. Naming according to intent sounds good 👍

src/v2/view-builder/views/TerminalView.js
magizh-okta: @gowthamidommety-okta Is there any harm in showing the link always rather thank checking the i18n key.

gowthamidommety-okta: Multiple flows use this view. my only concern is that the skip link will show up there. This check is what email is doing above too..

magizh-okta (approved):
LGTM 👍

src/v2/view-builder/views/TerminalView.js
mauriciocastillosilva-okta, line 8: nit: change this to `SAFE_MODE_KEY_PREFIX`

playground/mocks/data/idp/idx/safe-mode-credential-enrollment-intent.json
```diff
@@ -0,0 +1,22 @@
...
+  "intent": "CREDENTIAL_ENROLLMENT",
+  "remediation": {
+    "type": "array",
+    "value": []
+  },
+  "messages": {
+    "type": "array",
+    "value": [
+      {
+        "message": "Set up is temporarily unavailable due to server maintenance. Click 'Back to Settings' to continue.",
```
mauriciocastillosilva-okta, line 14: I want to discuss this a little more. "Back to settings" is only available if the SIW is okta-hosted and they are in our settings page. If at some point a custom hosted widget is boostraped with intent=CREDENTIAL_ENROLLMENT, this error message won't make sense, since "Back to settings" link lives on our page that hosts the widget, not within the widget itself. 

I think the error should be something like "Try again later" only
@anishashetty-okta let me know offline where is the backend PR, I can make comments there

mauriciocastillosilva-okta (approved):
Just one nit, if you can fix it that will be great!
then 🐑

gowthamidommety-okta:
opened another branch on main so closing this one (https://github.com/okta/okta-signin-widget/pull/1530)

PR #1530, opened 2020-11-24
feat: implements changes to support safe mode during authenticator en…

gowthamidommety-okta:
…rollment

Resolves: OKTA-345300

## Description:

See https://github.com/okta/okta-signin-widget/pull/1529 for approval

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-345300](https://oktainc.atlassian.net/browse/OKTA-345300)

### DIFF
--- a/src/v2/view-builder/utils/LinksUtil.js
+++ b/src/v2/view-builder/utils/LinksUtil.js
@@ -92,6 +92,21 @@ const goBackLink = (appState) => {
   return [];
 };
 
+const getSkipSetupLink = (appState) => {
+  if (appState.hasRemediationObject(RemediationForms.SKIP)) {
+    return [
+      {
+        'type': 'link',
+        'label': loc('oie.enroll.skip.setup', 'login'),
+        'name': 'skip-setup',
+        'actionPath': RemediationForms.SKIP,
+      }
+    ];
+  }
+
+  return [];
+};
+
 const getSignOutLink = (settings) => {
   if (settings?.get('signOutLink')) {
     return [
@@ -130,5 +145,6 @@ export {
   getForgotPasswordLink,
   goBackLink,
   getSignOutLink,
-  getBackToSignInLink
+  getBackToSignInLink,
+  getSkipSetupLink
 };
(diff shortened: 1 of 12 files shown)


magizh-okta (approved):

src/v2/view-builder/utils/LinksUtil.js
```diff
@@ -99,7 +99,7 @@ const getSkipSetupLink = (appState) => {
         'type': 'link',
         'label': loc('oie.enroll.skip.setup', 'login'),
         'name': 'skip-setup',
-        'formName': RemediationForms.SKIP,
+        'actionPath': RemediationForms.SKIP,
```
mauriciocastillosilva-okta, line 102: Why is this not a `formName`? From the response I can see that it is inside the remediation object...

I can see that we don't want this to submit a form, but instead just follow the link. So I am guessing that is the reason, which makes sense. With that said, then I think we should not have `RemediationForms` constants assigned to an `actionPath`, those should probably be used only for `formNames`
I see we already have a mix in our codebase though... so as a nit, maybe we change it to:
```
'actionPath': 'skip',
```
which is what we already have in [ReEnrollAuthenticatorWarningPasswordView](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/password/ReEnrollAuthenticatorWarningPasswordView.js#L46)

mauriciocastillosilva-okta (approved):
just one nit, rest looks good

src/v2/view-builder/utils/LinksUtil.js
gowthamidommety-okta: Yea i was thinking about this too.. I thought the action path needs to be the form name (https://github.com/okta/okta-signin-widget/blob/master/src/v2/controllers/FormController.js#L113), so I set it to this constant. Will change 👍

gowthamidommety-okta: Talked to Mau, leaving as is for now.. Will discuss with the team later

haishengwu-okta: It shall be `actionPath`.
- `actionPath` :  find the action from remediation list or other top level object and submit request
- `formName` : re-render page using this *remedation* form.

PR #1533, opened 2020-11-25
feat: idx interaction code flow integration

shuowu:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:


src/v2/BaseLoginRouter.js
```diff
@@ -32,11 +32,21 @@ import AppState from './models/AppState';
...
+const introspect = (settings) => {
+  const clientId = settings.get('clientId');
   const domain = settings.get('baseUrl');
+  const scopes = settings.get('scopes');
   const stateHandle = settings.get('stateToken');
+  const redirectUri = settings.get('redirectUri');
   const version = settings.get('apiVersion');
-  return idx.start({ domain, stateHandle, version });
+  
+  if (settings.get('useInteractionCodeFlow')) {
```
swiftone, line 43: I would check stateHandle first - A customized-okta-hosted widget could be configured with this switch, but we would want to maintain the okta-hosted flow.

src/v2/BaseLoginRouter.js
```diff
@@ -32,11 +32,21 @@ import AppState from './models/AppState';
 import idx from 'idx';
 
 
-const introspectStateToken = (settings) => {
+const introspect = (settings) => {
```
pradeepdewda-okta, line 35: I would like to maintain the separation between introspect and interact calls..

create two functions introspect and interact that returns idx.start()

create another calling function startLoginFlow() that calls the correct function from above

src/v2/BaseLoginRouter.js
```diff
@@ -32,11 +32,21 @@ import AppState from './models/AppState';
...
+  const clientId = settings.get('clientId');
   const domain = settings.get('baseUrl');
+  const scopes = settings.get('scopes');
   const stateHandle = settings.get('stateToken');
+  const redirectUri = settings.get('redirectUri');
   const version = settings.get('apiVersion');
-  return idx.start({ domain, stateHandle, version });
+  
+  if (stateHandle) {
```
pradeepdewda-okta, line 43: what if `stateHandle` and `useInteractionCodeFlow` both are present ? In that case we always call `introspect` ?

src/v2/BaseLoginRouter.js
```diff
@@ -32,11 +32,21 @@ import AppState from './models/AppState';
...
+    return idx.start({ domain, stateHandle, version });
+  } else if (settings.get('useInteractionCodeFlow')) {
+    return idx.start({ clientId, domain, scopes, stateHandle, redirectUri, version });
+  } else {
+    throw new Errors.ConfigError('Set "useInteractionCodeFlow" to true in configuration to enable the "interaction_code" flow.');
```
pradeepdewda-okta, line 48: need a better to prompt for this warning.. This should not be an error!!

src/v2/BaseLoginRouter.js
```diff
@@ -162,9 +179,11 @@ export default Router.extend({
     // introspect stateToken when widget is bootstrap with state token
     // and remove it from `settings` afterwards as IDX response always has
     // state token (which will be set into AppState)
-    if (this.settings.get('stateToken') || this.settings.get('proxyIdxResponse')) {
+    if (this.settings.get('stateToken') 
+        || this.settings.get('proxyIdxResponse')
+        || this.settings.get('useInteractionCodeFlow')) {
       const idxRespPromise = this.settings.get('proxyIdxResponse') ?
-        handleProxyIdxResponse(this.settings) : introspectStateToken(this.settings);
+        handleProxyIdxResponse(this.settings) : introspect(this.settings);
```
pradeepdewda-okta, line 186: call `startLoginFlow` here instead of `introspect`

src/widget/OktaSignIn.js
```diff
@@ -153,7 +153,9 @@ var OktaSignIn = (function () {
     var authClient = createAuthClient(authParams);
 
     var Router;
-    if ((options.stateToken && !Util.isV1StateToken(options.stateToken)) || options.proxyIdxResponse) {
+    if ((options.stateToken && !Util.isV1StateToken(options.stateToken)) 
+        || options.useInteractionCodeFlow
```
pradeepdewda-okta, line 157: at this moment we don't know if IDX flag is enabled or not... I think if stateToken exists, then only we can se IDX flag was enabled and hence we can use `useInteractionCodeFlow`..

src/v2/BaseLoginRouter.js
shuowu: The error here is for devs only, I assume the above two cases are the only cases we should support.

src/v2/BaseLoginRouter.js
```diff
@@ -32,11 +32,21 @@ import AppState from './models/AppState';
...
+  const scopes = settings.get('scopes');
   const stateHandle = settings.get('stateToken');
+  const redirectUri = settings.get('redirectUri');
   const version = settings.get('apiVersion');
-  return idx.start({ domain, stateHandle, version });
+  
+  if (stateHandle) {
+    return idx.start({ domain, stateHandle, version });
+  } else if (settings.get('useInteractionCodeFlow')) {
+    return idx.start({ clientId, domain, scopes, stateHandle, redirectUri, version });
```
swiftone, line 46: This should be issuer, not domain, if we have issuer.  (the stateHandle version can continue to use domain, or can also switch over to issuer)

src/widget/OktaSignIn.js
shuowu: The `useInteractionCodeFlow` runs the interaction code flow, which boostraps without stateToken. Should the logic be if any of the conditions satisfied, then use V2Router to use idx?

src/v2/BaseLoginRouter.js
shuowu: I think Brett's comment can answer your question: https://github.com/okta/okta-signin-widget/pull/1533#discussion_r530637890

src/v2/BaseLoginRouter.js
swiftone: interact is called inside idx.start(), so the widget doesn't explicitly call it.  If you mean two function calls to define the different params to idx.start(), then ignore this comment from me.

src/widget/OktaSignIn.js
pradeepdewda-okta: This issue is we need to check that IDX flag was enabled or not.. Till now we expected that if `stateToken` exists then IDX flag is enabled by default. But now it seems it's not the case.

src/v2/BaseLoginRouter.js
swiftone: If both are present, we should behave as if just stateHandle is present and ignore useInteractionCodeFlow.   This should only happen if there is a customized-okta-hosted widget, and in such a case we want to end with a redirect, so ignore useInteractionCodeFlow.

pradeepdewda-okta: @bretterer do we always call `introspect` with both exists ?

src/widget/OktaSignIn.js
shuowu: like a feature flag for `idx`? if there is one, only that flag should be enough here.

src/v2/BaseLoginRouter.js
pradeepdewda-okta: yes, for the purpose of clarity for devs.

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -402,9 +401,65 @@ Expect.describe('OktaSignIn v2 bootstrap', function () {
...
+    itp('throws an error if invalid version is passed to idx-js', function () {
+      return setupLoginFlow({
         apiVersion: '2.0.0',
-      }).catch(err => {
+        clientId: 'someClientId',
+        redirectUri: 'http://0.0.0.0:9999',
+        useInteractionCodeFlow: true
+      }, responses).catch(err => {
         expect(err.name).toBe('CONFIG_ERROR');
         expect(err.message.toString()).toEqual('Error: Unknown api version: 2.0.0.  Use an exact semver version.');
```
pradeepdewda-okta, line 464: can we get the exact api version number here which is being used ? instead of hardcoding

swiftone (approved):

src/v2/BaseLoginRouter.js
```diff
@@ -160,18 +210,20 @@ export default Router.extend({
...
-    if (this.settings.get('stateToken') || this.settings.get('proxyIdxResponse')) {
+    if (this.settings.get('oieEnabled')) {
       const idxRespPromise = this.settings.get('proxyIdxResponse') ?
-        handleProxyIdxResponse(this.settings) : introspectStateToken(this.settings);
+        handleProxyIdxResponse(this.settings) : startLoginFlow(this.settings);
       return idxRespPromise
         .then(idxResp => {
           this.settings.unset('stateToken');
           this.settings.unset('proxyIdxResponse');
+          this.settings.unset('useInteractionCodeFlow');
```
pradeepdewda-okta, line 220: Can you please create a new `unsetAll` method as well, that can take an array and unset all..

src/v2/BaseLoginRouter.js
```diff
@@ -160,18 +210,20 @@ export default Router.extend({
     // introspect stateToken when widget is bootstrap with state token
     // and remove it from `settings` afterwards as IDX response always has
     // state token (which will be set into AppState)
-    if (this.settings.get('stateToken') || this.settings.get('proxyIdxResponse')) {
+    if (this.settings.get('oieEnabled')) {
```
pradeepdewda-okta, line 213: thanks for this.. 🥇

PR #1536, opened 2020-11-30
Update package.json

### LINKED ISSUE #1520: Include source code with source maps
## About

Hey there! It would be super helpful for debugging purposes if you could include source code when publishing libraries 🙏 

Right now, the source maps aren't doing a whole lot because they point to source files that don't exist.

andyrichardson:
Fix #1520

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
     "url": "https://github.com/okta/okta-signin-widget/issues"
   },
   "files": [
-    "dist"
+    "dist",
+    "src"
   ],
   "browserslist": [
     "> 0.1%",


oleksandrpravosudko-okta:
hi @andyrichardson - thanks for you contribution
SIW's source map files already contain non-minified source code - including it into package doesn't seem to be necessary
are you able to provide more details on the environment in which source maps are not working as expected?

andyrichardson:
@oleksandrpravosudko-okta thanks for the response - just saw the `sourceContent` property.

I'm not too sure what the issue was - I'll see if I can get a reproduction at some point. From what I remember, when clicking the link for the source file in an error from the package, the resulting file was minified and chrome didn't seem to detect a source map.

PR #1548, opened 2020-12-03
fix: signinwidget lose context of idx pipeline in case of browser refresh

pradeepdewda-okta:
## Description:

When browser refreshes in the middle of the pipeline, SIW gets new token and looses context and state of pipeline.
To support the url refresh, store stateToken in sessionStorage. If user closes the tab, then sessionStorage gets cleared.

When to clean up sessionStorage
1. When login flow ends with success
2. When login flow ends at terminal state
3. When login flow ends with Cancel remediation
4. When login flow ends with user error, where only option is to refresh the page. for eg temrinal form errors.
5. At any success redirects

SIW stores the stateToken and clean it up on pipeline completions, that could be any from success, error, terminal states, cancel or redirects.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-332504](https://oktainc.atlassian.net/browse/OKTA-332504)


swiftone:
If okta-core is doing the storage, doesn't that mean the bug will continue to exist for customer-hosted widgets?

pradeepdewda-okta:
> If okta-core is doing the storage, doesn't that mean the bug will continue to exist for customer-hosted widgets?

Yes, I wanted to talk about that here. This is early work.

src/v2/BaseLoginRouter.js
```diff
@@ -95,6 +95,11 @@ export default Router.extend({
   },
 
   handleIdxResponseSuccess (idxResponse) {
+    //On success flush sessionStorage, 'success' object in remediation signifies end of login flow
+    if(idxResponse.rawIdxState.success || idxResponse.rawIdxState.successWithInteractionCode) {
+      sessionStorage.clear();
```
haishengwu-okta, line 100: seems too brutal since it could clear up other data from session storage that are set by other applications.

pradeepdewda-okta: Yup, @aarongranick-okta mentioned this in chat, i would just remove the item by name.

src/v2/BaseLoginRouter.js
```diff
@@ -31,10 +31,14 @@ import AppState from './models/AppState';
 import idx from 'idx';
 
 const startLoginFlow = (settings) => {
+  let stateHandle = sessionStorage.getItem('stateTokenInUse');
```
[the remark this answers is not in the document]
pradeepdewda-okta: Would it make sense to keep this not as clear as possible for security reasons ? for eg. just `widget-state-handle` ?

src/v2/BaseLoginRouter.js
```diff
@@ -207,25 +217,33 @@ export default Router.extend({
...
+      this.settings.unset('stateToken');
+      this.settings.unset('proxyIdxResponse');
+      this.settings.unset('useInteractionCodeFlow');
+      this.appState.trigger('remediationSuccess', idxResp);
+      this.render(Controller, options);
+    };
+
+    const renderErrorCallback = (errorResp) => {
+      this.settings.unset('stateToken');
+      this.settings.unset('useInteractionCodeFlow');
```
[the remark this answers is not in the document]
pradeepdewda-okta: Actually, I am not sure about this @yannongli-okta added.. It should be cleared in error as well i suppose, until if there is a special case.

src/v2/BaseLoginRouter.js
```diff
@@ -207,25 +217,33 @@ export default Router.extend({
...
-        })
+        .then(renderSuccessCallback)
         .catch(errorResp => {
-          this.settings.unset('stateToken');
-          this.settings.unset('useInteractionCodeFlow');
-          this.appState.trigger('remediationError', errorResp.error);
-          this.render(Controller, options);
+          // Try again if first introspect fails due to invalid token error, now with new token.
+          sessionStorage.setItem('stateTokenInUse', this.settings.get('stateToken'));
+          startLoginFlow(this.settings).then(renderSuccessCallback).catch(renderErrorCallback);
```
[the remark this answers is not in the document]
pradeepdewda-okta: Great points. I will update the code.

src/v2/BaseLoginRouter.js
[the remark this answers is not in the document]
yannongli-okta: I might miss this part as i was mainly testing the successful flow. Can you please help add the cleaning logic?

src/v2/controllers/FormController.js
```diff
@@ -167,6 +171,10 @@ export default Controller.extend({
   },
 
   showFormErrors (model, error) {
+    // If errors are terminal errors, then clear session storage, so that intentional hard refersh initiate a new loginflow
+    if(error.messages) {
```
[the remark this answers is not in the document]
pradeepdewda-okta: While testing in November, I figured there are some flows that ends up with form errors in a way that only thing you can do is to hard refresh the page.. I don't exactly remember what was the flow, but happens when user is locked out but still trying to login or password needs to be reset. If i came across that again i will update. But this is required.

src/v2/view-builder/views/TerminalView.js
```diff
@@ -54,6 +54,8 @@ const Body = BaseForm.extend({
 
       messagesObjs.value
         .forEach(messagesObj => {
+          // To support intentional hard refresh in case of terminal states, clear sessionStorage to initiate a new login flow
+          sessionStorage.removeItem('stateTokenInUse');
```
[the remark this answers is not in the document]
pradeepdewda-okta: make sense. I will move it.

src/widget/OktaSignIn.js
```diff
@@ -141,6 +141,11 @@ var OktaSignIn = (function () {
         See: https://developer.okta.com/code/javascript/okta_sign-in_widget#cdn
       `);
 
+    // Store state token to reuse in case of browser refresh.
+    if(!sessionStorage.getItem("stateTokenInUse")) {
+      sessionStorage.setItem("stateTokenInUse", options.stateToken);
```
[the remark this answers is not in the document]
pradeepdewda-okta: Good points. Let me see, if `initialize` method is good entry for session storage in BaseLoginRouter.

src/v2/BaseLoginRouter.js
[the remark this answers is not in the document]
pradeepdewda-okta: @yannongli-okta Cool. I will add that, if you don't think there wasn't any special error handling case when response was a proxy.

src/v2/BaseLoginRouter.js
```diff
@@ -207,25 +222,32 @@ export default Router.extend({
         });
     }
 
-    // introspect stateToken when widget is bootstrap with state token
-    // and remove it from `settings` afterwards as IDX response always has
-    // state token (which will be set into AppState)
+    // Remove stateToken and other attributes to sanitize settings object.
+    // IDX response always has stateToken, that will be set into AppState)
+    const renderCallback = (idxResp) => {
+      this.unsetAttributes();
```
shuowu, line 228: We may also want to check the existence and set `widget-state-token` here, as the custom hosted interaction code flow starts without stateToken. The first available one will be here (when idx response resolves).

src/v2/BaseLoginRouter.js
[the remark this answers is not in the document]
aarongranick-okta: security is worth considering. But I also think it is nice to namespace all our storage so it is clear to our customers and support what we own/manage vs. what their app is doing. In this case I think it would be better to keep the `okta-` prefix, matching other storage that we have for things like tokens (which are more sensitive from a security perspective)

src/v2/BaseLoginRouter.js
pradeepdewda-okta: nice catch 👍

src/v2/BaseLoginRouter.js
```diff
@@ -207,25 +222,32 @@ export default Router.extend({
...
+      if (!idxResp.error) {
+        this.appState.trigger('remediationSuccess', idxResp);
```
pradeepdewda-okta, line 230: @haishengwu-okta should i clear the storage here for all terminal messages ? instead of the view ? Wondering there are no other flow then terminal ends up here..

src/v2/controllers/FormController.js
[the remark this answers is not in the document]
haishengwu-okta: Are you sure that whenever `error.messages` exists, it's terminal error?
I had vague memory that `showFormErrors` is general error handler (e.g. in case of wrong password, it invokes `showFormErrors` to render form level error message.)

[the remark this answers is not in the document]
pradeepdewda-okta: Yeah, if error is at top level it is should be terminal state error. All form level errors will be present inside remediation object at field level.

src/v2/view-builder/views/TerminalView.js
```diff
@@ -50,8 +50,10 @@ const Body = BaseForm.extend({
   showMessages () {
     const messagesObjs = this.options.appState.get('messages');
     if (messagesObjs && Array.isArray(messagesObjs.value)) {
+      // To support intentional hard refresh in case of terminal states
+      // clear sessionStorage to initiate a new login flow
+      sessionStorage.removeItem('widget-state-token');
```
aarongranick-okta, line 55: Why does this logic depend on whether there are any messages? Is there a special case of being terminal with no messages?

pradeepdewda-okta: as far as i remember, if the top level idx response contains `message` object then it's a terminal error. i think messagesObjs should always be non-empty in terminal error case.

pradeepdewda-okta: For eg this one.. It's a terminal error and only choice now user has is it to refresh.. @haishengwu-okta also this errors shows up from FormController.js at line 175
Screen Shot 2021-01-11 at 11 30 16 AM

src/v2/controllers/FormController.js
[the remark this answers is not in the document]
haishengwu-okta: Here one example I was worrying. Please correct me wrong.

1. introspect gets `authenticator-verification-email.json`
2. user enter wrong OTP code 
3. server 403 with `error-email-verify.json`
4. widget invokes this `showFormErrors` to show form error and the error has `error.messages`

Don't think we shall clear session storage for this case, right?

src/v2/view-builder/views/TerminalView.js
```diff
@@ -50,8 +50,10 @@ const Body = BaseForm.extend({
   showMessages () {
     const messagesObjs = this.options.appState.get('messages');
     if (messagesObjs && Array.isArray(messagesObjs.value)) {
+      // To support intentional hard refresh in case of terminal states
+      // clear sessionStorage to initiate a new login flow
+      sessionStorage.removeItem('okta-siw-state-token');
```
haishengwu-okta, line 55: Looks like we can clear session storage independently with error messages.
Therefore we can override `postRender` of BaseView and clear up session there. (around line 85 below)

src/widget/OktaSignIn.js
```diff
@@ -140,7 +140,7 @@ var OktaSignIn = (function () {
         When you are ready to publish your app, embed the minified version to turn on production mode.
         See: https://developer.okta.com/code/javascript/okta_sign-in_widget#cdn
       `);
-
+
```
haishengwu-okta, line 143: revert this.

test/testcafe/spec/StateTokenRefreshUseCases_spec.js
```diff
@@ -0,0 +1,75 @@
...
+  return identityPage;
+}
+
+test.requestHooks(identifyRequestLogger, identifyMock)('should set and remove statetoken in sessionStorage', async t => {
+  const identityPage = await setup(t);
+  let stateToken = await t.eval(() => sessionStorage.getItem('okta-siw-state-token'));
+  await t.expect(stateToken !== null).eql(true);
+  await identityPage.fillIdentifierField('Test Identifier');
+  await identityPage.clickNextButton();
+  // On terminal error state token should be cleared
```
haishengwu-okta, line 48: It's unclear to me that we need to clear session storage at this point cause there are many reasons that identify step got error. (one example I would guess is user put wrong username and backend responds error, but user can correct that with right user name)

test/testcafe/spec/StateTokenRefreshUseCases_spec.js
```diff
@@ -0,0 +1,75 @@
...
+  return identityPage;
+}
+
+async function setupA(t) {
+  const identityPage = new SelectAuthenticatorPageObject(t);
+  await identityPage.navigateToPage();
+  return identityPage;
+}
+
+test.requestHooks(identifyRequestLogger, identifyMock)('should set and remove statetoken in sessionStorage', async t => {
```
haishengwu-okta, line 42: you don't need `identifyRequestLogger` if you didn't utilize logger result in the test.

src/v2/BaseLoginRouter.js
```diff
@@ -207,25 +217,34 @@ export default Router.extend({
         });
     }
 
-    // introspect stateToken when widget is bootstrap with state token
-    // and remove it from `settings` afterwards as IDX response always has
-    // state token (which will be set into AppState)
+    // Remove stateToken and other attributes to sanitize settings object.
+    // IDX response always has stateToken, that will be set into AppState)
+    const renderCallback = (idxResp) => {
```
haishengwu-okta, line 222: It seems more clear to have separated handler (success & error)

src/v2/BaseLoginRouter.js
```diff
@@ -253,6 +272,18 @@ export default Router.extend({
...
   },
 
+  unsetAttributes () {
+    this.settings.unset('stateToken');
+    this.settings.unset('useInteractionCodeFlow');
+    this.settings.unset('proxyIdxResponse');
+  },
+
+  updateStateHandleInStorage (stateToken) {
+    if(!sessionStorage.getItem('okta-siw-state-token') && stateToken) {
```
haishengwu-okta, line 282: Create utility which has
- constant for string `okta-siw-state-token`
- read/write/delete state token from session storage

Or
- constant for string `okta-siw-state-token` into `src/v2/view-builder/utils/Constants.js` and move it to upper folder like `src/v2`

src/v2/controllers/FormController.js
```diff
@@ -109,6 +109,10 @@ export default Controller.extend({
   },
 
   handleInvokeAction (actionPath = '') {
+    if(actionPath === 'cancel') {
```
haishengwu-okta, line 112: Curious what's the response of `cancel`. Any examples?

haishengwu-okta (commented):
How about clear up session storage only at
- `SuccessView.postRender`
- `TerminalView.postRender`

src/v2/BaseLoginRouter.js
pradeepdewda-okta: It made my code at line 239 much easier to read.. also the handlers are just one time use.

src/v2/controllers/FormController.js
pradeepdewda-okta: Any sign out after entering username is a cancel.. Any other actionPath could be `cancel` ?
Screen Shot 2021-01-12 at 10 25 54 AM

src/v2/controllers/FormController.js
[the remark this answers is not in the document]
pradeepdewda-okta: It doesn't.. But i see your point, we need a better way to cleanup storage from a single source.. i am exploring postRender suggestion.. 

Screen Shot 2021-01-12 at 10 31 36 AM

test/testcafe/spec/StateTokenRefreshUseCases_spec.js
pradeepdewda-okta: yeah make sense.. changed it to success case instead of user error.

src/v2/view-builder/views/TerminalView.js
pradeepdewda-okta: as discussed, added to Terminal and Success Views..

src/v2/controllers/FormController.js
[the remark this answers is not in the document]
pradeepdewda-okta: removed it.

src/v2/BaseLoginRouter.js
pradeepdewda-okta: changed, but i will move it to `src/v2` in a different PR.. It's been used at various places.

src/v2/BaseLoginRouter.js
pradeepdewda-okta: changed.

shuowu-okta (approved):
Interaction code flow part LGTM!

aarongranick-okta (approved):

swiftone (approved):

src/v2/BaseLoginRouter.js
```diff
@@ -253,6 +274,18 @@ export default Router.extend({
     this.controller.render();
   },
 
+  unsetAttributes () {
+    this.settings.unset('stateToken');
+    this.settings.unset('useInteractionCodeFlow');
+    this.settings.unset('proxyIdxResponse');
+  },
+
+  updateStateHandleInStorage (stateToken) {
```
haishengwu-okta, line 283: setStateHandleInStorage or addStateHandleInStorage

src/v2/view-builder/views/SuccessView.js
```diff
@@ -44,5 +46,9 @@ const Body = BaseForm.extend({
 });
 
 export default BaseView.extend({
-  Body
+  Body,
+
+  postRender () {
+    sessionStorage.removeItem(OKTA_STATE_TOKEN_KEY);
```
haishengwu-okta, line 52: always call super.postRender first.

src/v2/view-builder/views/TerminalView.js
```diff
@@ -86,5 +85,8 @@ export default BaseView.extend({
     HeaderBeacon: HeaderBeaconTerminal,
   }),
   Body,
-  Footer
+  Footer,
+  postRender () {
+    sessionStorage.removeItem(OKTA_STATE_TOKEN_KEY);
```
haishengwu-okta, line 90: ditto.

src/v2/BaseLoginRouter.js
```diff
@@ -149,6 +154,12 @@ export default Router.extend({
       uiSchemaTransformer,
       responseTransformer.bind({}, this.settings),
     )(idxResponse);
+
+    //On success flush sessionStorage, 'success' object in remediation signifies end of login flow
+    if(idxResponse.rawIdxState.success || idxResponse.rawIdxState.successWithInteractionCode) {
```
haishengwu-okta, line 159: do you still need this given the logic added to SuccessView?

src/v2/BaseLoginRouter.js
```diff
@@ -29,12 +29,16 @@ import uiSchemaTransformer from './ion/uiSchemaTransformer';
 import i18nTransformer from './ion/i18nTransformer';
 import AppState from './models/AppState';
 import idx from 'idx';
+import { OKTA_STATE_TOKEN_KEY } from './view-builder/utils/Constants';
 
 const startLoginFlow = (settings) => {
+  let stateHandle = sessionStorage.getItem(OKTA_STATE_TOKEN_KEY);
+  if (!stateHandle) {
+    stateHandle = settings.get('stateToken');
+  }
```
haishengwu-okta, line 38: Given the code in `initialize` function: `this.updateStateHandleInStorage`, is this check redundant ?

pradeepdewda-okta: Yeah that's correct it's redundant, i added it earlier, wanted to be extra careful. But now we have it in `initialize` method we can remove it.

src/v2/BaseLoginRouter.js
pradeepdewda-okta: Here also wanted to be extra careful.. I think i should remove it.

src/v2/view-builder/views/SuccessView.js
pradeepdewda-okta: Very good catch, could have been easily missed.

src/v2/BaseLoginRouter.js
```diff
@@ -29,12 +29,19 @@ import uiSchemaTransformer from './ion/uiSchemaTransformer';
...
+import { OKTA_STATE_TOKEN_KEY } from './view-builder/utils/Constants';
+
+const setStateHandleInStorage = (stateToken) => {
+  if(!sessionStorage.getItem(OKTA_STATE_TOKEN_KEY) && stateToken) {
+    sessionStorage.setItem(OKTA_STATE_TOKEN_KEY, stateToken);
+  }
+};
 
 const startLoginFlow = (settings) => {
+  let stateHandle = sessionStorage.getItem(OKTA_STATE_TOKEN_KEY);
```
aarongranick-okta, line 41: the stateHandle is being loaded from storage, but the codeVerifier will be re-generated ahead on line 51. Do we need to save the codeVerifier also? If we use a new codeVerifier we will be doing a new PKCE flow. Maybe we want that?

pradeepdewda-okta: interact will only be called when there is no stateToken (either from client or from storage).. So interact call will always be new.

PR #1549, opened 2020-12-07
fix: sign in button text wrap

ivandenysenko-okta:
sign in button text wrap

OKTA-350404

two line version:
Screen Shot 2020-12-08 at 1 10 48 AM

one line version:
Screen Shot 2020-12-08 at 5 06 12 PM

Screen Shot 2020-12-08 at 7 30 47 PM

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-350404](https://oktainc.atlassian.net/browse/OKTA-350404)

### DIFF
--- a/assets/sass/v2/_identify.scss
+++ b/assets/sass/v2/_identify.scss
@@ -1,7 +1,7 @@
 .sign-in-with-device-option {
   .okta-verify-container {
     .okta-verify-authenticator {
-      margin-right: 0.83em;
+      margin: 0 0.83em 0 0;
     }
   }
 }
@@ -23,10 +23,17 @@
       margin-bottom: 0.83em;
     }
     .button {
-      display: block;
+      display: flex;
       position: relative;
-      height: 50px;
-      line-height: 50px;
+      min-height: 50px;
+      height: auto;
+      padding-top: 10px;
+      padding-bottom: 10px;
+      line-height: 1.4;
+      align-items: center;
+      .okta-verify-authenticator {
+        flex-shrink: 0;
+      }
     }
   }
   .separation-line {


stevennguyen-okta (approved):
If we can go with the wrap + 5px top bottom that'd be good

pradeepdewda-okta:
@ivandenysenko-okta seems to be a very long text for a button.. If we support this, then in future people can create longer texts for button.. @stevennguyen-okta should we get a broader design review for this ?

assets/sass/v2/_identify.scss
```diff
@@ -23,10 +23,17 @@
       margin-bottom: 0.83em;
     }
     .button {
-      display: block;
+      display: flex;
       position: relative;
-      height: 50px;
-      line-height: 50px;
+      min-height: 60px;
```
jhe-okta, line 28: With this change, in the device with bigger screen that only one line is displayed , the button's height is 60px instead of 50px. Is this expected, @stevennguyen-okta ?

stevennguyen-okta: That is not expected, preferably the one-line button would match its existing height and only change to accommodate the button wrap.

stevennguyen-okta:
> @ivandenysenko-okta seems to be a very long text for a button.. If we support this, then in future people can create longer texts for button.. @stevennguyen-okta should we get a broader design review for this ?

We can followup post EA, as we're already too close and want to freeze it as much as possible. We reviewed/discussed this button a lot and the problem is if you removed "on this device" for example, it becomes less accurate of a descriptor and can make it sound like an express button to TOTP or Push on another device. If we change it to "Use" instead of "Sign in using" then it becomes more ambiguous of an action on this screen. We also didn't want to brand the action as "Fastpass" because we can't know if the user will experience a FastPass signin (zero interaction). It's kind of a tricky situation and we opted for as much clarity as possible for EA to be safe.

assets/sass/v2/_identify.scss
ivandenysenko-okta: made changes.
one line button is 50px now

pradeepdewda-okta:
@ivandenysenko-okta can you add screenshots of any other screen resolution where it's going to display the most and see how it looked.

ivandenysenko-okta:
@pradeepdewda-okta added iPhoneSE/5 emulation - the most common small resolution case

pradeepdewda-okta (approved):

jhe-okta (approved):

PR #1553, opened 2020-12-07
feat: updates sign on success message

sherizada-okta:
## Description:
Changes text from "You are being redirected" to "Signing in to {appInstanceName} as {identifier}" or "Signing in to {appInstanceName}" if no identifier has been set. 

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:
App set and User identifier set:
Screen Shot 2020-12-10 at 11 07 09 AM
App set and no User identifier set:
Screen Shot 2020-12-10 at 11 05 46 AM
No App or User set:
Screen Shot 2020-12-11 at 2 37 44 PM

### Issue:

- [OKTA-341975](https://oktainc.atlassian.net/browse/OKTA-341975)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -814,7 +814,9 @@ oie.select.authenticators.verify.title = Verify it\'s you with an authenticator
 oie.select.authenticators.verify.subtitle = Select from the following options
 
 ## Success Redirect
-oie.success.redirect = You are being redirected
+oie.success.text.signingIn = Signing in
+oie.success.text.signingIn.with.appName = Signing in to {0}
+oie.success.text.signingIn.with.appName.and.identifier = Signing in to {0} as {1}
 
 ## Footer
 oie.go.back = Go back
--- a/src/v2/view-builder/views/SuccessView.js
+++ b/src/v2/view-builder/views/SuccessView.js
@@ -1,11 +1,32 @@
+import { _, loc } from 'okta';
+
 import BaseView from '../internals/BaseView';
 import BaseForm from '../internals/BaseForm';
 import Util from '../../../util/Util';
-import { loc } from 'okta';
 
 const Body = BaseForm.extend({
   title () {
-    return  loc('oie.success.redirect', 'login');
+    let titleString = loc('oie.success.text.signingIn', 'login');
+    // For more info on the API response available in appState, see IdxResponseBuilder.java
+    const app = this.options.appState.get('app');
+    const user = this.options.appState.get('user');
+
+    if (_.isEmpty(app)) {
+      return titleString;
+    }
+
+    const {label: appInstanceName, name: appDisplayName} = app;
+    const {identifier: userEmail} = user;
+
+    const appName = appInstanceName ? appInstanceName : appDisplayName;
+
+    if (appName && userEmail) {
+      titleString = loc('oie.success.text.signingIn.with.appName.and.identifier', 'login', [appName, userEmail]);
+    } else if (appName) {
+      titleString = loc('oie.success.text.signingIn.with.appName', 'login', [appName]);
+    }
+
+    return titleString;
   },
   noButtonBar: true,
   initialize () {
--- a/test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js
+++ b/test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js
@@ -1,8 +1,16 @@
-import { RequestMock, Selector, ClientFunction } from 'testcafe';
+import { RequestMock, RequestLogger, Selector, ClientFunction } from 'testcafe';
 import IdentityPageObject from '../framework/page-objects/IdentityPageObject';
 import identifyWithIdpsIdentify from '../../../playground/mocks/data/idp/idx/identify-with-third-party-idps';
 import identifyWithIdpsNoIdentify from '../../../playground/mocks/data/idp/idx/identify-with-only-third-party-idps';
 import identifyOnlyOneIdp from '../../../playground/mocks/data/idp/idx/identify-with-only-one-third-party-idp';
+import identifyOnlyOneIdpAppUser from '../../../playground/mocks/data/idp/idx/identify-with-only-one-third-party-idp-app-user';
+
+const logger = RequestLogger(/introspect/,
+  {
+    logRequestBody: true,
+    stringifyRequestBody: true,
+  }
+);
 
 const mockWithIdentify = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/introspect')
@@ -16,11 +24,14 @@ const mockOnlyOneIdp = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/introspect')
   .respond(identifyOnlyOneIdp)
   .onRequestTo('http://localhost:3000/sso/idps/facebook-idp-id-123?stateToken=inRUXNhsc6Evt7GAb8DPAA')
-  .respond(async (req, res) => {
-    // Delay response for 2 seconds so that we can verify the `success redirect` screen
-    await new Promise((r) => setTimeout(r, 2000));
-    res.setBody('<html><h1>A external IdP login page for testcafe testing</h1></html>');
-  });
+  .respond('<html><h1>A external IdP login page for testcafe testing</h1></html>');
+
+const mockOnlyOneIdpAppUser = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(identifyOnlyOneIdpAppUser)
+  .onRequestTo('http://localhost:3000/sso/idps/facebook-idp-id-123?stateToken=inRUXNhsc6Evt7GAb8DPAA')
+  .respond('<html><h1>A external IdP login page for testcafe testing</h1></html>');
+
 
 fixture('Identify + IDPs');
 
@@ -87,13 +98,8 @@ test.requestHooks(mockWithoutIdentify)('should only render idp buttons with iden
   await t.expect(await identityPage.signoutLinkExists()).notOk();
 });
 
-test.requestHooks(mockOnlyOneIdp)('should auto redirect to 3rd party IdP login page', async t => {
+test.requestHooks(logger, mockOnlyOneIdp)('should auto redirect to 3rd party IdP login page with basic Signing in message', async t => {
   await setup(t);
-
-  // FIXME: assert landing on success redirect page
-  // import BasePageObject from '../framework/page-objects/BasePageObject';
-  // const successRedirectPage = new BasePageObject(t);
-  // await t.expect(successRedirectPage.getFormTitle()).eql('You are being redirected');
 
   const { log } = await t.getBrowserConsoleMessages();
   await t.expect(log.length).eql(3);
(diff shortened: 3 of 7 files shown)


sherizada-okta:
This needs to be tested against rain, once I figure that out...

pradeepdewda-okta:
@vincentvoong-okta need some input from UX.. Showing the email id like this (see screenshot) would be okay ? I see in SIW we use (***) to hide full email id.

playground/mocks/data/idp/idx/success.json
```diff
@@ -1,32 +1,43 @@
 {
-    "stateHandle": "022FJBvS6bHS7OjQSOhtfghhJvO2a2TELaMZ94s69K",
+    "stateHandle": "02EQcms7FOAByCjwRqpUCKfzxNJj2wJ_nh3x7zheyD",
```
pradeepdewda-okta, line 2: Instead of updating this view.. just create a new one for your usecase. for eg `success-with-email-identifier`

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -817,7 +817,8 @@ oie.select.authenticators.verify.title = Verify it\'s you with an authenticator
 oie.select.authenticators.verify.subtitle = Select from the following options
 
 ## Success Redirect
-oie.success.redirect = You are being redirected
+oie.success.text.signingIn = Signing in to {0} as {1}
```
pradeepdewda-okta, line 820: change to `oie.success.text.signingInWithIdentifier`

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -817,7 +817,8 @@ oie.select.authenticators.verify.title = Verify it\'s you with an authenticator
...
+oie.success.text.signingInWithoutIdentifier = Signing in to {0}
```
pradeepdewda-okta, line 821: change to `oie.success.text.signingIn`

src/v2/view-builder/views/SuccessView.js
```diff
@@ -5,7 +5,16 @@ import { loc } from 'okta';
 
 const Body = BaseForm.extend({
   title () {
-    return  loc('oie.success.redirect', 'login');
+    // For more info on the API response available in appState, see IdxResponse:
+    // https://github.com/okta/okta-core/blob/master/server/api/src/main/java/com/okta/api/mediation/
```
pradeepdewda-okta, line 9: we can just keep the file name `IdxResponseBuilder.java` instead of whole link..

src/v2/view-builder/views/SuccessView.js
```diff
@@ -5,7 +5,16 @@ import { loc } from 'okta';
...
+    // services/auth/IdxResponseBuilder.java
+    const {label: appInstanceName, name: appDisplayName} = this.options.appState.get('app');
+    const {identifier: userEmail} = this.options.appState.get('user');
+
+    const appName = appInstanceName ? appInstanceName : appDisplayName;
```
pradeepdewda-okta, line 14: I think we should always show appDisplayName only.. appInstanceName can be shown as variable (google_1, salesforce1 et)

src/v2/view-builder/views/SuccessView.js
```diff
@@ -5,7 +5,16 @@ import { loc } from 'okta';
 
 const Body = BaseForm.extend({
   title () {
-    return  loc('oie.success.redirect', 'login');
+    // For more info on the API response available in appState, see IdxResponse:
```
pradeepdewda-okta, line 8: Add a testcafe test for this change.

src/v2/view-builder/views/SuccessView.js
sherizada-okta: Product designer on the ticket said to favor `appInstanceName` over `appDisplayName`: https://oktainc.atlassian.net/browse/OKTA-341975?focusedCommentId=1656267
On dev when signing into the IDX org, `appDisplayName` is saasure and `appInstanceName` is Okta Administration. 
`appInstanceName` is favored because it is admin configurable, do you know how I can configure it to ensure this is the right value?

btw I got that label is the instanceName and name is displayName from here: https://github.com/okta/okta-core/blob/master/server/api/src/main/java/com/okta/api/mediation/services/auth/IdxResponseBuilder.java#L1501-L1502

pradeepdewda-okta: If `appInstanceName` is `label`.. then i think this is fine.

You can just go to `Applications` and add different apps with different names. The name you add would appear as  `label`

pradeepdewda-okta:
> @vincentvoong-okta need some input from UX.. Showing the email id like this (see screenshot) would be okay ? I see in SIW we use (***) to hide full email id.

@taylorlaubach-okta +1

pradeepdewda-okta:
> > @vincentvoong-okta need some input from UX.. Showing the email id like this (see screenshot) would be okay ? I see in SIW we use (***) to hide full email id.
> 
> @taylorlaubach-okta +1

@stevennguyen-okta : cc

taylorlaubach-okta:
@pradeepdewda-okta Is the user already authenticated?

I believe we generally hide the email address because we don't want to expose information about the account to anyone who makes a request. There are two instances where it shouldn't be a problem to show the username: 1. We've already verified the identity; 2. The value was entered by the requestor, and the SIW is repeating that same value (valid or invalid) rather than producing a valid value from the backend.

One other consideration here: I believe @vincentvoong-okta and @thanh-hanguyen-okta have a change in the works that will show the username at the top of the SIW. So this inline message with the username may be redundant once that gets implemented.

stevennguyen-okta:
> @pradeepdewda-okta Is the user already authenticated?
> 
> I believe we generally hide the email address because we don't want to expose information about the account to anyone who makes a request. There are two instances where it shouldn't be a problem to show the username: 1. We've already verified the identity; 2. The value was entered by the requestor, and the SIW is repeating that same value (valid or invalid) rather than producing a valid value from the backend.
> 
> One other consideration here: I believe @vincentvoong-okta and @thanh-hanguyen-okta have a change in the works that will show the username at the top of the SIW. So this inline message with the username may be redundant once that gets implemented.

Yea they should already have authenticated at this point. For some more context, this is to replace the "You are being redirected" text on the very last SIW step before we pass the user onto the app. While the devices team was working on Fastpass, we looked at the user experience across the SIW and made this suggestion to the Redirect screen to try to provide more context than "you are being redirected". This is particularly useful for the completely silent sign ins with FastPass. This was reviewed and approved by various groups like Wils, devices, and Alex Dahl. I think mfa team was in the review.

taylorlaubach-okta (approved):
Awesome! @stevennguyen-okta thanks for the context!

Yeah LGTM

playground/mocks/data/idp/idx/success.json
sherizada-okta: What about responseConfig.js? Now SuccessView won't work with the original version of success.json because the API response in there is incomplete.

pradeepdewda-okta: ? slack.

test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js
```diff
@@ -105,7 +104,7 @@ test.requestHooks(mockOnlyOneIdp)('should auto redirect to 3rd party IdP login p
   });
 
   // assert redirect to IdP login page eventually
-  await t.expect(await Selector('h1').innerText).eql('A external IdP login page for testcafe testing');
```
sherizada-okta, line 108: There was a warning about this await being incorrect

test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js
```diff
@@ -90,10 +91,8 @@ test.requestHooks(mockWithoutIdentify)('should only render idp buttons with iden
 test.requestHooks(mockOnlyOneIdp)('should auto redirect to 3rd party IdP login page', async t => {
   await setup(t);
 
-  // FIXME: assert landing on success redirect page
-  // import BasePageObject from '../framework/page-objects/BasePageObject';
-  // const successRedirectPage = new BasePageObject(t);
-  // await t.expect(successRedirectPage.getFormTitle()).eql('You are being redirected');
+  const successRedirectPage = new BasePageObject(t);
+  await t.expect(successRedirectPage.getFormTitle()).eql('Signing in to Okta Administration as admin@idx.com');
```
sherizada-okta, line 95: @pradeepdewda-okta here's the e2e test. All the work was already there.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -817,7 +817,9 @@ oie.select.authenticators.verify.title = Verify it\'s you with an authenticator
 oie.select.authenticators.verify.subtitle = Select from the following options
 
 ## Success Redirect
-oie.success.redirect = You are being redirected
+oie.success.text.signingIn = Signing in
+oie.success.text.signingInWithAppName = Signing in to {0}
```
pradeepdewda-okta, line 821: we generally don't use camel case.. instead `oie.success.text.signingIn.with.appName`.. same for the other one.

test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js
```diff
@@ -22,6 +24,17 @@ const mockOnlyOneIdp = RequestMock()
     res.setBody('<html><h1>A external IdP login page for testcafe testing</h1></html>');
   });
 
+const mockOnlyOneIdpAppUser = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(identifyOnlyOneIdpAppUser)
+  .onRequestTo('http://localhost:3000/sso/idps/facebook-idp-id-123?stateToken=inRUXNhsc6Evt7GAb8DPAA')
+  .respond(async (req, res) => {
+    // Delay response for 2 seconds so that we can verify the `success redirect` screen
+    await new Promise((r) => setTimeout(r, 2000));
```
pradeepdewda-okta, line 33: Avoid adding `delays` in our tests. Instead add a RequestLogger and check for network request was made or not..

test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js
```diff
@@ -105,7 +138,7 @@ test.requestHooks(mockOnlyOneIdp)('should auto redirect to 3rd party IdP login p
   });
 
   // assert redirect to IdP login page eventually
-  await t.expect(await Selector('h1').innerText).eql('A external IdP login page for testcafe testing');
+  await t.expect(Selector('h1').innerText).eql('A external IdP login page for testcafe testing');
```
pradeepdewda-okta, line 141: why we needed to remove `await` from here ?

sherizada-okta: When I ran the test locally I got this warning
```Warnings (1):
 --
  You passed a DOM snapshot property to the assertion's 't.expect()' method. The
  property value is assigned when the snapshot is resolved and this value is no
  longer updated. To ensure that the assertion verifies an up-to-date value, pass
  the selector property without 'await'.

     136 |    controller: null,
     137 |    formName: 'success-redirect',
     138 |  });
     139 |
     140 |  // assert redirect to IdP login page eventually
   > 141 |  await t.expect(await Selector('h1').innerText).eql('A external IdP
  login page for testcafe testing');
     142 |  const pageUrl = await ClientFunction(() => window.location.href)();
     143 |  await
  t.expect(pageUrl).eql('http://localhost:3000/sso/idps/facebook-idp-id-123?stateToken=inRUXNhsc6Evt7GAb8DPAA');
     144 |
     145 |});
     146 |`

Comment (pradeepdewda-okta) [reply test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js:141]: was it working earlier ?

Comment (sherizada-okta) [reply test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js:141]: It's a warning so it would work/pass anyway. Warning went away once I removed the await.

**Review by pradeepdewda-okta — APPROVED** [review]
Do not squash the last commit, otherwise it's hard to keep track what changed between the reviews.

File: src/v2/view-builder/views/SuccessView.js [hunk 12]
```diff
@@ -1,11 +1,32 @@
...
+      return titleString;
+    }
+
+    const {label: appInstanceName, name: appDisplayName} = app;
+    const {identifier: userEmail} = user;
+
+    const appName = appInstanceName ? appInstanceName : appDisplayName;
+
+    if (appName && userEmail) {
+      titleString = loc('oie.success.text.signingIn.with.appName.and.identifier', 'login', [appName, userEmail]);
```
haishengwu-okta, line 24: - ` return loc(..)` 
- change `titleString` to be `const`

sherizada-okta: I think separating all the title variants from a single variable is harder to read in the end. This function is currently just returning titleString, but if I return the loc functions without naming them and the first titleString, there'll be 3 return statements that look unrelated. But they're not unrelated, they're all just a titleString.

PR #1556, opened 2020-12-08
Hotfix 5.1 fix main container width in IE

sherizada-okta:
## Description:
Required for IE11 to apply the main container's width. 

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-351767](https://oktainc.atlassian.net/browse/OKTA-351767)

### DIFF
--- a/assets/sass/okta-theme.scss
+++ b/assets/sass/okta-theme.scss
@@ -81,6 +81,8 @@
 }
 
 #okta-sign-in.auth-container {
+  // for IE to support main tag's width
+  display: block;
 
   &.main-container {
     /* -- Fonts and Text Colors -- */


assets/sass/okta-theme.scss
```diff
@@ -81,6 +81,8 @@
 }
 
 #okta-sign-in.auth-container {
+  // for IE to support main tag's width
+  display: block;
```
pradeepdewda-okta, line 85: are we removing `main` ?

sherizada-okta: Adding display: block so IE recognize's main tag's width.

pradeepdewda-okta: should we also add that this change is only for IE in CSS ? like make it a browser specific change instead of a general css change.

sherizada-okta: I just saw this comment

sherizada-okta: `display` does not have browser vendor specific prefixes, so how would one do that in this codebase?

pradeepdewda-okta: there is way to do it, sending you SO link. Now this is closed, we can do it later (probably not a priority)

PR #1560, opened 2020-12-09
fix: hide error before switching to the next UI

jhe-okta:
## Description:

delay switching to the next UI for universal link approach in case the Okta Verify is not installed

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Reviewers:
@sshen-okta @pradeepdewda-okta 

### Issue:

- [OKTA-350084](https://oktainc.atlassian.net/browse/OKTA-350084)

### DIFF
--- a/src/v2/view-builder/views/signin/SignInWithDeviceOption.js
+++ b/src/v2/view-builder/views/signin/SignInWithDeviceOption.js
@@ -2,6 +2,7 @@ import { View, createButton, loc } from 'okta';
 import hbs from 'handlebars-inline-precompile';
 import Util from '../../../../util/Util';
 import Enums from '../../../../util/Enums';
+import { UNIVERSAL_LINK_POST_DELAY } from '../../utils/Constants';
 
 export default View.extend({
   className: 'sign-in-with-device-option',
@@ -30,13 +31,26 @@ export default View.extend({
       icon: 'okta-verify-authenticator',
       title: loc('oktaVerify.button', 'login'),
       click () {
-        if (deviceChallenge.challengeMethod && deviceChallenge.challengeMethod === Enums.UNIVERSAL_LINK_CHALLENGE) {
+        const isUVapproach = deviceChallenge.challengeMethod &&
+          deviceChallenge.challengeMethod === Enums.UNIVERSAL_LINK_CHALLENGE;
+        if (isUVapproach) {
+          // launch the Okta Verify app
           Util.redirect(deviceChallenge.href);
         }
         if (this.options.isRequired) {
           appState.trigger('saveForm', this.model);
         } else {
-          appState.trigger('invokeAction', Enums.LAUNCH_AUTHENTICATOR_REMEDIATION_NAME);
+          // OKTA-350084
+          // For the universal link (iOS) approach,
+          // we need to 1. launch the Okta Verify app
+          // and 2. take the enduser to the next step right away
+          // In Safari, when Okta Verify app is not installed, step 1 responds with error immediately,
+          // then step 2 will respond with error.
+          // To avoid showing the error right before switching to the next UI,
+          // wait for 500 milliseconds before invoking step 2
+          Util.callAfterTimeout(() => {
+            appState.trigger('invokeAction', Enums.LAUNCH_AUTHENTICATOR_REMEDIATION_NAME);
+          }, isUVapproach ? UNIVERSAL_LINK_POST_DELAY : 0);
         }
       }
     }), '.okta-verify-container');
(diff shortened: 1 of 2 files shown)


sshen-okta (approved):

src/v2/view-builder/views/signin/SignInWithDeviceOption.js
```diff
@@ -30,13 +30,26 @@ export default View.extend({
...
-          appState.trigger('invokeAction', Enums.LAUNCH_AUTHENTICATOR_REMEDIATION_NAME);
+          // OKTA-350084
+          // For the universal link (iOS) approach,
+          // we need to 1. launch the Okta Verify app
+          // and 2. take the enduser to the next step right away
+          // In Safari, when Okta Verify app is not installed, step 1 responds with error immediately,
+          // then step 2 will respond with error.
+          // To avoid showing the error right before switching to the next UI,
+          // wait for 500 milliseconds before invoking step 2
+          setTimeout(() => {
```
pradeepdewda-okta, line 50: I personally don't like introducing setTimeout that blocks the IO.. Can you give me more details about this ?

jhe-okta: @pradeepdewda-okta I left the reason in the comment. Basically the problem is in step 2 where we switch the user to the next SIW screen. If the step 1 has problem, the step 2 will return error and show on the UI. Basically we want to give it time before we know if the step 1 is okay. If it is okay (the Okta Verify app is installed), the enduser will be temporarily redirected to login.okta.com. So that the step 2 will not even be invoked. Ideally we won't want to use time out, but since we can't get the status of the step 1, this is what we think the best workaround is. Let me know if you have better ideas.

pradeepdewda-okta: What if 500ms is not enough for slower internet connections ? Do you explore polling instead ?

jhe-okta: That's possible too. But base on the testing we did, 500 is ideal that it does not affect user experience. Also, worst case -- if the enduser gets to the step 2, there's cancel link they can click on to get out of it. This case is very unlikely to happen, we set up a very unrealistic sign on rule just to reproduce the issue.

pradeepdewda-okta: To meet your deadline, i think it would be fine. But i would recommend to circle back to this issue and open a ticket to explore Polling and reimplement it.

I see this view is used at two places. You tested both right ?

jhe-okta: To be clear, the error is not from the polling. It is from the step 2 `/idp/idx/authenticators/okta-verify/launch` where switching to the next screen of polling happens. Maybe we can look into if there's improvement we can do in the launch api in order for better UX. cc @sshen-okta

pradeepdewda-okta: @jhe-okta @sshen-okta Yes since we have polling framework, we can leverage that. If there is any need to make certain API tweaks to support polling for this use case, we should do that.

jhe-okta: This is NOT about the polling API.

pradeepdewda-okta: @jhe-okta May be i miss understood. You says we go to step 2 when we can confirm the step 1 was okay.. So to confirm step 1 was okay or not we are waiting for sometime because server takes time to return the value ? Is that right ?

jhe-okta: If we proceed to step 2 right after step 1 without delay, when step 1 fails, step 2 will respond with error which we show in the UI and it is bad UX. When step 1 succeeds, the enduser will be taken out of the SIW. The time out is to make sure that the redirect happens first before the step 2, so that the UI does not display error. Again, this case is not pratical, it depends on an unrealistic sign on rule to be set up first.

pradeepdewda-okta: Okay. One more thing i want to clear. If Step 1 fails then expectation is still to show Step 2 screen? But introducing delay would not cause error to show up ?

Also can you add this `500` constant to `../../utils/Constants`

pradeepdewda-okta: @jhe-okta can you explore if we can use some kind of promise to resolve first before executing step 2 ?

jhe-okta: @pradeepdewda-okta updated to use the `Util.callAfterTimeout`.

pradeepdewda-okta (approved):
@jhe-okta @jhe-okta Jintai would create a jira to further investigate this approach to avoid using setTimeOut. Although not an immediate requirement.

jhe-okta:
> @jhe-okta @jhe-okta Jintai would create a jira to further investigate this approach to avoid using setTimeOut. Although not an immediate requirement.

https://oktainc.atlassian.net/browse/OKTA-352373

PR #1563, opened 2020-12-09
feat: poll requests when server is in safe or readonly mode

pradeepdewda-okta:
## Description:
1. When server is not in active mode, then requests that needs database write access can error out.
2. To avoid such errors introduce polling.
3. use cases are server is in safe-mode and rate limiting.

https://oktawiki.atlassian.net/wiki/spaces/eng/pages/1411254114/IdX+Polling+Transaction+Framework+1-Page

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
**update**: 
https://okta.box.com/s/9sazdq0sb6ffgnvkyswx0tsp9948gg97

**update:** Messages changed to:
Screen Shot 2020-12-14 at 12 03 46 PM

on poll success
https://okta.box.com/s/f58w8wp5af891g235uebsyx524vbis1x

on poll expires
https://okta.box.com/s/acf08b6cpgb676s5v8a6vnq5n27bacnh

Poll usecase in IDX Org:
https://okta.box.com/s/31t9fu102g69u5wlosputsl15egnp495

### Issue:

- [OKTA-349707](https://oktainc.atlassian.net/browse/OKTA-349707)

### DIFF
--- a/src/v2/view-builder/views/PollView.js
+++ b/src/v2/view-builder/views/PollView.js
@@ -1,0 +1,55 @@
+import { loc, View } from 'okta';
+import hbs from 'handlebars-inline-precompile';
+import BaseView from '../internals/BaseView';
+import BaseForm from '../internals/BaseForm';
+import polling from './shared/polling';
+import { MS_PER_SEC } from '../utils/Constants';
+
+const PollMessageView = View.extend({
+  template: hbs`
+    <div class="ion-messages-container">
+      {{i18n code="poll.form.message" 
+        bundle="login" arguments="countDownCounterValue" $1="<span class='strong'>$1</span>"}}
+    </div>
+    <div class="okta-waiting-spinner"></div>`
+  ,
+  getTemplateData () {
+    const countDownCounterValue = this.options;
+    return {
+      countDownCounterValue,
+    };
+  },
+});
+
+const Body = BaseForm.extend(Object.assign(
+  {
+    title () {
+      return  loc('poll.form.title', 'login');
+    },
+
+    noButtonBar: true,
+    
+    initialize () {
+      BaseForm.prototype.initialize.apply(this, arguments);
+      this.startPolling();
+    },
+
+    render () {
+      BaseForm.prototype.render.apply(this, arguments);
+      this.countDownCounterValue = Math.ceil(this.options.appState.getCurrentViewState().refresh / MS_PER_SEC);
+      this.add(new PollMessageView(this.countDownCounterValue));
+      this.startCountDown('.ion-messages-container span', MS_PER_SEC);
+    },
+
+    remove () {
+      BaseForm.prototype.remove.apply(this, arguments);
+      this.stopPolling();
+    }
+  },
+
+  polling,
+));
+
+export default BaseView.extend({
+  Body
+});
--- a/src/v2/view-builder/views/shared/polling.js
+++ b/src/v2/view-builder/views/shared/polling.js
@@ -1,9 +1,12 @@
 import { _ } from 'okta';
+import { MS_PER_SEC } from '../../utils/Constants';
 
 export default {
   startPolling () {
+    this.pollingInterval = this.options.currentViewState.refresh;
+    this.countDownCounterValue = Math.ceil(this.pollingInterval / MS_PER_SEC);
     // Poll is present in remediation form
-    if (this.options.currentViewState.refresh) {
+    if (this.pollingInterval) {
       this._startRemediationPolling();
     } else {
       // Poll is present in authenticator/ authenticator Enrollment obj.
@@ -35,17 +38,39 @@ export default {
     });
   },
 
+  startCountDown (selector , interval) {
+    if(this.countDown) {
+      clearInterval(this.countDown);
+    }
+    this.counterEl = this.$el.find(selector);
+    this.countDown = setInterval(() => {
+      if(this.counterEl.text() !== '1') {
+        this.counterEl.text(this.counterEl.text() - 1);
+      } else {
+        // reset the countdown counter visible to enduser, if still polling
+        this.counterEl.text(this.countDownCounterValue);
+        this.startCountDown(selector, interval);
+      }
+    }, interval, this);
+  },
+
+  _stopCountDown () {
+    if(this.countDown) {
+      clearInterval(this.countDown);
+    }
+  },
+
   _startRemediationPolling () {
-    const pollingInterval = this.options.currentViewState.refresh;
-    if (_.isNumber(pollingInterval)) {
+    if (_.isNumber(this.pollingInterval)) {
       this.polling = setInterval(() => {
         this.options.appState.trigger('saveForm', this.model);
-      }, pollingInterval);
+      }, this.pollingInterval);
     }
   },
 
   stopPolling () {
     if (this.polling) {
+      this._stopCountDown();
       clearInterval(this.polling);
     }
   }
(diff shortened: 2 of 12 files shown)


iancorcoran-okta (approved):

taylorlaubach-okta:
This seems pretty different from the other polling design that we discussed. A couple of questions:
1. Where did the "Waiting for server to respond" text come from?
2. How did we decide to remove the count down for retrying?

pradeepdewda-okta:
> This seems pretty different from the other polling design that we discussed. A couple of questions:
> 
> 1. Where did the "Waiting for server to respond" text come from?
> 2. How did we decide to remove the count down for retrying?

1. This is the Poll Form fixed title that i added. I do welcome any suggestions here.
2. We can add the count, but is that relevant ?

pradeepdewda-okta:
> > This seems pretty different from the other polling design that we discussed. A couple of questions:
> > 
> > 1. Where did the "Waiting for server to respond" text come from?
> > 2. How did we decide to remove the count down for retrying?
> 
> 1. This is the Poll Form fixed title that i added. I do welcome any suggestions here.
> 2. We can add the count, but is that relevant ?

Also We need to put a generalized message for this one, because SIW don't actual know In what context it's doing this immediate polling. The context could be the server is in readonly mode or safemode or the endpoint has hit the rate limiting constraints

taylorlaubach-okta:
I see, thanks for the context!

The count is helpful only because it makes what is happening a little more transparent. When we designed the other polling UI, we were concerned that a constant spinner would give the impression that nothing was happening, and that the user would be more likely to give up or refresh the page. So we added the count, and intentionally slowed it down enough to make sure it was clear what was happening. That way the user sees that things are happening: we tried, we failed, we are waiting, now we are trying again.

For the generic text, what you've got is super clear to me. That said, I'm concerned that for a non-technical end-user it might be confusing or intimidating. I'd prefer: "The system was unable to complete your request. We will automatically retry in {countdown} seconds"

src/v2/view-builder/views/PollView.js
```diff
@@ -0,0 +1,35 @@
...
+  {
+    title () {
+      return  loc('safemode.poll.title', 'login');
+    },
+
+    noButtonBar: true,
+    
+    initialize () {
+      BaseForm.prototype.initialize.apply(this, arguments);
+      this.startPolling();
```
sherizada-okta, line 16: I looked at `startPolling`, it doesn't call the callback immediately but waits for {refresh} ms to do the first call. Kind of a waste, unless it's unlikely the first call will ever be successful?

pradeepdewda-okta: That's intentional.. If any remediation resolves to `poll` means server asks sign-widget to wait for X seconds before trying out the same request again.

src/v2/view-builder/views/PollView.js
```diff
@@ -0,0 +1,54 @@
...
+    },
+
+    render () {
+      BaseForm.prototype.render.apply(this, arguments);
+      this.pollingInterval = Math.ceil(this.options.appState.getCurrentViewState().refresh / 1000);
```
sherizada-okta, line 38: What are the units for `refresh` and `pollingInterval`?  Ultimately this is for `setInterval` so `pollingInterval` should be milliseconds? But `refresh` from the mock looks like it's already in ms? 
It would be easier to reason what the calculation is doing if the constant 1000 was a variable like `MS_PER_SEC` and `pollingInterval` included its units, `pollingIntervalMS`.

src/v2/view-builder/views/shared/polling.js
```diff
@@ -2,8 +2,9 @@ import { _ } from 'okta';
 
 export default {
   startPolling () {
+    this.pollingInterval = this.options.currentViewState.refresh;
```
sherizada-okta, line 5: Why doesn't this have the calculation like
`this.pollingInterval = Math.ceil(this.options.appState.getCurrentViewState().refresh / 1000);`
from
 https://github.com/okta/okta-signin-widget/pull/1563/files#diff-09e81d05f4347e1aeffa0830d7f8f711b4496f57f24cc51ed7331b0932aef815R38

test/testcafe/spec/PollView_spec.js
```diff
@@ -0,0 +1,50 @@
...
+  const identityPage = new IdentityPageObject(t);
+  const selectFactorPageObject = new SelectFactorPageObject(t);
+
+  await identityPage.navigateToPage();  
+  return [identityPage , selectFactorPageObject];
+}  
+
+test.requestHooks(requestLogger, identifyMock)('should poll', async t => {
+  let identityPage, selectFactorPage;
+  [identityPage, selectFactorPage]  = await setup(t);
```
sherizada-okta, line 38: Why is the `let` declaration on a separate line?

src/v2/view-builder/views/shared/polling.js
pradeepdewda-okta: Good question.. because the one in PollingView shows up as seconds to enduser.. We don't want to show that in `ms`

test/testcafe/spec/PollView_spec.js
pradeepdewda-okta: fixed it.

src/v2/view-builder/views/PollView.js
pradeepdewda-okta: I think i should change the variable name.. it's not for setInterval, it's a countDown counter.. for if `refresh:5000` then enduser would see `5 seconds` `4seconds`.... and so on. See the screenshot.

I will change the variable name.

sherizada-okta: To clear up possible future confusion, there should be an explanation of what's going on here. I think this would do it?
```
// Display countdown to next poll in seconds, convert refresh (ms) to seconds
const pollingIntervalSeconds = Math.ceil(this.options.appState.getCurrentViewState().refresh / MS_PER_S);
this.pollingInterval = pollingIntervalSeconds;
```
and declare the constant outside of this method so it can be re-used each time render is called.

src/v2/view-builder/views/shared/polling.js
sherizada-okta: Ah I get it.

src/v2/view-builder/views/PollView.js
pradeepdewda-okta: I changed the variable name to make it more clear what it does..

sherizada-okta (approved):

sherizada-okta:
The demo video should be updated for the countdown and include the poll network requests to show they match up.

src/v2/view-builder/views/PollView.js
```diff
@@ -0,0 +1,55 @@
...
+    noButtonBar: true,
+    
+    initialize () {
+      BaseForm.prototype.initialize.apply(this, arguments);
+      this.startPolling();
+    },
+
+    render () {
+      BaseForm.prototype.render.apply(this, arguments);
+      this.countDownCounterValue = Math.ceil(this.options.appState.getCurrentViewState().refresh / MS_PER_SEC);
```
haishengwu-okta, line 39: looks like `this.countDownCounterValue` has been set at `polling` prototype object.

src/v2/view-builder/views/PollView.js
```diff
@@ -0,0 +1,55 @@
...
+      this.add(new PollMessageView(this.countDownCounterValue));
```
haishengwu-okta, line 40:
```
{ countDownCounterValue: this.countDownCounterValue }
```

src/v2/view-builder/views/PollView.js
pradeepdewda-okta: this one initializes `PollMessageView` value right before render completes.

haishengwu-okta: I meant `this.startPolling();` has create `this.countDownCounterValue` already.

PR #1575, opened 2020-12-16
fix: cancel loopback probing when polling is cancelled

jhe-okta:
## Description:
In loopback server, probing (check port and challenge) still happens even after the cancel link (cancel polling) from the footer is clicked. The change is to abort the probing once polling is cancelled.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Reviewers:
@pradeepdewda-okta 

### Issue:

- [OKTA-353152](https://oktainc.atlassian.net/browse/OKTA-353152)

### DIFF
--- a/src/v2/view-builder/views/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/DeviceChallengePollView.js
@@ -11,6 +11,7 @@ import DeviceFingerprint from '../../../util/DeviceFingerprint';
 import polling from './shared/polling';
 import Util from '../../../util/Util';
 import Enums from '../../../util/Enums';
+import { CANCEL_POLLING_ACTION } from '../utils/Constants';
 import Link from '../components/Link';
 import { getIconClassNameForBeacon } from '../utils/AuthenticatorUtil';
 
@@ -50,6 +51,7 @@ const Body = BaseForm.extend(Object.assign(
 
     remove () {
       BaseForm.prototype.remove.apply(this, arguments);
+      this.stopProbing();
       this.stopPolling();
     },
 
@@ -140,9 +142,13 @@ const Body = BaseForm.extend(Object.assign(
       };
 
       const doProbing = () => {
-        return checkPort()
+        this.checkPortXhr = checkPort();
+        return this.checkPortXhr
           // TODO: can we use standard ES6 promise methods, then/catch?
-          .done(onPortFound)
+          .done(() => {
+            this.probingXhr = onPortFound();
+            return this.probingXhr;
+          })
           .fail(onFailure);
       };
 
@@ -160,7 +166,7 @@ const Body = BaseForm.extend(Object.assign(
             Logger.error(`Authenticator is not listening on port ${currentPort}.`);
             if (countFailedPorts === ports.length) {
               Logger.error('No available ports. Loopback server failed and polling is cancelled.');
-              this.options.appState.trigger('invokeAction', 'authenticatorChallenge-cancel');
+              this.options.appState.trigger('invokeAction', CANCEL_POLLING_ACTION);
             }
           });
       });
@@ -171,6 +177,11 @@ const Body = BaseForm.extend(Object.assign(
       this.ulDom = this.add(`
         <iframe src="${this.customURI}" id="custom-uri-container" style="display:none;"></iframe>
       `).last();
+    },
+
+    stopProbing () {
+      this.checkPortXhr && this.checkPortXhr.abort();
+      this.probingXhr && this.probingXhr.abort();
     },
   },
 
@@ -197,7 +208,7 @@ const Footer = BaseFooter.extend({
       this.add(Link, { options: {
         name: 'cancel-authenticator-challenge',
         label: loc('loopback.polling.cancel.link', 'login'),
-        actionPath: 'authenticatorChallenge-cancel',
+        actionPath: CANCEL_POLLING_ACTION,
       }});
     }
   }
--- a/src/v2/view-builder/utils/Constants.js
+++ b/src/v2/view-builder/utils/Constants.js
@@ -1,3 +1,4 @@
 export const SHOW_RESEND_TIMEOUT = 30000;
 export const WARNING_TIMEOUT = 30000;
 export const UNIVERSAL_LINK_POST_DELAY = 500;
+export const CANCEL_POLLING_ACTION = 'authenticatorChallenge-cancel';


pradeepdewda-okta:
@jhe-okta please add a video.

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -14,6 +14,7 @@ import Enums from '../../../util/Enums';
 import Link from '../components/Link';
 import { getIconClassNameForBeacon } from '../utils/AuthenticatorUtil';
 
+const CANCEL_POLLING_ACTION = 'authenticatorChallenge-cancel';
```
pradeepdewda-okta, line 17: lets move this to `Constants.js`

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -140,12 +141,23 @@ const Body = BaseForm.extend(Object.assign(
       };
 
       const doProbing = () => {
-        return checkPort()
+        this.checkPortXhr = checkPort();
+        return this.checkPortXhr
           // TODO: can we use standard ES6 promise methods, then/catch?
-          .done(onPortFound)
+          .done(() => {
```
pradeepdewda-okta, line 147: I think you can improve this using `ES6 promise methods` now ?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -140,12 +141,23 @@ const Body = BaseForm.extend(Object.assign(
...
+            this.probingXhr = onPortFound();
+            return this.probingXhr;
+          })
           .fail(onFailure);
       };
 
+      this.listenTo(this.options.appState, 'invokeAction', (action) => {
```
pradeepdewda-okta, line 154: Use `handleInvokeAction` method of `FormController` instead of intercepting it here.

pradeepdewda-okta (commented):
Add a testcafe test.

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: I think we should do it later. It requires some more investigation and testing on how to convert the jquery ajax with timeout option to use promise so we can use then/catch.

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: `handleInvokeAction` is in the controller, not in the view. This change is only needed in this view. How do you suggest we handle it?

src/v2/view-builder/views/DeviceChallengePollView.js
pradeepdewda-okta: Sorry, I don't get it. What's so different about this ?

src/v2/view-builder/views/DeviceChallengePollView.js
pradeepdewda-okta: All the form actions, We try to handle in FormController irrespective of which view is this.. I also see that `this.listenTo(this.options.appState, 'invokeAction')` invokeAction is not listened by any other view or used anywhere else but FormController..

In my opinion actions should be handled by FormController only.. Try to debug `handleInvokeAction` method and see if you can generalize `cancel` polling action.. You can pass the context of your view to handler to get the methods that you want to abort..

pradeepdewda-okta (commented):
What if we just stopPolling? doesn't it stop server calls ?

jhe-okta:
> What if we just stopPolling? doesn't it stop server calls ?

No. They are different. Stop polling, which is basically triggered by invoking `authenticatorChallenge-cancel`, takes user to the next step. But at the same time, device probing could still be happening which is invoked by setTimeout. If probing happens after user being taken to the next UI, it results in showing error message in the new UI.

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: This is out of scope for the fix considering the time urgency. This should be tackled along with https://oktainc.atlassian.net/browse/OKTA-276652.

pradeepdewda-okta:
> > What if we just stopPolling? doesn't it stop server calls ?
> 
> No. They are different. Stop polling, which is basically triggered by invoking `authenticatorChallenge-cancel`, takes user to the next step. But at the same time, device probing could still be happening which is invoked by setTimeout. If probing happens after user being taken to the next UI, it results in showing error message in the new UI.

Can we just clear the timeout then?

pradeepdewda-okta (approved):

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -160,7 +166,7 @@ const Body = BaseForm.extend(Object.assign(
             Logger.error(`Authenticator is not listening on port ${currentPort}.`);
             if (countFailedPorts === ports.length) {
               Logger.error('No available ports. Loopback server failed and polling is cancelled.');
-              this.options.appState.trigger('invokeAction', 'authenticatorChallenge-cancel');
+              this.options.appState.trigger('invokeAction', CANCEL_POLLING_ACTION);
```
pradeepdewda-okta, line 169: Do you still need this ?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -197,7 +208,7 @@ const Footer = BaseFooter.extend({
       this.add(Link, { options: {
         name: 'cancel-authenticator-challenge',
         label: loc('loopback.polling.cancel.link', 'login'),
-        actionPath: 'authenticatorChallenge-cancel',
+        actionPath: CANCEL_POLLING_ACTION,
```
pradeepdewda-okta, line 211: and this ?

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: Yes. This is to cancel the polling.

src/v2/view-builder/views/DeviceChallengePollView.js
jhe-okta: Yes. This is to cancel the polling.

yannongli-okta:
Do we also need this change in https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/DeviceChallengePollView.js?

PR #1578, opened 2020-12-19
fix: show error when polling returns 403

pradeepdewda-okta:
## Description:
shows error when polling request returns 403 and stop polling.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Screen Shot 2020-12-18 at 5 19 34 PM

### Issue:

- [OKTA-354467](https://oktainc.atlassian.net/browse/OKTA-354467)

### DIFF
--- a/src/v2/view-builder/views/PollView.js
+++ b/src/v2/view-builder/views/PollView.js
@@ -8,10 +8,11 @@ import { MS_PER_SEC } from '../utils/Constants';
 const PollMessageView = View.extend({
   template: hbs`
     <div class="ion-messages-container">
-      {{i18n code="poll.form.message" 
-        bundle="login" arguments="countDownCounterValue" $1="<span class='strong'>$1</span>"}}
+      <p>{{i18n code="poll.form.message" 
+        bundle="login" arguments="countDownCounterValue" $1="<span class='strong'>$1</span>"}} </p>
     </div>
-    <div class="okta-waiting-spinner"></div>`
+    <div class="okta-waiting-spinner"></div>
+    `
   ,
   getTemplateData () {
     const countDownCounterValue = this.options;
@@ -44,7 +45,13 @@ const Body = BaseForm.extend(Object.assign(
     remove () {
       BaseForm.prototype.remove.apply(this, arguments);
       this.stopPolling();
-    }
+    },
+
+    triggerAfterError () {
+      BaseForm.prototype.triggerAfterError.apply(this, arguments);
+      this.stopPolling();
+      this.$el.find('.o-form-fieldset-container').empty();
+    },
   },
 
   polling,
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -526,9 +526,10 @@ const safeModePoll = {
   ],
   '/idp/idx/poll': [
     'safe-mode-polling',
+    // 'safe-mode-polling',
     'safe-mode-polling',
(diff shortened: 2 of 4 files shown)


src/v2/view-builder/views/PollView.js
```diff
@@ -44,7 +45,13 @@ const Body = BaseForm.extend(Object.assign(
     remove () {
       BaseForm.prototype.remove.apply(this, arguments);
       this.stopPolling();
-    }
+    },
+
+    triggerAfterError (model, error) {
+      this.stopPolling();
```
haishengwu-okta, line 51: are you able to do
```
super.triggerAfterError()
this.stopPolling();
this.model.trigger('clearFormError');
```

pradeepdewda-okta: Can not use `clearFormError ` because need is to remove form content (that shows polling in progress) and just show the error..

Added a new handler instead. Thanks for the pointer, its much better now.

haishengwu-okta: Hide form content on error seems very specific to this view hence don't add event/handler to BaseForm but keep in this PollView?
Sorry for missing detail earlier.

pradeepdewda-okta: It would be helpful, there are many instances where devices also does the polling so having a generalized method will help, i think it's a good addition to keep it in baseForm.

haishengwu-okta: Oh, interesting, would that make sense to have a Base View for Polling to abstract the common logic?

pradeepdewda-okta: polling.js shares the basic workings of polling functions.. No harm creating a new BasePolling view, thats a good idea. Device's team use cases are complex for eg. `DeviceChallengePollView`

pradeepdewda-okta: In my opinion start and stop poll controls should be handled by the view and not by the baseview. I don't see there is much common logic left then.

haishengwu-okta: > there are many instances where devices also does the polling so having a generalized method will help, i think it's a good addition to keep it in baseForm.

My point is the removing polling view from Form is specific to Polling related views but not BaseForm hence worth to consider not to put into BaseForm but use other fashions.

pradeepdewda-okta: oh.. why i added to baseForm is it will be useful in future for the other views if they want to clear out form fields, like we clear out the form errors. I am also planning to add handler that can clear form errors instead of using `this.$el.find('.o-form-error-container').empty();` explicitly..

haishengwu-okta: iirc, BaseForm.clearError will clear all errors.

pradeepdewda-okta: we have an handler for that in BaseForm, but at some places devs don't used it. ```  handleClearFormError () {
    if (this.$('.o-form-error-container').hasClass('o-form-has-errors')) {
      this.clearErrors();
    }
  }```

haishengwu-okta: yeah, that's what I meant. It meant for people to use to clear errors. (could be because we introduce this event later and didn't find time to refactor all places to avoid redundancy.)

haishengwu-okta (approved):

PR #1588, opened 2020-12-30
sms v1 resend button css

anipendakur-okta:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Before -
https://oktainc.atlassian.net/browse/OKTA-348315?focusedCommentId=1656555

After - 
image

### Reviewers:
@haishengwu-okta

### Issue:

- [OKTA-348315](https://oktainc.atlassian.net/browse/OKTA-348315)

### DIFF
--- a/assets/sass/modules/_btns.scss
+++ b/assets/sass/modules/_btns.scss
@@ -49,17 +49,27 @@ $od-secondary-button-action-text-color: #124a94;
   }
 }
 
-.sms-request-button,
-.call-request-button,
-.email-request-button {
+.sms-request-button.link-button,
+.call-request-button.link-button {
   text-align: center;
   width: 140px;
-
+  padding-left: 6px;
+  padding-right: 6px;
   // This css will break the text into two lines if it's too long.
   white-space: normal;
   overflow: visible;
   line-height: 1;
   padding-top: 12px;
+
+  @include max-width-mq($container-width) {
+    width: 100%;
+    margin-bottom: 15px;
+  }
+}
+
+.email-request-button {
+  text-align: center;
+  @include max-width-text(130px);
 
   @include max-width-mq($container-width) {
     width: 100%;
--- a/assets/sass/modules/_btns.scss
+++ b/assets/sass/modules/_btns.scss
@@ -53,7 +53,13 @@ $od-secondary-button-action-text-color: #124a94;
 .call-request-button,
 .email-request-button {
   text-align: center;
-  @include max-width-text(130px);
+  width: 140px;
+
+  // This css will break the text into two lines if it's too long.
+  white-space: normal;
+  overflow: visible;
+  line-height: 1;
+  padding-top: 12px;
 
   @include max-width-mq($container-width) {
     width: 100%;
--- a/assets/sass/modules/_enroll.scss
+++ b/assets/sass/modules/_enroll.scss
@@ -137,7 +137,7 @@ $enroll-content-spacing: 15px;
 .enroll-sms {
 
   .enroll-sms-phone {
-    width: 171px;
+    width: 170px;
     float: left;
 
     @include device-mq(x-small) {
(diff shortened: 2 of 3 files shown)


anipendakur-okta:
@evelynteo-okta - I had to go with adjusting the width rather than a line-break which doesn't work with single line texts. Tagging you here as I've attached a screenshot in the description.

evelynteo-okta:
> @evelynteo-okta - I had to go with adjusting the width rather than a line-break which doesn't work with single line texts. Tagging you here as I've attached a screenshot in the description.

@anipendakur-okta Thank you, Ani. Works for Norwegian... Other languages will still have the text-clipping challenge.

assets/sass/modules/_enroll.scss
```diff
@@ -152,10 +152,16 @@ $enroll-content-spacing: 15px;
...
+    width: 136px;
     margin-top: 0;
     padding: 0;
 
+    // This css will break the text into two lines if it's too long.
+    white-space: normal;
+    overflow: visible;
+    line-height: 1;
+    padding-top: 12px;
+
```
anipendakur-okta, line 164: Adjusted the width from line 140 and 155 here. Then instead of ellipsis on long text, I've made it to break onto the next line - 
image

With the current mark-up (anchor tag for button and no parent div around it), this is the best I could do for multi-line texts that wouldn't change vertical placement of text for single-line texts on that button. Please let me know if you have a better solution.

anipendakur-okta: Without these changes, it would have been - 
image

Note: `extra` part is something I added for testing purposes, it's not a part of the actual norwegian string.

assets/sass/modules/_btns.scss
```diff
@@ -53,7 +53,13 @@ $od-secondary-button-action-text-color: #124a94;
...
 .email-request-button {
   text-align: center;
-  @include max-width-text(130px);
+  width: 140px;
+
+  // This css will break the text into two lines if it's too long.
+  white-space: normal;
+  overflow: visible;
+  line-height: 1;
+  padding-top: 12px;
```
anipendakur-okta, line 62: I could update the `max-width-text` mixin itself as it is not used anywhere else. But, I didn't do it for these reasons -
* Every element that includes it in the future will have the text fall to next line.
* Here since this PR only addresses special cases of buttons-next-to-inputs, I'd keep them subjected only to those cases.
* Important to note `padding-top` here as I'm changing the `line-height`.
* Given a chance and time, I'd want to change the mark-up and css. Mark-up changes would be to - 1. Use button element instead of anchor tag that would take care of these special i18n long text scenarios mostly. 2. Create a parent element to enclose the input and the button and make it a `flex` box. This would give a lot more control over how we want it rendered.

assets/sass/modules/_btns.scss
```diff
@@ -49,8 +49,24 @@ $od-secondary-button-action-text-color: #124a94;
...
+
+  @include max-width-mq($container-width) {
+    width: 100%;
+    margin-bottom: 15px;
+  }
+}
+
```
anipendakur-okta, line 69: Fix for verify flows (https://oktainc.atlassian.net/browse/OKTA-356087) for both `call` and `send sms` buttons. (In  verify, both call and sms buttons are next to the input)

anipendakur-okta:
Tested on IE11 (saucelabs), safari, FF and chrome today. Tested French, German & English mostly.
Some screenshots - 
image
image

image
image

haishengwu-okta (approved):

anipendakur-okta:
Note: Ran MFAVerifySMSIT selenium test via bacon on IE11 and it passed: 
https://bacon-go.aue1e.saasure.net/tasks/CB_TESTS_SAUCELABS?taskId=d86513d6b1074257b8413a623849ae97
This test had given us problems in the past and it was necessary to make sure it doesn't cause any problems when this is merged.

PR #1596, opened 2021-01-06
fix: webauthn enroll and verificaiton on v1

magizh-okta:
* Safari requires a direct user action to trigger the webauthn prompt, to allow use of platform authenticator.
 
 * For enroll, making the enroll activate call to backend and then invoking model save seems very convoluted and makes safari think that it is not a direct user interaction. Moving initial call outside save and to initialize section.

* For verify now the user has to click verify before seeing the prompt and we dont show webauthn prompts programmatically anymore.  

* Additionally for verify 
    
* Since other browsers also might adopt this security requirement making this fix for all browser flows. This serves as a general security fix for all browser and future proof.
    
RESOLVES: OKTA-357718

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/17267130/103944673-e2a48680-50e8-11eb-985f-a6abfa235199.mov

**Verify demo**
https://okta.box.com/s/g7xv07cyzxiszp9jeh2yhp3kppp24bm2 

### Issue:

- [OKTA-357718](https://oktainc.atlassian.net/browse/OKTA-357718)

### DIFF
--- a/assets/sass/modules/_enroll.scss
+++ b/assets/sass/modules/_enroll.scss
@@ -430,6 +430,7 @@ $enroll-content-spacing: 15px;
   .okta-waiting-spinner {
     margin-top: 20px;
     margin-bottom: 20px;
+    display: none;
   }
 
   .u2f-devices-images {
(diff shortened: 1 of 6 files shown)


magizh-okta:
came up with simpler solution https://github.com/okta/okta-signin-widget/pull/1599

PR #1598, opened 2021-01-11
feat: accept autocomplete attr on textbox - OKTA-326235

shuowu:
## Description:

Accepts `autoComplete` option in TextBox and PasswordBox components to override the `autocomplete` attribute (default as `off`) of form input.

This PR only adds autocomplete attr for `PrimaryAuthForm`.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-326235](https://oktainc.atlassian.net/browse/OKTA-326235)


src/widget/OktaSignIn.js
```diff
@@ -1,12 +1,17 @@
...
 import Errors from 'util/Errors';
 import OAuth2Util from 'util/OAuth2Util';
 import Util from 'util/Util';
 import createAuthClient from 'widget/createAuthClient';
 import V1Router from 'LoginRouter';
 import V2Router from 'v2/WidgetRouter';
 
+// Enable "autoComplete" attribute
+const { FormUtil } = internal.views.forms.helpers;
+FormUtil.INPUT_OPTIONS.push('autoComplete');
```
haishengwu-okta, line 13: Why not change Courage?

haishengwu-okta (changes requested):

src/widget/OktaSignIn.js
shuowu: Initially, I was planning to limit the change in SIW only, but looks like the change should be safe to stay in courage. PR added here: https://github.com/okta/okta-ui/pull/3371

src/models/Settings.js
```diff
@@ -472,4 +472,8 @@ export default Model.extend({
 
     return options;
   },
+
+  isDsTheme: function () {
```
haishengwu-okta, line 476: FYI:
BaseForm expects the `this.settings` is instance of `Courage/SettingsModel` but apparently SIW creates its own settings (`src/models/Settings`) which is incompatible with Courage's SettingsModel.

haishengwu-okta (approved):
🚀

aarongranick-okta (commented):
I tested with the latest build `5.3.0-g8bf77f4`. I see that it is now saving and suggesting the username field! However, it is not offering to save password.

PR #1599, opened 2021-01-11
fix: webauthn enroll and verify fix on v1

magizh-okta:
* Webauthn enroll and verify now works on v1 in safari on bigsur and ios13.
    * The change to verify flow is common across all browsers to make sure we are future proof
    and follow security best practices ( explicit user click before webauthn prompt)
    
    * This needs upgrade of auth js to 4.6.0 to consume fix that uses borwser fetch instead of cross fetch as
    cross fetch wraps fetch with unnecessary promises causing safari to think the call is programatic and not due to direct user interaction
    even if webauthn prompt is triggered on api callback after user click.
    
auth-js pr -    https://github.com/okta/okta-auth-js/pull/585
    
    RESOLVES: OKTA-357718
    

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://drive.google.com/file/d/1yJeN_b85EkM2gxh5JtqYAjt1WRKi0eRd/view?usp=sharing

### Issue:

- [OKTA-357718](https://oktainc.atlassian.net/browse/OKTA-357718)

### DIFF
--- a/src/VerifyWebauthnController.js
+++ b/src/VerifyWebauthnController.js
@@ -204,7 +204,7 @@ export default FormController.extend({
 
     _stopEnrollment: function () {
       this.$('.okta-waiting-spinner').hide();
-      this.$('.o-form-button-bar [type="submit"]')[0].value = loc('retry', 'login');
+      this.$('.o-form-button-bar [type="submit"]')[0].value = loc('verify.u2f.retry', 'login');
       this.$('.o-form-button-bar').show();
     },
   },
--- a/src/VerifyWebauthnController.js
+++ b/src/VerifyWebauthnController.js
@@ -139,10 +139,10 @@ export default FormController.extend({
   Form: {
     autoSave: true,
     hasSavingState: false,
-    title: _.partial(loc, 'factor.webauthn.biometric', 'login'),
+    title: loc('factor.webauthn.biometric', 'login'),
     className: 'verify-webauthn-form',
     noCancelButton: true,
-    save: _.partial(loc, 'verify.u2f.retry', 'login'),
+    save: loc('mfa.challenge.verify', 'login'),
     noButtonBar: function () {
       return !webauthn.isNewApiAvailable();
     },
@@ -197,16 +197,6 @@ export default FormController.extend({
       return children;
     },
 
-    postRender: function () {
-      _.defer(() => {
-        if (webauthn.isNewApiAvailable()) {
-          this.model.save();
-        } else {
-          this.$('[data-se="webauthn-waiting"]').hide();
-        }
-      });
-    },
-
     _startEnrollment: function () {
       this.$('.okta-waiting-spinner').show();
       this.$('.o-form-button-bar').hide();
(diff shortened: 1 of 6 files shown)


test/unit/helpers/mocks/Util.js
```diff
@@ -64,6 +65,8 @@ fn.mockDuo = function () {
 };
 
 fn.mockAjax = function (responses) {
+  globalFetch = window.fetch;
+  window.fetch = null;
```
gowthamidommety-okta, line 69: Why was this change made?

gowthamidommety-okta (approved):
Just a question for my understanding. Rest looks good! 🚢

test/unit/helpers/mocks/Util.js
magizh-okta: since we mock ajax requests and dont want to use fetch we need to force using ajax and not fetch.

aarongranick-okta (approved):

src/VerifyWebauthnController.js
```diff
@@ -139,10 +139,10 @@ export default FormController.extend({
   Form: {
     autoSave: true,
     hasSavingState: false,
-    title: _.partial(loc, 'factor.webauthn.biometric', 'login'),
```
mauriciocastillosilva-okta, line 142: we shouldn't change this. If I remember correctly we needed `_partial` because we had some problem loading a new language since the language was not loaded in time when this was calculated. Adding this as a function was the workaround

src/VerifyWebauthnController.js
```diff
@@ -197,23 +197,14 @@ export default FormController.extend({
       return children;
     },
 
-    postRender: function () {
-      _.defer(() => {
-        if (webauthn.isNewApiAvailable()) {
-          this.model.save();
```
mauriciocastillosilva-okta, line 203: who makes this `save` call now? Or we didn't need it before? 
Oh wait... or is it that now we don't move the UI automatically to CHALLENGE, but instead now the user clicks 'verify' button, which submits the form and makes a `model.save` request, which is what replaces this specific call. In summary we don't make this call automatically for a user, which to me makes sense

magizh-okta: yeah and this is what Safari expects and allows.

src/VerifyWebauthnController.js
magizh-okta: nice catch.

magizh-okta: @haishengwu-okta  how do we avoid this issue in idx ? Also I wonder if this issue can be solved in some other way.

haishengwu-okta: we need to use same approach in IDX.

src/VerifyWebauthnController.js
```diff
@@ -197,23 +197,14 @@ export default FormController.extend({
...
-    },
-
     _startEnrollment: function () {
       this.$('.okta-waiting-spinner').show();
       this.$('.o-form-button-bar').hide();
     },
 
     _stopEnrollment: function () {
       this.$('.okta-waiting-spinner').hide();
+      this.$('.o-form-button-bar [type="submit"]')[0].value = loc('retry', 'login');
```
mauriciocastillosilva-okta, line 207: should be key `verify.u2f.retry` instead... I think that is what we had before, in case someone has customized it

mauriciocastillosilva-okta (approved):
just one comment, fix that properties key and 🐑

PR #1617, opened 2021-01-20
feat: accept authClient option

aarongranick-okta:
## Description:
Adds new option `authClient` which is an [AuthJS](https://github.com/okta/okta-auth-js) instance.
If `authClient` is passed, `authParams` will be ignored.

Used together with one of our other SDKs like [okta-react](https://github.com/okta/okta-react), this allows the SDKs to share a single instance of AuthJS. This will avoid some issues that can occur with multiple instances of AuthJS running on the same storage. A single instance ensures the same code is accessing the same storage with the same config in each SDK.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-361258](https://oktainc.atlassian.net/browse/OKTA-361258)

### DIFF
--- a/README.md
+++ b/README.md
@@ -590,6 +590,37 @@ signIn.authClient.session.exists()
 });
 ```
 
+The `authClient` can be set directly in the configuration:
+
+```javascript
+var authClient = new OktaAuth({
+  issuer: 'https://{yourOktaDomain}/oauth2/default',
+  clientId: '{yourClientId}'
+});
+var config = {
+  baseUrl: 'https://{yourOktaDomain}',
+  authClient: authClient
+};
+
+var signIn = new OktaSignIn(config);
+// signIn.authClient === authClient
+```
+
+If no `authClient` option is set, an instance will be created using `authParams`:
+
+```javascript
+var config = {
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1206,7 +1237,10 @@ OIDC flow is required for [Social Login][].
     oAuthTimeout: 300000 // 5 minutes
     ```
 
-- **authParams:** An object containing configuration which is passed directly to the internal `authClient`. Selected options are described below. See the full set of [Configuration options](https://github.com/okta/okta-auth-js#configuration-options)
+- **authClient:** An [AuthJS](https://github.com/okta/okta-auth-js) instance. This will be available on the widget instance as the [authClient](#authclient) property. 
+ **Note:** If the `authClient` option is used, `authParams` will be ignored.
+
+- **authParams:** An object containing configuration which is used to create the internal [authClient`](#authclient). Selected options are described below. See the full set of [Configuration options](https://github.com/okta/okta-auth-js#configuration-options)
```
swiftone, line 1243: Should we state explicitly that this is ignored if `authClient` is set?

swiftone (approved):

PR #1622, opened 2021-01-26
feat: Adds support for Custom Links for MFA pages (OKTA-359531)

denysoblohin-okta:
## Description:

Adds support for Custom Links for MFA pages.
To enable, add `helpLinks.factorPageCustomLink` object to config.
Example:
```js
{
  helpLinks: {
    factorPageCustomLink: {
      text: 'Need help with MFA?',
      href: 'https://acme.com/mfa-help',
    },
    // ... other custom links
  },
  // ... other config options
}
```
(both `text` and `href` should be set to enable custom link)
Scope of this feature: v1 widget only.
Implemented with new view `FooterMFA` extended from `FooterSignout`

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
How it looks for all factor verification pages:
Screenshot 2021-02-08 at 20 40 09
Screenshot 2021-02-08 at 20 40 40
Screenshot 2021-02-08 at 20 41 01

If `features.hideSignOutLinkInMFA` is true or `features.mfaOnlyFlow` is true:
Screenshot 2021-01-27 at 21 14 57

### Reviewers:
Haisheng Wu  @haishengwu-okta

### Issue:

- [OKTA-359531](https://oktainc.atlassian.net/browse/OKTA-359531)

### DIFF
--- a/README.md
+++ b/README.md
@@ -942,6 +942,10 @@ Set the following config options to override the help link URLs on the Primary A
 helpLinks: {
   help: 'https://acme.com/help',
   forgotPassword: 'https://acme.com/forgot-password',
+  factorPageCustomLink: {
+    text: 'Need help with MFA?',
+    href: 'https://acme.com/mfa-help',
+  },
   unlock: 'https://acme.com/unlock-account',
   custom: [
     {
@@ -962,6 +966,8 @@ helpLinks: {
 - **helpLinks.forgotPassword** - Custom link href for the "Forgot Password" link
 
 - **helpLinks.unlock** - Custom link href for the "Unlock Account" link. For this link to display, `features.selfServiceUnlock` must be set to `true`, and the self service unlock feature must be enabled in your admin settings.
+
+- **helpLinks.factorPageCustomLink** - Custom link object `{text, href}` that will be added to all MFA pages.
 
 - **helpLinks.custom** - Array of custom link objects `{text, href, target}` that will be added to the *"Need help signing in?"* section. The `target` of the link is optional.
 
--- a/README.md
+++ b/README.md
@@ -942,7 +942,7 @@ Set the following config options to override the help link URLs on the Primary A
 helpLinks: {
   help: 'https://acme.com/help',
   forgotPassword: 'https://acme.com/forgot-password',
-  factorPageCustomLink: {
+  factorPage: {
     text: 'Need help with MFA?',
     href: 'https://acme.com/mfa-help',
   },
@@ -967,7 +967,7 @@ helpLinks: {
 
 - **helpLinks.unlock** - Custom link href for the "Unlock Account" link. For this link to display, `features.selfServiceUnlock` must be set to `true`, and the self service unlock feature must be enabled in your admin settings.
 
-- **helpLinks.factorPageCustomLink** - Custom link object `{text, href}` that will be added to all MFA pages.
+- **helpLinks.factorPage** - Custom link object `{text, href}` that will be added to all MFA pages.
 
 - **helpLinks.custom** - Array of custom link objects `{text, href, target}` that will be added to the *"Need help signing in?"* section. The `target` of the link is optional.
 
--- a/src/MfaVerifyController.js
+++ b/src/MfaVerifyController.js
@@ -11,7 +11,7 @@
  */
 
 /* eslint complexity: [2, 21] max-statements: [2, 25] max-params: 0 */
-import { _, loc, internal } from 'okta';
+import { loc, internal } from 'okta';
 import BaseLoginController from 'util/BaseLoginController';
 import EmailMagicLinkForm from 'views/factor-verify/EmailMagicLinkForm';
 import InlineTOTPForm from 'views/mfa-verify/InlineTOTPForm';
@@ -133,10 +133,7 @@ export default BaseLoginController.extend({
       }
     });
 
-    this.add(new FooterMFA(_.extend(this.toJSON(), {
-      showLink: !this.settings.get('features.hideSignOutLinkInMFA') &&
-        !this.settings.get('features.mfaOnlyFlow')
-    })));
+    this.add(new FooterMFA(this.toJSON()));
   },
 
   findModel: function (factorType, options) {
--- a/src/MfaVerifyController.js
+++ b/src/MfaVerifyController.js
@@ -11,7 +11,7 @@
  */
 
 /* eslint complexity: [2, 21] max-statements: [2, 25] max-params: 0 */
-import { loc, internal } from 'okta';
+import { _, loc, internal } from 'okta';
 import BaseLoginController from 'util/BaseLoginController';
 import EmailMagicLinkForm from 'views/factor-verify/EmailMagicLinkForm';
 import InlineTOTPForm from 'views/mfa-verify/InlineTOTPForm';
@@ -22,7 +22,7 @@ import SecurityQuestionForm from 'views/mfa-verify/SecurityQuestionForm';
 import SendEmailAndVerifyCodeForm from 'views/mfa-verify/SendEmailAndVerifyCodeForm';
 import TOTPForm from 'views/mfa-verify/TOTPForm';
 import YubikeyForm from 'views/mfa-verify/YubikeyForm';
-import FooterSignout from 'views/shared/FooterSignout';
+import FooterMFA from 'views/shared/FooterMFA';
 let { CheckBox } = internal.views.forms.inputs;
 export default BaseLoginController.extend({
   className: 'mfa-verify',
@@ -133,10 +133,10 @@ export default BaseLoginController.extend({
       }
     });
 
-    if (!this.settings.get('features.hideSignOutLinkInMFA') &&
-        !this.settings.get('features.mfaOnlyFlow')) {
-      this.add(new FooterSignout(this.toJSON()));
-    }
+    this.add(new FooterMFA(_.extend(this.toJSON(), {
+      showLink: !this.settings.get('features.hideSignOutLinkInMFA') &&
+        !this.settings.get('features.mfaOnlyFlow')
+    })));
   },
 
   findModel: function (factorType, options) {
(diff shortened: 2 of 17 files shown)


src/views/shared/FooterSignout.js
```diff
@@ -17,31 +17,32 @@ const { Util } = internal.util;
 export default View.extend({
   template: hbs(
     '\
-      <a href="#" class="link {{linkClassName}}" data-se="signout-link">\
+      <a href="#" class="link signout {{linkClassName}}" data-se="signout-link">\
```
haishengwu-okta, line 20: why want to add `signout` all the time. It was either `signout` or `back` before?

src/RecoveryChallengeController.js
```diff
@@ -15,7 +15,7 @@ import hbs from 'handlebars-inline-precompile';
 import Enums from 'util/Enums';
 import FormController from 'util/FormController';
 import FormType from 'util/FormType';
-import FooterSignout from 'views/shared/FooterSignout';
+import FooterMFA from 'views/shared/FooterMFA';
```
haishengwu-okta, line 18: Don't think we need to add MFA verify link for recovery flow (RecoveryChallenge, RecoveryQuestion).
Let's follow up with PM in Slack.

src/VerifyDuoController.js
```diff
@@ -122,7 +122,7 @@ export default FormController.extend({
     },
   },
 
-  Footer: FooterSignout,
+  Footer: FooterMFA,
```
haishengwu-okta, line 125: Looks like in `VerifyDuo`, `VerifyU2F`, `VerifyWebauthn`, `VerifyWindowsHello`
Footer was added without checking feature `features.hideSignOutLinkInMFA` and `features.mfaOnlyFlow`.
@gowthamidommety-okta @mauriciocastillosilva-okta do you know if this is intentional or bug?

test/unit/spec/MfaVerify_spec.js
```diff
@@ -2283,6 +2317,70 @@ Expect.describe('MFA Verify', function () {
...
+      itp('has the correct text', function () {
+        return setupWithFirstFactor({ factorType: 'question' }, configWithCustomLink).then(function (test) {
+          expect(test.form.factorPageCustomLinkLabel($sandbox).trim()).toBe('Need help with MFA?');
+        });
+      });
+      itp('has the correct url', function () {
+        return setupWithFirstFactor({ factorType: 'question' }, configWithCustomLink).then(function (test) {
+          expect(test.form.factorPageCustomLinkHref($sandbox).trim()).toBe('https://acme.com/mfa-help');
+        });
+      });
```
haishengwu-okta, line 2335: merged into one test case

src/views/shared/FooterSignout.js
denysoblohin-okta: I added class `signout` to bind click event `handleSignout` only for this link and to not bind it to custom help link.
I've removed class `signout` and now use `click a[data-se="signout-link"]` as selector for click event that will b handles with `handleSignout`

src/RecoveryChallengeController.js
denysoblohin-okta: Removed

test/unit/spec/MfaVerify_spec.js
denysoblohin-okta: Done

src/VerifyDuoController.js
gowthamidommety-okta: Seems like something we should fix. The only reason I can think of for why we did not add this feature here, is that these factors may not be used for mfa only flows. However, I don't see any harm in adding a check for this feature. In fact, i feel like the feature would be more complete if we added it. @mauriciocastillosilva-okta thoughts?

mauriciocastillosilva-okta: Sounds right, we probably missed those. The original change for `hideSignOutLinkInMFA` only covered the generic `MFAVerifyController`. 
So +1, it is a bug, since all factors should behave the same about show/hide the footer.

haishengwu-okta: Thank you @gowthamidommety-okta @mauriciocastillosilva-okta !

@denysoblohin-okta we could
- either fix in this PR
- open another PR to fix this inconsistent first 

Then your PR would looks a little clear (e.g. instead of passing `showLink`, we can check the feature flag within `FooterMFA`)

denysoblohin-okta: @haishengwu-okta https://github.com/okta/okta-signin-widget/pull/1635

src/MfaVerifyController.js
```diff
@@ -133,10 +133,10 @@ export default BaseLoginController.extend({
       }
     });
 
-    if (!this.settings.get('features.hideSignOutLinkInMFA') &&
-        !this.settings.get('features.mfaOnlyFlow')) {
-      this.add(new FooterSignout(this.toJSON()));
-    }
+    this.add(new FooterMFA(_.extend(this.toJSON(), {
```
haishengwu-okta, line 136: Since the FooterMFA is only meant for MFA page, we can move `!this.settings.get('features.hideSignOutLinkInMFA') &&
         !this.settings.get('features.mfaOnlyFlow')` logic inside `FooterMFA` ?

denysoblohin-okta: Done

haishengwu-okta (approved):

README.md
```diff
@@ -942,6 +942,10 @@ Set the following config options to override the help link URLs on the Primary A
 helpLinks: {
   help: 'https://acme.com/help',
   forgotPassword: 'https://acme.com/forgot-password',
+  factorPageCustomLink: {
```
mauriciocastillosilva-okta, line 945: all of these, inside `helpLinks` are custom links right? so maybe we don't need to name them "custom*"? Would `factorPage` be enough?

README.md
```diff
@@ -942,6 +942,10 @@ Set the following config options to override the help link URLs on the Primary A
...
+    text: 'Need help with MFA?',
+    href: 'https://acme.com/mfa-help',
+  },
   unlock: 'https://acme.com/unlock-account',
   custom: [
```
mauriciocastillosilva-okta, line 950: in the same topic as my comment above, this should probably have been named "signIn" or so? Not anything to change in this PR, just pointing it out to see if it makes sense to you and others, and we can have a discussion if needed later

src/views/shared/FooterMFA.js
```diff
@@ -0,0 +1,47 @@
...
+  template: hbs(
+    '\
+      {{#if hasFactorPageCustomLink}}\
+        <a href="{{factorPageCustomLinkHref}}" data-se="factor-page-custom-link" \
+          class="link js-factor-page-custom-link" \
+          rel="noopener noreferrer" target="_blank">\
+          {{factorPageCustomLinkText}}\
+        </a>\
+      {{/if}}\
+      {{#if showLink}}\
```
mauriciocastillosilva-okta, line 26: we should have each link in a different line. Reason is that in other languages, if we have them side by side, there might be some overlap or inconsistencies

src/views/shared/FooterMFA.js
```diff
@@ -0,0 +1,47 @@
...
+        <a href="#" class="link {{linkClassName}}" data-se="signout-link">\
```
mauriciocastillosilva-okta, line 27: would it make sense to have this link string in a constant in FooterSignout and just append it here? Just wondering so you don't have to rewrite the exact same html for it here and in FooterSignout

src/views/shared/FooterMFA.js
haishengwu-okta: @vincentvoong-okta ^ any thoughts ?

README.md
denysoblohin-okta: Done

README.md
denysoblohin-okta: You mean replace "custom" -> "signIn", right?

src/views/shared/FooterMFA.js
denysoblohin-okta: Implemented this with reusing FooterSignout's precompiled template function:
https://github.com/okta/okta-signin-widget/pull/1622/commits/bd59ea36ee6a9c838711cd205d1649b7c3693750

README.md
haishengwu-okta: @denysoblohin-okta 
Just to be clear, this is unrelated to the change you're working and can be deal with separately.
Also it's breaking change (interface changed for external devs) hence probably meant for next major release.

src/MfaVerifyController.js
```diff
@@ -133,10 +133,7 @@ export default BaseLoginController.extend({
       }
     });
 
-    if (!this.settings.get('features.hideSignOutLinkInMFA') &&
-        !this.settings.get('features.mfaOnlyFlow')) {
-      this.add(new FooterSignout(this.toJSON()));
-    }
+    this.add(new FooterMFA(this.toJSON()));
```
haishengwu-okta, line 136: do you have to pass `this.toJSON()` ?

denysoblohin-okta: 1. I didn't look into okta core, but looks like we need to pass `this.toJSON()` to okta views constructor
2. Other controllers are extended from FormController and can use `Footer` property: https://github.com/okta/okta-signin-widget/blob/master/src/util/FormController.js#L98
MfaVerifyController is extended from BaseLoginController, which don't have such functionality

haishengwu-okta: In `VerifyCustomFactorController`, you did `this.addFooter(FooterMFA);`, looks like it's working?

src/views/shared/FooterMFA.js
vincentvoong-okta: @mauriciocastillosilva-okta @haishengwu-okta 
that sounds good to me. it will look like this: 
image

denysoblohin-okta: Done

PR #1625, opened 2021-01-27
feat: update registration input length error messages

sherizada-okta:
## Description:

On the v1 Registration form, update string length validation error messages to be more helpful:

1.  
    - old message: "This field cannot be less than the minimum required characters"
    - new message: "Please enter something at least {x} characters long"
2.
   - old message: "This field cannot exceed the maximum allowed characters"
   - new message: "Please enter something shorter than {x} characters"

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Why does the css intentionally break on characters, pretty annoying
Screen Shot 2021-01-27 at 1 57 39 PM
Screen Shot 2021-01-27 at 1 57 29 PM
Screen Shot 2021-01-27 at 1 29 44 PM

### Issue:

- [OKTA-347240](https://oktainc.atlassian.net/browse/OKTA-347240)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -642,6 +642,8 @@ registration.default.callbackhook.error=We could not process your registration a
 registration.error.userName.invalidEmail=Invalid email address
 registration.error.password.passwordRequirementsNotMet=Password requirements were not met
 registration.error.userName.notUniqueWithinOrg=An account with that email already exists
+registration.model.validation.field.string.too.short = Please enter something at least {0} characters long
+registration.model.validation.field.string.too.long = Please enter something shorter than {0} characters
 
 # PIV Auth
 piv.card=PIV Card
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -198,6 +198,18 @@ export default BaseLoginController.extend({
         noCancelButton: true,
         title: loc('registration.form.title', 'login'),
         save: loc('registration.form.submit', 'login'),
+        initialize() {
+          this.listenTo(this.model, 'invalid error', (model, errorResp) => {
+            // overwrite courage errorResp object to show custom error message
+            for (let formFieldName in errorResp) {
+              if (errorResp[formFieldName] === 'model.validation.field.string.minLength') {
+                errorResp[formFieldName] = loc('registration.model.validation.field.string.tooShort', 'login', [formFieldName, model.props[formFieldName].minLength]);
+              } else if (errorResp[formFieldName] === 'model.validation.field.string.maxLength') {
+                errorResp[formFieldName] = loc('registration.model.validation.field.string.tooLong', 'login', [formFieldName, model.props[formFieldName].maxLength]);
+              }
+            }
+          });
+        },
         parseErrorMessage: function (resp) {
           const hasErrorCauses = resp.errorCauses && resp.errorCauses.length;
 
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -199,7 +199,7 @@ export default BaseLoginController.extend({
         title: loc('registration.form.title', 'login'),
         save: loc('registration.form.submit', 'login'),
         initialize () {
-          this.listenTo(this.model, 'invalid error', function (model, errorResp) {
+          this.listenTo(this.model, 'invalid', function (model, errorResp) {
             // overwrite courage errorResp object to show custom error message
             for (let formFieldName in errorResp) {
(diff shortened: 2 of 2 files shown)


sherizada-okta:
I know this is a brittle way to add custom error messages, but it's the only way I have found so far to contain this change to just the SIW registration controller.
I wasn't able to override `validate` here. I would prefer to make the change to the courage model validation like here https://github.com/okta/okta-ui/pull/3420/files, but that would impact all forms.

src/RegistrationController.js
```diff
@@ -198,6 +198,18 @@ export default BaseLoginController.extend({
         noCancelButton: true,
         title: loc('registration.form.title', 'login'),
         save: loc('registration.form.submit', 'login'),
+        initialize() {
+          this.listenTo(this.model, 'invalid error', (model, errorResp) => {
```
haishengwu-okta, line 202: are you able to modify in function `parseErrorMessage` below?

sherizada-okta: `parseErrorMessage` is not triggered on form field validation

pradeepdewda-okta: @haishengwu-okta It's on-focus errors, not server errors. It should go to validate() function of the model. Can you suggest any better way to do this ?

sherizada-okta: and by form field validation, I validating when an input form field loses focus

src/RegistrationController.js
```diff
@@ -198,6 +198,22 @@ export default BaseLoginController.extend({
         noCancelButton: true,
         title: loc('registration.form.title', 'login'),
         save: loc('registration.form.submit', 'login'),
+        initialize () {
```
pradeepdewda-okta, line 201: If we can't find a better to do this early morning.. then i am fine with this approach.

we can add
`
modelEvents : {
'invalid error' : 'showErrors'
}`

and let showError handle the logic..

pradeepdewda-okta: until if @haishengwu-okta or @magizh-okta have any better idea..

sherizada-okta: @pradeepdewda-okta `showErrors` is not available to the Form, so I would need to add another Form method to call the controller.showErrors. And the invalid error object for the form field is different than the ajax error object showErrors expects.

src/RegistrationController.js
haishengwu-okta: oh I see. my bad. does that mean we totally ignore the message from backend?

sherizada-okta: @haishengwu-okta that is not the intention, I think I should remove the `error` event listener, this should only change model validation messages

src/RegistrationController.js
pradeepdewda-okta: @sherizada-okta add a new handler.. not the base showError method. sorry for the confusion.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -642,6 +642,8 @@ registration.default.callbackhook.error=We could not process your registration a
 registration.error.userName.invalidEmail=Invalid email address
 registration.error.password.passwordRequirementsNotMet=Password requirements were not met
 registration.error.userName.notUniqueWithinOrg=An account with that email already exists
+registration.model.validation.field.string.too.short = Please enter something at least {0} characters long
```
haishengwu-okta, line 645: @pradeepdewda-okta @sherizada-okta 
feels like we really show update Model props and validation function (https://github.com/okta/okta-ui/blob/master/packages/courage/src/framework/Model.js#L304)
or even change Courage (there was discussion on this but didn't find it now)

src/RegistrationController.js
```diff
@@ -198,6 +198,21 @@ export default BaseLoginController.extend({
         noCancelButton: true,
         title: loc('registration.form.title', 'login'),
         save: loc('registration.form.submit', 'login'),
+        modelEvents : { 'invalid' : 'modifyErrors' },
+        modifyErrors: function (model, errorResp) {
```
haishengwu-okta, line 202: @pradeepdewda-okta @sherizada-okta 
feels like we really should update Model props and validation function (https://github.com/okta/okta-ui/blob/master/packages/courage/src/framework/Model.js#L304)
or even change Courage (there was discussion on this but didn't find it now)

pradeepdewda-okta: @sherizada-okta We can create a refactor ticket for this to do it properly.

haishengwu-okta: Thank you!

PR #1627, opened 2021-01-27
OKTA-364172: reload playground when mock data and configs are changed

lesterchoi-okta:
## Description:
Allows Okta SIW developers to see changes to playground configs reflected automatically without the need to manually reload the browser.

* uses `nodemon` and webpack `devServer.proxy` to restart mock server when files in `/playground` change.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Reviewers:

@haishengwu-okta 

### Issue:

- [OKTA-364172](https://oktainc.atlassian.net/browse/OKTA-364172)

### DIFF
--- a/webpack.playground.config.js
+++ b/webpack.playground.config.js
@@ -3,11 +3,12 @@
 
 const path = require('path');
 const fs = require('fs');
-const dyson = require('dyson');
+const nodemon = require('nodemon');
 
 const TARGET = path.resolve(__dirname, 'target');
 const PLAYGROUND = path.resolve(__dirname, 'playground');
-const DEFAULT_SERVER_PORT = 3000;
+const DEV_SERVER_PORT = 3000;
+const MOCK_SERVER_PORT = 3030;
 const WIDGET_RC_JS = '.widgetrc.js';
 const WIDGET_RC = '.widgetrc';
 
@@ -34,7 +35,6 @@ module.exports = {
     filename: 'playground.bundle.js',
   },
   devtool: 'source-map',
-
   module: {
     rules: [
       {
@@ -47,7 +47,7 @@ module.exports = {
       },
     ]
   },
-
+  watch: true,
   devServer: {
     contentBase: [
       PLAYGROUND,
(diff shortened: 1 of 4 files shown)


webpack.playground.config.js
```diff
@@ -55,8 +57,9 @@ module.exports = {
       // webpack-dev-server v2 only watch contentbase on root level
```
lesterchoi-okta, line 57: This comment already explains exactly why the changes weren't being detected.
The added helper function helps set the paths explicitly.
Alternatively, we could hardcode the paths if they don't change frequently.

helpers.js
```diff
@@ -0,0 +1,32 @@
...
+const path = require('path');
+
+/**
+ * Recursively get directories in the path. Adjust maximum depth by setting maxDepth (default: -1)
+ * Paths will be relative or absoluate based on startPath
+ * @param {string} startPath the path to start from
+ * @param {number} maxDepth maximum depth to recurse
+ * @returns {array} array of paths as strings
+ */
+module.exports.getDirs = (startPath, maxDepth = -1) => {
```
lesterchoi-okta, line 11: Should move to a separate package or module and be tested.
Checked using `console.time`; takes ~4ms on `playground` directory.

lesterchoi-okta (commented):
Open to suggestions on implementation strategy.
1. Hardcode paths
2. Move helper to package

webpack.playground.config.js
haishengwu-okta: wow... this is simply. 🤦 
I think we can hard code the path, which are not changed frequently.

lesterchoi-okta: Will do. Did you confirm it's reloading the way you want?

pradeepdewda-okta:
@lesterchoi-okta wow thanks for doing this.. very helpful. Was on mind for a while.

webpack.playground.config.js
haishengwu-okta: Did you verify if the mock data has been reflected after webpack reload?

lesterchoi-okta: @haishengwu-okta Oops. Missed your edit. I just pushed the hardcoded version.
I can revert if you think the helper version is better, but I'd say the hardcoded paths makes it clear that subdirectories need to be added explicitly. If anything I should add a note to that section explaining that devs should update the webpack config with new paths if they're adding new folders.

lesterchoi-okta: > Did you verify if the mock data has been reflected after webpack reload?

Yes, but I'll check again.

webpack.playground.config.js
```diff
@@ -55,8 +54,30 @@ module.exports = {
...
       // explicitly list folders to watch for browser auto reload
       // sub-folders can be removed when upgrade to webpack-dev-server v3
-      `${TARGET}/js`,
-      `${TARGET}/css`,
+      path.join(TARGET, 'js'),
```
pradeepdewda-okta, line 57: this is too verbose.. I like helper better. we can add comment for explanation.

webpack.playground.config.js
haishengwu-okta: looks like it will not reload the mock server

lesterchoi-okta: IMHO the real issue is that these changes to `responseConfig.json` are tracked in git.

The **better** setup would be to have a sensible/common default for `responseConfig.json`, then have untracked "local" (i.e., personal or team) variations so devs aren't clobbering each other's test flows.

The **ideal** setup might be if it's a completely dynamic response that either...
1. uses magic test values to trigger specific responses (documentation required of course)
2. allows you to configure the responses via CLI (wishful)

lesterchoi-okta: > looks like it will not reload the mock server

Looking into it

haishengwu-okta: yeah, that's kind of real issue and #2 sounds good to me (CLI, or Web). 
but I'd assume it would take extra effort.
The ultimate goal is to make dev easier to use different set up of mocks.

#1 would require set up too many combinations.

lesterchoi-okta: > looks like it will not reload the mock server

Okay. Yes, definitely jumped the gun on this (I must've restarted the dev-server manually without realizing). The routes are registered in the `before` hook in webpack-dev-server which is run once at the start, and not triggered for re-compilations. I'll investigate ways to make it happen on file changes.

webpack.playground.config.js
lesterchoi-okta: @pradeepdewda-okta turns out this won't update the mock server anyway.
I'm going to cancel this PR and re-submit with a fix.

lesterchoi-okta (commented):
Any thoughts on what, if any, updates I should make on `/playground/README.md`?

haishengwu-okta (approved):

haishengwu-okta:
this is partially feature/partially bug fixing. so didn't think of change to the readme.
but if you find anything can improve for the readme, feel free to open PR.

lesterchoi-okta:
@pradeepdewda-okta Passes [on Bacon](https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&sha=16b7143076896a3b40a4f828831e71b5b8ff0370)

I think this just needs "code owner review from okta/devex-engineering"

NOTE: Travis CI failure changes from run to run, guess there are some flaky tests?

PR #1628, opened 2021-01-28
OKTA-353857: test SDK errors are emitted through triggerAfterError handler

oleksandrpravosudko-okta:
## Description:
* test SDK errors are emitted through triggerAfterError handler
* add missing tests for OAuth2Util

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-353857](https://oktainc.atlassian.net/browse/OKTA-353857)

### DIFF
--- a/test/unit/spec/OAuth2Util_spec.js
+++ b/test/unit/spec/OAuth2Util_spec.js
@@ -1,14 +1,122 @@
 /* eslint max-len: [2, 140] */
-import Util from 'util/OAuth2Util';
+import OAuth2Util from 'util/OAuth2Util';
+import Util from 'util/Util';
+import createAuthClient from 'widget/createAuthClient';
+import Settings from '../../../src/models/Settings';
+import { AuthSdkError } from '@okta/okta-auth-js';
+import Enums from '../../../src/util/Enums';
 
-// TODO: rewrite
+
 describe('util/OAuth2Util', function () {
-  
+  class MockModel {
+    constructor () {
+      this.trigger = function () {};
+    }
+  }
+
+  class MockController {
+    constructor () {
+      this.model = new MockModel();
+      this.trigger = function () {};
+    }
+  }
   describe('getTokens', function () {
-    // TODO
+    let controller;
+    let authClient;
+    let settings;
+
+    beforeEach(function () {
+      controller = new MockController();
+      authClient = createAuthClient({issuer: 'https://foo/default'});
+      settings = new Settings({
+        baseUrl: 'https://foo'
+      });
+      settings.setAuthClient(authClient);
+    });
 
     it('exists', () => {
-      expect(Util.getTokens).toBeTruthy();
+      expect(OAuth2Util.getTokens).toBeTruthy();
+    });
+
+    it('emits SDK errors through \'triggerAfterError\' event', function (done) {
+      spyOn(authClient.token, 'getWithPopup').and.callFake(function () {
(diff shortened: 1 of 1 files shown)


aarongranick-okta (approved):

pradeepdewda-okta:
@oleksandrpravosudko-okta @aarongranick-okta Reveling this kinds of errors to enduser would be a head scratcher to them.. I have never seen such types of errors on any login page of any website. We should reveal errors only that makes sense to them.

pradeepdewda-okta (changes requested):
We should not reveal the errors to the enduser that doesn't make any sense to them.. Why would an  enduser should worry about `jwt, iss and aud arguments ?`

aarongranick-okta:
@pradeepdewda-okta raises a good point. We should be restrictive on what is shown in the UI. All errors can be handled programmatically in the afterError handler.

If I remember correctly, this issue began with an error having to do with token verification failing with custom domains, a form of config error. This may not be the type of error we want to show to the developer, not the end-user.

pradeepdewda-okta:
Probably we can use `logger` to show such errors on browser console.

oleksandrpravosudko-okta:
Thanks for the feedback @pradeepdewda-okta, @aarongranick-okta.

It looks like SDK errors are already being [logged](https://github.com/okta/okta-signin-widget/blob/969ac40dd07f47516eb66ae1384fa2aa0cbb8270/src/util/Util.js#L149) during `triggerAfterError` call.

oleksandrpravosudko-okta:
@pradeepdewda-okta looking for another review round on this PR.

pradeepdewda-okta:
@oleksandrpravosudko-okta we need to change the description of PR right ? so you just added the test and we are not bubbling the error to enduser anymore right ?

oleksandrpravosudko-okta:
@pradeepdewda-okta that's correct - no changes to error propagation, only a test verifying SDK errors are passed to `triggerAfterError`

test/unit/spec/OAuth2Util_spec.js
```diff
@@ -1,14 +1,61 @@
...
-import Util from 'util/OAuth2Util';
+import OAuth2Util from 'util/OAuth2Util';
+import Util from 'util/Util';
+import createAuthClient from 'widget/createAuthClient';
+import Settings from '../../../src/models/Settings';
+import { AuthSdkError } from '@okta/okta-auth-js';
 
-// TODO: rewrite
+
+// TODO: add more tests
```
pradeepdewda-okta, line 9: what are those tests ?

pradeepdewda-okta (approved):

test/unit/spec/OAuth2Util_spec.js
oleksandrpravosudko-okta: There should be tests for a happy path(i.e. [success](https://github.com/okta/okta-signin-widget/blob/master/src/util/OAuth2Util.js#L28) function is called) and a test for picking [correct](https://github.com/okta/okta-signin-widget/blob/master/src/util/OAuth2Util.js#L62) token retrieval method according to provided token options.

pradeepdewda-okta: Please add that as comment..

oleksandrpravosudko-okta: added missing tests

PR #1634, opened 2021-02-02
Stop polling in base authenticator view

pradeepdewda-okta:
## Description:
Since we changed idx response status code from 200 to 401 for errors. We are seeing Rate Limit hitting for authenticator views where magic link is involved.

Server responds with 401, but widget don't stop polling and keeps adding a new startPolling timeout that causes a flurry of server calls which increase exponentially.

1. Removed the generic listener on model errors
2. If stateToken is invalid or expired then stop polling
3. continue polling if there is form error but session is still valid

Similar issue: OKTA-365525

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-366306](https://oktainc.atlassian.net/browse/OKTA-366306)
- [OKTA-365525](https://oktainc.atlassian.net/browse/OKTA-365525)


pradeepdewda-okta:
Adding e2e test in next commit.

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -74,6 +71,16 @@ const Body = BaseForm.extend(Object.assign(
...
       this.stopPolling();
+    },
+
+    triggerAfterError (model, error) {
+      BaseForm.prototype.triggerAfterError.apply(this, arguments);
+      this.stopPolling();
+
+      // Polling needs to be resumed if it's a form error and session is still valid
+      if(error.responseJSON.errorMessageKey !== 'idx.session.expired') {
+        this.startPolling();
```
magizh-okta, line 82: Does session expire when we hit ratelimit ?

pradeepdewda-okta: It hits the rate limit after session expires.. before that it works fine as expected.

magizh-okta: nvm

src/v2/ion/IonResponseHelper.js
```diff
@@ -123,9 +123,13 @@ const getGlobalErrors = (res) => {
 const convertFormErrors = (response) => {
   let errors = {
     errorCauses: getRemediationErrors(response),
-    errorSummary: getGlobalErrors(response)
+    errorSummary: getGlobalErrors(response),
   };
 
+  if (response.messages?.value[0]?.i18n.key) {
```
magizh-okta, line 129: this has already been added in 5.3 https://github.com/okta/okta-signin-widget/commit/3c1559f00e8bfbde3e3da9f6f657b57e720832a0#diff-1a8fa5d91b81ac1f2b5a724f95f37a6cf2e065b8d78f82be8dffa5dd8f527260R140

pradeepdewda-okta: yeah i was looking for that change, but didn't knew it was merged to 5.3

aarongranick-okta:
I am not seeing anything in the tests that tests 200 and 401 responses. Is this case being adequately tested? The behavior can be turned on/off with a switch (IDX_RESPONSE_ENHANCEMENTS_PRE_VERSIONING). 

@swiftone is working on something similar for the terminal view: OKTA-361545

pradeepdewda-okta:
> I am not seeing anything in the tests that tests 200 and 401 responses. Is this case being adequately tested? The behavior can be turned on/off with a switch (IDX_RESPONSE_ENHANCEMENTS_PRE_VERSIONING).
> 
> @swiftone is working on something similar for the terminal view: OKTA-361545

I think no one tested the scenarios for session token expiration while polling.. Probably not feasible also to wait for 5 mins in selenium tests.. Not sure how to invalidate the stateToken manually during testing.. I have raised this question earlier to QA but they don't have a answer yet.

aarongranick-okta:
> I think no one tested the scenarios for session token expiration while polling.. 

I am looking for a unit test or testcafe which uses a mock XHR/fetch response. An E2E test would be nice but I don't think we'll have 2 environments to test against. The FF will be either on or off, but we won't know which.

pradeepdewda-okta:
> > I think no one tested the scenarios for session token expiration while polling..
> 
> I am looking for a unit test or testcafe which uses a mock XHR/fetch response. An E2E test would be nice but I don't think we'll have 2 environments to test against. The FF will be either on or off, but we won't know which.

Search for `403` we have mocks for those.. any 40x errors

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -74,6 +71,16 @@ const Body = BaseForm.extend(Object.assign(
...
       BaseForm.prototype.remove.apply(this, arguments);
       this.stopPolling();
+    },
+
+    triggerAfterError (model, error) {
+      BaseForm.prototype.triggerAfterError.apply(this, arguments);
+      this.stopPolling();
+
+      // Polling needs to be resumed if it's a form error and session is still valid
+      if(error.responseJSON.errorMessageKey !== 'idx.session.expired') {
```
haishengwu-okta, line 81: Curious if API can respond 409 (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429)
and we check error code at UI?

pradeepdewda-okta: I don't think making api changes are in the scope... Other simple thing we can do is just stop polling on any error and do not resume polling at all. In case of form error enduser will quickly correct the problem and would proceed, and in case if token has expired enduser will see a `session expired error`.. 

So what i am saying is we may not need to support this `resume poll on error` edge case.. What do you say?

PR #1637, opened 2021-02-03
add Consent flow to v2/OIE

sherizada-okta:
## Description:
Adding Admin Consent and Enduser Consent views to OIE SIW to have feature parity with v1.

## PR Checklist

- [ ] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:
Flow:
Feb-22-2021 16-09-10

Admin Consent:
Screenshot 2021-02-22 at 4 06 00 PM

Enduser Consent:
Screenshot 2021-02-10 at 1 04 12 PM

Without logo, display alt text:
Screenshot 2021-02-22 at 2 02 24 PM
Screenshot 2021-02-22 at 2 02 00 PM

### Issue:

- [OKTA-366545](https://oktainc.atlassian.net/browse/OKTA-366545)
- [OKTA-366548](https://oktainc.atlassian.net/browse/OKTA-366548)


haishengwu-okta (commented):
TestCafe tests?

src/v2/view-builder/views/AdminConsent.js
```diff
@@ -0,0 +1,67 @@
...
+
+    changeScopes() {
+      const scopes = this.getUISchema().map(({value}) => {
+        return value;
+      });
+      this.model.set('scopes', scopes);
+    },
+
+    postRender() {
+      this.$el.find('.o-form-content').remove();
```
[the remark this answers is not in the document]
sherizada-okta: @haishengwu-okta 
Screenshot 2021-02-03 at 2 19 07 PM

sherizada-okta:
I am going to reuse the view for Admin Consent and End User since they mostly overlap. Commit or separate PR to come...

sherizada-okta:
@pradeepdewda-okta @haishengwu-okta this is ready for a review

src/v2/view-builder/views/AdminConsent.js
```diff
@@ -0,0 +1,82 @@
...
+
+    postRender () {
+      this.$el.find('.o-form-head').remove();
+      this.$el.find('.o-form-fieldset-container').remove();
+    }
+  },
+);
+
+export default BaseView.extend({
+  className: 'admin-consent-required',
```
haishengwu-okta, line 73: This is probably need to be done `ViewClassNamesFactory.js`

src/v2/view-builder/views/AdminConsent.js
```diff
@@ -0,0 +1,82 @@
...
+      this.add(ScopeList, {options: { model }});
+    },
+    /**
+     * Format scopes to match the schema required by the server
+     * POST /idp/idx/consent
+        {
+          "scopes": [ {{value}} ]
+        }
+     */
+    saveForm () {
```
haishengwu-okta, line 56: - Did you double check the request body to see it didn't send extra data?
- Looks like we shall override `BaseView.createModelClass`
- In additions, feel like we shall update `ion-object-handler#createUiSchemaForObject` to create a customized view for `scopes` field. Similar to how we handle `select-authenticator-authenticate`. @pradeepdewda-okta @nikhilvenkatraman-okta any thoughts?

src/v2/view-builder/views/AdminConsent.js
```diff
@@ -0,0 +1,82 @@
+import { loc, View, Model } from 'okta';
+import BaseView from '../internals/BaseView';
+import BaseForm from '../internals/BaseForm';
+import ScopeList from '../../../views/admin-consent/ScopeList';
+import consentLogoHeaderTemplate from '../../../views/shared/templates/consentLogoHeaderTemplate';
```
pradeepdewda-okta, line 5: Interesting. But we don't want to share code with v1. Right ? this resides outside of V2 folder. Mostly For such scenarios we just create another view. This template won't be reusable apart from consent views.

src/v2/view-builder/views/AdminConsent.js
pradeepdewda-okta: @haishengwu-okta 
yeah i would discourage to override saveForm. `scopes` is a new attribute for v2, so yes that feels right place to add this.

pradeepdewda-okta: But i also think that could be overkill to do this? It's a specific use case and not reusable..

swiftone: Note: `scopes` has always been part of `authParams`, and that it now exists OUTSIDE of authParams is likely behavior we don't want to keep.  We have a bug about this issue.

haishengwu-okta: `scopes` is just field name in a remediation form (which will be show up in request body).
why it's being conflict/confused with `authParams` (OIDC `scope`) ?

src/v2/view-builder/views/AdminConsent.js
sherizada-okta: Why not share view code? They're using the same UI framework. But I can also copy and move them to the v2 folder.

src/v2/view-builder/views/AdminConsent.js
sherizada-okta: ^ @swiftone, the scopes here are for a form that will be submitted to /idx/consent (if the user accepts these scopes) and then those scopes will be incorporated in the state token. I believe the authParams would be referring to the accepted scopes in the token?

sherizada-okta: I am sending the correct data on saveForm (you can see PR giphy for what that is), but another way is to use an event listener instead like this https://github.com/okta/okta-signin-widget/pull/1637/commits/f83396c5cf57506e8cdff79e717997c71a7d8e9f#diff-8a1abd3d64283eaee376fa7479ddf4070fbfaa7c93b59fe3164d2be7521a6394R34

src/v2/view-builder/views/AdminConsent.js
pradeepdewda-okta: Even with slightest difference most dev prefer to create their own templates, so it's generally not reusable most of the time.. Instead just create a plain view that returns you this screen in V2.

src/v2/view-builder/views/AdminConsent.js
```diff
@@ -0,0 +1,82 @@
...
+
+
+const Header = View.extend({
```
pradeepdewda-okta, line 8: put this in a new file `ConsentViewHeader.js`

src/v2/view-builder/views/AdminConsent.js
pradeepdewda-okta: Nice see, if you can trigger a event.. Also as mentioned by @haishengwu-okta super method should always be called first and then override statements .. ```BaseForm.prototype.saveForm.apply(this, arguments);```

playground/mocks/data/idp/idx/admin-consent.json
```diff
@@ -16,35 +16,34 @@
...
-                  "desc": "Allows the app to manage clients in your Okta organization."
-                },
-                {
-                  "name": "okta.authenticators.manage",
-                  "label": "okta.authenticators.manage",
-                  "value": "okta.authenticators.manage",
-                  "desc": "Allows the app to manage all authenticators (e.g. enrollments, reset)."
-                }
-              ]
+        {
```
sherizada-okta, line 19: I sent this schema update request to Shantanu

src/v2/view-builder/views/AdminConsent.js
sherizada-okta: https://github.com/okta/okta-signin-widget/pull/1637/commits/5c38fe6dbc14fc42c30a7aededad0f7c937ff8d3

playground/mocks/data/idp/idx/admin-consent.json
```diff
@@ -0,0 +1,99 @@
...
+        "name": "admin-consent",
+        "href": "http://localhost:3000/idp/idx/consent",
+        "method": "POST",
+        "accepts": "application/ion+json; okta-version=1.0.0",
+        "produces": "application/ion+json; okta-version=1.0.0",
+        "relatesTo": ["$.termsOfService", "$.privacyPolicy", "$.logo"],
+        "value": [
+        {
+          "name": "scopes",
+          "type": "object",
```
haishengwu-okta, line 21: I'm guessing it's probably making more sense as `array` as we send back list of scopes in request body (hence need to have some code changes in transformer layer)

Let's see what Shantanu get back to us.

src/v2/ion/ui-schema/ion-object-handler.js
```diff
@@ -120,6 +120,12 @@ const createUiSchemaForObject = (ionFormField, remediationForm, transformedResp,
     // 2. when `select-authenticator-enroll`,
     // create customize type so that display authenticator list as selectable list
     Object.assign(uiSchema, getAuthenticatorsEnrollUiSchema(ionFormField));
+  } else if (ionFormField.name === 'scopes')  {
+    const options = ionFormField.options.map(({label, desc}) => {
+      return {name: label, displayName: label, description: desc};
+    });
+    // setting 'type' here to add a specific View in FormInputFactory.create
+    Object.assign(uiSchema, {options, type: 'scopes'});
```
haishengwu-okta, line 128: - Be more explicit like `oauth2-admin-scopes` ?
- wonder if the remediation form name could be more explicit like `admin-scopes` ? (question to api team)

src/v2/view-builder/views/consent/ConsentView.js
```diff
@@ -0,0 +1,39 @@
...
+    noButtonBar: false,
+    noCancelButton: false,
+    save: () => loc('consent.required.consentButton', 'login'),
+    cancel: () => loc('consent.required.cancelButton', 'login'),
+    title: false,
+  },
+);
+
+export default BaseView.extend({
+  className: 'admin-consent-required',
```
haishengwu-okta, line 17: can we leverage `ViewClassNamesFactory` ?

src/v2/view-builder/views/AdminConsent.js
sherizada-okta: What's the reasoning behind separating the className from the View?

src/v2/view-builder/views/consent/ConsentView.js
sherizada-okta: What's the reasoning behind separating the className from the View?

playground/mocks/data/idp/idx/admin-consent.json
sherizada-okta: I asked Shantanu to change it to this, so that's what we'll get back

src/v2/ion/ui-schema/ion-object-handler.js
sherizada-okta: The remediation form name `admin-consent` seems more appropriate since the view/flow is referred to as Consent, where scopes is a property of Consent.
I can change the `type`.

playground/mocks/data/idp/idx/admin-consent.json
sherizada-okta: It looks like other instances of `options` either have the type `object` or no type, so `type: array` would be a new inconsistency.

haishengwu-okta: iirc, the `type` dictates request body. default to `string` when it's missing.

sherizada-okta: After reading about the options field, it looks like they're usually paired with type array: https://ionspec.org/#form-field-option-members
There'll be a design review of this response so I'll bring this up then

src/v2/ion/ui-schema/ion-object-handler.js
sherizada-okta: I changed type to `adminConsentScopes`

sherizada-okta:
I added Enduser Consent: https://github.com/okta/okta-signin-widget/pull/1637/commits/ee2433e6f3fe8f4b0bed43bfe03731ed38f82041
It re-uses Consent View but has some additional subviews

src/v2/view-builder/views/consent/ConsentViewHeader.js
```diff
@@ -0,0 +1,31 @@
+import { View, _ } from 'okta';
+
+import consentLogoHeaderTemplate from '../../../../views/shared/templates/consentLogoHeaderTemplate';
+
+const defaultLogo = '/img/logos/default.png';
+
+const ConsentViewHeader = View.extend({
+  className: 'consent-title detail-row',
+  template: consentLogoHeaderTemplate,
```
pradeepdewda-okta, line 9: i would put the template here instead.

pradeepdewda-okta (approved):
This looks very neat.. !! 🥇 

one nit: remove the template file and add the template in the view itself.

src/v2/view-builder/views/consent/ConsentViewHeader.js
```diff
@@ -2,20 +2,21 @@ import { View, _ } from 'okta';
...
+const defaultLogo = '/img/logos/default.png';
 
 const ConsentViewHeader = View.extend({
   className: 'consent-title detail-row',
   template: consentLogoHeaderTemplate,
   getTemplateData: function () {
     const { appState } = this.options;
-    const { label, clientUri } =  appState.get('app');
+    const { currentFormName } = appState.get('currentFormName');
+    const { label, clientUri, logo } =  appState.get('app');
```
sherizada-okta, line 13: @haishengwu-okta @pradeepdewda-okta I see these values can be set in the widgetrc, how does that work with appState in v2? I don't see these custom values overriding the AppState model. Is that something that's supposed to be a part of the SIW Customization project?

src/v2/view-builder/views/consent/ConsentViewHeader.js
sherizada-okta: https://github.com/okta/okta-signin-widget/pull/1637/commits/ce2298ed42df1b22be3dca6cfb4e897b19aefe63

playground/mocks/data/idp/idx/admin-consent.json
sherizada-okta: It'll be changed to `array` so I'll handle that new type in uiSchemaTransformer in an upcoming commit

sherizada-okta:
@pradeepdewda-okta thanks but there will be more changes in this PR from the API response updates, and I noticed a visual discrepancy with the Enduser buttons so I added https://github.com/okta/okta-signin-widget/pull/1637/commits/779af66e62e4ef08f064f45cc59562f5dce76baf

src/v2/view-builder/views/consent/ConsentViewHeader.js
haishengwu-okta: > I see these values can be set in the widgetrc

Could you give example how those can be customized via widgetrc for v1?
I'm not aware.
If so, need to create ticket to the customization epic.

assets/sass/modules/_consent.scss
```diff
@@ -90,5 +92,15 @@ $consent-light-text-color: #6e6e78; // Gray 600 in Odyssey
     input[data-type="save"] {
       float: right;
     }
+
+    .button-primary {
```
haishengwu-okta, line 96: why not just remove the `button-primary` class for the `Allow` button?

src/v2/view-builder/views/consent/ConsentViewHeader.js
```diff
@@ -0,0 +1,55 @@
...
+      <img class="client-logo custom-logo" src="{{customLogo}}" alt="aria logo" aria-hidden="true" />
+    {{else}}
+      <img class="client-logo default-logo" src="{{defaultLogo}}" alt="aria logo" aria-hidden="true" />
+    {{/if}}
+    {{#if clientURI}}
+      </a>
+    {{/if}}
+    <h1>
+      <span class="title-text">
+        {{#if isAdminConsent}}
```
haishengwu-okta, line 22: hmm.. why such difference? did you check with UX to see if we can use same text?

sherizada-okta: I am changing it to the text used in the mock which didn't carryover in v1 development: https://oktawiki.atlassian.net/wiki/spaces/PM/pages/1074627985/Spec+End+to+End+Identity+Attestation#User-Flow-Diagram

We shouldn't re-use the text for Enduser because it's grammatically incorrect.
For example, if the text is not updated it would read:
```
App wants to access
- View your phone number
```
vs 
```
App wants to
- View your phone number
```
For admin it works because 
```
App wants to access
- Resource and Policies
```

assets/sass/modules/_consent.scss
sherizada-okta: That is another way, but I believe CSS overrides should be in one place so it's easier to maintain.

src/v2/view-builder/views/consent/ConsentViewHeader.js
sherizada-okta: You can set the `logo` in widgetrc to whatever path you want. That is different than the value in `appState.logo` which is coming from the server.

src/v2/view-builder/views/consent/ConsentViewHeader.js
sherizada-okta: I'll confirm with a designer though.

sherizada-okta:
I've implemented all the prospective changes for the API updates. So this is ready for a final review and then I believe I can ship because no one can hit this consent path in production.

src/v2/view-builder/views/consent/ConsentViewHeader.js
sherizada-okta: Confirmed with @ZoeMonosson-okta that the header texts should be different

sherizada-okta:
The new Consent API remediation responses have gone through an API design review and we have come to an agreement.

Instead of sending scopes to the server, going to send {consent: true|false}.
- Scenario 1: User sees Consent view and clicks "Don't Allow" -> POST /consent { consent: false }
- Scenario 2: User sees Consent view and clicks "Allow Access" -> POST /consent { consent: true }

@pradeepdewda-okta @haishengwu-okta https://github.com/okta/okta-signin-widget/pull/1637/commits/8dc2de226528ec5ce95ef0959b7fc4a94b7906c4

src/v2/ion/RemediationConstants.js
```diff
@@ -49,7 +55,7 @@ const FORMS = {
   // for rendering a page that user has nothing to remediate.
   TERMINAL: 'terminal',
 };
-
+// Add your FORM in FORMS to FORMS_WITHOUT_SIGNOUT if you don't want to display a 'Sign out' link
```
pradeepdewda-okta, line 58: change it to -- `Forms added here do not show Sign Out link`

test/testcafe/spec/ConsentView_spec.js
```diff
@@ -0,0 +1,120 @@
...
+  const jsonBody = JSON.parse(body);
+
+  await t.expect(jsonBody.consent).eql(true);
+  await t.expect(method).eql('post');
+  await t.expect(url).eql('http://localhost:3000/idp/idx/consent');
+
+  await testRedirect(t);
+});
+
+test.requestHooks(requestLogger, consentAdminMock)('should call /consent and send {consent: false} on form cancel', async t => {
```
pradeepdewda-okta, line 69: cancel mean `Don't Allow` here right ? let's allow and not allowed in test description instead of submit/cancel

src/v2/view-builder/views/consent/ConsentAgreementText.js
```diff
@@ -0,0 +1,12 @@
+import { View } from 'okta';
+import hbs from 'handlebars-inline-precompile';
+
+
+export default View.extend({
+  className: 'consent-description detail-row',
```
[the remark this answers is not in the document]
sherizada-okta: grabs the v1 css styles

test/testcafe/spec/ConsentView_spec.js
sherizada-okta: right

src/v2/view-builder/views/consent/ConsentView.js
```diff
@@ -0,0 +1,63 @@
...
+    title: false,
+    events: {
+      'click input[data-type="save"]': 'setConsent',
+    },
+    setConsent () {
+      this.model.set('consent', true);
+    },
+    cancelForm () {
+      // override BaseForm.prototype.cancelForm which cancels auth flow
+      this.model.set('consent', false);
```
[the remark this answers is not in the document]
sherizada-okta: which ones' the handler?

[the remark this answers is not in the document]
pradeepdewda-okta: `setConsent`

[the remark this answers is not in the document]
sherizada-okta: Sureee, I'd need to rewrite it to this:
```
events: {
      'click input[data-type="save"]': function setConsentTrue() {
        this.setConsent(true);
      },
    },
    setConsent (bool) {
      this.model.set('consent', bool);
    },
    cancelForm () {
      this.setConsent(false);
```
Is that preferable?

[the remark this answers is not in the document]
sherizada-okta: `setConsent` in its current state should really be named `setConsentTrue`

src/v2/view-builder/internals/FormInputFactory.js
```diff
@@ -48,9 +52,29 @@ const createAuthenticatorVerifySelectView = (opt) => {
...
+      }
+    })
+  };
+};
+
 const inputCreationStrategy = {
   authenticatorEnrollSelect: createAuthenticatorEnrollSelectView,
-  authenticatorVerifySelect: createAuthenticatorVerifySelectView
+  authenticatorVerifySelect: createAuthenticatorVerifySelectView,
+  [RemediationForms.CONSENT_ADMIN]: createAdminScopesView,
```
haishengwu-okta, line 76: nit:
I'd rather use different naming (InputTypeConstant.consent_admin or just hard-code `consent-admin`) even thought the value is same to avoid confusing.

src/v2/view-builder/views/consent/ConsentAgreementText.js
```diff
@@ -0,0 +1,12 @@
+import { View } from 'okta';
```
haishengwu-okta, line 1: isn'n worth to be separated file?

src/v2/view-builder/views/consent/ConsentView.js
```diff
@@ -0,0 +1,63 @@
...
+  },
+);
+
+export default BaseView.extend({
+  Header: ConsentViewHeader,
+  Body,
+  initialize () {
+    BaseView.prototype.initialize.apply(this, arguments);
+
+    if (this.options.appState.get('currentFormName') === 'consent') {
```
haishengwu-okta, line 36: - `=== RemediationForm.CONSENT_ENDUSER` ?
- I'd rather to have class hierarchy as we already be able to dispatch to different class at View Factory.

```
AdminConsentView.js {
  ...
}
EnduserConsentView extends AdminContentView {
  Footer: EnduserConsentViewFooter,
}
```

src/v2/view-builder/views/consent/ConsentViewHeader.js
haishengwu-okta: `widgetrc.logo` is meant for identify view only (primary auth, passwordless auth, idp discovery  in v1)

src/v2/view-builder/views/consent/ConsentViewHeader.js
haishengwu-okta: It's not ideal to have diverged logic here.
Ideally, View Factory is only place doing dispatching (find the right view for particular remediation form)
If those views share some logic, extra a base view and extends from it.

haishengwu-okta (commented):
- Just comments on naming and class hierarchy convention.

src/v2/view-builder/views/consent/ConsentAgreementText.js
sherizada-okta: We should be striving for 1:1 view to file because it makes code easier to read.

src/v2/view-builder/views/consent/ConsentView.js
sherizada-okta: I understand and have added this change https://github.com/okta/okta-signin-widget/pull/1637/commits/d6e4a3e83458efa8e13a89b0b846cad2f755d51f

sherizada-okta:
Last 2 commits haven't been reviewed yet @pradeepdewda-okta @haishengwu-okta

haishengwu-okta (approved):

pradeepdewda-okta (approved):
Perfect 👍

PR #1638, opened 2021-02-03
feat: Adds IDP factor enrollment to OIE

jmelberg-okta:
## Description:

- Adds IDP factor enrollment and verification on the OIE pipeline
- Adds `/oauth2/v1/authorize/callback` mock logic for playground

### Screenshot/Video:

| | Video | Description |
| ---- | ------ | ---------- |
| **Enrollment** | Enrollment | Basic enrollment flow given a variety of factors. <br><br> In practice, when a user selects **Enroll** the widget performs a full-page redirect to the IDP login page. Upon successful authentication, the IDP will redirect back to the login page via the `oauth2/v1/authorize/callback` URI scheme.
| **Verification** | Verification | Factor challenge/verification flow given a variety of factors. <br><br> When a user selects **Verify**, the widget performs a full-page redirect to the IDP login page, similar to the enrollment flow.  |

### Issue:

- [OKTA-364505](https://oktainc.atlassian.net/browse/OKTA-364505)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -263,6 +263,7 @@ mfa.challenge.enterCode.placeholder = Enter Code
 mfa.challenge.enterCode.tooltip = Enter Code
 mfa.challenge.password.placeholder = Password
 mfa.backToFactors = Back to factor list
+mfa.enroll = Enroll
 mfa.phoneNumber.placeholder = Phone number
 mfa.phoneNumber.ext.placeholder = Extension
 mfa.sendCode = Send code
@@ -840,6 +841,14 @@ oie.duo.authenticator.description = Verify your identity using Duo Security.
 oie.duo.enroll.title = Set up Duo Security
 oie.duo.verify.title = Verify with Duo Security
 oie.duo.iFrameError = Error loading Duo. Try again or contact your admin for assistance.
+
+## IdP Authenticator
+# {0} refers to the display name of the authenticator for all IdP Authenticator values
+oie.idp.authenticator.description = Redirect to verify with {0}
+oie.idp.enroll.title = Set up {0}
+oie.idp.enroll.description = Clicking below will redirect to enrollment in {0}
+oie.idp.challenge.title=  Verify with {0}
+oie.idp.challenge.description = You will be redirected to verify with {0}
 
 ## Select authenticator enrollment
 oie.select.authenticators.enroll.title = Set up Authenticators
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -49,7 +49,7 @@ const idx = {
     // 'identify-with-third-party-idps',
     // 'identify-with-only-third-party-idps',
     // 'identify-with-only-one-third-party-idp',
-    //'identify-with-password',
+    // 'identify-with-password',
     // 'identify-with-universal-link',
     // 'success',
     // 'success-with-app-user',
@@ -74,7 +74,7 @@ const idx = {
   ],
   '/idp/idx/identify': [
     'authenticator-enroll-select-authenticator',
-    //'identify-with-only-one-third-party-idp',
+    // 'identify-with-only-one-third-party-idp',
     // 'error-identify-access-denied',
     // 'error-identify-user-locked-unable-challenge'
   ],
@@ -85,7 +85,7 @@ const idx = {
     'error-authenticator-enroll-security-question'
   ],
   '/idp/idx/challenge/send': [
-    //'authenticator-enroll-ov-sms',
+    // 'authenticator-enroll-ov-sms',
     'authenticator-enroll-ov-email',
   ],
   '/idp/idx/challenge/resend': [
@@ -94,7 +94,7 @@ const idx = {
   ],
   '/idp/idx/challenge/poll': [
     'success',
-    //'enroll-profile-new'
+    // 'enroll-profile-new'
     // 'authenticator-enroll-email',
     // 'authenticator-verification-okta-verify-push',
   ],
@@ -589,6 +589,20 @@ const passwordRecovery = {
     // 'error-identify-access-denied',
     'authenticator-verification-select-authenticator',
   ],
+};
+
+const idpAuthenticator = {
+  '/idp/idx/introspect': [
+    'authenticator-enroll-select-authenticator',
+    // 'authenticator-verification-select-authenticator',
+    'success',
+  ],
+  '/idp/idx/challenge': [
+    'authenticator-verification-idp',
+  ],
+  '/idp/idx/credential/enroll': [
+    'authenticator-enroll-idp',
+  ]
 };
 
 module.exports = {
--- a/playground/mocks/data/idp/idx/authenticator-enroll-idp.json
+++ b/playground/mocks/data/idp/idx/authenticator-enroll-idp.json
@@ -65,7 +65,6 @@
       "type": "federated",
       "id": "aut4mhtS1b84AR0iQ0g4",
       "displayName": "IDP Authenticator",
-      "contextualData": { "integrationType": "REDIRECT" },
       "methods": [
         { "type": "idp" }
       ]
--- a/playground/mocks/data/idp/idx/authenticator-enroll-idp.json
+++ b/playground/mocks/data/idp/idx/authenticator-enroll-idp.json
@@ -1,0 +1,132 @@
+{
+  "stateHandle": "02TptqPN4BOLIwMAGUVLPlZVJEnONAq7xkg19dy6Gk",
+  "version": "1.0.0",
+  "expiresAt": "2021-01-19T15:10:35.000Z",
+  "intent": "LOGIN",
+  "remediation": {
+    "type": "array",
+    "value": [
+      {
+        "name": "redirect-idp",
+        "type": "OIDC",
+        "idp": {
+          "id": "0oa69chx4bZyx8O7l0g4",
+          "name": "IDP Authenticator"
+        },
+        "href": "http://localhost:3000/sso/idps/0oa69chx4bZyx8O7l0g4?stateToken=02TptqPN4BOLIwMAGUVLPlZVJEnONAq7xkg19dy6Gk",
+        "method": "GET",
+        "relatesTo" : [ "$.currentAuthenticator" ]
+      },
+      {
+        "rel": [ "create-form" ],
+        "name": "select-authenticator-enroll",
+        "href": "http://localhost:3000/idp/idx/credential/enroll",
+        "method": "POST",
+        "accepts": "application/json; okta-version=1.0.0",
+        "produces": "application/ion+json; okta-version=1.0.0",
+        "value": [
+          {
+            "name": "authenticator",
+            "type": "object",
+            "options": [
+              {
+                "label": "IDP Authenticator",
+                "value": {
+                  "form": {
+                    "value": [
+                      {
+                        "name": "id",
+                        "required": true,
+                        "value": "aut4mhtS1b84AR0iQ0g4",
+                        "mutable": false
+                      }
+                    ]
+                  }
+                },
+                "relatesTo": "$.authenticators.value[0]"
+              }
+            ]
+          },
+          {
+            "name": "stateHandle",
+            "required": true,
+            "value": "02TptqPN4BOLIwMAGUVLPlZVJEnONAq7xkg19dy6Gk",
+            "visible": false,
+            "mutable": false
+          }
+        ]
+      }
+    ]
+  },
+  "currentAuthenticator": {
+    "type": "object",
+    "value": {
+      "key": "external_idp",
+      "type": "federated",
+      "id": "aut4mhtS1b84AR0iQ0g4",
+      "displayName": "IDP Authenticator",
+      "contextualData": { "integrationType": "REDIRECT" },
+      "methods": [
+        { "type": "idp" }
+      ]
+    }
+  },
+  "authenticators": {
+    "type": "array",
+    "value": [
+      {
+        "key": "external_idp",
+        "type": "federated",
+        "id": "aut4mhtS1b84AR0iQ0g4",
+        "displayName": "IDP Authenticator",
+        "methods": [
+          { "type": "idp" }
+        ]
+      }
+    ]
+  },
+  "authenticatorEnrollments": {
+    "type": "array",
+    "value": []
+  },
+  "enrollmentAuthenticator": {
+    "type": "object",
+    "value": {
+      "key": "external_idp",
+      "type": "federated",
+      "id": "aut4mhtS1b84AR0iQ0g4",
+      "displayName": "IDP Authenticator",
+      "methods": [
+        { "type": "idp" }
+      ]
+    }
+  },
+  "user": {
+    "type": "object",
+    "value": { "id": "00u2m55pu8UZyeMMl0g4" }
+  },
+  "cancel": {
+    "rel": [ "create-form" ],
+    "name": "cancel",
+    "href": "http://localhost:3000/idp/idx/cancel",
+    "method": "POST",
+    "value": [
+      {
+        "name": "stateHandle",
+        "required": true,
+        "value": "02TptqPN4BOLIwMAGUVLPlZVJEnONAq7xkg19dy6Gk",
+        "visible": false,
+        "mutable": false
+      }
+    ],
+    "accepts": "application/ion+json; okta-version=1.0.0"
+  },
+  "app": {
+    "type": "object",
+    "value": {
+      "name": "oidc_client",
+      "label": "Test OIDC App",
+      "id": "0oa11ch8m94eTn0Qe0g4"
+    }
+  }
+}
--- a/playground/mocks/data/idp/idx/authenticator-verification-idp.json
+++ b/playground/mocks/data/idp/idx/authenticator-verification-idp.json
@@ -65,7 +65,6 @@
       "type": "federated",
       "id": "aut4mhtS1b84AR0iQ0g4",
       "displayName": "IDP Authenticator",
-      "contextualData": { "integrationType": "REDIRECT" },
       "methods": [
         { "type": "idp" }
       ]
@@ -78,7 +77,6 @@
       "type": "federated",
       "id": "aut4mhtS1b84AR0iQ0g4",
       "displayName": "IDP Authenticator",
-      "contextualData": { "integrationType": "REDIRECT" },
       "methods": [
         { "type": "idp" }
       ]
--- a/playground/mocks/data/idp/idx/authenticator-verification-idp.json
+++ b/playground/mocks/data/idp/idx/authenticator-verification-idp.json
@@ -1,0 +1,144 @@
+{
+  "stateHandle": "02TptqPN4BOLIwMAGUVLPlZVJEnONAq7xkg19dy6Gk",
+  "version": "1.0.0",
+  "expiresAt": "2021-01-19T15:10:35.000Z",
+  "intent": "LOGIN",
+  "remediation": {
+    "type": "array",
+    "value": [
+      {
+        "name": "redirect-idp",
+        "type": "OIDC",
+        "idp": {
+          "id": "0oa69chx4bZyx8O7l0g4",
+          "name": "IDP Authenticator"
+        },
+        "href": "http://localhost:3000/sso/idps/0oa69chx4bZyx8O7l0g4?stateToken=02TptqPN4BOLIwMAGUVLPlZVJEnONAq7xkg19dy6Gk",
+        "method": "GET",
+        "relatesTo" : [ "$.currentAuthenticatorEnrollment" ]
+      },
+      {
+        "rel": [ "create-form" ],
+        "name": "select-authenticator-authenticate",
+        "href": "http://localhost:3000/idp/idx/challenge",
+        "method": "POST",
+        "accepts": "application/json; okta-version=1.0.0",
+        "produces": "application/ion+json; okta-version=1.0.0",
+        "value": [
+          {
+            "name": "authenticator",
+            "type": "object",
+            "options": [
+              {
+                "label": "IDP Authenticator",
+                "value": {
+                  "form": {
+                    "value": [
+                      {
+                        "name": "id",
+                        "required": true,
+                        "value": "aut4mhtS1b84AR0iQ0g4",
+                        "mutable": false
+                      }
+                    ]
+                  }
+                },
+                "relatesTo": "$.authenticatorEnrollments.value[0]"
+              }
+            ]
+          },
+          {
+            "name": "stateHandle",
+            "required": true,
+            "value": "02TptqPN4BOLIwMAGUVLPlZVJEnONAq7xkg19dy6Gk",
+            "visible": false,
+            "mutable": false
+          }
+        ]
+      }
+    ]
+  },
+  "currentAuthenticatorEnrollment": {
+    "type": "object",
+    "value": {
+      "key": "external_idp",
+      "type": "federated",
+      "id": "aut4mhtS1b84AR0iQ0g4",
+      "displayName": "IDP Authenticator",
+      "contextualData": { "integrationType": "REDIRECT" },
+      "methods": [
+        { "type": "idp" }
+      ]
+    }
+  },
+  "currentAuthenticator": {
+    "type": "object",
+    "value": {
+      "key": "external_idp",
+      "type": "federated",
+      "id": "aut4mhtS1b84AR0iQ0g4",
+      "displayName": "IDP Authenticator",
+      "contextualData": { "integrationType": "REDIRECT" },
+      "methods": [
+        { "type": "idp" }
+      ]
+    }
+  },
+  "authenticators": {
+    "type": "array",
+    "value": [
+      {
+        "key": "external_idp",
+        "type": "federated",
+        "id": "aut4mhtS1b84AR0iQ0g4",
+        "displayName": "IDP Authenticator",
+        "methods": [
+          { "type": "idp" }
+        ]
+      }
+    ]
+  },
+  "authenticatorEnrollments": {
+    "type": "array",
+    "value": [
+      {
+        "profile": { "provider": "Custom OIDC Provider" },
+        "type": "federated",
+        "key": "external_idp",
+        "id": "aut4mhtS1b84AR0iQ0g4",
+        "displayName": "IDP Authenticator",
+        "methods": [
+          { "type": "idp" }
+        ]
+      }
+    ]
+  },
+  "user": {
+    "type": "object",
+    "value": { "id": "00u2m55pu8UZyeMMl0g4" }
+  },
+  "cancel": {
+    "rel": [ "create-form" ],
+    "name": "cancel",
+    "href": "http://localhost:3000/idp/idx/cancel",
+    "method": "POST",
+    "value": [
+      {
+        "name": "stateHandle",
+        "required": true,
+        "value": "02TptqPN4BOLIwMAGUVLPlZVJEnONAq7xkg19dy6Gk",
+        "visible": false,
+        "mutable": false
+      }
+    ],
+    "accepts": "application/ion+json; okta-version=1.0.0"
+  },
+  "app": {
+    "type": "object",
+    "value": {
+      "name": "oidc_client",
+      "label": "Test OIDC App",
+      "id": "0oa11ch8m94eTn0Qe0g4"
+    }
+  }
+}
--- a/playground/mocks/spec-okta-api/oauth2/v1/authorize/callback.js
+++ b/playground/mocks/spec-okta-api/oauth2/v1/authorize/callback.js
@@ -1,0 +1,22 @@
+const apis = [
+  {
+    delay: [1000, 3000],
+    proxy: false,
+    path: '/oauth2/v1/authorize/callback',
+    method: 'GET',
+    template () {
+      // In a real case, `/oauth2/v1/authorize` would parse the `code`
+      // and `state` and handle it as any OAuth 2.0 request.
+      // To simplify the test flow, respond via a redirect with a new `stateToken`.
+      return `
+      <html>
+        <head>
+          <meta http-equiv="refresh" content="2;url=http://localhost:3000?stateToken=mockedStateToken123" />
+        </head>
+        <body></body>
+      </html>`;
+    },
+  },
+];
+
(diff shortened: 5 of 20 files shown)


src/v2/controllers/FormController.js
```diff
@@ -147,7 +149,19 @@ export default Controller.extend({
...
 
     this.toggleFormButtonState(true);
+
+    // NOTE: okta-idx-js does not know whether remediations
+    // are required to redirect or perform a GET request.
+    // As a result, we handle this accordingly by checking the formName
+    const fullPageRedirectForms = [ FORMS.REDIRECT_IDP ];
+    if (fullPageRedirectForms.includes(formName)) {
+      const currentViewState = this.options.appState.getCurrentViewState();
+      Util.redirectWithFormGet(currentViewState.href);
```
pradeepdewda-okta, line 159: why not only do this in the view that requires it ? instead of here ? for eg `SuccessView.js` does that

jmelberg-okta: Main reason is to centralize the redirect logic to the IDX handler, as opposed to having multiple implementations for each View/Controller. Keeping this logic at the `FormController` layer ensures any remediation response that contains `redirect-idp` will redirect when a user interacts with the **save** button.

pradeepdewda-okta: What are the other use cases that sends `redirect-idp` ?

pradeepdewda-okta: better to create a separate handler or function? @nikhilvenkatraman-okta mentioned it as TODO https://oktainc.atlassian.net/browse/OKTA-250473

jmelberg-okta: I'm pushing back on adding extra complexity for developers - as we're essentially forking here when processing the IDX response. Originally, I went down the path of adding a new handler, but quickly found that this requires:

- Additional model events need to be subscribed to and handled
- Disable the default `saveForm` function in favor of the new behavior

@haishengwu-okta - Would appreciate any input you have here!

src/v2/controllers/FormController.js
```diff
@@ -147,7 +149,19 @@ export default Controller.extend({
     }
 
     this.toggleFormButtonState(true);
+
+    // NOTE: okta-idx-js does not know whether remediations
+    // are required to redirect or perform a GET request.
+    // As a result, we handle this accordingly by checking the formName
+    const fullPageRedirectForms = [ FORMS.REDIRECT_IDP ];
```
pradeepdewda-okta, line 156: we are opening a pandora box here, where any form can be added into this array that can redirect.. We need some kind of restriction for developers not to make that mistake.

src/v2/controllers/FormController.js
```diff
@@ -147,7 +149,19 @@ export default Controller.extend({
...
+    if (fullPageRedirectForms.includes(formName)) {
+      const currentViewState = this.options.appState.getCurrentViewState();
+      Util.redirectWithFormGet(currentViewState.href);
+      return;
```
pradeepdewda-okta, line 160: also i don't like the idea of breaking the remediation process here, specially if the user actions ended up successful

src/v2/controllers/FormController.js
```diff
@@ -105,7 +107,7 @@ export default Controller.extend({
   },
 
   handleSwitchForm (formName) {
-    // trigger formname change to change view
+    // trigger formName change to change view
```
haishengwu-okta, line 110: 👍

src/v2/controllers/FormController.js
haishengwu-okta: - For your case, maybe we can follow same pattern as other IDP buttons added to identify view? `FormInputFactory.createIdpButtons`?
- But, a generic redirect handler for ion form is still needed. There are a few cases as Pradeep point out (hence the JIRA). Don't have a good idea now. We can have a group discussion. What Jordan have done seems good.

haishengwu-okta: Another idea:
- Looks like for those redirect case, the remediation form doesn't have `rel: [ 'create-form' ]`
- Maybe confirm with IDX and Device team (has redirect flow for OktaVerify as well) to see if it is always like this.
- IF so, we can change idx-js
- Otherwise,  🤔

jmelberg-okta: Thanks for the feedback! From what I can see, we'll need to handle redirects in a couple of use cases currently:

1. Social login - where the user needs to interact with a custom IdP (Facebook, Google, etc.) as their primary authentication method.
2. When we're enrolling a factor (includes `select-authenticator-enroll` form) with a remediation of type `redirect-idp` (Custom claims, SSO Extension).
3. We have factor enrollment and verification where we encourage the user to enroll through a native application (similar to OV) using a universal link (opens up a native app on desktop/mobile devices). Normal `GET` requests unfortunately will not open up these applications, only full-page redirects.

While the `redirect-idp` type is closely related for both Social and custom factor enrollment, the login flows do not require an "enroll authenticator" page like the factors do. As a result, we need a generic solution to handle this type when we encounter it.

@swiftone and @haishengwu-okta - We _could_ make this change in `idx-js`, but I think we'd need a formal discussion around whether or not this is the best idea to have the library follow redirect requests. Specifically, this change will require us to modify our existing [fetch](https://github.com/okta/okta-idx-js/blob/master/src/v1/generateIdxAction.js#L19-L47) functionality to consider `redirect-idp` types.

@haishengwu-okta and @pradeepdewda-okta - If the current implementation is still not our ideal case, I'll set aside some time for us to chat today or tomorrow to figure out a different path forward. Sound good?

src/v2/controllers/FormController.js
jmelberg-okta: Can you elaborate a bit more on this concern?

From what I can see, we're not breaking the remediation process here, we're just handling it via navigation instead of passing the `formName` into `idx.proceed`.

pradeepdewda-okta: exactly that's my point.. we are circumventing `idx.proceed` and i feel it should not do that.

changing idx.js for this use case right now could be overkill and i don't think this PR should be blocked because of that. We can create a new jira for Idx.js change.

Can we handle this use case after passing it to idx.proceed, in `handleIdxResponseSuccess` ? Also should we handle failure in redirection ?

jmelberg-okta: Yep - agree with your concerns and believe `idx-js` is a good place to handle this (more info called out [here](https://github.com/okta/okta-signin-widget/pull/1638#discussion_r572211893)). We should schedule a follow-up with the IDX/Dev-Ex team to discuss if `idx-js` is the best place to handle the redirect.

Is the SIW design review the best place for this, or are there alternate office hours for IDX that would be more beneficial?

Also, in regards to errors: we're going to need to rely on the `idx` response to correctly handle possible error states for us. These include:
- Errors that may occur within an OAuth 2.0 redirect (invalid config, token exp/iat skew, etc) which _should_ already be handled by the Widget's parsing of the [`error_description`](https://tools.ietf.org/html/rfc6749#section-5.2)
- User selects "sign-out" or there is a timeout in the factor enrollment/challenge.

pradeepdewda-okta: One more thing to explore is to add this redirect to `handleInvokeAction` instead of `handleSaveForm`.. Basically hitting on `Enroll` button for this flow we are actually not saving any form data?

src/v2/controllers/FormController.js
jmelberg-okta: Discussed February 9, 2021:

**tl;dr**: This implementation is fine for now. Will work with the `okta-idx-js` team to standardize on a solution for handling redirects in the near future.

Follow-up: Migrate `SuccessView` and and `SSOExtension` to use this centralized logic.

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -0,0 +1,20 @@
...
+    const currentAuthenticator = this.options.appState.get('currentAuthenticator');
+    const authenticatorEnrollments = this.options.appState.get('authenticatorEnrollments')?.value || [];
+    authenticatorEnrollments.filter(authenticator => authenticator.id === currentAuthenticator.id);
+
+    // Set the view based on whether or not we're in a challenge flow
+    this.Body = authenticatorEnrollments.length
+      ? ChallengeIdpBody
+      : EnrollIdPBody;
+  },
+  Footer: AuthenticatorEnrollFooter,
```
mauriciocastillosilva-okta, line 19: We have an `AuthenticatorEnrollFooter` and an `AuthenticatorVerifyFooter`, so you might want to do the same thing you are doing with `Body`

src/v2/view-builder/views/idp/ChallengeAuthenticatorIdPBody.js
```diff
@@ -0,0 +1,23 @@
+import { loc } from 'okta';
+import BaseForm from '../../internals/BaseForm';
+
+const Body = BaseForm.extend({
+
+  className: 'challenge-idp',
```
mauriciocastillosilva-okta, line 6: learned from Haisheng, we don't need classnames here, instead we should use the ones we have in a `div` a couple of levels above. BaseView is taking care of creating those classnames for you based on the authenticator key

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -0,0 +1,20 @@
+import EnrollIdPBody from './EnrollAuthenticatorIdPBody';
+import ChallengeIdpBody from './ChallengeAuthenticatorIdPBody';
```
mauriciocastillosilva-okta, line 2: nit: we can name this file and the one above `*IdPForm` instead of `*IdPBody`, since those extend `BaseForm`

test/testcafe/framework/page-objects/IdPAuthenticatorPageObject.js
```diff
@@ -0,0 +1,29 @@
...
+
+  getPageTitle() {
+    return this.form.getElement('.okta-form-title').textContent;
+  }
+
+  getPageSubtitle() {
+    return this.form.getElement('.okta-form-subtitle').textContent;
+  }
+
+  enroll() {
```
mauriciocastillosilva-okta, line 22: nit: `clickSave` or `submitForm` might be reusable instead of having `enroll` and `verify`? Or maybe `clickEnrollButton` and `clickVerifyButton` would be similar to what we currently have in other page objects?

test/testcafe/spec/AuthenticatorIdPView_spec.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+);
+
+const enrollMock = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(xhrEnrollIdPAuthenticator)
+  .onRequestTo('http://localhost:3000/idp/idx/credential/enroll')
+  .respond(xhrEnrollIdPAuthenticator)
+  .onRequestTo('http://localhost:3000/sso/idps/0oa69chx4bZyx8O7l0g4?stateToken=02TptqPN4BOLIwMAGUVLPlZVJEnONAq7xkg19dy6Gk')
+  .respond('<html><h1>A external IdP login page for testcafe testing</h1></html>');
```
mauriciocastillosilva-okta, line 20: nit: "An external..."

test/unit/spec/v2/ion/uiSchemaTransformer_spec.js
```diff
@@ -1083,7 +1117,7 @@ describe('v2/ion/uiSchemaTransformer', function () {
     });
   });
 
-  it('converts identify remidiation response', done => {
+  it('converts identify remediation response', done => {
```
mauriciocastillosilva-okta, line 1120: 👍 thanks for fixing spelling errors!

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -0,0 +1,20 @@
...
+export default BaseAuthenticatorView.extend({
+  initialize () {
+    BaseAuthenticatorView.prototype.initialize.apply(this, arguments);
+
+    const currentAuthenticator = this.options.appState.get('currentAuthenticator');
+    const authenticatorEnrollments = this.options.appState.get('authenticatorEnrollments')?.value || [];
+    authenticatorEnrollments.filter(authenticator => authenticator.id === currentAuthenticator.id);
+
+    // Set the view based on whether or not we're in a challenge flow
+    this.Body = authenticatorEnrollments.length
```
haishengwu-okta, line 15: aha... this is interesting.
For other authenticators, they are `challenge-authenticator` and `authenticator-enrollment-data/enroll-authenticator` respectively for challenge and enroll.
I'm curious why we don't use same pattern (I meant same remediation forms) then either backend decide to redirect to IDP (responds 302) or still widget handle the redirect (`when form name is challenge-authenticator and authenticator is external idp, redirect`)

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -828,6 +828,15 @@ oie.on_prem.enroll.passcode.label = Enter authenticator passcode
 oie.on_prem.verify.title = Verify with {0}
 oie.on_prem.verify.passcode.label = Enter code
 
+## Custom Factor
+# {0} refers to to name of an authenticator for all Custom Factor values
+oie.customFactor.authenticator.description=Redirect to verify with {0}
+oie.customFactor.enroll.title=Set up {0}
+oie.customFactor.enroll.description=Clicking below will redirect to enrollment in {0}
+oie.customFactor.enroll.save=Enroll
```
haishengwu-okta, line 836: nit:
`enroll = Enroll` ?

playground/mocks/data/idp/idx/authenticator-enroll-idp.json
```diff
@@ -0,0 +1,132 @@
...
+        "method": "POST",
+        "accepts": "application/json; okta-version=1.0.0",
+        "produces": "application/ion+json; okta-version=1.0.0",
+        "value": [
+          {
+            "name": "authenticator",
+            "type": "object",
+            "options": [
+              {
+                "label": "IDP Authenticator",
```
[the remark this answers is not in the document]
jmelberg-okta: Confirmed with Ajay that these are synonymous. Will modify the mock files to reflect this!

playground/mocks/data/idp/idx/authenticator-verification-idp.json
```diff
@@ -0,0 +1,144 @@
...
+                      {
+                        "name": "id",
+                        "required": true,
+                        "value": "aut4mhtS1b84AR0iQ0g4",
+                        "mutable": false
+                      }
+                    ]
+                  }
+                },
+                "relatesTo": "$.authenticators.value[0]"
```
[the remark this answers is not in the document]
jmelberg-okta: Good catch! Will fix.

src/v2/controllers/FormController.js
```diff
@@ -147,6 +149,17 @@ export default Controller.extend({
 
     this.toggleFormButtonState(true);
     model.trigger('request');
+
+    // NOTE: okta-idx-js does not know whether a remediation
+    // is required to redirect or perform a GET request.
+    // As a result, we handle this accordingly by checking the formName
+    const fullPageRedirectForms = [ FORMS.REDIRECT_IDP ];
```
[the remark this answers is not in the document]
jmelberg-okta: Since this variable is intended for only the `handleSaveForm` method, I'd prefer to keep this as function scope instead of global. That ok?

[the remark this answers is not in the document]
pradeepdewda-okta: @jmelberg-okta did you explore the possibility of triggering invokeAction instead of using default save ? We should put this code under `handleInvokeAction`.. We don't need to run this code for every form submit that doesn't make any sense.. This is special case so handle this as a special case in `handleInvokeAction` and should be triggered only when it's required..

playground/mocks/spec-okta-api/oauth2/v1/authorize/callback.js
```diff
@@ -0,0 +1,22 @@
+const apis = [
```
pradeepdewda-okta, line 1: why to put this under `v1` code ?

packages/@okta/i18n/src/properties/login.properties
jmelberg-okta: We don't have this property yet, but I've added a new one so future factors can reuse it!

playground/mocks/spec-okta-api/oauth2/v1/authorize/callback.js
jmelberg-okta: We do not have an `/oauth2/v2` endpoint. This flow is disjoint from our IDX remediation flows.

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
pradeepdewda-okta, line 15: Instead of using length we should use something more concrete to make this decision..

src/v2/view-builder/views/idp/EnrollAuthenticatorIdPBody.js
```diff
@@ -0,0 +1,23 @@
+import { loc } from 'okta';
+import BaseForm from '../../internals/BaseForm';
+
+const Body = BaseForm.extend({
```
pradeepdewda-okta, line 4: I don't see any visual difference between this and `ChallengeAuthenticatorIdPBody.js` view.. what i am missing ? Can't we use the save view ?

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
jmelberg-okta: Glad this caught your attention!

The IdP Authenticator is intended [not to use the `idp/idx/challenge/answer` endpoint](https://oktawiki.atlassian.net/wiki/spaces/~ajay.thomas/pages/1749414459/IDP+Authenticator+on+OIE+Assurance?focusedCommentId=1776779962#Enrollment-flow) as we do not control the verification logic. I believe 

When discussing this with the backend team, it sounds like they wanted to avoid the additional forms completely - treating "redirects" as a redirect form.

> when form name is challenge-authenticator and authenticator is external idp, redirect

This is interesting, but wouldn't it require us to check multiple objects inside of the `remediation` response before acting?

src/v2/view-builder/views/idp/EnrollAuthenticatorIdPBody.js
jmelberg-okta: The primary difference between the two views are the labels used for the title, subtitle, and save action.

pradeepdewda-okta: those can be passed as option to the view ?

jmelberg-okta: Can you give an example of how we can conditionally do this? I wasn't able to find an example of us doing something like the previously, and figured this was the cleanest solution. 😄

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
jmelberg-okta: Can you provide an example? To me, as long as this value is truthy we're OK proceeding.

src/v2/view-builder/views/idp/EnrollAuthenticatorIdPBody.js
pradeepdewda-okta: When you are adding the this view just past the correct options  ?

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
pradeepdewda-okta: take a look at DeviceEnrollmentTerminalView.js

jmelberg-okta: Hmm. I don't see anything within `DeviceEnrollmentTerminalView` that pertains to this. Can you please link directly to the code or write out an example?

pradeepdewda-okta: Okay.. Got it, i think it should be fine.

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -0,0 +1,25 @@
...
+    const authenticatorEnrollments = this.options.appState.get('authenticatorEnrollments')?.value || [];
+    const existingAuthenticator = authenticatorEnrollments
+      .filter(authenticator => authenticator.id === currentAuthenticator.id);
+
+    // Set the view based on whether or not we're in a challenge flow
+    this.Body = existingAuthenticator.length
+      ? ChallengeIdpForm
+      : EnrollIdPForm;
+
+    this.Footer = existingAuthenticator.length
```
pradeepdewda-okta, line 21: AuthenticatorVerifyFooter and AuthenticatorEnrollFooter both have exactly same code.. Not sure why it was added and can't be reused.

src/v2/view-builder/views/idp/EnrollAuthenticatorIdPBody.js
jmelberg-okta: Figured out what you were trying to say here -

Essentially, we can initialize a set of `options` that our child `Body` can use within our `AuthenticatorIdPView` object. In turn, these can be used to conditionally render our labels (`title`, `subtitle`, and `save`).

Updated the PR to do this, as it removes the need to maintain an Enroll/Challenge view!

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -0,0 +1,25 @@
...
+  initialize () {
+    BaseAuthenticatorView.prototype.initialize.apply(this, arguments);
+
+    const currentAuthenticator = this.options.appState.get('currentAuthenticator');
+    const authenticatorEnrollments = this.options.appState.get('authenticatorEnrollments')?.value || [];
+    const existingAuthenticator = authenticatorEnrollments
+      .filter(authenticator => authenticator.id === currentAuthenticator.id);
+
+    // Set the view based on whether or not we're in a challenge flow
+    this.Body = existingAuthenticator.length
```
pradeepdewda-okta, line 17: this could be 
```
let options = existingAuthenticator.length ? options1 : options2;
this.Body = new IdpForm({options}) // we can avoid duplicate forms
```

src/v2/view-builder/views/idp/EnrollAuthenticatorIdPBody.js
mauriciocastillosilva-okta: You could also do what I did for Duo, which is have an IdPBaseForm and reuse it in Challenge and Enroll views, extending to have the correct title and the correct Footer

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
jmelberg-okta: @pradeepdewda-okta - Can you create a ticket so we can clean this up? We're putting these findings under the `widget-help-wanted` label for the Client SDK team?

pradeepdewda-okta: I don't think SDK team is the right team that will do all these cleanups, I think it will be responsibility of home teams. I will create a jira for this though.

https://oktainc.atlassian.net/browse/OKTA-369626

src/v2/view-builder/views/idp/EnrollAuthenticatorIdPBody.js
jmelberg-okta: ⬆️ This is what I ended up doing - keeping all logic within a single view. Thanks for the suggestions all!

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -832,6 +833,14 @@ oie.numberchallenge.warning = Haven't received a push notification yet? Try open
 ## {0} refers to a number which is the correct answer for the number challenge
 oie.numberchallenge.instruction = On your mobile device, open the Okta Verify prompt, then tap <$1>{0}</$1> in Okta Verify to continue.
 
+## IdP Authenticator
+# {0} refers to the display name of the authenticator for all IdP Authenticator values
+oie.customFactor.authenticator.description=Redirect to verify with {0}
```
mauriciocastillosilva-okta, line 838: should we change also the keys to `oie.idpAuthenticator.*`? Just to keep the same names everywhere and not confuse any future dev

playground/mocks/config/responseConfig.js
```diff
@@ -560,6 +560,20 @@ const passwordRecovery = {
   ],
 };
 
+const idpFactor = {
```
mauriciocastillosilva-okta, line 563: nit: `const idpAuthenticator`

src/v2/view-builder/utils/AuthenticatorUtil.js
```diff
@@ -110,6 +110,21 @@ const getAuthenticatorData = function (authenticator, isVerifyAuthenticator) {
...
+  case AUTHENTICATOR_KEY.IDP: {
+    const idpName =  authenticator.relatesTo?.displayName;
+    const factorDescription = idpName
+      ? loc('oie.customFactor.authenticator.description', 'login', [idpName])
+      : loc('oie.customFactor.authenticator.description.default', 'login');
+    Object.assign(authenticatorData, {
+      description: isVerifyAuthenticator
+        ? ''
+        : factorDescription,
+      iconClassName: 'mfa-custom-factor',
```
mauriciocastillosilva-okta, line 123: nit: maybe this can be also `mfa-idp-authenticator`?

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -0,0 +1,47 @@
...
+});
+
+export default BaseAuthenticatorView.extend({
+  initialize () {
+    BaseAuthenticatorView.prototype.initialize.apply(this, arguments);
+
+    const currentAuthenticator = this.options.appState.get('currentAuthenticator');
+    const authenticatorEnrollments = this.options.appState.get('authenticatorEnrollments')?.value || [];
+    const existingAuthenticator = authenticatorEnrollments
+      .filter(authenticator => authenticator.id === currentAuthenticator.id);
```
mauriciocastillosilva-okta, line 36: It might be better to use `find` or `findWhere` instead of `filter`, since you only need to know if there is a one model that matches the id, right? And that will return you just one model or undefined...
as opposed to `filter` that will go through all the elements in the array and will return you another array

test/testcafe/spec/AuthenticatorIdPView_spec.js
```diff
@@ -0,0 +1,74 @@
...
+  .onRequestTo('http://localhost:3000/sso/idps/0oa69chx4bZyx8O7l0g4?stateToken=02TptqPN4BOLIwMAGUVLPlZVJEnONAq7xkg19dy6Gk')
+  .respond('<html><h1>An external IdP login page for testcafe testing</h1></html>');
+
+async function setup(t) {
+  const pageObject = new IdPAuthenticatorPageObject(t);
+  await pageObject.navigateToPage();
+
+  const { log } = await t.getBrowserConsoleMessages();
+
+  await t.expect(log.length).eql(3);
```
mauriciocastillosilva-okta, line 36: heard someone is improving how we assert all these log messages... you should take a look at that person's PR, it will greatly improve how we do this! 😉

packages/@okta/i18n/src/properties/login.properties
jmelberg-okta: Updated to `oie.idp` to be consistent with other factors!

src/v2/view-builder/utils/AuthenticatorUtil.js
jmelberg-okta: This will require CSS-level changes to extend the existing `mfa-custom-factor` icon to `mfa-idp-authenticator`.

Since this value is internal only and will not be read via a screen reader - I think we should keep this as-is to avoid the extra overrides. That work?

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
jmelberg-okta: Good suggestion! Will update.

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -0,0 +1,46 @@
...
+export default BaseAuthenticatorView.extend({
+  initialize () {
+    BaseAuthenticatorView.prototype.initialize.apply(this, arguments);
+
+    const currentAuthenticator = this.options.appState.get('currentAuthenticator');
+    const authenticatorEnrollments = this.options.appState.get('authenticatorEnrollments')?.value || [];
+    const existingAuthenticator = _.find(authenticatorEnrollments, { id: currentAuthenticator.id });
+
+    // If the authenticator is already enrolled, we're in a challenge flow as opposed to enrollment
+    this.options.isChallenge = existingAuthenticator || false;
```
haishengwu-okta, line 38: can we do this for body?
```
this.Body = isChallenge ? Foo : Bar
```

haishengwu-okta (commented):
one question

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -0,0 +1,46 @@
...
+
+});
+
+export default BaseAuthenticatorView.extend({
+  initialize () {
+    BaseAuthenticatorView.prototype.initialize.apply(this, arguments);
+
+    const currentAuthenticator = this.options.appState.get('currentAuthenticator');
+    const authenticatorEnrollments = this.options.appState.get('authenticatorEnrollments')?.value || [];
+    const existingAuthenticator = _.find(authenticatorEnrollments, { id: currentAuthenticator.id });
```
mauriciocastillosilva-okta, line 35: nit: I think backbone collection has a `findWhere` that does exactly the same as `_.find`, but you might save importing underscore... just a nit, either works

mauriciocastillosilva-okta (approved):
looks good 🐑

haishengwu-okta (approved):

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
jmelberg-okta: I _think_ this comment is referencing an outdated commit, as we now utilize a single `BaseView` and pass through `options` into the `Body`.

swiftone (approved):

brettritter-okta (approved):

PR #1640, opened 2021-02-03
fix: fetch via authjs 4.6.2. bump version to 5.3.1

aarongranick-okta:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@okta/okta-signin-widget",
   "description": "The Okta Sign-In Widget",
-  "version": "5.3.0",
+  "version": "5.3.1",
   "homepage": "https://github.com/okta/okta-signin-widget",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@okta/okta-signin-widget",
   "description": "The Okta Sign-In Widget",
-  "version": "5.3.0",
+  "version": "5.3.1",
```
magizh-okta, line 4: Why this ?

aarongranick-okta: I am combining two PRs into one. Both would be changing the `package.json`. We can add other commits to the branch before we promote.

haishengwu-okta: not ideal as mess up the commit history a bit.

haishengwu-okta (approved):

package.json
aarongranick-okta: i can remove it if you want. i thought we were in a hurry

aarongranick-okta: i can't do separate commits on this PR because it is also modifying package.json

PR #1644, opened 2021-02-04
Add support for number challenge polling view

nikhilvenkatraman-okta:
Resolves: OKTA-366354

## Description:

Add support for number challenge polling view

## PR Checklist

- [x] Added e2e tests

### Screenshot/Video:

Screen Shot 2021-02-03 at 5 13 49 PM

Screen Shot 2021-02-03 at 5 14 25 PM

https://okta.box.com/s/ycs6flyqb5bbp9j6s65io2gm2y2swn85

### Reviewers:

@magizh-okta @pradeepdewda-okta @haishengwu-okta @TianGan-okta @jasleennuno-okta 

### Issue:

- [OKTA-366354](https://oktainc.atlassian.net/browse/OKTA-366354)


src/v2/view-builder/views/ov/NumberChallengePushView.js
```diff
@@ -0,0 +1,122 @@
...
+      this.showWarningWithDelay();
+    },
+
+    title () {
+      return loc('oie.okta_verify.push.sent', 'login');
+    },
+
+    initialize () {
+      BaseForm.prototype.initialize.apply(this, arguments);
+      this.listenTo(this.model, 'error', this.stopPoll);
```
pradeepdewda-okta, line 68: Instead of adding a listener (which is at many places), override `TriggerAfterError()`

test/testcafe/spec/NumberChallengePushView_spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+const numberChallengeWaitMock = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(numberChallenge)
+  .onRequestTo('http://localhost:3000/idp/idx/challenge/poll')
+  .respond(numberChallenge)
+  .onRequestTo('http://localhost:3000/idp/idx/resend')
+  .respond(numberChallenge);
+
+fixture('Challenge Okta Verify Push');
```
pradeepdewda-okta, line 28: Can you add a test that stop poll once session expires or in case of any errors?

src/v2/view-builder/views/ov/NumberChallengePushView.js
```diff
@@ -0,0 +1,122 @@
...
+      this.clearWarning();
+    },
+
+    showWarning () {
+      this.clearWarning();
+      this.add(warningTemplate, '.o-form-error-container');
+    },
+
+    clearWarning () {
+      if (this.$('.o-form-error-container div').hasClass('okta-form-infobox-warning')) {
```
pradeepdewda-okta, line 99: see if you can trigger `clearFormError` instead..

src/v2/view-builder/views/ov/NumberChallengePushView.js
```diff
@@ -0,0 +1,122 @@
...
+        this.$('.okta-form-infobox-warning').remove();
+      }
+      clearTimeout(this.warningTimeout);
+    },
+
+    showWarningWithDelay () {
+      const warningInfo = this.$('.okta-form-infobox-warning');
+      warningInfo.hide();
+      this.showWarningTimeout = _.delay(() => {
```
pradeepdewda-okta, line 108: what is `warningTimeout` and `showWarningTimeout` ?

src/v2/view-builder/views/ov/NumberChallengePushView.js
```diff
@@ -0,0 +1,122 @@
...
+      this.addView();
+      correctAnswer = this.options.appState.get('currentAuthenticator').contextualData.correctAnswer;
+    },
+
+    addView () {
+      this.add(numberChallengeTemplate);
+    },
+
+    postRender () {
+      this.startPoll();
```
pradeepdewda-okta, line 78: please make sure that on stateToken expiration, polling does stop.

src/v2/view-builder/views/ov/ChallengeOktaVerifyView.js
```diff
@@ -2,16 +2,21 @@ import BaseAuthenticatorView from '../../components/BaseAuthenticatorView';
...
+import NumberChallengePushView from './NumberChallengePushView';
 
 export default BaseAuthenticatorView.extend({
   initialize () {
     BaseAuthenticatorView.prototype.initialize.apply(this, arguments);
-
     const currentAuthenticator = this.options?.appState?.get('currentAuthenticator');
     const selectedMethod = currentAuthenticator?.methods[0];
-    if (selectedMethod?.type === 'push') {
+    if (selectedMethod?.type === 'push' &&
```
pradeepdewda-okta, line 12: do we have a constant key for `push`, we use it, otherwise add it.

src/v2/view-builder/views/ov/NumberChallengePushView.js
```diff
@@ -0,0 +1,122 @@
...
+
+    clearWarning () {
+      if (this.$('.o-form-error-container div').hasClass('okta-form-infobox-warning')) {
+        this.$('.okta-form-infobox-warning').remove();
+      }
+      clearTimeout(this.warningTimeout);
+    },
+
+    showWarningWithDelay () {
+      const warningInfo = this.$('.okta-form-infobox-warning');
```
pradeepdewda-okta, line 106: possible to avoid direct DOM access ?

src/v2/view-builder/views/ov/NumberChallengePushView.js
```diff
@@ -0,0 +1,122 @@
+import { _, loc, View } from 'okta';
+import hbs from 'handlebars-inline-precompile';
+import BaseForm from '../../internals/BaseForm';
+import polling from '../shared/polling';
+import { WARNING_TIMEOUT } from '../../utils/Constants';
+
+let correctAnswer = '';
+const warningTemplate = View.extend({
```
pradeepdewda-okta, line 8: I would rather create their own views and handle show/hide/delay in those views like `ResendView.js` ?

src/v2/view-builder/views/ov/NumberChallengePushView.js
```diff
@@ -0,0 +1,122 @@
...
+    <span class="icon warning-16"></span>
+     <p>{{i18n
+      code="oie.numberchallenge.warning"
+      bundle="login"
+      $1="<a href='#' class='resend-number-challenge'>$1</a>"}}
+    </p>
+  `
+});
+
+const numberChallengeTemplate = View.extend({
```
pradeepdewda-okta, line 20: same here ^^

assets/sass/v2/_number-challenge.scss
```diff
@@ -0,0 +1,33 @@
+.number-challenge-section {
```
haishengwu-okta, line 1: Shall we scope all v2 css under the v2 specific top level class name?
Not specific to `number-challenge.scss` but also some other files.

src/v2/view-builder/views/ov/NumberChallengePushView.js
```diff
@@ -0,0 +1,122 @@
...
+  className: 'number-challenge-section',
+  template: hbs`
+     <p>
+     {{i18n
+      code="oie.numberchallenge.subtitle"
+      bundle="login"
+      arguments="correctAnswer"
+      $1="<span class='strong'>$1</span>"}}
```
haishengwu-okta, line 28: aha.. courage already support the new i18n syntax ?!!

src/v2/view-builder/views/ov/NumberChallengePushView.js
nikhilvenkatraman-okta: Curious what is the advantage of using `TriggerAfterError`?

assets/sass/v2/_number-challenge.scss
nikhilvenkatraman-okta: Are you suggesting we have some prefix that we add to top level class names ? Something like `oie-number-challenge-section` and so on for others ?

We already have all the sass inside a v2 folder so maybe it might be an overkill. Another thing to think about is are we breaking any CSS customizations a customer may have by adding new classnames.

src/v2/view-builder/views/ov/NumberChallengePushView.js
nikhilvenkatraman-okta: This is needed to hide the warning when a user hits resend push notification and then show it again after a timeout.

src/v2/view-builder/views/ov/NumberChallengePushView.js
nikhilvenkatraman-okta: Going to remove this method and just call `this.stopPoll(); this.startPoll();` from `handleResendNumberChallenge` . Calling `stopPoll` should clear the warning.

src/v2/view-builder/views/ov/NumberChallengePushView.js
nikhilvenkatraman-okta: Triggering `clearFormError` to clear the warning seems a bit odd also I don't see any other place doing this. Also handleClearFormError looks for the className `o-form-has-errors` where as we are working with `okta-form-infobox-warning`.

src/v2/view-builder/views/ov/NumberChallengePushView.js
pradeepdewda-okta: we have the ability, so we can use it. We can do more in it apart from just `stopPoll`.. just like we override `remove` instead of adding a listener

src/v2/view-builder/views/ov/NumberChallengePhoneView.js
```diff
@@ -0,0 +1,35 @@
+import { View } from 'okta';
+import hbs from 'handlebars-inline-precompile';
+
+let correctAnswer = '';
+const NumberChallengePhoneView = View.extend({
+  initialize () {
+    correctAnswer = this.options.appState.get('currentAuthenticator').contextualData.correctAnswer;
```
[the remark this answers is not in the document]
pradeepdewda-okta: make sense to make it `this.correctAnswer` ? Instead of making it available globally

src/v2/view-builder/views/ov/NumberChallengePushView.js
```diff
@@ -0,0 +1,73 @@
...
+    className: 'okta-verify-number-challenge',
+
+    events: {
+      'click a.resend-number-challenge': 'handleResendNumberChallenge'
+    },
+
+    handleResendNumberChallenge () {
+      this.options.appState.trigger('invokeAction', 'currentAuthenticator-resend');
+      // remove the warning
+      this.stopPoll();
```
pradeepdewda-okta, line 21: instead startPoll can always call `this.stopPoll` first before starting polling...

src/v2/view-builder/views/ov/NumberChallengePhoneView.js
```diff
@@ -0,0 +1,35 @@
+import { View } from 'okta';
+import hbs from 'handlebars-inline-precompile';
+
+let correctAnswer = '';
+const NumberChallengePhoneView = View.extend({
+  initialize () {
```
jasleennuno-okta, line 6: do we need the initialize? can we just say `return this.options.appState.get('currentAuthenticator').contextualData.correctAnswer` in `getTemplateData`?

src/v2/view-builder/views/ov/ResendNumberChallengeView.js
```diff
@@ -0,0 +1,16 @@
+import { View } from 'okta';
+import hbs from 'handlebars-inline-precompile';
+
+const ResendNumberChallengeView = View.extend({
```
pradeepdewda-okta, line 4: I was wondering better would be to let this view show/clear warning instead of `NumberChallengePushView.js`.. main view can trigger events to do that.. Basically `this.warningTimeout` can completely resides in this view.

pradeepdewda-okta (commented):
this looks good. few more comments.

src/v2/view-builder/views/ov/NumberChallengePhoneView.js
nikhilvenkatraman-okta: Good call !

src/v2/view-builder/views/ov/NumberChallengePhoneView.js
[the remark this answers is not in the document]
nikhilvenkatraman-okta: Moved this to a local variable within `getTemplateData`

src/v2/view-builder/views/ov/NumberChallengePushView.js
nikhilvenkatraman-okta: That might not apply for when the polling start the first time.

pradeepdewda-okta: yeah that would be a `null` action then..

src/v2/view-builder/views/ov/ResendNumberChallengeView.js
nikhilvenkatraman-okta: Yeah, I thought about that but it feels like it adds more complexity and I was trying to avoid that. To do this it would mean that the `ResendNumberChallenge` view would also need to know when polling started and stopped, for which I would have to trigger a few more events and have listeners in the `ResendNumberChallengeView`.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -819,6 +819,10 @@ oie.verify.google_authenticator.otp.title = Verify with Google Authenticator
 oie.verify.google_authenticator.otp.description = Enter the temporary code generated in your Google Authenticator app
 oie.google_authenticator.otp.enterCodeText = Enter code
 
+## Number Challenge
+oie.numberchallenge.warning = Haven't received a push notification yet? Try opening the Okta Verify app on your device or <$1>resend the push notification</$1>.
+oie.numberchallenge.subtitle = On your mobile device, open the Okta Verify prompt, then tap <$1>{0}</$1> in Okta Verify to continue.
```
jmelberg-okta: Love seeing this tag syntax used!

nit: For positional arguments, let's add a comment above the key for our translators to know what it is in reference to.

src/v2/view-builder/views/ov/ChallengeOktaVerifyView.js
```diff
@@ -2,16 +2,22 @@ import BaseAuthenticatorView from '../../components/BaseAuthenticatorView';
...
 export default BaseAuthenticatorView.extend({
   initialize () {
     BaseAuthenticatorView.prototype.initialize.apply(this, arguments);
-
     const currentAuthenticator = this.options?.appState?.get('currentAuthenticator');
     const selectedMethod = currentAuthenticator?.methods[0];
-    if (selectedMethod?.type === 'push') {
+    if (selectedMethod?.type === AUTHENTICATOR_METHODS.PUSH &&
+      currentAuthenticator.contextualData &&
+      currentAuthenticator.contextualData.correctAnswer) {
```
jmelberg-okta, line 15: Can we utilize [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here instead of the nested check?

src/v2/view-builder/views/ov/ChallengeOktaVerifyView.js
```diff
@@ -2,16 +2,22 @@ import BaseAuthenticatorView from '../../components/BaseAuthenticatorView';
...
+      this.Body = NumberChallengePushView;
+    } else if (selectedMethod?.type === AUTHENTICATOR_METHODS.PUSH) {
       this.Body = ChallengeOktaVerifyPushView;
-    } else {
+    }
+    else {
```
jmelberg-okta, line 20: style-nit: Did prettier modify this?

src/v2/view-builder/views/ov/NumberChallengePushView.js
```diff
@@ -0,0 +1,78 @@
...
+    },
+
+    stopPoll () {
+      this.stopPolling();
+      this.clearWarning();
+    },
+
+    showWarning () {
+      this.clearWarning();
+      this.add(ResendNumberChallengeView, '.o-form-error-container');
```
jmelberg-okta, line 59: Why do we need to call `this.clearWarning` beforehand? It appears that we'll only call `showWarning` once after the view renders.

assets/sass/v2/_number-challenge.scss
haishengwu-okta: Check `_device-challenge.scss`, it scopes css under `.siw-main-view`.
Such pattern probably works all all other css that meant for the widget page body.
(could do as separated refactoring PR)

src/v2/view-builder/views/ov/NumberChallengePhoneView.js
```diff
@@ -0,0 +1,33 @@
...
+        <div class="phone--screen">
+          <span class="phone--number" data-se="challenge-number">{{correctAnswer}}</span>
+        </div>
+        <div class="phone--home-button"></div>
+      </div>
+    </div>
+  `,
+  getTemplateData () {
+    const correctAnswer = this.options.appState.get('currentAuthenticator').contextualData &&
+      this.options.appState.get('currentAuthenticator').contextualData.correctAnswer;
```
haishengwu-okta, line 25: use `?.` operator?

assets/sass/v2/_number-challenge.scss
nikhilvenkatraman-okta: Ah makes sense to scope it under .siw-main-view

src/v2/view-builder/views/ov/ChallengeOktaVerifyView.js
nikhilvenkatraman-okta: My editor might have messed the formatting. Fixed!

src/v2/view-builder/views/ov/NumberChallengePushView.js
nikhilvenkatraman-okta: you are right, we can remove it

stevennguyen-okta (changes requested):
Overall looks great. One minor fix request: In the "havent received a push" infobox, could you add the comma after "device"? so that it reads out as "... on your device, or resend the push notification."

Also, the number should change when the push is resent, is that going to be handled separately from this PR?

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -819,6 +819,10 @@ oie.verify.google_authenticator.otp.title = Verify with Google Authenticator
 oie.verify.google_authenticator.otp.description = Enter the temporary code generated in your Google Authenticator app
 oie.google_authenticator.otp.enterCodeText = Enter code
 
+## Number Challenge
+oie.numberchallenge.warning = Haven't received a push notification yet? Try opening the Okta Verify app on your device or <$1>resend the push notification</$1>.
```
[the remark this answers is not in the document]
nikhilvenkatraman-okta: @magizh-okta good call out. The string is slightly different 

```
Haven't received a push notification yet? Try opening the Okta Verify app on your device or <$1>resend the push notification</$1>.
``` 
```
Haven't received a push notification yet? Try opening the Okta Verify App on your phone. 
```
@stevennguyen-okta  Are we ok to use the existing string from V1 ? `Haven't received a push notification yet? Try opening the Okta Verify App on your phone. `. 

The main reason for this is that if a V1 customer has customized this string using the old key `oktaverify.warning` it will be lost if we create a new string for OIE.

nikhilvenkatraman-okta:
> Also, the number should change when the push is resent, is that going to be handled separately from this PR?

@stevennguyen-okta Good eye. This is working based on mock API response, when its integrated with the real API a new number will be sent back from the API after the user hits resend.

src/v2/view-builder/views/ov/ResendNumberChallengeView.js
```diff
@@ -0,0 +1,16 @@
...
+  className: 'okta-form-infobox-warning infobox infobox-warning',
```
magizh-okta, line 5: can we use createCallout instead ?

test/testcafe/spec/NumberChallengePushView_spec.js
```diff
@@ -0,0 +1,104 @@
...
+    let warningBox = challengeOktaVerifyPushPageObject.getWarningBox();
+    await t.expect(warningBox.visible).ok();
+    await t.expect(warningBox.innerText)
+      .eql('Haven\'t received a push notification yet? Try opening the Okta Verify app on your device or resend the push notification.');
+    // click resend number challenge
+    challengeOktaVerifyPushPageObject.clickResendNumberChallenge();
+    await t.expect(logger.count(() => true)).eql(8);
+    const req = logger.requests[7].request;
+    const reqBody = JSON.parse(req.body);
+    await t.expect(reqBody).eql({
```
magizh-okta, line 84: these tests have some parts in common. Since both add 30s wait we can get rid of first one ?

src/v2/view-builder/views/ov/ResendNumberChallengeView.js
magizh-okta: I thoght we dint have callout but I see that it is used in other places.

magizh-okta (approved):
Looking good just a couple of improvements.

src/v2/view-builder/views/ov/NumberChallengePushView.js
```diff
@@ -0,0 +1,77 @@
...
+
+    initialize () {
+      BaseForm.prototype.initialize.apply(this, arguments);
+      this.add(NumberChallengePhoneView);
+    },
+
+    triggerAfterError () {
+      BaseForm.prototype.triggerAfterError.apply(this, arguments);
+      this.stopPolling();
+      this.$el.find('.o-form-fieldset-container').empty();
```
pradeepdewda-okta, line 38: @haishengwu-okta see, for this i added a common handler in baseForm to clear fieldset content in my last PollView  PR.. do you think we should add a common handler now ? 

```
this.trigger('handleClearFieldsetContent');
```

src/v2/view-builder/views/ov/ResendNumberChallengeView.js
pradeepdewda-okta: `ResendNumberChallenge` should not know about polling.. it can just receive an event when to start and stop right?

```
    startPoll () {
      this.startPolling();
      this.trigger('toggleWarning'); //can handle start and stop 
    }
```

src/v2/view-builder/views/ov/ResendNumberChallengeView.js
nikhilvenkatraman-okta: Yeah, we have a mix. We have used a template for warning in some places as well  https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/ov/ChallengeOktaVerifyPushView.js#L7-L13

nikhilvenkatraman-okta: Using a template works well here as we are using the new i18n syntax to add the resend link, which doesnt seem to render correctly with `createCallout`

pradeepdewda-okta: wow.. didn't knew we already have similar code. We should create something reusable that can handle such views ? 
like handling warnings, clearing them out, common template ?

We can have more such views in future.

nikhilvenkatraman-okta: Having something re-usable would be great, but it needs some more thought as each of these views have some nuance   that this component needs to be able to handle.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -819,6 +819,11 @@ oie.verify.google_authenticator.otp.title = Verify with Google Authenticator
 oie.verify.google_authenticator.otp.description = Enter the temporary code generated in your Google Authenticator app
 oie.google_authenticator.otp.enterCodeText = Enter code
 
+## Number Challenge
+oie.numberchallenge.warning = Haven't received a push notification yet? Try opening the Okta Verify app on your device, or <$1>resend the push notification</$1>.
+## argument {0} in oie.numberchallenge.instruction is a number
```
TianGan-okta, line 824: Please refer to the i18n-smartling format: ## {0} refers to xxx

TianGan-okta (approved):

src/v2/view-builder/views/ov/ResendNumberChallengeView.js
nikhilvenkatraman-okta: @pradeepdewda-okta  There are a few things that will need to happen for this 

-  The `ResendNumberChallengeView ` will be added immediately, gets hidden on load and is then listening to 
`showWarning`
-  `NumberChallengePushView`  then triggers `showWarning` 
-  `ResendNumberChallengeView` will then need to start the timeout and shows the warning. 
-  `NumberChallengePushView` will trigger another evert to hide the warning when user hits resend.

These will all be events on the `appState` as `state` is not available within these views. Let me know if this is inline with what you are thinking.

packages/@okta/i18n/src/properties/login.properties
nikhilvenkatraman-okta: Thanks, can you point me to a wiki ?

nikhilvenkatraman-okta: Updating this to `## {0} refers to a number which is the correct answer for the number challenge`

src/v2/view-builder/views/ov/NumberChallengePushView.js
haishengwu-okta: - curious why need to hide the form?
- would it be more generic like `trigger('hideFormContent')`

pradeepdewda-okta (approved):
@nikhilvenkatraman-okta thanks for making the changes.. Looks good.

src/v2/view-builder/views/ov/NumberChallengePushView.js
pradeepdewda-okta: It's not all form content.. it's just the fieldset.

stevennguyen-okta (approved):
Looks great

PR #1654, opened 2021-02-10
Update button text for push and num challenge error state

nikhilvenkatraman-okta:
Resolves: OKTA-366355

## Description:

- Update button text for push and number challenge error state

Screen Shot 2021-02-10 at 9 36 54 AM

## PR Checklist

- [x] Added e2e tests

### Reviewers:

@gowthamidommety-okta @stevennguyen-okta 

### Issue:

- [ OKTA-366355](https://oktainc.atlassian.net/browse/ OKTA-366355)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -803,7 +803,7 @@ oie.okta_verify.totp.title = Enter a code
 oie.okta_verify.totp.enterCodeText = Enter code from Okta Verify app
 oie.okta_verify.push.title = Get a push notification
 oie.okta_verify.push.sent = Push notification sent
-oie.okta_verify.push.send = Send push notification
+oie.okta_verify.push.resend = Resend push notification
 oie.okta_verify.signed_nonce.label = Use Okta Verify on this device
 
 ## Google authenticator
--- a/src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
+++ b/src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
@@ -12,7 +12,7 @@ const Body = BaseForm.extend(Object.assign(
     },
 
     save () {
-      return loc('oie.okta_verify.push.send', 'login');
+      return loc('oie.okta_verify.push.resend', 'login');
     },
 
     showMessages () {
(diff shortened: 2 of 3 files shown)


stevennguyen-okta (changes requested):
Minor request: Could you change "Push" and "Notification" to lowercase? This was an error on my part for going out casing standardization. I have corrected the mock.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -803,7 +803,7 @@ oie.okta_verify.totp.title = Enter a code
 oie.okta_verify.totp.enterCodeText = Enter code from Okta Verify app
 oie.okta_verify.push.title = Get a push notification
 oie.okta_verify.push.sent = Push notification sent
-oie.okta_verify.push.send = Send push notification
```
gowthamidommety-okta, line 806: I vaguely remember we auto push by default.. but i can't remember if that changed. Is there no case where we need `Send push notification`?

gowthamidommety-okta (commented):
One question

packages/@okta/i18n/src/properties/login.properties
gowthamidommety-okta: If we don't need `Send push notification` does it make sense to change the key to `oie.okta_verify.push.resend`

nikhilvenkatraman-okta: @gowthamidommety-okta  I can update the key name. Currently I only see this key used in one place https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js#L15

gowthamidommety-okta: got it. If you can change the key name / add a new key that would be great :)

stevennguyen-okta (approved):
looks good!

gowthamidommety-okta (approved):
lgtm! One small thing to check before merge: is it ok to delete existing strings?

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -803,7 +803,7 @@ oie.okta_verify.totp.title = Enter a code
...
+oie.okta_verify.push.resend = Resend push notification
```
pradeepdewda-okta, line 806: when we do the i18n translations this key name change gets taken care of ? @evelynteo-okta

PR #1655, opened 2021-02-10
fix: Resolve render promise only on SUCCESS status

denysoblohin-okta:
## Description:

Forgot password flow (as well as unlock account flow, sign-up flow) should not resolve `showSigninToGetTokens` promise

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-360093](https://oktainc.atlassian.net/browse/OKTA-360093)

### DIFF
--- a/src/models/LoginModel.js
+++ b/src/models/LoginModel.js
@@ -19,7 +19,7 @@ export default BaseLoginModel.extend({
   },
   loginWithActivationToken: function (activationToken) {
     return this.startTransaction(function (authClient) {
-      return authClient.signIn({ token: activationToken });
+      return authClient.signInWithCredentials({ token: activationToken });
     });
   },
 });
--- a/src/widget/createRouter.js
+++ b/src/widget/createRouter.js
@@ -1,4 +1,5 @@
 import { $ } from 'okta';
+import Enums from 'util/Enums';
 
 export default function createRouter (Router, widgetOptions, renderOptions, authClient, successFn, errorFn) {
   let router;
@@ -8,7 +9,9 @@ export default function createRouter (Router, widgetOptions, renderOptions, auth
       authClient,
       globalSuccessFn: (res) => {
         successFn && successFn(res); // call success function if provided
-        resolve(res);
+        if (res.status === Enums.SUCCESS) {
+          resolve(res);
+        }
       },
       globalErrorFn: (error) => {
         errorFn && errorFn(error); // call error function if provided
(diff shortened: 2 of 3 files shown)


aarongranick-okta:
@denysoblohin-okta Let's also add logic for `UNLOCK_ACCOUNT_EMAIL_SENT` (same as for this, silently ignore).

https://github.com/okta/okta-signin-widget/blob/master/playground/main.js#L30

src/widget/OktaSignIn.js
```diff
@@ -22,6 +22,10 @@ var OktaSignIn = (function () {
           _.extend({}, widgetOptions, renderOptions, {
             authClient: authClient,
             globalSuccessFn: (res) => {
+              const statusesToIgnore = ['UNLOCK_ACCOUNT_EMAIL_SENT', 'FORGOT_PASSWORD_EMAIL_SENT'];
```
aarongranick-okta, line 25: The early return logic should only affect the returned promise. The success callback should still fire as it has before.

denysoblohin-okta: Done

denysoblohin-okta:
Maybe instead of blacklisting status ('UNLOCK_ACCOUNT_EMAIL_SENT', 'FORGOT_PASSWORD_EMAIL_SENT') we should check for status 'SUCCESS' when resolving promise?

aarongranick-okta:
> Maybe instead of blacklisting status ('UNLOCK_ACCOUNT_EMAIL_SENT', 'FORGOT_PASSWORD_EMAIL_SENT') we should check for status 'SUCCESS' when resolving promise?

Yes.

src/widget/OktaSignIn.js
```diff
@@ -23,7 +24,9 @@ var OktaSignIn = (function () {
             authClient: authClient,
             globalSuccessFn: (res) => {
               successFn && successFn(res); // call success function if provided
-              resolve(res);
+              if (res.status === Enums.SUCCESS) {
```
pradeepdewda-okta, line 27: what if the status is not success.. but the pipeline still ended without any error?  for eg in case of `redirects` ?

aarongranick-okta: in this case the promise will not resolve (this is current behavior for redirects)

pradeepdewda-okta: what's the difference between calling `globalSuccessFn` and than again check for `res.status` ? what was missing here ?

aarongranick-okta: The global success function can be called multiple times (checking res.status), but the promise can only resolve one time, on success. Apps that need to handle `res.status` that is not `SUCCESS` will still need to pass a successFn.

aarongranick-okta (approved):

aarongranick-okta (approved):

PR #1657, opened 2021-02-10
fix: authParams passed to showSignInToGetTokens

aarongranick-okta:
## Description:
Additional "authParams" which are passed to `showSignInToGetTokens` and `showSignInAndRedirect` will be passed to `renderEl`

We are calling this a "fix" since it is changing the behavior to align with customer expectations. The logic has technically been unchanged since the methods were created, but their usage has increased since we feature them in our samples now.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-361428](https://oktainc.atlassian.net/browse/OKTA-361428)

### DIFF
--- a/src/util/OAuth2Util.js
+++ b/src/util/OAuth2Util.js
@@ -16,6 +16,21 @@ import Errors from './Errors';
 import Util from './Util';
 const util = {};
 
+// https://github.com/okta/okta-auth-js#authorize-options
+const AUTH_PARAMS = [
+  'responseType',
+  'scopes',
+  'state',
+  'nonce',
+  'idp',
+  'idpScope',
+  'display',
+  'prompt',
+  'maxAge',
+  'loginHint'
+];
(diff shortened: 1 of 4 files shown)


swiftone (approved):

aarongranick-okta:
Screen Shot 2021-02-10 at 3 42 06 PM

davidantonwalters-okta (commented):
Should there be a selenium case for this so we have a non-mock validation?

aarongranick-okta:
> Should there be a selenium case for this so we have a non-mock validation?

These methods are added as convenience wrappers around `renderEl`, which all of our E2E and testcafe tests exercise in many ways.  The new unit tests are verifying that the wrapper functions are passing the correct options to `renderEl`. There should be no difference in behavior between calling `showSignInAndRedirect` or calling `renderEl` directly with the same set of options (mode="remediation")

PR #1658, opened 2021-02-11
Centralize console assertions for testcafe suite

jmelberg-okta:
## Description:

- Extract duplicative log assertions for the testcafe suite into a single source file.

## PR Checklist

## Resolves

- [OKTA-369480](https://oktainc.atlassian.net/browse/OKTA-369480)

### DIFF
--- a/test/testcafe/spec/WidgetCustomization_spec.js
+++ b/test/testcafe/spec/WidgetCustomization_spec.js
@@ -123,14 +123,14 @@ test.requestHooks(identifyMock)('should show custom buttons links', async t => {
       'title': 'Custom Button 1',
       'className': 'btn-customAuth-1',
       'click': function () {
-        window.location.href = 'http://www.example.com/';
+        window.location.href = 'https://www.example.com/';
       }
     },
     {
       'title': 'Custom Button 2',
       'className': 'btn-customAuth-1',
       'click': function () {
-        window.location.href = 'http://www.example2.com/';
+        window.location.href = 'https://www.example2.com/';
       }
     }]
   });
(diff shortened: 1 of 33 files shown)


test/testcafe/spec/WidgetCustomization_spec.js
```diff
@@ -123,14 +123,14 @@ test.requestHooks(identifyMock)('should show custom buttons links', async t => {
       'title': 'Custom Button 1',
       'className': 'btn-customAuth-1',
       'click': function () {
-        window.location.href = 'http://www.example.com/';
+        window.location.href = 'https://www.example.com/';
```
aarongranick-okta, line 126: should we use a URL that will not load, or will load locally?

aarongranick-okta:
Looks good. Is it always checking console messages for `afterRender`, are there never messages between `ready` and `afterRender` ?

test/testcafe/spec/WidgetCustomization_spec.js
jmelberg-okta: This URL is actually fetched during our tests, so Ideally we'll add a mock for it. Interestingly... this website is down for some reason, causing our tests to fail.

haishengwu-okta (approved):
🏎️  Thanks!

aarongranick-okta (approved):

mauriciocastillosilva-okta (approved):

jmelberg-okta:
> Looks good. Is it always checking console messages for afterRender, are there never messages between ready and afterRender ?

Good question @aarongranick-okta!

From what I can see, assuming the view loads we'll sequentially call `ready` and then `afterRender`. On success (view loads), the third object is always going to be `context`, as defined [here](https://github.com/okta/okta-signin-widget/blob/30bdf93dd9e0e8a0f536c92a184b194a642dc335/playground/main.js#L75). When an error gets triggered via `afterError`, then we'll have a [4th log](https://github.com/okta/okta-signin-widget/blob/30bdf93dd9e0e8a0f536c92a184b194a642dc335/playground/main.js#L84), which is tested via UI selectors instead of capturing console messages.

PR #1663, opened 2021-02-12
feat: show errors related to OIE migration

aarongranick-okta:
## Description:

The purpose of this PR is to improve the user and developer experience around migrating from V1 to OIE.

Detects 3 special case errors:

1. Attempting to get tokens without redirect (`showSignInToGetTokens()`) when MFA is enabled in sign on policy. With V1, MFA must be done on the Okta hosted page. With OIE, MFA can be done client-side. So this error may occur after switching the OIE feature off. Currently the user experience for this error in an endless spinner. With this PR:

Screen Shot 2021-02-18 at 11 06 09 AM

2. Setting `useInteractionCodeFlow` to true when OIE is not enabled. This can also happen after migrating to OIE, but then turning the feature off. The current user experience is an empty collapsed widget. With this PR:

Screen Shot 2021-02-11 at 5 18 18 PM

3. Will throw an error immediately if `useInteractionCodeFlow` is set to `true`, but PKCE is not enabled. This avoids a confusing error about "codeVerifier" on the `/introspect` call, which occurs later. This config error may be seen by a developer while migrating an app that is using implicit flow to OIE.

Adds tests and mocks for these errors, and a few for interaction code in general.
Adds oauth mocks so that PKCE can be enabled in playground / karma.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-369757](https://oktainc.atlassian.net/browse/OKTA-369757)


swiftone (approved):

src/v2/BaseLoginRouter.js
```diff
@@ -153,7 +153,25 @@ export default Router.extend({
   },
 
   handleIdxResponseFailure (error = {}) {
-    // 1. loosely check whether is IDX error response
+    // special case: `useInteractionCodeFlow` is true but the Org is not using OIE
+    const featureMsg = 'The requested feature is not enabled in this environment.';
```
pradeepdewda-okta, line 157: the error messages needs to be translated..

pradeepdewda-okta (changes requested):
i18n missing ? :)

playground/mocks/data/idp/idx/error-feature-not-enabled.json
```diff
@@ -0,0 +1,4 @@
+{
+  "error": "access_denied",
+  "error_description": "The requested feature is not enabled in this environment."
```
pradeepdewda-okta, line 3: are we okay to expose this to enduser ? should we more specific or clear in our messaging ?

src/v2/BaseLoginRouter.js
```diff
@@ -153,7 +153,25 @@ export default Router.extend({
...
+    if (error?.error === 'access_denied' && error.error_description === featureMsg) {
```
pradeepdewda-okta, line 158: instead of `error_description` i would use `i18nkey`

src/v2/BaseLoginRouter.js
aarongranick-okta: This is the error that comes from the backend

src/v2/BaseLoginRouter.js
aarongranick-okta: Unfortunately the mock JSON is the complete server response. This is the error returned in a V1 environment when a client tries to use the interaction code flow.

We could show ALL "access_denied" errors.

playground/mocks/data/idp/idx/error-feature-not-enabled.json
aarongranick-okta: this is mock, but I copied from the actual response I received from the server

src/util/OAuth2Util.js
```diff
@@ -44,11 +44,24 @@ util.getTokens = function (settings, params, controller) {
...
-    // OKTA-104330- Handle error case where user is not assigned to OIDC client
+    let showError = false;
+
+    // user is not assigned to OIDC client
     if (error.errorCode === 'access_denied') {
+      showError = true;
+    }
+
+    // MFA is required but prompt=none (authn V1)
+    const mfaRequiredMsg = 'The client specified not to prompt, but the client app requires re-authentication or MFA.';
```
pradeepdewda-okta, line 55: this needs to get translated as well.. 
```
const mfaRequiredMsg = loc('newkey', 'login');
```

src/v2/BaseLoginRouter.js
```diff
@@ -153,7 +153,25 @@ export default Router.extend({
...
+      // simulate an IDX error response
+      error = Object.assign({
+        details: {
+          messages: {
```
pradeepdewda-okta, line 162: I think this should work if you add message like this instead 
```
“messages”: {
    “type”: “array”,
    “value”: [
      {
        “message”: “Any message",
        “i18n”: {
          “key”: “error.feature.disabled”
        },
        “class”: “ERROR”
      }
    ]
  } (
```

pradeepdewda-okta (approved):
one nit. then 🐑

src/v2/BaseLoginRouter.js
haishengwu-okta, line 158: will message always be English even on different locale (backend)

nikhilvenkatraman-okta: This can be problematic to check the english text. Its defined as an [enum](https://github.com/okta/okta-core/blob/master/components/core/tokens/api/src/main/java/com/okta/monolith/platform/tokens/dto/OAuth2ErrorCode.java#L82) in the monolith. The [description is constructed using a locale in context](https://github.com/okta/okta-core/blob/master/components/core/tokens/api/src/main/java/com/okta/monolith/platform/tokens/dto/OAuth2ErrorCode.java#L338) which may result in a non english text.

pradeepdewda-okta: @aarongranick-okta you mentioned over slack not to use this as a check right ?

aarongranick-okta: We can remove the check on description, but then I think we'd want to change the error message to reflect the loss of certainty related to "feature not enabled". It would now be an "access denied" message.

src/v2/BaseLoginRouter.js
aarongranick-okta: OK, so we can use the raw error as message and then provide an i18n key for translation. Should I provide a test for error when i18n key is not set?

src/v2/BaseLoginRouter.js
pradeepdewda-okta: We can check for `error.feature.disabled` key instead.. ?

src/util/OAuth2Util.js
```diff
@@ -44,11 +44,26 @@ util.getTokens = function (settings, params, controller) {
...
+  
+    // user is not assigned to OIDC client
     if (error.errorCode === 'access_denied') {
-      controller.model.trigger('error', controller.model, { responseJSON: error });
+      showError = true;
+    }
+
+    // MFA is required but prompt=none (authn V1)
+    const mfaRequiredMsg = 'The client specified not to prompt, but the client app requires re-authentication or MFA.';
+    if (error.errorCode === 'login_required' && error.errorSummary === mfaRequiredMsg) {
```
jmelberg-okta, line 57: Does this message get returned via our API? If so, this creates a tight coupling between an expected API response (which may be localized by the backend) and the Widget - and will only work for `en` scenarios.

Instead, can we drop the `error.errorSummary` check?

src/v2/BaseLoginRouter.js
```diff
@@ -153,7 +153,27 @@ export default Router.extend({
   },
 
   handleIdxResponseFailure (error = {}) {
-    // 1. loosely check whether is IDX error response
+    // will happen if `useInteractionCodeFlow` is true but the Org is not using OIE
+    if (error?.error === 'access_denied') {
+      // simulate an IDX error response
+      error = Object.assign({
```
jmelberg-okta, line 160: We'll likely need to reuse this error creation logic. Want to move it out to a shared file?

src/widget/OktaSignIn.js
```diff
@@ -142,6 +142,13 @@ var OktaSignIn = (function () {
 
     var authClient = options.authClient ? options.authClient : createAuthClient(authParams);
 
+    // validate authClient configuration against widget options
+    if (options.useInteractionCodeFlow && authClient.isPKCE() === false) {
+      throw new Errors.ConfigError(
+        'The "useInteractionCodeFlow" option requires PKCE to be enabled on the authClient.'
```
jmelberg-okta, line 148: Why don't we do this for the developer automatically?

src/v2/BaseLoginRouter.js
aarongranick-okta: This is a very special case, receiving a "V1" error while operating on an IDX code path. The error says IDX is not enabled. For other errors on this path, it is expected they would be in an IDX format

src/widget/OktaSignIn.js
aarongranick-okta: PKCE is enabled by default. If it is off, the developer has set it so explicitly. This is a config error, should only be seen by developers. Without this extra code, the error would be a runtime error, increasing the possibility that users would run into it because the developer thought it worked (but didn't try to login) then pushed it up.

PR #1666, opened 2021-02-12
build: adds tests to cover IDX_RESPONSE_ENHANCEMENTS_PRE_VERSIONING

swiftone:
## Description:

Adds tests to cover the addition of new `session expired` responses with error status codes

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-361545](https://oktainc.atlassian.net/browse/OKTA-361545)

### DIFF
--- a/playground/mocks/data/idp/idx/error-pre-versioning-ff-session-expired.json
+++ b/playground/mocks/data/idp/idx/error-pre-versioning-ff-session-expired.json
@@ -1,0 +1,15 @@
+{
+  "version":"1.0.0",
+  "messages":{
+    "type":"array",
+    "value":[
+      {
+        "message":"The session has expired.",
+        "i18n":{
+          "key":"idx.session.expired"
+        },
+        "class":"ERROR"
+      }
+    ]
+  }
+}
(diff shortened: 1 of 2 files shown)


aarongranick-okta (changes requested):
Looking good. Could you also add verification that the "cancel" link is rendered? (I think it usually says "Sign Out". This is particularly important in the terminal error view because this is how the user is able to start over (even page refresh may not clear the error state)

aarongranick-okta (approved):

playground/mocks/data/idp/idx/error-pre-versioning-ff-session-expired.json
```diff
@@ -0,0 +1,15 @@
+{
+  "version":"1.0.0",
+  "messages":{
+    "type":"array",
+    "value":[
+      {
+        "message":"The session has expired.",
```
haishengwu-okta, line 7: we have `error-session-expired.json` already. curious what's the difference?

swiftone: That was added by Pradeep to deal with polling.  I don't know much about that one, but when I compared that JSON with the responses I was getting from the server for this case, the non-polling responses did not contain the stateHandle, yet `error-session-expired.json` does, so I created this second more explicit mock json.

PR #1669, opened 2021-02-17
Adds e2e tests for OIE enabled org

vijetmahabaleshwar-okta:
## Description:
- All existing e2e tests are being run against Okta Classic org
- We need the same coverage against OIE enabled org
- This PR addresses this issue by running tests against OIE org in a new test suite

## PR Checklist

- [ ] Added unit tests?

- [x] Added e2e tests

### Issue:

- [OKTA-369030](https://oktainc.atlassian.net/browse/OKTA-369030)


test/e2e/conf.js
```diff
@@ -48,6 +48,13 @@ if (process.env.TRAVIS  || process.env.CHROME_HEADLESS) {
     };
   }
 
+  if (process.env.OIE_TESTS) {
```
vijetmahabaleshwar-okta, line 51: These specs run against okta classic org only.

scripts/e2e-oie.sh
```diff
@@ -0,0 +1,27 @@
...
+export TEST_SUITE_TYPE="junit"
+export TEST_RESULT_FILE_DIR="${REPO}/build2/reports/junit"
+echo $TEST_SUITE_TYPE > $TEST_SUITE_TYPE_FILE
+echo $TEST_RESULT_FILE_DIR > $TEST_RESULT_FILE_DIR_FILE
+
+# This file contains all the env vars we need for e2e tests
+aws s3 --quiet --region us-east-1 cp s3://ci-secret-stash/prod/signinwidget/export-test-credentials.sh $OKTA_HOME/$REPO/scripts/export-test-credentials.sh
+source $OKTA_HOME/$REPO/scripts/export-test-credentials.sh
+
+# We use the below OIE enabled org and clients for OIE tests
```
haishengwu-okta, line 17: Shall we create a new file to s3 and merge those config to it?

test/e2e/page-objects/OIEPrimaryAuthPage.js
```diff
@@ -0,0 +1,35 @@
...
+'use strict';
+
+const FormPage = require('./FormPage');
+
+class OIEPrimaryAuthPage extends FormPage {
+
+  constructor () {
+    super();
+    this.identifierUsernameInput = element(by.css('[id="okta-signin-username"]'));
+    this.credentialPasscodeInput = element(by.css('[id="okta-signin-password"]'));
```
haishengwu-okta, line 21: how come this would work? we don't have those IDs for OIE login form.

haishengwu-okta:
Might be nice to break test into separated sub folder like `test/e2e/legacy`, `test/e2e/oie`

test/e2e/page-objects/OIEPrimaryAuthPage.js
vijetmahabaleshwar-okta: I realized that these tests are running in interop mode i.e. running /v1/authn pipeline against OIE enabled org. So it'll still render the v1 login form. I don't need to add new page objects for this. 🤦 
I still think that we should have the same specs that we run against v1 org, run against an OIE enabled org as well to test backward compatibility.

scripts/e2e-oie.sh
vijetmahabaleshwar-okta: The variables are the same in both the orgs. It's easier to have a single file I think.

test/e2e/page-objects/OIEPrimaryAuthPage.js
haishengwu-okta: Make sense, so we'll have ?
- `/authn` against classical org
- `/authn` against OIE org
- `/idx` against OIE org

vijetmahabaleshwar-okta:
I have updated the PR to run the same specs against an OIE enabled org. (Backward compatibility mode)
Since these tests are running against a self-hosted widget, to test pure OIE features, I will need to add interaction code flag. I'll address that in a subsequent PR.

Though the question I have is should these tests against the OIE enabled org run as a separate bacon test suite or can I run them serially in the existing e2e test suite?

haishengwu-okta:
Separated suite seems nicer!

test/e2e/page-objects/OIEPrimaryAuthPage.js
vijetmahabaleshwar-okta: Correct. I have already filed a bug for the 2nd case (/authn against OIE org) just by running these tests - https://oktainc.atlassian.net/browse/OKTA-375434
Next would be to run a subset of these tests + new E2E tests against /idx endpoints.

scripts/e2e-oie.sh
haishengwu-okta: I meant those 3 variables below. they shall be different with non-OIE org?

vijetmahabaleshwar-okta: We can't merge both into the same config, because these 3 variables would be different for OIE and non-OIE orgs.
I can add these in a separate config file in S3 but these variables are not secret. Do we need them in S3?

haishengwu-okta: Feel it's better to in s3 file as well, just be consistent?

test/e2e/specs/dev_spec.js
```diff
@@ -39,10 +39,15 @@ describe('Dev Mode flows', function () {
     util.loadTestPage('basic-dev');
   });
 
+  afterEach(function () {
+    // Logout of Okta session
+    browser.get('{{{WIDGET_TEST_SERVER}}}/login/signout');
```
jmelberg-okta, line 44: nit: Consider adding an expectation here to verify we're back on the primary auth page to confirm logging out was successful.

jmelberg-okta (approved):
One suggestion - but overall LGTM!

As a follow-up, we should explore ways to combine our `e2e-*` scripts (i.e. `setup-e2e.sh`) to condense the code needed for new run this new e2e test suite.

PR #1690, opened 2021-02-20
feat: don't rerender SIW when only poll refresh value is updated in remediation

jane-kuang-okta:
## Description:
- Add support for when the poll refresh value changes
- Currently, when the SIW detects any change in the IDX response, the SIW will re-render. However, we want to ensure that we don't rerender the SIW if the only value that has changed in the remediation response is the poll `refresh` value, otherwise the user will see the SIW flickering at each poll call

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
|                        | Video             | Description    |
|--------------| ------------- | ------------- |
| SIW with backend change only  | https://okta.box.com/s/pdlnwsviya43wb5duslsts4a0eequ3f3 | During Custom URI flow when the user clicks the 'Sign in with Okta Verify' button, the SIW will keep rerendering because it detected a change in the IDX response from the server (poll `refresh` value)
| SIW with backend AND SIW changes  | https://okta.box.com/s/2zif6ckmi5fmparz1o59j8fm0eeifv3e | Added logic that excludes `refresh` value when comparing the previous IDX response to the incoming response. Now the SIW does not refresh at every poll call.

### Issue:

- [OKTA-370335](https://oktainc.atlassian.net/browse/OKTA-370335)


src/v2/models/AppState.js
```diff
@@ -105,9 +105,33 @@ export default Model.extend({
...
+    const identicalResponse = _.isEqual(_.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'refresh']),
+      _.nestedOmit(previousRawState, ['expiresAt', 'refresh']));
+    const noPollRefreshNeeded = _.isEqual(_.omit(transformedResponse.idx.rawIdxState, 'expiresAt'),
       _.omit(previousRawState, 'expiresAt') );
+
+    if (identicalResponse && !noPollRefreshNeeded) {
+      const currentFormName = this.get('currentFormName');
+      const currentViewState = transformedResponse.remediations.filter(r => r.name === currentFormName)[0];
+
+      this.trigger('refreshUpdated', currentViewState);
```
pradeepdewda-okta, line 132: I don't remember the reason to trigger an event from here ? can't be updating `reRender` to `false` is suffice ?

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -24,7 +25,7 @@ const request = (opts) => {
   return $.ajax(ajaxOptions);
 };
 
-const Body = BaseForm.extend(Object.assign(
+const Body = BaseFormWithPolling.extend(Object.assign(
```
pradeepdewda-okta, line 28: Also, I don't think we need a baseForm just for polling.. BaseForm should be able to handle this as well..

src/v2/models/AppState.js
```diff
@@ -105,9 +105,33 @@ export default Model.extend({
   },
 
   shouldReRenderView (transformedResponse) {
+    _.mixin({
+      nestedOmit: function (obj, iteratee, context) {
```
pradeepdewda-okta, line 109: this is neat 👍

src/v2/models/AppState.js
```diff
@@ -105,9 +105,33 @@ export default Model.extend({
   },
 
   shouldReRenderView (transformedResponse) {
+    _.mixin({
```
pradeepdewda-okta, line 108: create a folder under V2 called `mixins` and add a file `mixins.js` and add this function their and use it here.

src/v2/models/AppState.js
```diff
@@ -105,9 +105,33 @@ export default Model.extend({
...
+
+        return result;
+      }
+    });
+
     const previousRawState = this.has('idx') ? this.get('idx').rawIdxState : null;
-    const identicalResponse = _.isEqual(_.omit(transformedResponse.idx.rawIdxState, 'expiresAt'),
+    const identicalResponse = _.isEqual(_.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'refresh']),
+      _.nestedOmit(previousRawState, ['expiresAt', 'refresh']));
+    const noPollRefreshNeeded = _.isEqual(_.omit(transformedResponse.idx.rawIdxState, 'expiresAt'),
```
pradeepdewda-okta, line 125: i am not sure why this needed.. `noPollRefreshNeeded`

src/v2/models/AppState.js
jane-kuang-okta: Since the SIW will not rerender with these new nested omit changes, the SIW won't be updating the polling interval value anymore in DeviceChallengePollView, meaning it'll keep polling the server at the initial fast `refresh` interval. So, this event is used to trigger that update.

src/v2/view-builder/views/DeviceChallengePollView.js
jane-kuang-okta: I originally had the listener in the `initialize()` function in BaseForm, but per @haishengwu-okta 's suggestion, that's not ideal because that logic is only applicable to certain views. Wasn't 100% sure where to put this logic exactly, so I'm open to any suggestions.

src/v2/models/AppState.js
jane-kuang-okta: The point of having `noPollRefreshNeeded` was to trigger the update polling interval event if **_only_** `refresh` in the IDX response changed. `if (identicalResponse && !noPollRefreshNeeded)` means only refresh value changed. 

If we use the "Forgot Password" flow and we poll for email as an example, that interval is a steady 4000ms. In that case `identicalResponse` and `noPollRefreshNeeded` are true -- then we don't need to enter the new event triggering logic.

src/v2/models/AppState.js
pradeepdewda-okta: okay that make sense... Let's try another approach instead.. 
1. add a local boolean attribute in line 23 as `dynamicRefresh`
2. set it's value to true/false accordingly in `shouldReRenderView` 
3. Listen to change `change` event on this property in your DeviceChallengePollView and update the interval

src/v2/models/AppState.js
pradeepdewda-okta: `noPollRefreshNeeded` is confusing, possible change the name to `isDynamicRefreshInterval`

src/v2/view-builder/views/DeviceChallengePollView.js
```diff
@@ -43,13 +43,23 @@ const Body = BaseForm.extend(Object.assign(
...
     },
 
     onPollingFail () {
       this.$('.spinner').hide();
       this.stopPolling();
     },
 
+    updateRefresh () {
+      if (this.polling) {
+        this.options.currentViewState = this.options.appState.get('dynamicRefresh');
```
haishengwu-okta, line 57: Changing the `currentViewState` this way a bit _hacky_.
I was more leaning toward to only pass around the new refresh interval; then change the `startPolling` function
like
```
this.startPolling(newRefresh);

-----
  startPolling (newInterval) {
    this.pollingInterval = newInterval  || this.options.currentViewState.refresh;
  ...
}
```

what do you think @pradeepdewda-okta @jane-kuang-okta

src/v2/view-builder/views/DeviceChallengePollView.js
haishengwu-okta: Either leave to BaseForm or BaseFormWithPolling.
The reason I'd suggest `BaseFormWithPolling` is polling is only meant for several views and doesn't quite make sense to blow up BaseForm this way.

haishengwu-okta (commented):
Looks good in general! 2 questions around Class heirachy and data structure.

src/v2/view-builder/views/DeviceChallengePollView.js
pradeepdewda-okta: With her new implementation we don't even need `BaseFormWithPolling` at all.. Concern view can look for change in value for `dynamicRefresh` attribute..

haishengwu-okta: Other view will need this soon, like email polling.

haishengwu-okta:
```
import polling from './shared/polling';

const BaseFormWithPolling = BaseForm.extend({
   initialize() {
     // where can listen to dynamic refresh event
  },
...polling
});

----------

// using `polling` can be removed from `DeviceChallengePollView.Body` now.
DeviceChallengePollView.Body = BaseFormWithPolling.extend({...});

```

Potentially, all view (its body) using `polling` today can extend from BaseFormWithPolling.

There is general issue around inherent v.s. composition (cause no support multiple parent class) but we can worry it about later.

src/v2/models/AppState.js
```diff
@@ -27,6 +28,7 @@ export default Model.extend({
     currentFormName: 'string',
     idx: 'object',
     remediations: 'array',
+    dynamicRefresh: 'object',
```
haishengwu-okta, line 31: `number` ?

src/v2/view-builder/views/DeviceChallengePollView.js
jane-kuang-okta: In that case, I think @pradeepdewda-okta 's argument would be we can just listen for `change:dynamicRefresh` in the email polling view. I do think `BaseFormWithPolling` would help with redundancy

src/v2/view-builder/internals/BaseFormWithPolling.js
```diff
@@ -0,0 +1,19 @@
...
+import BaseForm from './BaseForm';
+
+export default BaseForm.extend(Object.assign(
+  {
+    initialize () {
+      BaseForm.prototype.initialize.apply(this, arguments);
+      this.listenTo(this.options.appState, 'change:dynamicRefreshInterval', this.updateRefresh);
+    },
+
+    updateRefresh () {
```
pradeepdewda-okta, line 11: change it to more specific `updateRefreshInterval`

src/v2/view-builder/views/DeviceChallengePollView.js
pradeepdewda-okta: I am good with this change, make sense to separate it out and create a reusable class. We can refactor other views later. Can we change the name to `BaseFormPolling.js`?

pradeepdewda-okta (approved):
Overall i am good with the changes. Make sure all testcafe tests passes, in case of failure you may need to update them.

src/v2/view-builder/views/DeviceChallengePollView.js
jane-kuang-okta: FYI, will be keeping as is in latest commit. ^Comment above was based on outdated code.

haishengwu-okta (approved):
🎖️ Thanks!

src/v2/view-builder/views/shared/polling.js
```diff
@@ -2,8 +2,8 @@ import { _ } from 'okta';
 import { MS_PER_SEC } from '../../utils/Constants';
 
 export default {
-  startPolling () {
-    this.pollingInterval = this.options.currentViewState.refresh;
+  startPolling (newInterval) {
+    this.pollingInterval = newInterval || this.options.currentViewState.refresh;
```
haishengwu-okta, line 6: A further refactoring I can see is always set the interval to `dynamicRefreshInterval`.
1. `in the BaseFormPolling.initialize`, `this.appState.set('dynamicRefreshInterval', .., {silent: true})`
2. then in this method, only read from appState.
3. also need to refractor other views that using polling.

just FYI now. In case you'll be working on this piece of code again, we can try to do it.

PR #1691, opened 2021-02-22
Adds a11y fixes for push button

jmelberg-okta:
## Description:

Fixes for accessibility issues where text rendered inside of the "Push" verification button is not read aloud for screen readers.

## PR Checklist

- [x] Added unit tests?
- [x] Added e2e tests
- [x] [Bacon link](https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&sha=84f05ff48377a9936d725b847f23160139d7de6c)

### Screenshot/Video:

| Flow | Video |
| ---- | ----- |
| Classic | Classic |
| OIE | OIE |

### Reviewers:

- @edburyenegren-okta 
- @mauriciocastillosilva-okta 

### Issue:

- [OKTA-362496](https://oktainc.atlassian.net/browse/OKTA-362496)

### DIFF
--- a/src/v2/view-builder/views/ov/ChallengeOktaVerifyPushView.js
+++ b/src/v2/view-builder/views/ov/ChallengeOktaVerifyPushView.js
@@ -36,6 +36,9 @@ const Body = BaseForm.extend(Object.assign(
           e.preventDefault();
         }
       }));
+      this.add(
+        `<span class='accessibility-text' role='alert'>${loc('oie.okta_verify.push.sent', 'login')}</span>`,
+      );
     },
 
     postRender () {
--- a/src/views/mfa-verify/PushForm.js
+++ b/src/views/mfa-verify/PushForm.js
@@ -76,16 +76,22 @@ export default Form.extend({
   },
   setSubmitState: function (ableToSubmit) {
     const button = this.$el.find('.button');
-
+    const a11ySpan = this.$el.find('.accessibility-text');
     this.enabled = ableToSubmit;
     if (ableToSubmit) {
       button.removeClass('link-button-disabled');
       button.prop('value', loc('oktaverify.send', 'login'));
       button.prop('disabled', false);
+      if (a11ySpan) {
+        a11ySpan.remove();
+      }
     } else {
       button.addClass('link-button-disabled');
       button.prop('value', loc('oktaverify.sent', 'login'));
       button.prop('disabled', true);
+      this.add(
+        `<span class='accessibility-text' role='alert'>${loc('oktaverify.sent', 'login')}</span>`,
+      );
     }
   },
   submit: function (e) {
--- a/test/testcafe/spec/ChallengeOktaVerifyPush_spec.js
+++ b/test/testcafe/spec/ChallengeOktaVerifyPush_spec.js
@@ -52,7 +52,9 @@ test
 
     const pageTitle = challengeOktaVerifyPushPageObject.getPageTitle();
     const pushBtn = challengeOktaVerifyPushPageObject.getPushButton();
+    const a11ySpan = challengeOktaVerifyPushPageObject.getA11ySpan();
     await t.expect(pushBtn.textContent).contains('Push notification sent');
+    await t.expect(a11ySpan.textContent).contains('Push notification sent');
     await t.expect(pushBtn.hasClass('link-button-disabled')).ok();
     await t.expect(pageTitle).contains('Get a push notification');
   });
--- a/test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
+++ b/test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
@@ -13,6 +13,10 @@ export default class ChallengeOktaVerifyPushPageObject extends ChallengeFactorPa
     return this.form.getElement('.send-push');
   }
 
+  getA11ySpan () {
+    return this.form.getElement('.accessibility-text');
+  }
+
   getResendPushButton () {
     return this.form.getElement('.button-primary');
   }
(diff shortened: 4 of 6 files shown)


src/v2/view-builder/views/ov/ChallengeOktaVerifyPushView.js
```diff
@@ -36,6 +36,9 @@ const Body = BaseForm.extend(Object.assign(
           e.preventDefault();
         }
       }));
+      this.add(
```
jmelberg-okta, line 39: With OIE flows, this label is **always** disabled as the push is sent **automatically**. When we encounter an error state (i.e. the user rejects the notification) - this button is re-enabled with text matching the state.

src/views/mfa-verify/PushForm.js
```diff
@@ -76,16 +76,22 @@ export default Form.extend({
...
       button.prop('disabled', false);
+      if (a11ySpan) {
+        a11ySpan.remove();
+      }
     } else {
       button.addClass('link-button-disabled');
       button.prop('value', loc('oktaverify.sent', 'login'));
       button.prop('disabled', true);
+      this.add(
+        `<span class='accessibility-text' role='alert'>${loc('oktaverify.sent', 'login')}</span>`,
```
jmelberg-okta, line 93: Using `role="alert"` isn't ideal, but since the user interacts with this button with the intent to perform an action - it seemed appropriate.

@edburyenegren-okta - Think this is OK?

src/v2/view-builder/views/ov/ChallengeOktaVerifyPushView.js
haishengwu-okta: Just for my curiosity, isn't there a way to tell screen reader that there is disable button and reads loud the text?
Adding extra element for a11y seems hacky (I'm very new to a11y world. :D ) ?

jmelberg-okta: Good question! As a user navigates the page using a reader, a `disabled` button will never receive focus. As a result, all text within a `disabled` state (input, button, etc) will not be read aloud.

There are a few approaches to solve this problem:

1. Utilize a hidden element with an `alert` or `notify` role (what we're doing here). While it may seem "hacky" - this forces a screen reader to see updated text and read it aloud.
2. Style inputs/buttons accordingly without marking them as `disabled`. There is a great [article here](https://medium.com/@izzmo/anti-pattern-using-disabled-or-read-only-for-form-input-e2c24c669f5b) that goes a bit deeper into the accessibility concerns with marking elements as `disabled`. This option nests too many changes and creates more of an opportunity to release a regression.
3. Move visual messaging out of buttons (Odyssey approach). There are valid use cases where users should be informed of an action taking place after interacting with a page. For this scenario, a banner or toast could be leveraged to notify the user about the sent push notification. When we tackle a11y with our redesign/rewrite on Odyssey, we'll explore a more suitable UX.

src/views/mfa-verify/PushForm.js
edburyenegren-okta: I think so! The push response is time-sensitive, so it's best if they hear the trigger asap. Great work here, @jmelberg-okta

aarongranick-okta (approved):

src/v2/view-builder/views/ov/ChallengeOktaVerifyPushView.js
mauriciocastillosilva-okta: Do we need to remove it if the user rejects the push or if it times out?

test/testcafe/spec/ChallengeOktaVerifyPush_spec.js
```diff
@@ -52,7 +52,9 @@ test
 
     const pageTitle = challengeOktaVerifyPushPageObject.getPageTitle();
     const pushBtn = challengeOktaVerifyPushPageObject.getPushButton();
+    const a11ySpan = challengeOktaVerifyPushPageObject.getA11ySpan();
     await t.expect(pushBtn.textContent).contains('Push notification sent');
+    await t.expect(a11ySpan.textContent).contains('Push notification sent');
     await t.expect(pushBtn.hasClass('link-button-disabled')).ok();
     await t.expect(pageTitle).contains('Get a push notification');
   });
```
mauriciocastillosilva-okta, line 60: test below checks for the button to not be disabled and to have a different text. Maybe we should test also that `a11ySpan` is removed? I am not as familiar with how OV works in v2/idx, but looking at what you are doing for v1 it seems we have something similar

src/v2/view-builder/views/ov/ChallengeOktaVerifyPushView.js
jmelberg-okta: @mauriciocastillosilva-okta - We don't need to remove the new element, as rejecting the push will render the `ChallengeOktaVerifyResendPushView` view and the time-out flow does not enable the button. As a result, this a11y hidden element can remain present in the view as long as we keep this behavior. 😄

test/testcafe/spec/ChallengeOktaVerifyPush_spec.js
jmelberg-okta: I thought about adding this test as well, but decided not to because that test is using a different view (more information in [this comment](https://github.com/okta/okta-signin-widget/pull/1691/files/af4fcc6b7407ac7675a97c5bd84d1c2b5f17320f#r582129603)). Since v2 uses multiple views for this scenario, I think we're OK testing only the disabled button state.

mauriciocastillosilva-okta (approved):
looks good

jmelberg-okta:
Closing this PR manually, as a hook failed. This change has been merged into master as a part of [this commit](https://github.com/okta/okta-signin-widget/commit/e06b7e405d173a4899304025458d9f20e335e26c).

PR #1697, opened 2021-02-25
adds new build tool: update-readme

aarongranick-okta:
## Description:
New buildtool command to update the CDN URLs in README.md.

It uses yarn to find the last published version of the widget, then updates matching URLs in README

```
node buildtools update-readme
```

Also includes updated README

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-304593](https://oktainc.atlassian.net/browse/OKTA-304593)


buildtools/commands/update-readme.js
```diff
@@ -0,0 +1,28 @@
+const { resolve, join } = require('path');
+const shell = require('shelljs');
+const replace = require('replace-in-file');
+
+const ROOT_DIR = resolve(__dirname, '../../');
+
+function getPublishedWidgetVersion () {
+  const stdout = shell.exec('yarn info @okta/okta-signin-widget versions', { silent: true });
```
haishengwu-okta, line 8: would be easier to check 
```
'dist-tags': {
    latest: '5.3.3',
    legacy: '5.2.4'
  },

```

from `yarn info @okta/okta-signin-widget` ?

buildtools/commands/update-readme.js
```diff
@@ -0,0 +1,28 @@
...
+  const arrayStr = stdout.substring(stdout.indexOf('['), stdout.lastIndexOf(']') + 1).replace(/'/g, '"');
```
swiftone, line 9: Why not use the --json option to `yarn info`?

buildtools/commands/update-readme.js
haishengwu-okta: `npm view ..` gives similar lighter result

buildtools/commands/update-readme.js
```diff
@@ -0,0 +1,28 @@
...
+  const versions = JSON.parse(arrayStr);
+  const version = versions[versions.length - 1];
+  console.log('Last published okta-signin-widget version: ', version);
+  return version;
+}
+
+exports.command = 'update-readme';
+exports.describe = 'Updates CDN URLs in the README file';
+exports.handler = () => {
+  console.log('updating readme..');
```
swiftone, line 19: nit: 3 dots for an incomplete ellipsis, 4 for a complete one.

buildtools/commands/update-readme.js
aarongranick-okta:
```
yarn info @okta/okta-signin-widget versions --json
{"type":"warning","data":"../package.json: No license field"}
{"type":"inspect","data":["1.4.0","1.5.0","1.6.0-rc.0","1.6.0","1.7.0","1.8.0","1.9.0","1.11.0","1.12.0","1.13.0","2.1.0","2.3.0","2.4.0","2.5.0","2.6.0","2.7.0","2.8.0","2.9.0","2.10.0","2.11.0","2.12.0","2.13.0","2.14.0","2.15.0","2.16.0","2.17.0","2.18.0","2.19.0","2.20.0","2.20.1","2.21.0","3.0.0","3.0.1","3.0.2","3.0.3","3.0.4","3.0.5","3.0.6","3.1.0","3.1.1","3.1.2","3.1.3","3.1.4","3.1.5","3.1.6","3.2.0","3.2.1","3.2.2","3.3.0","3.3.1","3.3.2","3.3.3","3.4.0","3.4.1","3.4.2","3.5.0","3.6.0","3.6.1","3.6.2","3.7.0","3.7.1","3.7.2","3.7.3","3.8.0","3.8.1","3.8.2","3.9.0","3.9.1","3.9.2","4.0.0","4.0.1","4.0.2","4.1.0","4.1.1","4.1.2","4.1.3","4.1.4","4.1.5","4.2.0","4.2.1","4.2.2","4.2.3","4.2.4","4.2.5","4.2.6","4.2.7","4.3.0","4.3.1","4.3.2","4.3.3","4.3.4","4.3.5","4.4.0","4.4.1","4.4.2","4.4.3","4.4.4","4.5.0","4.5.1","4.5.2","5.0.0","5.0.1","5.0.2","5.1.0","5.1.1","5.1.2","5.1.3","5.1.4","5.1.5","5.2.0","5.2.1","5.2.2","5.2.3","5.2.4","5.3.0","5.3.1","5.3.2","5.3.3"]}
```
Due to the "no license field" two JSON objects are actually returned. So it's not any easier to work with. For comparison the non --json output looks like:

```
[
  '1.4.0',
  '1.5.0',
  '1.6.0-rc.0',
  '1.6.0',
  '1.7.0',
  '1.8.0',
  '1.9.0',
  '1.11.0',
  '1.12.0',
  '1.13.0',
  '2.1.0',
  '2.3.0',
  '2.4.0',
  '2.5.0',
  '2.6.0',
  '2.7.0',
  '2.8.0',
  '2.9.0',
  '2.10.0',
  '2.11.0',
  '2.12.0',
  '2.13.0',
  '2.14.0',
  '2.15.0',
  '2.16.0',
  '2.17.0',
  '2.18.0',
  '2.19.0',
  '2.20.0',
  '2.20.1',
  '2.21.0',
  '3.0.0',
  '3.0.1',
  '3.0.2',
  '3.0.3',
  '3.0.4',
  '3.0.5',
  '3.0.6',
  '3.1.0',
  '3.1.1',
  '3.1.2',
  '3.1.3',
  '3.1.4',
  '3.1.5',
  '3.1.6',
  '3.2.0',
  '3.2.1',
  '3.2.2',
  '3.3.0',
  '3.3.1',
  '3.3.2',
  '3.3.3',
  '3.4.0',
  '3.4.1',
  '3.4.2',
  '3.5.0',
  '3.6.0',
  '3.6.1',
  '3.6.2',
  '3.7.0',
  '3.7.1',
  '3.7.2',
  '3.7.3',
  '3.8.0',
  '3.8.1',
  '3.8.2',
  '3.9.0',
  '3.9.1',
  '3.9.2',
  '4.0.0',
  '4.0.1',
  '4.0.2',
  '4.1.0',
  '4.1.1',
  '4.1.2',
  '4.1.3',
  '4.1.4',
  '4.1.5',
  '4.2.0',
  '4.2.1',
  '4.2.2',
  '4.2.3',
  '4.2.4',
  '4.2.5',
  '4.2.6',
  '4.2.7',
  '4.3.0',
  '4.3.1',
  '4.3.2',
  '4.3.3',
  '4.3.4',
  '4.3.5',
  '4.4.0',
  '4.4.1',
  '4.4.2',
  '4.4.3',
  '4.4.4',
  '4.5.0',
  '4.5.1',
  '4.5.2',
  '5.0.0',
  '5.0.1',
  '5.0.2',
  '5.1.0',
  '5.1.1',
  '5.1.2',
  '5.1.3',
  '5.1.4',
  '5.1.5',
  '5.2.0',
  '5.2.1',
  '5.2.2',
  '5.2.3',
  '5.2.4',
  '5.3.0',
  '5.3.1',
  '5.3.2',
  '5.3.3',
  '5.4.0'
]
```

buildtools/commands/update-readme.js
aarongranick-okta: The output of  `yarn info @okta/okta-signin-widget versions` comes out like this:

```
[
  '1.4.0',
  '1.5.0',
 <snip>
  '3.0.4',
  '5.3.1',
  '5.3.2',
  '5.3.3',
  '5.4.0'
]
```
so is it is easy to parse it as an array and grab the last entry. 

Are you thinking of a case where we want to tag an older version as latest because of a bad release?

buildtools/commands/update-readme.js
swiftone: That's....interesting.  It doesn't on my machine, good to know, I guess.

buildtools/commands/update-readme.js
haishengwu-okta: Could be. and I don't trust the Array :D instead of more **explicit** meta data like what I post above.

package.json
```diff
@@ -127,16 +127,18 @@
...
+    "shelljs": "^0.8.4",
     "source-map-loader": "^0.2.4",
     "stylelint": "^10.1.0",
     "stylelint-checkstyle-formatter": "^0.1.2",
     "stylelint-scss": "^3.9.2",
     "testcafe": "^1.4.0",
     "time-grunt": "^2.0.0",
     "wait-on": "^3.3.0",
-    "webdriver-manager": "^12.0.6",
+    "webdriver-manager": "^12.1.8",
```
nikhilvenkatraman-okta, line 141: Can we add a yarn command for it in the scripts ?

buildtools/commands/update-readme.js
```diff
@@ -0,0 +1,27 @@
...
+  const versions = JSON.parse(arrayStr);
+  const version = versions[versions.length - 1];
+  console.log('Last published okta-signin-widget version: ', version);
+  return version;
+}
+
+exports.command = 'update-readme';
+exports.describe = 'Updates CDN URLs in the README file';
+exports.handler = () => {
+  const version = getPublishedWidgetVersion();
```
jmelberg-okta, line 19: Been awhile since I've dug into the publish code, but why can't we reference the `package.json` version?

Since this build tool is intended to hook into our promotion flow, I'd assume that when the artifact is promoted the `version` can be treated as our source of truth.

This would also mitigate any delays in `npm publish` where the artifact may not be present at the time this script is executed.

buildtools/commands/update-readme.js
aarongranick-okta: turns out the warning is going to stderr, so there is only one JSON object. switching to --json

buildtools/commands/update-readme.js
aarongranick-okta: It would be nice to fit this into our promotion flow, but we also need to be able to manually update it at anytime. 

The package.json can only be considered a "source of truth" for this CDN url during a very narrow of slice of time, which you identified: after promote, but before version bump. And at that time the conductor release train has not been done yet so technically the URL would be invalid. If we wired it up to execute at that stage, then I think it would make sense for the script to accept a widget version as a command line argument. If the caller has a superior source of truth then it may provide the value.

Since the README update should link to valid URLs, it probably makes more sense to wire it up in the conductor release train so it runs after the CDN deployment. We will have to be careful if it is passing an argument and overriding the "latest" dist tag from NPM because sometimes we often release trains to patch older versions, and these should not update the README (or the "latest" dist tag).

nikhilvenkatraman-okta (approved):
🚀

haishengwu-okta (approved):

PR #1699, opened 2021-02-26
refactor: Move device authenticator views and refactor internals

jmelberg-okta:
## Description:

This is a refactoring commit that changes file locations and imports. Specifically:

- All device authenticator views for enroll and challenge are migrated into a `device` directory. This adheres to our new convention for authenticators.
- Allows `internals` to be imported using destructured ES6 named exports.

## PR Checklist

### Reviewers:

- @haishengwu-okta

### DIFF
--- a/src/v2/view-builder/internals/index.js
+++ b/src/v2/view-builder/internals/index.js
@@ -1,17 +1,7 @@
-import BaseHeader from './BaseHeader';
-import BaseFooter from './BaseFooter';
-import BaseForm from './BaseForm';
-import BaseFormWithPolling from './BaseFormWithPolling';
-import BaseModel from './BaseModel';
-import BaseView from './BaseView';
-import FormInputFactory from './FormInputFactory';
-
-export const { addCustomButton, createCustomButtons, createIdpButtons } = FormInputFactory;
-export {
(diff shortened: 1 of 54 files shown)


src/v2/view-builder/internals/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+export const { addCustomButton, createCustomButtons, createIdpButtons } = FormInputFactory;
+export {
+  BaseHeader,
+  BaseFooter,
+  BaseForm,
+  BaseFormWithPolling,
+  BaseModel,
+  BaseView,
+};
```
anipendakur-okta, line 17: Isn't exporting individually (not as a whole object) preferred in order to support tree-shaking (if we were to)?

So something like this is what I was thinking -
```
export { default as BaseHeader } from './BaseHeader';
export { default as BaseFooter } from './BaseFooter';
.
.
.
```

jmelberg-okta: Great suggestion! Completely overlooked the ability to do this. Will update!

pradeepdewda-okta:
This is nice 🥇

haishengwu-okta (approved):

PR #1704, opened 2021-03-01
adds jest for unit tests

aarongranick-okta:
## Description:
- adds a jest config and helper
- adds a config for debugging in VSCode
- runs all v2 + OktaSignIn tests using jest
- all tests still being run by karma as well

The V2 tests were very minimally modified so they can run in jest AND karma.  

The expectation is that most new unit tests will be written in jest, but also that existing unit tests will need to be maintained. We now have a choice of leaving these tests in Karma or migrating them to Jest with minimal changes.

### Issue:

- [OKTA-372526](https://oktainc.atlassian.net/browse/OKTA-372526)


aarongranick-okta:
Screen Shot 2021-03-01 at 12 21 33 PM

jest.config.js
```diff
@@ -0,0 +1,54 @@
...
+    '^jquery$': `${LOCAL_PACKAGES}/@okta/courage-dist/jquery.js`,
+    '^qtip$': '@okta/qtip2/dist/jquery.qtip.min.js',
+    '^duo$': 'duo_web_sdk/index.js',
+    '^typingdna$': 'TypingDnaRecorder-JavaScript/typingdna',
+  },
+  setupFiles: [
+    '<rootDir>/test/unit/jest/jest-setup-global.js'
+  ],
+  testMatch: [
+    '**/test/unit/spec/v2/**/*.{js,ts}',
```
haishengwu-okta, line 41: - maybe move tests into `test/jest/` folder?
- same for setup file.

haishengwu-okta (approved):
looks good to me and one question!

jest.config.js
```diff
@@ -0,0 +1,54 @@
+var path = require('path');
```
jmelberg-okta, line 1: Is there anything prohibiting us from favoring `const` over `var` where it is necessary?

jest.config.js
```diff
@@ -0,0 +1,54 @@
...
+  testMatch: [
+    '**/test/unit/spec/v2/**/*.{js,ts}',
+    '**/test/unit/spec/OktaSignIn_spec.js'
+  ],
+  roots: [
+    'test/unit/spec'
+  ],
+  testPathIgnorePatterns: [
+
+  ],
```
jmelberg-okta, line 49: nit: Let's remove the empty configuration.

package.json
```diff
@@ -35,7 +35,7 @@
     "lint:report": "mkdir -p build2 && run-p -c lint:eslint:report lint:grunt:report lint:stylelint:report",
     "test:dev": "grunt assets && karma start",
     "test:coverage": "yarn test --reporters=progress,coverage",
-    "test": "yarn test:dev --single-run",
+    "test": "yarn test:dev --single-run && yarn jest",
```
jmelberg-okta, line 38: Would prefer we separate Jest from our existing test suite, as we're now running Jest **on top** of our existing test suites. This adds extra test time that isn't necessary.

To start - I think splitting them into a separate `test:jest` command (and setup the necessary Bacon script) would be ideal. This allows developers to run jest tests independently.

test/unit/spec/v2/view-builder/views/ChallengeWebauthnView_spec.js
```diff
@@ -11,8 +11,10 @@ import ChallengeWebauthnResponse
   from '../../../../../../playground/mocks/data/idp/idx/authenticator-verification-webauthn.json';
 
 describe('v2/view-builder/views/webauthn/ChallengeWebauthnView', function () {
+  let test;
```
jmelberg-okta, line 14: nit: Consider using the variable `testContext` here to avoid confusion with the often reserved word `test`.

jmelberg-okta (changes requested):
Requesting that we avoid chaining together both test types here. If we don't, we're going to only report issues on the first failed suite before our jest tests are even triggered.

package.json
aarongranick-okta: `yarn jest` will run all the jest tests. We can add an alias `test:jest` if you think that is useful

swiftone: I know I tend to look in package.json for how to run things - far too often the configs are tucked away so I can't expect the simple case to work.

jest.config.js
aarongranick-okta: these tests are using the same mocks and helpers as the karma tests, so I think there's a good argument for leaving them in place.

I added an exclude list to the karma main.js so that tests will not be run twice.

.bacon.yml
```diff
@@ -0,0 +1,50 @@
+test_suites:
+    - name: lint
+      script_path: /root/okta/okta-signin-widget/scripts
+      sort_order: '1'
+      timeout: '60'
```
jmelberg-okta, line 5: Let's modify this to `15`. Any lint checks that take longer than that imply a serious issue.

.bacon.yml
```diff
@@ -0,0 +1,50 @@
...
+      script_name: lint
+      criteria: MERGE
+      queue_name: small
+    - name: unit-karma
+      script_path: /root/okta/okta-signin-widget/scripts
+      sort_order: '2'
+      timeout: '10'
```
jmelberg-okta, line 12: Looking at previous `main` merges, I _think_ this value for our `unit-*` tests will be OK. We may want to bump it to **15** just to be safe.

.bacon.yml
```diff
@@ -0,0 +1,50 @@
...
+      script_path: /root/okta/okta-signin-widget/scripts
+      sort_order: '3'
+      timeout: '10'
+      script_name: jest
+      criteria: MERGE
+      queue_name: small
+    - name: testcafe
+      script_path: /root/okta/okta-signin-widget/scripts
+      sort_order: '4'
+      timeout: '10'
```
jmelberg-okta, line 26: We'll want to bump these up a bit, as these tests take longer than the timeout specified here.

jmelberg-okta (approved):
Thanks for making the requested changes! Looks great once we update the timeouts!

aarongranick-okta:
@vijetmahabaleshwar-okta FYI bacon scripts on this are now controlled by local `.bacon.yml` file

PR #1707, opened 2021-03-01
feat: Adds Symantec VIP authenticator to OIE

jmelberg-okta:
## Description:

- Adds Symantec VIP enrollment and verification on the OIE pipeline

### Screenshot/Video:

| Enrollment | Verification |
| ---- | ------ |
|  Enrollment | Verification |

### Issue:

- [OKTA-364505](https://oktainc.atlassian.net/browse/OKTA-374394)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -857,11 +857,19 @@ oie.duo.iFrameError = Error loading Duo. Try again or contact your admin for ass
 
 ## IdP Authenticator
 # {0} refers to the display name of the authenticator for all IdP Authenticator values
-oie.idp.authenticator.description = Redirect to verify with {0}
+oie.idp.authenticator.description = Redirect to verify with {0}.
 oie.idp.enroll.title = Set up {0}
 oie.idp.enroll.description = Clicking below will redirect to enrollment in {0}
-oie.idp.challenge.title=  Verify with {0}
+oie.idp.challenge.title = Verify with {0}
 oie.idp.challenge.description = You will be redirected to verify with {0}
+
+## Symantec VIP Authenticator
+# {0} refers to the display name of the authenticator "Symantec VIP"
+oie.symantecVip.authenticator.description = Verify by entering a temporary code from the {0} app.
+oie.symantecVip.enroll.title = Set up {0}
+oie.symantecVip.enroll.description = From the {0} app, enter your credential ID and two consecutive generated codes
+oie.symantecVip.challenge.title = Verify with {0}
+oie.symantecVip.challenge.description = Enter the generated security code from the {0} app.
 
 ## Select authenticator enrollment
 oie.select.authenticators.enroll.title = Set up Authenticators
--- a/src/v2/ion/ViewClassNamesFactory.js
+++ b/src/v2/ion/ViewClassNamesFactory.js
@@ -24,6 +24,7 @@ const FORMNAME_CLASSNAME_MAPPINGS = {
     [AUTHENTICATOR_KEY.OV]: 'mfa-verify',
     [AUTHENTICATOR_KEY.GOOGLE_OTP]: 'mfa-verify',
     [AUTHENTICATOR_KEY.DUO]: 'mfa-verify-duo',
+    [AUTHENTICATOR_KEY.SYMANTEC_VIP]: 'mfa-symantec',
   },
   [FORMS.CHALLENGE_POLL]: {
     [AUTHENTICATOR_KEY.OV]: 'mfa-verify',
@@ -41,6 +42,7 @@ const FORMNAME_CLASSNAME_MAPPINGS = {
     [AUTHENTICATOR_KEY.ON_PREM]: 'enroll-onprem',
     [AUTHENTICATOR_KEY.RSA]: 'enroll-rsa',
     [AUTHENTICATOR_KEY.DUO]: 'enroll-duo',
+    [AUTHENTICATOR_KEY.SYMANTEC_VIP]: 'enroll-symantec',
   },
 
   [FORMS.SELECT_AUTHENTICATOR_ENROLL]: {
--- a/src/v2/view-builder/utils/AuthenticatorUtil.js
+++ b/src/v2/view-builder/utils/AuthenticatorUtil.js
@@ -133,13 +133,10 @@ const getAuthenticatorData = function (authenticator, isVerifyAuthenticator) {
 
   case AUTHENTICATOR_KEY.IDP: {
     const idpName =  authenticator.relatesTo?.displayName;
-    const factorDescription = idpName
-      ? loc('oie.idp.authenticator.description', 'login', [idpName])
-      : loc('oie.idp.authenticator.description.default', 'login');
     Object.assign(authenticatorData, {
       description: isVerifyAuthenticator
         ? ''
-        : factorDescription,
+        : loc('oie.idp.authenticator.description', 'login', [idpName]),
       iconClassName: 'mfa-custom-factor',
       buttonDataSeAttr: getButtonDataSeAttr(authenticator),
     });
@@ -151,6 +148,18 @@ const getAuthenticatorData = function (authenticator, isVerifyAuthenticator) {
         ? ''
         : loc('oie.custom_otp.description', 'login'),
       iconClassName: 'mfa-hotp',
+      buttonDataSeAttr: getButtonDataSeAttr(authenticator),
+    });
+    break;
+  }
+
+  case AUTHENTICATOR_KEY.SYMANTEC_VIP: {
+    const appName =  authenticator.relatesTo?.displayName;
+    Object.assign(authenticatorData, {
+      description: isVerifyAuthenticator
+        ? ''
+        : loc('oie.symantecVip.authenticator.description', 'login', [appName]),
+      iconClassName: 'mfa-symantec',
       buttonDataSeAttr: getButtonDataSeAttr(authenticator),
     });
     break;
--- a/src/v2/view-builder/views/symantec/AuthenticatorSymantecView.js
+++ b/src/v2/view-builder/views/symantec/AuthenticatorSymantecView.js
@@ -1,0 +1,46 @@
+import { _, loc } from 'okta';
+import BaseForm from '../../internals/BaseForm';
+import BaseAuthenticatorView from '../../components/BaseAuthenticatorView';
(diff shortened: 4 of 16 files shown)


src/v2/view-builder/utils/AuthenticatorUtil.js
```diff
@@ -133,13 +133,10 @@ const getAuthenticatorData = function (authenticator, isVerifyAuthenticator) {
...
   case AUTHENTICATOR_KEY.IDP: {
     const idpName =  authenticator.relatesTo?.displayName;
-    const factorDescription = idpName
-      ? loc('oie.idp.authenticator.description', 'login', [idpName])
-      : loc('oie.idp.authenticator.description.default', 'login');
     Object.assign(authenticatorData, {
       description: isVerifyAuthenticator
         ? ''
-        : factorDescription,
+        : loc('oie.idp.authenticator.description', 'login', [idpName]),
```
pradeepdewda-okta, line 139: is the other string getting deleted ? `oie.idp.authenticator.description.default`..

src/v2/view-builder/views/symantec/AuthenticatorSymantecView.js
```diff
@@ -0,0 +1,46 @@
...
+  initialize () {
+    BaseAuthenticatorView.prototype.initialize.apply(this, arguments);
+
+    const currentAuthenticator = this.options.appState.get('currentAuthenticator');
+    const authenticatorEnrollments = this.options.appState.get('authenticatorEnrollments')?.value || [];
+    const existingAuthenticator = _.find(authenticatorEnrollments, { id: currentAuthenticator.id });
+
+    // If the authenticator is already enrolled, we're in a challenge flow as opposed to enrollment
+    this.options.isChallenge = existingAuthenticator || false;
+    this.options.displayName = currentAuthenticator.displayName;
```
jmelberg-okta, line 39: Inferring whether we're in a challenge flow or not by checking the API response is **super valuable**, as we're able to remove duplicate view code. For most authenticators, all this logic could be extracted into the `BaseAuthenticatorView`.

This has a few benefits:

1. Assuming we move toward utilizing the `displayName` over a hardcoded name in the `title`, all `title` attributes can utilize the same `i18n` key. (See conversation above)
2. Whether to **Enroll** or **Verify** is implicitly decided by the API response
3. Code cleanup!
   - There are quite a few authenticator enroll/challenge views that have nearly identical code. For example, the OnPrem authenticator ([enroll](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/on-prem/EnrollAuthenticatorOnPremView.js)/[challenge](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/on-prem/ChallengeAuthenticatorOnPremView.js)), security question ([enroll/challenge](https://github.com/okta/okta-signin-widget/tree/master/src/v2/view-builder/views/security-question)), and IdP (condensed already).

If there is another +1 to this, I'll add a note to bring it up during the next design review.

src/v2/view-builder/utils/AuthenticatorUtil.js
jmelberg-okta: You commented too quickly! 😄 

This string [never existed](https://github.com/okta/okta-signin-widget/pull/1707/files#r585108406). We missed it during code review, but _luckily_ we would never encounter the case where the `*.default` string would be used.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -718,6 +718,11 @@ cert.authentication.title = Certificate authentication
 ##                       OIE strings            ##
 ##################################################
 
+## Generic Authenticator Messages
+# {0} refers to the authenticator name (Duo, Symantec VIP, etc.)
+oie.authenticator.enroll.title = Set up {0}
+oie.authenticator.challenge.title = Verify with {0}
```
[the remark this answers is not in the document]
pradeepdewda-okta: Yes, if the key exists for the exact same string then reuse it.

pradeepdewda-okta (approved):
Looks good to me.

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -24,6 +24,7 @@ const FORMNAME_CLASSNAME_MAPPINGS = {
     [AUTHENTICATOR_KEY.OV]: 'mfa-verify',
     [AUTHENTICATOR_KEY.GOOGLE_OTP]: 'mfa-verify',
     [AUTHENTICATOR_KEY.DUO]: 'mfa-verify-duo',
+    [AUTHENTICATOR_KEY.SYMANTEC_VIP]: 'mfa-verify-symantec',
```
TianGan-okta, line 27: Make sure we have that in v1

TianGan-okta (approved):
LGTM 🌟

src/v2/ion/ViewClassNamesFactory.js
nikhilvenkatraman-okta: I think this is called just `mfa-symantec` in V1. In that case we should add that instead ?

src/v2/view-builder/views/symantec/AuthenticatorSymantecView.js
mauriciocastillosilva-okta: I was going to say that we now have 2 different patterns to create enroll/verify views, so we should choose one! And seems the answer is here. So +1 to bring it up in a design review in case we are missing something from having 2 different files, and if we can stick to it (maybe even refactor the other authenticators views at some point). And yes, I think we should show displayName from the response

mauriciocastillosilva-okta (approved):
looks good

packages/@okta/i18n/src/properties/login.properties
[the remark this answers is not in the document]
nikhilvenkatraman-okta: Is it better to have unique keys per authenticator for customization support ?

src/v2/view-builder/views/symantec/AuthenticatorSymantecView.js
nikhilvenkatraman-okta: This is cool, but maybe we should move this computation of `isChallenge` into appState instead of doing it deep at the view level ? That way it is consistent for each view.

packages/@okta/i18n/src/properties/login.properties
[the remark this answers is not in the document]
jmelberg-okta: > Is it better to have unique keys per authenticator for customization support ?

I'm really torn on this. On one hand - I love the idea of grouping all titles under a single key (assuming the values are equivalent), but we do lose out on customizability. If we do allow these `title` attributes to be customized, do we also want to extend it to the `subTitle`, input fields, and the **Enroll** and **Verify** buttons?

I think to get the answer, we'll need to loop in PM to see what our customizability story is. For now, I'd prefer we try to avoid duplication/redundancies when we can.

src/v2/view-builder/views/symantec/AuthenticatorSymantecView.js
jmelberg-okta: Thanks for the feedback all! I'll capture this feedback for a future design review. In regards to moving it to the `appState`, I think this makes a **lot** of sense!

src/v2/view-builder/views/symantec/AuthenticatorSymantecView.js
```diff
@@ -0,0 +1,46 @@
...
+
+    this.Footer = this.options.isChallenge
+      ? AuthenticatorVerifyFooter
+      : AuthenticatorEnrollFooter;
+  },
+  Body,
```
haishengwu-okta, line 45: nit:
we really shall use same pattern creating Footer to create Body.

jmelberg-okta: Agree! Interestingly, the `AuthenticatorEnrollFooter` and `AuthenticatorVerifyFooter` are equivalent, rendering the exact same view. I'm still trying to track down why we have two implementations instead of a general one - but do believe we can safely generalize this into a single view.

If we want to adopt a pattern where we're moving more logic into the `BaseAuthenticatorView`, then we can rename this to `AuthenticatorFooter` and include it within the `BaseAuthenticatorView` by default. This'll cleanup a lot of duplicate code in our views!

I'll create a ticket to track this, in addition to bring it up within our channel to capture more feedback.

packages/@okta/i18n/src/properties/login.properties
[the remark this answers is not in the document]
nikhilvenkatraman-okta: We have had several discussions with the PM and the ask from is to definitely to have support for customization. I believe this is why we have created individual keys for authenticators.

src/v2/view-builder/views/symantec/AuthenticatorSymantecView.js
haishengwu-okta: - Good catch on the Footers.
- For the Body, I meant adding something like this to init function: `this.Body = this.option.isChallenge ? VerifyForm : EnrollForm`

packages/@okta/i18n/src/properties/login.properties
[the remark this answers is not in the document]
jmelberg-okta: Thanks @nikhilvenkatraman-okta! Will continue the v1 convention of keeping these view-specific.

src/v2/view-builder/views/symantec/AuthenticatorSymantecView.js
jmelberg-okta: Ah - I see what you mean.

This is actually a pattern I'm advocating we [move away from](https://github.com/okta/okta-signin-widget/pull/1707/files/f89ce3ec2b0919cc027282a1bc344b8b67ec8e30#r585114176). Two forms are not always necessary and create duplicate code for a few of our authenticators.

I'll open up a draft PR that condenses these sometime this week to help visualize this.

src/v2/ion/ViewClassNamesFactory.js
jmelberg-okta: Good catch! Re-mapped to `mfa-symantec`

src/v2/view-builder/views/symantec/AuthenticatorSymantecView.js
haishengwu-okta: Cannot see much duplication. we can chat in your draft PR later.

PR #1710, opened 2021-03-03
fix: changes OIE On-prem i18n strings to use v1 i18n strings

gowthamidommety-okta:
Resolves: OKTA-371493

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-371493](https://oktainc.atlassian.net/browse/OKTA-371493)


src/v2/ion/i18nTransformer.js
```diff
@@ -122,6 +122,35 @@ const I18N_OVERRIDE_MAPPINGS = {
   'oie.session.expired' : 'oie.idx.session.expired',
 };
 
+const getDisplayName = (remediation, defaultName) => {
+  return remediation.relatesTo?.value?.displayName || defaultName;
+};
```
jmelberg-okta, line 127: Thanks for doing this! This helper method will hopefully allow us to simplify a lot of our view complexity.

Since this isn't specific to `i18n` and retrieves an authenticator's `displayName`, I'd suggest moving this to `AuthenticatorUtil` instead.

src/v2/ion/i18nTransformer.js
```diff
@@ -187,7 +217,8 @@ const updateLabelForUiSchema = (remediation, uiSchema) => {
 
   if (uiSchema.label) {
     Logger.info('\t 2: ', i18nPath);
-    uiSchema.label = getI18NValue(i18nPath, uiSchema.label);
+    const params = getI18NParams(remediation, authenticatorKey);
+    uiSchema.label = getI18NValue(i18nPath, uiSchema.label, params);
```
jmelberg-okta, line 221: Suggestion: Instead of requiring a call to `getI18NParams` here, let's nest it within the `getI18nValue` method. That way, all our i18n lookup logic is centralized.

src/v2/ion/i18nTransformer.js
jmelberg-okta: Since the `displayName` is a required attribute with remediation forms, we can probably get away with not adding a default value here.

src/v2/ion/i18nTransformer.js
```diff
@@ -122,6 +122,35 @@ const I18N_OVERRIDE_MAPPINGS = {
...
+  [FORMS.ENROLL_AUTHENTICATOR]: {
+    [AUTHENTICATOR_KEY.ON_PREM]: {
+      getParam: getDisplayName,
+      defaultParam: loc('oie.on_prem.authenticator.default.vendorName', 'login'),
+    },
+    [AUTHENTICATOR_KEY.RSA]: {
+      getParam: getDisplayName,
+      defaultParam: loc('oie.on_prem.authenticator.default.vendorName', 'login'),
+    },
+  },
```
jmelberg-okta, line 139: As we add more enrollment forms, this dictionary is going to be a burden to manage. Any reason we can't apply this to all forms now?

jmelberg-okta (commented):
Couple of initial suggestions + questions.

Seeing some tests for different authenticators and future values we'd like to parse to visualize this new functionality a bit better!

src/v2/ion/i18nTransformer.js
nikhilvenkatraman-okta: +1 Should Custom OTP also be added to this list ? https://github.com/okta/okta-signin-widget/blob/master/packages/@okta/i18n/src/properties/login.properties#L829

src/v2/ion/i18nTransformer.js
```diff
@@ -122,6 +122,35 @@ const I18N_OVERRIDE_MAPPINGS = {
...
+
+const I18N_PARAMS_MAPPING = {
+  [FORMS.ENROLL_AUTHENTICATOR]: {
+    [AUTHENTICATOR_KEY.ON_PREM]: {
+      getParam: getDisplayName,
+      defaultParam: loc('oie.on_prem.authenticator.default.vendorName', 'login'),
+    },
+    [AUTHENTICATOR_KEY.RSA]: {
+      getParam: getDisplayName,
+      defaultParam: loc('oie.on_prem.authenticator.default.vendorName', 'login'),
```
nikhilvenkatraman-okta, line 137: nit: should this point to `oie.rsa.authenticator.default.vendorName` ?

src/v2/ion/i18nTransformer.js
nikhilvenkatraman-okta: If I understand this correct, we have the same logic to show a vendor name by reading `remediation.relatesTo?.value?.displayName` and  fallback to show a  default vendor name. 
 
This is  also only for a few use cases where we have vendor specific authenticators like (on prem, rsa, custom OTP) but the logic is the same for all.

gowthamidommety-okta: > As we add more enrollment forms, this dictionary is going to be a burden to manage. Any reason we can't apply this to all forms now?

I tried that first (to just pass displayName as a param always), but i cannot set params on a string that does not need them. courage throws an error - https://github.com/okta/okta-ui/blob/master/packages/courage/src/util/StringUtil.js#L64

src/v2/ion/i18nTransformer.js
gowthamidommety-okta: good catch! I use the same default vendor name for both since they are practically the same.. but `factor.totpHard.rsaSecurId` makes more sense here. Thank you!

src/v2/ion/i18nTransformer.js
gowthamidommety-okta: > +1 Should Custom OTP also be added to this list ? https://github.com/okta/okta-signin-widget/blob/master/packages/@okta/i18n/src/properties/login.properties#L829

The only reason i cannot is because im not sure if the uischema.label for custom totp string takes a param

src/v2/ion/i18nTransformer.js
gowthamidommety-okta: I looked into this, and it will affect all uiSchemas then. I thought we should restrict this to labels. What do you think?

src/v2/ion/i18nTransformer.js
nikhilvenkatraman-okta: Hmm not sure if I understand that. But can we not tell by looking at this ? https://github.com/okta/okta-signin-widget/blob/master/packages/@okta/i18n/src/properties/login.properties#L829

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -833,24 +833,17 @@ oie.custom_otp.verify.passcode.label = Enter code
 oie.on_prem.authenticator.description = Verify by entering a code generated by {0}.
 oie.on_prem.authenticator.default.vendorName = Custom On-prem Authenticator
 oie.on_prem.enroll.title = Set up {0}
-oie.on_prem.enroll.username.label = Enter authenticator username
-oie.on_prem.enroll.passcode.label = Enter authenticator passcode
 oie.on_prem.verify.title = Verify with {0}
-oie.on_prem.verify.passcode.label = Enter code
```
nikhilvenkatraman-okta, line 839: For the string `Enter code` we have created oie specifc keys 
oie.custom_otp.verify.passcode.label, oie.google_authenticator.otp.enterCodeText should they be replaced with `mfa.challenge.enterCode.placeholder` as well ?

nikhilvenkatraman-okta: I see that we are preserving any V1 customization a customer may have with this change where we are re-using `mfa.challenge.enterCode.placeholder` for on prem and rsa. 

But What if customer wanted to customize the string as 
```
Enter Rsa Code
Enter onPrem Code
```

With this change they wont be able to do that correct ? They can only do one customization for `mfa.challenge.enterCode.placeholder` that will show up in both authenticators.

playground/mocks/data/idp/idx/error-authenticator-enroll-passcode-change-on-prem.json
```diff
@@ -9,7 +9,7 @@
       {
         "message": "Wait for token to change, then enter the new tokencode.",
         "i18n": {
-          "key": "oie.on_prem.error.token.change",
+          "key": "errors.E0000106",
```
jmelberg-okta, line 12: I'd suggest we use an `oie.*` value here (or transform it), as we should try to avoid grouping certain authenticator error responses to our global error.

From a customization standpoint, this helps us be as clear as possible on keys that are mapped directly to API responses.

src/v2/ion/i18nTransformer.js
jmelberg-okta: I'd assume we'll also want this for additional schemas, but am fine if we keep this here for now!

src/v2/view-builder/utils/AuthenticatorUtil.js
```diff
@@ -195,5 +195,9 @@ export function removeRequirementsFromError (errorJSON, policy) {
   return errorJSON;
 }
 
+export function getAuthenticatorDisplayName (remediation) {
```
jmelberg-okta, line 198: nit: Let's add doc to this function so it shows up with IntelliSense.

jmelberg-okta (approved):
Overall the changes LGTM, but strongly suggest we do not rely on any `errors.*` properties directly.

packages/@okta/i18n/src/properties/login.properties
nikhilvenkatraman-okta: Resolved offline on slack

nikhilvenkatraman-okta (approved):

playground/mocks/data/idp/idx/error-authenticator-enroll-passcode-change-on-prem.json
nikhilvenkatraman-okta: I think the goal is to re-use an existing V1 error key ?
```
errors.E0000106 = Wait for token to change, then enter the new tokencode.
```

src/v2/ion/i18nTransformer.js
gowthamidommety-okta: Right but none of the strings (except title https://github.com/okta/okta-signin-widget/blob/master/packages/@okta/i18n/src/properties/login.properties#L826) takes a param

playground/mocks/data/idp/idx/error-authenticator-enroll-passcode-change-on-prem.json
gowthamidommety-okta: Yes, @nikhilvenkatraman-okta is right. I was trying to use our existing error override string

PR #1717, opened 2021-03-04
port 5.4.1

aarongranick-okta:
### Reviewers:
@jmelberg-okta 

### Issue:

- [OKTA-375492](https://oktainc.atlassian.net/browse/OKTA-375492)


src/widget/buildRenderOptions.js
```diff
@@ -0,0 +1,23 @@
...
+  if (!renderOptions.clientId) {
+    throw new Errors.ConfigError('"clientId" is required');
+  }
+
+  if (!renderOptions.redirectUri) {
+    throw new Errors.ConfigError('"redirectUri" is required');
+  }
+
+  return renderOptions;
+}
```
jmelberg-okta, line 23: style-lint: Newlines at the eof are generally preferred. Surprised we don't have a lint rule for this!

scripts/jest.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-setup_service google-chrome-stable 83.0.4103.61-1
-
 source $OKTA_HOME/$REPO/scripts/setup.sh
```
jmelberg-okta, line 3: Do we intend to include this in our `.travis.yml` config as well?

jmelberg-okta (approved):
Changes from the Symantec VIP authenticator look correct!

We have the opportunity to enable a few more lint rules to cleanup some common `style-lint` errors, maybe we pick this up sometime next week.

haishengwu-okta (approved):

scripts/jest.sh
aarongranick-okta: sounds like a reasonable proposition

PR #1725, opened 2021-03-05
bug: Fixes issues with IdP and Symantec authenticators

jmelberg-okta:
## Description:

- Fixes a couple of issues where single-enrollment authenticators would verify properly
- Addresses a css issue where the Symantec VIP authenticator extended an improper class 

## PR Checklist

- [x] Added unit tests?

### Screenshot/Video:

| Before | After |
| ----------- | ------ |
| Screen Shot 2021-03-05 at 11 45 03 AM | Screen Shot 2021-03-05 at 11 42 40 AM |

### Reviewers:
- @gowthamidommety-okta 

### Issue:

- [OKTA-376140](https://oktainc.atlassian.net/browse/OKTA-376140)


src/v2/BaseLoginRouter.js
```diff
@@ -133,7 +133,7 @@ export default Router.extend({
           ]
         };
       }
-      // Need to mimic IdxRespones as idx returns raw response at error case
+      // Need to mimic IdxResponses as idx returns raw response at error case
```
jmelberg-okta, line 136: I've addressed a few typos in files I touched while debugging - none of which should have an impact on functionality.

src/v2/models/AppState.js
```diff
@@ -106,6 +106,28 @@ export default Model.extend({
     return currentViewState;
   },
 
+  /**
+   * Checks to see if we're in an enrollment or challenge flow.
```
jmelberg-okta, line 110: @gowthamidommety-okta - Thanks for the ping yesterday about this.

There are two cases we want to account for:

1. The authenticator supports only one enrollment (email, duo, etc)
2. The authenticator supports multiple enrollments

As a result, when we're determining whether or not we're in a challenge flow - we can look into both of these objects for verification.

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -24,7 +24,6 @@ const FORMNAME_CLASSNAME_MAPPINGS = {
     [AUTHENTICATOR_KEY.OV]: 'mfa-verify',
     [AUTHENTICATOR_KEY.GOOGLE_OTP]: 'mfa-verify',
     [AUTHENTICATOR_KEY.DUO]: 'mfa-verify-duo',
-    [AUTHENTICATOR_KEY.SYMANTEC_VIP]: 'mfa-symantec',
```
gowthamidommety-okta, line 27: why was this removed? This would be `mfa-verify` right?

src/v2/ion/ViewClassNamesFactory.js
```diff
@@ -42,7 +41,6 @@ const FORMNAME_CLASSNAME_MAPPINGS = {
     [AUTHENTICATOR_KEY.ON_PREM]: 'enroll-onprem',
     [AUTHENTICATOR_KEY.RSA]: 'enroll-rsa',
     [AUTHENTICATOR_KEY.DUO]: 'enroll-duo',
-    [AUTHENTICATOR_KEY.SYMANTEC_VIP]: 'enroll-symantec',
```
gowthamidommety-okta, line 45: This is correct.. why was this removed?

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -28,13 +28,8 @@ export default BaseAuthenticatorView.extend({
...
 
-    const currentAuthenticator = this.options.appState.get('currentAuthenticator');
-    const authenticatorEnrollments = this.options.appState.get('authenticatorEnrollments')?.value || [];
-    const existingAuthenticator = _.find(authenticatorEnrollments, { id: currentAuthenticator.id });
-
-    // If the authenticator is already enrolled, we're in a challenge flow as opposed to enrollment
-    this.options.isChallenge = existingAuthenticator || false;
-    this.options.displayName = currentAuthenticator.displayName;
+    this.options.isChallenge = this.options.appState.isAuthenticatorChallenge();
+    this.options.displayName = this.options.appState.get('currentAuthenticator').displayName;
```
haishengwu-okta, line 32: `currentAuthenticator` v.s `currentAuthenticatorEnrollment`?

jmelberg-okta: This gets tricky based on if the enrolled authenticator is coming from `currentAuthenticatorEnrollment` or the list of `authenticatorEnrollments`. I think pulling this from the `currentAuthenticator` is our best bet.

src/v2/models/AppState.js
```diff
@@ -106,6 +106,28 @@ export default Model.extend({
...
+   * @returns {boolean}
+   */
+  isAuthenticatorChallenge () {
+    const currentAuthenticator = this.get('currentAuthenticator');
+    const currentAuthenticatorEnrollment = this.get('currentAuthenticatorEnrollment');
+    const authenticatorEnrollments = this.get('authenticatorEnrollments')?.value || [];
+
+    // For single-enrollment (currentAuthenticatorEnrollment)
+    if (currentAuthenticatorEnrollment && currentAuthenticator.id === currentAuthenticatorEnrollment.id) {
```
gowthamidommety-okta, line 119: Are `currentAuthenticator` and `currentAuthenticatorEnrollment` both returned for all responses? I also see some mock responses returning both `currentAuthenticatorEnrollment` and `authenticatorEnrollments`.. It is possible our mocks are not accurate though. cc @SamSanjabi-okta 

What about checking the form name instead to figure out if we are in challenge flow?

SamSanjabi-okta: I don't think `currentAuthenticator` and `currentAuthenticatorEnrollment` should ever _both_ be in the response.  In general:
- `currentAuthenticator` = current authenticator being challenged during enrollment, or during verification for authenticators verified by type (like webauthn) -- right now RSA & Symantec might also be doing they shouldn't and it's something we'll fix later.
- `currentAuthenticatorEnrollment` = current authenticator being challenged during verification for all other cases (single enrollment authenticators, multiple enrollment authenticators when a device is selected like phone, etc...)

`authenticatorEnrollments` I believe comes back in all responses after the user is identified and has all of user's enrollments, unless it's the "anonymized" password + email screen in which case it just has those two but the object will still be there.

src/v2/ion/ViewClassNamesFactory.js
jmelberg-okta: The `mfa-symantec` class is unfortunately mapped to CSS that adds a [background image](https://github.com/okta/okta-signin-widget/blob/master/assets/sass/okta-theme.scss#L285-L295). Will change this to `mfa-verify`!

src/v2/ion/ViewClassNamesFactory.js
jmelberg-okta: Accidental removal on my end - will fix!

src/v2/models/AppState.js
jmelberg-okta: From what I have seen, `currentAuthenticator` is always present for enroll or challenge flows. The `currentAuthenticatorEnrollment` may or may not be present, hence the conditional check here for its existence.

We could check the form name, but I figured a more future-proof solution was to always check if the `currentAuthenticator` exists in either the `currentAuthenticatorEnrollment` or `authenticatorEnrollments` fields.

gowthamidommety-okta: > currentAuthenticator is always present for enroll or challenge flows

What Sam said above is that it depends on the authenticator. You are probably seeing it for symantec because on the backend, it is verified by type (like webauthn). My understanding is that we will see `currentAuthenticator` for authenticators with type `app` and ones with type `security_key`. Are you seeing this for other authenticator types too? 

From the info in this thread, it doesn't seem like we can use these objects 

- `authenticatorEnrollments` is always returned after the user is identified - this means after identify, in either enroll or verify flows
- `currentAuthenticator` is always returned during enrollment. And only during verification of authenticators verified by type. It is not consistently there during the verify flow for all authenticators
- `currentAuthenticatorEnrollment` is always returned on verify but ONLY for those authenticators NOT verified by type.

I think the best way to check if we are in challenge flow, at this time, is by form name.

jmelberg-okta: Given that we `currentAuthenticator` and `currentAuthenticatorEnrollment` should _ideally_ not be present, I'll update this logic to rely on the formName

src/v2/models/AppState.js
```diff
@@ -107,25 +107,38 @@ export default Model.extend({
...
-      return true;
+    if (!isChallenge) {
+      // Always returned during enrollment
+      return currentAuthenticator.displayName;
     }
 
-    if (authenticatorEnrollments.length > 0) {
-      // For multi-enrollment - return if the current id matches
-      return !!_.find(authenticatorEnrollments, { id: currentAuthenticator.id });
+    // Challenge/verification flows may or may have this name in different
```
jmelberg-okta, line 121: Typo, will fix when squashing!

```diff
-// Challenge/verification flows may or may have this name in different
+// Challenge/verification flows may or may not have this name in different

File: src/v2/models/AppState.js [hunk 8]
```diff
@@ -106,6 +106,41 @@ export default Model.extend({
     return currentViewState;
   },
 
+  /**
+   * Returns the displayName of the authenticator
+   * @returns {string}
+   */
+  getAuthenticatorDisplayName () {
```
jmelberg-okta, line 113: Just to add a note - this is currently only used by the IdP and Symantec authenticators. As more authenticators conform this to pattern (conditionally load enroll or verify views) - we'll be able to utilize this more often!

gowthamidommety-okta: Thank you! I think i can use this for on-prem 😄 

Can we do this a little differently? I don't think we need to check challenge, but instead could do
`return currentAuthenticator.displayName || currentAuthenticatorEnrollment.displayName;`
Since displayName will be on either of those objects. 

Alternatively we could get the current view state `getCurrentViewState` and get `relatesTo` from there. relatesTo will always point to the right object. But that might be too much since we already have currentAuthenticator and currentAuthenticatorEnrollment.

src/v2/models/AppState.js
```diff
@@ -106,6 +106,41 @@ export default Model.extend({
...
+      return currentAuthenticator.displayName;
+    }
+
+    // Challenge/verification flows may or may have this name in different
+    // parts of the response. We walk through those here.
+    const currentAuthenticatorEnrollment = this.get('currentAuthenticatorEnrollment');
+    if (currentAuthenticatorEnrollment && currentAuthenticatorEnrollment.displayName) {
+      // Always returned for challenge/verify flows - except when the authenticator is
+      // verified by type.
+      return currentAuthenticatorEnrollment.displayName;
```
nikhilvenkatraman-okta, line 127: In [some cases](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/custom-otp/ChallengeCustomOTPAuthenticatorView.js#L11) we fallback to a default vendor name if urrentAuthenticatorEnrollment.displayName is not set.

nikhilvenkatraman-okta: Don't need to handle that in this PR but when we do this in the future, we should think about accepting the authenticator type as an argument to this method.

src/v2/models/AppState.js
jmelberg-okta: This simplifies the logic a lot! I guess this level of granularity is just extra noise 😄

src/v2/models/AppState.js
jmelberg-okta: Yeah - from what I understand from the few backend devs I've talked to, the `displayName` is a required property that'll always be returned via our API. Doesn't hurt to have this fallback though!

gowthamidommety-okta (approved):
lgtm!

nikhilvenkatraman-okta (approved):

PR #1729, opened 2021-03-08
Handle IdP Authenticator errors

jmelberg-okta:
## Description:

By default, the Widget will surface IDX errors from the `idx/introspect` response in the form of a **Terminal** error. This means there are the user is at a "dead stop" and must refresh the page to retry their login attempt. As a result, we will not be able to let the user proceed from the enroll/challenge page, as indicated in [this UX mock](https://www.figma.com/file/I1ebb1ApKB1j7EkIaAeWFH/OIE-Authenticators?node-id=783%3A400).

This is different than other authenticators because the IdP Authenticator relies on redirecting back to the Widget versus handling the response from `challenge/answer`. To address this, this PR adds:
- A `showMessages` override to display `messages` as a callout
- Special handling for the `redirect-idp` form **if and only if** we are in an error scenario.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

| Enrollment | Verification |
| ---------- | ----------- |
| Enrollment | Verification |

### DIFF
--- a/playground/mocks/data/idp/idx/error-authenticator-enroll-idp.json
+++ b/playground/mocks/data/idp/idx/error-authenticator-enroll-idp.json
@@ -1,0 +1,144 @@
+{
+  "stateHandle": "02TptqPN4BOLIwMAGUVLPlZVJEnONAq7xkg19dy6Gk",
+  "version": "1.0.0",
+  "expiresAt": "2021-01-19T15:10:35.000Z",
+  "intent": "LOGIN",
+  "messages": {
+    "type": "array",
+    "value": [
+      {
+        "message": "Unable to enroll. Try again or contact your admin for assistance.",
+        "i18n": {
+          "key": "",
+          "params": []
+        },
+        "class": "ERROR"
+      }
+    ]
+  },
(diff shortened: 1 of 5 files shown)


playground/mocks/data/idp/idx/error-authenticator-enroll-idp.json
```diff
@@ -0,0 +1,144 @@
...
+  "intent": "LOGIN",
+  "messages": {
+    "type": "array",
+    "value": [
+      {
+        "message": "Unable to enroll. Try again or contact your admin for assistance.",
+        "i18n": {
+          "key": "",
+          "params": []
+        },
```
jmelberg-okta, line 14: This message is similar to V1 and once the backend work is completed we'll add the correct key.

gowthamidommety-okta (commented):
approach lgtm! Does AuthenticatorIdPView.js only render for the error case? Can we explain that in comments / change the file name if that's true?

jmelberg-okta:
Closing w.r.t #1738

PR #1730, opened 2021-03-09
implement features.transformUsername on v2 OIE

sherizada-okta:
## Description:
Transforms `identifier` using the `settings.transformUsername` function.

An example transformUsername that is set in .widgetrc.js:
```
// The callback function is passed two arguments:
// 1) username: The name entered by the user
// 2) operation: The type of operation the user is trying to perform:
//      - PRIMARY_AUTH
//      - FORGOT_PASSWORD
//      - UNLOCK_ACCOUNT
transformUsername: function (username, operation) {
  // This example will append the '@acme.com' domain if the user has
  // not entered it
  return username.indexOf('@acme.com') > -1
    ? username
    : username + '@acme.com';
}
```
https://github.com/okta/okta-signin-widget#username-and-password

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:
Screenshot 2021-03-11 at 2 12 27 PM
Screenshot 2021-03-11 at 2 12 15 PM

### Issue:

- [OKTA-314098](https://oktainc.atlassian.net/browse/OKTA-314098)


src/v2/controllers/FormController.js
```diff
@@ -165,6 +165,12 @@ export default Controller.extend({
       return;
     }
 
+    const identifier = model.get('identifier');
+    const transformUsername = this.settings.get('features.transformUsername');
+    if (_.isFunction(transformUsername) && identifier) {
```
pradeepdewda-okta, line 170: what if the identifier is an email ? We should check with backend on this..

sherizada-okta: It's up to the dev what they want to do with transformUsername. 
Here's an example from SIW readme:
```
// The callback function is passed two arguments:
// 1) username: The name entered by the user
// 2) operation: The type of operation the user is trying to perform:
//      - PRIMARY_AUTH
//      - FORGOT_PASSWORD
//      - UNLOCK_ACCOUNT
transformUsername: function (username, operation) {
  // This example will append the '@acme.com' domain if the user has
  // not entered it
  return username.indexOf('@acme.com') > -1
    ? username
    : username + '@acme.com';
}
```

pradeepdewda-okta:
this looks good.. adding some tests ?

sherizada-okta:
@pradeepdewda-okta I'm thinking now of moving this logic inside BaseModel.toJSON.

src/v2/controllers/FormController.js
```diff
@@ -17,6 +17,7 @@ import { FORMS } from '../ion/RemediationConstants';
 import Util from '../../util/Util';
 import { clearTransactionMeta } from '../client/transactionMeta';
 
+
```
haishengwu-okta, line 20: remove extra empty line

src/v2/view-builder/internals/BaseModel.js
```diff
@@ -11,6 +11,13 @@
  */
 import { _, Model } from 'okta';
 
+const FORM_NAME_TO_OPERATION_MAP = {
+  'select-authenticator-unlock-account': 'UNLOCK_ACCOUNT',
+  'identify': 'PRIMARY_AUTH',
+  'select-identify': 'PRIMARY_AUTH',
```
haishengwu-okta, line 17: Doesn't find what view represented by `select-identify`?
Looks like it's just an action to switch from register page to identify page?

src/v2/view-builder/internals/BaseModel.js
```diff
@@ -80,6 +87,22 @@ const create = function (remediation = {}, optionUiSchemaConfig = {}) {
   const BaseModel = Model.extend({
     props,
     local,
+    toJSON: function () {
+      const modelJSON = Model.prototype.toJSON.apply(this, arguments);
+
+      const identifier = this.get('identifier');
+      if (identifier) {
```
haishengwu-okta, line 94: shall it be `identifier && currentFormName is one of Object.keys(FORM_NAME_TO_OPERATION_MAP)`?

haishengwu-okta (commented):
Do we have counter parter in v2 for `'signin/recovery-challenge'` which looks like also call transforUserName

src/v2/view-builder/internals/BaseModel.js
```diff
@@ -11,6 +11,13 @@
  */
 import { _, Model } from 'okta';
 
+const FORM_NAME_TO_OPERATION_MAP = {
+  'select-authenticator-unlock-account': 'UNLOCK_ACCOUNT',
```
haishengwu-okta, line 15: use constant from `Enum.js` ?

src/v2/view-builder/internals/BaseModel.js
```diff
@@ -80,6 +87,22 @@ const create = function (remediation = {}, optionUiSchemaConfig = {}) {
   const BaseModel = Model.extend({
     props,
     local,
+    toJSON: function () {
```
haishengwu-okta, line 90: I'm thinking another option is to call transformUserName at `FormController.handleSaveForm`, right before `idx.proceed`.
This option will work for `processCreds` which could be pure function and async function (return promise)

sherizada-okta: That option is what I originally had - https://github.com/okta/okta-signin-widget/pull/1730/commits/c01f10a98d8939d557aee90045d7fc114caffed1

But I didn't find an existing way to test the FormController, and now I think its easier for testing to put it here and a more intuitive place for it.

processCreds would make sense in the FormController.

src/v2/view-builder/internals/BaseModel.js
sherizada-okta: That would make this setting bound to our upkeep of this Enum, which would be annoying for us and users.

sherizada-okta:
> Do we have counter part in v2 for `'signin/recovery-challenge'` which looks like also call transformUserName

@haishengwu-okta I don't understand this question, can you rephrase? These changes are for v2.

src/v2/view-builder/internals/BaseModel.js
sherizada-okta: enroll-profile-new.json and enroll-profile.json have an identifier form value for it:
```
{
        "rel": [
          "create-form"
        ],
        "name": "select-identify",
        "href": "http://localhost:3000/idp/idx",
        "method": "POST",
        "accepts": "application/vnd.okta.v1+json",
        "value": [
          {
            "name": "identifier",
            "label": "identifier"
          },
          {
            "name": "stateHandle",
            "value": "01r2p5S9qaAjESMFuPzt7r3ZMcZZQ_vvS0Tzg56ajB",
            "visible": false
          }
        ]
      }

Comment (haishengwu-okta) [reply src/v2/view-builder/internals/BaseModel.js:17]: I didn't find a View represented for `select-identify` in ViewFactory hence the ask.

I may miss something or we need to double check if the mock is out of dated.

File: src/v2/view-builder/internals/BaseModel.js (hunk 6 above)
Comment (haishengwu-okta) [reply src/v2/view-builder/internals/BaseModel.js:90]: Hard to test seems not a good justification cause you'll need to test `processCreds` anyway? :D 
In test cafe, We could re-render ui with different set of options so that you'd be able to verify the identifier changed (in request body)

**Comment by haishengwu-okta** [comment]
Maybe we don't have `signin/recovery-challenge` in V2 (I don't quite remember). It's just question and worth to double check.

File: src/v2/view-builder/internals/BaseModel.js (hunk 6 above)
Comment (sherizada-okta) [reply src/v2/view-builder/internals/BaseModel.js:90]: > Hard to test seems not a good justification cause you'll need to test `processCreds` anyway? :D
heh I know, but I wanted to put off crossing that bridge until I really had to.

**Review by pradeepdewda-okta — APPROVED** [review]
💯

File: src/util/Enums.js [hunk 7]
```diff
@@ -79,4 +79,10 @@ export default {
   UNIVERSAL_LINK_CHALLENGE: 'UNIVERSAL_LINK',
 
   LAUNCH_AUTHENTICATOR_REMEDIATION_NAME: 'launch-authenticator',
+
+  FORM_NAME_TO_OPERATION_MAP: {
```
haishengwu-okta, line 83: this var probably shall go into `RemediationConstant` but uses `Enums.PRIMARY_AUTH`, etc.

src/v2/controllers/FormController.js
```diff
@@ -165,6 +167,22 @@ export default Controller.extend({
       return;
     }
 
+    const identifier = model.get('identifier');
```
haishengwu-okta, line 170: I'd rather to change json result instead of model itself.
```
const requestValue = model.toJSON();
if requestValue.identify then
  ...
idx.proceed(formName, requestValue);
```

haishengwu-okta (approved):
- I think testcafe is good enough without unit test.
- Move around the Enum variable and 🚀

sherizada-okta:
> Maybe we don't have `signin/recovery-challenge` in V2 (I don't quite remember). It's just question and worth to double check.

The recovery flows that take an identifier are unlock account and forgot password.

PR #1737, opened 2021-03-09
fix: fixes typo in error message

nikhilvenkatraman-okta:
Resolves: OKTA-376702

## Description:

Fixes typo

## PR Checklist

- [x] Added e2e tests

### Reviewers:

@greyyin-okta @davidantonwalters-okta 

### Issue:

- [OKTA-376702](https://oktainc.atlassian.net/browse/OKTA-376702)


greyyin-okta (approved):

pradeepdewda-okta (commented):
I think there is a selenium test needs to be fixed as well.. make sure to fix before you push to 5.4..

https://github.com/okta/okta-core/blob/94529e522d73508a7d85d36d143504ede21b2533/components/web/login/impl/src/main/java/com/okta/api/mediation/services/auth/IdentityEngineMediationServiceImpl.java#L1453

pradeepdewda-okta:
> @pradeepdewda-okta this is being addressed already https://github.com/okta/okta-core/pull/50455/files#diff-9d22ec1bc0189103568bfe9c8b7dc9412edbd0f6c4cf1a3427f2d1252837d9a1R1453

thanks.. I see few more instances when you search in okta-core.. A file name also has typo.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -840,7 +840,7 @@ oie.numberchallenge.warning = Haven't received a push notification yet? Try open
 ## {0} refers to a number which is the correct answer for the number challenge
 oie.numberchallenge.instruction = On your mobile device, open the Okta Verify prompt, then tap <$1>{0}</$1> in Okta Verify to continue.
 oie.numberchallenge.force.upgrade.title = Update Okta Verify
-oie.numberchallenge.force.upgrade = Your response was received, but your Okta Verify version is no longer supported by your organization. To verify your identity with push notifcations, update Okta Verify to the latest version, then try again.
+oie.numberchallenge.force.upgrade = Your response was received, but your Okta Verify version is no longer supported by your organization. To verify your identity with push notifications, update Okta Verify to the latest version, then try again.
```
greyyin-okta, line 843: IIUC, UI will retrieve the error msg from this file instead of from okta-core, will this lead to any problem?
For example, we have changed the error msg in okta-core, but not here, the user will be shown the out-of-date msg?

nikhilvenkatraman-okta: Yes, the UI looks at the i18n key sent by the server and does a look up in this file to show the message. The error message in okta-core should not have any effect on the widget as we dont use it at all, we only look at the key to add a message on the client side.

gowthamidommety-okta (approved):
🚢

mauriciocastillosilva-okta (approved):

PR #1738, opened 2021-03-09
Handle IdP Authenticator errors

jmelberg-okta:
## Description:

By default, the Widget will surface IDX errors from the `idx/introspect` response in the form of a **Terminal** error. This means there are the user is at a "dead stop" and must refresh the page to retry their login attempt, as all [remediations are dropped](https://github.com/okta/okta-signin-widget/blob/master/src/v2/BaseLoginRouter.js#L140).

This PR adds the ability for error responses that contain `remediation` values to be handled.

**Note**: Requires changes from `okta-idx-js` [#49](https://github.com/okta/okta-idx-js/pull/49) before merge.

## PR Checklist

- [x] Added e2e tests

### Screenshot/Video:

| Enrollment | Verification |
| ---------- | ----------- |
| Enrollment | Verification |

### Resolves

- [OKTA-374550](https://oktainc.atlassian.net/browse/OKTA-374550)

### DIFF
--- a/src/v2/BaseLoginRouter.js
+++ b/src/v2/BaseLoginRouter.js
@@ -101,44 +101,49 @@ export default Router.extend({
     // The response is not in IDX format. See playground/mocks/data/oauth2/error-feature-not-enabled.json
     if (error?.error === 'access_denied' && error.error_description) {
       // simulate an IDX error response
+      const idxMessages = {
+        messages: {
+          type: 'array',
+          value: [{
+            message: error.error_description,
+            i18n: {
+              key: 'oie.feature.disabled'
+            },
+            class: 'ERROR'
+          }],
+        },
+      };
+
+      // Format the error to resemble an IdX error
       error = Object.assign({
         details: {
-          messages: {
-            type: 'array',
-            value: [{
-              message: error.error_description,
-              i18n: {
-                key: 'oie.feature.disabled'
-              },
-              class: 'ERROR'
-            }]
-          }
+          rawIdxState: idxMessages,
+          context: idxMessages,
+          neededToProceed: [],
         }
       }, error);
     }
 
-        
     // loosely check whether is IDX error response
     // see idx for details: https://github.com/okta/okta-idx-js/blob/master/src/index.js
-    if (error?.details) {
-      // Populate generic error message if there isnt any.
-      if (!error.details.messages ) {
-        error.details.messages = {
+    if (error?.details?.rawIdxState) {
+      // Populate generic error message if there isn't any.
+      if (!error.details.rawIdxState.messages) {
+        const idxMessage = {
           type: 'array',
-          'value': [
+          value: [
             {
               message: loc('oform.error.unexpected', 'login'),
               class: 'ERROR'
             }
           ]
         };
+        // Format the error to resemble an IdX error
+        error.details.rawIdxState.messages = idxMessage;
+        error.details.context = { messages: idxMessage };
       }
-      // Need to mimic IdxResponses as idx returns raw response at error case
-      return this.handleIdxResponseSuccess({
-        rawIdxState: error.details,
-        context: _.pick(error.details, 'messages'),
-        neededToProceed: [],
-      });
+
+      return this.handleIdxResponseSuccess(error.details);
     }
 
     // assume it's a config error
@@ -146,7 +151,7 @@ export default Router.extend({
       error
     ));
 
-    // -- TODO: OKTA-244631 How to suface up the CORS error in IDX?
+    // -- TODO: OKTA-244631 How to surface up the CORS error in IDX?
     // -- The `err` object from idx.js doesn't have XHR object
     // Global error handling for CORS enabled errors
     // if (err.xhr && BrowserFeatures.corsIsNotEnabled(err.xhr)) {
--- a/src/v2/ion/responseTransformer.js
+++ b/src/v2/ion/responseTransformer.js
@@ -35,8 +35,8 @@ const isObject = x => _.isObject(x);
  */
 const getFirstLevelObjects = (resp) => {
   const result = {};
-  _.each(resp, (val, key) => {
-    // if key is remediation we dont do any transformation
+  _.each(resp, (val = {}, key) => {
+    // if key is remediation we don't do any transformation
     if (key === 'remediation') {
       return;
     }
@@ -87,7 +87,6 @@ const getRemediationValues = (idx) => {
         value: [],
       });
     }
-
   }
   return {
     remediations: [
@@ -194,7 +193,10 @@ const convert = (settings, idx = {}) => {
   );
 
   injectIdPConfigButtonToRemediation(settings, result);
-  convertRedirectIdPToSuccessRedirectIffOneIdp(result);
+  if (!result.messages) {
+    // Only redirect to the IdP if we are not in an error flow
+    convertRedirectIdPToSuccessRedirectIffOneIdp(result);
+  }
 
   return result;
 };
--- a/src/v2/view-builder/views/idp/AuthenticatorIdPView.js
+++ b/src/v2/view-builder/views/idp/AuthenticatorIdPView.js
@@ -1,4 +1,4 @@
-import { loc } from 'okta';
+import { createCallout, loc } from 'okta';
 import { BaseForm } from '../../internals';
 import BaseAuthenticatorView from '../../components/BaseAuthenticatorView';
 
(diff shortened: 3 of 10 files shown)


src/v2/ion/responseTransformer.js
```diff
@@ -35,8 +35,8 @@ const isObject = x => _.isObject(x);
  */
 const getFirstLevelObjects = (resp) => {
   const result = {};
-  _.each(resp, (val, key) => {
-    // if key is remediation we dont do any transformation
+  _.each(resp, (val = {}, key) => {
```
jmelberg-okta, line 38: In [some cases](https://github.com/okta/okta-idx-js/blob/master/src/v1/makeIdxState.js#L41), this can be `undefined`. Adding a default assignment here for safety.

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -18,6 +18,29 @@ const Body = BaseForm.extend({
       : loc('oie.idp.enroll.description', 'login', [displayName]);
   },
 
+  showMessages () {
```
jmelberg-okta, line 21: Long-term goal: This logic should move to a `BaseAuthenticatorView` which will surface any/all messages as default behavior. Without this override, they're displayed as [text on the page](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/internals/BaseForm.js#L93)

pradeepdewda-okta:
@jmelberg-okta the error should show above the subtitle as a global error in the form..

src/v2/BaseLoginRouter.js
```diff
@@ -133,20 +137,17 @@ export default Router.extend({
           ]
         };
       }
-      // Need to mimic IdxResponses as idx returns raw response at error case
-      return this.handleIdxResponseSuccess({
-        rawIdxState: error.details,
-        context: _.pick(error.details, 'messages'),
-        neededToProceed: [],
-      });
+      // Pull all nested messages to the root context
```
haishengwu-okta, line 140: why needs to do this? (responseTransformer shall have done it?)

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -18,6 +18,29 @@ const Body = BaseForm.extend({
...
+    const messages = this.options.appState.get('messages') || {};
+    if (Array.isArray(messages.value)) {
+      this.add('<div class="ion-messages-container”></div>', '.o-form-error-container');
+
+      messages
+        .value
+        .forEach(messagesObj => {
+          const msg = messagesObj.message;
+          if (messagesObj.class === 'ERROR') {
+            this.add(createCallout({
```
pradeepdewda-okta, line 33: Suppose if there are multiple errors.. is this going to add multiple callouts ??

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
haishengwu-okta: Are you able to leverage `form error event`? (`FormController.showFormErrors`)

src/v2/BaseLoginRouter.js
jmelberg-okta: This is needed for the errors we create by hand - specifically the `oie.feature.disabled`, and `oform.error.unexpected` generic errors.

As an alternative, I can populate the `context` above during error creation - if we think that's more straightforward.

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
```diff
@@ -18,6 +18,29 @@ const Body = BaseForm.extend({
...
   },
 
+  showMessages () {
+    // IdP Authenticator error messages are not form errors
+    // Parse and display them here.
+    const messages = this.options.appState.get('messages') || {};
+    if (Array.isArray(messages.value)) {
+      this.add('<div class="ion-messages-container”></div>', '.o-form-error-container');
+
+      messages
```
pradeepdewda-okta, line 28: Do you think it would be better to enhance `BaseForm` `showMessages` function to cover this case ??

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
jmelberg-okta: Not that I could see, as form events rely on an action to be invoked (i.e. submit pressed). Since these errors occur on initial load - a user hasn't interacted with the form yet.

pradeepdewda-okta: Baseform should have handled such errors on initial load.. 
https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/internals/BaseForm.js#L32

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
jmelberg-okta: Great question, and I believe it will.For this flow, we shouldn't have more than one error message - so I decided to copy the behavior used in other authenticators:
-  [Challenge Okta Verify (v2)](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js#L17-L37)
- [Terminal View](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/TerminalView.js#L54-L75)

As mentioned in [this comment](https://github.com/okta/okta-signin-widget/pull/1738/files/53a50163a5bcf39edb963e5a5f4d3ea672a54114#r592554832) I think there is value in moving all error logic to a single handler as a part of a `BaseAuthenticatorView`. When we do this, we can loop in UX to see if we should concatenate all the error messages together, or keep this separate as we're currently doing.

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
jmelberg-okta: @pradeepdewda-okta - In this method I provided a comment explaining why we cannot rely on the `BaseForm`'s implementation. We need to provide an override to surface callouts over plain text on the page.

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
jmelberg-okta: Yes! That would be ideal. However, I need to understand the history of the `BaseForm` implementation and why we and UX signed off on errors being surfaced in that way. Adding this as a fast-follow action item for Widget cleanup!

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
haishengwu-okta: - First of all, we need to consolidate/better approach to handle errors. seems too many variants.
- What I was thinking was 
```
postRender() {
  this.model.trigger('error', ...);
}
```

src/v2/BaseLoginRouter.js
haishengwu-okta: ah, I see. yeah, seems more straightforward to set `context` (strictly mimic what idx-js response is)

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
haishengwu-okta: Much appreaited!!

nikhilvenkatraman-okta: IIRC the showMessages in the BaseForm by default will render messages as a text. This is used in a few scenarios example password expiry message which can be reproduced using `authenticator-expiry-warning-password.json`

Screen Shot 2021-03-11 at 1 01 18 PM

In this case the "class":"INFO". When we clean Baseform to consolidate all the error logic, up we can make use of this class attribute to render text if its `INFO` and render the error callout if class is `ERROR`

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
jmelberg-okta: > First of all, we need to consolidate/better approach to handle errors. seems too many variants.

💯 ! We should be able to consolidate all of our `showMessage` functionality to a single `BaseForm` function. A quick POC of this shows that as long as the error message contains a `class` - this can be mapped properly:
- `ERROR`: Render an error callout
- `INFO`: Display error message as raw text on the form

We'll need to account for a few cases, which is why this isn't a trivial task:
-  Errors with `i18n` text should default to retrieving this value as the error content, falling-back to the API message if it cannot be located. Today, we have mixed implementations of this.
- Errors that are intended to be [displayed as](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/IdentifierView.js#L70) "warning" types. This could always be an override, but ideally we'll find a better solution.

I'll add these notes to a JIRA.

nikhilvenkatraman-okta: We also have a use case where we want to show the title with the error message. The current approach is to override the `showMessages` method and do some custom logic (check i18n key to add title)

haishengwu-okta (approved):

src/v2/view-builder/views/idp/AuthenticatorIdPView.js
jmelberg-okta: Created [OKTA-377735](https://oktainc.atlassian.net/browse/OKTA-377735) as a follow-up task.

nikhilvenkatraman-okta (approved):

aarongranick-okta (approved):

PR #1742, opened 2021-03-11
feat: trap errors from widget event handlers - OKTA-367626

shuowu:
## Description:

Trap and log event handler errors from customer apps based on the `logging` config. Defaults to log the error in browser console.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

throw error from event handler
```
signIn.on('ready', () => {
  // handle `ready` event.
  // use `console.log` in particular so that those logs can be retrived
  // in testcafe for assertion
  console.log('===== playground widget ready event received =====');
  throw new Error('event error');
});
```

Error is trapped and the widget still works properly.

Screen Shot 2021-03-12 at 12 54 36 PM

### Issue:

- [OKTA-367626](https://oktainc.atlassian.net/browse/OKTA-367626)

### DIFF
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -151,7 +151,9 @@ var OktaSignIn = (function () {
             try {
               callback.apply(this, callbackArgs);
             } catch (err) {
-              DEBUG && Logger.error(`[okta-signin-widget] "${event}" event handler error:`, err);
+              if (options.logging?.logEventsError !== false) {
+                Logger.error(`[okta-signin-widget] "${event}" event handler error:`, err);
+              }
             }
           };
         }
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -151,7 +151,7 @@ var OktaSignIn = (function() {
             try {
               callback.apply(this, callbackArgs);
             } catch (err) {
-              if (options.logging?.logEventsError !== false) {
+              if (options.logging?.external !== false) {
                 Logger.error(`[okta-signin-widget] "${event}" event handler error:`, err);
               }
             }
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -149,7 +149,7 @@ var OktaSignIn = (function () {
         if (EVENTS_LIST.includes(event)) {
           onArgs[1] = function (...callbackArgs) {
             try {
-              callback.apply(null, callbackArgs);
+              callback.apply(this, callbackArgs);
             } catch (err) {
               Logger.error(`Error happens on ${event} event:`, err);
             }
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -143,7 +143,7 @@ var OktaSignIn = (function () {
     }
 
     _.extend(this, Router.prototype.Events, {
-      on: function(...onArgs) {
+      on: function (...onArgs) {
         // custom events listener on widget instance to trap third-party callback errors
         const [event, callback] = onArgs;
         if (EVENTS_LIST.includes(event)) {
@@ -153,7 +153,7 @@ var OktaSignIn = (function () {
             } catch (err) {
               Logger.error(`Error happens on ${event} event:`, err);
             }
-          }
+          };
         }
         Router.prototype.Events.on.apply(this, onArgs);
       }
--- a/src/widget/OktaSignIn.js
+++ b/src/widget/OktaSignIn.js
@@ -1,13 +1,16 @@
 import _ from 'underscore';
 import Errors from 'util/Errors';
 import Util from 'util/Util';
+import Logger from 'util/Logger';
 import createAuthClient from 'widget/createAuthClient';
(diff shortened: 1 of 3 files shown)


src/widget/OktaSignIn.js
```diff
@@ -2,12 +2,15 @@
...
 import Errors from 'util/Errors';
 import Util from 'util/Util';
+import Logger from 'util/Logger';
 import createAuthClient from 'widget/createAuthClient';
 import buildRenderOptions from 'widget/buildRenderOptions';
 import createRouter from 'widget/createRouter';
 import V1Router from 'LoginRouter';
 import V2Router from 'v2/WidgetRouter';
 
+const EVENTS_LIST = ['ready', 'afterError', 'afterRender', 'pageRendered'];
```
haishengwu-okta, line 12: `pageRendered` is deprecated. don't think we shall care about it.

haishengwu-okta (commented):
@aaronbrodersen-okta @shuowu @nikhilvenkatraman-okta  looks like this is same issue that happened during unit test, if any unexpected exception happened, we didn't see any error from console hence quite hard to pin down the root cause quickly.

Usually we'll see it's been caught by Q library around
```
try {
            return typeof fulfilled === "function" ? fulfilled(value) : value;
        } catch (exception) {
            return reject(exception);
        }
```

I think the reason is `BaseLoginRouter.render` run in promise chain hence any unexpected exception will be caught by Q.
I'm wondering if adding log to `BaseLoginRouter.render` (around line 268, the `catch` callback) would be good enough?

shuowu:
@haishengwu-okta The goal of this change is to make sure third-party event callbacks do not tear down the widget (which causes confusion for both devs and support), and the error should be properly logged in the console. The errors can be categorized as `sync errors` and `async errors`. With debugging, only the `sync errors` cause the problem, the change here is to provide a customized third-party listener function (on) that can trap and log `sync errors` properly.

> I'm wondering if adding log to BaseLoginRouter.render (around line 268, the catch callback) would be good enough?

Debugged this suggestion, looks like the error cannot be trapped there.

haishengwu-okta:
I see. I didn't think of `trap` errors but just log to the console. It's critical to tear down widget when devs do something wrong and force them to fix.
I guess the argument is for those unexpected runtime error, shall we just log it and stop the world or keep widget working (but widget may or may not behavior as expected). If we stop the world, **maybe** developer could help discover the error earlier at development time.

shuowu:
I think we can think about the problem separately, as internal and external errors.

For internal error, it can be errors that happen inside the widget (most probably introduce by our team), for this kind of error, I agree that the widget **should** stop working, and let the devs figure out what happened, other than introduce unexpected behaviours silently.

For external error, like the errors from event handlers, I don't see how it can affect the behaviour of the widget, but stop the widget due to this kind of errors may confuse the developers that something wrong with the widget. In this case, log the error with proper stack trace maybe can convey clear information to the devs.

pradeepdewda-okta:
@shuowu 
Would it be possible to show these console errors only when widget is in development mode ? This error stack is visible to anyone makes me cringe.

src/widget/OktaSignIn.js
```diff
@@ -127,7 +130,22 @@ var OktaSignIn = (function () {
...
+    _.extend(this, Router.prototype.Events, {
+      on: function (...onArgs) {
+        // custom events listener on widget instance to trap third-party callback errors
+        const [event, callback] = onArgs;
+        if (EVENTS_LIST.includes(event)) {
+          onArgs[1] = function (...callbackArgs) {
+            try {
+              callback.apply(this, callbackArgs);
+            } catch (err) {
+              DEBUG && Logger.error(`[okta-signin-widget] "${event}" event handler error:`, err);
```
shuowu, line 142: @pradeepdewda-okta Comment addressed

haishengwu-okta: Isn't the intent to show error at product build so that developer can discover the error during development?

haishengwu-okta:
> For external error, like the errors from event handlers, I don't see how it can affect the behaviour of the widget...

IMHO, the problem with this approach is that customers would assume widget+their customization code works properly and may discover the issue after deployment to their users v.s. discover the issue at development phase.

Maybe we could chat more during SIW meeting ?

test/unit/spec/OktaSignIn_spec.js
```diff
@@ -250,6 +253,25 @@ Expect.describe('OktaSignIn initialization', function() {
...
+      });
+    });
+    it('does not trap errors non-registered events', () => {
+      try {
+        signIn.on('not-widget-event', function() {
+          throw new Error('mockerror');
+        });
+        signIn.trigger('not-widget-event');
+        expect(Logger.error).not.toHaveBeenCalled();
+      } catch (err) {} // eslint-disable-line no-empty
```
aarongranick-okta, line 273: having a catch block may trap and ignore unexpected errors.  Try using `bind` to set up the function call and then use expect(fn).toThrow(...) to assert the specific mock error was thrown

aarongranick-okta (approved):
see 1 comment about test

PR #1780, opened 2021-03-25
Refactoring browser redirect handler

haishengwu-okta:
## Description:

1. add `useRedirect` prop to Model to allow individual view to override the behavior for form submit.
2. refactoring the idp authenticator to follow `enroll-` & `verify-` pattern like other authenticators.

## PR Checklist

- [x] Added unit tests?
- [x] Added e2e tests

### Issue:

- OKTA-383681


.bacon.yml
```diff
@@ -1,57 +1,57 @@
...
-      queue_name: small
\ No newline at end of file
+      - name: lint
+        script_path: /root/okta/okta-signin-widget/scripts
+        sort_order: "1"
+        timeout: "10"
+        script_name: lint
+        criteria: MERGE
+        queue_name: small
+      - name: unit-karma
```
jmelberg-okta, line 9: Is this an intended change?

package.json
```diff
@@ -34,9 +29,9 @@
     "lint": "mkdir -p build2 && run-p -c lint:eslint lint:grunt lint:stylelint",
     "lint:report": "mkdir -p build2 && run-p -c lint:eslint:report lint:grunt:report lint:stylelint:report",
     "test:jest": "grunt assets && yarn jest",
-    "test:dev": "grunt assets && karma start",
-    "test:coverage": "yarn test --reporters=progress,coverage",
-    "test": "yarn test:dev --single-run",
+    "test:jest:coverage": "yarn test:jest --collectCoverage",
```
jmelberg-okta, line 32: Adding an extra script to capture coverage seems unnecessary. Instead, what if we took the `okta-ui` approach and added a README note about passing through `jest` options?

src/v2/controllers/FormController.js
```diff
@@ -137,8 +137,7 @@ export default Controller.extend({
     const actionFn = idx['actions'][actionPath];
 
     if (_.isFunction(actionFn)) {
-      // TODO: OKTA-243167
-      // 1. what's the approach to show spinner indicating API in fligh?
+      // TODO: OKTA-243167 what's the approach to show spinner indicating API in fligh?
```
jmelberg-okta, line 140: While we're here: 
```suggestion
      // TODO: OKTA-243167 what's the approach to show spinner indicating API in flight?
```

src/v2/controllers/FormController.js
```diff
@@ -151,41 +150,39 @@ export default Controller.extend({
...
     // NOTE: okta-idx-js does not know whether a remediation
     // is required to redirect or perform a GET request.
     // As a result, we handle this accordingly by checking the formName
-    const fullPageRedirectForms = [ FORMS.REDIRECT_IDP ];
-    if (fullPageRedirectForms.includes(formName)) {
-      const currentViewState = this.options.appState.getCurrentViewState();
+    // TODO: seems better to check `currentViewState?.rel`, but
+    // redirect-idp form does not have `rel`
+    // device-apple-sso-extension form does have `rel` and still need redirect
+    const useFormRedirect = !currentViewState?.hasOwnProperty('method')
```
jmelberg-okta, line 167: nit: Let's avoid prototype built-ins: https://eslint.org/docs/rules/no-prototype-builtins

src/v2/controllers/FormController.js
```diff
@@ -151,41 +150,39 @@ export default Controller.extend({
...
-      //      - UNLOCK_ACCOUNT
-      const operation = FORM_NAME_TO_OPERATION_MAP[formName];
-      modelJSON.identifier = this.settings.transformUsername(identifier, operation);
+    // Run hook: transform the user name (a.k.a identifier)
+    const modelJSON = this.transformIdentifier(formName, model);
+
+    // Error out when this is not a remediation form. Unexpected Exception.
+    const idx = this.options.appState.get('idx');
+    if (!this.options.appState.hasRemediationObject(formName)) {
+      this.options.settings.callGlobalError(`Cannot find http action for "${formName}".`);
```
jmelberg-okta, line 180: IIRC `callGlobalError` gets bubbled up to an unrecoverable error state to the end-user. Shouldn't this value (and the form error below) be localized?

src/v2/controllers/FormController.js
```diff
@@ -217,21 +214,36 @@ export default Controller.extend({
       });
   },
 
+  transformIdentifier(formName, model) {
```
jmelberg-okta, line 217: nit: Since we're refactoring - how about extracting out this method to a `util` file?

src/v2/view-builder/views/SuccessView.js
```diff
@@ -29,10 +28,7 @@ const Body = BaseForm.extend({
   noButtonBar: true,
   initialize() {
     BaseForm.prototype.initialize.apply(this, arguments);
-    // TODO OKTA-250473
-    // Form post for success redirect
-    const url = this.options.currentViewState.href;
-    Util.redirectWithFormGet(url);
+    this.trigger('save', this.model);
```
jmelberg-okta, line 31: 🎉  - One of the benefits to moving this logic to `handleSaveForm`!

src/v2/view-builder/views/idp/AuthenticatorIdPVerifyView.js
```diff
@@ -42,21 +38,11 @@ const Body = BaseForm.extend({
...
 
 export default BaseAuthenticatorView.extend({
-  initialize() {
-    BaseAuthenticatorView.prototype.initialize.apply(this, arguments);
-
-    // The IdP Authenticator doesn't have a traditional enroll/challenge form.
-    // We need to infer it based on if currentAuthenticatorEnrollment is included in the response
-    const currentAuthenticatorEnrollment = this.options.appState.get('currentAuthenticatorEnrollment');
-    this.options.isChallenge = !!currentAuthenticatorEnrollment;
-  },
```
jmelberg-okta, line 60: My motivation for keeping this as a single file was to reduce the duplicate code between enroll/challenge views. Some authenticators (Duo, IdP, Symantec) have nearly identical code, only changing the title, subtitle, and save form labels.

This may have been an over-optimization to condense code, but let's have a conversation about ways to solve this problem. At a high level:

- Authenticators always have an enroll/challenge view state. Ideally, the title, subtitle, and save form labels (`Enroll in {0}`/`Verify with {0}`, `Enroll`/`Verify`) should be constant throughout all authenticators.
- Form and top-level (`messages` present in response) error messages should be shown when present. This is opt-in today, as errors render as basic text on the page.
- Header/footer logic is also relatively constant throughout all authenticators.
- The decision to show the enroll or challenge page can be inferred based on whether there is a `currentAuthenticatorEnrollment`. This has been moved to an `appState` helper called [`isAuthenticatorChallenge `](https://github.com/okta/okta-signin-widget/blob/ff47bcfd36a5a44d3fff465c6407ba883acb774e/src/v2/models/AppState.js#L122-L129).

package.json
haishengwu-okta: Yeah, we could update README.

src/v2/view-builder/views/idp/AuthenticatorIdPVerifyView.js
haishengwu-okta: It's sort of deficit the dispatch logic in ViewFactory (hence going to be a little redundant work)
The ViewFactory already has such logic like 
- `if is challenge, use ViewA; if is enroll, use ViewB`

What you suggest here is to push down the dispatch one level down. (see the `if is challenge else` runs twice)
- In ViewFactory, `if is challenge, use ViewA; if is enroll, use ViewA`
- In ViewA, `if is challenge, does X; if is enroll, does B`

In this draft, The `messages` method is duplicated but it's easy to get rid of via inheritance: creates a base class for IdPAuthenticator.

src/v2/controllers/FormController.js
haishengwu-okta: Good catch!!
Would the tool you or Nikhil developed be able to catch this issue?

src/v2/controllers/FormController.js
haishengwu-okta: could be (would need another parameter `this.settings`).

.bacon.yml
haishengwu-okta: sort of.. I had config my IDE to auto format files (per file type) hence the changes..
revert it to minimize the changes.

src/v2/view-builder/views/idp/AuthenticatorIdPVerifyView.js
haishengwu-okta: Doesn't seem conflict with your goal to change more general label like `Verify with {0}`? 
I'm thinking we would have

- BaseEnrollAuthenticatorForm
- BaseVerifyAuthenticatorForm

which take care of Title/Subtitle/Save for all authenticators.

For handling `messages`, hopefully we could take care it BaseForm/BaseView with maybe couple of special case in individual View level.

src/v2/controllers/FormController.js
jmelberg-okta: It won't - as our rules apply for only Courage templates and known components (`createCallout`, etc.).

Adding a lint rule for ensuring `callGlobalError` receives a `function` could help us resolve future instances of this! cc: @nikhilvenkatraman-okta

src/v2/controllers/FormController.js
jmelberg-okta: Ah - missed this! We can always approach this later.

src/v2/view-builder/views/idp/AuthenticatorIdPVerifyView.js
jmelberg-okta: ☝🏼 These are all excellent points and resolves a variety of problems. Let's scope the inheritance modeling work for another time!

test/testcafe/spec/Interact_spec.js
```diff
@@ -71,7 +71,9 @@ const errorMock = RequestMock()
   .onRequestTo('http://localhost:3000/oauth2/default/.well-known/openid-configuration')
   .respond(xhrWellKnownResponse, 200)
   .onRequestTo('http://localhost:3000/oauth2/default/v1/interact')
-  .respond(xhrServerError, 400);
+  .respond(xhrServerError, 400)
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
```
haishengwu-okta, line 75: Avoid to depend default behavior of playground.

src/v2/controllers/FormController.js
nikhilvenkatraman-okta: Good callout. We can track localizing global errors separately https://oktainc.atlassian.net/browse/OKTA-383615

src/v2/controllers/FormController.js
```diff
@@ -153,39 +152,29 @@ export default Controller.extend({
...
-    const idx = this.options.appState.get('idx');
-    if (!idx['neededToProceed'].find(item => item.name === formName)) {
-      this.options.settings.callGlobalError(`Cannot find http action for "${formName}".`);
-      this.showFormErrors(this.formView.model, 'Cannot find action to proceed.');
-      return;
-    }
-
+    // Toggle Form saving status (e.g. disabling save button, etc)
     this.toggleFormButtonState(true);
     model.trigger('request');
```
jmelberg-okta, line 157: nit: Should these be handled via the controller for BaseForm/View? Seems like more of a form-level concern to me! 😄

src/v2/controllers/FormController.js
```diff
@@ -153,39 +152,29 @@ export default Controller.extend({
...
 
-    // NOTE: okta-idx-js does not know whether a remediation
-    // is required to redirect or perform a GET request.
-    // As a result, we handle this accordingly by checking the formName
-    const fullPageRedirectForms = [ FORMS.REDIRECT_IDP ];
-    if (fullPageRedirectForms.includes(formName)) {
+    // Use full page redirection if necessary
+    if (model.get('useRedirect') === true) {
```
jmelberg-okta, line 160: The `===` is unnecessary, right?

src/v2/ion/responseTransformer.js
```diff
@@ -155,11 +155,38 @@ const convertRedirectIdPToSuccessRedirectIffOneIdp = (result) => {
...
+  if (Array.isArray(result.remediations)) {
+    const idpAuthenticator = result.remediations.filter(
+      remediation => {
+        return remediation.name === RemediationForms.REDIRECT_IDP
+          && remediation?.relatesTo?.value?.key === 'external_idp';
+      }
+    );
+
+    if (idpAuthenticator.length === 1) {
+      const isVerifyFlow = result.hasOwnProperty('currentAuthenticatorEnrollment');
```
jmelberg-okta, line 182:
```suggestion
      const isVerifyFlow = Object.prototype.hasOwnProperty.call(result, 'currentAuthenticatorEnrollment');
```

src/v2/controllers/FormController.js
```diff
@@ -217,21 +206,36 @@ export default Controller.extend({
       });
   },
 
+  transformIdentifier(formName, model) {
+    const modelJSON = model.toJSON();
+    if (modelJSON.hasOwnProperty('identifier')) {
```
jmelberg-okta, line 211:
```suggestion
    if (Object.prototype.hasOwnProperty.call(modelJSON, 'identifier')) {
```

test/testcafe/spec/ChallengeOktaVerifySSOExtensionView_spec.js
```diff
@@ -12,7 +13,12 @@ const credentialSSOExtensionMock = RequestMock()
   .onRequestTo(/idp\/idx\/introspect/)
   .respond(identifyUserVerificationWithCredentialSSOExtension)
   .onRequestTo(verifyUrl)
-  .respond(identify);
+  .respond((req, res) => {
+    return new Promise((resolve) => setTimeout(function() {
+      res.setBody(identify);
+      resolve(res);
+    }, Constants.TESTCAFE_DEFAULT_AJAX_WAIT + 1000));
```
jmelberg-okta, line 20: Why is this wait needed?

jmelberg-okta (approved):
Looks pretty good! Couple of minor comments.

src/v2/controllers/FormController.js
pradeepdewda-okta: or `"identifier" in modelJSON`

src/v2/controllers/FormController.js
haishengwu-okta: Good point! I'll make sure it's got addressed at my next PR.

test/testcafe/spec/ChallengeOktaVerifySSOExtensionView_spec.js
haishengwu-okta: Add comments at line 51. 

Here is what testcafe does (my understanding)

1. `await ssoExtensionPage.navigateToPage();`  --->  wait **3s** because there is ajax in flight.
2. continue next expression.

(3s is hard-code by testcafe)

Let's say the ajax takes less than 3s, then while testcafe is wait (step 1), ajax finished and widget will proceed to next view (success view for instance), so when it runs into step 2, widget already at success view hence no way to verify ssoExtensionPage.
The trick it did before is to error out the ajax (remove `method` from API mock hence `idx.proceed` will show error therefore we stuck on the ssoExtensionPage and can verify page elements.

src/v2/controllers/FormController.js
haishengwu-okta: The problem was there are several cases  `saveForm` was triggered on `appState` directly instead of `form.trigger('save')`, e.g. polling. The nuance is whether want to clean up error (`form.trigger(save)` does clean up form errors).
Need to dig into further and may require more refactoring, like move `handeSaveForm` into BaseForm.

PR #1783, opened 2021-03-26
Adds device fingerprinting to OIE

jmelberg-okta:
## Description:

- Adds device fingerprinting support for [New Device Notifications on OIE](https://oktawiki.atlassian.net/wiki/spaces/~yuming.cao/pages/1845726231/New+device+sign-in+notification+checking+logic+on+OIE?focusedCommentId=1912082844#OIE-Fingerprint).

### DIFF
--- a/playground/mocks/spec-okta-api/auth/services/devicefingerprint.js
+++ b/playground/mocks/spec-okta-api/auth/services/devicefingerprint.js
@@ -1,0 +1,30 @@
+const apis = [
+  {
+    proxy: false,
+    path: '/auth/services/devicefingerprint',
+    method: 'GET',
+    template() {
+      // In a real case, `/auth/services/devicefingerprint` is loaded within an
+      // invisible iframe and generates a unique identifier for New Device Notifications
+      return `
+        <html>
+          <script>
+            const message = JSON.stringify({
+              type: 'FingerprintAvailable',
+              fingerprint: 'mock-device-fingerprint',
+            });
+            window.parent.postMessage(message, window.location.href);
+          </script>
+        </html>
+      `;
+    },
+    render(req, res) {
+      // Manually override the CSP for device-fingerprinting mocking
+      // This should never be performed in a production app
+      res.header('Content-Security-Policy', 'script-src \'unsafe-inline\'');
+      res.send(res.body);
+    }
+  },
+];
+
+module.exports = apis;
--- a/src/v2/client/introspect.js
+++ b/src/v2/client/introspect.js
@@ -12,10 +12,20 @@
 
 import idx from '@okta/okta-idx-js';
 
-export async function introspect(settings) {
+export async function introspect(settings, appState) {
   const domain = settings.get('baseUrl');
   const stateHandle = settings.get('stateToken');
   const version = settings.get('apiVersion');
+
+  // For certain flows, we need to generate a device fingerprint
+  // to determine if we need to send a "New Device Sign-on Notification".
+  // In the future, this should be handled by okta-auth-js
+  idx.client.interceptors.request.use(requestConfig => {
+    const fingerprint = appState.get('deviceFingerprint');
+    if (fingerprint) {
+      requestConfig.headers['X-Device-Fingerprint'] = fingerprint;
+    }
+  });
 
   return idx.start({ domain, stateHandle, version });
 }
--- a/src/v2/client/startLoginFlow.js
+++ b/src/v2/client/startLoginFlow.js
@@ -23,7 +23,7 @@ const handleProxyIdxResponse = async (settings) => {
   });
 };
 
-export async function startLoginFlow(settings) {
+export async function startLoginFlow(settings, appState) {
   // Return a preset response
   if (settings.get('proxyIdxResponse')) {
     return handleProxyIdxResponse(settings);
@@ -32,7 +32,7 @@ export async function startLoginFlow(settings) {
   // Use stateToken
   const stateHandle = settings.get('stateToken');
   if (stateHandle) {
-    return introspect(settings);
+    return introspect(settings, appState);
   }
   
   // Use or acquire interactionHandle
--- a/src/v2/view-builder/utils/DeviceFingerprinting.js
+++ b/src/v2/view-builder/utils/DeviceFingerprinting.js
@@ -1,0 +1,106 @@
+/*!
+ * Copyright (c) 2021-Present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { $ } from 'okta';
+
+function getUserAgent() {
+  return navigator.userAgent;
+}
+
+function isWindowsPhone(userAgent) {
+  return userAgent.match(/windows phone|iemobile|wpdesktop/i);
+}
+
+export default {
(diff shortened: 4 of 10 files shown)


src/v2/client/startLoginFlow.js
```diff
@@ -32,7 +32,7 @@ export async function startLoginFlow(settings) {
   // Use stateToken
   const stateHandle = settings.get('stateToken');
   if (stateHandle) {
-    return introspect(settings);
+    return introspect(settings, appState);
```
haishengwu-okta, line 35: when widget begin to introspect, `appState` shall be empty. Why we need it?

playground/mocks/spec-okta-api/auth/services/devicefingerprint.js
```diff
@@ -0,0 +1,30 @@
...
+              fingerprint: 'mock-device-fingerprint',
+            });
+            window.parent.postMessage(message, window.location.href);
+          </script>
+        </html>
+      `;
+    },
+    render(req, res) {
+      // Manually override the CSP for device-fingerprinting mocking
+      // This should never be performed in a production app
```
haishengwu-okta, line 23: how does it work in product without CSP override?

src/v2/client/startLoginFlow.js
haishengwu-okta: NVM, found the usage.

src/v2/client/introspect.js
```diff
@@ -12,10 +12,24 @@
...
+export async function introspect(settings, appState) {
   const domain = settings.get('baseUrl');
   const stateHandle = settings.get('stateToken');
   const version = settings.get('apiVersion');
+  const useDeviceFingerprint = settings.get('features.deviceFingerprinting');
+
+  // For certain flows, we need to generate a device fingerprint
+  // to determine if we need to send a "New Device Sign-on Notification".
+  // In the future, this should be handled by okta-auth-js
+  idx.client.interceptors.request.use(requestConfig => {
```
haishengwu-okta, line 24: maybe only add `interceptor` iff `useDeviceFingerprint` is `true`?

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -16,6 +17,33 @@ const Body = BaseForm.extend({
     return loc('oform.next', 'login');
   },
 
+  saveForm() {
+    // Ideally this can be added to a "preSaveForm" handler - but keeping this here for now.
+    if (!this.settings.get('features.deviceFingerprinting')) {
```
haishengwu-okta, line 22: `features.deviceFingerprinting` has been checked twice: one here and another in the interceptor config.
Similarly for checking `/idp/idx/identify` (we modify IdentifyView hence we already know figure-print coming from identify step but we check again at interceptor.)
I'd try to think if there is way to eliminate such redundancy. 
**maybe** in interceptor, we only need to check the availability of `deviceFingerprint` ?

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -16,6 +17,33 @@ const Body = BaseForm.extend({
     return loc('oform.next', 'login');
   },
 
+  saveForm() {
```
haishengwu-okta, line 20: We have briefly discussed to add general handle to BaseView (e.g. `preSave`) and FormController could invoke it before invoking `idx.proceeded`. Did you get chance to explore such idea?

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -16,6 +17,33 @@ const Body = BaseForm.extend({
...
+      BaseForm.prototype.saveForm.apply(this, arguments);
+      return;
+    }
+
+    // Before the XHR is made for "identify", we'll generate this one-time use fingerprint via
+    // a hidden-iframe (similar to authn/v1 flows)
+    const fingerprintData = {
+      oktaDomainUrl: this.settings.get('baseUrl'),
+      element: this.$el,
+      stateHandle: this.options.appState.get('idx')?.rawIdxState?.stateHandle,
```
haishengwu-okta, line 32: does `this.options.appState.get('stateHandle')` or `this.options.currentViewState.stateHandle` give same result?
try to avoid `rawIdxState`

src/v2/view-builder/views/IdentifierView.js
jmelberg-okta: Great suggestion - we should make sure the `IdentifierView` keeps this decision logic, so I updated this PR to remove the duplicate checks.

src/v2/view-builder/views/IdentifierView.js
jmelberg-okta: I did! Played around with a couple implementations, but couldn't settle on something scalable. Will discuss this offline with the team during our next sync.

src/v2/client/introspect.js
```diff
@@ -12,10 +12,23 @@
...
 
+  // For certain flows, we need to generate a device fingerprint
+  // to determine if we need to send a "New Device Sign-on Notification".
+  // In the future, this should be handled by okta-auth-js
+  idx.client.interceptors.request.use(requestConfig => {
+    const fingerprintRequestUrls = [
+      `${domain}/idp/idx/identify`,
+    ];
+    const fingerprint = appState.get('deviceFingerprint');
+    if (fingerprint && fingerprintRequestUrls.includes(requestConfig.url)) {
```
haishengwu-okta, line 28: what about even remove the url check?

haishengwu-okta (approved):
One question & LGTM

src/v2/client/introspect.js
jmelberg-okta: We actually need to keep this here - as we only want to **add the header** for this specific endpoint. Our presentation layer only notifies our app state that we have a device fingerprint, not to act on it.

haishengwu-okta: I was thinking there is no way (assume bug free) for `fingerprint` is defined but not at `identify` step?
Let's see tomorrow we want finger print for `/enroll`, we 'll need to modify Enroll View as well as add URL check. When 2 changes happen at same time, usually it's code smell ?

src/v2/view-builder/utils/DeviceFingerprinting.js
```diff
@@ -0,0 +1,106 @@
...
+          handleError('no data');
+          return;
+        }
+        try {
+          const message = JSON.parse(event.data);
+
+          if (message && message.type === 'FingerprintServiceReady') {
+            sendMessageToOkta({
+              type: 'GetFingerprint',
+              stateHandle: fingerprintData.stateHandle,
```
mauriciocastillosilva-okta, line 70: Is there a PR on the server service to support `stateHandle`? 

Question 2: is this the only change from v1 file?

mauriciocastillosilva-okta (commented):
looks good, just a couple of questions

src/v2/client/introspect.js
jmelberg-okta: This is a valid concern, and one we brought up with our server-side team. Luckily, `/identify` will be the **only** path that we're looking for this device fingerprint on OIE, as the backend will [include the fingerprint in the auth state](https://oktawiki.atlassian.net/wiki/spaces/~yuming.cao/pages/1845726231/New+device+sign-in+notification+checking+logic+on+OIE?focusedCommentId=1917097435#comment-1917097435) for all subsequent requests.

We _could_ find a way to clean up this code even more by moving the fingerprinting generation outside of the Identify view into this handler, but this seems like more of a view-level responsibility as we don't have access to the `$el` here.

What do you think?

src/v2/view-builder/utils/DeviceFingerprinting.js
jmelberg-okta: There is! You can see how we plan to support this [in #51397](https://github.com/okta/okta-core/pull/51397).

And yes, this new `DeviceFingerprinting` logic is embedded within `v2` and will **only** be triggered for OIE flows. There will be no impact to legacy fingerprinting Widget-side.

mauriciocastillosilva-okta: Sounds good. I meant to ask if this line is the only difference from the file used in v1, just to know exactly what are the differences.

mauriciocastillosilva-okta (approved):

PR #1791, opened 2021-03-30
email link vuln takeover

anipendakur-okta:
## Description:

* Email Link Vuln Takeover screens. -
* Figma - https://www.figma.com/file/C0WWCFlfFAT4yy60TkUOYp/Email-Interstitial?node-id=24%3A126
* Wiki - https://oktawiki.atlassian.net/wiki/spaces/EJ/pages/1884915878/API+Review+Email+link+consent+flow

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Consent screen -
image

On hitting allow btn -
image
On hitting deny btn -
image

The original tab where the user had clicked `Send me an email` link on denial - 
image

Video of consent flow - https://okta.box.com/s/6bhjhaa9xpu9cv3te2j8rrdlbtws3r9h
Video of consent denied flow - https://okta.box.com/s/331guvrdgios5eug03wz504dw4hu6w55
### Reviewers:
@haishengwu-okta 

### Issue:

- [OKTA-368273](https://oktainc.atlassian.net/browse/OKTA-368273)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -932,3 +932,8 @@ oie.feature.disabled = The requested feature is not enabled in this environment.
 # Consent
 oie.consent.admin.title = would like to access:
 oie.consent.enduser.title = would like to:
+oie.consent.enduser.accept.label = Yes, it's me
+oie.consent.enduser.deny.label = No, it's not me
+oie.consent.enduser.title = Did you just try to sign in?
+oie.consent.enduser.email.allow.description = To continue, return to the device or window where you requested the email link. You may close this window at any time.
```
anipendakur-okta, line 938: This is email specific, hence the key is intended to have `email` in it. The rest of the new properties need not be specific to email.

playground/mocks/data/idp/idx/terminal-return-email-consent-denied.json
```diff
@@ -0,0 +1,18 @@
+{
```
anipendakur-okta, line 1: SIW gets this remediation when the User clicks on the `No, it's not me` button from the `consent` screen.

playground/mocks/data/idp/idx/terminal-return-email-consent.json
```diff
@@ -0,0 +1,18 @@
+{
```
anipendakur-okta, line 1: SIW gets this terminal view remediation when the user clicks on `Yes, it's me` button on the `consent` screen.

playground/mocks/data/idp/idx/email-challenge-consent.json
```diff
@@ -0,0 +1,71 @@
+{
```
anipendakur-okta, line 1: This is the `consent` screen remediation. A widget will need to be bootstrapped and rendered with this remediation when the user clicks on `Sign in` link from their email.

src/v2/models/AppState.js
```diff
@@ -49,7 +49,7 @@ export default Model.extend({
       },
     },
     authenticatorMethodType: {
-      deps: ['currentAuthenticator', 'currentAuthenticatorEnrollment',],
+      deps: ['currentAuthenticator', 'currentAuthenticatorEnrollment'],
```
anipendakur-okta, line 52: Fixing a typo.

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
```diff
@@ -0,0 +1,91 @@
...
+export default BaseAuthenticatorView.extend({
+  className: 'enduser-email-consent',
+
+  Header: BaseHeader.extend({
+    HeaderBeacon: HeaderBeacon.extend({
+      getBeaconClassName: function() {
+        return getIconClassNameForBeacon(AUTHENTICATOR_KEY.EMAIL);
+      },
+    }),
+  }),
```
anipendakur-okta, line 18: This is needed because - 

1. This is a fresh SIW that is bootstrapped. There's no `authenticatorKey` set in `appState`. Also, `Beacon` is necessary here because the `TerminalView`displays Beacons based on the `i18n` key received in remediation.

2. One of the terminal states has been reused in this flow - `idx.return.to.original.tab` which already shows Email Beacon (`playground/mocks/data/idp/idx/terminal-return-email-consent.json`). So, if I change that flow to not show Email Beacon, I might be breaking other flows. Hence this was needed for consistency in this flow although UX preferred not to have Beacon for this.

playground/mocks/data/idp/idx/terminal-enduser-email-consent-denied.json
```diff
@@ -0,0 +1,18 @@
+{
```
anipendakur-okta, line 1: This is the remediation SIW gets in the first device after the user denied signin.

assets/sass/v2/_forms.scss
```diff
@@ -10,9 +10,11 @@
   }
 
   .siw-main-body {
+    .o-form-error-container {
+      margin-bottom: 1rem;
+    }
     .ion-messages-container {
       text-align: center;
-      margin-bottom: 1rem;
```
anipendakur-okta, line 15: Moving the margin to the parent container.

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
```diff
@@ -0,0 +1,91 @@
...
+          <div class="enduser-email-consent--info">
+            <i class="enduser-email-consent--icon app-icon"></i>
+            <div>${this.model.escape('app')}</div>
+          </div>
+        `);
+      }
+    },
+    getUISchema() {
+      const uiSchemas = BaseForm.prototype.getUISchema.apply(this, arguments);
+      return uiSchemas.filter(uiSchema => uiSchema.name !== 'consent' );
```
haishengwu-okta, line 59: what else besides `consent` in the uiSchema?

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
```diff
@@ -0,0 +1,91 @@
...
+  _getInfoObject(requestInfoArray, requestInfoName) {
+    return requestInfoArray.find(({ name }) => (name === requestInfoName) );
+  },
+
+  createModelClass({ requestInfo }) {
+    const ModelClass = BaseAuthenticatorView.prototype.createModelClass.apply(this, arguments);
+    const { value: browser} = this._getInfoObject(requestInfo, 'browser');
+    const { value: app } = this._getInfoObject(requestInfo, 'appName');
+
+    const props = Object.assign({
```
haishengwu-okta, line 72: shall be `local`?

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
```diff
@@ -0,0 +1,91 @@
...
+    },
+    save() {
+      return loc('oie.consent.enduser.accept.label', 'login');
+    },
+    cancel() {
+      return loc('oie.consent.enduser.deny.label', 'login');
+    },
+    initialize() {
+      BaseForm.prototype.initialize.apply(this, arguments);
+      if (this.model.get('browser')) {
```
haishengwu-okta, line 40: nit:
are you able to use handlebar template?
```
tpl({{if browser}}...{{if app}...)(this.model.pick('browser', 'app'))
```

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
```diff
@@ -0,0 +1,91 @@
...
+
+  postRender() {
+    // Move buttons in DOM to match visual hierarchy to fix tab order.
+    // TODO https://oktainc.atlassian.net/browse/OKTA-350521
+    const buttonContainer = this.$el.find('.o-form-button-bar');
+    buttonContainer.find('.button-primary').appendTo(buttonContainer);
+    buttonContainer.find('.button-primary').removeClass('button-primary');
+  },
+
+  Body: ConsentViewForm.extend({
```
haishengwu-okta, line 28: nit: 
define as separated variable
`const Body = ConsentViewForm.extend({...})`

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
```diff
@@ -0,0 +1,91 @@
...
+  }),
+
+  _getInfoObject(requestInfoArray, requestInfoName) {
+    return requestInfoArray.find(({ name }) => (name === requestInfoName) );
+  },
+
+  createModelClass({ requestInfo }) {
+    const ModelClass = BaseAuthenticatorView.prototype.createModelClass.apply(this, arguments);
+    const { value: browser} = this._getInfoObject(requestInfo, 'browser');
+    const { value: app } = this._getInfoObject(requestInfo, 'appName');
```
haishengwu-okta, line 70: looks like you can just reduce the `requestInfo` array. (use reduce or forEach depends your preference.)

```
const local = requestInfo.reduce((init, obj) => {
  init[obj.name] = {type: obj.type, value.obj.value};
  return init;
}, {});

```

haishengwu-okta (commented):
Looks good to me!

playground/mocks/data/idp/idx/email-challenge-consent.json
```diff
@@ -0,0 +1,71 @@
...
+    "value":[{
+      "rel":[
+        "create-form"
+      ],
+      "name":"email-challenge-consent",
+      "href":"http://localhost:3000/idp/idx/consent",
+      "method":"POST",
+      "accepts":"application/json; okta-version=1.0.0",
+      "produces":"application/ion+json; okta-version=1.0.0",
+      "requestInfo":[
```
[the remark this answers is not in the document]
anipendakur-okta: Hmm. Not sure, probably because it needs to be displayed to the user, hence they're in `remediation`. May be we can push it in uiSchema and expand our uiSchema transformer to just display the text as they are typed `string`. (It could have come directly in the `value` array instead too)

src/v2/view-builder/views/TerminalView.js
```diff
@@ -33,6 +38,19 @@ const HeaderBeaconTerminal = HeaderBeacon.extend({
 const Body = BaseForm.extend({
   noButtonBar: true,
 
+  initialize() {
+    BaseForm.prototype.initialize.call(this, arguments);
+    let description;
```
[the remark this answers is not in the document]
anipendakur-okta: Definitely, but these are client side i18n description texts. `showMessages` already shows an error callout for `error` types and for `info`s it just displays the texts directly without a callout.

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
anipendakur-okta: There is only `consent` as of now. But in future, if something else gets added, that needs to be displayed. Since `consent` is a `boolean` I had to filter it out to not show it as `checkbox`.

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
anipendakur-okta: Sure, I can change this. In `toJSON`, I'm returning `{consent: this.get('consent')}` hence they get stripped and not sent to the API which will give the same effect as `local`.

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
anipendakur-okta: There seems to be a lot of restrictions with our `lint`. Looks like the only way I am able to do this is -
```
// One looong line - so I'll need an eslint directive to ignore long lines here.
// note- i can not do string concatenation for some odd reason??
// neither can i use backticks here. it fails.
const getInfo = hbs('{{#if browser}}<div class="enduser-email-consent--info"><i class="enduser-email-consent--icon browser-icon"></i><div>{{browser}}</div></div>{{/if}}{{#if app}}<div class="enduser-email-consent--info"><i class="enduser-email-consent--icon app-icon"></i><div>{{app}}</div></div>{{/if}}');
.
.
.
const info = getInfo(this.model.pick('browser', 'app'));
this.add(info);
```

It makes the `initialize` function definition looks nice and tidy. What do you think?

src/v2/models/AppState.js
anipendakur-okta: So, the other similar places don't have a trailing `,` in the same file. It just makes it consistent (doesn't make any difference as you already know)

playground/mocks/data/idp/idx/email-challenge-consent.json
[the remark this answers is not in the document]
haishengwu-okta: We can do that as well but I was more thinking `requestInfo` as contextual data like other information hence shall sit at top level.

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
haishengwu-okta: ah, I see. enduser / admin consent handled this is `ion-boolean-handler.js`

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
haishengwu-okta: Create another View? 
or
```
 hbs`
...
`
```

Try to search usage of `hbs` in `src`

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
anipendakur-okta: I think `find` will fit better for the context since it preemptively returns without going through the entire array. I have changed the code a bit and removed `_getInfoObject` helper fn now.

test/testcafe/spec/EnduserEmailConsentView_spec.js
```diff
@@ -0,0 +1,94 @@
...
+        url,
+      },
+    } = requestLogger.requests[requestLogger.requests.length - 1];
+
+    const { consent } = JSON.parse(body);
+    await t.expect(consent).eql(true);
+    await t.expect(method).eql('post');
+    await t.expect(url).eql('http://localhost:3000/idp/idx/consent');
+
+    const successPage = new SuccessPageObject(t);
```
haishengwu-okta, line 66: shall use `TerminalPageObject`?

test/testcafe/framework/page-objects/EnduserConsentPageObject.js
```diff
@@ -0,0 +1,27 @@
...
+
+  clickDontAllowButton() {
+    return this.form.clickCancelButton();
+  }
+
+  getInfoItemTexts() {
+    return this.form.getInnerTexts('.enduser-email-consent--info');
+  }
+
+  getSaveButtonLabel() {
```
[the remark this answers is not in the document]
anipendakur-okta: It does, but here I'm removing the `button-primary` class so it can't be used. (both buttons are secondary here as per UX)

src/v2/view-builder/views/consent/EnduserEmailConsentView.js
anipendakur-okta: @vincentvoong-okta - so TLDR; we need to have the email beacon in this flow because of shared terminal flows.

haishengwu-okta (approved):

test/testcafe/framework/page-objects/EnduserConsentPageObject.js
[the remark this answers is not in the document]
sherizada-okta: Okay first of all, `clickAllowButton` and `clickDontAllowButton` methods are already defined on ConsentPageObject, so you didn't need to define them again.

Second,
BaseFormObject uses these selectors:
```
const SUBMIT_BUTTON_SELECTOR = '.o-form-button-bar input[data-type="save"]';
const CANCEL_BUTTON_SELECTOR = '.o-form-button-bar input[data-type="cancel"]';
```
for its label getter methods, and you are already using these selectors in the `clickAllowButton` and `clickDontAllowButton` methods, so they could have bee re-used for this page.

PR #1795, opened 2021-03-30
OKTA-382782: Variable IDX Challenge Polling

worldleelavivat-okta:
## Description:

Variable IDX Challenge Polling

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Screen Shot 2021-03-31 at 5 00 47 PM

### Reviewers:
- @thomasmaslen-okta 
- @haoyuli-okta 
- @anipendakur-okta 
- @haishengwu-okta 
- @pradeepdewda-okta 

### Issue:

- [OKTA-382782](https://oktainc.atlassian.net/browse/OKTA-382782)


thomasmaslen-okta (commented):
I definitely like the fact that changes this small do the trick.

But let me check whether I'm in sync on the intent of this POC and how we should think of it when looking at it or commenting on it.

IIUC it is pretty much as you said in your Email message:

> Here is a quick POC PR with minimal amount of code to get SIW to play nice with variable refresh interval based off of Jane's PR

In particular, I'm assuming that I should think of the change in `AppState.js` as a neat hack but not something that we would propose to merge?  (Both because it is modifying the `AppState` code and because it checks `currentAuthenticatorEnrollment` but not `currentAuthenticator`)

Or am I thinking of this the wrong way and the intent is that yes, this can be the basis for a merge-worthy fix?

worldleelavivat-okta:
@thomasmaslen-okta The final change would still be something along this line since `AppState.js` is where we maintain the state of the app and where/when the data is changed. We might need to check/test more to find where the refresh property are in difference scenario. Current change is
`const refreshInterval = currentViewState.refresh ||
        transformedResponse.currentAuthenticatorEnrollment?.poll?.refresh;`

`currentViewState.refresh` is the current implementation and it would take precedented if it is defined.
`transformedResponse.currentAuthenticatorEnrollment?.poll?.refresh` are all optional chain called which could return `undefined` and it would just maintain the previous refresh interval.

Currently, `currentViewState.refresh` somehow always return `undefined` so the `change:dynamicRefreshInterval` event never fired so the SIW just use the initial interval it set initially.

haishengwu-okta (approved):

src/v2/models/AppState.js
```diff
@@ -141,8 +141,10 @@ export default Model.extend({
       // make sure to update the correct poll view's refresh value
       const currentFormName = this.get('currentFormName');
       const currentViewState = transformedResponse.remediations.filter(r => r.name === currentFormName)[0];
+      const refreshInterval = currentViewState.refresh ||
+        transformedResponse.currentAuthenticatorEnrollment?.poll?.refresh;
```
haishengwu-okta, line 145: there might be case when refresh exists at `transformedResponse.currentAuthenticator?.poll?.refresh`?

pradeepdewda-okta: can you add a comment what this is about ? and why a specific authenticator(view) is being used here

pradeepdewda-okta: Instead of here, can we tackle this somewhere specific to AuthenticatorEnrollment ? or in `BaseFormWithPolling`

worldleelavivat-okta: @haishengwu-okta I don't believe I ever see `refresh` in `transformedResponse.currentAuthenticator?.poll?.refresh` but @thomasmaslen-okta if you don't mind confirming.

@pradeepdewda-okta I can add some comment regarding the variables. But I think this should remain here in `AppState.js` since the logic is already here with the `AppState.js` change. And `BaseFormWithPolling` just react to the change in state.

pradeepdewda-okta: @worldleelavivat-okta The issue is we don't want to add View specific code AppState.. If it's not required then just remove it

thomasmaslen-okta: > @haishengwu-okta I don't believe I ever see `refresh` in `transformedResponse.currentAuthenticator?.poll?.refresh` but @thomasmaslen-okta if you don't mind confirming.

(Per Slack conversation) you're right that, due to autoenrollment of the Email Authenticator, you may not see `transformedResponse.currentAuthenticator?.poll?.refresh` in the wild at present -- but, per Haisheng's comment, this PR should definitely anticipate it.

worldleelavivat-okta: @haishengwu-okta / @thomasmaslen-okta  added `currentAuthenticator`

@pradeepdewda-okta 
Added comment + I think this is the cleanest way to get it since this is how remidiations are currently getting the new value as well and we are reusing `BaseFormWithPolling`. 
The problem is that there is no way for the view to get new value since it isn't being re-render + AppState aren't being refresh. Unless we pass on the whole new `transformedResponse` down to the variable instead of just a number to the view then let the view handle extracting data but I feel like that a little heavy handed approach.

thomasmaslen-okta (commented):
I don't understand the `polling.js` changes (from the "Fix bug when sending new refresh interval conflate polling for both remidiation polling and authenticator polling" commit), but I trust that @haishengwu-okta / @pradeepdewda-okta understand them and can scrutinize them.

src/v2/view-builder/views/shared/polling.js
```diff
@@ -6,16 +6,16 @@ export default {
     this.pollingInterval = newRefreshInterval || this.options.currentViewState.refresh;
     this.countDownCounterValue = Math.ceil(this.pollingInterval / MS_PER_SEC);
     // Poll is present in remediation form
-    if (this.pollingInterval) {
+    if (this.options.currentViewState.refresh) {
```
worldleelavivat-okta, line 9: Passing new `newRefreshInterval` for Authenticator would trigger `this._startRemediationPolling();` instead of `this._startAuthenticatorPolling();` if it is checking for `this.pollingInterval`. Since `this.options.currentViewState.refresh` is only exist for remediation check for that instead.

src/v2/view-builder/views/shared/polling.js
```diff
@@ -6,16 +6,16 @@ export default {
     this.pollingInterval = newRefreshInterval || this.options.currentViewState.refresh;
```
pradeepdewda-okta, line 6: lets create two variables here.. 
`this.fixedPollingInterval = this.options.currentViewState.refresh;`
`this.dynamicPollingInterval = newRefreshInterval`

src/v2/view-builder/views/shared/polling.js
pradeepdewda-okta, line 9: use `this. fixedPollingInterval` here

src/v2/view-builder/views/shared/polling.js
```diff
@@ -6,16 +6,16 @@ export default {
...
     this.countDownCounterValue = Math.ceil(this.pollingInterval / MS_PER_SEC);
     // Poll is present in remediation form
-    if (this.pollingInterval) {
+    if (this.options.currentViewState.refresh) {
       this._startRemediationPolling();
     } else {
       // Poll is present in authenticator/ authenticator Enrollment obj.
       // Authenticator won't co-exists hence it's safe to trigger both.
-      this._startAuthenticatorPolling();
+      this._startAuthenticatorPolling(newRefreshInterval);
```
pradeepdewda-okta, line 14: don't add a param

src/v2/view-builder/views/shared/polling.js
```diff
@@ -24,7 +24,7 @@ export default {
       if (this.options.appState.has(responseKey)) {
         const authenticator = this.options.appState.get(responseKey);
         const authenticatorPollAction = `${responseKey}-poll`;
-        const pollInterval = authenticator?.poll?.refresh;
+        const pollInterval = newRefreshInterval || authenticator?.poll?.refresh;
```
pradeepdewda-okta, line 27: use `this. dynamicPollingInterval` here

src/v2/models/AppState.js
```diff
@@ -141,8 +141,12 @@ export default Model.extend({
       // make sure to update the correct poll view's refresh value
       const currentFormName = this.get('currentFormName');
       const currentViewState = transformedResponse.remediations.filter(r => r.name === currentFormName)[0];
+      // Get new refresh interval for either: remediations, authenticator, or authenticator enrollment
+      const refreshInterval = currentViewState.refresh ||
```
pradeepdewda-okta, line 145: create a new function `getRefreshInterval` and move this logic there.. In case in future if we add more objects like `Authenticators`

src/v2/view-builder/views/shared/polling.js
worldleelavivat-okta: created

src/v2/view-builder/views/shared/polling.js
worldleelavivat-okta: fixed

src/v2/view-builder/views/shared/polling.js
worldleelavivat-okta: fixed

src/v2/models/AppState.js
worldleelavivat-okta: added

test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
```diff
@@ -274,3 +284,25 @@ test
     await t.expect(await terminalPageObject.goBackLinkExists()).ok();
     await t.expect(terminalPageObject.getFormTitle()).eql('Verify with your email');
   });
+
+test
+  .requestHooks(dynamicRefreshIntervalMock)('challenge email authenticator screen has right labels', async t => {
```
pradeepdewda-okta, line 289: change the test name

test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
```diff
@@ -274,3 +284,25 @@ test
...
+    await setup(t);
+
+    await t.wait(31000);
```
pradeepdewda-okta, line 292: this is taking total more than a minute to finish.. lets just validates dynamic interval.. with shorter wait time say `t.wait(5000)` and refreshInterval even shorter say `1000`

worldleelavivat-okta: fixed to total of 8 sec (2 sec with 1 sec/req interval + 6 sec with 2 sec/req interval)

worldleelavivat-okta:
@jane-kuang-okta / @santoshmale-okta / @sshen-okta 
Just FYI, Cred Platform team is also working on variable refresh polling rate. This PR is for the UI part of the change.

test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
```diff
@@ -274,3 +285,26 @@ test
...
+    await t.expect(logger.count(
+      record => record.response.statusCode === 200 &&
+        record.request.url.match(/poll/)
+    )).eql(2);
+
+    await t.removeRequestHooks(dynamicRefreshShortIntervalMock);
+    await t.addRequestHooks(dynamicRefreshLongIntervalMock);
+
+    // 3 poll requests in 6 seconds at 2 sec interval
+    await t.wait(6000);
```
pradeepdewda-okta, line 305: Thanks for making the test much efficient..

One thing. This actually doesn't test dynamicRefresh right? it's always 2sec ?

I might have to add sequence of mocks that have different `refreshIntervals`
```
authenticator-verification-email-polling-long-1 - 1000ms
authenticator-verification-email-polling-long-2 - 2000ms
authenticator-verification-email-polling-long-3 - 3000 ms
```

worldleelavivat-okta: It does test it right now, the sequence is as follow:
```
authenticator-verification-email-polling-short-1 - refreshInterval 1000ms - elapsedTime 1000ms - wait 2000ms
authenticator-verification-email-polling-short-2 - refreshInterval 1000ms - elapsedTime 2000ms
authenticator-verification-email-polling-long-1 - refreshInterval 2000 ms - elapsedTime 4000ms - wait 6000ms
authenticator-verification-email-polling-long-2 - refreshInterval 2000ms - elapsedTime 6000ms
authenticator-verification-email-polling-long-3 - refreshInterval 2000 ms - elapsedTime 8000ms

Total Requests: 5
```

Assertion is done at 2000ms to check that there are 2 requests completed so far. Final assertion at the end for 5 total requests at the end.

pradeepdewda-okta: ah okay make sense. thanks for clarification.

pradeepdewda-okta (approved):

PR #1807, opened 2021-04-02
fix: poll view in OIE

magizh-okta:
* Fixes poll view to display new refresh interval on based on response.

* Uses refresh as a time interval after which UI should check for next update, rather than a poll interval.

* We could change polling.js to use setTimout and change the behavior every where (bigger effort and not sure if that would be the ideal intended behavior). Current v1 implementation is using setInterval except for the pollView. So just keeping it the same way for now. 

Video before fix -  https://drive.google.com/file/d/1SdQWnsFVoqe2xNx3VzrzQSvv-eZPjW7d/view?usp=sharing

after fix - https://drive.google.com/file/d/1B2ij_Hc3PI3G4N5uMK42D4Jr-xbFcHTg/view?usp=sharing

https://user-images.githubusercontent.com/17267130/113633816-7de2c480-9622-11eb-9ee5-3e9890e45182.mov

RESOLVES: OKTA-375200

## PR Checklist

- [ ] Added unit tests?

- [x] Added e2e tests

### Issue:

- [OKTA-375200](https://oktainc.atlassian.net/browse/OKTA-375200)


src/v2/models/AppState.js
```diff
@@ -157,6 +157,16 @@ export default Model.extend({
       reRender = false;
     }
 
+    if (identicalResponse && this.get('currentFormName') === 'poll') {
```
pradeepdewda-okta, line 160: there is one more PR.. may have a conflict..

https://github.com/okta/okta-signin-widget/pull/1795/files#r605108218

src/v2/models/AppState.js
```diff
@@ -157,6 +157,16 @@ export default Model.extend({
...
+      /**
+       * returns true: We want to force reRender when currentForm is poll because request has to reinitiate
+       * based on new refresh and UI has to reflect new timer.
+       * We dont technicall poll here we just make a request after the specified refresh time each time
```
pradeepdewda-okta, line 164: typo: `technicall`

src/v2/view-builder/views/PollView.js
```diff
@@ -43,12 +46,13 @@ const Body = BaseForm.extend(Object.assign(
...
       BaseForm.prototype.remove.apply(this, arguments);
-      this.stopPolling();
+      clearTimeout(this.refreshTimeout);
     },
 
     triggerAfterError() {
       BaseForm.prototype.triggerAfterError.apply(this, arguments);
-      this.stopPolling();
+      clearTimeout(this.refreshTimeout);
+      this.stopCountDown();
```
haishengwu-okta, line 55: maybe move `stop/start CountDown` to this view and remove line 60 (compose `polling.js`)

haishengwu-okta (commented):
Looks good to me! thank you for taking the work!

src/v2/models/AppState.js
```diff
@@ -152,6 +152,16 @@ export default Model.extend({
       reRender = false;
     }
 
+    if (identicalResponse && this.get('currentFormName') === 'poll') {
```
haishengwu-okta, line 155: nit:
looks like we can clean up the logic a bit.
- `return false` at line 152
- don't need to check `identicalResponse`

test/testcafe/framework/page-objects/PollingPageObject.js
```diff
@@ -12,6 +13,14 @@ export default class PollingPageObject extends BasePageObject {
     return this.form.getTitle();
   }
 
+  getContent() {
```
haishengwu-okta, line 16: already exists in BaseFormObject or move into that?

test/testcafe/spec/PollView_spec.js
```diff
@@ -1,65 +1,60 @@
...
-  await identityPage.navigateToPage();  
-  return [identityPage , selectFactorPageObject];
-}  
+  await identityPage.navigateToPage();
+  return identityPage;
+}
 
-test.requestHooks(requestLogger, identifyMock)('should poll', async t => {
-  let [identityPage, selectFactorPage]  = await setup(t);
+test.only.requestHooks(requestLogger, identifyMock)('should make request based on timer in response', async t => {
```
haishengwu-okta: remove `.only`

test/testcafe/framework/page-objects/PollingPageObject.js
magizh-okta: The selector in basePageObject is for getIonMessages but here the text in poll view is added manually.

src/v2/view-builder/views/PollView.js
```diff
@@ -43,17 +47,34 @@ const Body = BaseForm.extend(Object.assign(
...
       this.$el.find('.o-form-fieldset-container').empty();
     },
-  },
 
-  polling,
+    startCountDown(selector , interval) {
```
pradeepdewda-okta, line 60: devs can't reuse it if they are not using PollView.. what's the reason to move it ?

src/v2/view-builder/views/PollView.js
```diff
@@ -43,17 +47,34 @@ const Body = BaseForm.extend(Object.assign(
...
+      if (this.countDown) {
+        clearInterval(this.countDown);
+      }
+      this.counterEl = this.$el.find(selector);
+      this.countDown = setInterval(() => {
+        if(this.counterEl.text() !== '0') {
```
pradeepdewda-okta, line 66: It was intentionally checking for `this.counterEl.text() !== '1'` .. i think it adds an extra second or shows `0` seconds remaining otherwise ??

test/testcafe/spec/PollView_spec.js
```diff
@@ -1,65 +1,60 @@
...
-//   let [identityPage, selectFactorPage]  = await setup(t);
-//   await identityPage.fillIdentifierField('username');
-//   await identityPage.fillPasswordField('password');
-//   await identityPage.clickNextButton();
-//   selectFactorPage.selectFactorByIndex(0);
-//   const terminalPageObject = new TerminalPageObject(t);
-//   await t.expect(terminalPageObject.getErrorMessages().getTextContent()).eql('Something went wrong, Try again later.');
-// });
\ No newline at end of file
+test.requestHooks(requestLogger, identifyPollErrorMock)('not poll on error', async t => {
```
pradeepdewda-okta, line 52: it's not flaky anymore ?

src/v2/view-builder/views/PollView.js
magizh-okta: The count down was mostly for this one view only. no other view shows the countdown. I moved it per @haishengwu-okta's suggestion coz it made sense to remove any confusion in the polling.js

src/v2/view-builder/views/PollView.js
magizh-okta: yeah this is intended and per v1. In case the actual request is taking longer the user should see 0 seconds and spinner and not stuck in 1 secopnd.

test/testcafe/spec/PollView_spec.js
magizh-okta: wil run couple of times in bacon just to be sure.

src/v2/models/AppState.js
magizh-okta: This wont work since it iwll always return false since identicalResponse is true.

pradeepdewda-okta (approved):
💯

src/v2/view-builder/views/PollView.js
haishengwu-okta: yeah, the count down is only meant for PollView. We can move to polling.js until it's going to be reused by other cases.

haishengwu-okta (approved):

Issue #1741, opened 2021-03-10
Issue mismatch error results in Promise returned by showSignInToGetTokens never resolving or erroring.


sflanker:
### I'm submitting a
  - [x] bug report
  - [ ] feature request

### Background info

I apparently misconfigured something such that I am getting a token with an incorrect issuer (I'm using a custom domain, but the issuer URL is my ___.okta.com url instead). This misconfiguration results in completely silent failure and the Promise that is returned by the `showSignInToGetTokens` function never resolves or errors.

### Expected behavior

The handler specified using `Promise.catch` should be invoked with the error. 

### What went wrong?

The login dialog just hangs forever. There are no errors or warnings in the console.

### Steps to reproduce

Follow the basic [showSignInToGetTokens example](https://www.npmjs.com/package/@okta/okta-signin-widget#showsignintogettokens). Specify the custom domain for your Okta SPA application as the `baseUrl`.

https://repl.it/@paulinvio/okta-aurelia-webpack-3-app#src/app.ts

### Your environment

  - Okta Sign-In Widget Version:
  - Browser: Edge 88.0.705.68 (Chromium 88.0.4324.150)
  - OS: Mac OS 10.15.7
  - Language: English

sflanker:
It seems like there should be a call to `settings.callGlobalError` in here somewhere 😞:

https://github.com/okta/okta-signin-widget/blob/master/src/util/OAuth2Util.js#L46

shuowu:
@sflanker Thanks for the report!

Internal Ref: OKTA-377480

sflanker:
I'm curious if there's a rationale for this test: https://github.com/okta/okta-signin-widget/blob/master/test/unit/spec/PrimaryAuth_spec.js#L2986 or if it is just validating bad behavior.

PR #1815, opened 2021-04-06
Added call to callGlobalError when an error occurs during sign in. #1741

sflanker:
## Description:

Fixed https://github.com/okta/okta-signin-widget/issues/1741
Previously if an issuer was incorrect the promise returned by showSignInToGetTokens would never resolve or fail. Now it fails.

## PR Checklist

- [x] Added unit tests? **(not added, but fixed existing test)**
- [ ] Added e2e tests
- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)? Maybe you should link to a public wiki instead of this login-walled thing

### Issue:

- [OKTA-1741](https://github.com/okta/okta-signin-widget/issues/1741)

### DIFF
--- a/src/util/OAuth2Util.js
+++ b/src/util/OAuth2Util.js
@@ -65,6 +65,7 @@ util.getTokens = function (settings, params, controller) {
     }
 
     Util.triggerAfterError(controller, new Errors.OAuthError(error.message), settings);
+    settings.callGlobalError(error);
   }
 
   const authClient = settings.getAuthClient();
--- a/test/unit/spec/PrimaryAuth_spec.js
+++ b/test/unit/spec/PrimaryAuth_spec.js
@@ -2983,7 +2983,7 @@ Expect.describe('PrimaryAuth', function () {
     });
 
     // TODO: add test to verify the behavior when missing `state` in the data
-    itp('ignores messages with the wrong origin', function () {
+    itp('Rejects messages with the wrong origin', function () {
       const successSpy = jasmine.createSpy('successSpy');
       const errorSpy = jasmine.createSpy('errorSpy');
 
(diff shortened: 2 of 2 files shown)


oleksandrpravosudko-okta:
Thanks for the contribution - this change looks reasonable.
I have copied this over to https://github.com/okta/okta-signin-widget/pull/1818 to run tests.

haishengwu-okta:
Curious why cannot rely on `afterError` event?

src/util/OAuth2Util.js
```diff
@@ -65,6 +65,7 @@ util.getTokens = function (settings, params, controller) {
     }
 
     Util.triggerAfterError(controller, new Errors.OAuthError(error.message), settings);
+    settings.callGlobalError(error);
```
pradeepdewda-okta, line 68: triggerAfterError doesn't work because we are passing `error.message` instead of error ?

haishengwu-okta: - seems @sflanker concern didn't get error instead of actual content.?

sflanker: > Curious why cannot rely on `afterError` event?

Because that's not how Promises are supposed to work. If your API returns a promise expect that promise to eventually either resolve or reject. I shouldn't have to wire up some special event handler instead! Having some special event handler *in addition to* rejecting the promise would be fine if people need a special mechanism to subscribe to certain kinds of failure events.

haishengwu-okta: the default behavior of `settings.callGlobalError(error);`  is to throw error but user can customized it to do something different. 
shall we just throw errors?

sflanker: You're in a promise worker function so throwing exceptions would be sub optimal since that would also be un-handleable and leave the promise in a perpetually pending state. However that's not the behavior of `callGlobalError` in this context. It's behavior is to reject the promise, which it what it should be doing in this case.

I'm not an expert in your codebase, so it is possible that this same code path is invoked in other contexts where this behavior might not be ideal. This PR is just meant to be a suggestion.

Here's a simple repro of the issue: https://replit.com/@paulinvio/static-site-okta-signin#script.js
You get test the issue here: https://static-site-okta-signin.paulinvio.repl.co/
If you log in with valid credentials (I can provide you with some if you email me: paul@invioinc.com ), The result is no error or warning in the console, and the promise returned is forever stuck in a pending state (the catch handler is never called).

Issue #1512, opened 2020-11-18
idpDiscovery with requestContext does not provide access & Id token


ey52fpv:
:information_source: **If you have a question**, please post it on the [Okta Developer Forum](https://devforum.okta.com) instead. Issues in this repository are reserved for bug reports and feature requests.

### I'm submitting a
  - [ ] bug report
  - [ ] feature request

### Background info

I'm working on React App with OKTA Sign in Widget that has 2 users flows internal organization users vs external. For external users, the password are stored in the APP's OKTA org and I'm able to retrieve access & id token on successful authentication. For internal users, the APP's OKTA org make a SAML request to organization's OKTA org for authentication and on successful authentication when it redirects to APP it does not come with access & id token.

const widget = new OktaSignIn({
        baseUrl: XXX,
        clientId:XXX,
        redirectUri: window.location.origin + '/implicit/callback',
        i18n: {
          en: {
            'primaryauth.title': 'XXX',
          },
        },
        authParams: {
          pkce:true,
          issuer:XXX,
          display: 'page',
          scopes: ['openid', 'profile', 'email'],
          responseMode: 'query',
          responseType: ['id_token', 'token'],
        },
        tokenManager: { secure: false }, 
        cookies: {
          secure: false
        },
        features: {
           registration: false, 
           idpDiscovery: true 
        },
        idpDiscovery:{
          requestContext: '/home/oidc_client/1234/5678'
        }
      });

Code Snippet:

const { authState, authService } = useOktaAuth();
useEffect(() => {
    if (!authState.isPending) {
    if (!authState.isAuthenticated) {
      // When user isn't authenticated, forget any user info
      authService.login("/");
    } else {
      authService.getUser().then((info) => {
        getUsers(info.sub);
      });
    }
  }
  },[authState,authService]); 

### Expected behavior

What should have happened? 

### What went wrong?

Please provide log or error messages if applicable.

### Steps to reproduce

If the current behavior is a bug, please provide the steps to reproduce and a minimal demo if possible (you can use this template as a starting point: https://jsfiddle.net/m2gyLq4v/).

### Your environment

  - Okta Sign-In Widget Version:
  - Browser:
  - OS:
  - Language:

swiftone:
Hi @ey52fpv - Are you using the widget hosted on an okta site, or on your own site?  Unfortunately, for the latter case, idpDiscovery is not yet an officially supported feature.

ey52fpv:
[...]

PR #1817, opened 2021-04-06
Use token.getWithRedirect after positive IDP discovery. #1512

sflanker:
## Description:

Previously using the Sign-in Widget with IDP discovery caused the user to get stuck on the Okta dashboard for the Service Provider okta account, unless the poorly documented `idpDiscovery.requestContext` property, and even then the user gets redirected back to the application with no code and no tokens (just `fromLogin=true`). This PR fixes this by simply redirecting to the OAuth2 authorize endpoint. 

I'm not sure you'll want to take this PR as is since I've simply dropped the tests related to `features.redirectByFormSubmit` because I can't find any documentation on it so I don't know what the expected behavior for this would be.

### Issue:

- [Github Issue 1512](https://github.com/okta/okta-signin-widget/issues/1512)

### DIFF
--- a/src/models/IDPDiscovery.js
+++ b/src/models/IDPDiscovery.js
@@ -13,7 +13,7 @@
 import PrimaryAuthModel from 'models/PrimaryAuth';
 import CookieUtil from 'util/CookieUtil';
 import Enums from 'util/Enums';
-import Util from 'util/Util';
+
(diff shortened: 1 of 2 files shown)


oleksandrpravosudko-okta:
Thanks for the PR @sflanker.

Internal ticket for review and investigation:
OKTA-385002

PR #1819, opened 2021-04-06
AppState render check should ignore 'stateHandle' data OKTA-384206

anatoliiboiarshchykov-okta:
- update shouldReRenderView() to ignore 'stateHandle' to match previous & current data
- add a unit test to check 'stateHandle' field is ignored

## Description:
Due to ticket [OKTA-384206](https://oktainc.atlassian.net/browse/OKTA-384206)([PR](https://github.com/okta/okta-core/pull/50800)) **'**stateHanler**'** field is now containing encrypted context. As result, its value is changed with each **'/poll'** or any other API call, causing **FormController** page to re-render(it is triggered once any changes between previous calls data and current detected). When **FormController** page is re-rendering it is recalling OV **'/deviceChallenge'**, causing pop-up spamming.

In the previous implementation, 'stateHandle' was only an ID for Redis and it was not immutable.
## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/80226149/113718672-8c4de080-96f5-11eb-910c-1135f7915067.mov

### Reviewers:
@santoshmale-okta , @sshen-okta , @vakaevich-okta  ,@haishengwu-okta , @pradeepdewda-okta 
@jane-kuang-okta - can you please take a look, since you worked with this code some time ago.

### Issue:

- [OKTA-384206](https://oktainc.atlassian.net/browse/OKTA-384206)

### DIFF
--- a/src/v2/mixins/mixins.js
+++ b/src/v2/mixins/mixins.js
@@ -1,17 +1,24 @@
 import { _ } from 'okta';
 
 _.mixin({
-  nestedOmit: function(obj, iteratee, context) {
-    var result = _.omit(obj, iteratee, context);
-          
+  nestedOmit: function(obj, names) {
+    var result = _.omit(obj, names);
+
+    if (names.includes(result.name)) {
+      result = _.omit(result, 'value');
+    }
+
     _.each(result, function(val, key) {
-      if (typeof(val) === 'object') {
-        result[key] = _.nestedOmit(val, iteratee, context);
+      if (Array.isArray(val)) {
+        result[key] = val.map((v) => {
+          return _.nestedOmit(v, names);
+        });
+      } else if (typeof val === 'object') {
+        result[key] = _.nestedOmit(val, names);
       }
     });
-          
     return result;
-  }
+  },
 });
 
 export { _ };
--- a/src/v2/models/AppState.js
+++ b/src/v2/models/AppState.js
@@ -131,11 +131,13 @@ export default Model.extend({
 
   shouldReRenderView(transformedResponse) {
     const previousRawState = this.has('idx') ? this.get('idx').rawIdxState : null;
+    
     const identicalResponse = _.isEqual(
-      _.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'refresh']),
-      _.nestedOmit(previousRawState, ['expiresAt', 'refresh']));
-    const isSameRefreshInterval = _.isEqual(_.omit(transformedResponse.idx.rawIdxState, 'expiresAt'),
-      _.omit(previousRawState, 'expiresAt') );
+      _.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'refresh', 'stateHandle']),
+      _.nestedOmit(previousRawState, ['expiresAt', 'refresh', 'stateHandle']));
+    const isSameRefreshInterval = _.isEqual(
+      _.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'stateHandle']),
+      _.nestedOmit(previousRawState, ['expiresAt', 'stateHandle']));
 
     if (identicalResponse && !isSameRefreshInterval) {
       this.set('dynamicRefreshInterval', this.getRefreshInterval(transformedResponse));
--- a/test/unit/spec/v2/models/AppState_spec.js
+++ b/test/unit/spec/v2/models/AppState_spec.js
@@ -1,5 +1,5 @@
 import AppState from 'v2/models/AppState';
-import { FORMS_WITHOUT_SIGNOUT, FORMS_FOR_VERIFICATION, FORMS } from 'v2/ion/RemediationConstants';
+import {FORMS, FORMS_FOR_VERIFICATION, FORMS_WITHOUT_SIGNOUT} from 'v2/ion/RemediationConstants';
 
 describe('v2/models/AppState', function() {
   beforeEach(() => {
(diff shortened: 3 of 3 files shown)


src/v2/models/AppState.js
```diff
@@ -129,13 +129,31 @@ export default Model.extend({
...
   shouldReRenderView(transformedResponse) {
     const previousRawState = this.has('idx') ? this.get('idx').rawIdxState : null;
+    
     const identicalResponse = _.isEqual(
-      _.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'refresh']),
-      _.nestedOmit(previousRawState, ['expiresAt', 'refresh']));
-    const isSameRefreshInterval = _.isEqual(_.omit(transformedResponse.idx.rawIdxState, 'expiresAt'),
-      _.omit(previousRawState, 'expiresAt') );
+      _.nestedOmitStateHandleObjects(
+        _.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'refresh', 'stateHandle'])),
```
haishengwu-okta, line 151: why adding `stateHandle` to `nestedOmit` isn't enough?

src/v2/models/AppState.js
```diff
@@ -129,13 +129,31 @@ export default Model.extend({
     return FORMS_FOR_VERIFICATION.includes(currentFormName);
   },
 
+  removeStateHandleObjects(obj, parentObj, objKeyInParent) {
```
haishengwu-okta, line 132: Has this been moved to underscore wrapper already?

src/v2/models/AppState.js
jane-kuang-okta: +1

pradeepdewda-okta: is this because you want to remove it from under value object ?
```
            'value': {
              'name': 'stateHandle',
              'value': 'ABCDF'
            }
```

anatoliiboiarshchykov-okta: Correct. '**poll**' response contains '**value**' objects with '**stateHandle**' field witch changes.
**Here is how it looks:**
Screenshot 2021-04-07 at 08 32 45

**Full response:**
[poll_resp.json.zip](https://github.com/okta/okta-signin-widget/files/6269203/poll_resp.json.zip)

src/v2/models/AppState.js
anatoliiboiarshchykov-okta: I am not sure. Checked [Okta.js](https://github.com/okta/okta-ui/blob/master/packages/courage/src/util/Okta.js), but was not able to find anything similar. 
Please advise, is there any place where I can take a look?

src/v2/models/AppState.js
pradeepdewda-okta: Okay.. so that should also be part of _.nestedOmit() function.. probably add a new parameter to it and call `nestedOmitStateHandleObjects` internally based on that parameter.. 

basically we should only use `nestedOmit` in the code to do every deletion we need..

haishengwu-okta: oh I see. yeah, would be nice to modify `nestedOmit`

src/v2/mixins/mixins.js
```diff
@@ -1,17 +1,44 @@
 import { _ } from 'okta';
 
 _.mixin({
-  nestedOmit: function(obj, iteratee, context) {
-    var result = _.omit(obj, iteratee, context);
+  nestedOmit: function(obj, iteratee, context, dropObjectsByNameField) {
```
haishengwu-okta, line 4: I'm thinking something simpler (without extra `dropObjectsByNameField`)

- expect 2nd parameter is array. we don't need to be as generic as `_.omit` atm.
- adds `obj.name === "stateHandle"` is because drop by name is special case for `stateHandle`. But it's optional as it doesn't affect final result atm.

```
    nestedOmit: function(obj, names) {
    var result = _.omit(obj, names);

    if (names.includes(result.name) && result.name === "stateHandle") {
      result = _.omit(result, 'value');
    }

    _.each(result, function(val, key) {
      if (Array.isArray(val)) {
        result[key] = val.map((v) => {
          return _.nestedOmit(v, names);
        });
      } else if (typeof val === "object") {
        result[key] = _.nestedOmit(val, names);
      }
    });

    return result;
  },
```

test/unit/spec/v2/models/AppState_spec.js
```diff
@@ -111,4 +111,42 @@ describe('v2/models/AppState', function() {
...
+    it('rerender view should be false if stateHandle are different', () => {
+
+      const transformedResponse = {
+        'idx': {
+          'rawIdxState': {
+            'version': '1.0.0',
+            'intent': 'LOGIN',
+            'stateHandle': 'ABCDF',
+            'refresh' : '23',
+            'value': {
```
haishengwu-okta, line 125: better to mimic real data (the value is nested inside another array)

src/v2/models/AppState.js
```diff
@@ -131,11 +131,13 @@ export default Model.extend({
...
+    
     const identicalResponse = _.isEqual(
-      _.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'refresh']),
-      _.nestedOmit(previousRawState, ['expiresAt', 'refresh']));
-    const isSameRefreshInterval = _.isEqual(_.omit(transformedResponse.idx.rawIdxState, 'expiresAt'),
-      _.omit(previousRawState, 'expiresAt') );
+      _.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'refresh', 'stateHandle'], {}, true),
+      _.nestedOmit(previousRawState, ['expiresAt', 'refresh', 'stateHandle'], {}, true));
+    const isSameRefreshInterval = _.isEqual(
+      _.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'refresh', 'stateHandle'], {}, true),
```
pradeepdewda-okta, line 139: Why to omit `'refresh'` for `isSameRefreshInterval` ??

src/v2/mixins/mixins.js
pradeepdewda-okta: we can simplify this as @haishengwu-okta mentioned above ^^

src/v2/models/AppState.js
jane-kuang-okta: This should be 
```suggestion
      const isSameRefreshInterval = _.isEqual(
      _.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'stateHandle'], {}, true),
      _.nestedOmit(previousRawState, ['expiresAt', 'stateHandle'], {}, true));
```
@anatoliiboiarshchykov-okta please remove `refresh`. `isSameRefreshInterval` basically says if we remove the keys listed (that don't include refresh) and the two IDX responses are still the same, then the refresh interval is the same.

anatoliiboiarshchykov-okta: typo

test/unit/spec/v2/models/AppState_spec.js
anatoliiboiarshchykov-okta: Updated with real payload. (It is trimmed a little bit
)

src/v2/mixins/mixins.js
anatoliiboiarshchykov-okta: Used this approach but without  result.name === "stateHandle" check to omit value field
here ->
if (names.includes(result.name) && result.name === "stateHandle") {
      result = _.omit(result, 'value');
    }

haishengwu-okta (approved):

pradeepdewda-okta:
Nice..

PR #1822, opened 2021-04-06
fix: add webauthn oie i18n in login.properties, update mocks

TianGan-okta:
## Description:
Add the keys/values as https://github.com/okta/okta-core/pull/51526

Discussion:
1. we will use the prefix to map the api error keys
2. If we want to support the customization of the webauthn, we can do it later either fix the loc or fallback if error
3. The error `authfactor.webauthn.error.assertion_validation_failure` will show after browser error, so we barely show the api error, so it is fine to map to generic error

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
image (51)

### Reviewers:
@nikhilvenkatraman-okta @mpeng-okta 

### Issue:
- [OKTA-383350](https://oktainc.atlassian.net/browse/OKTA-383350)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -448,6 +448,16 @@ verify.webauthn.biometric.instructions=Your browser or device will prompt you to
 webauthn.biometric.error.factorNotSupported = Security key or biometric authenticator is not supported on this browser. Select another factor or contact your admin for assistance.
 webauthn.biometric.error.factorNotSupported.oneFactor = Security key or biometric authenticator is not supported on this browser. Contact your admin for assistance.
 enroll.webauthn.instructions.noSupportForBiometric = Note: Some browsers may not support biometric authenticators.
+authfactor.webauthn.error.clientdata_parsing_failed=Client data parsing failed
+authfactor.webauthn.error.clientdata_message_type_mismatch=Message type in client data doesn't match expected message type.
+authfactor.webauthn.error.clientdata_challenge_missing=No challenge parameter in client data
+authfactor.webauthn.error.clientdata_challenge_mismatch=Challenge in client data doesn't match with server side value
+authfactor.webauthn.error.clientdata_origin_missing=No origin parameter in client data
+authfactor.webauthn.error.clientdata_origin_mismatch=Origin parameter in client data doesnt match. clientData origin was {0}
```
mpeng-okta, line 456: `authfactor.webauthn.error.clientdata_origin_mismatch=Origin parameter in client data doesn't match. clientData origin was {0}`

Small update -- I forgot an apostrophe in my PR so this should be the above.

test/testcafe/spec/GenericErrorHandling_spec.js
```diff
@@ -2,11 +2,29 @@ import { RequestMock } from 'testcafe';
...
+import signatureVerificationFailure from '../../../playground/mocks/data/idp/idx/error-authenticator-webauthn-signature-verification-failure';
+import ChallengePasswordPageObject from '../framework/page-objects/ChallengePasswordPageObject';
 
 const securityAccessDeniedMock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/introspect')
   .respond(xhr403SecurityAccessDenied, 403);
 
+const challengeFailedResponse = (response) => RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(verificationPassword)
```
TianGan-okta, line 24: Make the verification page to password, since the webauthn will always failed in e2e as the following:
image (50)
I think it should be fine since we just need to verify the localization here. Please let me know, if it's ok or I need to revert the tests

nikhilvenkatraman-okta: I see that we have unit tests for [ChallengeWebAuthnView](https://github.com/okta/okta-signin-widget/blob/master/test/unit/spec/v2/view-builder/views/ChallengeWebauthnView_spec.js) should we move this e2e test to unit test ? cc: @magizh-okta

mpeng-okta: I think this particular error is a WebAuthn-library-specific error for timeout or when the user clicks cancel (it's part of the set of errors that the browser itself surfaces), and this isn't localized since we're not necessarily sure what error would be returned (it's also opaque on purpose here since the browsers aren't supposed to show time out vs not allowed). The new error messages we show would be localized though.

mpeng-okta: I think this particular error is a WebAuthn-library-specific error for timeout or when the user clicks cancel (it's part of the set of errors that the browser itself surfaces), and this isn't localized since we're not necessarily sure what error would be returned (it's also opaque on purpose here since the browsers aren't supposed to show time out vs not allowed). The new error messages we show would be localized though.

TianGan-okta: If use the unit test, we only test the keys, but not the mocks, if that is fine?

TianGan-okta: I think the best way now is testing in the `i18nTransformer` test with its response mock

mpeng-okta (approved):

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -949,3 +959,6 @@ oie.consent.enduser.deny.label = No, it's not me
 oie.consent.enduser.title = Did you just try to sign in?
 oie.consent.enduser.email.allow.description = To continue, return to the device or window where you requested the email link. You may close this window at any time.
 oie.consent.enduser.deny.description = You may close this window at any time.
+
+# Browser thrown errors
+browser.error.NotAllowedError = The operation either timed out or was not allowed. See: https://www.w3.org/TR/webauthn-2/#sctn-privacy-considerations-client.
```
magizh-okta, line 964: If we are doing overriding we can give it a user friendly error message and not include links. 

We can just leave it as "The operation either timed out or was not allowed."

src/v2/ion/i18nTransformer.js
```diff
@@ -281,6 +281,21 @@ const getMessage = (message) => {
   return message.message;
 };
 
+/**
+ * @param {Object} error
+ */
+const getMessageFromBrowserError = (error) => {
+  if (error.name) {
+    const key = `browser.error.${error.name}`;
+    if (Bundles.login[key]) {
```
nikhilvenkatraman-okta, line 290: 👍

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -448,6 +448,16 @@ verify.webauthn.biometric.instructions=Your browser or device will prompt you to
...
+authfactor.webauthn.error.clientdata_parsing_failed=Client data parsing failed
+authfactor.webauthn.error.clientdata_message_type_mismatch=Message type in client data doesn't match expected message type.
+authfactor.webauthn.error.clientdata_challenge_missing=No challenge parameter in client data
+authfactor.webauthn.error.clientdata_challenge_mismatch=Challenge in client data doesn't match with server side value
+authfactor.webauthn.error.clientdata_origin_missing=No origin parameter in client data
+authfactor.webauthn.error.clientdata_origin_mismatch=Origin parameter in client data doesn't match. clientData origin was {0}
+authfactor.webauthn.error.invalid_enrollment_request_data=Invalid data in the WebAuthn enrollment request. {0}
+authfactor.webauthn.error.assertion_validation_failure=None of the WebAuthn authenticators the user has enrolled for was able to validate the assertion
+authfactor.webauthn.error.authdata_validation_failure_during_authentication_challenge=Failed to validate authenticator data during WebAuthn verification.
+authfactor.webauthn.error.signature_verification_failure=Signature verification was unsuccessful.
```
magizh-okta, line 460: @mpeng-okta I know these are api error messages. Now seems like a good time to make sure they are user friendly too.  @vincentvoong-okta  thoughts ?

playground/mocks/data/idp/idx/error-authenticator-webauthn-assertion-validation-failure.json
```diff
@@ -0,0 +1,299 @@
+{
```
magizh-okta, line 1: @TianGan-okta  wondering if we can just clone and add respective messages in test rather than having n json files

playground/mocks/data/idp/idx/error-authenticator-webauthn-invalid-enrollment-request-data.json
```diff
@@ -0,0 +1,300 @@
...
+    ]
+  },
+  "messages":{
+    "type":"array",
+    "value":[
+      {
+        "message":"Invalid data in the WebAuthn enrollment request. {0}",
+        "i18n":{
+          "key":"authfactor.webauthn.error.invalid_enrollment_request_data",
+          "params": ["my data"]
```
nikhilvenkatraman-okta, line 158: Is this intentionally left for testing ?

magizh-okta (commented):
🔥  Nice work @TianGan-okta

playground/mocks/data/idp/idx/error-authenticator-webauthn-invalid-enrollment-request-data.json
nikhilvenkatraman-okta: Can we do this transformation within the test file ?

playground/mocks/data/idp/idx/error-authenticator-webauthn-assertion-validation-failure.json
TianGan-okta: This is a mock for playground, doesn't hurt to have these mock for dev to play with :) what do you think?

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -448,6 +448,16 @@ verify.webauthn.biometric.instructions=Your browser or device will prompt you to
 webauthn.biometric.error.factorNotSupported = Security key or biometric authenticator is not supported on this browser. Select another factor or contact your admin for assistance.
 webauthn.biometric.error.factorNotSupported.oneFactor = Security key or biometric authenticator is not supported on this browser. Contact your admin for assistance.
 enroll.webauthn.instructions.noSupportForBiometric = Note: Some browsers may not support biometric authenticators.
+authfactor.webauthn.error.clientdata_parsing_failed=Client data parsing failed
```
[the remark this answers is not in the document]
mpeng-okta: I guess it's this way mainly because we're following the patten from the v1 messages bundle, where we'll want to be reusing for v1 as well https://github.com/okta/okta-core/pull/51526/files#diff-0c5a124f8509ab561a3675183396cc89b5c5688b20c817c95a6728ef50e6b9a9R360

playground/mocks/data/idp/idx/error-authenticator-webauthn-assertion-validation-failure.json
magizh-okta: I am just worried that there are too many json files and we need to update everything if we need to change anything. Since these are all the same api format I dont think dev would test this often. Dont have a strong opinion here. Just wanted to keep the code base clean.

packages/@okta/i18n/src/properties/login.properties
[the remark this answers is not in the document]
pradeepdewda-okta: @TianGan-okta Is this the error visible to end user? If yes, then it doesn't make sense, we should avoid blindly copying API errors and display to enduser..

Recently many customers are complaining about such errors.

packages/@okta/i18n/src/properties/login.properties
pradeepdewda-okta: Yes.. We should override these messages here, which make sense to Enduser.

pradeepdewda-okta (changes requested):
Update enduser facing error messages to something that make sense to them.

packages/@okta/i18n/src/properties/login.properties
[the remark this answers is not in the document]
TianGan-okta: FYI, I asked @vincentvoong-okta and @mattloughran-okta's review on this thread: https://okta.slack.com/archives/C8WSD3TM4/p1617835519328000

test/unit/spec/v2/ion/i18nTransformer_spec.js
```diff
@@ -1752,4 +1753,128 @@ describe('v2/ion/i18nTransformer', function() {
...
+          params: ['User present bit was not set'],
+        },
+      },
+      expectedWebAuthnGenericError,
+    ], [
+      'authfactor.webauthn.not.api.error',
+      {
+        message:'This is not a api error that can be generic.',
+        i18n:{
+          key:'authfactor.webauthn.not.api.error',
```
magizh-okta, line 1871: still an api error. 

`authfactor.webauthn.not.known.api.error`  ?

magizh-okta (commented):
LGTM 👍  @nikhilvenkatraman-okta  @jmelberg-okta  looking for prefix to map multiple error keys to one seems like good approach to avoid hardcoding of keys in UI . Do you have any objections to these ?

mpeng-okta (approved):

packages/@okta/i18n/src/properties/login.properties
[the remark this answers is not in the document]
haishengwu-okta: Even though, why need `authfactor` prefix. `webauthn` shall be good enough?

playground/mocks/data/idp/idx/error-authenticator-webauthn-failure.json
```diff
@@ -0,0 +1,299 @@
...
+      }
+    ]
+  },
+  "messages":{
+    "type":"array",
+    "value":[
+      {
+        "message":"None of the WebAuthn authenticators the user has enrolled for was able to validate the assertion",
+        "i18n":{
+          "key":"authfactor.webauthn.error.assertion_validation_failure"
```
jmelberg-okta, line 157: I'm not a fan of this approach, as we're returning keys that will always be unused by clients.

What is the issue with returning the `authfactor.webauthn.error` key here? Developers can still read from the `message` value to understand what the root error is.

packages/@okta/i18n/src/properties/login.properties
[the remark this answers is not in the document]
TianGan-okta: i think @mpeng-okta is following the existing pattern in the okta-core which we are using in V1. If we do have strong opinion here, I can map it to a new key for UI only. The reason why I am using prefix instead of a need key is if will easy for the develop know it is mapped from API, and easy to debug, for example if we get a error key from backend, we just use it's prefix to find the mapped key in login bundle instead of look in to the code and figure out which key is the mapped one.

playground/mocks/data/idp/idx/error-authenticator-webauthn-failure.json
TianGan-okta: Yes, it helps the developers to look in to the response if there is any error occur

magizh-okta: if backend is adding this key for this sole purpose of client to use its own translation  then it makes sense for backend to return the generic key. I am not sure if there is some other purpose for backend. 

Also seems like it is a clients decision to determine what message to show so it feels weird from backend to have a common error code. /cc @mpeng-okta

TianGan-okta: And if we change the error object on the back-end now, it will cause to many effort, since webauthn probably the only one have this issue.

packages/@okta/i18n/src/properties/login.properties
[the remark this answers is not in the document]
mpeng-okta: Yeah the current pattern has the full `authfactor.webauthn.error` prefix, and we're sharing keys across OIE and v1, so it would make sense to keep the same keys (not repeating ourselves, not re-translating keys, etc).

playground/mocks/data/idp/idx/error-authenticator-webauthn-failure.json
mpeng-okta: Yeah, the full key is returned basically as a pipeline-level mechanism. When the IDX pipeline internationalizes keys, it does a translation/conversion for the message value itself, and returns the corresponding i18n key as well for the UI to grab and do its own translation for display in the SIW (the message value isn't necessarily displayed in the SIW AFAIK).

Would probably require some strange rewiring to have the pipeline do i18n on the full key (e.g. the assertion validation failure key above), but then only return the genericized i18n key, since the platform "layer" is using the full key to obtain the specific message above.

magizh-okta (approved):
@jmelberg-okta  seems like backend cannot change the key due to valid reasons. I am okay with this approach. Let us know if you still have concerns/ alternate suggestions.

src/v2/ion/i18nTransformer.js
```diff
@@ -46,6 +46,8 @@ import Logger from 'util/Logger';
 import { getAuthenticatorDisplayName } from '../view-builder/utils/AuthenticatorUtil';
 import { FORMS, AUTHENTICATOR_KEY } from './RemediationConstants';
 
+const WEBAUTHN_API_GENERIC_ERROR_KEY = 'authfactor.webauthn.error';
```
pradeepdewda-okta, line 49: what is the confidence level that this won't change ?

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -950,3 +951,6 @@ oie.consent.enduser.deny.label = No, it's not me
 oie.consent.enduser.title = Did you just try to sign in?
 oie.consent.enduser.email.allow.description = To continue, return to the device or window where you requested the email link. You may close this window at any time.
 oie.consent.enduser.deny.description = You may close this window at any time.
+
+# Browser thrown errors
+browser.error.NotAllowedError = The operation either timed out or was not allowed.
```
pradeepdewda-okta, line 956: append `oie.` to key

pradeepdewda-okta (approved):
looks better.. Now we only show this two error messages to enduser. perfect.

packages/@okta/i18n/src/properties/login.properties
nikhilvenkatraman-okta: Where is this key being used ?

src/v2/ion/i18nTransformer.js
mpeng-okta: It would require a breaking change, and we have a comment to include the prefix for any new WebAuthn error strings, so the likelihood of it changing is probably quite low.

packages/@okta/i18n/src/properties/login.properties
TianGan-okta: That is for the browser error that throw from the chrome/firefox or others

test/unit/spec/v2/ion/i18nTransformer_spec.js
```diff
@@ -1752,4 +1753,128 @@ describe('v2/ion/i18nTransformer', function() {
...
+    })).toEqual('browser message');
+  });
+
+  const expectedWebAuthnGenericError = 'You are currently unable to use a Security key or biometric authenticator. Try again.';
+
+  it.each([
+    [
+      'authfactor.webauthn.error.assertion_validation_failure',
+      {
+        message:'None of the WebAuthn authenticators the user has enrolled for was able to validate the assertion',
```
magizh-okta, line 1777: qq @mpeng-okta @vincentvoong-okta  why dont we want to surface this error ? In case user had many yubikeys and was using the one he/she dint enroll and was wondering what happened wouldnt this be helpful ?

mpeng-okta: Since we use `allowCredentials` anyway, if the user uses an unenrolled one we'd get a client error anyway I think:
https://github.com/okta/okta-signin-widget/blob/5689b077fd8f5b0ea14841cc94fa7e5dcc044187/src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js#L66-L75

magizh-okta (approved):

Issue #1071, opened 2020-03-03
"Sign Out" link doesn't invalidate user sessions


wgknowles:
### I'm submitting a
  - [X] bug report
  - [ ] feature request

### Background info

There is a "Sign Out" link that appears in several places on the Okta sign-on Widget (e.g. on the MFA prompt screen). Clicking this link doesn't actually invalidate an active Okta user session.

### Expected behavior

Clicking a "Sign Out" link should _always_ invalidate active sessions.

### What went wrong?

Clicking the "Sign Out" link takes you back to the first screen asking for a username and password, but if you navigate to a non-MFA URL e.g. [tenant].okta.com/enduser/UserHome it doesn't actually invalidate the user's okta session.

### Steps to reproduce

Note: These steps assume that your Sign-on Policy _does not_ require MFA for all URLs.

1) Log into an okta tenant and establish a session (e.g. https://[tenant].okta.com/app/UserHome
2) Now, navigate to a URL that triggers a MFA prompt (e.g. an MFA protected App URL)
3) Click the Sign-out link on the MFA prompt page
4) Now, browse to a URL that doesn't require MFA (e.g. https://[tenant].okta.com/app/UserHome)
5) You are still logged in!

### Your environment

  - Okta Sign-In Widget Version: 3.3

swiftone:
Internal ref: OKTA-282546

wgknowles:
This issue only appears after you change versions of the sign-on widget.  I've only tested on 3.3 but it may apply to other versions.

aarongranick-okta:
Hi @wgknowles and thank you for bringing this issue to our attention.

The link on some views of the widget says “Sign Out” but it really means “Cancel / Back to Beginning”. It definitely does NOT sign you out. The issue here seems to be a bad choice of wording, on our part, for the "cancel" button.

The assumption was that if the widget is shown you are not signed in, but in this case it looks like the default system policy does not require MFA, but the app policy does. So logging into the app requires a "step up" with additional authentication factor(s).

I think it is possible this case was not properly considered when "Sign Out" was chosen as the text for this button. I am adding notes to this issue `OKTA-282546`, hopefully we can find a better, less misleading label.

PR #1825, opened 2021-04-07
Close session on 'Sign out'

denysoblohin-okta:
## Description:
- In MFA pages for v1 link 'Sign Out' changed to 'Back to Sign In'  ('Sign out' should now be only on password reset and expired pages)
- By default `FooterSignout` now shows 'Back to Sign In' link instead of 'Sign out'
- Close session on 'Sign Out'

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-282546](https://oktainc.atlassian.net/browse/OKTA-282546)
https://github.com/okta/okta-signin-widget/issues/1071


src/PasswordResetController.js
```diff
@@ -125,7 +125,7 @@ export default FormController.extend({
     });
 
     if (!this.settings.get('features.hideBackToSignInForReset')) {
-      this.addFooter(FooterSignout);
+      this.addFooter(FooterSignout, { linkText: loc('signout', 'login'), linkClassName: 'goto', closeSession: true });
```
haishengwu-okta, line 128: why passwordReset has special ?

haishengwu-okta (commented):
Could you consolidate 
```
goback = Back to Sign In
## used in OIE.
backToSignin = Back to sign in
```
to
```
goback = Back to sign in
```
and change `backToSignin` to `goback` for v2 code.

src/PasswordResetController.js
denysoblohin-okta: `FooterSignout` is used in 3 controllers: RecoveryChallenge, RecoveryQuestion, PasswordReset.
For first 2 ones `FooterSignout` renders 'Back' link and does not clear session. For PasswordReset it renders 'Sign Out' and closes session.
If I'm not mistaken, in case of password reset user is already logged in, and in first 2 cases user is in recovery process and is not logged in

denysoblohin-okta:
Changed `backToSignin` to `goback`

src/views/expired-password/Footer.js
```diff
@@ -35,6 +37,10 @@ export default View.extend({
         .doTransaction(function(transaction) {
           return transaction.cancel();
         })
+        .then(function() {
+          const authClient = self.options.appState.settings.authClient;
```
haishengwu-okta, line 41: - probably can just be `self.settigs.authClient` or `self.settings.getAuthClient()`. (see line 45 below)
- ditto for another file

src/PasswordResetController.js
haishengwu-okta: A more general solution would be 
```
 if authClient.session.sessionExists() then
  authClient.session.closeSession();
```

The implication is it triggers 2 API calls.

@aarongranick-okta any thoughts?

aarongranick-okta: It would be good to make fewest XHR calls in this case. calling `sessionExists` does a GET on `/sessions/me` and `closeSession` does a DELETE on the same endpoint. I think we can probably just call `closeSession` and catch/ignore any error thrown from it.

haishengwu-okta: Fair enough.
It might trigger Splunk alert on the error when `closeSession` on page where there are no session at all.

sherizada-okta: @denysoblohin-okta poking in because I'm working on a customer issue where on v1 password reset they expect only "Back to sign in" not "Sign out" even when they have the `signOutLink` customization because the user is not signed in so it doesn't make sense. This UX has also been confirmed by @vincentvoong-okta on[ this thread](https://okta.slack.com/archives/CCA77QVL2/p1619479941015500).

denysoblohin-okta: @haishengwu-okta @sherizada-okta I've changed 'Sign out' to 'Back to sign in' for password reset flow.
The only flow that uses 'Sign out' link with closing session for v1 is password expiring.

aarongranick-okta: @haishengwu-okta "It might trigger Splunk alert on the error when closeSession on page where there are no session at all."

Do we know if it will or not? This seems fairly important. May be worth 2 xhrs to avoid this.

src/RecoveryChallengeController.js
```diff
@@ -154,7 +154,7 @@ export default FormController.extend({
     }
 
     if (!this.settings.get('features.hideBackToSignInForReset')) {
-      this.add(new FooterSignout(_.extend(this.toJSON(), { linkText: loc('goback', 'login'), linkClassName: '' })));
+      this.add(new FooterSignout(this.toJSON()));
```
haishengwu-okta, line 157: `this.add(FooterSignout)` probably good enough like other places.

src/views/shared/FooterSignout.js
```diff
@@ -46,8 +46,8 @@ export default View.extend({
   },
   getTemplateData: function() {
     return {
-      linkClassName: _.isUndefined(this.options.linkClassName) ? 'goto' : this.options.linkClassName,
```
haishengwu-okta, line 49: why remove `goto`

haishengwu-okta (commented):
you need rebase to `5.6`

src/views/shared/FooterSignout.js
denysoblohin-okta: `.goto` will change float to right.
Right float seems good for 'Sign out', but for 'Back to sign in' left float seems more appropriate (because of "Back" word).
Need thought from UX.

src/RecoveryChallengeController.js
denysoblohin-okta: Done

src/PasswordResetController.js
denysoblohin-okta: @haishengwu-okta Added checking for session existence before session deletion

denysoblohin-okta:
@haishengwu-okta Same PR for 5.6 branch - https://github.com/okta/okta-signin-widget/pull/1932

PR #1837, opened 2021-04-12
chore: update description text for webauthn enroll

TianGan-okta:
## Description:
Updated the discription as designer requested

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
image

### Reviewers:
@magizh-okta

### Issue:

- [OKTA-382766](https://oktainc.atlassian.net/browse/OKTA-382766)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -765,7 +765,7 @@ oie.email.return.link.expired.title = Verify with your email
 oie.webauthn.label = Security Key or Biometric Authenticator
 oie.webauthn.description = Use a security key or a biometric authenticator to sign in
 oie.enroll.webauthn.title = Set up security key or biometric authenticator
-oie.enroll.webauthn.instructions = Your browser will prompt to register a security key or biometric authenticator (Windows Hello, Touch ID, etc.). Follow the instructions to complete enrollment.
+oie.enroll.webauthn.instructions = You will be prompted to register a security key or biometric authenticator (Windows Hello, Touch ID, Face ID, etc.). Follow the instructions to complete set up.
```
magizh-okta, line 768: I keep forgetting this. For translators to look at this and update other strings dont we have to change the key when we update text ? /cc @santhoshbalakrishnan-okta

TianGan-okta: I think only if there is parameter change. But yeah @santhoshbalakrishnan-okta please confirm

mauriciocastillosilva-okta: we don't have to update keys anymore, that was before because we wanted to prevent potential errors in other languages (say there was 1 param now, as Tian was pointing out). But we have moved away from that idea and now all edits are allowed. (still, if you are changing the parameters for the string in a login flow, which is critical, might be good to double check if it possible to do it in the same week we get translations, but not a requirement)

magizh-okta (approved):

PR #1843, opened 2021-04-14
Adds test for button state on Identify

jmelberg-okta:
## Description:

- Adds tests to verify the button state is toggled back to normal when errors occur on the Identify page

## PR Checklist

- [x] Added unit tests?

### Issue:

- [OKTA-386981](https://oktainc.atlassian.net/browse/OKTA-386981)

### DIFF
--- a/test/testcafe/spec/Identify_spec.js
+++ b/test/testcafe/spec/Identify_spec.js
@@ -33,6 +33,24 @@ const identifyMockWithFingerprint = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/challenge')
   .respond(xhrAuthenticatorOVTotp);
 
+const identifyMockWithFingerprintError = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(xhrIdentify)
+  .onRequestTo('http://localhost:3000/auth/services/devicefingerprint')
+  .respond(`
+  <html>
+    <script>
+      const message = JSON.stringify({
+        type: 'FingerprintAvailable',
+        fingerprint: 'mock-device-fingerprint',
+      });
+      window.parent.postMessage(message, window.location.href);
+    </script>
+  </html>
+  `)
+  .onRequestTo('http://localhost:3000/idp/idx/identify')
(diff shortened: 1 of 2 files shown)


test/testcafe/spec/Identify_spec.js
```diff
@@ -246,3 +266,27 @@ test.requestHooks(identifyRequestLogger, identifyMockWithFingerprint)('should co
...
+
+  await rerenderWidget({
+    features: {
+      deviceFingerprinting: true,
+    }
+  });
+
+  await identityPage.fillIdentifierField('Test Identifier');
+  await identityPage.clickNextButton();
+
```
pradeepdewda-okta, line 281: we should also check that `Next` button is now in `reloading` state once you click on it.. and then at line 290 this check `await t.expect(identityPage.getSaveButtonLabel()).eql('Next');` would make more sense..

jmelberg-okta: This is going to be problematic to test in `testcafe`, as the behavior of the test is to validate that we can proceed into the next step of the idx pipeline. We _could_ add `setTimeout` handlers to trigger the response after some set amount of time, but this a discouraged pattern.

IMO - this would be much better handled at the unit test layer, asserting the button changes state when pressed. Adding code coverage there, and additional `FormController` logic would be a great follow-up task!

test/testcafe/spec/Identify_spec.js
```diff
@@ -246,3 +266,27 @@ test.requestHooks(identifyRequestLogger, identifyMockWithFingerprint)('should co
   const factorReqHeaders = factorReq.headers;
   await t.expect(factorReqHeaders['x-device-fingerprint']).notOk();
 });
+
+test.requestHooks(identifyRequestLogger, identifyMockWithFingerprintError)('should continue to compute device fingerprint if there are errors', async t => {
```
mauriciocastillosilva-okta, line 270: nit: test is to know if the header is attached right? maybe change the description to 
`adds device fingerprint header and resets submit button if there is an error`?

mauriciocastillosilva-okta (approved):
looks good to me, just a suggestion to change a test description

PR #1844, opened 2021-04-14
fix: handle rate limit for email authenticator

worldleelavivat-okta:
## Description:

- Add 60 seconds delay before restart polling when encounter rate limit error
- Hide error message when encounter rate limit error

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Before:
Screen Shot 2021-04-14 at 3 51 41 PM
Screen Shot 2021-04-14 at 3 51 37 PM
Screen Shot 2021-04-14 at 3 51 29 PM
Screen Shot 2021-04-14 at 3 51 23 PM

After:
Screen Shot 2021-04-14 at 3 54 15 PM

https://user-images.githubusercontent.com/39062268/114920207-31a03d00-9df7-11eb-9101-8d86e0d69e3a.mov

### Reviewers:
- @pradeepdewda-okta 
- @anipendakur-okta 
- @haishengwu-okta 

### Issue:

- [OKTA-384169](https://oktainc.atlassian.net/browse/OKTA-384169)

### DIFF
--- a/playground/mocks/data/idp/idx/error-429-api-limit-exceeded.json
+++ b/playground/mocks/data/idp/idx/error-429-api-limit-exceeded.json
@@ -1,0 +1,7 @@
+{
+  "errorCode": "E0000047",
+  "errorSummary": "API call exceeded rate limit due to too many requests.",
+  "errorLink": "E0000047",
+  "errorId": "oaeD09ai4xzT4WpXsSwHn0Pxw",
+  "errorCauses": []
+}
--- a/src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
+++ b/src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
@@ -77,6 +77,18 @@ const Body = BaseFormWithPolling.extend(Object.assign(
       BaseForm.prototype.triggerAfterError.apply(this, arguments);
       this.stopPolling();
 
+      // Wait 1 min before polling again & do not show error message when encounter rate limit error
+      if (error.responseJSON?.errorSummaryKeys?.includes('tooManyRequests') ||
+        error.responseJSON?.errorCode === 'E0000047') {
+        setTimeout(() => {
+          model.trigger('clearFormError');
+        }, 0);
+        setTimeout(() => {
+          this.startPolling();
+        }, 60000);
+        return;
+      }
+
       // Polling needs to be resumed if it's a form error and session is still valid
       if(!error.responseJSON?.errorSummaryKeys?.includes('idx.session.expired')) {
         this.startPolling();
--- a/test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
+++ b/test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
@@ -15,6 +15,8 @@ import magicLinkReturnTab from '../../../playground/mocks/data/idp/idx/terminal-
 import magicLinkExpired from '../../../playground/mocks/data/idp/idx/terminal-return-expired-email';
 import terminalTransferedEmail from '../../../playground/mocks/data/idp/idx/terminal-transfered-email';
 import sessionExpired from '../../../playground/mocks/data/idp/idx/error-session-expired';
+import tooManyRequest from '../../../playground/mocks/data/idp/idx/error-429-too-many-request';
+import apiLimitExeeeded from '../../../playground/mocks/data/idp/idx/error-429-api-limit-exceeded';
 
 const emailVerificationEmptyProfile = JSON.parse(JSON.stringify(emailVerificationNoProfile));
 // add empty profile to test
@@ -98,6 +100,18 @@ const dynamicRefreshShortIntervalMock = RequestMock()
 const dynamicRefreshLongIntervalMock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/challenge/poll')
   .respond(emailVerificationPollingLong);
+
+const tooManyRequestMock = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(emailVerificationPolling)
+  .onRequestTo('http://localhost:3000/idp/idx/challenge/poll')
+  .respond(tooManyRequest, 429);
+
+const apiLimitExeededMock = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(emailVerificationPolling)
+  .onRequestTo('http://localhost:3000/idp/idx/challenge/poll')
+  .respond(apiLimitExeeeded, 429);
 
 
 fixture('Challenge Email Authenticator Form');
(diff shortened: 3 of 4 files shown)


src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -77,6 +77,16 @@ const Body = BaseFormWithPolling.extend(Object.assign(
       BaseForm.prototype.triggerAfterError.apply(this, arguments);
       this.stopPolling();
 
+      // Wait 1 min before polling again & do not show error message when encounter rate limit error
+      if (error.responseJSON?.errorSummaryKeys?.includes('tooManyRequests') ||
+        error.responseJSON?.errorCode === 'E0000047') {
+        model.trigger('clearFormError');
```
worldleelavivat-okta, line 83: @pradeepdewda-okta is there a way to hide error message somewhere here? I can't seem to remove it

worldleelavivat-okta: Hiding error message can be discussed if the UI folks have different opinion

playground/mocks/data/idp/idx/error-429-api-limit-exceeded.json
```diff
@@ -0,0 +1,7 @@
+{
+  "errorCode": "E0000047",
```
haishengwu-okta, line 2: curious why api limit has different structure with `too many request`

haishengwu-okta (commented):
LGTM

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -77,6 +77,18 @@ const Body = BaseFormWithPolling.extend(Object.assign(
       BaseForm.prototype.triggerAfterError.apply(this, arguments);
       this.stopPolling();
 
+      // Wait 1 min before polling again & do not show error message when encounter rate limit error
+      if (error.responseJSON?.errorSummaryKeys?.includes('tooManyRequests') ||
+        error.responseJSON?.errorCode === 'E0000047') {
+        setTimeout(() => {
+          model.trigger('clearFormError');
+        }, 1);
```
anipendakur-okta, line 85: Any reason why you wanted `1` here instead of `0` ? `0` basically implies as soon as possible.

anipendakur-okta:
Can you attach a screenshot of the widget in the `after` scenario as well pls?

anipendakur-okta:
The only other question I have is - 
This seems like a generic 429 response with `tooManyRequests` key. This isn't specific for `email` or any authenticator. Are we ok to have the 60 seconds delay before we restart polling regardless what authenticator/scenario it is? Just wanted to confirm..

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
worldleelavivat-okta: Sure let me update 0. I thought I need a little delay to make sure that the error is processed first. Tested with 0 and it work.

worldleelavivat-okta:
@anipendakur-okta 
Added the video in `after` section.

The code is in `BaseAuthenticatorEmailView.js` so it should impact and handled ratelimit for email polling only without impacting other scenario.

anipendakur-okta:
> @anipendakur-okta
> Added the video in `after` section.
> 
> The code is in `BaseAuthenticatorEmailView.js` so it should impact and handled ratelimit for email polling only without impacting other scenario.

Right, since the code is generic, do we not want to move it one step up to `BaseAuthenticatorWithPolling` instead of `BaseAuthenticatorEmailView` ? I guess I'm trying to understand if this is implemented only for Email in backend or in a common part of code. It might be better to be inline with that.

worldleelavivat-okta:
@anipendakur-okta 
I am not sure if the other team is planning to handle ratelimit the same way as we did here. Moving the code up to `BaseAuthenticatorWithPolling`, means it would impact the `DeviceChallengePollView.js` and I would rather have it impact limited scope and if device team would like to handle ratelimit the same way then we can move the code up to that level afterward.

anipendakur-okta:
> @anipendakur-okta
> I am not sure if the other team is planning to handle ratelimit the same way as we did here. Moving the code up to `BaseAuthenticatorWithPolling`, means it would impact the `DeviceChallengePollView.js` and I would rather have it impact limited scope and if device team would like to handle ratelimit the same way then we can move the code up to that level afterward.

I get that. So currently we get 429s for only `Email` related polling then? If so we're good to go. If we're getting 429s for other authenticators/ scenarios, we don't want the error to be shown for those on the SIW. Sorry if my question was vague..

playground/mocks/data/idp/idx/error-429-api-limit-exceeded.json
thomasmaslen-okta: > curious why api limit has different structure with `too many request`

My 0.02:  API rate-limiting such as this is generic Okta monolith HTTP / REST API functionality, not specific to IDX, so I look at this from the other end of the spectrum -- IMO having any sort of JSON response body is sort of an optional extra, not fundamental, so I wouldn't necessarily expect it to be an IdxResponse.  To me the fundamental pieces are the HTTP 429 response code and the HTTP response headers (not body) that convey rate-limiting information, but at present we can't get at any of those (OKTA-386750);  absent that, this PR has to fall back to trying to glean information from the Okta (but not IDX) JSON response body.

test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
```diff
@@ -310,4 +324,46 @@ test
...
+    await t.expect(logger.count(
+      record => record.response.statusCode === 429 &&
+        record.request.url.match(/poll/)
+    )).eql(1);
+
+    await t.removeRequestHooks(tooManyRequestMock);
+    await t.addRequestHooks(validOTPmock);
+
+    // Pause for 60 sec before sending request
+    await t.wait(60000);
```
aarongranick-okta, line 343: Is this a real 60 seconds, or is the clock being mocked? Could these tests be done in jest?

worldleelavivat-okta: real 60 secs...i thought it might be okay since we have above with 30 seconds wait as well. I am open to move it if possible.

worldleelavivat-okta:
@anipendakur-okta 
429s can occur with any endpoint if it exceed the threshold. I didn't want to move it up to limit the scope since I do not know how to do testing for devices challenge.

thomasmaslen-okta:
> > I get that. So currently we get 429s for only Email related polling then? If so we're good to go. If we're getting 429s for other authenticators/ scenarios, we don't want the error to be shown for those on the SIW. Sorry if my question was vague..

> @anipendakur-okta
> 429s can occur with any endpoint if it exceed the threshold. I didn't want to move it up to limit the scope since I do not know how to do testing for devices challenge.

👍 

My 0.02:  while I agree that HTTP 429 responses are a general problem for the SIW (and particularly for the various flavours of IDX polling) and it would behoove us to consider and solve the overall problem, I think it's a good thing to start out with this PR addressing the problem only for Email Authenticator polling (particularly since this PR has to use an approach that is just the least-worst we can do at present), then see how well it works out for this use case before we try to generically solve it for all IDX polling.

anipendakur-okta (approved):

PR #1846, opened 2021-04-14
bug: Fixes auto-redirect for IdP and adds error messages for i18n

jmelberg-okta:
## Description:

- Fixes an issue where the Widget would auto-redirect when there was only **one** remediation form
- Adds i18n messages for IdP Authenticator flows

## PR Checklist

- [x] Added e2e tests

### Reviewers:

- @haishengwu-okta 

### Issue:

- [OKTA-380834](https://oktainc.atlassian.net/browse/OKTA-380834)
- [OKTA-386387](https://oktainc.atlassian.net/browse/OKTA-386387)

### DIFF
--- a/src/v2/ion/responseTransformer.js
+++ b/src/v2/ion/responseTransformer.js
@@ -228,8 +228,11 @@ const convert = (settings, idx = {}, lastResult = null) => {
     { idx }
   );
 
+  // Override the `result` to handle custom IdP login buttons
+  // and update the form for IdP Authenticators.
   injectIdPConfigButtonToRemediation(settings, result);
   modifyFormNameForIdPAuthenticator(result);
+
   if (!result.messages) {
     // Only redirect to the IdP if we are not in an error flow
     convertRedirectIdPToSuccessRedirectIffOneIdp(settings, result, lastResult);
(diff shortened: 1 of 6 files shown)


src/v2/ion/responseTransformer.js
```diff
@@ -229,13 +229,12 @@ const convert = (settings, idx = {}, lastResult = null) => {
   );
 
   injectIdPConfigButtonToRemediation(settings, result);
+  modifyFormNameForIdPAuthenticator(result);
```
jmelberg-okta, line 232: @haishengwu-okta - When we made this change with #1780, we didn't account for the IdP Authenticator having a single remediation response of `redirect-idp`. As a result, we'd auto-redirect to the IdP when we're receive a response [like this](https://github.com/okta/okta-signin-widget/blob/998bcba0855c0ed32f27a79710745c12ae6215f5/playground/mocks/data/idp/idx/authenticator-verification-idp-single-remediation.json).

Invoking the form-name change here ensures the `convertRedirectIdPToSuccessRedirectIffOneIdp` doesn't accidentally trigger the redirect.

haishengwu-okta, line 232: 🤦 🤦 

seems worth to mention orders of function call matters as comments

haishengwu-okta (approved):

src/v2/ion/responseTransformer.js
jmelberg-okta: Good idea - will add a comment here!

PR #1851, opened 2021-04-19
Add support for processCreds for SIW v2

conorhanrahan-okta:
## Description:

This PR adds support for processCreds in an IDX sign in widget. Not every view has access to the current user's login. Some views contain only a password input, and the server response contains a user that looks like this:
```
  "user": {
    "type": "object",
    "value": {
      "id": "00uwb8GLwf1HED5Xs0g3"
    }
  },
```

If we want those views to call `processCreds`, we will need to have the backend send something like this:
```
  "user": {
    "type": "object",
    "value": {
      "id": "I9bvFiq01cRFgbn",
      "passwordChanged": "2019-05-03T19:00:00.000Z",
      "profile": {
        "login": "foo@example.com",
        "firstName": "Foo",
        "lastName": "Bar",
        "locale": "en-us",
        "timeZone": "UTC"
      }
    }
  },
```
I'm guessing the user's profile is not being sent over for security reasons.

These views will call processCreds
- IdentifierView (with a password)
- EnrollAuthenticatorPasswordView

These views will not call processCreds because the FE doesn't have access to the user's username:
- ChallengeAuthenticatorPasswordView
- ReEnrollAuthenticatorPasswordView
- ReEnrollAuthenticatorWarningPasswordView
- ResetAuthenticatorPasswordView

## PR Checklist

- [ ] Added unit tests?

- [x] Added e2e tests

### Issue:

- [OKTA-314099](https://oktainc.atlassian.net/browse/OKTA-XXXXXX)

### DIFF
--- a/src/v2/view-builder/views/IdentifierView.js
+++ b/src/v2/view-builder/views/IdentifierView.js
@@ -20,28 +20,41 @@ const Body = BaseForm.extend({
 
   saveForm() {
     // Ideally this can be added to a "preSaveForm" handler - but keeping this here for now.
+    this.model.trigger('request');
+
+    this.getDeviceFingerprintPromise()
+      .then(() => this.getProcessCredsPromise())
+      .then(() => {
+        BaseForm.prototype.saveForm.apply(this, arguments);
+      });
+  },
+
+  getDeviceFingerprintPromise() {
     if (!this.settings.get('features.deviceFingerprinting')) {
-      BaseForm.prototype.saveForm.apply(this, arguments);
-      return;
+      return Promise.resolve();
     }
 
-    // Before the XHR is made for "identify", we'll generate this one-time use fingerprint via
-    // a hidden-iframe (similar to authn/v1 flows)
     const fingerprintData = {
       oktaDomainUrl: this.settings.get('baseUrl'),
       element: this.$el,
     };
 
-    this.model.trigger('request');
-
-    DeviceFingerprinting.generateDeviceFingerprint(fingerprintData)
+    return DeviceFingerprinting.generateDeviceFingerprint(fingerprintData)
       .then(fingerprint => {
         this.options.appState.set('deviceFingerprint', fingerprint);
       })
-      .catch(() => { /* Keep going even if device fingerprint fails */ })
-      .finally(() => {
-        BaseForm.prototype.saveForm.apply(this, arguments);
-      });
+      .catch(() => { /* Keep going even if device fingerprint fails */ });
+  },
+
+  getProcessCredsPromise() {
+    const creds = {
+      username: this.model.get('identifier'),
+    };
+    const password = this.model.get('credentials.passcode');
+    if (password) {
+      creds.password = password;
+    }
+    return this.settings.processCreds(creds);
   },
 
   render() {
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -147,6 +147,7 @@ const authn = {
     // 'mfa-required-email',
     // 'unauthenticated',
     'admin-consent-required',
+    // 'password-expired',
   ],
   '/api/v1/authn': [
     'unauthenticated',
--- a/playground/mocks/data/api/v1/authn/password-expired.json
+++ b/playground/mocks/data/api/v1/authn/password-expired.json
@@ -1,0 +1,47 @@
+{
+    "stateToken": "007ucIX7PATyn94hsHfOLVaXAmOBkKHWnOOLG43bsb",
+    "expiresAt": "2015-11-03T10:15:57.000Z",
+    "status": "PASSWORD_EXPIRED",
+    "relayState": "/myapp/some/deep/link/i/want/to/return/to",
+    "_embedded": {
+        "user": {
+        "id": "00ub0oNGTSWTBKOLGLNR",
+        "passwordChanged": "2015-09-08T20:14:45.000Z",
+        "profile": {
+            "login": "dade.murphy@example.com",
+            "firstName": "Dade",
+            "lastName": "Murphy",
+            "locale": "en_US",
+            "timeZone": "America/Los_Angeles"
+        }
+        },
+        "policy": {
+        "complexity": {
+            "minLength": 8,
+            "minLowerCase": 1,
+            "minUpperCase": 1,
+            "minNumber": 1,
+            "minSymbol": 0
+        }
+        }
+    },
+    "_links": {
+        "next": {
+        "name": "changePassword",
+        "href": "http://localhost:3000/api/v1/authn/credentials/change_password",
+        "hints": {
+            "allow": [
+            "POST"
+            ]
+        }
+        },
+        "cancel": {
+        "href": "http://localhost:3000/api/v1/authn/cancel",
+        "hints": {
+            "allow": [
+            "POST"
+            ]
+        }
+        }
+    }
+}
--- a/test/testcafe/spec/IdentifyWithPassword_spec.js
+++ b/test/testcafe/spec/IdentifyWithPassword_spec.js
@@ -31,6 +31,11 @@ const rerenderWidget = ClientFunction((settings) => {
   window.renderPlaygroundWidget(settings);
 });
 
+const fakeProcessCreds = ({ username, password }) => {
+  // eslint-disable-next-line no-console
+  console.log(`processCreds called with ${username} and ${password}`);
+};
+
 fixture('Identify + Password');
 
 async function setup(t) {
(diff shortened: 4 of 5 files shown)


conorhanrahan-okta:
Got all Selenium tests to pass on the downstream bacon: https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&branch=d16t-okta-signin-widget-8fa47ee-607d04ce

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
```diff
@@ -73,6 +73,18 @@ const Body = BaseForm.extend({
...
+  saveForm() {
+    let processCredsPromise = Promise.resolve();
+    const username = this.options.appState.get('user')?.profile?.login;
+    if (username) {
+      const password = this.model.get('credentials.passcode');
+      processCredsPromise = this.settings.processCreds({ username, password });
+    }
+    processCredsPromise.then(() => {
+      BaseForm.prototype.saveForm.apply(this, arguments);
+    });
```
sherizada-okta, line 87: It looks like the Settings method `processCreds` takes care of ensuring `this.settings.processCreds` always returns a promise.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -20,28 +20,42 @@ const Body = BaseForm.extend({
...
+  getProcessCredsPromise() {
+    const password = this.model.get('credentials.passcode');
+    if (!password) {
+      return Promise.resolve();
+    }
+    const creds = {
+      username: this.model.get('identifier'),
+      password
+    };
+    return this.settings.processCreds(creds);
```
sherizada-okta, line 58: I think the checks for identifier/password should either
- go inside `this.settings.processCreds`
or
- not happen at all and just let the user handle it

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
sherizada-okta: And same thing about the identifer check here (let processCreds handle it either on our end of theirs)

sherizada-okta:
Can you clean up the PR title?

conorhanrahan-okta:
Will this need a Selenium test or is the TestCafe tests enough? Not sure if it's possible to do `widgetrc` stuff during a Selenium test

src/models/Settings.js
```diff
@@ -397,9 +397,9 @@ export default Model.extend({
 
   processCreds: function(creds) {
     const processCreds = this.get('processCreds');
-
+    const credentialsAreComplete = creds.username && creds.password;
```
haishengwu-okta, line 400: why need add this check?

haishengwu-okta (commented):
looks like any password related View shall invoke this call? 
- in v1, we have PrimaryAuth, PasswordExpired, PasswordReset

src/v2/view-builder/views/IdentifierView.js
haishengwu-okta: what about avoid to invoke `this.settings.processCreds` if it's identify first flow (only has user name field)

conorhanrahan-okta: That's how I did it originally in this commit: https://github.com/okta/okta-signin-widget/pull/1851/commits/8fa47ee77a4940a66d1c56ab295364a94bfec53d

After I saw I changed it to call processCreds with partial credentials and do the check inside `this.settings.processCreds`. Which approach do we want to take here? I don't feel strongly either way

src/models/Settings.js
conorhanrahan-okta: See the discussion here: https://github.com/okta/okta-signin-widget/pull/1851#discussion_r616153242

conorhanrahan-okta:
> looks like any password related View shall invoke this call?
> 
> * in v1, we have PrimaryAuth, PasswordExpired, PasswordReset

The call to `processCreds` will only happen in the views listed in the description. In most of the password views we don't have access to the user's identifier, so for those views there will be no underlying call to the user's `processCreds`.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -20,28 +20,38 @@ const Body = BaseForm.extend({
 
   saveForm() {
     // Ideally this can be added to a "preSaveForm" handler - but keeping this here for now.
+    this.model.trigger('request');
+
+    this.getDeviceFingerprintPromise()
+      .then(() => this.getProcessCredsPromise())
+      .then(() => {
+        BaseForm.prototype.saveForm.apply(this, arguments);
```
sherizada-okta, line 28: `arguments` here should be `undefined` because you're inside an arrow function

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
```diff
@@ -73,6 +73,15 @@ const Body = BaseForm.extend({
...
       }
     ]);
+  },
+
+  saveForm() {
+    this.settings.processCreds({
+      username: this.options.appState.get('user')?.profile?.login,
+      password: this.model.get('credentials.passcode')
+    }).then(() => {
+      BaseForm.prototype.saveForm.apply(this, arguments);
```
sherizada-okta, line 83: arrow functions have no `arguments`

test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
```diff
@@ -84,3 +88,18 @@ test('should have the correct reqiurements', async t => {
   await t.expect(enrollPasswordPage.getRequirements()).contains('Your password cannot be any of your last 4 passwords');
   await t.expect(enrollPasswordPage.getRequirements()).contains('A lowercase letter');
 });
+
+test('should call processCreds when the user submits the form', async t => {
+  const enrollPasswordPage = await setup(t);
+  await rerenderWidget({
+    processCreds: ({ username, password }) => {
+      // eslint-disable-next-line no-console
+      console.log(`processCreds called with ${username} and ${password}`);
```
sherizada-okta, line 97: what about when processCreds is a promise

src/v2/view-builder/views/IdentifierView.js
haishengwu-okta: Don't have strong opinion either way but leaning toward to not make changes to `this.settings.processCreds` method. hence the ask.

src/v2/view-builder/views/password/EnrollAuthenticatorPasswordView.js
conorhanrahan-okta: I just looked into this, it looks like the arguments variable works fine here. Arrow function's don't set the `arguments` variable at all, which means it get the `arguments` value from the outer scope. I set a breakpoint to double check. The existing device fingerprinting logic takes the same approach.

test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
conorhanrahan-okta: I modified one of these tests to use the `processCreds(creds, callback)` function signature instead, I am guessing that's what you were referring to. All the promise stuff happens in `this.settings.processCreds`, not the`this.settings.get('processCreds')` that the user supplies.

src/models/Settings.js
conorhanrahan-okta: I've reverted this change

src/v2/view-builder/views/IdentifierView.js
conorhanrahan-okta: Reverted this change

sherizada-okta (dismissed):

sherizada-okta:
Can you also add a row for this new OIE behavior here: https://oktawiki.atlassian.net/wiki/spaces/eng/pages/1894023233/OIE+Vs.+Legacy+behaviour+drift

If you search for "SIW customization" you'll get some examples. It sounds like the drift is OIE hook won't support the same flows.

conorhanrahan-okta:
I spent some time looking at the v1 behavior and realized a few places where I was differing from existing processCreds behavior:
- In the Identifier view, I now call processCreds with an `undefined` username when there is no password in the view. This is similar to the v1 behavior in `PrimaryAuthForm.js` when `features.passwordlessAuth` is enabled
- I realized that processCreds is not being called when enrolling a password in v1. So I deleted the commit where I added process creds to v2's `EnrollAuthenticatorPasswordView`

To summarize the new behavior:
- `processCreds` will only be enabled for the Identifier which includes a password field.
- `processCreds` will not be called when in the expired password or reset password views. This is different than the v1 behavior. If we want to support this, we will need the BE to pass the profile in with the user (see [this Slack comment](https://okta.slack.com/archives/CAZH2MWEL/p1618513936470300))

Sorry, I regret not double checking this earlier! 😖

test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
sherizada-okta: Right, like a test case supporting this example from the [readme](https://github.com/okta/okta-signin-widget#username-and-password):
```
// Passed a creds object {username, password} and a callback for further processing
processCreds: function (creds, callback) {
  // This example demonstrates optional legacy form-based login
  $.ajax({
    method: "POST",
    url: "/logintype",
    data: {
      username : creds.username
    },
    success: function (logintype) {
      if (logintype == "LEGACY") {
        $('#legacyUser').val(creds.username);
        $('#legacyPassword').val(creds.password);
        $('#legacyLogonForm').submit();
      } else {
        callback();
      }
    }
  });
}
```

sherizada-okta:
So in these instances, it's best to ask the PM what the best course of action is. Which I've done [here](https://okta.slack.com/archives/G01NZUZJCBH/p1619543746072800).

conorhanrahan-okta:
Closing because this will be too different from the old behavior. See https://okta.slack.com/archives/G01NZUZJCBH/p1619543746072800

PR #1859, opened 2021-04-21
test: Another round of fixes for en leaks suite

nikhilvenkatraman-okta:
Resolves: OKTA-389024

## Description:

- Improved logic to exclude multiple no translate sections from UI 
- Use `innerText` instead of `textContent` to avoid including elements with display:none attributes on UI 
- Unignore few more mocks 
- We have over 100 test passing now 🎉 

## PR Checklist

- [x] Added e2e tests

### Screenshot/Video:

Screen Shot 2021-04-21 at 11 51 51 AM

### Reviewers:

@haishengwu-okta @jmelberg-okta 

### Issue:

- [OKTA-389024](https://oktainc.atlassian.net/browse/OKTA-389024)

### DIFF
--- a/playground/main.js
+++ b/playground/main.js
@@ -2,7 +2,7 @@
 /* eslint no-console: 0 */
 
 import signinWidgetOptions from '../.widgetrc.js';
-import { assertNoEnglishLeaks } from '../LocaleUtils';
+import { assertNoEnglishLeaks } from '../playground/LocaleUtils';
 
 let signIn;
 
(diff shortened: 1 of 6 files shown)


LocaleUtils/index.js
```diff
@@ -7,7 +7,10 @@ export async function assertNoEnglishLeaks(mockFile, viewText, noTranslationCont
   }
   // Exclude all elements with a class of no-translate. Example phone numbers
   if (noTranslationContent.length) {
-    viewText = viewText.split(noTranslationContent).join('');
+    //replace each noTranslationContent
+    for (var i = 0; i < noTranslationContent.length; i++) {
```
jmelberg-okta, line 11: Can we use `forEach` here since `noTranslationContent` is going to be an Array? IMO this is a bit easier to follow.

playground/main.js
```diff
@@ -85,15 +85,19 @@ const renderPlaygroundWidget = (options = {}) => {
 
     // assert english leaks if locale set to ok-PL
     if (signinWidgetOptions.language === 'ok-PL') {
-      // wait for view to finish rendering
-      const viewText = document.getElementById('okta-sign-in').textContent;
+      //Use innerText to avoid including hidden elements
```
jmelberg-okta, line 88: This is tricky, as some "hidden" elements may still have HTML properties that need to be read via screen readers. Probably OK to hide this for now, but something to consider when we ramp up our a11y testing.

LocaleUtils/index.js
```diff
@@ -7,7 +7,10 @@ export async function assertNoEnglishLeaks(mockFile, viewText, noTranslationCont
   }
   // Exclude all elements with a class of no-translate. Example phone numbers
   if (noTranslationContent.length) {
-    viewText = viewText.split(noTranslationContent).join('');
```
jmelberg-okta, line 10: Can we also move `LocaleUtils/index.js` outside the root of the project and into `playground`?

jmelberg-okta (approved):

LocaleUtils/index.js
nikhilvenkatraman-okta: We could, but its also used by the EnglishLeaks.spec file which is within testcafe. I wanted a common place that made sense for both the playground and our testcafe specs

jmelberg-okta: We already share mocks from the playground in our testcafe specs - so I'd suggest we nest it there for the time being.

Once we migrate to a mono-repo, we'll have a better location for shared files like this. 😄

haishengwu-okta (commented):
👍

PR #1860, opened 2021-04-21
add data-se attr to username/password on OIE Identify view

sherizada-okta:
## Description:
adds 'data-se' attribute for selenium tests, to easily grab values.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-389175](https://oktainc.atlassian.net/browse/OKTA-389175)

### DIFF
--- a/src/v2/view-builder/views/IdentifierView.js
+++ b/src/v2/view-builder/views/IdentifierView.js
@@ -115,21 +115,26 @@ const Body = BaseForm.extend({
     const newSchemas = schemas.map(schema => {
       let newSchema = { ...schema };
 
-      if (schema.name === 'identifier' && isCustomizedI18nKey(identifierExplainLabeli18nKey, settings)) {
-        newSchema = {
-          ...newSchema,
-          explain: loc(identifierExplainLabeli18nKey, 'login'),
-          'explain-top': true,
-          'data-se': 'username-form-container',
-        };
-      } else if (schema.name === 'credentials.passcode' &&
-          isCustomizedI18nKey(passwordExplainLabeli18nKey, settings)) {
-        newSchema = {
-          ...newSchema,
-          explain: loc(passwordExplainLabeli18nKey, 'login'),
-          'explain-top': true,
-          'data-se': 'password-form-container',
-        };
+      if (schema.name === 'identifier') {
+        newSchema['data-se'] = `o-form-fieldset-${schema.name}`;
+
+        if (isCustomizedI18nKey(identifierExplainLabeli18nKey, settings)) {
+          newSchema = {
+            ...newSchema,
+            explain: loc(identifierExplainLabeli18nKey, 'login'),
+            'explain-top': true,
+          };
+        }
+      } else if (schema.name === 'credentials.passcode') {
+        newSchema['data-se'] = `o-form-fieldset-${schema.name}`;
+
+        if (isCustomizedI18nKey(passwordExplainLabeli18nKey, settings)) {
+          newSchema = {
+            ...newSchema,
+            explain: loc(passwordExplainLabeli18nKey, 'login'),
+            'explain-top': true,
+          };
+        }
       }
 
       return newSchema;
--- a/src/v2/view-builder/views/IdentifierView.js
+++ b/src/v2/view-builder/views/IdentifierView.js
@@ -116,7 +116,7 @@ const Body = BaseForm.extend({
       let newSchema = { ...schema };
 
       if (schema.name === 'identifier') {
-        newSchema['data-se'] = `o-form-fieldset-${schema.name}`;
+        newSchema['data-se'] = 'o-form-fieldset-identifier';
 
         if (isCustomizedI18nKey(identifierExplainLabeli18nKey, settings)) {
           newSchema = {
@@ -126,7 +126,7 @@ const Body = BaseForm.extend({
           };
         }
       } else if (schema.name === 'credentials.passcode') {
-        newSchema['data-se'] = `o-form-fieldset-${schema.name}`;
+        newSchema['data-se'] = 'o-form-fieldset-credentials.passcode';
 
         if (isCustomizedI18nKey(passwordExplainLabeli18nKey, settings)) {
           newSchema = {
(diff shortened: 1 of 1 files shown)


src/v2/view-builder/views/IdentifierView.js
```diff
@@ -113,15 +113,23 @@ const Body = BaseForm.extend({
     const passwordExplainLabeli18nKey = 'primaryauth.password.tooltip';
 
     const newSchemas = schemas.map(schema => {
-      const newSchema = { ...schema };
+      let newSchema = { ...schema };
```
magizh-okta, line 117: Screen Shot 2021-04-21 at 6 56 20 PM


The inputs already have a data-se am I missing something

sherizada-okta: That's not the same spot where I need the selector. 
My use case is I want to test the label and sub-labels for username and password and they're currently not using any descriptive classnames or attrs, so I would need to base my tests off of their position in the DOM. For example: 
```
usernameContainer = cssSelector([data-se="o-form-label"]).get(0)
passwordContainer = cssSelector([data-se="o-form-label"]).get(1)
```

haishengwu-okta: what tests need those selector?
what we usually do is locate the `input` and look up for its label (`parent -> sibling`)

sherizada-okta: https://github.com/okta/okta-core/pull/52020

sherizada-okta: In v1, the form had more descriptive attributes to facilitate testing 
image

magizh-okta: ahh I see. We can pass inputId attribute in here so all inputs get a selector for the label :) 

https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/uiSchemaTransformer.js#L61

eg) http://courage.oktaskylab.com/gist/vcZejrtNR

/cc @haishengwu-okta  if you have any other suggestions.

pradeepdewda-okta (approved):
this looks good to me..

src/v2/view-builder/views/IdentifierView.js
sherizada-okta: I decided to use `data-se` instead of `inputId` because then I can also grab the sub-label (.o-form-explain)

haishengwu-okta: If it's merely for testing, I think we could enhance `FormComponent.java` by adding method like `getLabelForField`.
Otherwise, @magizh-okta 's suggestion is nice.

sherizada-okta: The sub/explain label I am also testing is not a label, so inputId would not affect those elements.
> I decided to use `data-se` instead of `inputId` because then I can also grab the sub-label (.o-form-explain)

magizh-okta: @haishengwu-okta  I guess this is for selenium tests and getting outer element in selenium is tricky. @sherizada-okta  can we add the data-se={label+fieldName} in the UISchemaTransfer so all inputs gets this goodness.
We can create a follow up too if u want to get this before cF.

sherizada-okta: I want to get this in before CF because it's blocking my monolith changes, and adding them to all elements might break existing selenium tests so that would need a monolith CI run.

haishengwu-okta: oh, yeah, it would need to utilize xpath.
the changes are too specific like v1 we specially add ID for username and password.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -113,15 +113,23 @@ const Body = BaseForm.extend({
...
 
       if (schema.name === 'identifier' && isCustomizedI18nKey(identifierExplainLabeli18nKey, settings)) {
-        newSchema.explain = loc(identifierExplainLabeli18nKey, 'login');
-        newSchema['explain-top'] = true;
+        newSchema = {
+          ...newSchema,
+          explain: loc(identifierExplainLabeli18nKey, 'login'),
+          'explain-top': true,
+          'data-se': 'username-form-container',
```
haishengwu-okta, line 123: - `o-form-fieldset-${newSchema.name}`?
- turns out this will override existing `data-se=o-form-fieldset`.  seems not really ideal.

src/v2/view-builder/views/IdentifierView.js
sherizada-okta: I think adding a data-se that includes the form name to the default ui schema config is a good idea, but it'll require
- applying it to schemas that are not extending from the uiSchemaDefaultConfig in uiSchemaTransformer
- fix breaking testcafe tests
- check against monolith se tests

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -113,15 +113,28 @@ const Body = BaseForm.extend({
...
-        newSchema.explain = loc(identifierExplainLabeli18nKey, 'login');
-        newSchema['explain-top'] = true;
-      } else if (schema.name === 'credentials.passcode' &&
-          isCustomizedI18nKey(passwordExplainLabeli18nKey, settings)) {
-        newSchema.explain = loc(passwordExplainLabeli18nKey, 'login');
-        newSchema['explain-top'] = true;
+      let newSchema = { ...schema };
+
+      if (schema.name === 'identifier') {
+        newSchema['data-se'] = `o-form-fieldset-${schema.name}`;
```
haishengwu-okta, line 119: this is generic logic and don't have to check on `schema.name`.

src/v2/view-builder/views/IdentifierView.js
haishengwu-okta: Please refactoring for general solution as follow up and get rid of this one off.

sherizada-okta: https://oktainc.atlassian.net/browse/OKTA-389414

PR #1872, opened 2021-04-24
fix: Reflects error message change in okta-core for unknown user

huanliu-okta:
## Description:

HTML tag is removed from the idx.unknown.user error message in [a PR in okta-core](https://github.com/okta/okta-core/pull/51872) because it is not desirable to return HTML tag in API return messages. 

This PR reflects the change in okta-core. Okta-core's change has green bacon regardless of this change, so I suspect this PR is not strictly necessary, i.e., the mocked output just need to match the test expectation. But adding this PR to make sure we are consistent. 

## PR Checklist

   - Run `yarn test -t jest`, which passes both before and after the change. (The test would fail if the mocked output does not match the test expectation.)
   - `yarn test -t testcafe` and `yarn test -t karma` both fail locally before and after for the same reasons that look unrelated to this change. 

- [ ] Added unit tests?
   no

- [ ] Added e2e tests
   no

    Yes 
### Reviewers:

@haishengwu-okta 

### Issue:

- [OKTA-379880](https://oktainc.atlassian.net/browse/OKTA-379880)

### DIFF
--- a/playground/mocks/data/idp/idx/identify-unknown-user.json
+++ b/playground/mocks/data/idp/idx/identify-unknown-user.json
@@ -52,7 +52,7 @@
     "type": "array",
     "value": [
       {
-        "message": "There is no account with the email <strong> test@rain.com </strong>. <a href=\"#\" class=\"js-sign-up\"> Sign up </a> for an account",
+        "message": "There is no account with the email test@rain.com. Error: idx.unknown.user",
         "i18n": {
           "key": "idx.unknown.user",
           "params": []
--- a/playground/mocks/data/idp/idx/identify-unknown-user.json
+++ b/playground/mocks/data/idp/idx/identify-unknown-user.json
@@ -52,7 +52,7 @@
     "type": "array",
     "value": [
       {
-        "message": "There is no account with the email test@rain.com. Error: idx.unknown.user",
+        "message": "There is no account with the email test@rain.com.",
         "i18n": {
           "key": "idx.unknown.user",
           "params": []
--- a/test/testcafe/spec/IdentifyUnknownUser_spec.js
+++ b/test/testcafe/spec/IdentifyUnknownUser_spec.js
@@ -33,7 +33,7 @@ test('should show messages callout for unknown user', async t => {
   await identityPage.fillIdentifierField('unknown');
   await identityPage.clickNextButton();
   await t.expect(identityPage.getUnknownUserCalloutContent())
-    .eql('There is no account with the email test@rain.com. Error: idx.unknown.user');
+    .eql('There is no account with the email test@rain.com.');
 });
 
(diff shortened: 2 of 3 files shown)


playground/mocks/data/idp/idx/identify-unknown-user.json
```diff
@@ -52,7 +52,7 @@
     "type": "array",
     "value": [
       {
-        "message": "There is no account with the email <strong> test@rain.com </strong>. <a href=\"#\" class=\"js-sign-up\"> Sign up </a> for an account",
+        "message": "There is no account with the email test@rain.com. Error: idx.unknown.user",
```
nikhilvenkatraman-okta, line 55: We have a [click handler](https://github.com/okta/okta-signin-widget/blob/f53b68168988903b0d63d4c584f68ef5693a03ee/src/v2/view-builder/views/IdentifierView.js#L155) that was listening to the click event from this anchor tag. If we are no longer supporting clicking on the link we should remove the handler as well.

pradeepdewda-okta, line 55: can you use the exact error, that server is sending ?

pradeepdewda-okta: I hope server doesn't send `Error: idx.unknown.user` with the message..

nikhilvenkatraman-okta: yeah, I am not sure why we need `Error: idx.unknown.user` as part of the message.

huanliu-okta: @nikhilvenkatraman-okta thanks for the review. The idea is to support click handler, but not with HTML tag, rather key off i18n key. See comment from @iancorcoran-okta  @pradeepdewda-okta from the accompanying https://github.com/okta/okta-core/pull/51872

pradeepdewda-okta: I haven't reviewed [this one](https://github.com/okta/okta-core/pull/51872#issuecomment-825786136) 

So we don't have to insert the error code in the message..  Following is enough to insert insert new error message, with link..
```
"i18n": {
          "key": "idx.unknown.user"
}
```

huanliu-okta:
Updated to remove the error code from message. This PR only updates the error message returned from the server. Will create a separate JIRA ticket to update the click handler to key off the error key.

huanliu-okta:
The [okta-core change](https://github.com/okta/okta-core/pull/51872) is merged. I created a [follow up ticket](https://oktainc.atlassian.net/browse/OKTA-391232) to insert the "Sign Up" link back to the error message, so that the click handler would still work.

haishengwu-okta (approved):

PR #1874, opened 2021-04-26
test: Fix English leak test for rate limit error

worldleelavivat-okta:
## Description:

Fix English leak test for rate limit error

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Reviewers:
- @haishengwu-okta 
- @nikhilvenkatraman-okta 
- @anipendakur-okta 

### Issue:

- [OKTA-390215](https://oktainc.atlassian.net/browse/OKTA-390215)


haishengwu-okta (approved):

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -983,3 +983,6 @@ oie.browser.error.NotAllowedError = The operation either timed out or was not al
 # soft_token API errors
 authfactor.challenge.soft_token.invalid_passcode = Your code doesn't match our records. Please try again.
 authfactor.challenge.soft_token.used_passcode = Each code can only be used once. Please wait for a new code and try again.
+
+# rate limit
+tooManyRequests = Too many requests
```
nikhilvenkatraman-okta, line 988: Can we add an `oie/idx` prefix to this key ?. It is a format that we follow for all oie keys. We are also looking to add some automation for customization where we need to know if a key is an oie key or not.

worldleelavivat-okta: I double checked the error and look like `tooManyRequests` in the currently format is mainly used for OIE. 

I have added `oie` prefix to this key. The V1 error look more like error in `error-429-api-limit-exceeded.json`.

src/v2/ion/i18nTransformer.js
```diff
@@ -128,6 +128,8 @@ const I18N_OVERRIDE_MAPPINGS = {
   'oie.session.expired' : 'oie.idx.session.expired',
 
   'activate-device.userCode' : 'oie.device.code.activate.label',
+
+  'tooManyRequests' : 'oie.tooManyRequests',
```
nikhilvenkatraman-okta, line 132: I don't think you need this change. This is only for input labels that are sent from the backend.

nikhilvenkatraman-okta (approved):
I think we can remove the i18nTransformer change.

nikhilvenkatraman-okta (changes requested):
Looks like we dont have this change in the backend so this will break the widget. We should change the backend or revert the change for oie prefix

worldleelavivat-okta:
@nikhilvenkatraman-okta reverted the prefix change

anipendakur-okta (approved):
lgtm

nikhilvenkatraman-okta (approved):

PR #1875, opened 2021-04-27
fix: enable lint for new mocks and ignore existing mocks

nikhilvenkatraman-okta:
Resolves: OKTA-390370

## Description:
- Enable lint for new mocks and ignore existing mocks

## PR Checklist

### Reviewers:

@haishengwu-okta @jmelberg-okta @aarongranick-okta 

### Issue:

- [OKTA-390370](https://oktainc.atlassian.net/browse/OKTA-390370)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -40,8 +40,47 @@ module.exports = {
     {
       'files': ['playground/**/*.json'],
       'rules': {
-        'local-rules/no-missing-api-keys': 1,
-      }
+        'local-rules/no-missing-api-keys': 2,
+      },
+      'overrides': [
+        {
+          'files': [
+            'playground/**/authenticator-expiry-warning-password.json',
+            'playground/**/authenticator-verification-okta-verify-reject-push.json',
+            'playground/**/error-403-security-access-denied.json',
+            'playground/**/error-with-failure-redirect.json',
+            'playground/**/error-authenticator-verify-password.json',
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -40,8 +40,47 @@ module.exports = {
...
       'files': ['playground/**/*.json'],
       'rules': {
-        'local-rules/no-missing-api-keys': 1,
-      }
+        'local-rules/no-missing-api-keys': 2,
+      },
+      'overrides': [
+        {
+          'files': [
+            'playground/**/authenticator-expiry-warning-password.json',
```
nikhilvenkatraman-okta, line 48: These mocks have known issues. Will create tickets for home teams to address.

jmelberg-okta: This is a good way to make sure the rule gets enforced by default for all **new mock files**. 🎉 

Am curious if some of these error files may actually be **correct**, as the error message is localized before the Widget consumes it. If so - are we encouraging developers to make backend changes to add an i18n property or do we intend to have another category if "ignored" files?

jmelberg-okta (approved):

.eslintrc.js
nikhilvenkatraman-okta: We want to make sure all API response have an i18n key. These are mocks that either have a i18n key itself missing from the mock or they have it in the mock but is not added to the login bundle. For cases where its missing from the mock, we want an API change to be made by the home team.

PR #1877, opened 2021-04-28
fix: add en leaks for enroll profile errors 

sherizada-okta:
## Description:
fixes english leaks in the following mocks:
    'error-new-signup-email',
    'error-new-signup-email-exists'

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:
- [OKTA-389420](https://oktainc.atlassian.net/browse/OKTA-389420)


haishengwu-okta (approved):

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -651,6 +651,9 @@ registration.default.callbackhook.error=We could not process your registration a
 registration.error.userName.invalidEmail=Invalid email address
 registration.error.password.passwordRequirementsNotMet=Password requirements were not met
 registration.error.userName.notUniqueWithinOrg=An account with that email already exists
+registration.error.notUniqueWithinOrg = A user with this Email already exists
+registration.error.invalidLoginEmail = 'Email' must be in the form of an email address
+registration.error.doesNotMatchPattern = Provided value for property 'Email' does not match required pattern
```
sherizada-okta, line 656: This is not the correct string, needs a param in place of the registration attr Email

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -651,6 +651,9 @@ registration.default.callbackhook.error=We could not process your registration a
 registration.error.userName.invalidEmail=Invalid email address
 registration.error.password.passwordRequirementsNotMet=Password requirements were not met
 registration.error.userName.notUniqueWithinOrg=An account with that email already exists
+registration.error.notUniqueWithinOrg = A user with this {0} already exists
```
jmelberg-okta, line 654: This will produce invalid translations, as the argument **also** needs to be translated.

We need to make sure we use the entire message (`A user with this email already exists`) as a part of multiple i18n key entries.

sherizada-okta: That makes sense @jmelberg-okta. 
But I am guessing the params object indicates that there is string interpolation. For the following example, the value in the param could be anything, not just "Email". So the fix here is to have the server respond with a more specific key?

```
 {
    "message": "Provided value for property 'Email' does not match required pattern",
    "i18n": {
      "key": "registration.error.doesNotMatchPattern",
      "params": [
        "Email"
      ]
    },
    "class": "ERROR"
  }
  ```

PR #1883, opened 2021-04-30
fix: english leak authenticator-enroll-webauthn

TianGan-okta:
## Description:
1. Fix webauthn english leak
2. remove outdated mock in test

## PR Checklist

- [] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:
image

### Reviewers:
@nikhilvenkatraman-okta @jmelberg-okta 

### Issue:

- [OKTA-389456](https://oktainc.atlassian.net/browse/OKTA-389456)


test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
```diff
@@ -37,7 +37,6 @@ const ignoredMocks = [
   'error-okta-verify-totp.json',
   'error-internal-server-error.json',
   'error-identify-access-denied.json',
-  'error-google-authenticator-otp.json',
```
TianGan-okta, line 40: This file has been removed when I was working on softtoken i18n, so removing it from the filter.

src/v2/view-builder/views/webauthn/EnrollWebauthnView.js
```diff
@@ -20,7 +20,9 @@ function getExcludeCredentials(authenticatorEnrollments = []) {
 }
 
 const Body = BaseForm.extend({
-  title: loc('oie.enroll.webauthn.title', 'login'),
+  title() {
+    return loc('oie.enroll.webauthn.title', 'login');
```
jmelberg-okta, line 24: Why does this need to be a function now?

TianGan-okta: This will fix the i18n leak issue. I think there is some timing issue if we are using it as properties

jmelberg-okta: Looks like we also do this elsewhere, as the English value is loaded in the browser by default - and returning a `function` here ensures the `loc` method gets executed a bit later.

There are quite a few occurrences where we're mapping this attribute and others (subtitle, save, etc) to a `string`, so @nikhilvenkatraman-okta what do you think about creating a lint rule for this?

jmelberg-okta (approved):

PR #1887, opened 2021-04-30
fix: Resolve i18n issues in some authenticators

jmelberg-okta:
## Description:

- Updates mock JSON files to updated API values to resolve transformation issues.
- Use existing translations for authenticator key names as opposed to using `displayName` prop. This keeps v1 parity.
- Adds `no-translate` to enroll/verify challenge views

### Screenshot/Video:

| Description | Image |
| ----------- | ------ |
| Use legacy i18n keys for authenticator names | Screen Shot 2021-04-30 at 3 33 14 PM |

### Reviewers:
- @TianGan-okta 
- @nikhilvenkatraman-okta 

### Issue:

- [OKTA-389458](https://oktainc.atlassian.net/browse/OKTA-389458)


src/v2/view-builder/components/AuthenticatorEnrollOptions.js
```diff
@@ -22,7 +22,7 @@ const AuthenticatorRow = View.extend({
           </div>\
         </div>\
         <div class="authenticator-description">\
-          <h3 class="authenticator-label">{{label}}</h3>\
+          <h3 class="authenticator-label no-translate">{{label}}</h3>\
```
jmelberg-okta, line 25: The intention here is to use the `displayName` of the authenticator the user is enrolling or challenging, or falling back to v1 customizations (see Duo and Symantec authenticators). Given that this is a DB stored value configured by the user/admin - we won't be able to guarantee this will be translated.

src/v2/ion/i18nTransformer.js
```diff
@@ -60,25 +60,29 @@ const I18N_OVERRIDE_MAPPINGS = {
 
   'identify-recovery.identifier': 'password.forgot.email.or.username.placeholder',
 
+  'select-authenticator-enroll.authenticator.duo': 'factor.duo',
```
nikhilvenkatraman-okta, line 63: Are we just re-arranging entries here ?

src/v2/ion/i18nTransformer.js
```diff
@@ -60,25 +60,29 @@ const I18N_OVERRIDE_MAPPINGS = {
...
-  'select-authenticator-authenticate.authenticator.okta_verify.signed_nonce': 'oie.okta_verify.signed_nonce.label',
   'select-authenticator-authenticate.authenticator.okta_verify.push': 'oie.okta_verify.push.title',
+  'select-authenticator-authenticate.authenticator.okta_verify.signed_nonce': 'oie.okta_verify.signed_nonce.label',
   'select-authenticator-authenticate.authenticator.okta_verify.totp': 'oie.okta_verify.totp.title',
-  'select-authenticator-authenticate.authenticator.google_otp': 'oie.google_authenticator.label',
+  'select-authenticator-authenticate.authenticator.phone_number': 'oie.phone.label',
   'select-authenticator-authenticate.authenticator.rsa_token': 'factor.totpHard.rsaSecurId',
+  'select-authenticator-authenticate.authenticator.security_question': 'oie.security.question.label',
+  'select-authenticator-authenticate.authenticator.symantec_vip': 'factor.totpHard.symantecVip',
+  'select-authenticator-authenticate.authenticator.webauthn': 'oie.webauthn.label',
```
TianGan-okta, line 85: Noticed some of label have the key for oie some of them are reusing the v1 key? Do we want to keep them consistent?

TianGan-okta (approved):

src/v2/ion/i18nTransformer.js
jmelberg-okta: Yep - keeping them alphabetic so it is easier to follow!

src/v2/ion/i18nTransformer.js
jmelberg-okta: We **don't** want to be consistent here.

If there is an existing **non-OIE** key, it should be used. Reusing the existing key allows customers who have [i18n customizations](https://github.com/okta/okta-signin-widget#language-and-text) to migrate without a breaking change and/or updating their translations.

PR #1873, opened 2021-04-26
fixes registration hook issue

pradeepdewda-okta:
## Description:

1. it caused post to send old form data instead of new, in case of registration errors

## PR Checklist

- [ ] Have you verified the basic functionality for this change?

- [ ] Added unit tests?

- [ ] Added e2e tests

- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?

### Reviewers:
@sherizada-okta 

### Issue:

- [OKTA-388105](https://oktainc.atlassian.net/browse/OKTA-388105)

PR #1896, opened 2021-05-03
Refactor Captcha to utilize execute flow

kamaldeepcheema-okta:
## Description:
- During our testing of Captcha, we discovered an issue with the reCaptcha library that causes the Captcha lib to become unresponsive after multiple failed attempts. The source of this issue is an open bug against the library itself: https://github.com/google/recaptcha/issues/269
- A workaround to prevent this issue is to use API's `execute` method rather than solely relying on `render`
- The solution has been refactored to prevent this issue.
- As an added benefit, client-side validation now occurs prior to displaying the Captcha challenge (https://oktainc.atlassian.net/browse/OKTA-391321)

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/77295104/116939095-bf728980-ac39-11eb-87e6-a58499527a47.mov

### Issue:

- [OKTA-391741](https://oktainc.atlassian.net/browse/OKTA-391741)


src/v2/view-builder/views/captcha/CaptchaView.js
```diff
@@ -23,6 +23,31 @@ const RECAPTCHAV2_URL =
...
+          data-callback="OktaSignInWidgetOnCaptchaSolved"
+          data-size="invisible">
+      </div>
+    {{/if}}\
+  `,
+
+  getTemplateData: function() {
+    const captchaConfig = this.options.appState.get('captcha');
+    let className = null;
+    if (captchaConfig) {
```
haishengwu-okta, line 40:
```
if (this.captchaConfig) {
  return { className, siteKey, isCaptchaConfigured: true};
} else {
  return {};
}
```

src/v2/view-builder/views/captcha/CaptchaView.js
```diff
@@ -74,7 +135,11 @@ export default View.extend({
 
     // Attaching the callback to the window object so that the CAPTCHA script that we dynamically render
     // can have access to it since it won't have access to this view's scope.
-    window.OktaSignInWidgetOnCaptchaLoaded = onCaptchaLoaded;
+    window.OktaSignInWidgetOnCaptchaLoaded = this.onCaptchaLoaded;
```
haishengwu-okta, line 138: do you have to bind to `this` context as you can access the function under `window` object in test?

src/v2/view-builder/views/captcha/CaptchaView.js
```diff
@@ -23,6 +23,31 @@ const RECAPTCHAV2_URL =
 export default View.extend({
   className: 'captcha-view',
 
+  template: hbs`
+    {{#if isCaptchaConfigured}}\
+      <div id ="captcha-container"
+          class="{{class}}"
+          data-sitekey="{{sitekey}}"
+          data-callback="OktaSignInWidgetOnCaptchaSolved"
```
haishengwu-okta, line 31: Could 'OktaSignInWidgetOnCaptchaSolved` be defined as variable (just because this string has been used at 2 places)?

src/v2/view-builder/views/captcha/CaptchaView.js
```diff
@@ -49,22 +74,58 @@ export default View.extend({
...
+      const captchaId = this.$el.find('#captcha-container').attr('data-captcha-id');
+      captchaObject.reset(captchaId);
 
       // Set the token in the model
       const fieldName = this.options.name;
       this.model.set(fieldName, token);
 
-      // Clear form errors before re-validation
-      this.model.trigger('clearFormError');
+      // By the time we reach this callback, the model has already unflattened the field names.
```
haishengwu-okta, line 90: Doesn't seem right.
`this.model` shall always stay flatten. `this.model.toJSON` would return unflattened data.

haishengwu-okta: It's fixed at https://github.com/okta/okta-signin-widget/pull/1873/files

haishengwu-okta (approved):

haishengwu-okta:
Accidentally approved.. but the only one concern is around unflatten data.

src/v2/view-builder/views/captcha/CaptchaView.js
kamaldeepcheema-okta: The issue that I found is that `this.model.toJSON` is invoked before this `onCaptchaLoaded` is invoked in the case where there is validation errors (because the view has to re-render itself to display client side errors). Because of that, by the time it gets to this callback, the model itself contains both the flattened and unflattened data - so it essentially has two copies of the underlying. I'm not sure if this is an intended behavior or a bug but here's a little demo of this:

https://user-images.githubusercontent.com/77295104/117023965-939de500-acc7-11eb-9757-bf778fe41722.mov

src/v2/view-builder/views/captcha/CaptchaView.js
kamaldeepcheema-okta: Yeah this is a good point! I don't need to bind to `this` because it was mostly done to facilitate unit testing.

src/v2/view-builder/views/captcha/CaptchaView.js
kamaldeepcheema-okta: Just to clarify, I'm assuming you're suggesting to use a template variable here?

src/v2/view-builder/views/captcha/CaptchaView.js
kamaldeepcheema-okta: UDPATE: The issue I demoed above is caused by the underlying problem that this PR is solving: https://github.com/okta/okta-signin-widget/pull/1873/files. Thank you @haishengwu-okta for pointing this out. All this extra logic around unflattening is not needed and I've removed it now.

src/v2/view-builder/internals/BaseForm.js
```diff
@@ -33,12 +32,9 @@ export default Form.extend({
...
 
     // Render CAPTCHA if one of the form fields requires us to.
-    this.listenTo(this.options.appState, 'onCaptchaLoaded', (captchaConfig, callback) => {
-      renderCaptcha(
-        captchaConfig,
-        this,
-        callback
-      );
+    this.listenTo(this.options.appState, 'onCaptchaLoaded', (captchaObject) => {
+      this.isCaptchaConfigured = true;
```
haishengwu-okta, line 36: This flag is redundant and can be removed.

haishengwu-okta (approved):
One comment & 🚢

PR #1898, opened 2021-05-04
chore: Fix issue with local testcafe runner and changes default reporter

jmelberg-okta:
## Description:

Fixes an issue with the `test` runner that omitted filtering for testcafe specs.

## PR Checklist

### Screenshot/Video:

```bash
[okta-signin-widget]$ yarn test -t testcafe chrome test/testcafe/spec --filter "should have correct display texts"
...
Running: testcafe chrome test/testcafe/spec  --test-grep "should have correct display texts"
The "src", "browsers", "filter" options from the configuration file will be ignored.
 Running tests in:
 - Chrome 90.0.4430.93 / macOS 10.15.7

 Identify Recovery - reset flow
 ✓ should have correct display texts

 1 passed (3s)
✨  Done in 23.25s.
```

### Reviewers:
- @pradeepdewda-okta 

### Issue:

- [OKTA-392638](https://oktainc.atlassian.net/browse/OKTA-392638)


.testcaferc.json
```diff
@@ -3,9 +3,5 @@
     "chrome:headless"
   ],
   "concurrency": 2,
```
pradeepdewda-okta, line 5: change it to `1`, otherwise it opens two browsers locally and also default to `chrome` instead of headless ?

test/testcafe/README.md
```diff
@@ -15,19 +15,19 @@ Run All in headless chrome
...
 
 Run all similar Fixtures that matches the pattern
 
-- `yarn test -t testcafe chrome -F "Unknown.*"`
+- `yarn test -t testcafe chrome --filter "Unknown.*"`
 
 Run a single test
 
-- `yarn test -t testcafe chrome -t "should have editable fields"`
+- `yarn test -t testcafe chrome --filter "should have editable fields"`
```
pradeepdewda-okta, line 26: how would it identify between -F, -f, -t, -T, --live, --debug-mode ?

.testcaferc.json
jmelberg-okta: You'll need to coordinate this change with DevEx's test team. This parallelism was added in #1188, so I'd prefer to leave this unchanged.

test/testcafe/README.md
jmelberg-okta: Right now it's filtering basted on the test name, not the fixture. This should account for most cases and we can address the edge cases when they come up.

In regards to other flags like `--live` and `--debug-mode`, those get captured via the [`additionalArgs` logic](https://github.com/okta/okta-signin-widget/blob/744d2997a0a9a51eeb41f9e140c6001b05ce9844/scripts/buildtools/commands/test.js#L69), and will continue to work.

.testcaferc.json
pradeepdewda-okta: That change didn't force how developers wanted to run testcafe through command line.. It has it's own command to run.. But here we are forcing every developer to change `testcaferc.js`, I think `testcaferc.js` should be optional to use when dev wants to run command line, NOT the default.

test/testcafe/README.md
pradeepdewda-okta: With this change we are loosing flexibility and usefulness of testcafe command line.. We should not enforce use of `testcaferc.json` config on the developer, instead it should be optional..

test/testcafe/README.md
```diff
@@ -23,7 +23,7 @@ Run all similar Fixtures that matches the pattern
 
 Run a single test
 
-- `yarn test -t testcafe chrome -t "should have editable fields"`
+- `yarn test -t testcafe chrome --test "should have editable fields"`
```
pradeepdewda-okta, line 26: We need to update Readme

1. It's not required to pass `chrome`, because it can default to what we have in rc file.
2. but we can have a single usage example on how to change the browser
   `yarn test -t testcafe <browser> --test "should have editable fields"`

pradeepdewda-okta (approved):
🥇  for prompt resolution..

I tested locally with all our existing parameters and `--debug-mode`. Now existing behavior seems to be intact.

PR #1906, opened 2021-05-05
phone enroll engleaks fix

anipendakur-okta:
## Description:
* "SMS" and "Voice call" radio options are translated on method selection screen during enroll now. A note -"+1" is not translated still hence the mock can't be removed from EngLeaks spec. - 
image

* "Enter code" is translated on enroll screen for SMS now - 
image
* "Enter code" is translated on enroll screen for Voice now -
image

## PR Checklist

- [X] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-389449](https://oktainc.atlassian.net/browse/OKTA-389449)


test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
```diff
@@ -44,8 +44,6 @@ const ignoredMocks = [
   'authenticator-expired-password.json',
   'authenticator-expired-password-with-enrollment-authenticator.json',
   'authenticator-expired-password-no-complexity.json',
-  'authenticator-enroll-phone.json',
-  'authenticator-enroll-phone-voice.json',
   'authenticator-enroll-email.json',
   'authenticator-enroll-data-phone.json',
   'authenticator-enroll-data-phone-voice.json',
```
anipendakur-okta, line 49: These cannot be removed yet because we still have `+1` country code that's not translated. Please correct me otherwise, it doesn't make sense to translate the country code too. We'll have to re-implement how contrycode is retrieved if we intend to do this.

src/v2/ion/i18nTransformer.js
```diff
@@ -93,9 +93,11 @@ const I18N_OVERRIDE_MAPPINGS = {
   'authenticator-verification-data.okta_verify.authenticator.methodType.totp': 'oie.okta_verify.totp.title',
 
   'authenticator-enrollment-data.phone_number.authenticator.phoneNumber': 'mfa.phoneNumber.placeholder',
+  'authenticator-enrollment-data.phone_number.authenticator.methodType.sms': 'oie.phone.enroll.sms.label',
+  'authenticator-enrollment-data.phone_number.authenticator.methodType.voice': 'oie.phone.enroll.voice.label',
 
   'enroll-authenticator.okta_password.credentials.passcode': 'oie.password.passwordLabel',
-  'enroll-authenticator.phone.credentials.passcode': 'mfa.challenge.enterCode.placeholder',
+  'enroll-authenticator.phone_number.credentials.passcode': 'mfa.challenge.enterCode.placeholder',
```
anipendakur-okta, line 100: @pradeepdewda-okta - I was wondering if you can take a look at this change. The previous key `enroll-authenticator.phone.credentials.passcode` didn't make sense as there's no `phone` named authenticator (it seems to be `phone_number` instead). I have also added a test for this.

test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
nikhilvenkatraman-okta: yeah, I think that makes sense, but let's just check with PM once on that and try to ignore that section from the test suite by adding a class of `no-translate` to the html element. Refer to the [guideline](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2021917320/siw-localization-bug-fixing+guideline) for details.

nikhilvenkatraman-okta (approved):

src/v2/ion/i18nTransformer.js
pradeepdewda-okta: @anipendakur-okta this looks good to me..

pradeepdewda-okta (approved):

PR #1917, opened 2021-05-06
Change email verification sent text to not specify email address

ethanhsiung-okta:
Resolves: OKTA-369683

## Description:

- Update email verification sent text to not specify the email/login of the user
- The original problem was that the text did not specify that the email was sent to both primary and secondary emails. Instead it was only mentioning the primary email address
- Slack thread: https://okta.slack.com/archives/C9AJXB5GC/p1615928217016900?thread_ts=1615923255.009200&cid=C9AJXB5GC

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Before:
image
After:
Screen Shot 2021-05-05 at 4 23 32 PM

### Reviewers:
@anipendakur-okta @rameshbhojan-okta 

### Issue:

- [OKTA-369683](https://oktainc.atlassian.net/browse/OKTA-369683)


src/v2/view-builder/views/email/ChallengeAuthenticatorEmailView.js
```diff
@@ -11,11 +11,7 @@ const CheckYourEmailTitle = View.extend({
     'data-se': 'o-form-explain',
   },
   template: hbs`
-    {{#if email}}
-      {{i18n code="oie.email.verify.sentText" bundle="login" arguments="email" $1="<span class='strong'>$1</span>"}}
-    {{else}}
-      {{i18n code="oie.email.verify.alternate.sentText" bundle="login"}}
-    {{/if}}
+    {{i18n code="oie.email.verify.alternate.sentText" bundle="login"}}
```
haoyuli-okta, line 14: could we handle this with a secondary email instead of sending a general message?
the second option is used for user enumeration, which does not show you anything on purpose.
Can someone keep me honest here, we want to have the option for you to choose which emails to use to recover ?

test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
```diff
@@ -136,7 +136,7 @@ test
     await t.expect(saveBtnText).contains('Verify');
     await t.expect(pageTitle).contains('Verify with your email');
     await t.expect(challengeEmailPageObject.getFormSubtitle())
-      .contains('Check i**a@h***o.net for a verification message. Click the verification button in your email or enter the code below to continue.');
+      .contains('Check your email for a verification message. Click the verification button in your email or enter the code below to continue.');
```
pradeepdewda-okta, line 139: we can remove these two tests that might not make sense anymore since the message is same all the time.. ?
`challenge email authenticator screen has right labels when profile is null`
`challenge email authenticator screen has right labels when profile is empty`

src/v2/view-builder/views/email/ChallengeAuthenticatorEmailView.js
ethanhsiung-okta: I believe the verification email is always sent to both email addresses when possible. 
You can look at this slack thread https://okta.slack.com/archives/C9AJXB5GC/p1615928217016900?thread_ts=1615923255.009200&cid=C9AJXB5GC for why we picked this option

test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
ethanhsiung-okta: I left them in on purpose because I figured it couldn't hurt to cover a few more cases. If you think they're unnecessary, I can remove them

pradeepdewda-okta (approved):

haoyuli-okta (approved):

test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
rameshbhojan-okta: Let's remove one of them, if the test(s) are redundant.

rameshbhojan-okta (approved):

test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
ethanhsiung-okta: The three test scenarios now are:
1. Profile is provided
2. Profile is null
3. Profile is empty
I'd say 2 and 3 are different enough to keep both

anipendakur-okta:
I was wondering if this fix is a temporary fix and if there's a plan to change the behavior to send the email to either primary / secondary?

src/v2/view-builder/views/email/ChallengeAuthenticatorEmailView.js
```diff
@@ -11,11 +11,7 @@ const CheckYourEmailTitle = View.extend({
     'data-se': 'o-form-explain',
   },
   template: hbs`
-    {{#if email}}
-      {{i18n code="oie.email.verify.sentText" bundle="login" arguments="email" $1="<span class='strong'>$1</span>"}}
```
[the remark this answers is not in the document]
ethanhsiung-okta: Sure, I'll remove the key. 
Also, good point on the selenium tests as well. I'll do a bacon run with these changes and see what breaks

ethanhsiung-okta:
> I was wondering if this fix is a temporary fix and if there's a plan to change the behavior to send the email to either primary or secondary email later on?

@anipendakur-okta The email is sent to both primary and secondary addresses currently. As far as I know, there's no plan to change that, and this would not be a temporary fix

src/v2/view-builder/views/email/ChallengeAuthenticatorEmailView.js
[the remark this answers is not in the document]
anipendakur-okta: You'll need to run bacon on the downstream artifact for this change.. just fyi.. :)

PR #1918, opened 2021-05-06
When calling /challenge, if it timeout, the xhr.statusText === 'timeo…

sshen-okta:
## Description:
When calling /challenge, if it timeout, the xhr.statusText === 'timeout'.

When it is timeout, we should not cancel the polling.

some context about this fix:
We implemented the following probing logic:

SIW iterates the loopback ports with the following two calls

/probe - detect if the port is available

/challenge - send device challenge to OV

If both calls above for a particular port get success, we stop probing, start /poll to the server.

until we iterate all the loopback ports. 

If non of the port get success, we "cancel" the polling.

When a call to (/probe, /challenge) is successful, the xhr.status == 200.

timeout actually means fail (xhr.status==0)

In the case of /challenge, we have 3000 ms timeout, but for inline enrollment, OV will reply the /challenge after enrollment completes by the user. 

Usually it takes a few minutes. As a result, SIW will end up the following situation.

"non of the port get success, we "cancel" the polling."

The fix is that we should not treat timeout of /challenge as failure so that SIW will continue to poll the result until either user completes the enrollment or IdX transaction timeout(5 mintues).

Resolves: OKTA-384095

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
before fix, failure video:
https://okta.box.com/s/z16eqtkt78gigmfwmc77loumn3m1bay4

after fix, success video:
https://okta.box.com/s/lq3qxwyj26p60uv44m69ukupcc6sazko

### Reviewers:
@anatoliiboiarshchykov-okta @haishengwu-okta @magizh-okta 

### Issue:

- [OKTA-384095](https://oktainc.atlassian.net/browse/OKTA-384095)

### DIFF
--- a/src/v2/view-builder/views/device/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/device/DeviceChallengePollView.js
@@ -113,7 +113,12 @@ const Body = BaseFormWithPolling.extend(
               return this.trigger('save', this.model);
             }).fail((xhr) => {
               Logger.error(`OV challenge response with HTTP code ${xhr.status} ${xhr.responseText}`);
-              onPortFail();
+              if (xhr.statusText === 'timeout') {
+                foundPort = true;
+                return this.trigger('save', this.model);
+              } else {
+                onPortFail();
+              }
             });
           })
           .fail(onFailure);
--- a/test/testcafe/spec/DeviceChallengePollView_spec.js
+++ b/test/testcafe/spec/DeviceChallengePollView_spec.js
@@ -55,6 +55,33 @@ const loopbackSuccesskMock = RequestMock()
     'access-control-allow-methods': 'POST, OPTIONS'
   });
 
+const loopbackSuccessForChallengeTimeoutLogger = RequestLogger(/introspect|probe|challenge|poll|cancel/, { logRequestBody: true, stringifyRequestBody: true });
+const loopbackSuccessForChallengeTimeoutMock = RequestMock()
+  .onRequestTo(/\/idp\/idx\/introspect/)
+  .respond(identifyWithDeviceProbingLoopback)
+  .onRequestTo(/\/idp\/idx\/authenticators\/poll/)
+  .respond(identifyWithDeviceProbingLoopback)
+  .onRequestTo(/2000\/probe/)
+  .respond(null, 500, {
+    'access-control-allow-origin': '*',
+    'access-control-allow-headers': 'X-Okta-Xsrftoken, Content-Type'
+  })
+  .onRequestTo(/6511\/probe/)
+  .respond(null, 200, {
+    'access-control-allow-origin': '*',
+    'access-control-allow-headers': 'X-Okta-Xsrftoken, Content-Type'
+  })
+  .onRequestTo(/6511\/challenge/)
+  .respond(async (req, res) => {
+    await new Promise((r) => setTimeout(r, 3100));
+    res.statusCode = req.method !== 'POST' ? 204 : 200;
+    res.headers = {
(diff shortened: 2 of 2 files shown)


sshen-okta:
before fix, failure video:
https://user-images.githubusercontent.com/5872429/117342687-5ae14580-ae58-11eb-80c0-d8fc95e2e0ff.mov

after fix, success video:

https://user-images.githubusercontent.com/5872429/117342782-751b2380-ae58-11eb-9e56-9af8a8ccc846.mov

sshen-okta:
before fix, failure video:
https://user-images.githubusercontent.com/5872429/117342687-5ae14580-ae58-11eb-80c0-d8fc95e2e0ff.mov

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -204,6 +231,36 @@ test
     await t.expect(identityPage.getIdentifierValue()).eql('Test Identifier');
   });
 
+test
+  .requestHooks(loopbackSuccessForChallengeTimeoutLogger, loopbackSuccessForChallengeTimeoutMock)('in loopback server approach, with pool but not cancel when challenge timeout', async t => {
```
magizh-okta, line 235: nit: will poll*

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -55,6 +55,33 @@ const loopbackSuccesskMock = RequestMock()
...
+    'access-control-allow-headers': 'X-Okta-Xsrftoken, Content-Type'
+  })
+  .onRequestTo(/6511\/probe/)
+  .respond(null, 200, {
+    'access-control-allow-origin': '*',
+    'access-control-allow-headers': 'X-Okta-Xsrftoken, Content-Type'
+  })
+  .onRequestTo(/6511\/challenge/)
+  .respond(async (req, res) => {
+    await new Promise((r) => setTimeout(r, 10000));
```
magizh-okta, line 76: can we have a shorter timeout 3100 ?

magizh-okta (approved):
minor nits after that 🚢

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -113,7 +113,12 @@ const Body = BaseFormWithPolling.extend(
               return this.trigger('save', this.model);
             }).fail((xhr) => {
               Logger.error(`OV challenge response with HTTP code ${xhr.status} ${xhr.responseText}`);
-              onPortFail();
+              if (xhr.statusText === 'timeout') {
```
pradeepdewda-okta, line 116: create a constant
CHALLENGE_TIMEOUT = 'timeout'

sshen-okta: Hey @pradeepdewda-okta I feel we might not want to make a const for this since there is only one reference of 'timeout' in this file. Ideally, we might want to have a constants collection object somewhere at framework/util level to capture all keyword in JavaScript. for example,
'object', 'image', 'function', 'title' etc. today, I saw we use these string (not const) everywhere.
but it is out of the scope of this issue.

test/testcafe/spec/DeviceChallengePollView_spec.js
sshen-okta: Looks like our test execution is faster (less than 3s), the test is teared down after completion. the long timeout here will not have impact. but I'm going to change it anyway.

PR #1920, opened 2021-05-06
Remove Verify with something else link from the Device poll page

yutongwang-okta:
## Description:
Remove "Verify with something else" link from the Device poll page. In the Device poll page, the user has not been detected, so the link leads back to login page. But we should keep the link in the switch authenticator page, where the user is already detected.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

1. Before the change: the "Verify with something else" link shows up in the Device poll page
Screen Shot 2021-05-06 at 4 48 31 PM

2. After the change: link is removed
Screen Shot 2021-05-06 at 5 05 19 PM

3. We still need to keep the "Verify with something else" in the fastpass assurance page
Screen Shot 2021-05-10 at 12 19 59 PM

### Issue:

- [OKTA-387045](https://oktainc.atlassian.net/browse/OKTA-387045)

### DIFF
--- a/src/v2/view-builder/views/device/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/device/DeviceChallengePollView.js
@@ -156,14 +156,6 @@ const Footer = BaseFooter.extend({
       Enums.UNIVERSAL_LINK_CHALLENGE
     ].includes(this.options.currentViewState.relatesTo.value.challengeMethod);
     if (isFallbackApproach) {
-      this.links = [
-        {
-          name: 'sign-in-options',
-          type: 'link',
-          label: loc('oie.verification.switch.authenticator', 'login'),
-          href: this.settings.get('baseUrl')
-        }
-      ];
       BaseFooter.prototype.initialize.apply(this, arguments);
     } else {
       this.add(Link, {
--- a/test/testcafe/spec/DeviceChallengePollView_spec.js
+++ b/test/testcafe/spec/DeviceChallengePollView_spec.js
@@ -231,7 +231,6 @@ test
     await t.expect(content).contains('Don’t have Okta Verify?');
     await t.expect(content).contains('Download here');
     await t.expect(deviceChallengePollPageObject.getDownloadOktaVerifyLink()).eql('https://apps.apple.com/us/app/okta-verify/id490179405');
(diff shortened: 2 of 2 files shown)


jane-kuang-okta:
@yutongwang-okta could you show a "Before" image of the Custom Uri screen? And also specify the "After" picture so it's easier to compare. Some questions:

1. In the video provided in Jira, the link to go back to the sign in screen says "Back to sign in" but your image says "Cancel and take me to sign in." Which should it be?
2. The Jira says to remove the link when a user comes back from the IdP, but this change seems to impact all Custom Uri/UL screens. Should this UI change be for all Custom Uri/UL screens?

yutongwang-okta:
@jane-kuang-okta Good point. I have updated the section with new screenshots. The previous ones are a bit old, and the text should be "Back to sign in".  @sshen-okta reported the problem in the IDP case, but the issue is actually for all Custom URI/UL cases. During that phase, the OV does not have user information, thus switching authenticator will lead user back to login.

santoshmale-okta:
There seem to be two places where we show the Custom URI screen. Once in the `Identify` Step which is where we seem to use `DeviceChallengePollView`. The second place is during the `Assurance` step where Okta Verify shows up as a factor in the factor list. We seem to use a different view there `ChallengeOktaVerifyFastPassView`. Can we please see whats the current behavior for this screen in the Assurance Step? 
Probably it makes sense to have the `Verify with Something else` in the Assurance step as the user might have already verified one factor by then and might just be wanting to go back and select diff factor rather than `Back to Sign In`. Please verify.

sshen-okta:
> There seem to be two places where we show the Custom URI screen. Once in the `Identify` Step which is where we seem to use `DeviceChallengePollView`. The second place is during the `Assurance` step where Okta Verify shows up as a factor in the factor list. We seem to use a different view there `ChallengeOktaVerifyFastPassView`. Can we please see whats the current behavior for this screen in the Assurance Step?
> Probably it makes sense to have the `Verify with Something else` in the Assurance step as the user might have already verified one factor by then and might just be wanting to go back and select diff factor rather than `Back to Sign In`. Please verify.

can you paster the screen for the page in the "Assurance" step?

nbories-okta (approved):
looks good visually. Thanks 🍶

PR #1929, opened 2021-05-07
security question error mocks engleaks

anipendakur-okta:
## Description:
The error string on length is translated now - 
image
image

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-389448](https://oktainc.atlassian.net/browse/OKTA-389448)


playground/mocks/data/idp/idx/error-authenticator-enroll-security-question-create-question.json
```diff
@@ -1,65 +1,133 @@
...
-                              "class": "ERROR",
-                              "i18n": {
-                                "key": "securityQuestion.answer.tooShort",
-                                "params": []
+                              "message":"The security question answer must be at least 4 characters in length",
+                              "i18n":{
+                                "key":"securityQuestion.answer.tooShort.arg",
+                                "params":[
+                                  4
+                                ]
```
anipendakur-okta, line 128: It seems that there is no more `securityQuestion.answer.tooShort` and instead `securityQuestion.answer.tooShort.arg` with params coming from backend. I have discussed in detail with @rameshbhojan-okta .

`securityQuestion.answer.tooShort` seems to be a V1 thing now.

Hence updated this mock and the next one as well.

rameshbhojan-okta: I don't remember any recent changes w.r.t this and AFAIK this was always the case - not sure why mocks here in SIW were different all this while. 

`securityQuestion.answer.tooShort` was used only in the property level validation from MFA Question in V1. 

> It seems that there is no more securityQuestion.answer.tooShort and instead securityQuestion.answer.tooShort.arg with params coming from backend. I have discussed in detail with @rameshbhojan-okta .
> 
> securityQuestion.answer.tooShort seems to be a V1 thing now.

nikhilvenkatraman-okta (approved):
Nice to see the mocks getting updated to match the APIs

PR #1938, opened 2021-05-10
Remove Verify with something else link from the Device poll page

yutongwang-okta:
This is the recreation of https://github.com/okta/okta-signin-widget/pull/1920 which was branched off master.

## Description:

Remove "Verify with something else" link from the Device poll page. In the Device poll page, the user has not been detected, so the link leads back to login page. But we should keep the link in the switch authenticator page, where the user is already detected.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

1. Before the change: the "Verify with something else" link shows up in the identify page
Screen Shot 2021-05-06 at 4 48 31 PM

2. After the change: link is removed
Screen Shot 2021-05-06 at 5 05 19 PM

3. We still need to keep the "Verify with something else" in the FastPass page during assurance stage
Screen Shot 2021-05-10 at 12 19 59 PM

4. We still need to keep the "Verify with something else" in other factor page during assurance stage
Screen Shot 2021-05-06 at 5 06 27 PM

### Issue:

- OKTA-387045

### DIFF
--- a/src/v2/view-builder/views/device/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/device/DeviceChallengePollView.js
@@ -161,14 +161,6 @@ const Footer = BaseFooter.extend({
       Enums.UNIVERSAL_LINK_CHALLENGE
     ].includes(this.options.currentViewState.relatesTo.value.challengeMethod);
     if (isFallbackApproach) {
-      this.links = [
-        {
-          name: 'sign-in-options',
-          type: 'link',
-          label: loc('oie.verification.switch.authenticator', 'login'),
-          href: this.settings.get('baseUrl')
-        }
-      ];
       BaseFooter.prototype.initialize.apply(this, arguments);
     } else {
       this.add(Link, {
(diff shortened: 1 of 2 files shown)


howardlee-okta (approved):

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -161,14 +161,6 @@ const Footer = BaseFooter.extend({
       Enums.UNIVERSAL_LINK_CHALLENGE
     ].includes(this.options.currentViewState.relatesTo.value.challengeMethod);
     if (isFallbackApproach) {
-      this.links = [
```
haishengwu-okta, line 164: Isn't really related to your changes but we shall try to override `links` method instead of `initialize`.
Check the `BaseFooter.js`

yutongwang-okta: Thanks @howardlee-okta and @haishengwu-okta

PR #1947, opened 2021-05-12
enleaks fix for security.access_denied server side error

anipendakur-okta:
## Description:
image

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-389253](https://oktainc.atlassian.net/browse/OKTA-389253)


nikhilvenkatraman-okta (approved):

playground/mocks/data/idp/idx/error-email-verify.json
```diff
@@ -1,18 +1,17 @@
...
+  "messages":{
+    "type":"array",
+    "value":[
       {
-        "message": "Authentication failed",
-        "i18n": {
-          "key": "security.access_denied"
+        "message":"You do not have permission to perform the requested action.",
+        "i18n":{
+          "key":"security.access_denied"
```
jmelberg-okta, line 12: Can we ask the server-side to update this key with our reusable one? This message is already used for the [`E0000006`](https://github.com/okta/okta-signin-widget/blob/b9a0cca776bcd2f3d2e4ee14e2e1baeb0de55aaf/packages/%40okta/i18n/src/properties/login.properties#L69) code.

As a result, customer overrides for this error will continue to apply when OIE is enabled.

anipendakur-okta: At a glance, it looks like a lot of re-work and many teams will have to be involved in achieving it. I just searched for the key in `okta-core` and found many occurrences of it. Perhaps this might be an exception/compromise ?

anipendakur-okta: Note: E0000006 - `You do not have permission to perform the requested action`
vs security.access_denied - `You do not have permission to perform the requested action.` (the trailing `.`). I will be doing this change, but since the strings are different, several selenium tests are expected to fail in the monolith.

anipendakur-okta: Spoke with @jmelberg-okta and we decided to do this a client-side mapping in i18nTransformer instead of server-side changes.

jmelberg-okta (approved):

PR #1961, opened 2021-05-17
fix: OV Totp UV error on verify

jasleennuno-okta:
## Description:
- OV Totp UV error on verify

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Screen Shot 2021-05-17 at 1 57 44 PM

### Reviewers:
- @mauriciocastillosilva-okta @stevennguyen-okta @kapilduraphe-okta 

### Issue:

- [OKTA-394842](https://oktainc.atlassian.net/browse/OKTA-394842)


stevennguyen-okta (changes requested):
Could you add some spacing beneath the error so that it is evenly spaced between the title and the rest of the UI? It is currently a bit close to the form field

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -833,6 +833,12 @@ oie.authenticator.app.method.push.verify.enable.biometrics.point2 = Okta Verify
 oie.authenticator.app.method.push.verify.enable.biometrics.point3 = In Okta Verify, biometrics are enabled for your account
 oie.authenticator.app.method.push.enroll.enable.biometrics = Your organization requires biometrics. To proceed, ensure your device supports biometrics, then add your account and enable biometrics when prompted.
 oie.authenticator.app.method.push.enroll.enable.biometrics.title = Enable biometrics to add an account in Okta Verify
+oie.authenticator.oktaverify.method.totp.verify.enable.biometrics = Your response was received, but your organization requires biometrics. Make sure your device supports biometrics, Okta Verify is up-to-date and biometrics are enabled for your account in Okta Verify, then try again.
```
pradeepdewda-okta, line 836: used `Your` 4 times in one message. Seems overused.

stevennguyen-okta: Feedback noted, but I think if we shifted the framing of the user requirements to reduce the amount of direct language (you, your), I am afraid that the messages will be potentially less concise or unclear

stevennguyen-okta (approved):
new spacing looks good to me

kapilduraphe-okta:
key and message look good

assets/sass/v2/_okta-verify.scss
```diff
@@ -49,6 +49,8 @@
 }
 
 .okta-verify-uv-callout-content {
+  margin-bottom: 15px;
```
pradeepdewda-okta, line 52: since we are adding a new css.. can we use `em` instead of pixels

jasleennuno-okta: everything else in this file uses 15px so i wanted to keep that consistent

mauriciocastillosilva-okta (approved):
looks good

src/v2/controllers/FormController.js
```diff
@@ -248,8 +250,24 @@ export default Controller.extend({
 
     if (IonResponseHelper.isIonErrorResponse(error)) {
       const convertedErrors = IonResponseHelper.convertFormErrors(error);
-      const showBanner = convertedErrors.responseJSON.errorCauses.length ? false : true;
-      model.trigger('error', model, convertedErrors, showBanner);
+      if (IonResponseHelper.containsMessageWithI18nKey(convertedErrors, OV_UV_ENABLE_BIOMETRIC_SERVER_KEY)) {
```
mauriciocastillosilva-okta, line 253: I don't think this belongs to the FormController, this seems to be specific to OktaVerify. What I can see from the last change is that we moved the code to show a callout from `showMessages` in ChallengeOktaVerifyTotpView.js to here, but I think it was fine in the OktaVerify View... 

`showMessages` is called when we re-render the view, which happens in `FormController.handleFormNameChange`, which maybe does not happen for TOTP. So what I think we can do as well is add a listener to the model in ChallengeOktaVerifyTotpView.js, something like
```
    initialize() {
      this.listenTo(this.model, 'error', () => {
        showMessages();
      });
    },
```
Did you try that?

jasleennuno-okta: yea, i agree this isn't the best place for it
i did try doing what you suggested above and it didn't work - i'm not sure if maybe at the point we are in this method we are too far removed from the view?

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
```diff
@@ -13,8 +15,40 @@ const Body = BaseForm.extend(Object.assign(
     save() {
       return loc('mfa.challenge.verify', 'login');
     },
+
+    initialize: function() {
+      BaseForm.prototype.initialize.apply(this, arguments);
+      this.listenTo(this.options.appState, 'afterError', (convertedErrors) => {
+        // Need to wait until BaseForm renders the error so we can replace it
+        setTimeout(() => {
```
pradeepdewda-okta, line 23: can you try to avoid `setTimeout` here ?

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
```diff
@@ -13,8 +15,40 @@ const Body = BaseForm.extend(Object.assign(
...
+          this._showUserVerifcationError(convertedErrors);
+        });
+      });
+    },
+
+    _showUserVerifcationError(convertedErrors) {
```
pradeepdewda-okta, line 29: did you try overriding `showMessages` instead of creating a new fn `_showUserVerifcationError` here ? Since you want to display this enhanced callout based on a key `OV_UV_ENABLE_BIOMETRIC_SERVER_KEY`. It just make sense to override `showMessages`

pradeepdewda-okta (changes requested):
try overriding showMessages instead of creating a new fn

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
mauriciocastillosilva-okta: any suggestions? I need to form to go its regular flow and render the error message, so I can replace it. I have no access to the exact part where the BaseForm renders that message, so this seems to me a good alternative

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
mauriciocastillosilva-okta: if you look at this PR there were attempts to show this error in other ways. `showMessages` specifically does not work for this use case, because `showMessages` is only called in initialize for a BaseForm (so it works when we change from a remediation form to another). In this case we are not changing forms, we are getting an error from the API and we just need to replace the format of that error we got to show it more user friendly, that is why I need to replace it this way

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
pradeepdewda-okta: for see how `IdentifierView.js` overrides `showMessages()`

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
magizh-okta: I guess we do the same for push and even there we dont rerender the form after we get back error messages I guess. 
https://github.com/okta/okta-signin-widget/blob/f53b68168988903b0d63d4c584f68ef5693a03ee/src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js#L20

mauriciocastillosilva-okta: Push is different, we are polling, and the poll requests are returned with http 200 status. When the poll response comes with `messages` object, it also comes with a different remediation form name. This is why `showMessages` works there, sine we move to a different form

For TOTP, there is no polling. We make the request with the code and the API returns a 400 indicating it was wrong. At this time, BaseForm renders the error messages, and what I am trying to do here is replace that error messages with a different one (better format for our UI)

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
```diff
@@ -13,6 +15,38 @@ const Body = BaseForm.extend(Object.assign(
     save() {
       return loc('mfa.challenge.verify', 'login');
     },
+
+    initialize: function() {
+      BaseForm.prototype.initialize.apply(this, arguments);
+      this.listenTo(this.options.appState, 'afterError', (convertedErrors) => {
```
magizh-okta, line 21: Isnt afterError  triggered only after form is rendered and then an error occurs on submit.

magizh-okta (commented):
Looks good just one question.

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
pradeepdewda-okta: `showFormErrors` trigger `error` that trigger `afterError`.. It is on submit but doesn't do re-render

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
```diff
@@ -13,6 +15,38 @@ const Body = BaseForm.extend(Object.assign(
...
+        this.model.trigger('clearFormError');
+        // add custom error for User Verification
+        this.add('<div class="ion-messages-container"></div>', '.o-form-error-container');
+        const options = {
+          type: 'error',
+          className: 'okta-verify-uv-callout-content',
+          title: loc('oie.authenticator.app.method.push.verify.enable.biometrics.title', 'login'),
+          subtitle: loc('oie.authenticator.app.method.push.verify.enable.biometrics.description', 'login'),
+          bullets: [
+            loc('oie.authenticator.app.method.push.verify.enable.biometrics.point1', 'login'),
```
pradeepdewda-okta, line 42: I think we should avoid hardcoding these error messages. Widget should strictly respect what contains in API response, that was the purpose of `ION` framework.. If API is not sending this part of error message then we should ask backend team to come up with some solution so that it would be easy for widget to follow API structure and show this enhance error messaging.. 

It could be done this way
```
"messages": {
    "type": "array",
    "value": [
      {
        "message": "Your response was received, but your organization requires biometrics. Make sure your device supports biometrics, Okta Verify is up-to-date and biometrics are enabled for your account in Okta Verify, then try again.",
        "i18n": {
          "key": "oie.authenticator.oktaverify.method.totp.verify.enable.biometrics",
          "params": []
        },
        "class": "ERROR",
        "bullets": ['bullet1", "bullet2", "bullet3"]
      }
    ]
  }
 ```
IDX is doing a cleanup for `ResendView` that hardcodes all of warning messages.. remediation should send those messages.. https://oktainc.atlassian.net/browse/OKTA-408316

On similar note this can be done as well..

File: src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js (hunk 6 above)
Comment (magizh-okta) [reply src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js:21]: what I was getting at was that we might not need the timeout here since the afterError would be triggered only on error during form submit at which the form is already rendered.

File: src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js (hunk 7 above)
Comment (mauriciocastillosilva-okta) [reply src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js]: we already have this pattern in another [OV view](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js#L37). I agree we should stick with what the API returns, but this is also harder since what we are doing here is cosmetic changes to show a better error message (improving ux), and for a pure API client that is really not needed.
Probably we should talk to idx to see how we can improve the error message object that is returned. We can do a follow up for that, and if/when something changes, we can come back and adjust this code

File: src/v2/view-builder/internals/BaseForm.js [hunk 8]
```diff
@@ -100,6 +100,23 @@ export default Form.extend({
...
+   */
+  isErrorMessageCustomized() {
+    return false;
+  },
+
+  /*
+   * Renders an error message
+   * This should be overriden if isErrorMessageCustomized() returns true
+   */
+  showCustomErrorMessage() {
```
magizh-okta, line 116: not should we name this  showCustomFormErrorMessage just to clearly differentiate form showMessages. 

Also add a comment before showMessages " this function is called only on initialize and displays messages only on view load and not after that.

magizh-okta (commented):
@mauriciocastillosilva-okta  I like this approach. I am not sure if you had any other use cases in mind. Using afterError and cleaning up existing error and displaying a customized error also works right. We are anyway firing the afterError event for a reason.

magizh-okta (approved):
looks like we can use afterError here and we now have a cleaner approach. After adding comments 🚢

src/v2/controllers/FormController.js
```diff
@@ -143,7 +143,7 @@ export default Controller.extend({
       idx.proceed(actionPath, {})
         .then(this.handleIdxSuccess.bind(this))
         .catch(error => {
-          this.showFormErrors(this.formView.model, error);
+          this.showFormErrors(this.formView.model, error, this.formView.form);
```
pradeepdewda-okta, line 146: we can save an extra argument if we just pass `this.formView` as first argument.. and extract `model` in the function

src/v2/controllers/FormController.js
```diff
@@ -251,7 +251,11 @@ export default Controller.extend({
     } else if (error.errorSummary) {
       errorObj = { responseJSON: error };
     }
-    model.trigger('error', model, errorObj || { responseJSON: { errorSummary: String(error) } }, true);
+    const showErrorBanner = !form?.isErrorMessageCustomized(errorObj);
```
pradeepdewda-okta, line 254: we should rename this to `isErrorBannerCalloutCustomized` to be more precise on what it does. So that dev doesn't get confused with customizing error message strings..

src/v2/controllers/FormController.js
```diff
@@ -251,7 +251,11 @@ export default Controller.extend({
...
+    model.trigger('error', model, errorObj || { responseJSON: { errorSummary: String(error) } }, showErrorBanner);
+    if (!showErrorBanner) {
+      form.showCustomErrorMessage(errorObj);
```
pradeepdewda-okta, line 257: should rename to `showCustomErrorCallout`  or `showCustomErrorBanner` ?

pradeepdewda-okta (commented):
@mauriciocastillosilva-okta 🔥  looks neat.. 

few nits then :shipit:

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
pradeepdewda-okta: @mauriciocastillosilva-okta thanks Mau.. can we create a jira for this and assign to appropriate team so that we don't miss this cleanup.

pradeepdewda-okta (approved):

src/v2/controllers/FormController.js
mauriciocastillosilva-okta: `handleSaveForm` line 166 receives the `model` as an argument, so I don't want to change that. It is probably ok, but I don't want to introduce any risk in case anyone calling that method (which is called when we trigger 'saveForm` event on the appState) needs to pass a different model

pradeepdewda-okta: `showFormErrors` only used in `FormController`..  your choice though. 👍

src/v2/controllers/FormController.js
mauriciocastillosilva-okta: renamed to `isErrorCalloutCustomized`

src/v2/controllers/FormController.js
mauriciocastillosilva-okta: renamed to `showCustomErrorCallout`

jasleennuno-okta (commented):
i can't approve since this is my PR 😂 but this LGTM! looks a lot cleaner
thanks, mau!

PR #1962, opened 2021-05-17
Change label to Email as per API labels

pradeepdewda-okta:
## Description:

1. from Primary Email to Email
2. removed other instances of `Primary Email` references from mocks that can cause confusion

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Screen Shot 2021-05-17 at 3 25 41 PM

### Issue:

- [OKTA-394156](https://oktainc.atlassian.net/browse/OKTA-394156)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -963,7 +963,7 @@ oie.registration.form.title = Sign up
 oie.registration.form.submit = Sign Up
 oie.user.profile.lastname= Last name
 oie.user.profile.firstname= First name
-oie.user.profile.primary.email= Primary email
+oie.user.profile.primary.email= Email
 oie.selfservice.unlock_user.success.message = You can log in using your existing username and password.
 oie.selfservice.unlock_user.failed.message = We are unable to unlock your account at this time, please contact your administrator
 
--- a/playground/mocks/data/idp/idx/authenticator-verification-email-no-profile.json
+++ b/playground/mocks/data/idp/idx/authenticator-verification-email-no-profile.json
@@ -52,7 +52,7 @@
             "type":"object",
             "options":[
               {
-                "label":"Primary Email Address",
+                "label":"Email",
                 "value":{
                   "form":{
                     "value":[
@@ -129,7 +129,7 @@
       "type":"email",
       "key": "okta_email",
       "id":"eae1d4u7aVB3VBbI70g4",
-      "displayName":"Primary Email Address",
+      "displayName":"Email",
       "methods":[
         {
           "type":"email"
@@ -144,7 +144,7 @@
         "type":"email",
         "key": "okta_email",
         "id":"eae1d4u7aVB3VBbI70g4",
-        "displayName":"Primary Email Address",
+        "displayName":"Email",
         "methods":[
           {
             "type":"email"
--- a/src/v2/ion/i18nTransformer.js
+++ b/src/v2/ion/i18nTransformer.js
@@ -267,6 +267,9 @@ const getI18nKey = (i18nPath) => {
  */
 const getI18NValue = (i18nPath, defaultValue, params = []) => {
   const i18nKey = getI18nKey(i18nPath);
+  // TODO : OKTA-397225
+  // here defaultValue is uiSchema label or placeholders, some lables may be customized by 
+  // admin to anything string. We should not localize and replace these customized labels even if i18nkey exists
   if (i18nKey) {
     return loc(i18nKey, 'login', params);
   } else {
--- a/playground/mocks/data/idp/idx/authenticator-verification-email-polling-long.json
+++ b/playground/mocks/data/idp/idx/authenticator-verification-email-polling-long.json
@@ -52,7 +52,7 @@
             "type":"object",
             "options":[
               {
-                "label":"Primary Email Address",
+                "label":"Email",
                 "value":{
                   "form":{
                     "value":[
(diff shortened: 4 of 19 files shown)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -963,7 +963,7 @@ oie.registration.form.title = Sign up
 oie.registration.form.submit = Sign Up
 oie.user.profile.lastname= Last name
 oie.user.profile.firstname= First name
-oie.user.profile.primary.email= Primary email
+oie.user.profile.primary.email= Email
```
sherizada-okta, line 966: This label can be edited in Admin. It's only "Email" by default.

playground/mocks/data/idp/idx/authenticator-verification-email-no-profile.json
```diff
@@ -52,7 +52,7 @@
             "type":"object",
             "options":[
               {
-                "label":"Primary Email Address",
+                "label":"Email",
```
sherizada-okta, line 55: What about a test where the label is like "Custom Email" to check that "Email" is not being hardcoded.

packages/@okta/i18n/src/properties/login.properties
anipendakur-okta, line 966: This may break the monolith selenium tests. Please create a downstream build to verify before merging.

anipendakur-okta: There's one mapping in `i18nTransformer` to this key - `enroll-profile.userProfile.email` - The related selenium tests (in monolith, if any) may break.

anipendakur-okta:
Even though only mocks are changed here, I'd suggest changing the entire mock instead of just the label. That way the mocks can act as the source-of-truth (for things like i18n work that we're doing) for near-future.

packages/@okta/i18n/src/properties/login.properties
pradeepdewda-okta: Yes i will verify that first.

pradeepdewda-okta:
> Even though only mocks are changed here, I'd suggest changing the entire mock instead of just the label. That way the mocks can act as the source-of-truth (for things like i18n work that we're doing) for near-future.

This PR is about changing `Primary Email` to `Email`. We basically hardcoded that into widget and was not displaying correct label that comes from API. So replacing entire mock doesn't make sense where we don't anticipate any changes.

sherizada-okta (approved):

anipendakur-okta (approved):

src/v2/ion/i18nTransformer.js
```diff
@@ -267,7 +267,9 @@ const getI18nKey = (i18nPath) => {
  */
 const getI18NValue = (i18nPath, defaultValue, params = []) => {
   const i18nKey = getI18nKey(i18nPath);
-  if (i18nKey) {
+  // defaultValue is uiSchema label, these lables may be customized by 
+  // admin to anything. We should not localize customized labels
+  if (i18nKey && Bundles.login[i18nKey] === defaultValue) {
```
pradeepdewda-okta, line 272: seems this will cause the localization to be ignored.. for eg

localized string will not be equal to similar english string coming from server..

We need to provide some backend solution through which widget can identify between custom label and default labels..  Our current solution always replaces the schema label from the value exists in login.properties.. that can't be right for `custom labels`

pradeepdewda-okta: @anipendakur-okta @sherizada-okta fyi

pradeepdewda-okta: For eg. API should return something like this if admin changed the labels for any input fields
{
label: "Username",
name: "identifier",
customLabel: "true"
}

sherizada-okta: You can look up the key in the bundle, if it's not there, then use the custom key, which is also defined in the bundle. 
In this case it would just be `enroll.userProfile.email,custom = {0}`. Then you pass the param as a label.

sherizada-okta: Default key scenario: If the i18n key is in the bundle, then use that. User's customization is preserved.

Custom key scenario: If the i18n key is not in the bundle, use the custom i18n key. In this case, the user needs to customize the custom key. We can log when a custom i18n key is being used, for the user to easily find if they want to customize?

sherizada-okta: @pradeepdewda-okta do you see any gaps for a good UX in that solution ^ that would be covered by returning a new API key/value pair?

pradeepdewda-okta: > Default key scenario: Look up the i18n key using the API label value, if the i18n key is in the bundle, then use that. User's customization is preserved.

Then one issue here would be what if they customize the default key like `email` to `custom email`.. then it will get replaced by our existing key that is what currently happening..

pradeepdewda-okta: >We can log when a custom i18n key is being used, for the user to easily find if they want to customize?

that would be not ideal as an enduser to see such things if they open up browser console.. It has to go somewhere in our profile enrollment admin page.. where they can enter localization keys as well..

pradeepdewda-okta:
Based on https://okta.slack.com/archives/CCA77QVL2/p1621529364119000

I am undoing custom label changes which is causing localization to break.. will be tracked here post BC OKTA-397225

PR #1964, opened 2021-05-18
Adding Captcha TestCafe tests

kamaldeepcheema-okta:
## Description:

- Adding some testCafe tests for Captcha for the Sign-in, SSR, and SSPR flows.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-389457](https://oktainc.atlassian.net/browse/OKTA-389457)

### DIFF
--- a/test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
+++ b/test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
@@ -29,7 +29,8 @@ const ignoredMocks = [
   'authenticator-verification-okta-verify-signed-nonce-loopback.json',
   'authenticator-expired-password-with-enrollment-authenticator.json',
   'error-with-failure-redirect.json',
-  'identify-recovery-with-recaptcha-v2.json'
+  'identify-recovery-with-recaptcha-v2.json', // flaky on bacon
+  'identify-with-password-with-recaptcha-v2.json' // flaky on bacon
 ];
 
 const parseMockData = () => {
--- a/test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
+++ b/test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
@@ -29,8 +29,8 @@ const ignoredMocks = [
   'authenticator-verification-okta-verify-signed-nonce-loopback.json',
   'authenticator-expired-password-with-enrollment-authenticator.json',
   'error-with-failure-redirect.json',
-  'identify-recovery-with-recaptcha-v2.json', // flaky on bacon
-  'identify-with-password-with-recaptcha-v2.json' // flaky on bacon
+  'identify-recovery-with-recaptcha-v2.json', // No english leaks for this, just flaky on bacon due to loading the reCaptcha lib 
+  'identify-with-password-with-recaptcha-v2.json' // No english leaks for this, just flaky on bacon due to loading the reCaptcha lib 
 ];
 
 const parseMockData = () => {
--- a/test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
+++ b/test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
@@ -1,8 +1,8 @@
 import { RequestMock, RequestLogger, Selector } from 'testcafe';
 import IdentityPageObject from '../framework/page-objects/IdentityPageObject';
 import { checkConsoleMessages } from '../framework/shared';
-import xhrIdentifyRecoveryWithRecaptcha from '../../../playground/mocks/data/idp/idx/identify-recovery-with-recaptcha-v2.json';
-import xhrIdentifyRecoveryWithHcaptcha from '../../../playground/mocks/data/idp/idx/identify-recovery-with-hcaptcha.json';
+import xhrIdentifyRecoveryWithRecaptcha from '../../../playground/mocks/data/idp/idx/identify-recovery-with-recaptcha-v2';
+import xhrIdentifyRecoveryWithHcaptcha from '../../../playground/mocks/data/idp/idx/identify-recovery-with-hcaptcha';
 import xhrAuthenticatorVerifySelect from '../../../playground/mocks/data/idp/idx/authenticator-verification-select-authenticator';
 
 const identifyRecoveryWithReCaptchaMock = RequestMock()
--- a/test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
+++ b/test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
@@ -25,6 +25,14 @@ const identifyRequestLogger = RequestLogger(
   }
 );
 
+const reCaptchaRequestLogger = RequestLogger(
+  /\/recaptcha\/api2\/userverify/,
+  {
+    logRequestBody: true,
+    stringifyRequestBody: true,
+  }
+);
+
 fixture('Identify Recovery - reset flow with Captcha');
 
 async function setup(t) {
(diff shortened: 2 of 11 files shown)


test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  return identityPage;
+}
+
+// https://oktainc.atlassian.net/browse/OKTA-393059
+// We're disabling this test for now because there seems to be an underlying issue with with this test
+// in Bacon. Locally this test runs with no issues but it's very flaky on Bacon. OKTA-393059 is created to investigate
+// further.
+test.requestHooks(identifyRequestLogger, identifyRecoveryWithReCaptchaMock).skip('should be able to submit identifier with reCaptcha enabled', async t => {
```
kamaldeepcheema-okta, line 45: We skip this test (and the reCaptcha ones in other files) for now because there's additional investigation required to know why the reCaptcha eng-leaks and testCafe tests fail on Bacon only (https://oktainc.atlassian.net/browse/OKTA-393059). It's currently not clear as to why they can't pass clicking the submit button. In the test we already assert that the Captcha related DOM loads correctly but despite this, we can't get past the submit button sometimes.

test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
```diff
@@ -29,7 +29,8 @@ const ignoredMocks = [
   'authenticator-verification-okta-verify-signed-nonce-loopback.json',
   'authenticator-expired-password-with-enrollment-authenticator.json',
   'error-with-failure-redirect.json',
-  'identify-recovery-with-recaptcha-v2.json'
+  'identify-recovery-with-recaptcha-v2.json', // flaky on bacon
+  'identify-with-password-with-recaptcha-v2.json' // flaky on bacon
```
nikhilvenkatraman-okta, line 33: @kamaldeepcheema-okta  Are there known english leaks in this mock ? And are we planning to fix these later ?  or are they going to be ignored always ? 

FYI all ignored mocks need to be fixed by the 06/11 OIE deadline. If they have known leaks, I would recommend creating a follow up JIRA and add the labels `oie-new-org` `oie-frc-l10n` `oie-new-org-bug` 
Example https://oktainc.atlassian.net/browse/OKTA-389430

kamaldeepcheema-okta: @nikhilvenkatraman-okta there are no english-leaks for any of these two mocks. This is the issue that we've discussed offline and one that is tracked by https://oktainc.atlassian.net/browse/OKTA-393059. Here's the slack where we discussed this: https://okta.slack.com/archives/CAZH2MWEL/p1620156477193100. This is more to do with an issue specific to the Bacon env regarding reCaptcha, because these run fine with no issues locally. With this being said, I don't think this falls under the regular list of known leaks since there's no localization issue here.

nikhilvenkatraman-okta: While you are here, can we add a comment here saying that there are not known english leaks but just that it does not run on bacon ?

nikhilvenkatraman-okta (approved):

test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
kamaldeepcheema-okta: UPDATE: the test has been re-written in a different way to avoid the limitation we see in Bacon regarding reCaptcha lib. Now during the reCaptcha tests, we just validate whether we hit google's API with the expected siteKey during submission. From my troubleshooting, this is the best option we have as it's otherwise hard to mock the reCaptcha service and the overall flow. For our purposes this will suffice for validating that the functionality is working as expected.

test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
kamaldeepcheema-okta: Yep done

test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
```diff
@@ -0,0 +1,98 @@
...
+
+  const identityPage = await setup(t);
+
+  await identityPage.fillIdentifierField('test.identifier');
+
+  // Wait for the reCaptcha container to appear in the DOM and become visible.
+  await t.expect(Selector('#captcha-container').find('.grecaptcha-badge').exists).ok({timeout: 3000});
```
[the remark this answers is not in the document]
kamaldeepcheema-okta: Synced with @pradeepdewda-okta on slack regarding this. This is an important step of the test and will be kept. No flakiness comes from this as we're simply validating whether the captcha related DOMs get loaded properly before moving forward. This is a practice we follow in other tests as well

test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
```diff
@@ -0,0 +1,94 @@
...
+  const identityPage = await setup(t);
+  
+  await identityPage.fillIdentifierField('test.identifier');
+  
+  // Wait for the reCaptcha container to appear in the DOM and become visible.
+  await t.expect(Selector('#captcha-container').find('.grecaptcha-badge').exists).ok({timeout: 3000});
+
+  await identityPage.clickNextButton();
+
+  await t.wait(3000);
```
pradeepdewda-okta, line 59: We should not need this wait..

test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
```diff
@@ -0,0 +1,94 @@
...
+  return identityPage;
+}
+
+test.requestHooks(identifyRequestLogger, reCaptchaRequestLogger, identifyRecoveryWithReCaptchaMock)('should be able to submit identifier with reCaptcha enabled', async t => {
+  const identityPage = await setup(t);
+  
+  await identityPage.fillIdentifierField('test.identifier');
+  
+  // Wait for the reCaptcha container to appear in the DOM and become visible.
+  await t.expect(Selector('#captcha-container').find('.grecaptcha-badge').exists).ok({timeout: 3000});
```
pradeepdewda-okta, line 55: Not sure we discussed, trying to remember. Why this can't be rewritten in a way to avoid timeout ?

test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
kamaldeepcheema-okta: We do need this because otherwise in some cases the call to the reCaptcha API will not get registered in our request logger fast enough, and the line immediately after this could fail - causing flakiness. In my local tests, I noticed 1s wasn't enough all the time, so I increased it to 3s so it avoids any flakiness.

pradeepdewda-okta: Never a good fan of introducing `wait`..  what screen shows up after you click next button ?

Can we check for some other element after hitting submit and before checking t.expect(reCaptchaRequestLogger.count(() => true)).eql(1);

test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
kamaldeepcheema-okta: These are DOM elements that get added to the page by the Captcha libs and since this happens on the fly and after all the related scripts load, we can't always guarantee that it will always be there without some buffer. Also, this timeout doesn't necessarily mean that the test will always wait 3s, if it detects the element before then it continues. It just guarantees that we wait **at most** 3s to check the selector: https://testcafe.io/documentation/402827/guides/concepts/built-in-wait-mechanisms#wait-mechanism-for-selectors

pradeepdewda-okta: But shouldn't it fail the expect and not proceed if it can't find the DOM element in the first place.. ?

kamaldeepcheema-okta: transferred this conversation over to Slack

test/testcafe/spec/IdentifyRecoveryWithCaptcha_spec.js
kamaldeepcheema-okta: transferred this conversation over to Slack

pradeepdewda-okta (approved):
💯

PR #1970, opened 2021-05-19
Fixing the spacing in the social auth links when displaying multiple IDPs

kamaldeepcheema-okta:
## Description:

- When we have multiple IDPs enabled, the spacing between the footer links and the IDP buttons is more than it should be. Please refer to the images below. 

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

Before:
Before

After:
After

### Issue:

- [OKTA-396709](https://oktainc.atlassian.net/browse/OKTA-396709)

### DIFF
--- a/assets/sass/v2/_identify.scss
+++ b/assets/sass/v2/_identify.scss
@@ -15,6 +15,16 @@
   }
 }
 
+.sign-in-with-idp {
+  .okta-idps-container {
+    .social-auth-button.link-button {
+      &:last-of-type {
+        margin-bottom: 0px;
+      }   
+    }   
+  }
+}
+
 .sign-in-with-device-option,
 .sign-in-with-idp,
 .custom-buttons {
--- a/assets/sass/v2/_identify.scss
+++ b/assets/sass/v2/_identify.scss
@@ -19,9 +19,9 @@
   .okta-idps-container {
     .social-auth-button.link-button {
       &:last-of-type {
-        margin-bottom: 0px;
-      }   
-    }   
+        margin-bottom: 0;
+      }
+    }
   }
 }
 


assets/sass/v2/_identify.scss
```diff
@@ -15,6 +15,16 @@
   }
 }
 
+.sign-in-with-idp {
+  .okta-idps-container {
+    .social-auth-button.link-button {
+      &:last-of-type {
+        margin-bottom: 0;
```
kamaldeepcheema-okta, line 22: This is meant to override the styling here: https://github.com/okta/okta-signin-widget/blob/master/assets/sass/modules/_btns.scss#L94.

We don't remove that since that might have an unwanted side effect of potentially breaking the style of a v1 feature. Instead we override it for the multiple IDPs (OIE only) case.

pradeepdewda-okta: @kamaldeepcheema-okta can you verify that ? If not, then then we should make the fix there.

kamaldeepcheema-okta: @pradeepdewda-okta Yep I spent some time into this before going forward with this solution. Here's the summary of what I found:

- The associated css has been in _btnc.scss for 5+ years, part of the original codebase ported over from the monolith. Back then we had a SocialAuth.js view (now deleted) which contained the logic for social auth
- Now this logic is contained within our src/models/Settings.js file - this was implemented around 4 years ago: https://github.com/okta/okta-signin-widget/blob/master/src/models/Settings.js#L287
- More recently, we added this in our FormInputFactory to support the new pipeline - https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/internals/FormInputFactory.js#L104
- Although I don't have full context on the original social-auth feature, from the information above it does seem like we support two versions of it. That's why I opted to override instead of totally removing that css since it's safer.

@gowthamidommety-okta I noticed you added the original in Settings.js through https://oktainc.atlassian.net/browse/OKTA-119408. Can you just confirm if we still support this today?

gowthamidommety-okta: @kamaldeepcheema-okta the change I added was specific to PIV auth, this css looks specific to social auth. I tested this locally, and looks like we do need that `margin-bottom: 25px;`

socialAuthCss

kamaldeepcheema-okta: thank you very much @gowthamidommety-okta for looking into this. 

@pradeepdewda-okta it does look like we need to keep the original css then and overriding in this manner is the more appropriate approach.

pradeepdewda-okta: Perfect finding. 👍

pradeepdewda-okta (approved):

PR #1973, opened 2021-05-20
fix(v2): support config['features.showPasswordToggleOnSignInPage']

sherizada-okta:
## Description:
For customizing SIW feature flags, both of the following methods should work:
- will work with this change: `config['features.showPasswordToggleOnSignInPage'] = boolean`
- already works: `config.features.showPasswordToggleOnSignInPage = boolean`

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

### Issue:

- [OKTA-397629](https://oktainc.atlassian.net/browse/OKTA-397629)

### DIFF
--- a/src/models/Settings.js
+++ b/src/models/Settings.js
@@ -171,7 +171,10 @@ export default Model.extend({
       fn: function() {
         // showPasswordToggle is for OIE only.
         // Used to default showPasswordToggleOnSignInPage to true.
-        return this.options?.features?.showPasswordToggleOnSignInPage !== false;
+        const defaultValue = true;
+        const customizedValue = this.options?.features?.showPasswordToggleOnSignInPage ??
+          this.options?.['features.showPasswordToggleOnSignInPage'];
+        return customizedValue ?? defaultValue;
       },
       cache: true,
     },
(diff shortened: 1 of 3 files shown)


src/models/Settings.js
```diff
@@ -163,7 +163,10 @@ export default Model.extend({
       fn: function() {
         // showPasswordToggle is for OIE only.
         // Used to default showPasswordToggleOnSignInPage to true.
-        return this.options?.features?.showPasswordToggleOnSignInPage !== false;
+        const defaultValue = true;
+        const customizedValue = this.options?.features?.showPasswordToggleOnSignInPage ??
```
pradeepdewda-okta, line 167: so many questions marks..

pradeepdewda-okta: I would rather create it own function

pradeepdewda-okta (approved):

src/models/Settings.js
sherizada-okta: @pradeepdewda-okta what kind of function?

pradeepdewda-okta: just a wrapper `isCustomized()` ?

sherizada-okta: This is an edge case. The reason I have to do this logic for `customizedValue` is because this setting has a different default value in OIE, so I don't want to use the value that would have been passed to this derived function, `this.settings.get('features.showPasswordToggleOnSignInPage')`. The wrapper would go unused.

PR #1531, opened 2020-11-24
fix: add Okta Verify icon to the sign in using Okta Verify button

jhe-okta:
## Description:

add Okta Verify icon to the sign in using Okta Verify button

## PR Checklist

- [x] Have you verified the basic functionality for this change?

- [x] Added unit tests?

- [ ] Added e2e tests

- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?

### Screenshot/Video:
Screen Shot 2020-11-24 at 12 48 08 PM
Screen Shot 2020-11-24 at 12 51 03 PM

### Reviewers:
@stevennguyen-okta @pradeepdewda-okta 

### Issue:

- [OKTA-347753](https://oktainc.atlassian.net/browse/OKTA-347753)

PR #1983, opened 2021-05-24
fix: update OV icon during setup - download

jasleennuno-okta:
## Description:
- fix OV icon during set up

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

#### Before
Screen Shot 2021-05-21 at 1 32 00 PM

#### After
Screen Shot 2021-05-24 at 3 52 25 PM

### Reviewers:
@mauriciocastillosilva-okta @stevennguyen-okta 

### Issue:

- [OKTA-398364](https://oktainc.atlassian.net/browse/OKTA-398364)


mauriciocastillosilva-okta (commented):
I see we already have 6 okta verify icons in https://github.com/okta/okta-signin-widget/tree/master/assets/img/icons/mfa, can we reuse one of those instead of adding a new one? or maybe we need to clean up those icons too?

jasleennuno-okta:
> I see we already have 6 okta verify icons in https://github.com/okta/okta-signin-widget/tree/master/assets/img/icons/mfa, can we reuse one of those instead of adding a new one? or maybe we need to clean up those icons too?

@mauriciocastillosilva-okta i think the problem is that we added backgrounds for the icons, but in this particular case we don't want a background so that the image can be bigger
cc: @stevennguyen-okta

stevennguyen-okta:
> > I see we already have 6 okta verify icons in https://github.com/okta/okta-signin-widget/tree/master/assets/img/icons/mfa, can we reuse one of those instead of adding a new one? or maybe we need to clean up those icons too?
> 
> @mauriciocastillosilva-okta i think the problem is that we added backgrounds for the icons, but in this particular case we don't want a background so that the image can be bigger
> cc: @stevennguyen-okta

If we can find a way to minimize the amount of icons I would be for that, good Q though. I assumed we'd keep the same method we had originally because the OV Logo has to account for combinations of:

- Different sizes in some places (I have sizes 70, 48, 38@1x, and 38@2x for example)
- Some icons have white padding around them and others don't (for example beacon icon has white padding around it)

jasleennuno-okta:
it looks to me like these are the same - they look the same in IntelliJ, same file size, same dimensions
oktaVerify@2x.png, oktaVerify_ico@2x.png, oktaVerify_70x70.png
let me know if that's not true..

would it make sense to merge this change first since it's a UI bug and create a jira to clean up the icons at a later time?

mauriciocastillosilva-okta:
Majority of authenticators/factors have only 4 icons, which account for the sizes you are mentioning, and as far as I can tell in all other factors/authenticators we have padding around the icon. We have those 4 icons for okta verify, so all good there.
The 2 icons that we already have extra for OV are oktaVerify_authenticator*.png, and those do not have the white padding around
- https://github.com/okta/okta-signin-widget/blob/master/assets/img/icons/mfa/oktaVerify_authenticator%402x.png
- https://github.com/okta/okta-signin-widget/blob/master/assets/img/icons/mfa/oktaVerify_authenticator.png

Both were added in https://github.com/okta/okta-signin-widget/pull/1531. Can we reuse one of those here?

assets/sass/modules/_enroll.scss
```diff
@@ -283,7 +283,8 @@ $enroll-content-spacing: 15px;
 }
 
 .okta-verify-38 {
```
mauriciocastillosilva-okta, line 285: can we rename this to something like `.okta-verify-download-icon`? it is no longer the 38x38 we had, so maybe this will help us know what each css class is for

mauriciocastillosilva-okta (commented):
just a small change, so we can identify each css class better

jasleennuno-okta:
Follow up jira - https://oktainc.atlassian.net/browse/OKTA-398547

mauriciocastillosilva-okta (approved):

PR #1985, opened 2021-05-25
fix: set display:none for search input when not visible for a11y

TiffanyNg-okta:
## Description:

- The hidden search input was getting picked up by screen readers when the country dropdown is collapsed
- Set `display:none` for search input when not visible for a11y
- Note that the country dropdown itself doesn’t get picked up by screen readers currently. This is a known issue and is in Katie Le’s list of accessibility issues that will be handled separately.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Before fix:

https://user-images.githubusercontent.com/76003796/119437288-a9089c80-bcd2-11eb-8fd6-6fc0b4d46448.mov

After fix:

https://user-images.githubusercontent.com/76003796/119437293-ac9c2380-bcd2-11eb-8c83-768cc4b00b57.mov

### Issue:

- [OKTA-390748](https://oktainc.atlassian.net/browse/OKTA-390748)

### DIFF
--- a/assets/sass/widgets/_chosen-override.scss
+++ b/assets/sass/widgets/_chosen-override.scss
@@ -1,8 +1,0 @@
-/* Override _chosen.scss styles */
-.chzn-drop > .chzn-search > input {
-  display: none;
-}
-
-.chzn-with-drop > .chzn-drop > .chzn-search > input {
-  display: inline-block;
-}
--- a/assets/sass/widgets/_chosen-override.scss
+++ b/assets/sass/widgets/_chosen-override.scss
@@ -1,0 +1,8 @@
+/* Override _chosen.scss styles */
+.chzn-drop > .chzn-search > input {
+  display: none;
+}
+
+.chzn-with-drop > .chzn-drop > .chzn-search > input {
+  display: inline-block;
+}
--- a/test/unit/spec/EnrollSms_spec.js
+++ b/test/unit/spec/EnrollSms_spec.js
@@ -215,7 +215,7 @@ Expect.describe('EnrollSms', function() {
       });
     });
     itp('has hidden country search input by default', function() {
-      return setup().then(function(test) {
+      return setup().then(function() {
         const searchInput = $sandbox.find('.chzn-search > input');
         expect(searchInput.css('display')).toBe('none');
       });
--- a/test/unit/spec/EnrollSms_spec.js
+++ b/test/unit/spec/EnrollSms_spec.js
@@ -214,10 +214,15 @@ Expect.describe('EnrollSms', function() {
         expect(test.form.getCodeFieldAutocomplete()).toBe('off');
       });
     });
-    itp('has search input display set correctly', function() {
+    itp('has hidden country search input by default', function() {
       return setup().then(function(test) {
         const searchInput = $sandbox.find('.chzn-search > input');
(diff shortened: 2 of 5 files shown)


assets/sass/widgets/_chosen-override.scss
```diff
@@ -0,0 +1,8 @@
+/* Override _chosen.scss styles */
+.chzn-drop > .chzn-search > input {
+  display: none;
+}
+
+.chzn-with-drop > .chzn-drop > .chzn-search > input {
+  display: inline-block;
+}
```
sherizada-okta, line 8: do you want these overrides to apply to all components or just a certain one?

test/unit/spec/EnrollSms_spec.js
```diff
@@ -214,6 +214,15 @@ Expect.describe('EnrollSms', function() {
...
     });
+    itp('has search input display set correctly', function() {
+      return setup().then(function(test) {
+        const searchInput = $sandbox.find('.chzn-search > input');
+        expect(searchInput.css('display')).toBe('none');
+        const mousedown = $.Event('mousedown');
+        test.form.countryDropdown().trigger(mousedown);
+        expect(searchInput.css('display')).toBe('inline-block');
+      });
+    });
```
sherizada-okta, line 225: This test case could be split up so the test titles are more informative, like 'has a hidden search input by default' and 'shows search input on mousedown'

assets/sass/widgets/_chosen-override.scss
TiffanyNg-okta: I was specifically looking at the country chosen dropdown in the Enroll SMS view. But I think these could be applied to any chosen dropdown with search as they probably all have the same a11y issue that I'm trying to resolve.

test/unit/spec/EnrollSms_spec.js
TiffanyNg-okta: Good point, I split up the test.

assets/sass/widgets/_chosen-override.scss
sherizada-okta: I would check with the other views that also use the chosen component that these styles make sense.

sherizada-okta: These styles should go in the existing https://github.com/okta/okta-signin-widget/blob/master/assets/sass/widgets/_chosen.scss ?

TiffanyNg-okta: Hmm I think you brought up a good point. I'm not sure I know all the views that this chosen dropdown with search is being used in. I think for the purposes of this ticket, I should focus on the Enroll SMS view as it mentioned this view specifically and not others. If we find that there are similar a11y issues in other views, I can open a separate PR/ticket. Moving the styles under `.enroll-sms` selector for now which should limit the scope to just the Enroll SMS view. This would be in the okta-signin-widget/assets/sass/modules/_enroll.scss. Let me know if you have other ideas.

Also side note, I originally asked in the eng-sign-in-widget slack channel if I should update the _chosen.scss directly but was recommended to not make changes there and have a separate override file: https://okta.slack.com/archives/CAZH2MWEL/p1621981282066500. Seems the _chosen.scss was forked from okta-ui repo.

sherizada-okta (approved):

PR #2002, opened 2021-06-01
feat: Expose ability to hide "Back to sign in" link for all MFA views

jmelberg-okta:
## Description:

- When the configuration option `features.hideSignOutLinkInMFA` is provided views should **not contain** the "Back to sign in" text.

## PR Checklist

- [x] Added e2e tests

### Reviewers:

- @mauriciocastillosilva-okta

### Issue:

- [OKTA-395411](https://oktainc.atlassian.net/browse/OKTA-395411)

### DIFF
--- a/src/v2/models/AppState.js
+++ b/src/v2/models/AppState.js
@@ -194,18 +194,15 @@ export default Model.extend({
       transformedResponse.currentAuthenticator?.poll?.refresh;
   },
 
-  // Sign Out link will be displayed in the footer of a form, unless
-  // - widget config hideSignOutLinkInMFA=true or mfaOnlyFlow=true
-  // and form is for identity verification (FORMS_FOR_VERIFICATION)
+  // Sign Out link will be displayed in the footer of a form, unless:
+  // - widget configuration set hideSignOutLinkInMFA or mfaOnlyFlow to true
   // - cancel remediation form is not present in the response
   // - form is part of our list FORMS_WITHOUT_SIGNOUT
   shouldShowSignOutLinkInCurrentForm(hideSignOutLinkInMFA) {
     const idxActions = this.get('idx') && this.get('idx').actions;
     const currentFormName = this.get('currentFormName');
-    const hideSignOutConfigOverride = hideSignOutLinkInMFA
-      && FORMS_FOR_VERIFICATION.includes(currentFormName);
 
-    return !hideSignOutConfigOverride
+    return !hideSignOutLinkInMFA
       && _.isFunction(idxActions?.cancel)
       && !FORMS_WITHOUT_SIGNOUT.includes(currentFormName);
   },
--- a/test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
+++ b/test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
@@ -1,5 +1,6 @@
 import { RequestMock, RequestLogger } from 'testcafe';
 
+import { renderWidget } from '../framework/shared';
 import SelectFactorPageObject from '../framework/page-objects/SelectAuthenticatorPageObject';
 import ChallengeFactorPageObject from '../framework/page-objects/ChallengeFactorPageObject';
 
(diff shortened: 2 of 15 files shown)


pradeepdewda-okta:
you can change the title that says `Expose ability to hide "Back to sign in" link for all mfa views`

src/v2/models/AppState.js
```diff
@@ -194,18 +194,15 @@ export default Model.extend({
       transformedResponse.currentAuthenticator?.poll?.refresh;
   },
 
-  // Sign Out link will be displayed in the footer of a form, unless
-  // - widget config hideSignOutLinkInMFA=true or mfaOnlyFlow=true
-  // and form is for identity verification (FORMS_FOR_VERIFICATION)
+  // Sign Out link will be displayed in the footer of a form, unless:
```
pradeepdewda-okta, line 197: it's unfortunate that the name of the parameter has `Sign Out` in it and now we call it `Back to sign in`.. can we at least mention that in comments and probably change the variable names ?

pradeepdewda-okta (approved):

src/v2/models/AppState.js
jmelberg-okta: Agree - we should come up with a deprecation strategy for fields that don't map the best with OIE enabled. I'll create a follow-up for a discussion on this.

pradeepdewda-okta: here in comment we can change to `Back to sign in formerly Sign Out link will be displayed .....`

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -209,7 +210,21 @@ test.requestHooks(mockChallengePassword)('should load select authenticator list'
...
+  const selectFactorPage = await setup(t);
+  await renderWidget({
+    features: { hideSignOutLinkInMFA: true },
+  });
+  await t.expect(selectFactorPage.getFormTitle()).eql('Verify it\'s you with an authenticator');
+  // signout link is not visible
+  await t.expect(await selectFactorPage.signoutLinkExists()).notOk();
+});
+
+test.requestHooks(mockChallengePassword)('should load select authenticator list', async t => {
```
mauriciocastillosilva-okta, line 225: do we need this test? Seems it is just testing the title, and I think we have that already in tests above

test/unit/spec/v2/models/AppState_spec.js
```diff
@@ -25,21 +25,17 @@ describe('v2/models/AppState', function() {
       this.initAppState({ idx: { actions: { cancel: 'invalid' } } });
       expect(this.appState.shouldShowSignOutLinkInCurrentForm()).toBe(false);
     });
-    it('returns false if the currentFormName is in FORMS_WITHOUT_SIGNOUT', () => {
-      this.initAppState({ idx: { actions: { cancel: () => {} } } }, FORMS_WITHOUT_SIGNOUT[0]);
```
mauriciocastillosilva-okta, line 29: we still need a test for `FORMS_WITHOUT_SIGNOUT`, can we add it back? The test for `FORMS_FOR_VERIFICATION` is that one that is no longer needed

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
jmelberg-okta: I think this was a copy + paste error.. it shouldn't be here! Will remove. 😞

test/unit/spec/v2/models/AppState_spec.js
jmelberg-okta: Woah - good catch. I think I got too aggressive with my code deletion and missed this when reviewing. Thank you!

test/unit/spec/v2/models/AppState_spec.js
```diff
@@ -25,21 +25,21 @@ describe('v2/models/AppState', function() {
...
-      expect(this.appState.shouldShowSignOutLinkInCurrentForm()).toBe(false);
+    it('returns false if param hideSignOutLinkInMFA is true but there is no remediation action', () => {
```
mauriciocastillosilva-okta, line 28: this is a double criteria to return false, right? hideSignOutLinkInMFA=true and no actions.
Not sure if that was the intention, or if we should have one more for hideSignOutLinkInMFA=false and no actions

mauriciocastillosilva-okta (approved):
looks good, just left one comment I missed in the previous round. 
🐑

test/unit/spec/v2/models/AppState_spec.js
jmelberg-okta: This should actually be `false`, as we'd never execute the check for remediation actions unless `hideSignOutLinkInMFA` is set to `false`. Updating!

PR #1989, opened 2021-05-25
added back to sign link to terminal views

pradeepdewda-okta:
## Description:

1. By default add Back to sign in link to all terminal views
2. Add views that do not need `Back to sign in` link

## PR Checklist

- [ ] Have you verified the basic functionality for this change?

- [ ] Added unit tests?

- [ ] Added e2e tests

- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?

### Screenshot/Video:
Screen Shot 2021-06-02 at 1 15 50 PM
Screen Shot 2021-06-02 at 5 23 34 PM
Screen Shot 2021-06-02 at 10 26 03 AM
Screen Shot 2021-06-02 at 10 35 37 AM

### Issue:

- [OKTA-392835](https://oktainc.atlassian.net/browse/OKTA-392835)

PR #2004, opened 2021-06-01
Change OIE session expiration message

ethanhsiung-okta:
Resolves: OKTA-394634

## Description:

- Change session expiration message from `The session has expired.` to `The session has expired. Refresh the page and try again.`
- Passing downstream artifact bacon run: https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&branch=d16t-okta-signin-widget-c54a556-60b6682d&page=1&pageSize=6

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Before:
Screen Shot 2021-05-18 at 3 31 30 PM

After:
Screen Shot 2021-05-25 at 4 33 44 PM

### Reviewers:

- @rameshbhojan-okta @anipendakur-okta 

### Issue:

- [OKTA-394634](https://oktainc.atlassian.net/browse/OKTA-394634)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -972,7 +972,7 @@ oie.selfservice.unlock_user.failed.message = We are unable to unlock your accoun
 oie.selfservice.user.unlock.not.allowed = Self Service Unlock is not allowed at this time. Please contact support for assistance.
 
 ## Terminal errors
-idx.session.expired = The session has expired.
+idx.session.expired = The session has expired. Refresh the page and try again.
 oie.registration.is.not.enabled = Sign up is not enabled for this organization.
 oie.forgot.password.is.not.enabled = Forgot password is not enabled for this organization.
 idx.return.error = Could not process this email link. Return to the screen where you requested it.
--- a/test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
+++ b/test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
@@ -220,7 +220,7 @@ test
     const challengeEmailPageObject = await setup(t);
     await t.expect(challengeEmailPageObject.resendEmailView().hasClass('hide')).ok();
     await t.wait(5000);
-    await t.expect(challengeEmailPageObject.getErrorFromErrorBox()).eql('The session has expired.');
+    await t.expect(challengeEmailPageObject.getErrorFromErrorBox()).eql('The session has expired. Refresh the page and try again.');
     // Check no poll requests were made further. There seems to be no way to interrupt a poll with mock response.
     await t.expect(logger.count(
       record => record.response.statusCode === 200 &&
--- a/test/testcafe/spec/ChallengeAuthenticatorPassword_spec.js
+++ b/test/testcafe/spec/ChallengeAuthenticatorPassword_spec.js
@@ -116,7 +116,7 @@ test.requestHooks(sessionExpiresDuringPassword)('challege password authenticator
   await challengePasswordPage.switchAuthenticatorExists();
   await challengePasswordPage.verifyFactor('credentials.passcode', 'test');
   await challengePasswordPage.clickNextButton();
-  await t.expect(challengePasswordPage.getErrorFromErrorBox()).eql('The session has expired.');
+  await t.expect(challengePasswordPage.getErrorFromErrorBox()).eql('The session has expired. Refresh the page and try again.');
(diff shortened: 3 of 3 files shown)


vincentvoong-okta (commented):
LGTM

rameshbhojan-okta:
> LGTM

@vincentvoong-okta - Did we intentionally make a choice to not use fillers like `Please` etc.? e.g. `Please refresh the page...`?

rameshbhojan-okta (approved):

pradeepdewda-okta:
How to make sure that the action `Refresh the page and try again` will always work for all the flows ?

pradeepdewda-okta (changes requested):
I am working on this ticket https://oktainc.atlassian.net/browse/OKTA-392835 that adds `Back to sign in` link which basically does a page refresh and clears any locally stored stateToken.. I feel that should be the right solution.

ethanhsiung-okta:
> I am working on this ticket https://oktainc.atlassian.net/browse/OKTA-392835 that adds `Back to sign in` link which basically does a page refresh and clears any locally stored stateToken.. I feel that should be the right solution.

@pradeepdewda-okta I agree, this makes more sense. @vincentvoong-okta @rameshbhojan-okta thoughts?

rameshbhojan-okta:
> > I am working on this ticket https://oktainc.atlassian.net/browse/OKTA-392835 that adds `Back to sign in` link which basically does a page refresh and clears any locally stored stateToken.. I feel that should be the right solution.
> 
> @pradeepdewda-okta I agree, this makes more sense. @vincentvoong-okta @rameshbhojan-okta thoughts?

Ok, so then @pradeepdewda-okta - Are we saying we will never throw the Session Expired? I understand your fix would take care of the primary issue reported in the bug - but this PR which fixes the error message was for general scenarios to improve the message with feedback for user when the session indeed expires. (e.g. even from API calls' perspective and not just our SIW UI).

pradeepdewda-okta:
@rameshbhojan-okta User will still see the error message, as they used to see in `v1` but they can also take action to go back to sign in page once we add `Back to sign in` link to all terminal states. @unniravindranathan-okta

vincentvoong-okta:
@ethanhsiung-okta @pradeepdewda-okta 
I didn't realize we're also adding `Back to sign in` on this page. At least from the UX side, that would solve the problem of just dead-ending the user.

@rameshbhojan-okta 
Yes, removing "Please" was intentional - generally speaking, we want to use please as little as possible in our language unless we're asking the user to do something fairly inconvenient like waiting for something to download.

rameshbhojan-okta:
> @rameshbhojan-okta User will still see the error message, as they used to see in `v1` but they can also take action to go back to sign in page once we add `Back to sign in` link to all terminal states. @unniravindranathan-okta

Ok sounds good, @pradeepdewda-okta . So, because of the reasoning above and for the benefit of the API consumers - I think it is still good to get this error message fix in? I am trying to understand if you still need to block this PR or can you go ahead and remove the change request from the PR? 😄

pradeepdewda-okta:
> > @rameshbhojan-okta User will still see the error message, as they used to see in `v1` but they can also take action to go back to sign in page once we add `Back to sign in` link to all terminal states. @unniravindranathan-okta
> 
> Ok sounds good, @pradeepdewda-okta . So, because of the reasoning above and for the benefit of the API consumers - I think it is still good to get this error message fix in? I am trying to understand if you still need to block this PR or can you go ahead and remove the change request from the PR? 😄

@rameshbhojan-okta just wanted to understand how it will benefit API consumers ? this PR doesn't change API response.. 
Here is my PR that covers generalized use cases for terminal states. https://github.com/okta/okta-signin-widget/pull/1989

ethanhsiung-okta:
Chatted offline with @rameshbhojan-okta and @pradeepdewda-okta about this.
Pradeep's changes here https://github.com/okta/okta-signin-widget/pull/1989 would add a `Back to sign in`, so we no longer think this change (or any text change) is necessary

rameshbhojan-okta:
> @rameshbhojan-okta just wanted to understand how it will benefit API consumers ? 

@pradeepdewda-okta - Yes, this doesn't change the API response/attributes as such but it does affect the value of the error message being returned by the API right, which is what our UI currently reads and displays? 
So, where I was getting at was e.g. CIAM customers that may want to consume APIs directly (or through our SDKs), the error messages often provide feedback for them to decide on the next step of action(s).

But, overall - that was just my general comment and not very specific to this change/PR - as I don't feel very strongly about the error message as we can leave it as-is like we discussed offline on slack.

PR #2006, opened 2021-06-02
bug: show switch authenticator link only when needed

mauriciocastillosilva-okta:
## Description:

- Show the link only if there are more authenticators to choose from. If there is only 1 authenticator, there is no point of showing it
- Context: API will now return the link all the time, so the logic to hide or show it is on the SIW

Resolves: OKTA-399497

## PR Checklist

- [x] Added unit tests?
- [x] Added e2e tests

### Screenshot/Video:

For more context, these are the links:
| Set up | Verify |
| --- | --- |
| SIW-setup authenticator | SIW-verify authenticator |

### Issue:

- [OKTA-399497](https://oktainc.atlassian.net/browse/OKTA-399497)

### DIFF
--- a/src/v2/ion/ui-schema/ion-object-handler.js
+++ b/src/v2/ion/ui-schema/ion-object-handler.js
@@ -18,7 +18,7 @@ import { AUTHENTICATOR_KEY } from '../RemediationConstants';
  * @param {AuthenticatorOption[]} options
  * @param {( AuthenticatorEnrollment[] || Authenticator[] )} authenticators
  */
-const createOVOptions = (options = []) => {
+export const createOVOptions = (options = []) => {
   // Split OV into individual entries for verification (one for each method).
   const ovItem = options.find((option) => option.relatesTo.key === AUTHENTICATOR_KEY.OV);
   const methodTypeObj = ovItem?.value?.form?.value?.find(v => v.name === 'methodType');
--- a/src/v2/ion/ui-schema/ion-object-handler.js
+++ b/src/v2/ion/ui-schema/ion-object-handler.js
@@ -18,7 +18,7 @@ import { AUTHENTICATOR_KEY } from '../RemediationConstants';
  * @param {AuthenticatorOption[]} options
  * @param {( AuthenticatorEnrollment[] || Authenticator[] )} authenticators
  */
-function createOVOptions(options = []) {
+export function createOVOptions(options = []) {
   // Split OV into individual entries for verification (one for each method).
   const ovItem = options.find((option) => option.relatesTo.key === AUTHENTICATOR_KEY.OV);
   const methodTypeObj = ovItem?.value?.form?.value?.find(v => v.name === 'methodType');
@@ -109,7 +109,8 @@ function getAuthenticatorsVerifyUiSchema({ options }) {
 /**
   * Create ui schema for ION field that has type 'object'.
   */
-function createUiSchemaForObject(ionFormField, remediationForm, transformedResp, createUISchema, settings) {
+export default function createUiSchemaForObject(ionFormField, remediationForm, transformedResp,
+  createUISchema, settings) {
   const uiSchema = {};
 
   if (ionFormField.name === 'authenticator' &&
@@ -148,8 +149,3 @@ function createUiSchemaForObject(ionFormField, remediationForm, transformedResp,
 
   return uiSchema;
 }
-
-export {
-  createUiSchemaForObject as default,
-  createOVOptions
-};
--- a/src/v2/ion/ui-schema/ion-object-handler.js
+++ b/src/v2/ion/ui-schema/ion-object-handler.js
@@ -18,7 +18,7 @@ import { AUTHENTICATOR_KEY } from '../RemediationConstants';
  * @param {AuthenticatorOption[]} options
  * @param {( AuthenticatorEnrollment[] || Authenticator[] )} authenticators
  */
-export const createOVOptions = (options = []) => {
+function createOVOptions(options = []) {
   // Split OV into individual entries for verification (one for each method).
   const ovItem = options.find((option) => option.relatesTo.key === AUTHENTICATOR_KEY.OV);
   const methodTypeObj = ovItem?.value?.form?.value?.find(v => v.name === 'methodType');
@@ -63,9 +63,9 @@ export const createOVOptions = (options = []) => {
       ovItem.relatesTo.deviceKnown ? options.unshift(signedNonceOption) : options.push(signedNonceOption);
     }
   }
-};
+}
 
-const createAuthenticatorOptions = (options = []) => {
+function createAuthenticatorOptions(options = []) {
   createOVOptions(options);
   return options.map(option => {
     const value = option.value?.form?.value || [];
@@ -90,26 +90,26 @@ const createAuthenticatorOptions = (options = []) => {
       authenticatorKey: option.relatesTo?.key,
     };
   });
-};
+}
 
-const getAuthenticatorsEnrollUiSchema = ({ options }) => {
+function getAuthenticatorsEnrollUiSchema({ options }) {
   return {
     type: 'authenticatorEnrollSelect',
     options: createAuthenticatorOptions(options),
   };
-};
+}
 
-const getAuthenticatorsVerifyUiSchema = ({ options }) => {
+function getAuthenticatorsVerifyUiSchema({ options }) {
   return {
     type: 'authenticatorVerifySelect',
     options: createAuthenticatorOptions(options),
   };
-};
+}
 
 /**
   * Create ui schema for ION field that has type 'object'.
   */
-const createUiSchemaForObject = (ionFormField, remediationForm, transformedResp, createUISchema, settings) => {
+function createUiSchemaForObject(ionFormField, remediationForm, transformedResp, createUISchema, settings) {
   const uiSchema = {};
 
   if (ionFormField.name === 'authenticator' &&
@@ -147,6 +147,9 @@ const createUiSchemaForObject = (ionFormField, remediationForm, transformedResp,
   }
 
   return uiSchema;
-};
+}
 
-export default createUiSchemaForObject;
+export {
+  createUiSchemaForObject as default,
+  createOVOptions
+};
--- a/src/v2/models/AppState.js
+++ b/src/v2/models/AppState.js
@@ -17,6 +17,7 @@ import {
   FORMS_WITH_STATIC_BACK_LINK,
   FORMS_FOR_VERIFICATION,
 } from '../ion/RemediationConstants';
+import { createOVOptions } from '../ion/ui-schema/ion-object-handler';
 import { _ } from '../mixins/mixins';
 
 /**
@@ -75,6 +76,19 @@ export default Model.extend({
 
   hasRemediationObject(formName) {
     return this.get('idx').neededToProceed.find((remediation) => remediation.name === formName);
+  },
+
+  hasMoreThanOneAuthenticatorOption(formName) {
+    const form = this.hasRemediationObject(formName);
+    if (form) {
+      const authenticator = form.value.find((value) => value.name === 'authenticator');
+      let authenticatorOptions = authenticator && authenticator.options;
+      // OV is a special case, so process OV options
+      authenticatorOptions = [...authenticatorOptions]; //clone it since we are changing it for OV
+      createOVOptions(authenticatorOptions);
+      return authenticatorOptions.length > 1;
+    }
+    return false;
   },
 
   getActionByPath(actionPath) {
--- a/src/v2/models/AppState.js
+++ b/src/v2/models/AppState.js
@@ -78,7 +78,7 @@ export default Model.extend({
     return this.get('idx').neededToProceed.find((remediation) => remediation.name === formName);
   },
 
-  hasMoreThanOneAuthenticatorOption(formName) {
+  getRemediationAuthenticationOptions(formName) {
     const form = this.hasRemediationObject(formName);
     if (!form) {
       return false;
@@ -88,7 +88,7 @@ export default Model.extend({
     // OV is a special case, so process OV options
     authenticatorOptions = [...authenticatorOptions]; //clone it since we are changing it for OV
     createOVOptions(authenticatorOptions);
-    return authenticatorOptions.length > 1;
+    return authenticatorOptions;
   },
 
   getActionByPath(actionPath) {
--- a/src/v2/models/AppState.js
+++ b/src/v2/models/AppState.js
@@ -80,15 +80,15 @@ export default Model.extend({
 
   hasMoreThanOneAuthenticatorOption(formName) {
     const form = this.hasRemediationObject(formName);
-    if (form) {
-      const authenticator = form.value.find((value) => value.name === 'authenticator');
-      let authenticatorOptions = authenticator && authenticator.options;
-      // OV is a special case, so process OV options
-      authenticatorOptions = [...authenticatorOptions]; //clone it since we are changing it for OV
-      createOVOptions(authenticatorOptions);
-      return authenticatorOptions.length > 1;
+    if (!form) {
+      return false;
     }
-    return false;
+    const authenticator = form.value.find((value) => value.name === 'authenticator');
+    let authenticatorOptions = authenticator?.options || [];
+    // OV is a special case, so process OV options
+    authenticatorOptions = [...authenticatorOptions]; //clone it since we are changing it for OV
+    createOVOptions(authenticatorOptions);
+    return authenticatorOptions.length > 1;
   },
 
   getActionByPath(actionPath) {
--- a/src/v2/models/AppState.js
+++ b/src/v2/models/AppState.js
@@ -81,7 +81,7 @@ export default Model.extend({
   getRemediationAuthenticationOptions(formName) {
     const form = this.hasRemediationObject(formName);
     if (!form) {
-      return false;
+      return [];
     }
     const authenticator = form.value.find((value) => value.name === 'authenticator');
     let authenticatorOptions = authenticator?.options || [];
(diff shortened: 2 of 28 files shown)


src/v2/ion/ui-schema/ion-object-handler.js
```diff
@@ -18,7 +18,7 @@ import { AUTHENTICATOR_KEY } from '../RemediationConstants';
  * @param {AuthenticatorOption[]} options
  * @param {( AuthenticatorEnrollment[] || Authenticator[] )} authenticators
  */
-const createOVOptions = (options = []) => {
+export const createOVOptions = (options = []) => {
```
anipendakur-okta, line 21: Since you're changing this anyway, can we change it to a function instead of a const. -
```
export function createOVOptions(options = []) {
```

anipendakur-okta: It might be in our best interest to change other exports here too. There's no need for them to be consts as they all are functions. And exporting individual functions supports tree-shaking better (if we were to tree-shake soon.)

src/v2/models/AppState.js
```diff
@@ -77,6 +78,19 @@ export default Model.extend({
...
   },
 
+  hasMoreThanOneAuthenticatorOption(formName) {
+    const form = this.hasRemediationObject(formName);
+    if (form) {
+      const authenticator = form.value.find((value) => value.name === 'authenticator');
+      let authenticatorOptions = authenticator && authenticator.options;
+      // OV is a special case, so process OV options
+      authenticatorOptions = [...authenticatorOptions]; //clone it since we are changing it for OV
+      createOVOptions(authenticatorOptions);
```
anipendakur-okta, line 88: How about some more ES6 love?
```
let authenticatorOptions = authenticator?.options || [];
authenticatorOptions = [...authenticatorOptions];
```

OR

```
const authenticatorOptions = [...(authenticator?.options || [])];
```

src/v2/models/AppState.js
```diff
@@ -77,6 +78,19 @@ export default Model.extend({
     return this.get('idx').neededToProceed.find((remediation) => remediation.name === formName);
   },
 
+  hasMoreThanOneAuthenticatorOption(formName) {
```
jmelberg-okta, line 81: name-nit: There may be future cases where we'd like to get the total number of authenticator options. In the spirit of generic function signatures, how about renaming this and changing the return type to an array?

```suggestion
  getAuthenticatorOptions(formName) {
```

src/v2/models/AppState.js
```diff
@@ -77,6 +78,19 @@ export default Model.extend({
...
+    const form = this.hasRemediationObject(formName);
+    if (form) {
```
jmelberg-okta, line 83: nit: To avoid a large conditional block, consider returning when the following isn't met:

```js
if (!form || !form.value.find(value => value.name === 'authenticator')) {
  // return
}
```

src/v2/view-builder/utils/LinksUtil.js
```diff
@@ -4,7 +4,8 @@ import { ACTIONS, FORMS as RemediationForms } from '../../ion/RemediationConstan
 const { ENROLLED_PASSWORD_RECOVERY_LINK, ORG_PASSWORD_RECOVERY_LINK } = ACTIONS;
 
 const getSwitchAuthenticatorLink = (appState) => {
-  if (appState.hasRemediationObject(RemediationForms.SELECT_AUTHENTICATOR_AUTHENTICATE)) {
+  if (appState.hasRemediationObject(RemediationForms.SELECT_AUTHENTICATOR_AUTHENTICATE) &&
+    appState.hasMoreThanOneAuthenticatorOption(RemediationForms.SELECT_AUTHENTICATOR_AUTHENTICATE)) {
```
jmelberg-okta, line 8: Since the new method `hasMoreThanOneAuthenticatorOption` checks if this form name exists in the remediation object, we should be able to condense this here and below to the latter part of the conditional.

test/testcafe/spec/ChallengeAuthenticatorDuo_spec.js
```diff
@@ -53,11 +53,7 @@ test
   .requestHooks(mock)('renders an iframe for Duo', async t => {
     const challengeDuoPage = await setup(t);
 
-    const { log } = await t.getBrowserConsoleMessages();
-    await t.expect(log.length).eql(3);
-    await t.expect(log[0]).eql('===== playground widget ready event received =====');
-    await t.expect(log[1]).eql('===== playground widget afterRender event received =====');
-    await t.expect(JSON.parse(log[2])).eql({
```
jmelberg-okta, line 60: 👏🏼

test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
```diff
@@ -143,6 +143,11 @@ test
     await t.expect(pageTitle).contains('Verify with your email');
     await t.expect(challengeEmailPageObject.getFormSubtitle())
       .contains('Check your email for a verification message. Click the verification button in your email or enter the code below to continue.');
+
+    // Verify links (switch authenticator link not present since there are no other authenticators available)
+    await t.expect(await challengeEmailPageObject.switchAuthenticatorLinkExists()).notOk();
```
jmelberg-okta, line 148: Probably OK - but if we modify the mocks at any time we may lose out on this test coverage if responses no longer have a single authenticator present.

For future safety - can we modify the `validOTPmock` above to remove **all but one** authenticator as a separate test?

test/unit/spec/v2/view-builder/utils/LinksUtil_spec.js
```diff
@@ -0,0 +1,43 @@
+import AppState from 'v2/models/AppState';
+import { FORMS } from 'v2/ion/RemediationConstants';
+import { getSwitchAuthenticatorLink } from 'v2/view-builder/utils/LinksUtil';
+
+describe('v2/utils/LinksUtil', function() {
+  const mockAppState = (remediationFormName, hasMoreThanOneAuthenticator) => {
+    const appState = new AppState();
+    spyOn(appState, 'hasRemediationObject').and.callFake(formName => formName === remediationFormName);
+    spyOn(appState, 'hasMoreThanOneAuthenticatorOption').and.returnValue(hasMoreThanOneAuthenticator);
```
jmelberg-okta, line 9: Let's make sure we use `jest` spies instead of `jasmine`:

```suggestion
    jest.spyOn(appState, 'hasRemediationObject').mockImplementation(formName => formName === remediationFormName);
    jest.spyOn(appState, 'hasMoreThanOneAuthenticatorOption').mockReturnValue(hasMoreThanOneAuthenticator);
```

jmelberg-okta (commented):
Great work! Couple of suggestions.

test/unit/spec/v2/view-builder/utils/LinksUtil_spec.js
TianGan-okta: Bye jasmine

gowthamidommety-okta:
Still reviewing.. question I have is, what will this look like with `ChallengeAuthenticatorDataOktaVerifyView`? This is the view to render OV options when it is the only authenticator

src/v2/models/AppState.js
mauriciocastillosilva-okta: We don't have the use case for this yet, so I would prefer to keep it as returning a boolean. Once we need this, it will be very easy to refactor this function to support both `hasMoreThanOneAuthenticatorOption` and `getAuthenticatorOptions `

test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
mauriciocastillosilva-okta: I don't think we change mocks that often, so I was thinking that we can rely on that mock. If we change it in the future, this test will fail, and I am hoping at that point the developer making the change will do whatever needed to keep the test coverage good. 
PS All the tests I am changing are relying on the mocks as they are today. 

Does this sound good? I can add the test too, but I think we are covered with what we have

mauriciocastillosilva-okta:
@gowthamidommety-okta if OV is the only authenticator we look at the methods that are available. This is what `createOVOptions` function does. So anytime, OV been the single authenticator or not, OV will be counted depending how many methods are available

src/v2/ion/ui-schema/ion-object-handler.js
```diff
@@ -147,6 +147,9 @@ const createUiSchemaForObject = (ionFormField, remediationForm, transformedResp,
...
 
   return uiSchema;
-};
+}
 
-export default createUiSchemaForObject;
+export {
+  createUiSchemaForObject as default,
+  createOVOptions
+};
```
anipendakur-okta, line 155: Exporting an object will actually import the whole object into the bundle as compared to individual exports. In order to support tree-shaking we would need something like
```
export function createOVOptions() {}
.
.
.
export default createUiSchemaForObject;
```

src/v2/ion/ui-schema/ion-object-handler.js
```diff
@@ -18,7 +18,7 @@ import { AUTHENTICATOR_KEY } from '../RemediationConstants';
  * @param {AuthenticatorOption[]} options
  * @param {( AuthenticatorEnrollment[] || Authenticator[] )} authenticators
  */
-const createOVOptions = (options = []) => {
+function createOVOptions(options = []) {
```
pradeepdewda-okta, line 21: why this change ??

src/v2/ion/ui-schema/ion-object-handler.js
pradeepdewda-okta: `export default` make sense..

src/v2/models/AppState.js
pradeepdewda-okta: we can make this function more generic right away..

hasRequiredAuthenticators(formName, count) {
    .
    .
   return authenticatorOptions.length > count;
}

test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js
```diff
@@ -143,6 +143,11 @@ test
...
+    await t.expect(await challengeEmailPageObject.signoutLinkExists()).ok();
+    await t.expect(challengeEmailPageObject.getSignoutLinkText()).eql('Back to sign in');
```
pradeepdewda-okta, line 150: There could two types of `Back to sign in` links..  `js-cancel` and `js-goback` . If `cancel` object is not present `js-goback` will be rendered.. just fyi

test/testcafe/spec/EnrollAuthenticatorPhoneView_spec.js
```diff
@@ -83,6 +85,9 @@ test
     await t.expect(pageSubtitle).contains('your phone');
     await t.expect(pageSubtitle).contains('Enter the code below to verify.');
 
+    // Verify links
```
pradeepdewda-okta, line 88: dups ?

test/unit/spec/v2/models/AppState_spec.js
```diff
@@ -180,4 +185,130 @@ describe('v2/models/AppState', function() {
     });
 
   });
+
+  describe('hasRemediationObject', () => {
+    it('returns null if there form is not present', (done) => {
```
pradeepdewda-okta, line 190: `if there form` ??

test/unit/spec/v2/view-builder/utils/LinksUtil_spec.js
```diff
@@ -0,0 +1,37 @@
+import AppState from 'v2/models/AppState';
```
pradeepdewda-okta, line 1: 🥇

src/v2/ion/ui-schema/ion-object-handler.js
mauriciocastillosilva-okta: if you scroll up the comments in the PR, Ani pointed out why https://github.com/okta/okta-signin-widget/pull/2006#discussion_r645057750

src/v2/models/AppState.js
mauriciocastillosilva-okta: ok, I got a couple of comments on this function, so I will refactor it. I prefer Jordan's proposal, which is to return the options objects and let the consumer of the function decide what to do with them (either look at the length or iterate through them). Will make the change

src/v2/models/AppState.js
```diff
@@ -77,6 +78,19 @@ export default Model.extend({
     return this.get('idx').neededToProceed.find((remediation) => remediation.name === formName);
   },
 
+  getRemediationAuthenticationOptions(formName) {
+    const form = this.hasRemediationObject(formName);
+    if (!form) {
+      return false;
```
jmelberg-okta, line 84: Now that this isn't returning a `boolean`, let's return an empty array here to ensure a uniform type.

anipendakur-okta (approved):
LGTM

test/testcafe/spec/ChallengeOktaVerifyTotp_spec.js
```diff
@@ -52,6 +52,12 @@ test
     await t.expect(await challengeOktaVerifyTOTPPageObject.getTotpLabel())
       .contains('Enter code from Okta Verify app');
 
+    // Verify links
+    await t.expect(await challengeOktaVerifyTOTPPageObject.switchAuthenticatorLinkExists()).ok();
+    await t.expect(challengeOktaVerifyTOTPPageObject.getSwitchAuthenticatorLinkText()).eql('Verify with something else');
+    await t.expect(await challengeOktaVerifyTOTPPageObject.signoutLinkExists()).ok();
+    await t.expect(challengeOktaVerifyTOTPPageObject.getSignoutLinkText()).eql('Back to sign in');
```
jmelberg-okta, line 59: Looking at this again, I think we have an opportunity to condense these tests. I created a follow-up for us to look into breaking out the footer checking/content logic out to a shared assertion ([OKTA-402002](https://oktainc.atlassian.net/browse/OKTA-402002)).

test/testcafe/spec/EnrollAuthenticatorGoogleAuthenticator_spec.js
```diff
@@ -117,6 +122,10 @@ test
     await t.expect(enrollGoogleAuthenticatorPageObject.getSharedSecret()).eql('ZR74DHZTG43NBULV');
     await enrollGoogleAuthenticatorPageObject.goToNextPage();
 
+    // Verify links (switch authenticator link not present since there are no other authenticators available)
```
jmelberg-okta, line 125: These comments are extremely helpful as reviewers walk through the code. Thank you!

test/unit/spec/v2/view-builder/utils/LinksUtil_spec.js
```diff
@@ -0,0 +1,42 @@
...
+import { FORMS } from 'v2/ion/RemediationConstants';
+import { getSwitchAuthenticatorLink } from 'v2/view-builder/utils/LinksUtil';
+
+describe('v2/utils/LinksUtil', function() {
+  const mockAppState = (remediationFormName, hasMoreThanOneAuthenticator) => {
+    const appState = new AppState();
+    spyOn(appState, 'getRemediationAuthenticationOptions').and.callFake(formName => {
```
jmelberg-okta, line 8: Similar to the other changes, let's make sure there are no `spyOn` methods being called that aren't methods on `jest`. This helps us eventually migrate to future major versions of Jest, [which plans](https://jestjs.io/blog/2020/05/05/jest-26#a-new-way-to-consume-jest---jestglobals) to eventually drop these as a part of their global objects.

```suggestion
    jest.spyOn(appState, 'getRemediationAuthenticationOptions').mockImplementation(formName => {
```

jmelberg-okta (approved):
Solid work! 💪🏼 

Couple of minor comments + changes.

test/testcafe/spec/ChallengeOktaVerifyTotp_spec.js
```diff
@@ -52,6 +52,12 @@ test
     await t.expect(await challengeOktaVerifyTOTPPageObject.getTotpLabel())
       .contains('Enter code from Okta Verify app');
 
+    // Verify links
+    await t.expect(await challengeOktaVerifyTOTPPageObject.switchAuthenticatorLinkExists()).ok();
+    await t.expect(challengeOktaVerifyTOTPPageObject.getSwitchAuthenticatorLinkText()).eql('Verify with something else');
```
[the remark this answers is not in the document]
mauriciocastillosilva-okta: You know, this is a very good question. I don't know either why we check for both, but I have also seen it in different places so I copied the pattern. My only guess is that we do this in case there is more than one link and one of them is empty? Which is really unlikely... would it be ok if we clean this up in the jira you filed? [OKTA-402002](https://oktainc.atlassian.net/browse/OKTA-402002) (which I know is for the signout link... or I can file another jira too)

gowthamidommety-okta (approved):
Sorry for the late comment. Everything looks good, but I'm still concerned about OV when it is the only authenticator. Can we add tests here https://github.com/okta/okta-signin-widget/blob/master/test/testcafe/spec/ChallengeAuthenticatorOktaVerify_spec.js#L5 to make sure it works as expected?

test/testcafe/spec/ChallengeAuthenticatorOktaVerify_spec.js
```diff
@@ -128,3 +137,36 @@ test.requestHooks(requestLogger, mockChallengeOVSelectMethod)('should send right
     'stateHandle': '02ciZ1YTWakSanNu8GNNTnTXzhL5hoLzTlR0JewfG3'
   });
 });
+
+test.requestHooks(requestLogger, mockChallengeOVTotpMethod)('should show switch authenticator link only when needed', async t => {
+  const selectAuthenticatorPage = await setup(t);
+  await t.expect(selectAuthenticatorPage.getFormTitle()).eql('Verify it\'s you with an authenticator');
+  await t.expect(selectAuthenticatorPage.getFormSubtitle()).eql('Select from the following options');
+
+  // This view (only OV authenticator, but with 3 methods) is custom
```
gowthamidommety-okta, line 146: 🙌

gowthamidommety-okta (approved):
Looks great! Thank you so much for checking 🙏

PR #2020, opened 2021-06-05
OKTA-395319 Update cancel polling link in terminal view

yannongli-okta:
## Description:
- Issue: When loopback succeeds but SIW receives terminal state like `user is not assigned to this application`, SIW already stops polling and server knows it. If we click on `Cancel and take me to sign in`, it will make an API call to /poll/cancel which doesn't make sense and server keeps returning same terminal response. User is stuck in the page.
- Fix: In device challenge poll view, when we receive terminal state from backend, the footer link will be then updated with new text and url. This will make a call to /idp/idx/cancel and user is able to get back to main loginpage.
- Note: This does not apply to use cases in factor challenge view, since there is already a `Back to Sign In` link in that page.

## PR Checklist

- [ ] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:
- Demo: https://okta.box.com/s/mfjcfhu6tu1ju5k6b9hfq05xfp3fkl25

### Reviewers:
@okta/device-platform @pradeepdewda-okta 

### Issue:

- [OKTA-395319](https://oktainc.atlassian.net/browse/OKTA-395319)


YangChen-okta:
Since we agreed this is just a temp solution to meet the 6/11 deadline, the fix looks good enough for me to move on.

In the long run, I think we should do the following fixes:
1. on server side, remove "launch OV" remediation from response as it doesn't make sense when device user is not assigned to the app.
2. make SIW "dummy" and consistent, i.e. simply follow the remediation to present the "identify" view with errors on top.  "cancel" or "back/take to ..." seems not necessary.

pradeepdewda-okta:
> "cancel" or "back/take to ..." seems not necessary.
There needs to be some action for user to take on terminal screens, until and unless it's totally irrelevant for some flows.

YangChen-okta:
In this specific case, the user cannot do anything except maybe switch to a different user.  When "app not  assigned" error happens with existing session, SIW presents the login page with username, just to make it consistent.

pradeepdewda-okta:
> In this specific case, the user cannot do anything except maybe switch to a different user. When "app not assigned" error happens with existing session, SIW presents the login page with username, just to make it consistent.

Then may be we should add to the error message that.. `You can close this window and try with a different user` involve UX/PM to make this clarification.

santoshmale-okta (approved):
LGTM!

santoshmale-okta:
I am not sure if this is the only scenario in this bucket. The thing to note is that whenever the server is denying access and returning an error message, it rolls back all contexts in the state token and adds an identification transaction here https://github.com/okta/okta-core/blob/master/components/platform/idx/api/src/main/java/com/okta/platform/services/idx/pipeline/BaseIdxPipelineStepHandler.java#L504.  This would always result in the server returning back an error message and the remediation corresponding to the new identification transaction that was added. The SIW just treats the error message as a terminal message and displays the message on an already existing screen (which makes sense in almost all of the cases). This is one of the corner cases where we were running into an issue as the transaction corresponding to the previous page no longer exists. If there are any other scenarios where the already existing screens refer to the transactions that existed before the rollback, we would run into the issue in those places. But, I doubt if there are any other cases. This solution LGTM (even for the longer term).

src/v2/models/AppState.js
```diff
@@ -33,6 +33,7 @@ export default Model.extend({
     remediations: 'array',
     dynamicRefreshInterval: 'number',
     deviceFingerprint: 'string',
+    updatePollingCancelLink: 'boolean',
```
pradeepdewda-okta, line 36: change this too `updateFooterLink`

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -39,6 +40,9 @@ const Body = BaseFormWithPolling.extend(
     onPollingFail() {
       this.$('.spinner').hide();
       this.stopPolling();
+      // When SIW receives form error, polling is already stopped
+      // SIW needs to update footer link from /poll/cancel to /idp/idx/cancel
+      this.options.appState.set('updatePollingCancelLink', true);
```
pradeepdewda-okta, line 45: when polling starts again do we need to update this to `false` ?

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -171,7 +172,27 @@ const Footer = BaseFooter.extend({
         }
       });
     }
-  }
+  },
+
+  updateLink() {
```
pradeepdewda-okta, line 177: change to `updateFooterLinks`

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -264,6 +290,38 @@ test
     await t.expect(loopbackSuccessForChallengeTimeoutLogger.contains(record => record.request.url.match(/\/idp\/idx\/authenticators\/poll\/cancel/))).eql(false);
   });
 
+test
+  .requestHooks(loopbackSuccessLogger, loopbackSuccessButNotAssignedAppMock)('loopback succeeds but user is not assigned to app, then clicks cancel link', async t => {
+    const deviceChallengePollPageObject = await setup(t);
+    await t.expect(deviceChallengePollPageObject.getBeaconClass()).contains(BEACON_CLASS);
```
pradeepdewda-okta, line 296: you don't need to repeat checks that are tested already in previous tests.. keep the test short specific to what it intends.

src/v2/models/AppState.js
pradeepdewda-okta: instead of this.. did you explore triggering an event from `onPollingFail` ??

this.trigger('updateFooterLink');

yannongli-okta:
@pradeepdewda-okta Updated code to use event instead of variable. Please take a look again

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -39,6 +40,9 @@ const Body = BaseFormWithPolling.extend(
     onPollingFail() {
       this.$('.spinner').hide();
       this.stopPolling();
+      // When SIW receives form error, polling is already stopped
+      // SIW needs to update footer link from /poll/cancel to /idp/idx/cancel
+      this.options.appState.trigger('updateFooterLink', '');
```
pradeepdewda-okta, line 45: you can pass label as data here
```
const data = { label: loc('loopback.polling.cancel.link.with.form.error', 'login') }
```

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -171,7 +172,27 @@ const Footer = BaseFooter.extend({
...
-  }
+  },
+
+  handleUpdateFooterLink() {
+    // only update link for loopback
+    if (!this.isFallbackApproach()) {
+      const link = getSignOutLink(this.options.settings, {
+        label: loc('loopback.polling.cancel.link.with.form.error', 'login')
+      })[0];
+      this.removeChildren();
```
pradeepdewda-okta, line 183: in future if this view contains some other link then this can cause issue.. See if you can remove only what you want to replace?

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -171,7 +173,24 @@ const Footer = BaseFooter.extend({
         }
       });
     }
-  }
+  },
+
+  handleUpdateFooterLink(data) {
+    // only update link for loopback
+    if (!this.isFallbackApproach()) {
+      this.$('.js-cancel-authenticator-challenge').remove();
```
pradeepdewda-okta, line 181: this.$el.find(`.js-cancel-authenticator-challenge`)

pradeepdewda-okta (approved):
Looks good to me..

jchan-okta:
@yannongli-okta @sshen-okta @santoshmale-okta  did we decide to change "Cancel and take me to sign in" to "Take me to sign in"? Is it from PM or UX input?

Also, how come the url in the browser input field doesn't get updated after clicking "Take me to sign in" based on the video https://okta.box.com/s/mfjcfhu6tu1ju5k6b9hfq05xfp3fkl25 ?

yannongli-okta:
> @yannongli-okta @sshen-okta @santoshmale-okta did we decide to change "Cancel and take me to sign in" to "Take me to sign in"? Is it from PM or UX input?
> 
> Also, how come the url in the browser input field doesn't get updated after clicking "Take me to sign in" based on the video https://okta.box.com/s/mfjcfhu6tu1ju5k6b9hfq05xfp3fkl25 ?

@jchan-okta Yep this is decision made by Tanvir and our team had a discussion on it. I double checked use case in custom uri, and clicking on `Back to Sign In` also doesn't change the browser url. Did you see different behaviors somewhere else?

jchan-okta:
> > @yannongli-okta @sshen-okta @santoshmale-okta did we decide to change "Cancel and take me to sign in" to "Take me to sign in"? Is it from PM or UX input?
> > Also, how come the url in the browser input field doesn't get updated after clicking "Take me to sign in" based on the video https://okta.box.com/s/mfjcfhu6tu1ju5k6b9hfq05xfp3fkl25 ?
> 
> @jchan-okta Yep this is decision made by Tanvir and our team had a discussion on it. The main reason is we didn't COME from Sign in page so there is no BACK to sign in.
> 
> And I double checked use case in custom uri, and clicking on `Back to Sign In` also doesn't change the browser url. Did you see different behaviors somewhere else?

My bad, clicking "Take me to sign in" or "Cancel and take me to sign in" or "Back to sign in" will just cancel the polling, the url in the browser url field stays the same, it is same as before.

yannongli-okta:
Closing this PR in favor of https://github.com/okta/okta-signin-widget/pull/2039

PR #2029, opened 2021-06-07
Fixing eng leak with password reset error

kamaldeepcheema-okta:
## Description:

- Fixing eng leak with the password reset error. We had an existing error message defined for this, we just had to map it appropriately.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
image

### Issue:

- [OKTA-401412](https://oktainc.atlassian.net/browse/OKTA-401412)

### DIFF
--- a/src/v2/ion/i18nTransformer.js
+++ b/src/v2/ion/i18nTransformer.js
@@ -155,6 +155,7 @@ const I18N_OVERRIDE_MAPPINGS = {
   'api.users.auth.error.POST_PASSWORD_UPDATE_AUTH_FAILURE': 'oie.post.password.update.auth.failure.error',
   'security.access_denied': 'errors.E0000006',
   'api.factors.error.sms.invalid_phone': 'oie.phone.invalid',
+  'app.ldap.password.reset.failed': 'errors.E0000017'
 };
 
 const I18N_PARAMS_MAPPING = {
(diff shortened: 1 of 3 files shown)


src/v2/ion/i18nTransformer.js
```diff
@@ -155,6 +155,7 @@ const I18N_OVERRIDE_MAPPINGS = {
   'api.users.auth.error.POST_PASSWORD_UPDATE_AUTH_FAILURE': 'oie.post.password.update.auth.failure.error',
   'security.access_denied': 'errors.E0000006',
   'api.factors.error.sms.invalid_phone': 'oie.phone.invalid',
+  'app.ldap.password.reset.failed': 'errors.E0000017'
```
nikhilvenkatraman-okta, line 158: Do we know if this key is sent back on for OIE or is it for V1 as well ? Asking because we try to differentiate by adding the `oie.` prefix for OIE only keys. We can leave it as is, if it is used in V1 as well.

kamaldeepcheema-okta: @nikhilvenkatraman-okta according to this line: https://github.com/okta/okta-core/blob/master/components/apps/directories/impl/src/main/java/com/saasure/application/ldap_sun_one/auth/SunONELoginHandlerImpl.java#L1263, it looks like this is used in V1 as well.

nikhilvenkatraman-okta (approved):

PR #2039, opened 2021-06-09
fix: Update cancel polling link when receiving terminal response

yannongli-okta:
Resolves: OKTA-395319

## Description:
Issue: When loopback succeeds but SIW receives terminal state like user is not assigned to this application, SIW already stops polling and server knows it. If we click on Cancel and take me to sign in, it will make an API call to /poll/cancel which doesn't make sense and server keeps returning same terminal response. User is stuck in the page.

Fix: In device challenge poll view, when we receive terminal state from backend, the footer link will be then updated with new text and url. This will make a call to /idp/idx/cancel and user is able to get back to main loginpage.

Note: This does not apply to use cases in factor challenge view, since there is already a Back to Sign In link in that page.

## PR Checklist

- [ ] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:
Demo: https://okta.box.com/s/mfjcfhu6tu1ju5k6b9hfq05xfp3fkl25

### Reviewers:
@okta/device-platform @aarongranick-okta 

### Issue:

- [OKTA-395319](https://oktainc.atlassian.net/browse/OKTA-395319)


yannongli-okta:
This PR is the same as https://github.com/okta/okta-signin-widget/pull/2020 but is intended to merge into 5.7 branch.

aarongranick-okta (approved):

PR #2040, opened 2021-06-09
Support idpDisplay config for OIE flow

kamaldeepcheema-okta:
## Description:

- Currently we don't support the idpDisplay config for OIE. For parity with v1, we add support for this. If idpDisplay is PRIMARY, the idp buttons are rendered above the main form, otherwise they're rendered below the main form if SECONDARY.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/77295104/121367849-a203a000-c908-11eb-8d91-400a53ac5059.mov

### Issue:

- [OKTA-398647](https://oktainc.atlassian.net/browse/OKTA-398647)

### DIFF
--- a/src/v2/view-builder/views/IdentifierView.js
+++ b/src/v2/view-builder/views/IdentifierView.js
@@ -74,13 +74,30 @@ const Body = BaseForm.extend({
         });
       }
 
-      this.add(signInWithIdps, {
-        selector: '.o-form-button-bar',
-        options: {
-          idpButtons,
-          addSeparateLine: true,
-        }
-      });
+      // We check the 'idpDisplay' option config to determine whether to render the idp buttons 
+      // above or below the form
+      const idpDisplay = this.options.settings.get('idpDisplay');
+      if (idpDisplay.toUpperCase() === 'PRIMARY') {
+        // Render idp buttons above the form
+        this.add(signInWithIdps, {
+          prepend: true,
+          selector: '.o-form-fieldset-container',
+          options: {
+            idpButtons,
+            addSeparateLine: true,
+            addSeparateLineAtBottom: true
+          }
+        });
+      } else {
+        // Render idp buttons below the form
+        this.add(signInWithIdps, {
+          selector: '.o-form-button-bar',
+          options: {
+            idpButtons,
+            addSeparateLine: true,
+          }
+        });
+      }
     }
 
     const customButtons = createCustomButtons(this.options.settings);
--- a/src/v2/view-builder/views/IdentifierView.js
+++ b/src/v2/view-builder/views/IdentifierView.js
@@ -60,33 +60,12 @@ const Body = BaseForm.extend({
       this.add(signInWithDeviceOption, '.o-form-fieldset-container', false, true, { isRequired: false });
     }
 
-    // add external idps buttons
+    // add forgot password link and external idps buttons if needed
     const idpButtons = createIdpButtons(this.options.appState.get('remediations'));
     if (Array.isArray(idpButtons) && idpButtons.length) {
-
       // Add the forgot password link before the buttons for multiple IDPs
-      const forgotPasswordLink = getForgotPasswordLink(this.options.appState, this.options.settings);
-      if (forgotPasswordLink.length) {
-        this.add('<div class="links-primary"></div>', { selector: '.o-form-button-bar' });
-        this.add(Link, {
-          selector: '.links-primary',
-          options: forgotPasswordLink[0],
-        });
-      }
-
-      // We check the 'idpDisplay' option config to determine whether to render the idp buttons 
-      // above or below the login fields
-      const idpDisplay = this.options.settings.get('idpDisplay');
-      const isPrimaryIdpDisplay = idpDisplay && idpDisplay.toUpperCase() === 'PRIMARY';
-
-      this.add(signInWithIdps, {
-        prepend: isPrimaryIdpDisplay,
-        selector: isPrimaryIdpDisplay ? '.o-form-fieldset-container' : '.o-form-button-bar',
-        options: {
-          idpButtons,
-          isPrimaryIdpDisplay
-        }
-      });
+      this._addForgotPasswordView();
+      this._addIdpView(idpButtons);
     }
 
     const customButtons = createCustomButtons(this.options.settings);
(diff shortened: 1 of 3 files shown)


src/v2/view-builder/views/IdentifierView.js
```diff
@@ -74,11 +74,18 @@ const Body = BaseForm.extend({
         });
       }
 
+      // We check the 'idpDisplay' option config to determine whether to render the idp buttons 
+      // above or below the form
```
pradeepdewda-okta, line 78: `above or below login fields`

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -74,11 +74,18 @@ const Body = BaseForm.extend({
...
+      const idpDisplay = this.options.settings.get('idpDisplay');
+      const isPrimaryDisplay = idpDisplay.toUpperCase() === 'PRIMARY';
```
pradeepdewda-okta, line 80: check if `idpDisplay` exists first

src/v2/view-builder/views/signin/SignInWithIdps.js
```diff
@@ -4,10 +4,13 @@ import hbs from 'handlebars-inline-precompile';
 export default View.extend({
   className: 'sign-in-with-idp',
   template: hbs`
-    {{#if addSeparateLine}}
+    {{#if addSeparateLineAtTop}}
```
pradeepdewda-okta, line 7: since we got a chance to refactor.. can you please rename this to `addDividerTop` and `addDividerBottom` 
`class="separation-line"` to `class="form-divider"`

src/v2/view-builder/views/signin/SignInWithIdps.js
```diff
@@ -19,7 +22,8 @@ export default View.extend({
     const jsonData = View.prototype.getTemplateData.apply(this, arguments);
 
     return Object.assign(jsonData, {
-      addSeparateLine: this.options.addSeparateLine,
+      addSeparateLineAtTop: this.options.addSeparateLine && !this.options.addSeparateLineAtBottom,
```
pradeepdewda-okta, line 25: instead of this flag `this.options.addSeparateLine` we can just use idpButtons value ? For eg.. if buttons exists then we have to add the divider..

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -74,11 +74,18 @@ const Body = BaseForm.extend({
...
+
       this.add(signInWithIdps, {
-        selector: '.o-form-button-bar',
+        prepend: isPrimaryDisplay,
+        selector: isPrimaryDisplay ? '.o-form-fieldset-container' : '.o-form-button-bar',
         options: {
           idpButtons,
           addSeparateLine: true,
+          addSeparateLineAtBottom: isPrimaryDisplay
```
pradeepdewda-okta, line 88: we don't need another variable.. we can just pass `isPrimaryIdpDisplay`

src/v2/view-builder/views/signin/SignInWithIdps.js
pradeepdewda-okta: we can simplify this too..  If it's not `isPrimaryIdpDisplay` then obviously it's default.  We don't need these two variables.

jsonData.isPrimaryIdpDisplay = this.option. isPrimaryIdpDisplay
return jsonData;

test/unit/spec/v2/view-builder/views/IdentifierView_spec.js
```diff
@@ -49,4 +54,23 @@ describe('v2/view-builder/views/IdentifierView', function() {
...
+    jest.spyOn(AppState.prototype, 'getActionByPath').mockReturnValue(true);
+    jest.spyOn(AppState.prototype, 'isIdentifierOnlyView').mockReturnValue(false);
+    idpDisplay = 'PRIMARY';
+    testContext.init();
+
+    // The idp buttons should be rendered above the main form.
+    expect(testContext.view.$el.find('.o-form-fieldset-container .sign-in-with-idp').length).toEqual(1);
+    expect(testContext.view.$el.find('.o-form-button-bar .sign-in-with-idp').length).toEqual(0);
+
+    idpDisplay = 'SECONDARY';
```
pradeepdewda-okta, line 69: what if `idpDisplay` is undefined/null/or something garbage.

src/v2/view-builder/views/signin/SignInWithIdps.js
kamaldeepcheema-okta: I can update the arguments to `addDividerTop` and `addDividerBottom`, however updating the class is a bit more risky since it's referenced in other places: in RedirectIdPView.js and of course the _identify.css file. Also I noticed that `CustomButtons.js` and `SignInWithDeviceOption.js` follow the same pattern in using the `seperation-line` class. I don't think it makes sense to just update it for `SignInWithIdps.js` and not others. Lastly, `form-divider` is already a special class that has some css attached to it in _forms.scss, I don't think we want this css applied in our case here hence why we may have chosen a different class to begin with.

test/unit/spec/v2/view-builder/views/IdentifierView_spec.js
kamaldeepcheema-okta: I will add a case for this.

src/v2/view-builder/views/signin/SignInWithIdps.js
kamaldeepcheema-okta: sounds good, I'll simplify the arguments here

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -74,11 +74,17 @@ const Body = BaseForm.extend({
         });
       }
 
+      // We check the 'idpDisplay' option config to determine whether to render the idp buttons
```
pradeepdewda-okta, line 77: we should put this logic into it's own function now, since `render` is getting larger.. say `this.addIdpView()`
```
      // We check the 'idpDisplay' option config to determine whether to render the idp buttons 
      // above or below the login fields
      const idpDisplay = this.options.settings.get('idpDisplay');
      const isPrimaryIdpDisplay = idpDisplay && idpDisplay.toUpperCase() === 'PRIMARY';

      this.add(signInWithIdps, {
        prepend: isPrimaryIdpDisplay,
        selector: isPrimaryIdpDisplay ? '.o-form-fieldset-container' : '.o-form-button-bar',
        options: {
          idpButtons,
          isPrimaryIdpDisplay
        }
      });
    }
 ```

File: test/unit/spec/v2/view-builder/views/IdentifierView_spec.js [hunk 8]
```diff
@@ -49,4 +54,29 @@ describe('v2/view-builder/views/IdentifierView', function() {
...
     expect(testContext.view.$el.find('.links-primary .js-forgot-password').length).toEqual(0);
   });
+
+  it('view renders IDP buttons correctly with idpDisplay property', function() {
+    jest.spyOn(AppState.prototype, 'hasRemediationObject').mockReturnValue(true);
+    jest.spyOn(AppState.prototype, 'getActionByPath').mockReturnValue(true);
+    jest.spyOn(AppState.prototype, 'isIdentifierOnlyView').mockReturnValue(false);
+    testContext.init();
+
+    // The idp buttons should be rendered below the main login fields when no idpDisplay defined.
```
pradeepdewda-okta, line 64: also add a check that if there are no buttons configured then do not render any buttons..

src/v2/view-builder/views/IdentifierView.js
kamaldeepcheema-okta: yeah this is a good idea, let me make this change.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -60,27 +60,12 @@ const Body = BaseForm.extend({
...
-      }
-
-      this.add(signInWithIdps, {
-        selector: '.o-form-button-bar',
-        options: {
-          idpButtons,
-          addSeparateLine: true,
-        }
-      });
+      this._addForgotPasswordView();
```
pradeepdewda-okta, line 67: Nice..

pradeepdewda-okta (approved):

PR #2050, opened 2021-06-14
Clean up en leaks ignore list

nikhilvenkatraman-okta:
Resolves: OKTA-404422

## Description:

Clean up en leaks ignore list

## PR Checklist

### Reviewers:

@jmelberg-okta @aarongranick-okta 

### Issue:

- [OKTA-404422](https://oktainc.atlassian.net/browse/OKTA-404422)

### DIFF
--- a/src/v2/view-builder/utils/AuthenticatorUtil.js
+++ b/src/v2/view-builder/utils/AuthenticatorUtil.js
@@ -56,6 +56,7 @@ const getAuthenticatorData = function(authenticator, isVerifyAuthenticator) {
         ? authenticator.relatesTo?.profile?.phoneNumber
         : loc('oie.phone.authenticator.description', 'login'),
       iconClassName: 'mfa-okta-phone',
+      noTranslateClassName: isVerifyAuthenticator ? 'no-translate' : '',
       buttonDataSeAttr: getButtonDataSeAttr(authenticator),
     });
     break;
--- a/src/v2/view-builder/components/AuthenticatorVerifyOptions.js
+++ b/src/v2/view-builder/components/AuthenticatorVerifyOptions.js
@@ -22,7 +22,7 @@ const AuthenticatorRow = View.extend({
       <div>
         <h3 class="authenticator-label no-translate {{#if description}}authenticator-label--small{{/if}}">{{label}}</h3>
         {{#if description}}
-          <p class="authenticator-description--text">{{description}}</p>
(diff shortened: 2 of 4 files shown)


src/v2/view-builder/components/AuthenticatorEnrollOptions.js
```diff
@@ -24,7 +24,7 @@ const AuthenticatorRow = View.extend({
         <div class="authenticator-description">\
           <h3 class="authenticator-label no-translate">{{label}}</h3>\
           {{#if description}}\
-            <p class="authenticator-description--text">{{description}}</p>\
+            <p class="authenticator-description--text no-translate">{{description}}</p>\
```
jmelberg-okta, line 27: Shouldn't this `description` point to what is defined [here](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/utils/AuthenticatorUtil.js#L75-L77)? Seems like we **always** want these to be localized.

nikhilvenkatraman-okta: yeah, it is not just the authenticator description. In some cases we also have a phone number which is what I was trying to ignore in this case.

Screen Shot 2021-06-15 at 12 16 39 PM

nikhilvenkatraman-okta: Its for the phone number cases https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/utils/AuthenticatorUtil.js#L55

nikhilvenkatraman-okta: @jmelberg-okta updated it to add the no-translate className conditionally.

src/v2/view-builder/utils/AuthenticatorUtil.js
```diff
@@ -56,6 +56,7 @@ const getAuthenticatorData = function(authenticator, isVerifyAuthenticator) {
         ? authenticator.relatesTo?.profile?.phoneNumber
         : loc('oie.phone.authenticator.description', 'login'),
       iconClassName: 'mfa-okta-phone',
+      noTranslateClassName: isVerifyAuthenticator? 'no-translate': '',
```
jmelberg-okta, line 59: style-nit:

```suggestion
      noTranslateClassName: isVerifyAuthenticator ? 'no-translate' : '',
```

jmelberg-okta (approved):

PR #2055, opened 2021-06-16
fix: Generate fingerprint for IdP discovery flows

anishashetty-okta:
## Description:
When an org has an IDP set up with active IDP rules , the SIW goes through the IDP Discovery flow
When the `authn` call was made, we did not generate a device fingerprint here due to which new device Sign in email notifications were not being sent

When passwordless is ENABLED
image

When passwordless is disabled
image

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:
Attached above

### Reviewers:
@mauriciocastillosilva-okta 

### Issue:

- [OKTA-402247](https://oktainc.atlassian.net/browse/OKTA-402247)

### DIFF
--- a/src/IDPDiscoveryController.js
+++ b/src/IDPDiscoveryController.js
@@ -16,6 +16,7 @@ import PrimaryAuthModel from 'models/PrimaryAuth';
 import BaseLoginController from 'util/BaseLoginController';
 import IDPDiscoveryForm from 'views/idp-discovery/IDPDiscoveryForm';
 import CustomButtons from 'views/primary-auth/CustomButtons';
+import DeviceFingerprint from 'util/DeviceFingerprint';
 
 export default PrimaryAuthController.extend({
   className: 'idp-discovery',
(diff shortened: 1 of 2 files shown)


src/IDPDiscoveryController.js
```diff
@@ -60,25 +61,47 @@ export default PrimaryAuthController.extend({
...
       } else {
-        this.options.appState.set('disableUsername', true);
-        this.options.appState.trigger('navigate', 'signin');
+        self.doPrimaryAuth();
       }
     });
   },
+
+  doPrimaryAuth : function() {
+    this.options.appState.trigger('loading', false);
```
jmelberg-okta, line 85: Why do we need to change the loading state here now?

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -853,6 +855,58 @@ Expect.describe('IDPDiscovery', function() {
...
+          test.form.submit();
+          return Expect.waitForMfaVerify(test);
+        })
+        .then(function() {
+          expect(DeviceFingerprint.generateDeviceFingerprint).toHaveBeenCalled();
+          const ajaxArgs = Util.getAjaxRequest(0);
+          expect(ajaxArgs.requestHeaders['x-device-fingerprint']).toBe('thisIsTheDeviceFingerprint');
+        });
+    });
+    itp('renders primary auth with a device fingerprint when passwordless is disabled during idp discovery',
```
jmelberg-okta, line 881: Can we also add a test for when fingerprinting fails please?

jmelberg-okta (approved):
Changes overall LGTM! Couple comments.

Also - let's update our final commit text to address the typo:

```diff
-fix: Generate fngerprint for discovery flows
+fix: Generate fingerprint for IdP discovery flows
```

src/IDPDiscoveryController.js
mauriciocastillosilva-okta: same question

mauriciocastillosilva-okta (approved):
looks good. Just same question as Jordan

src/IDPDiscoveryController.js
anishashetty-okta: Thats my bad. I can take that out.

PR #2061, opened 2021-06-21
when stateToken changes between two remediations

pradeepdewda-okta:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:


src/v2/controllers/FormController.js
```diff
@@ -141,9 +147,12 @@ export default Controller.extend({
 
     if (idx['neededToProceed'].find(item => item.name === actionPath)) {
       idx.proceed(actionPath, {})
-        .then(this.handleIdxSuccess.bind(this))
+        .then(this.handleIdxResponse.bind(this))
```
pradeepdewda-okta, line 150: if Idx responded with either success or error.. for widget it doesn't matter if it was a success or failure. Previously, API errors were not formatted as idx responses.. https://github.com/okta/okta-idx-js/pull/66/files

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -101,15 +101,17 @@ const Body = BaseForm.extend({
...
-    const messagesObj = this.options.appState.get('messages');
-    if (messagesObj?.value.length
-        && this.options.appState.get('currentFormName') === 'identify') {
-      const content = messagesObj.value[0].message;
-      const messageCallout = createCallout({
-        content: content,
-        type: 'warning',
-      });
-      this.add(messageCallout, '.o-form-error-container');
+    if(!this.options.appState.get('idx')?.formError) {
```
pradeepdewda-okta, line 104: this avoids duplicated error rendering. If it's a `formError` then `FormController` will render it..

src/v2/models/AppState.js
```diff
@@ -155,11 +155,9 @@ export default Model.extend({
     const previousRawState = this.has('idx') ? this.get('idx').rawIdxState : null;
 
     const identicalResponse = _.isEqual(
-      _.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'refresh', 'stateHandle']),
-      _.nestedOmit(previousRawState, ['expiresAt', 'refresh', 'stateHandle']));
-    const isSameRefreshInterval = _.isEqual(
-      _.nestedOmit(transformedResponse.idx.rawIdxState, ['expiresAt', 'stateHandle']),
-      _.nestedOmit(previousRawState, ['expiresAt', 'stateHandle']));
```
haishengwu-okta, line 162: why those changes?

src/v2/BaseLoginRouter.js
```diff
@@ -116,7 +116,7 @@ export default Router.extend({
     // transform response
     const ionResponse = transformIdxResponse(this.settings, idxResponse, lastResponse);
 
-    this.appState.setIonResponse(ionResponse);
+    this.appState.setIonResponse(ionResponse, this.controller);
```
haishengwu-okta, line 119: Didn't feel good if we have to pass around previous `controller`. 
Maybe we miss something somewhere?

src/v2/models/AppState.js
pradeepdewda-okta: since stateHandle can change between the remediations we do now want to check for it, a response can not be identical if `stateHandle` changed

src/v2/BaseLoginRouter.js
pradeepdewda-okta: needs `controller` context to trigger errors from AppState, right before we change the app state.

src/v2/models/AppState.js
haishengwu-okta: Even stateHandle changed but remediation array isn't changed, I guess we don't need re-render, correct?

src/v2/BaseLoginRouter.js
pradeepdewda-okta: removed this.

src/v2/models/AppState.js
pradeepdewda-okta: That is right, but i am using this function to update the AppState and not to re-render, At least till we fix polling. that would be right after this PR.

Updating AppState without fixing polling is causing multiple polling views to fail..

test/unit/spec/v2/models/AppState_spec.js
```diff
@@ -118,7 +118,7 @@ describe('v2/models/AppState', function() {
   });
 
   describe('shouldReRenderView', () => {
-    it('rerender view should be false if stateHandle are different', () => {
+    it('rerender view should be true if stateHandle are different', () => {
```
haishengwu-okta, line 121: if only stateHandle is different, we don't expect to re-render (but only refresh `appState`), do we?

pradeepdewda-okta: i will remove this

PR #2064, opened 2021-06-23
Load Captcha lib based on SIW language

kamaldeepcheema-okta:
## Description:

- Previously the Captcha lib relied on the browser's locale to set its language, but the SIW language could differ from the browser locale. So now we drive the Captcha language based on the SIW language exclusively.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/77295104/123118276-7c9a8a00-d410-11eb-8b6f-a9b39233231a.mov

### Issue:

- [OKTA-406089](https://oktainc.atlassian.net/browse/OKTA-406089)

### DIFF
--- a/src/v2/view-builder/views/captcha/CaptchaView.js
+++ b/src/v2/view-builder/views/captcha/CaptchaView.js
@@ -24,7 +24,7 @@ const RECAPTCHAV2_URL =
   `https://www.google.com/recaptcha/api.js?onload=${OktaSignInWidgetOnCaptchaLoadedCallback}&render=explicit`;
 
 const getCaptchaUrl = (baseURL, locale) => {
-  return `${baseURL}&hl=${locale || 'en'}`;
+  return `${baseURL}&hl=${locale || navigator.language}`;
 };
 
 export default View.extend({
--- a/src/v2/view-builder/views/captcha/CaptchaView.js
+++ b/src/v2/view-builder/views/captcha/CaptchaView.js
@@ -23,6 +23,10 @@ const HCAPTCHA_URL =
 const RECAPTCHAV2_URL = 
   `https://www.google.com/recaptcha/api.js?onload=${OktaSignInWidgetOnCaptchaLoadedCallback}&render=explicit`;
 
+const getCaptchaUrl = (baseURL, locale) => {
+  return `${baseURL}&hl=${locale || 'en'}`;
+};
+
 export default View.extend({
   className: 'captcha-view',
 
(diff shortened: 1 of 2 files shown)


pradeepdewda-okta (approved):

src/v2/view-builder/views/captcha/CaptchaView.js
```diff
@@ -131,10 +135,11 @@ export default View.extend({
     // the 'data-callback' attribute which the Captcha library uses to invoke the callback.
     window[OktaSignInWidgetOnCaptchaSolvedCallback] = onCaptchaSolved;
 
+    const locale = this.options.settings.get('language');
```
nikhilvenkatraman-okta, line 138: nit: can we move this locale computation `const locale = this.options.settings.get('language');`  into `getCaptchaUrl` method itself ?

nikhilvenkatraman-okta (approved):
Just one comment

src/v2/view-builder/views/captcha/CaptchaView.js
```diff
@@ -23,6 +23,10 @@ const HCAPTCHA_URL =
 const RECAPTCHAV2_URL = 
   `https://www.google.com/recaptcha/api.js?onload=${OktaSignInWidgetOnCaptchaLoadedCallback}&render=explicit`;
 
+const getCaptchaUrl = (baseURL, locale) => {
```
jmelberg-okta, line 26: May need to doc that not all of the Widget's supported languages map properly.

For example, our code for Norwegian is `nb`, while [Google's is](https://developers.google.com/recaptcha/docs/language) `no`.

There appears to be one other instance, so we may need to manage a remapping object:
- `nb` 👉🏼  `no`
- `nl-NL` 👉🏼 `nl`

src/v2/view-builder/views/captcha/CaptchaView.js
kamaldeepcheema-okta: As part of this I'll move the method inside the view because otherwise it won't have the right context for `this`

src/v2/view-builder/views/captcha/CaptchaView.js
kamaldeepcheema-okta: @jmelberg-okta I actually noticed this as well when I was reading their docs but thankfully when I tested those scenarios, the mapping worked as you have above, both for reCaptcha and hCaptcha. I think they already perform this mapping on their side.

PR #2068, opened 2021-06-24
feat: Show Can't verify link for WebAuthn verify flow

jasleennuno-okta:
## Description:
- Show Can't verify link for WebAuthn verify flow
     Verify - show info about biometrics and security keys
     OV Enrollment - show alternative steps for enrolling

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

#### WebAuthn Verification Flow
Screen Shot 2021-06-28 at 3 17 45 PM
Screen Shot 2021-06-28 at 3 17 55 PM

#### OV Enrollment Case
Screen Shot 2021-06-28 at 3 18 24 PM

### Reviewers:
- @mauriciocastillosilva-okta @magizh-okta @vincentvoong-okta @panchamyadav-okta 

### Issue:

- [OKTA-405384](https://oktainc.atlassian.net/browse/OKTA-405384)


src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
```diff
@@ -115,8 +148,93 @@ const Body = BaseForm.extend({
   }
 });
 
+const Footer = AuthenticatorFooter.extend({
+  props: {
```
mauriciocastillosilva-okta, line 152: this is a view, I don't think we should add `props` to it, those belong to a model.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -801,6 +801,16 @@ oie.webauthn.error.not.supported = Security key or biometric authenticator is no
...
 oie.verify.webauthn.instructions = You will be prompted to use a security key or biometric verification (Windows Hello, Touch ID, etc.). Follow the instructions to complete verification.
 oie.verify.webauthn.uv.required.instructions = Biometric verification or a PIN is required to sign in with this authenticator.
+oie.verify.webauthn.cant.verify = Can't verify?
+oie.verify.webauthn.cant.verify.biometric.authenticator.title = Are you trying to use a biometric authenticator?
+oie.verify.webauthn.cant.verify.biometric.authenticator.description1 = Biometric authenticators (fingerprint, face recognition, PIN) will only work on the same device on which they were set up.
+oie.verify.webauthn.cant.verify.biometric.authenticator.description2 = If available, set up another security method that supports verification from a mobile browser.
+oie.verify.webauthn.cant.verify.security.key.title = Are you trying to use a security key?
+oie.verify.webauthn.cant.verify.security.key.description = Insert your security key into a USB port when propmted by the browser and tap on the button or gold disk. Security keys can work on multiple devices.
```
magizh-okta, line 809: This part is redundant with the on in the main description do we need this again ? cc @vincentvoong-okta

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -801,6 +801,16 @@ oie.webauthn.error.not.supported = Security key or biometric authenticator is no
 oie.verify.webauth.title = Verify with Security Key or Biometric Authenticator
 oie.verify.webauthn.instructions = You will be prompted to use a security key or biometric verification (Windows Hello, Touch ID, etc.). Follow the instructions to complete verification.
 oie.verify.webauthn.uv.required.instructions = Biometric verification or a PIN is required to sign in with this authenticator.
+oie.verify.webauthn.cant.verify = Can't verify?
+oie.verify.webauthn.cant.verify.biometric.authenticator.title = Are you trying to use a biometric authenticator?
+oie.verify.webauthn.cant.verify.biometric.authenticator.description1 = Biometric authenticators (fingerprint, face recognition, PIN) will only work on the same device on which they were set up.
+oie.verify.webauthn.cant.verify.biometric.authenticator.description2 = If available, set up another security method that supports verification from a mobile browser.
```
magizh-okta, line 807: nit: a mobile browser

Should it say something like "If available, set up another security method on the device where biometric enrollment was done."

src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
```diff
@@ -115,8 +148,93 @@ const Body = BaseForm.extend({
   }
 });
 
+const Footer = AuthenticatorFooter.extend({
```
magizh-okta, line 151: Lets move out this logic to a separate file like this 

https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/components/IdentifierFooter.js

May be not under shared

src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
```diff
@@ -115,8 +148,93 @@ const Body = BaseForm.extend({
...
+  props: {
+    enabled: ['boolean', false]
+  },
+
+  events: {
+    'click .js-cant-verify': function(e) {
+      e.preventDefault();
+      if (!this.props.enabled) {
```
magizh-okta, line 159: I guess rather than using the prop we can directly do the toggle here

src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
mauriciocastillosilva-okta, line 151: Looking at this Footer, I don't think we need to create a complete new one. I think we can just override `links` and return the link we need, otherwise in this new View we are basically copying all the logic that is already in BaseFooter

and a nit: name this `WebAuthnFooter`

magizh-okta (commented):
couple of nits and question for @vincentvoong-okta  on the text.

src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
```diff
@@ -115,8 +148,93 @@ const Body = BaseForm.extend({
...
+      this.add(footerInfo, '.footer-info');
+    }
+
+    this.listenTo(this.props.enabled, 'change:enabled', function(model, enable) {
+      this.$('.link').toggleClass('o-form-disabled', !enable);
+    });
+  },
+
+  links: function() {
+    let links = AuthenticatorFooter.prototype.links.apply(this, arguments);
```
mauriciocastillosilva-okta, line 210: can be a `const` since this is an array and you are just adding a new object to it, not changing the reference to it.

src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
```diff
@@ -115,8 +148,93 @@ const Body = BaseForm.extend({
...
+    links.push({
+      'label': loc('oie.verify.webauthn.cant.verify', 'login'),
+      'name': 'cant-verify',
+      'href': '#',
+      'aria-expanded': false,
+      'aria-controls': 'help-description-container',
+      'class': 'link help js-help',
```
mauriciocastillosilva-okta, line 217: in here what is missing is 
```
clickHandler: () => {
  // toggle the view that has the instructions
}
```

What I am missing I believe is where to add the view that has the instructions. So probably what would be better is to change `Link` so we can have a different type (maybe `toggle-link`) and that will add a view during initialize. Or something along those lines

packages/@okta/i18n/src/properties/login.properties
jasleennuno-okta: @panchamyadav-okta thoughts?

src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
jasleennuno-okta: @mauriciocastillosilva-okta how do you feel about leaving the new initialize() function?
there are 2 things different here:
one is that we use `getBackToSignInLink()` instead of `getSignOutLink()`
and we have the extra logic to add the new view when we add the link

packages/@okta/i18n/src/properties/login.properties
vincentvoong-okta: @magizh-okta which part is redundant? the use of "security key"?

packages/@okta/i18n/src/properties/login.properties
vincentvoong-okta: @magizh-okta @jasleennuno-okta 
Thanks Magizh, I'd change that slightly to: `If available, set up another security method on the device you used to set up your biometric authenticator.`

packages/@okta/i18n/src/properties/login.properties
magizh-okta: @vincentvoong-okta  yeah for security key we just seem to be providing instructions on how to verify and I feel it is not necessary in the 'can't verify' section.  But if we just want to add it for novice users I am fine.

assets/sass/v2/_footer.scss
```diff
@@ -8,6 +8,15 @@
       padding: 0.3rem 0;
       word-wrap: break-word;
     }
+
+    .help-description {
+      .ov-enrollment-info {
+        list-style: decimal;
+        margin-left: 15px;
+        margin-top: 15px;
+        margin-bottom: 15px;
```
magizh-okta, line 17: nit: margin: 15px 0 15px 15px;

test/unit/spec/v2/view-builder/views/ChallengeWebauthnView_spec.js
```diff
@@ -59,7 +59,8 @@ describe('v2/view-builder/views/webauthn/ChallengeWebauthnView', function() {
     expect(testContext.view.$('.retry-webauthn').css('display')).toBe('none');
     expect(testContext.view.$('.okta-waiting-spinner').css('display')).toBe('block');
     expect(testContext.view.$('.webauthn-not-supported').length).toBe(0);
-
+    expect(testContext.view.$('.js-cant-verify').length).toBe(1);
+    expect(testContext.view.$('.js-help-description').css('display')).toBe('none');
```
magizh-okta, line 63: nit: Rather than having this in all tests I would just have a single test for this.

magizh-okta (approved):
Looking good. Lets wait for Vince to confirm the text. after that :Ship

packages/@okta/i18n/src/properties/login.properties
vincentvoong-okta: @magizh-okta @jasleennuno-okta 
Yeah the security key part was included to let the user know that if they did also enroll in one before they can use it instead of trying to sign in with a platform authenticator. We can change the copy slightly to say: `If you have set up a security key, insert it in a USB port when prompted by the browser and tap on the button or gold disk. Security keys can work on multiple devices.`

magizh-okta (approved):

src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
mauriciocastillosilva-okta: Why do this view has a "back to sign in" link instead of a "sign out" like the other challenge views? This will mean that if I am in an authentication flow using say Phone and OV, then Phone will show "sign out" and OV will show "back to sign in"? That sounds inconsistent, can you please check?

I am still in favor of not duplicating the logic of BaseFooter, and I think we can get away with enhancing Link to support a "toggle text link". Option B can be to refactor BaseFooter and have functions for the common code so we avoid duplicating it.

pradeepdewda-okta (changes requested):
remove duplicate code.

src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
pradeepdewda-okta: see `IdentifierFooter.js ` for example.

jasleennuno-okta: @pradeepdewda-okta i don't think `footerInfo` works for what i'm needing to do here..
it looks like `footerInfo` adds an extra section at the VERY end of the footer..
i'm just trying to implement a sort of expand/collapse - take a look at the screenshots in the PR description

jasleennuno-okta: @mauriciocastillosilva-okta it looks like other authenticators also have `Back to sign in`

Screen Shot 2021-06-28 at 1 28 37 PM

Screen Shot 2021-06-28 at 1 29 49 PM

src/v2/view-builder/components/ChallengeWebauthnFooter.js
```diff
@@ -0,0 +1,98 @@
...
+      <h3>{{i18n code="oie.verify.webauthn.cant.verify.biometric.authenticator.title" bundle="login"}}</h3><br>
+      <p>{{i18n code="oie.verify.webauthn.cant.verify.biometric.authenticator.description1" bundle="login"}}</p><br>
+      <p>{{i18n code="oie.verify.webauthn.cant.verify.biometric.authenticator.description2" bundle="login"}}</p><br>
+      <h3>{{i18n code="oie.verify.webauthn.cant.verify.security.key.title" bundle="login"}}</h3><br>
+      <p>{{i18n code="oie.verify.webauthn.cant.verify.security.key.description" bundle="login"}}</p><br>
+    </div>
+  `,
+});
+
+const CantVerifyInfoOVEnrollmentFlowTemplate = View.extend({
```
pradeepdewda-okta, line 21: we generally call them view not `templates` .. `CantVerifyInfoOVEnrollmentFlowHelpLinkView`

src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
jasleennuno-okta: @pradeepdewda-okta this is what i'm trying to do - i think this is very different behavior than what footerInfo is intended to do

pradeepdewda-okta: @jasleennuno-okta you right for this you have to enhance `Links.js` https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/components/Link.js

you can add another param similar to `footerInfo` say.. `linkInfoView` that could be passed as a view from your `Footer` class.. `Link.js` can add such a view under any link..

vincentvoong-okta:
@jasleennuno-okta 
Text looks good! One thing about the order of the footer links. Can we swap "Can't verify?" with "Verify with something else"? If both are available, I think "Can't verify?" should be closer in proximity to WebAuthN UI since that is what it's addressing. It seems a bit strange if we separate it with "Verify with something else"

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -801,6 +801,16 @@ oie.webauthn.error.not.supported = Security key or biometric authenticator is no
...
 oie.verify.webauthn.uv.required.instructions = Biometric verification or a PIN is required to sign in with this authenticator.
+oie.verify.webauthn.cant.verify = Can't verify?
+oie.verify.webauthn.cant.verify.biometric.authenticator.title = Are you trying to use a biometric authenticator?
+oie.verify.webauthn.cant.verify.biometric.authenticator.description1 = Biometric authenticators (fingerprint, face recognition, PIN) will only work on the same device on which they were set up.
+oie.verify.webauthn.cant.verify.biometric.authenticator.description2 = If available, set up another security method on the device you used to set up your biometric authenticator.
+oie.verify.webauthn.cant.verify.security.key.title = Are you trying to use a security key?
+oie.verify.webauthn.cant.verify.security.key.description = If you have set up a security key, insert it in a USB port when prompted by the browser and tap on the button or gold disk. Security keys can work on multiple devices.
+oie.verify.webauthn.cant.verify.enrollment.step1 = Open your Okta Dashboard (e.g. yourcompany.okta.com) on the device you used to setup your security key or biometric authenticator
```
jmelberg-okta, line 810: May be pretty neat to use the actual value for the domain here, since we have it stored in our auth client.

src/v2/view-builder/components/ChallengeWebauthnFooter.js
```diff
@@ -0,0 +1,98 @@
...
+      links = links.filter(l => $.isPlainObject(l));
+    }
+
+    if (this.options.appState.shouldShowSignOutLinkInCurrentForm(
+      this.options.settings.get('features.hideSignOutLinkInMFA') ||
+      this.settings.get('features.mfaOnlyFlow')) && hasBackToSignInLink) {
+      links = links.concat(getBackToSignInLink(this.options.settings));
+    }
+
+    links.forEach(link => {
```
jmelberg-okta, line 55: nit: If possible, can we avoid multiple iterations over `links`?

test/unit/spec/v2/view-builder/views/ChallengeWebauthnView_spec.js
```diff
@@ -223,4 +222,70 @@ describe('v2/view-builder/views/webauthn/ChallengeWebauthnView', function() {
       })
       .catch(done.fail);
   });
+
+  it('shows correct text when Can\'t verify? is clicked', function() {
+    spyOn(webauthn, 'isNewApiAvailable').and.callFake(() => true);
+    spyOn(BrowserFeatures, 'isSafari').and.callFake(() => false);
```
jmelberg-okta, line 228: Please use `jest` objects for `spyOn`:

```suggestion
    jest.spyOn(BrowserFeatures, 'isSafari').mockReturnValue(false);
```

test/unit/spec/v2/view-builder/views/ChallengeWebauthnView_spec.js
```diff
@@ -223,4 +222,70 @@ describe('v2/view-builder/views/webauthn/ChallengeWebauthnView', function() {
...
+    testContext.init();
+    expect(testContext.view.$('.idx-webauthn-verify-text').text()).toBe(
+      'You will be prompted to use a security key or biometric verification (Windows Hello, Touch ID, etc.). Follow the instructions to complete verification.'
+    );
+    expect(testContext.view.$('.js-cant-verify').length).toBe(1);
+    expect(testContext.view.$('.js-help-description').css('display')).toBe('none');
+    const cantVerifyLink = document.getElementsByClassName('link js-cant-verify');
+    cantVerifyLink[0].click();
+    expect(testContext.view.$('.js-help-description').css('display')).toBe('block');
+    expect(testContext.view.$('.js-help-description')[0].textContent).toBe(
```
jmelberg-okta, line 238: nit: Can we use `inlineSnapshot` here instead?

src/v2/view-builder/components/ChallengeWebauthnFooter.js
```diff
@@ -0,0 +1,98 @@
+import AuthenticatorFooter from './AuthenticatorFooter';
+import { $, _, loc, View } from 'okta';
+import { getBackToSignInLink } from '../utils/LinksUtil';
+import Link from './Link';
+import hbs from 'handlebars-inline-precompile';
```
jmelberg-okta, line 5: style-nit: Module-like or alias imports usually are located at the top of the file. Can we keep the `okta` and `handlebars-inline-precompile` imports up top and together?

src/v2/view-builder/components/ChallengeWebauthnFooter.js
```diff
@@ -0,0 +1,98 @@
+import AuthenticatorFooter from './AuthenticatorFooter';
```
jmelberg-okta, line 1: All source files require legal headers - let's make sure we add them to all new files please!

jmelberg-okta (commented):
Nice work! Few suggestions called out below.

jasleennuno-okta:
> @jasleennuno-okta
> Text looks good! One thing about the order of the footer links. Can we swap "Can't verify?" with "Verify with something else"? If both are available, I think "Can't verify?" should be closer in proximity to WebAuthN UI since that is what it's addressing. It seems a bit strange if we separate it with "Verify with something else"

@vincentvoong-okta please take a look at the updated screenshots in the PR description

src/v2/view-builder/components/Link.js
```diff
@@ -19,6 +19,23 @@ const Link = View.extend({
...
 
+  /**
+   * View
+   * adds any view under the link
+   */
+  linkInfo: null,
+
+  initialize() {
+    View.prototype.initialize.apply(this, arguments);
+    if (this.options.linkInfo) {
```
jasleennuno-okta, line 30: adding the view in link, means we inherit the styling

Screen Shot 2021-06-28 at 3 31 05 PM

src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
mauriciocastillosilva-okta: discussed offline with @jasleennuno-okta . This is the plan
- Enhance `Links.js` so it supports a way to add a view inside the link that can be toggled. This is similar to what Jasleen had initially, but we will move the logic to `Links.js` instead of been in the Footer. This will also make it generic for any other view to consume.
- `linkInfoView` or something similar sounds good to me, and we will need a small css override for that new view to not inherit the link styles
- Remove the override of Footer once we confirm that `getSignOutLink` works properly, which it should since we have a cancel remediation form, but need to check details

@pradeepdewda-okta we can't use `footerInfo` for the reasons Jasleen mentioned, but we should be good with the plan I just described

pradeepdewda-okta: @mauriciocastillosilva-okta sounds about right. thanks @jasleennuno-okta for improving widget code..

src/v2/view-builder/internals/BaseFooter.js
```diff
@@ -62,9 +63,16 @@ export default View.extend({
     }
 
     links.forEach(link => {
-      this.add(Link, {
-        options: link,
-      });
+      if (link.linkInfo) {
```
pradeepdewda-okta, line 66: Nice. this worked as i mentioned over slack. ❤️  that you created a new component, much cleaner.

src/v2/view-builder/components/ChallengeWebauthnFooter.js
```diff
@@ -0,0 +1,68 @@
...
+      CantVerifyInfoOVEnrollmentFlowView: CantVerifyInfoVerifyFlowView;
+    links.unshift({
+      'label': loc('oie.verify.webauthn.cant.verify', 'login'),
+      'name': 'cant-verify',
+      'aria-expanded': false,
+      'aria-controls': 'help-description-container',
+      'class': 'link help js-help',
+      'linkInfo': cantVerifyInfoView,
+      'clickHandler': function() {
+        $('.js-help-description').slideToggle(200, () => {
```
pradeepdewda-okta, line 56: possible to use `this.$el.find('.js-help-description')` ?

src/v2/view-builder/components/ChallengeWebauthnFooter.js
```diff
@@ -0,0 +1,68 @@
...
+      </ol>
+  `,
+});
+
+export default AuthenticatorFooter.extend({
+  links: function() {
+    const links = AuthenticatorFooter.prototype.links.apply(this, arguments);
+
+    const cantVerifyInfoView = this.options.appState.get('app')
+      && this.options.appState.get('app').name === 'Okta_Authenticator' ?
```
pradeepdewda-okta, line 46: make this constant ?

src/v2/view-builder/components/ToggleTextLink.js
```diff
@@ -0,0 +1,38 @@
...
+   * View
+   * adds any view under the link
+   */
+  linkInfo: null,
+
+  initialize() {
+    View.prototype.initialize.apply(this, arguments);
+    let link = _.resultCtx(this, 'link', this);
+    const linkInfo = _.resultCtx(this, 'footerInfo', this);
+    this.add(link);
```
pradeepdewda-okta, line 33: if linkInfo is not present then it is just like `Link` ?

jasleennuno-okta: then it just adds a link, yes

src/v2/view-builder/internals/BaseFooter.js
```diff
@@ -62,9 +63,18 @@ export default View.extend({
...
+        this.add(ToggleTextLink, {
+          options : {
+            link: new Link({ options : link }),
+            linkInfo: link.linkInfo,
+          }
+        });
+      } else {
```
pradeepdewda-okta, line 73: can't we just add `ToggleTextLink` and remove `else` part, because it only adds simple `Link` when `linkInfo` is not available ?

src/v2/view-builder/components/ChallengeWebauthnFooter.js
```diff
@@ -0,0 +1,70 @@
...
+    const cantVerifyInfoView = this.options.appState.get('app')
+      && this.options.appState.get('app').name === OKTA_AUTHENTICATOR ?
+      CantVerifyInfoOVEnrollmentFlowView: CantVerifyInfoVerifyFlowView;
+    links.unshift({
+      'label': loc('oie.verify.webauthn.cant.verify', 'login'),
+      'name': 'cant-verify',
+      'aria-expanded': false,
+      'aria-controls': 'help-description-container',
+      'class': 'link help js-help',
+      'linkInfo': cantVerifyInfoView,
```
mauriciocastillosilva-okta, line 56: can we instead send an object with the data ToggleTextLink needs? I am thinking something like
```
type: 'toggle-text-link',
additionalOptions: {
  view: cantVerifyInfoView,
  selector: 'js-help-description',
}
```

src/v2/view-builder/components/ToggleTextLink.js
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * View
+   * adds any view under the link
+   */
+  linkInfo: null,
+
+  initialize() {
+    BaseView.prototype.initialize.apply(this, arguments);
+    this.add(Link, {
```
mauriciocastillosilva-okta, line 33: here is what I am thinking to add the clickHandler and anything else related to toggling the view
```
BaseView.prototype.initialize.apply(this, arguments);
const textViewOptions = this.options.link.options.additionalOptions;
const linkOptions: Object.assign({}, this.options.link.options.options, {
  'aria-expanded': false,
  'clickHandler': function() {
    $(textViewOptions.textViewSelector).slideToggle(200, () => {
      this.attr('aria-expanded', $(textViewOptions.textViewSelector).is(':visible'));
    });
  });
});

this.add(Link, linkOptions);
this.add(textViewOptions.view);
```

src/v2/view-builder/components/ChallengeWebauthnFooter.js
```diff
@@ -0,0 +1,68 @@
...
+      'additionalOptions': {
+        view: cantVerifyInfoView,
+        selector: '.js-help-description',
+      },
+    });
+
+    return links;
+  },
+
+  postRender: function() {
```
mauriciocastillosilva-okta, line 65: let's move this to ToggleTextLink (and use the `selector` from the options)

src/v2/view-builder/components/ToggleTextLink.js
```diff
@@ -0,0 +1,34 @@
...
+  initialize() {
+    BaseView.prototype.initialize.apply(this, arguments);
+    const textViewOptions = this.options.additionalOptions;
+    const linkOptions = Object.assign({}, this.options,
+      {
+        'type': 'link',
+        'aria-expanded': false,
+        'clickHandler': function() {
+          $(textViewOptions.selector).slideToggle(200, () => {
+            $(textViewOptions.selector).attr('aria-expanded', false);
```
mauriciocastillosilva-okta, line 27: this should be set to the value of the selector right? not to `false`
I think this is something like
```
this.attr('aria-expanded', $(textViewOptions.textViewSelector).is(':visible'));
```

test/unit/spec/v2/view-builder/views/ChallengeWebauthnView_spec.js
```diff
@@ -223,4 +222,70 @@ describe('v2/view-builder/views/webauthn/ChallengeWebauthnView', function() {
...
+    testContext.init();
+    expect(testContext.view.$('.idx-webauthn-verify-text').text()).toMatchInlineSnapshot(
+      '"You will be prompted to use a security key or biometric verification (Windows Hello, Touch ID, etc.). Follow the instructions to complete verification."'
+    );
+    expect(testContext.view.$('.js-cant-verify').length).toBe(1);
+    expect(testContext.view.$('.js-help-description').css('display')).toBe('none');
+    const cantVerifyLink = document.getElementsByClassName('link js-cant-verify');
+    cantVerifyLink[0].click();
+    expect(testContext.view.$('.js-help-description').css('display')).toBe('block');
+    expect(testContext.view.$('.js-help-description')[0].textContent).toMatchInlineSnapshot(
```
mauriciocastillosilva-okta, line 238: nit: either just compare to a simple string`$('.js-help-description').first().text()).toEqual('Are you trying...');`
or
compare the inline snapshot to html `$('.js-help-description').first().html()).toMatchInlineSnapshot`

test/unit/spec/v2/view-builder/views/ChallengeWebauthnView_spec.js
```diff
@@ -223,4 +222,70 @@ describe('v2/view-builder/views/webauthn/ChallengeWebauthnView', function() {
...
```
mauriciocastillosilva-okta, line 287: same nit here, either a string or inline snapshot for html. Probably html is better

mauriciocastillosilva-okta (commented):
Looking great! I like how ToggleTextLink works now. Thank you

src/v2/view-builder/components/ChallengeWebauthnFooter.js
pradeepdewda-okta: also selector should be `.js-cant-verify`  ??  this is not a `help` link ..

pradeepdewda-okta (approved):
looks good to me. thank you guys again.

src/v2/view-builder/internals/BaseFooter.js
```diff
@@ -62,9 +63,11 @@ export default View.extend({
...
 
     links.forEach(link => {
-      this.add(Link, {
-        options: link,
-      });
+      let LinkView = Link;
+      if (link.type === 'toggle-text-link') {
+        LinkView = ToggleTextLink;
+      }
+      this.add(LinkView, { options: link, });
```
mauriciocastillosilva-okta, line 70: nit: remove the extra comma at the end, or leave it as it was before in a different line

test/unit/spec/v2/view-builder/views/ChallengeWebauthnView_spec.js
```diff
@@ -223,4 +222,70 @@ describe('v2/view-builder/views/webauthn/ChallengeWebauthnView', function() {
...
+    const cantVerifyLink = document.getElementsByClassName('link js-cant-verify');
+    cantVerifyLink[0].click();
+    expect(testContext.view.$('.js-help-description').css('display')).toBe('block');
+    expect(testContext.view.$('.js-help-description').first().html()).toMatchInlineSnapshot(
+      '"<h3>Are you trying to use a biometric authenticator?</h3><br><p>Biometric authenticators (fingerprint, face recognition, PIN) will only work on the same device on which they were set up.</p><br><p>If available, set up another security method on the device you used to set up your biometric authenticator.</p><br><h3>Are you trying to use a security key?</h3><br><p>If you have set up a security key, insert it in a USB port when prompted by the browser and tap on the button or gold disk. Security keys can work on multiple devices.</p><br>"'
+    );
+  });
+
+  it('shows additional text when Can\'t verify? is clicked from Okta_Authenticator', function() {
+    testContext.init = (
```
mauriciocastillosilva-okta, line 244: instead of overriding `init` here, just make the changes to the `init` function at the beginning of this file. The only difference is that here you have a value for `app`, so you can handle that in the common init for all tests
```
const appStateOptions = {
  currentAuthenticator,
  authenticatorEnrollments,
};
if (app) {
  appStateOptions.app = app;
}
const appState = new AppState(appStateOptions);
```
The other changes to use jest spies and expects are good for all tests

src/v2/view-builder/components/ToggleTextLink.js
```diff
@@ -0,0 +1,39 @@
...
+  initialize() {
+    BaseView.prototype.initialize.apply(this, arguments);
+    const textViewOptions = this.options.additionalOptions;
+    const linkOptions = Object.assign({}, this.options,
+      {
+        'type': 'link',
+        'aria-expanded': false,
+        'clickHandler': function() {
+          $(textViewOptions.selector).slideToggle(200, () => {
+            $(textViewOptions.selector).attr('aria-expanded', $(textViewOptions.select).is(':visible'));
```
mauriciocastillosilva-okta, line 27: this needs a small change, I don't think it will work as is: we need to change the attribute `aria-expanded` in the link view, not in the text view. I think it should be something like
```
'clickHandler': function(e) {
  $(textViewOptions.selector).slideToggle(200, () => {
    e.currentTarget.setAttribute('aria-expanded', $(textViewOptions.selector).is(':visible'));
  });
}
```
check if this works, and you can also add a small check for this in the tests you already have

src/v2/view-builder/components/ToggleTextLink.js
```diff
@@ -18,13 +18,14 @@ export default View.extend({
...
     const textViewOptions = this.options.additionalOptions;
+    const linkName = this.options.name;
     const linkOptions = Object.assign({}, this.options,
       {
         'type': 'link',
         'aria-expanded': false,
         'clickHandler': function() {
           $(textViewOptions.selector).slideToggle(200, () => {
-            $(textViewOptions.selector).attr('aria-expanded', $(textViewOptions.select).is(':visible'));
+            $(`js-${linkName}`).attr('aria-expanded', $(textViewOptions.selector).is(':visible'));
```
jasleennuno-okta, line 28: `js-${linkName}` - this is how `Link.js` forms the name

src/v2/view-builder/components/ToggleTextLink.js
jasleennuno-okta: `e` was still undefined.. let me know what you think of the way i came up with it

src/v2/view-builder/components/ToggleTextLink.js
mauriciocastillosilva-okta: sounds good.
Can we add a line in the test to check for it? something like
```
expect(testContext.view.$('.js-cant-verify').attr('aria-expanded')).toBe(true/false);
```

mauriciocastillosilva-okta (approved):
looks good. Just one comment to add to a test and then 🚢

PR #2075, opened 2021-06-28
Upgrade jQuery to latest version 3.6.0

aarongranick-okta:
## Description:
- overrides / replaces courage jQuery version with current latest `3.6.0`
- fixes issues in tests (1 test disabled, see OKTA-407603) 

Most common issues were related to time/race conditions.
The $(function() ) syntax operates asynchronously in current jQuery.
Created a new `BaseController` to override this behavior from courage.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-377873](https://oktainc.atlassian.net/browse/OKTA-377873)


package.json
```diff
@@ -143,19 +144,19 @@
...
-    "webpack-dev-server": "2.11.3",
-    "@okta/eslint-plugin-okta": "0.4.0"
+    "webpack-dev-server": "2.11.3"
   },
   "dependencies": {
     "@babel/polyfill": "^7.10.1",
     "@babel/runtime": "^7.10.3",
     "cross-fetch": "^3.0.4",
     "handlebars": "^4.5.3",
+    "jquery": "^3.6.0",
```
haishengwu-okta, line 154: seems not necessary to be public dependence given we bundle everything ?

jmelberg-okta: Adding onto the comment above, does this change our [multiple output configurations](https://github.com/okta/okta-signin-widget/#using-the-npm-module)?

test/unit/helpers/mocks/Util.js
```diff
@@ -311,6 +316,10 @@ fn.mockSetTimeout = function() {
   }
 };
 
+fn.hasTimeouts = function() {
+  return !!timeouts.length;
+};
```
jmelberg-okta, line 321: Is this used?

test/unit/spec/ForgotPassword_spec.js
```diff
@@ -18,7 +18,7 @@ import $sandbox from 'sandbox';
 const SharedUtil = internal.util.Util;
 const itp = Expect.itp;
 
-function setup(settings, startRouter) {
+async function setup(settings, startRouter) {
```
jmelberg-okta, line 21: Is this now a requirement for all Karma-based tests that use this approach for setup?

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -362,6 +362,12 @@ Expect.describe('IDPDiscovery', function() {
       });
     });
     itp('does not show a beacon if features.securityImage is false', function() {
+
+      // BaseLoginRouter will render twice if language bundles are not loaded:
+      // https://github.com/okta/okta-signin-widget/blob/master/src/util/BaseLoginRouter.js#L202
+      // We are not testing i18n, so we can mock language bundles as loaded
+      Util.mockBundles();
```
jmelberg-okta, line 369: Hmm.. these should be loaded by default given our language code should be `en`. Is a different locale being used in this test?

jmelberg-okta (commented):
Few questions to kick things off!

package.json
aarongranick-okta: The convention that I follow is to include all code which is part of the bundle in dependencies. When we publish as an ESM package this will be important. I also think it helps for security audits to know whether something is a "dev" dependency (tooling, tests) vs. a production dependency (code exists in the bundle and is shipped to end-users). @jmelberg-okta this has no effect on the output, but dependencies are installed along with the module so it would increase install time slightly.

test/unit/spec/ForgotPassword_spec.js
aarongranick-okta: no it's not a requirement, but I find it a little easier to debug and catch exceptions

test/unit/spec/IDPDiscovery_spec.js
aarongranick-okta: I don't think this test does any setup for i18n. But the double render was causing a race condition on the beacon. I guess the race existed but was resolving consistently with previous jquery version.

test/unit/helpers/mocks/Util.js
aarongranick-okta: it was used while I was hunting down these async issues. Most of those issues were resolved by patching the `render` in BaseController

aarongranick-okta:
@haishengwu-okta I have moved BaseController logic to CourageForSigninWidget

haishengwu-okta (approved):

test/unit/helpers/mocks/Util.js
jmelberg-okta: Great - let's remove the unused function then, since it is now unused.

test/unit/spec/IDPDiscovery_spec.js
jmelberg-okta: Can we look into this further? The code block you've linked to in the comment shouldn't be executed, as we're not testing a different locale.

Perhaps something else is happening that may be indicating a bigger problem?

test/unit/spec/IDPDiscovery_spec.js
```diff
@@ -634,7 +640,7 @@ Expect.describe('IDPDiscovery', function() {
     });
     itp('toggles "focused-input" css class on focus in and focus out', function() {
       return setup().then(function(test) {
-        test.form.usernameField().focus();
+        test.form.usernameField().focusin();
```
jmelberg-okta, line 643: Can you help me understand why `focus` doesn't work anymore?

test/unit/spec/LoginRouter_spec.js
```diff
@@ -823,65 +823,63 @@ Expect.describe('LoginRouter', function() {
...
+    test.setNextResponse(resRecovery);
+    test.router.refreshAuthState('dummy-token');
+    await Expect.waitForRecoveryQuestion(test);
+    
+    test.setNextResponse(errorInvalidToken);
+    let form = new RecoveryForm($sandbox);
+
+    form.setAnswer('4444');
+    form.submit();
+    await Expect.waitForPrimaryAuth(test);
```
jmelberg-okta, line 841: Since we aren't handling the resolved/rejected values for these anymore, should we wrap these tests in a `try catch` or ensure all expectations are called and there are no silent failures?

test/unit/spec/LoginRouter_spec.js
```diff
@@ -1399,6 +1397,11 @@ Expect.describe('LoginRouter', function() {
       Util.loadWellKnownAndKeysCache();
       const successSpy = jasmine.createSpy('successSpy');
 
+      // In this test the id token will be returned succesfully. It must pass all validation.
+      // Mock the date to 10 seconds after token was issued.
+      const AUTH_TIME = (1451606400) * 1000; // The time the "VALID_ID_TOKEN" was issued
```
jmelberg-okta, line 1402: Does this mean whenever our mock token changes, we'll need to update this `AUTH_TIME` as well?

webpack.common.config.js
```diff
@@ -21,7 +21,9 @@ module.exports = function(outputFilename) {
         'nls': '@okta/i18n/src/json',
         'okta': `${LOCAL_PACKAGES}/@okta/courage-dist/okta.js`,
         'okta-i18n-bundles': 'util/Bundles',
-        'jquery': `${LOCAL_PACKAGES}/@okta/courage-dist/jquery.js`,
+
+        // Uncomment the line below to use the version of jQuery from courage
+        // 'jquery': `${LOCAL_PACKAGES}/@okta/courage-dist/jquery.js`,
```
jmelberg-okta, line 26: I don't think we have instructions for this in the README, correct? Should we remove this code?

jmelberg-okta (commented):
Few questions about test changes that seem suspicious.

test/unit/spec/LoginRouter_spec.js
aarongranick-okta: if any exception throws within the test it will fail

test/unit/spec/IDPDiscovery_spec.js
aarongranick-okta: the problem was a race condition. while the bundle is loading it wants to put a "loading" beacon on. But are test is trying to make assertions about the beacon. The loading logic is just in the way.

webpack.common.config.js
aarongranick-okta: I left this in only for us, in case we needed to do some quick A/B test. There is no reason an end-user should need / want to use the courage jquery

test/unit/spec/LoginRouter_spec.js
aarongranick-okta: yes the mock time must match the issued time of the token, or it will fail validation

test/unit/spec/IDPDiscovery_spec.js
aarongranick-okta: Not entirely sure. May be related to this: https://github.com/jquery/jquery/issues/4382

test/unit/spec/PrimaryAuth_spec.js
```diff
@@ -466,6 +481,13 @@ Expect.describe('PrimaryAuth', function() {
   });
 
   Expect.describe('elements', function() {
+    beforeEach(() => {
+      // BaseLoginRouter will render twice if language bundles are not loaded:
+      // https://github.com/okta/okta-signin-widget/blob/master/src/util/BaseLoginRouter.js#L202
+      // We are not testing i18n, so we can mock language bundles as loaded
+      Util.mockBundles();
```
jmelberg-okta, line 488: Since this is a race condition related to the beacon, let's move this into the test below so it doesn't extend to all other tests.

jmelberg-okta (approved):

PR #2082, opened 2021-06-29
Adding Back to sign in link for redirect-idp view

kamaldeepcheema-okta:
## Description:

- If the current view only has the social idps buttons, there's no way for the user to re-initiate the sign-in flow. This makes for a bad user experience especially when they enter the wrong username - see https://oktainc.atlassian.net/browse/OKTA-405042 for more details.
- Now we render the "Back to sign in" link for these views.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/77295104/123866923-ad405f00-d8fb-11eb-8b66-a439caf2d558.mov

### Issue:

- [OKTA-405042](https://oktainc.atlassian.net/browse/OKTA-405042)


test/testcafe/framework/page-objects/IdentityPageObject.js
```diff
@@ -189,4 +190,8 @@ export default class IdentityPageObject extends BasePageObject {
   getCustomIdpButtonLabel(index) {
     return Selector(CUSTOM_IDP_BUTTON).nth(index).textContent;
   }
+
+  async hasBackToSignInLink() {
```
pradeepdewda-okta, line 194: there is `goBackLinkExists` exists in `BasePageObject`.. probably you can reuse that.

src/v2/view-builder/views/RedirectIdPView.js
```diff
@@ -1,6 +1,24 @@
 import IdentifierView from './IdentifierView';
+import IdentifierFooter from '../components/IdentifierFooter';
+import { getBackToSignInLink } from '../utils/LinksUtil';
+
 
 export default IdentifierView.extend({
+
+  initialize() {
```
pradeepdewda-okta, line 8: what is the remediation ? does the api response doesn't contain `cancel` object

test/testcafe/framework/page-objects/IdentityPageObject.js
kamaldeepcheema-okta: good catch! I didn't realize this existed already.

src/v2/view-builder/views/RedirectIdPView.js
kamaldeepcheema-okta: In my testing, it didn't render in either case (the remediation with or without the cancel object). It's because our logic flows through IdentifierFooter.js and this does not take that into account when generating the links: https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/components/IdentifierFooter.js#L34

pradeepdewda-okta: check if BaseFooter intentionally not adding it ?  or at `AppState.js` `shouldShowSignOutLinkInCurrentForm` returning false

kamaldeepcheema-okta: @pradeepdewda-okta Yeah it looks like `shouldShowSignOutLinkInCurrentForm` in `AppState` is returning false ultimately because: https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/RemediationConstants.js#L74. So we can simply just remove this from the list and it would do the job. Although I'm not sure why this was added here to begin with. The history doesn't reveal much.

pradeepdewda-okta (approved):
yeah if this is a bug. then removing the form from here make sense. @haishengwu-okta see if you have any context.

haishengwu-okta (approved):

PR #2085, opened 2021-06-30
Fix issue with long IDP button labels

kamaldeepcheema-okta:
## Description:

- Fix an issue where long names for IDP buttons caused an overflow issue in the UI (for more details: https://oktainc.atlassian.net/browse/OKTA-405181).
- Now we truncate the string with an ellipsis and introduce a tooltip. Signed off solution with @unniravindranathan-okta  (PM) and @vincentvoong-okta (Product Designer) 

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/77295104/124012382-1af99300-d9af-11eb-8b1e-221520f330fd.mov

### Issue:

- [OKTA-405181](https://oktainc.atlassian.net/browse/OKTA-405181)


assets/sass/modules/_btns.scss
```diff
@@ -88,7 +88,11 @@ $od-secondary-button-action-text-color: #124a94;
   line-height: 50px;
   margin-top: 15px;
   padding-left: 90px;
+  padding-right: 40px;
```
kamaldeepcheema-okta, line 91: we don't seem to have pre-defined variables or standards for margins/paddings hence why I hardcoded this value here.

jmelberg-okta: Not a designer, but shouldn't padding on the left and right be consistent so the content is centered properly?

kamaldeepcheema-okta: Yeah this is a good question. The answer is the buttons have the icon rendered on the left side and left padding takes this into account. It takes up about 50px according to what I saw hence why it's set to 40px. You can also refer to the video and see that the string is centered.

jmelberg-okta: Thanks for the great explanation! Hopefully we don't have any icon variants that push/pull this padding outside of the bounds we expect, but do agree this is a good approach for now.

kamaldeepcheema-okta: Yeah I don't think we'll have a case like that, at least by looking at this: https://github.com/okta/okta-signin-widget/blob/master/assets/sass/modules/_btns.scss#L198. According to that the background gradient color is set for a 50px X 50px square.

src/v2/view-builder/internals/FormInputFactory.js
```diff
@@ -137,6 +137,7 @@ const createIdpButtons = (remediations) => {
     const button = {
       attributes: {
         'data-se': `social-auth-${type}-button`,
+        'title': displayName // We add a tooltip in case the name gets truncated if too long
```
jmelberg-okta, line 140: One problem with this is a screen reader will read both the button text **and** the tooltip `title`. If we can, we should try to avoid the duplicate reading.

@edburyenegren-okta - Do you have a suggestion here?

kamaldeepcheema-okta: Yeah great point! This didn't cross my mind. I'll wait for @edburyenegren-okta to chime in.

test/unit/spec/v2/view-builder/views/IdentifierView_spec.js
```diff
@@ -90,4 +90,19 @@ describe('v2/view-builder/views/IdentifierView', function() {
     expect(testContext.view.$el.find('.o-form-fieldset-container .sign-in-with-idp').length).toEqual(0);
     expect(testContext.view.$el.find('.o-form-button-bar .sign-in-with-idp').length).toEqual(0);
   });
+
+  it('view renders IDP buttons correctly with tooltip', function() {
```
pradeepdewda-okta, line 94: so button tooltip will always exists now ? I think make it available only when there is a long name..

src/v2/view-builder/internals/FormInputFactory.js
edburyenegren-okta: This is partly the reason Odyssey uses a separate Tooltip component, but that feels far out of scope for this change.

One solution would be wrapping the truncated text leaf in a `span` with `aria-hidden`, and giving the `button` an `aria-label` or `title` (but not both) that contains the full, untruncated text. That would require swapping _all_ buttons over or detecting truncation somehow; both feel less than ideal...

Is Design comfortable with any other behaviors? Showing the full title on `:hover` or using a full Tooltip?

test/unit/spec/v2/view-builder/views/IdentifierView_spec.js
kamaldeepcheema-okta: @pradeepdewda-okta, yes for now that's the behavior. If we want to dynamically add a tooltip, then we'd need to come up with a character limit that would be sensible for all languages past which we would generate that. I didn't go down that route because this was the safer approach (I didn't know about the screen reading issue). 

Any ideas on how best to come up with this threshold?

pradeepdewda-okta: how about we do this based on overlapping text? If text overlaps the button width then add a tooltip ?

kamaldeepcheema-okta: @pradeepdewda-okta @edburyenegren-okta I just pushed some updates to only add the tooltip when the text is longer than the allowed width. Please have a look.

src/v2/view-builder/internals/FormInputFactory.js
```diff
@@ -102,6 +118,10 @@ const create = function(uiSchemaObj) {
  *
  */
 const createIdpButtons = (remediations) => {
+  // This is the max allowable text width that we can display for IDP buttons.
+  // Generated by taking the 316px width of button minus padding-left (90 px) and padding-right (50px)
+  const MAX_WIDTH = 186;
```
edburyenegren-okta, line 123: Can we handle this via CSS instead?

src/v2/view-builder/internals/FormInputFactory.js
```diff
@@ -10,6 +10,22 @@ import EnduserScopeList from '../../../views/consent/ScopeList';
 import CaptchaView from '../views/captcha/CaptchaView';
 
 
+const isTextOverflowing = (text, maxWidth) => {
```
pradeepdewda-okta, line 13: perfect 👍

src/v2/view-builder/internals/FormInputFactory.js
```diff
@@ -142,6 +162,12 @@ const createIdpButtons = (remediations) => {
       title: displayName,
       href: idpObject.href,
     };
+
+    if (isTextOverflowing(displayName, MAX_WIDTH)) {
```
pradeepdewda-okta, line 166: should we rename, `isTextOverflow`?

src/v2/view-builder/internals/FormInputFactory.js
pradeepdewda-okta, line 123: MAX_IDP_BUTTON_WIDTH

pradeepdewda-okta (approved):

src/v2/view-builder/internals/FormInputFactory.js
kamaldeepcheema-okta: @edburyenegren-okta this is strictly for doing the calculation on the fly of whether the IDP buttons we render will overflow. Although not ideal, we need the max width in here to compare against, I don't see a way to determine that part via CSS. We ultimately use this to determine whether we add a `title` attr. We still handle the actual truncating and displaying of the ellipsis via CSS.

src/v2/view-builder/internals/FormInputFactory.js
kamaldeepcheema-okta: yep will do

jmelberg-okta (approved):

PR #2090, opened 2021-07-04
[OKTA-408185] Restart login flow on cancel

denysoblohin-okta:
## Description:
Issue: User can't login after clicking 'Back to sign in' (see video)
JS error: 
```js
TypeError: Cannot read property 'state' of null
```
on this line:
https://github.com/okta/okta-signin-widget/blob/8d27ecb7f46fb1ea11ad49cf8538b661de546239/src/v2/client/interactionCodeFlow.js#L23

Source of error: After click on 'Back to sign in' idx transaction is canceled and transaction meta is cleared.
Fix method is to restart login flow after clear-up

## PR Checklist

- [ ] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video of error this PR is fixing:

https://user-images.githubusercontent.com/72614880/124826104-1fa8e300-df7d-11eb-9c52-750f36ca0a3e.mov

Screen Shot 2021-06-28 at 10 21 27 PM

### Issue:

- [OKTA-408185](https://oktainc.atlassian.net/browse/OKTA-408185)


src/v2/controllers/FormController.js
```diff
@@ -156,6 +156,19 @@ export default Controller.extend({
         .then(this.handleIdxSuccess.bind(this))
         .catch(error => {
           this.showFormErrors(this.formView.model, error);
+        })
```
shuowu, line 159: How about emit an event and let the `BaseLoginRouter` listen on it? in this way, we can centralize the reset/init logic in one place.

src/models/Settings.js
```diff
@@ -530,4 +533,21 @@ export default Model.extend({
   isDsTheme: function() {
     return false;
   },
+
+  unsetOieSettings: function() {
```
shuowu, line 537: When the transaction is canceled, a new one should be started when back to the login screen. No preserve logic is needed here.

denysoblohin-okta: After starting login flow these 3 options are unset:
https://github.com/okta/okta-signin-widget/blob/8d27ecb7f46fb1ea11ad49cf8538b661de546239/src/v2/BaseLoginRouter.js#L216-L218
This will prevent further interaction:
https://github.com/okta/okta-signin-widget/blob/8d27ecb7f46fb1ea11ad49cf8538b661de546239/src/v2/client/startLoginFlow.js#L37-L40

shuowu: `stateToken` should be unset in a new transaction. `proxyIdxResponse` is probably used for debugging only. I only setting you want to preserve is `useInteractionCodeFlow`, which can be set to true programmatically when an event is triggered. We can probably name it as `interactionCanceled`.

src/v2/controllers/FormController.js
pradeepdewda-okta: We should not initiate login flow from FormController..

denysoblohin-okta: @shuowu @pradeepdewda-okta Please review changes

pradeepdewda-okta:
Can you add a video? Also add debugger screen.

Can you verify that we are cleaning up `stateToken` from sessionStorage properly.. You can check that in debugger

src/v2/controllers/FormController.js
pradeepdewda-okta: I think instead, you should trigger BaseLoginRouter's `render` function instead that starts the loginflow and handle cleanup..

src/v2/BaseLoginRouter.js
```diff
@@ -185,6 +186,18 @@ export default Router.extend({
     // }
   },
 
+  handleIdxInteractionCanceled() {
```
pradeepdewda-okta, line 189: instead of creating this new handler, it should use existing `startLoginFlow` of `render`

src/v2/BaseLoginRouter.js
```diff
@@ -185,6 +186,18 @@ export default Router.extend({
...
+    // Restart login flow
+    this.settings.set('useInteractionCodeFlow', true);
+    startLoginFlow(this.settings)
```
pradeepdewda-okta, line 192: this creates duplication and alienate `interact` from main flow, hard to maintain widget state properly in case of errors..

denysoblohin-okta:
@pradeepdewda-okta 
I've created `startLoginFlow` method in login router to reuse in `render` and `interactionCanceled` handler.
In my understanding `interact` is necessary to start a new login flow after cancel, because `interactionHandle` has been cleared on `cancel`. So not sure how to eliminate this alienation.

Added video and screenshot of error.

src/v2/BaseLoginRouter.js
```diff
@@ -80,6 +80,7 @@ export default Router.extend({
     configIdxJsClient(this.appState);
     this.listenTo(this.appState, 'remediationSuccess', this.handleIdxResponseSuccess);
     this.listenTo(this.appState, 'remediationError', this.handleIdxResponseFailure);
+    this.listenTo(this.appState, 'interactionCanceled', this.startLoginFlow);
```
pradeepdewda-okta, line 83: does it not work if you do `this.render` instead of `this.startLoginFlow`

pradeepdewda-okta:
Couldn't understand the video.. do we expect to see error here?  that doesn't seem right

denysoblohin-okta:
@pradeepdewda-okta 
Video of error:

https://user-images.githubusercontent.com/72614880/124838218-10329580-df8f-11eb-920d-5d56df1120c2.mov


Video of fix:

https://user-images.githubusercontent.com/72614880/124837987-99959800-df8e-11eb-84fc-3959fa62278f.mov

We do not expect error. This PR should fix the error

src/v2/BaseLoginRouter.js
denysoblohin-okta: No, this would produce error `TypeError: Controller is not a constructor` in this line:
https://github.com/okta/okta-signin-widget/blob/8d27ecb7f46fb1ea11ad49cf8538b661de546239/src/v2/BaseLoginRouter.js#L247

denysoblohin-okta: Changed `render` method for using it for re-render current controller (without parameters)

src/v2/controllers/FormController.js
denysoblohin-okta: Done

src/v2/BaseLoginRouter.js
```diff
@@ -238,17 +239,20 @@ export default Router.extend({
...
+        settings: this.settings,
+        appState: this.appState
+      }, options);
+      this.controller = new Controller(controllerOptions);
+
+      // Bubble up all controller events
+      this.listenTo(this.controller, 'all', this.trigger);
+    }
+    
     this.controller.render();
```
aarongranick-okta, line 256: In what case do we not have `Controller` ? In that case, won't the next line
```
this.controller.render()
```
throw an exception? (no property render on undefined)

test/testcafe/spec/SessionStorage_spec.js
```diff
@@ -261,3 +263,62 @@ test.requestHooks(credentialSSONotExistLogger, credentialSSONotExistMock)('shall
...
+  // Add mocks for interaction code flow
+  const challengeSuccessMock = RequestMock()
+    .onRequestTo('http://localhost:3000/idp/idx/challenge/answer')
+    .respond(xhrSuccessWithInteractionCode)
+    .onRequestTo('http://localhost:3000/oauth2/default/v1/token')
+    .respond(xhrSuccessTokens);
+  await t.addRequestHooks(challengeSuccessMock);
+
+  // Setup widget with interaction code flow
+  const optionsForInteractionCodeFlow = {
```
aarongranick-okta, line 281: It looks like this may have been copied from `EnglishLeaks_spec`. This config appears to be incorrect / misleading.

- `codeVerifier` is not a widget option. (It may exist in transaction meta)
- `codeChallenge` option will put the widget into a different mode (remediation, meant for server-side web apps). In this mode the widget will not retrieve tokens. Although the token endpoints are mocked for success, I believe this part of the flow is not being tested.
- `codeChallengeMethod` default value is `S256`. Recommend not setting this option in the test

It would be great if you could clean up the config here and in `EnglishLeaks`:
- remove `codeVerifier`, `codeChallenge` and `codeChallengeMethod` options. This should not break the tests. If, for some reason, removing these options does cause any change, we should test both cases. (with/without codeChallenge)

src/v2/controllers/FormController.js
```diff
@@ -132,8 +132,13 @@ export default Controller.extend({
 
   handleInvokeAction(actionPath = '') {
     const idx = this.options.appState.get('idx');
+    let useInteractionCodeFlow;
 
     if (actionPath === 'cancel') {
+      const authClient = this.options.settings.getAuthClient();
+      const transactionMeta = authClient.transactionManager.load();
+      useInteractionCodeFlow = !!transactionMeta?.interactionHandle;
```
aarongranick-okta, line 140: Why are we using presence of interactionHandle in storage as "source of truth" instead of settings.get('useInteractionFlow')? How could we have transaction storage containing an interactionHandle if our config does not have `useInteractionCodeFlow`?

Because the widget is currently accepting an option, `interactionHandle` which is accepted instead of `useInteractionCodeFlow`:
https://github.com/okta/okta-signin-widget/blob/master/src/v2/client/startLoginFlow.js#L37
https://github.com/okta/okta-signin-widget/blob/master/src/models/Settings.js#L247

There is a task to remove this option, as it has been deemed an anti-pattern: OKTA-405474
Please add a code comment here
```
// TODO: use settings as SoT for `useInteractionCodeFlow` OKTA-405474
```

src/v2/controllers/FormController.js
```diff
@@ -156,6 +161,12 @@ export default Controller.extend({
         .then(this.handleIdxSuccess.bind(this))
         .catch(error => {
           this.showFormErrors(this.formView.model, error);
+        })
+        .then(() => {
+          if (actionPath === 'cancel' && useInteractionCodeFlow) {
+            this.options.settings.set('useInteractionCodeFlow', true);
```
aarongranick-okta, line 167: Please add a code comment here
```
// TODO: remove this line after OKTA-405474
```

src/v2/BaseLoginRouter.js
denysoblohin-okta: `Controller` will be undefined only when `render` is called on `interactionCanceled` handler:
https://github.com/okta/okta-signin-widget/blob/d3b02d7a0a3de1befdd71eddd5948d9fc64d8c95/src/v2/BaseLoginRouter.js#L83
In this case `this.controller` should be already defined.

pradeepdewda-okta: It is because when you trigger `interactionCanceled` the `controller` context should be passed and then you can call render on that context?

src/v2/controllers/FormController.js
```diff
@@ -156,6 +161,13 @@ export default Controller.extend({
         .then(this.handleIdxSuccess.bind(this))
         .catch(error => {
           this.showFormErrors(this.formView.model, error);
+        })
+        .then(() => {
+          if (actionPath === 'cancel' && useInteractionCodeFlow) {
```
pradeepdewda-okta, line 166: does this need to be taken care in `handleSaveForm` as well ?

src/v2/controllers/FormController.js
denysoblohin-okta: We can't rely on `useInteractionFlow` because it's cleared on this line:
https://github.com/okta/okta-signin-widget/blob/9d9dd22e24a2938548ecfbeb3b69f16a8bfb20fd/src/v2/BaseLoginRouter.js#L219

src/v2/BaseLoginRouter.js
pradeepdewda-okta: `Back to sign in` is basically a redirect to login page and also clears any stateToken stored in sessionStorage. Can't we do something similar for `interactionCodeFlow` ? like just delete any existing token and do a page refresh to restart the login flow

src/v2/controllers/FormController.js
aarongranick-okta: That's odd, I wonder why it is being unset. Will be investigated along with OKTA-405474.

Can you also tag these lines with 
```
// TODO: do not unset  `useInteractionCodeFlow` OKTA-405474
```

test/testcafe/spec/SessionStorage_spec.js
denysoblohin-okta: Done. It has not broken tests

src/v2/controllers/FormController.js
denysoblohin-okta: Added

src/v2/BaseLoginRouter.js
denysoblohin-okta: Reverted changes to `render` method. Pass controller as context to `interactionCanceled` event

src/v2/controllers/FormController.js
denysoblohin-okta: In my understanding, no. As we need to restart login flow only on cancel, which is triggered from here
https://github.com/okta/okta-signin-widget/blob/194575de499cb4ce91453362a517623256dbab87/src/v2/view-builder/components/Link.js#L57
and handled in `handleInvokeAction`

pradeepdewda-okta: Why only on `cancel` ? can there be token/code expiration ? what if enduser do a hard refresh on this page when token is expired ?

denysoblohin-okta: If token is expired, user will see screenshot 1. Click on 'Back to sign-in' will restart login flow correctly.
Screenshot 2021-07-12 at 22 19 17

If user will do a hard refresh with expired token, user will see screenshot 2. Click on  'Back to sign-in' will do nothing as it's just goto link. I fixed this in last commit by setting `actionPath: cancel` and removing `'href': settings?.get('baseUrl'),` 
Screenshot 2021-07-12 at 22 19 46

WDYT?

src/v2/view-builder/utils/LinksUtil.js
```diff
@@ -117,14 +117,15 @@ const getSignOutLink = (settings, options = {}) => {
...
+const getBackToSignInLink = () => {
   return [
     {
+      'actionPath': 'cancel',
       'type': 'link',
       'label': loc('goback', 'login'),
       'name': 'go-back',
       // TODO: OKTA-381328 back to baseUrl only works for default login page
-      'href': settings?.get('baseUrl'),
+      //'href': settings?.get('baseUrl'),
```
pradeepdewda-okta, line 128: why this ? this can't be commented.. It will break `Back to sign in` link for various flows.

denysoblohin-okta: I've removed `href` and added `actionPath` because they can't work together according to this code:
https://github.com/okta/okta-signin-widget/blob/56930aa682bf8e0aee294b5a1c8588c5b0d4c6a8/src/v2/view-builder/components/Link.js#L41-L57

pradeepdewda-okta: @denysoblohin-okta there is already a version for that `getSignOutLink`

denysoblohin-okta: Another possible solution I see is to catch specific `idx.session.expired` error and only in this case trigger `cancel` action on click on 'Back to sign-in' , in other cases use href

src/v2/controllers/FormController.js
```diff
@@ -156,7 +161,22 @@ export default Controller.extend({
         .then(this.handleIdxSuccess.bind(this))
         .catch(error => {
           this.showFormErrors(this.formView.model, error);
+        })
+        .then(() => {
```
jmelberg-okta, line 165: This is going to create problems for us, as it appears we executing this handler after the resolved promise from the [`handleIdxResponseSuccess`](https://github.com/okta/okta-signin-widget/blob/55a5d9cad217477abe56de8ab5dd828546e54888/src/v2/BaseLoginRouter.js#L85) method.

We already check for the `actionPath` above (see line 141) - why can't this logic sit there?

test/testcafe/spec/SessionStorage_spec.js
```diff
@@ -154,7 +156,7 @@ test.requestHooks(identifyChallengeMock)('shall clear session.stateHandle when c
 
   // Go back to Identify page after sign-out
   await t.expect(identityPage.form.getTitle()).eql('Sign In');
-  await t.expect(getStateHandleFromSessionStorage()).eql(null);
+  await t.expect(getStateHandleFromSessionStorage()).notEql(xhrEmailVerification.stateHandle);
```
jmelberg-okta, line 159: This test is asserting the `stateHandle` is empty, so we're breaking existing behavior with this change.

What do we expect this value to be now?

test/testcafe/spec/SessionStorage_spec.js
```diff
@@ -261,3 +263,59 @@ test.requestHooks(credentialSSONotExistLogger, credentialSSONotExistMock)('shall
...
+  pageTitle = challengeEmailPageObject.form.getTitle();
+  await t.expect(pageTitle).eql('Verify with your email');
+  await challengeEmailPageObject.verifyFactor('credentials.passcode', '1234');
+  await challengeEmailPageObject.clickNextButton();
+  
+  // Check success
+  const history = await t.getNativeDialogHistory();
+  await t
+    .expect(history.length).eql(1)
+    .expect(history[0].text).eql('SUCCESS: OIDC with single responseType. Check Console');
```
jmelberg-okta, line 320: Instead of checking for a prompt, we can get more value from parsing the console logs to validate the correct [response](https://github.com/okta/okta-signin-widget/blob/17682c785c97f00a6e53911c455c61841b2798ae/playground/main.js#L62).

src/v2/BaseLoginRouter.js
```diff
@@ -215,13 +216,15 @@ export default Router.extend({
...
           this.settings.unset('proxyIdxResponse');
+          // TODO: do not unset  `useInteractionCodeFlow` OKTA-405474
           this.settings.unset('useInteractionCodeFlow');
           this.appState.trigger('remediationSuccess', idxResp);
           this.render(Controller, options);
         })
         .catch(errorResp => {
           this.settings.unset('stateToken');
           this.settings.unset('proxyIdxResponse');
+          // TODO: do not unset  `useInteractionCodeFlow` OKTA-405474
```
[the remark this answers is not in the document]
denysoblohin-okta: Adding this comment was requested by @aarongranick-okta : https://github.com/okta/okta-signin-widget/pull/2090#discussion_r667134521
Option `useInteractionCodeFlow ` will be removed in future.

src/v2/controllers/FormController.js
```diff
@@ -133,6 +133,11 @@ export default Controller.extend({
   handleInvokeAction(actionPath = '') {
     const idx = this.options.appState.get('idx');
 
+    // TODO: use useInteractionCodeFlow from settings instead after OKTA-405474
+    const authClient = this.options.settings.getAuthClient();
+    const transactionMeta = authClient.transactionManager.load();
```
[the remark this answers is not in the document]
denysoblohin-okta: Because `useInteractionCodeFlow` is unset after starting login flow here:
https://github.com/okta/okta-signin-widget/blob/56930aa682bf8e0aee294b5a1c8588c5b0d4c6a8/src/v2/BaseLoginRouter.js#L218
The only way I see to "compute" this property after being deleted is to check `interactionHandle` in transaction meta.
And to load transaction meta we need to use auth-js's manager: 
https://github.com/okta/okta-signin-widget/blob/56930aa682bf8e0aee294b5a1c8588c5b0d4c6a8/src/v2/client/interactionCodeFlow.js#L22

src/v2/controllers/FormController.js
denysoblohin-okta: I assumed that first we need to cancel current login flow, and then after `/cancel` response we can  start a new one.

denysoblohin-okta: But this logic can be moved to line 141

test/testcafe/spec/SessionStorage_spec.js
denysoblohin-okta: Because of starting new login flow after cancel (the purpose of this PR), state handle should be a new one.

src/v2/controllers/FormController.js
```diff
@@ -156,7 +161,22 @@ export default Controller.extend({
...
+        .then(() => {
+          if (actionPath === 'cancel') {
+            // TODO: remove this line after OKTA-405474
+            if (useInteractionCodeFlow) {
+              this.options.settings.set('useInteractionCodeFlow', true);
+            }
+            this.options.appState.trigger('interactionCanceled', this.constructor);
+          }
         });
+    } else if (actionPath === 'cancel') {
```
haishengwu-okta, line 174: Is there real use case for this code branch? 
IIRC, There is no way that widget triggers `cancel` action but `actions[cancel]` is not function.

src/v2/controllers/FormController.js
```diff
@@ -156,7 +161,22 @@ export default Controller.extend({
         .then(this.handleIdxSuccess.bind(this))
```
haishengwu-okta, line 161: I wonder if you shall suppose to change here instead of adding another `.then` ?
```
.then(function () {
  if (action === 'cancel' AND useInteractionFlow) {
    this.options.appState.trigger('interactionCanceled');
  } else {
     this.handleIdxSuccess();
  }
});
```

src/v2/controllers/FormController.js
```diff
@@ -156,7 +161,22 @@ export default Controller.extend({
...
         .catch(error => {
           this.showFormErrors(this.formView.model, error);
+        })
+        .then(() => {
+          if (actionPath === 'cancel') {
+            // TODO: remove this line after OKTA-405474
+            if (useInteractionCodeFlow) {
+              this.options.settings.set('useInteractionCodeFlow', true);
+            }
+            this.options.appState.trigger('interactionCanceled', this.constructor);
```
haishengwu-okta, line 171: why need to pass `this.constructor` ?

src/v2/controllers/FormController.js
```diff
@@ -133,6 +133,11 @@ export default Controller.extend({
   handleInvokeAction(actionPath = '') {
     const idx = this.options.appState.get('idx');
 
+    // TODO: use useInteractionCodeFlow from settings instead after OKTA-405474
+    const authClient = this.options.settings.getAuthClient();
+    const transactionMeta = authClient.transactionManager.load();
+    const useInteractionCodeFlow = !!transactionMeta?.interactionHandle;
```
haishengwu-okta, line 139: Could create a help function in `client/transactionMeta.js`.

test/testcafe/spec/SessionStorage_spec.js
denysoblohin-okta: It would be great, but TestCafe can't return object used in `console.log`, it returns string `[object Object]`
There is open issue for this:
https://github.com/DevExpress/testcafe/issues/5425

src/v2/controllers/FormController.js
denysoblohin-okta: Moved this logic instead of `this.handleIdxSuccess`  as suggested by @haishengwu-okta 
https://github.com/okta/okta-signin-widget/pull/2090#discussion_r668471099

src/v2/controllers/FormController.js
denysoblohin-okta: Done

src/v2/controllers/FormController.js
denysoblohin-okta: Because this will be passed as first parameter (`Controller`) to `BaseLoginRouter.render()` 
https://github.com/okta/okta-signin-widget/blob/1bca146103174f2993c2412b3fe7bc7a3e7fd7a8/src/v2/BaseLoginRouter.js#L83
https://github.com/okta/okta-signin-widget/blob/1bca146103174f2993c2412b3fe7bc7a3e7fd7a8/src/v2/BaseLoginRouter.js#L189

src/v2/controllers/FormController.js
denysoblohin-okta: Done

src/v2/client/transactionMeta.js
```diff
@@ -95,4 +95,10 @@ export function isTransactionMetaValid(settings, meta) {
   }
 
   return true;
-}
\ No newline at end of file
+}
+
+export function useInteractionCodeFlow(settings) {
+  const authClient = settings.getAuthClient();
+  const transactionMeta = authClient.transactionManager.load();
```
jmelberg-okta, line 102: Can we reuse the [`getTransactionMeta`](https://github.com/okta/okta-signin-widget/blob/c07ce5a87879a4bf98faf40cf8239eda8f4d86e9/src/v2/client/transactionMeta.js#L21) method here?

test/testcafe/spec/SessionStorage_spec.js
jmelberg-okta: Right, my comment was mostly focused on ensuring the assertions match the description.

Let's make sure we update the test description and have an assertion for the new `stateHandle` we'd expect.

test/testcafe/spec/SessionStorage_spec.js
jmelberg-okta: We should still be able to compare object equality though, as we're doing using the `checkConsoleMessages` helper. Here is an example where we're [checking for context](https://github.com/okta/okta-signin-widget/blob/d0f55fee688fe6c79620cf7e37e7a53b236311e4/test/testcafe/spec/IdentifyRegistrationWithCaptcha_spec.js#L49-L52).

jmelberg-okta (commented):
Overall these changes appear good to me. Let's wait for final approval from Aaron before proceeding with this change.

package.json
```diff
@@ -40,7 +40,7 @@
     "test:e2e": "yarn build:release && grunt test-e2e",
     "test:e2e:windows": "export SAUCE_PLATFORM_NAME=windows && yarn test:e2e",
     "test:testcafe-setup": "mkdir -p test-reports/testcafe && yarn start --mock mockDuo --open false",
-    "test:testcafe-run": "yarn test -t testcafe -c 2 chrome:headless --reporter xunit:build2/reports/junit/testcafe-results.xml",
```
haishengwu-okta, line 43: assume this is not intentional?

denysoblohin-okta: Yes, reverted.
(I changed this temporary to see Testcafe errors in Travis)

src/v2/controllers/FormController.js
denysoblohin-okta: Will move fix for expired session to another PR

src/v2/view-builder/utils/LinksUtil.js
denysoblohin-okta: Reverted these changes. Will move fix for expired session to another PR

src/v2/controllers/FormController.js
denysoblohin-okta: Reverted these changes. Will move fix for expired session to another PR

src/v2/client/transactionMeta.js
denysoblohin-okta: Yes.. but I think it's better to not use `getTransactionMeta` because it's async and will create new transaction meta if one does not exist

test/testcafe/spec/SessionStorage_spec.js
denysoblohin-okta: Changed.
See https://github.com/okta/okta-signin-widget/pull/2090/files#diff-d5619ca161ed5c74752d62d68598cc9038e1a588eee3755c9f3a103ed3ad43e0R317

aarongranick-okta (approved):

pradeepdewda-okta (changes requested):
Beta version of this change needs to pass bacon first..

src/v2/controllers/FormController.js
pradeepdewda-okta: Let's setup a meeting to discuss this change.. I am not comfortable with this one-off chaining here for interaction flow.. why it needs a special treatment.

test/testcafe/spec/SessionStorage_spec.js
```diff
@@ -261,3 +263,88 @@ test.requestHooks(credentialSSONotExistLogger, credentialSSONotExistMock)('shall
   )).eql(1);
   await t.expect(getStateHandleFromSessionStorage()).eql(null);
 });
+
+test.requestHooks(identifyChallengeMock)('shall back to sign-in and authenticate succesfully', async t => {
```
pradeepdewda-okta, line 267: typo: succesfully

test/testcafe/spec/SessionStorage_spec.js
```diff
@@ -3,14 +3,16 @@ import xhrEmailVerification from '../../../playground/mocks/data/idp/idx/authent
 import xhrSessionExpried from '../../../playground/mocks/data/idp/idx/error-session-expired';
 import xhrIdentify from '../../../playground/mocks/data/idp/idx/identify';
 import xhrSuccess from '../../../playground/mocks/data/idp/idx/success';
+import xhrSuccessWithInteractionCode from '../../../playground/mocks/data/idp/idx/success-with-interaction-code';
+import xhrSuccessTokens from '../../../playground/mocks/data/oauth2/success-tokens.json';
```
pradeepdewda-okta, line 7: remove `.json`

playground/main.js
```diff
@@ -55,11 +55,11 @@ const renderPlaygroundWidget = (options = {}) => {
       //    an array of tokens or a single token, depending on the
       //    initial configuration.
       else if (Array.isArray(res)) {
-        console.log(res);
+        console.log('===== playground widget success =====', res);
```
jmelberg-okta, line 58: I don't think we need this extra notice for success, as it'll complicate the assertion in our testcafe helpers.

I suggest keeping this as it was before, and letting the `JSON.parse(log[i])` logic handle parsing it so we can check the content through `checkConsoleMessages`

pradeepdewda-okta: +1

pradeepdewda-okta:
@denysoblohin-okta when you click on `Back to sign in` what does `interact` request use as `interactionHandle` ? is it using a stale token from previous request.. basically upon `cancel` everything should be wiped out..

playground/main.js
denysoblohin-okta: Changed to parsing JSON

denysoblohin-okta:
@pradeepdewda-okta `/interact` will return new interaction handle

pradeepdewda-okta:
> @pradeepdewda-okta `/interact` will return new interaction handle. `code_challenge ` and `state` in request are new

Was it null when it does an `interact` without your changes.. ?

denysoblohin-okta:
@pradeepdewda-okta Without my changes after `cancel` there is no `interact` request

src/v2/controllers/FormController.js
pradeepdewda-okta: @denysoblohin-okta will create jira to remove exclusive handling of cancel action for `interaction code flow` for devex team to resolve

pradeepdewda-okta (approved):
As part of OKTA-405474, TODO's should be fixed

src/v2/BaseLoginRouter.js
```diff
@@ -215,19 +210,33 @@ export default Router.extend({
...
+        .finally(() => {
+          this._render(Controller, options);
         });
     }
 
+    this._render(Controller, options);
+  },
+
+  _render: function(Controller, options = {}) {
+    // Since we have a wrapper view, render our wrapper and use its content
```
pradeepdewda-okta, line 229: can you explain why we need this in comments, instead what it is doing?

src/v2/BaseLoginRouter.js
```diff
@@ -215,19 +210,29 @@ export default Router.extend({
...
-          this.settings.unset('useInteractionCodeFlow');
           this.appState.trigger('remediationError', errorResp.error || errorResp);
-          this.render(Controller, options);
+        })
+        .finally(() => {
+          this._render(Controller, options);
         });
     }
 
+    this._render(Controller, options);
```
pradeepdewda-okta, line 225: if we are doing the `render` in finally why we need it here as well?

denysoblohin-okta: The problem I'm trying to solve is the recursive call of `this.render` in then/catch.
If we not unset `useInteractionCodeFlow `, `startLoginFlow` will be called again and result in infinite loop.
`useInteractionCodeFlow ` setting was loop-breaker.

Instead of this, I moved actual render logic into `_render` method and call it in `finally`.

src/v2/BaseLoginRouter.js
denysoblohin-okta: I just moved these lines from `render` method into `_render`

pradeepdewda-okta: i mean why it needs to be in `render`

src/v2/BaseLoginRouter.js
pradeepdewda-okta: there seems to be some flaw with this flow, i am not super confident with this PR.. @aarongranick-okta you should take a look on why `useInteractionCodeFlow` unset is a loop breaker. It should not have that dependency.

aarongranick-okta: It seems that `render` is following a recursive pattern, used here to load languages:
https://github.com/okta/okta-signin-widget/blob/master/src/v2/BaseLoginRouter.js#L206

The lines of code to "unset" stateToken, etc. seem to be related to this function, which is not designed to be called recursively.
https://github.com/okta/okta-signin-widget/blob/master/src/v2/client/startLoginFlow.js#L26

Conceptually we want render to move forward, not in a loop. One way to accomplish this would be to use `async/await`

- await LanguageUtil.loadLanguage()
- await startLoginFlow
- continue with normal render

Although some methods return Q promises, I believe it should work. (Some tests may need minor adjustments).

src/v2/BaseLoginRouter.js
```diff
@@ -192,6 +192,7 @@ export default Router.extend({
     if (sessionStorageHelper.getLastInitiatedLoginUrl() !== window.location.href) {
       sessionStorageHelper.removeStateHandle();
     }
+
```
pradeepdewda-okta, line 195: remove this line

test/unit/spec/v2/client/transactionMeta_spec.js
```diff
@@ -231,4 +231,5 @@ describe('v2/client/transactionMeta', () => {
       expect(isTransactionMetaValid(testContext.settings, testContext.transactionMeta)).toBe(true);
     });
   });
+
```
pradeepdewda-okta, line 234: nit-- use lint to remove unnecessary changes from PR

aarongranick-okta (approved):
lgtm

PR #2091, opened 2021-07-06
fix: Show error when email OTP submission hit org ratelimit

worldleelavivat-okta:
## Description:
- Show error when email OTP submission hit org ratelimit

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
otp_ratelimit

### Reviewers:
- @pradeepdewda-okta 
- @haishengwu-okta 
- @mauriciocastillosilva-okta 
- @jmelberg-okta 
- @cred-okta 

### Issue:

- [OKTA-408448](https://oktainc.atlassian.net/browse/OKTA-408448)


pradeepdewda-okta:
@worldleelavivat-okta can this PR wait for couple of weeks ? We are making changes on how widget handle errors and how polling works..

worldleelavivat-okta:
@pradeepdewda-okta OKTA-408448 is marked for `OIE-Tuscany` so it needed to go in this release before July 9th CF.

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -53,6 +53,7 @@ const Body = BaseFormWithPolling.extend(Object.assign(
     },
     initialize() {
       BaseFormWithPolling.prototype.initialize.apply(this, arguments);
+      this.formSubmitted = false;
```
pradeepdewda-okta, line 56: instead of controlling the form state.. before next poll request the error should be cleared.

worldleelavivat-okta: @pradeepdewda-okta 
I don't think we should clear the error before the next poll since the next poll could occur in a few seconds which mean the error would stay on for only a few seconds.
Polling should be somethings in a background so I think we should keep it in the form state so things only get update when user interact with the UI rather than update the UI based on something in the background.

pradeepdewda-okta: There is nothing like background polling. Widget polls if there is a poll remediation, if poll contains a form error then API should send a form error with remediation ie. [ remediation + error ]..

So what i am suggesting is in this particular case next poll happens in 60 sec then the error will remain there for 60sec and as soon as we get the remediation from next poll, then new error or state should be rendered..

pradeepdewda-okta: @mauriciocastillosilva-okta @jmelberg-okta can you guys please help review this PR ??

worldleelavivat-okta: @pradeepdewda-okta 
If we clear error before polling in this case, then error would only stay on till next poll which is 60 sec. But if the next poll has no error, shouldn't the error stay on longer?

pradeepdewda-okta: if next response contains an error then ideally widget should render the error again...

worldleelavivat-okta: I mean if the next polling response contains no error, shouldn't the previous error stay (if it is from user interaction by submitting OTP)?

pradeepdewda-okta: Ideally no. new remediation ==> new screen (UI).

thomasmaslen-okta:
> @pradeepdewda-okta [OKTA-408448](https://oktainc.atlassian.net/browse/OKTA-408448) is marked for `OIE-Tuscany` so it needed to go in this release before July 9th CF.

I know that this JIRA got tagged as `OIE-Tuscany`, but since the problem only occurs if you hit the org rate limit, have the relevant authorities (@shayekhazaeli-okta ?) determined that yes, this positively definitely needs to be in `OIE-Tuscany`?  (Quite possibly they have done so and I'm just not up to speed).

worldleelavivat-okta:
@thomasmaslen-okta from talking to @shayekhazaeli-okta previously, this one is required for `OIE-Tuscany`

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
mauriciocastillosilva-okta: I am not entirely following: why are we polling if there is a rate limit error? iIf we get a 429 response, there is no remediation that says we should keep polling, so we should stop there and show the error message we have that says "Try again later" and stop any other processing.

worldleelavivat-okta: @mauriciocastillosilva-okta 
Current behaviour is that if poling encounter rate limit error, SIW should:
1. Stop polling and wait 60 secs before start polling again (60 sec should be enough time for rate limit to reset)
2. Hide rate limit error if polling encounter that since it being handled in number 1.

The bug is that current logic doesn't distinguish rate limit error between when user OTP submission (saving form) and when polling. So rate limit error is always hidden to user during OTP submission. 

This logic should help with that. Note that `triggerAfterError(model, error)` isn't aware of what is causing error (user vs polling), so the solution is tracking of it in the form state. (Rename `isPollingRateLimitError` to `isRateLimitError` since it a little misleading).So if the user submit the OTP and encounter rate limit error, the error should stay on and the polling would pause for 60 sec to give time for rate limit to reset.

pradeepdewda-okta: @worldleelavivat-okta the behavior to resume polling after 1 min should be dictated by API response, Widget is NOT going to handle such one offs.. with JWE token introduction where stateToken can change between any two remediation this is going to break.

worldleelavivat-okta: @pradeepdewda-okta 
While I agree as it should be handled more gracefully by API response in the future when JWE token is introduced. This bug fix is for `OIE-Tuscany` so can we leave that as part of future enhancement as part of JWE token introduction as it would change the current behaviour that was signed off from Product before?

pradeepdewda-okta: @worldleelavivat-okta Well it's not JWE token specific. JWE token actually magnifies the issue with this implementation and a genuine use case where stateToken changes.. Currently this flow works because there is no change in stateToken, but you should not assume that stateToken is persistent even in current idx implementation.. 

See if you can change this to `terminal` error state, if changing API is not possible.. `Resuming` polling without IDX remediation is a NO NO.

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -82,21 +84,24 @@ const Body = BaseFormWithPolling.extend(Object.assign(
         return;
       }
 
-      if (this.isPollingRateLimitError(error)) {
+      if (this.isRateLimitError(error)) {
         // When polling encounter rate limit error, wait 60 sec for rate limit bucket to reset
         // before polling again & hide error message
-        setTimeout(() => {
-          model.trigger('clearFormError');
```
mauriciocastillosilva-okta, line 89: an idea, can we just not clear the errors here? As in, if we have a rate limit error, then don't clear any error messages, doesn't matter if we are polling or not. That way we will keep the error message for 1 minute, AND we don't have to keep a boolean in this view to keep status. At the end, a rate limit error is not something we need to hide the message

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -82,21 +84,24 @@ const Body = BaseFormWithPolling.extend(Object.assign(
...
-        }, 0);
+        if (this.formSubmitted === false) {
```
mauriciocastillosilva-okta, line 90: instead of using a client side boolean to keep status, we can use the current polling status we have
```
triggerAfterError(model, error) {
     const isFormPolling = this.polling;
     this.stopPolling();

      if (error.responseJSON?.errorSummaryKeys?.includes('idx.session.expired')) {
        // Do NOT resume polling since session is invalid and polling is already stopped
        return;
      }

      if (this.isRateLimitError(error)) {
        // When polling encounter rate limit error, wait 60 sec for rate limit bucket to reset
        // before polling again & hide error message
        if (!isFormPolling) {
```

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
worldleelavivat-okta: We would like to hide error if it is not from user interactions, in otherwords, if it is from polling, do not show rate limit error.

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
worldleelavivat-okta: sound good to me. Let me update and test it.

worldleelavivat-okta: Unfortunately, `saveForm` always call `this.stopPolling()` so we would still need additional variable in form state to know if it is user submitted.

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -77,6 +75,7 @@ const Body = BaseFormWithPolling.extend(Object.assign(
 
     triggerAfterError(model, error) {
       BaseForm.prototype.triggerAfterError.apply(this, arguments);
+      const isFormPolling = this.polling;
```
pradeepdewda-okta, line 78: wondering we `stopPolling` after and assigned variable before?

src/v2/view-builder/views/shared/polling.js
```diff
@@ -50,6 +50,7 @@ export default {
   stopPolling() {
     if (this.polling) {
       clearInterval(this.polling);
+      this.polling = false;
```
pradeepdewda-okta, line 53: change to `this.polling = null`

worldleelavivat-okta: updated

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
worldleelavivat-okta: It is to be on the safer side so we won't have multiple polling going without a way to cancel.

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
worldleelavivat-okta: For reference, discussion in: https://okta.slack.com/archives/C0131CX1MGB/p1625683893070300

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
mauriciocastillosilva-okta: nit: `const isFormPolling = !!this.polling;`

mauriciocastillosilva-okta (approved):
just a nit, after that looks good

mauriciocastillosilva-okta (commented):
last commit looks good ⛵

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -91,9 +91,7 @@ const Body = BaseFormWithPolling.extend(Object.assign(
             model.trigger('clearFormError');
           }, 0);
         }
-        setTimeout(() => {
-          this.startPolling(this.options.appState.get('dynamicRefreshInterval'));
-        }, 60000);
+        this.startPolling(60000);
```
pradeepdewda-okta, line 94: that would be polling every 1 min ?

worldleelavivat-okta: It would for 1 min and revert back to interval based on the next polling response.

PR #2094, opened 2021-07-06
feat: Adds enrollment and challenge flows for YubiKey on OIE

jmelberg-okta:
## Description:

- Adds happy-path support for the YubiKey authenticator enrollment/challenge on OIE
- The various error scenarios will be added once the feature is server-side complete.

## PR Checklist

- [x] Added e2e tests

### Screenshot/Video:

| Enroll | Challenge | Pseudo-Loc |
| ------ | --------- | ------------ |
| Enroll | Challenge |  | 
| Screen Shot 2021-07-07 at 10 04 26 AM | Screen Shot 2021-07-07 at 10 11 41 AM | Screen Shot 2021-07-07 |

### Reviewers:

- @magizh-okta 

### Issue:

- [OKTA-409867](https://oktainc.atlassian.net/browse/OKTA-409867)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -930,6 +930,13 @@ oie.symantecVip.challenge.title = Verify with {0}
 oie.symantecVip.challenge.description = Enter the generated security code from the {0} app.
 oie.symantecVip.verify.passcode.label = Enter security code
 
+## YubiKey Authenticator
+oie.yubikey.authenticator.description = Verify your identity using YubiKey
+oie.yubikey.enroll.title = Set up YubiKey
+oie.yubikey.challenge.title = Verify with YubiKey
+oie.yubikey.description = Insert the YubiKey into a USB port and tap it to generate a verification code.
+oie.yubikey.passcode.label = Insert then tap your YubiKey
```
jmelberg-okta, line 938: These keys are consistent with our [mock files](https://www.figma.com/file/I1ebb1ApKB1j7EkIaAeWFH/OIE-Authenticators?node-id=570%3A1155), and are currently displaying slightly different text than V1.

@nikhilvenkatraman-okta - Can you remind me where we stand with v1 customizations in regards to the "Set up" vs. "Setup" conversation? If we're OK with using the existing wording (and are OK with removing the period), we can eliminate two extra keys and keep v1 customizations.

assets/sass/v2/_yubikey.scss
```diff
@@ -0,0 +1,8 @@
+.yubikey-demo {
+  background: url('../img/icons/mfa/yubikeyDemo.png') no-repeat center center;
+  background-size: contain;
+  width: 100%;
+  min-height: 102px;
+  margin-bottom: 20px;
+  background-color: 'red';
+}
```
jmelberg-okta, line 8: FYI this is [identical to v1](https://github.com/okta/okta-signin-widget/blob/master/assets/sass/modules/_enroll.scss#L122-L128), just scoping it within our v2 code.

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
```diff
@@ -0,0 +1,107 @@
...
+import xhrSuccess from '../../../playground/mocks/data/idp/idx/success';
+
+const logger = RequestLogger(/introspect/,
+  {
+    logRequestBody: true,
+    stringifyRequestBody: true,
+  }
+);
+
+const enrollMock = RequestMock()
```
pradeepdewda-okta, line 15: does it have any terminal error states ? or any form errors which are coming from API ?

jmelberg-okta: It does, but the enduser facing errors are still under active development. This PR is to provide "happy path" support/testing.

We'll have a follow-up task to handle various error states once they become available.

pradeepdewda-okta: Cool. I just wanted to make sure that form error has correct idx responses. that is (remediation + error), widget should no longer render a terminal idx response as form error.

test/unit/spec/v2/ion/uiSchemaTransformer_spec.js
```diff
@@ -763,6 +789,14 @@ describe('v2/ion/uiSchemaTransformer', function() {
                     authenticatorKey: 'symantec_vip',
                     relatesTo: XHRAuthenticatorEnrollSelectAuthenticators.authenticators.value[11]
                   },
+                  {
+                    label: 'YubiKey Authenticator',
```
magizh-okta, line 793: The label here shows up in select authenticator list UI and this doesnt seem to be translated or mapped to a i18n string in SIW. Also do we want to call this  'Yubikey Security Method' ?

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
```diff
@@ -0,0 +1,23 @@
...
+  getPageTitle() {
+    return this.form.getElement('.okta-form-title').textContent;
+  }
+
+  getPageSubtitle() {
+    return this.form.getElement('.okta-form-subtitle').textContent;
+  }
+
+  submit() {
+    return this.form.clickSaveButton();
```
magizh-okta, line 21: I think they are in base page object itself

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
```diff
@@ -0,0 +1,107 @@
+import { RequestMock, RequestLogger } from 'testcafe';
```
magizh-okta, line 1: minor nit we usually have enroll and challenge flows tested separately 

eg) 
ChallengeAuthenticatorDuo_spec.js
EnrollAuthenticatorDuo_spec.js

magizh-okta (commented):
Looking good. Just some nits.

test/unit/spec/v2/ion/uiSchemaTransformer_spec.js
jmelberg-okta: This was a miss on my end, as I didn't map the label to the `select-authenticator-enroll` and `select-authenticator-authenticate` forms - thank you for catching that! Pushing up a new commit in a few to resolve it.

General rule I've been following is **if** the name of the authenticator is customizable/editable, then we wan't to leverage the API response in favor of relying on our own i18n translation. For this authenticator, it shouldn't be modifiable.

In regards to "Security Method" vs "Authenticator", we are only changing occurrences **where more than one authenticator** is being referenced. Here, this should remain `YubiKey Authenticator`.

Great questions!

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
jmelberg-okta: The convention I tend to follow (and what appears to be present in this codebase as well) is there is a `_spec` file for every source file. Unlike most of our other authenticators, YubiKey only has one major difference between enroll and verify - the `title` text. To reduce the number of files and duplicate code, we opted for a single View Controller that conditionally renders the title given the scenario:

https://github.com/okta/okta-signin-widget/blob/9172dd832540e3fcf8c6225ab4a7e6290ed18b92/src/v2/view-builder/views/yubikey/AuthenticatorYubiKeyView.js#L12-L15

Given this, I'd prefer to couple them together. Any objection?

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
jmelberg-okta: Not that I could find. It looks like the `BasePageObject` provides helpers for header/footer interactions, as opposed to the body content.

I [created a ticket](https://oktainc.atlassian.net/browse/OKTA-410074) to have a `BaseAuthenticatorPageObject`, which should help us reduce the number of extended classes. I'll plan on taking this on next week.

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
magizh-okta: Well these are e2e test and test file name doesnt have to map to the src file name.  I would slightly prefer it to be isolated tests or have a common file name like YubikeyAuthenticatorEnrollChallenge_spec.js


I am fine as is too.

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
magizh-okta: I can find methods for title and subtitle. 

https://github.com/okta/okta-signin-widget/blob/master/test/testcafe/framework/page-objects/BasePageObject.js#L69

jmelberg-okta: Ah - somehow I overlooked those. Updated!

magizh-okta (approved):

PR #2097, opened 2021-07-07
Hw fix polling n error

haishengwu-okta:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:


src/v2/models/AppState.js
```diff
@@ -236,17 +241,22 @@ export default Model.extend({
...
     // clear cache for derived props.
     this.trigger('cache:clear');
   },
 
   setIonResponse(transformedResponse) {
-    if (!this.shouldReRenderView(transformedResponse)) {
-      return;
-    }
+    const doRerender = this.shouldReRenderView(transformedResponse);
+    this.clearAppStateCache(doRerender);
```
pradeepdewda-okta, line 259: this seems to be very cryptic way to re-render the view..  :)

It is much clear to trigger a trigger an event on when to re-render and unbinds re-rendering from changing formName

haishengwu-okta: Could be.
Fix the reset currentFormName logic a bit. Plus we do have comment down below to explain the critical part when change currentFormName.

src/v2/controllers/FormController.js
```diff
@@ -248,6 +249,16 @@ export default Controller.extend({
     let errorObj;
     if (IonResponseHelper.isIonErrorResponse(error)) {
       errorObj = IonResponseHelper.convertFormErrors(error);
+
+      // When error happened, we need to try to refresh `appState` due to `stateHandle` changed.
+      // A little sloppy logic to determine how to determine whether need to refresh appState.
+      // 1. set `hasFormError` to avoid re-render the page
+      // 2. sometime error response is very different (especially no remediation) which is only
```
pradeepdewda-okta, line 256: this doesn't feel right where we are updating `appState` selectively in error cases.. 

**Correct flow here--**
Always update the AppState if there was error response, But widget can choose to render or re-render
If updating AppState in error cases fails testcafe then that is a flaw in that flow. We should fix that flow instead allowing the flow to trigger error without updating the app state.. _Widget should just blindly update AppSate after every idx response._

For eg -- `pause polling when encounter 429 too many request` flow show terminal errors as form error.. Which is ideally not right. If they want the form error then the flow's remediation should dictate that...

PR #2115, opened 2021-07-27
Increasing poll timeout to 5min

vakaevich-okta:
## Description:

When user tries to access an app that requires device context, device probing kicks in and SIW makes a /challenge call on the port that OV is running on. Then OV initiates inline enrollment since there is no account. User waits for some time and clicks cancel on the inline enrollment prompt in OV
Now, on step 2 the http connection timeout from SIW to OV is 3s.
So, OV has no way to respond back to SIW that user cancelled the inline enrollment (if it happens after 3s).

This PR increases the timeout to 5min

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Reviewers:

@santoshmale-okta , @johannesstockmann-okta , @diptishiralkar-okta , @arunthotta-okta , @saurinpandya-okta 

### Issue:

- [OKTA-406443](https://oktainc.atlassian.net/browse/OKTA-406443)

### DIFF
--- a/src/v2/view-builder/views/device/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/device/DeviceChallengePollView.js
@@ -3,7 +3,7 @@ import { BaseForm, BaseFormWithPolling, BaseFooter, BaseView } from '../../inter
 import Logger from '../../../../util/Logger';
 import BrowserFeatures from '../../../../util/BrowserFeatures';
 import Enums from '../../../../util/Enums';
-import { CANCEL_POLLING_ACTION } from '../../utils/Constants';
+import { CANCEL_POLLING_ACTION, CHALLENGE_TIMEOUT } from '../../utils/Constants';
 import Link from '../../components/Link';
 import { doChallenge } from '../../utils/ChallengeViewUtil';
 import OktaVerifyAuthenticatorHeader from '../../components/OktaVerifyAuthenticatorHeader';
@@ -90,7 +90,7 @@ const Body = BaseFormWithPolling.extend(
           url: getAuthenticatorUrl('challenge'),
           method: 'POST',
           data: JSON.stringify({ challengeRequest }),
-          timeout: 3000 // authenticator should respond within 3000ms for challenge request
+          timeout: CHALLENGE_TIMEOUT // authenticator should respond within 5 min (300000ms) for challenge request
         });
       };
 
--- a/src/v2/view-builder/utils/Constants.js
+++ b/src/v2/view-builder/utils/Constants.js
@@ -1,5 +1,6 @@
 export const SHOW_RESEND_TIMEOUT = 30000;
 export const WARNING_TIMEOUT = 30000;
+export const CHALLENGE_TIMEOUT = 300000;
 export const MS_PER_SEC = 1000;
 export const UNIVERSAL_LINK_POST_DELAY = 500;
 export const CANCEL_POLLING_ACTION = 'authenticatorChallenge-cancel';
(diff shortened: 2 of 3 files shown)


src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -90,7 +90,7 @@ const Body = BaseFormWithPolling.extend(
           url: getAuthenticatorUrl('challenge'),
           method: 'POST',
           data: JSON.stringify({ challengeRequest }),
-          timeout: 3000 // authenticator should respond within 3000ms for challenge request
+          timeout: 300000 // authenticator should respond within 5 min (300000ms) for challenge request
```
pradeepdewda-okta, line 93: create a constant and put it in `Constants.js`.

pradeepdewda-okta: see if you need to change any `testcafe` tests..

santoshmale-okta:
We need to test the basic fastpass flows for this change with Windows/MacoS/Android clients and make sure things are working as expected.

src/v2/view-builder/views/device/DeviceChallengePollView.js
jhe-okta: Can you please update the comment to give more context?

vakaevich-okta:
Comments addressed.
@jhe-okta , PR description is updated with more context, please take a look.
@santoshmale-okta , on macOS OV works as expected, but I say let's merge this pretty low risk change, and wait till @saurinpandya-okta executes planned DT bugs verification under Win, macOS and this particular one under Android. wdyt?

pradeepdewda-okta:
so for the next 5 min, if user stays on the screen `/poll` would continue ? Could there be any rate limit issue, since the window is increase to 5 min ( which is a lot )...

vakaevich-okta:
> so for the next 5 min, if user stays on the screen `/poll` would continue ? Could there be any rate limit issue, since the window is increase to 5 min ( which is a lot )...

Ratelimit issue could raise in case we'd multiply the number of requests, not the time to wait for the single response from a user under non-basic flow. Common use case for a user is to answer within seconds, hence if your concern is about the amount of hanging requests that would hit the API ratelimit, - no indications they will.

santoshmale-okta:
>@santoshmale-okta , on macOS OV works as expected, but I say let's merge this pretty low risk change, and wait till @saurinpandya-okta executes planned DT bugs verification under Win, macOS and this particular one under Android. wdyt?

santoshmale-okta:
> so for the next 5 min, if user stays on the screen /poll would continue ? Could there be any rate limit issue, since the window is increase to 5 min ( which is a lot )...  

@pradeepdewda-okta Actually, this PR is not modifying the transaction timeout (which coincidentally is already 5 mins). So, even without this PR, currently, in some corner case end user might end up on a polling screen for 5 minutes. That's why we have a backoff polling logic. This PR is only increasing the connection timeout from SIW to OV so that OV has a chance to respond back to SIW after 3seconds (which can happen in at least one known use case, i.e, when enduser does not want to enroll into OV)

vakaevich-okta:
Sanity testing is done, no regression is found

santoshmale-okta:
@vakaevich-okta can we please merge this into the correct release branch instead of merging into master

vakaevich-okta:
Closing this one in favor of a new one with the proper base branch: https://github.com/okta/okta-signin-widget/pull/2139

PR #2125, opened 2021-07-29
fix(v1): when location starts with data.userProfile in registration api error, render the error summary instead of generic error

yannongli-okta:
## Description:
This is a fix for v1 self service registration inline hook. 

After https://github.com/okta/okta-ui/commit/03387a3452a4f2c65b8e769ca84b27b9a8ea3b68, the new expected behavior for showing error is: when received an api error, a generic error message is shown in error banner and each individual field error is shown down below that field.
Screen Shot 2021-07-23 at 3 50 56 PM

But in v1 inline hooks, the location property was in format that starts with `data.userProfile` according to our developer doc 
https://developer.okta.com/docs/reference/registration-hook/#sample-json-payload-of-request The commit above is a general enhancement and thus didn't cover well on this specific use case, and it's causing a regression where only a generic error message is shown in the error banner.
Screen Shot 2021-07-23 at 3 52 25 PM

The proposed fix within this PR is: when we detect the location property starts with `data.userProfile`, we will replace the error banner with errorSummary returned in api response and the behavior is the same as before.
Screen Shot 2021-07-29 at 3 03 01 PM

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Reviewers:
@sherizada-okta @pradeepdewda-okta 

### Issue:

- [OKTA-409142](https://oktainc.atlassian.net/browse/OKTA-409142)

### DIFF
--- a/src/RegistrationController.js
+++ b/src/RegistrationController.js
@@ -117,7 +117,7 @@ export default BaseLoginController.extend({
 
         if (responseJSON && responseJSON.errorCauses.length) {
           const { errorCode, errorCauses } = responseJSON;
-          const { errorSummary, reason } = errorCauses[0];
+          const { errorSummary, reason, location } = errorCauses[0];
 
           const isNotUniqueValue =
             errorCode === 'E0000001' &&
@@ -126,6 +126,8 @@ export default BaseLoginController.extend({
           if (isNotUniqueValue) {
             this.renderIsNotUniqueError(responseJSON);
           }
+
+          this.renderLegacyLocationErrorIfNeeded(location, errorSummary);
 
           Util.triggerAfterError(
             this,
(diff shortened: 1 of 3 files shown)


src/RegistrationController.js
```diff
@@ -148,6 +150,16 @@ export default BaseLoginController.extend({
     this.$el.find('.okta-form-infobox-error p').text(errorSummary);
   },
 
+  renderLegacyLocationErrorIfNeeded: function(location, errorSummary) {
+    // replace generic error message with errorSummary for v1 SIW
+    // this makes sure that with legacy location that starts with `data.userProfile`
+    // we still see the errorSummary in the error banner instead of only a generic error
+    // See example in https://developer.okta.com/docs/reference/registration-hook/#sample-json-payload-of-request
+    if (location && location.startsWith('data.userProfile')) {
```
sherizada-okta, line 158: `startsWith` is [not supported on ie11](https://caniuse.com/?search=startsWith), a browser version we support, and I'm not sure if SIW's babel version polyfills it. You can try using another method or test ie11 using the Sauce Labs chiclet.

sherizada-okta (commented):
Looks good

yannongli-okta:
closed in favor of https://github.com/okta/okta-signin-widget/pull/2126 to merge to 5.9 branch

PR #2127, opened 2021-08-02
chore: Updates codeowners

jmelberg-okta:
## Description:

- Updates the CODEOWNER file to remove previous owners and scope requests for certain users.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -9,4 +9,9 @@
 #  - [CODEOWNERS] https://help.github.com/articles/about-codeowners
 #  - [Widget] https://github.com/orgs/okta/teams/widget-contributors/members
 
-*    @pradeepdewda-okta @sherizada-okta @mauriciocastillosilva-okta @haishengwu-okta @magizh-okta @jasleennuno-okta @jmelberg-okta @nikhilvenkatraman-okta @gowthamidommety-okta @robertdamphousse-okta @brettritter-okta @TianGan-okta @aarongranick-okta @shuowu-okta @anipendakur-okta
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -9,4 +9,9 @@
 #  - [CODEOWNERS] https://help.github.com/articles/about-codeowners
 #  - [Widget] https://github.com/orgs/okta/teams/widget-contributors/members
 
-*    @pradeepdewda-okta @sherizada-okta @mauriciocastillosilva-okta @haishengwu-okta @magizh-okta @jasleennuno-okta @jmelberg-okta @nikhilvenkatraman-okta @gowthamidommety-okta @robertdamphousse-okta @brettritter-okta @TianGan-okta @aarongranick-okta @shuowu-okta @anipendakur-okta
+*    @pradeepdewda-okta @sherizada-okta @mauriciocastillosilva-okta @haishengwu-okta @magizh-okta @robertdamphousse-okta @TianGan-okta @aarongranick-okta @shuowu-okta
+
+# Localization: pseudo-loc, i18n, etc.
+packages/@okta/pseudo-loc/    @nikhilvenkatraman-okta @jmelberg-okta
+packages/@okta/i18n/          @nikhilvenkatraman-okta @jmelberg-okta
```
jmelberg-okta, line 16: This is a bit aggressive, but will help ensure we catch common issues before we have linting in place (coming soon).

Specifically:

- Use of arguments (`{n}`) include a comment in the `properties` file explaining what the value is in reference to
- No concatenated strings
- Punctuation is present within the resource file instead of embedded within the template
- HTML is owned by the template instead of the resource file

sherizada-okta (approved):

nikhilvenkatraman-okta (approved):

.github/CODEOWNERS
nikhilvenkatraman-okta: We have an item to add a lint rule for the first item. Adding a task for us to add guideline/documentation for the others

PR #2135, opened 2021-08-06
Support email verify flow using OTP; Use AuthJS as IDX client

aarongranick-okta:
## Description:
Support email verify flow using OTP

- Removes okta-idx-js as a dependency and uses AuthJS as the IDX client
- Upgrades to auth-js 6.0.0 (breaking change)
- removes `stateTokenExternalId` option
- adds `otp` option

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

- [] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?

### Issue:

- [OKTA-406314](https://oktainc.atlassian.net/browse/OKTA-406314)


src/v2/BaseLoginRouter.js
```diff
@@ -84,12 +85,21 @@ export default Router.extend({
...
     this.listenTo(this.appState, 'updateAppState', this.handleUpdateAppState);
     this.listenTo(this.appState, 'remediationError', this.handleIdxResponseFailure);
     this.listenTo(this.appState, 'restartLoginFlow', this.restartLoginFlow);
   },
 
+  updateDeviceFingerprint() {
+    const authClient = this.settings.getAuthClient();
+    const fingerprint = this.appState.get('deviceFingerprint');
+    if (fingerprint) {
+      authClient.http.setRequestHeader('X-Device-Fingerprint', fingerprint);
```
haishengwu-okta, line 99: IIRC, we choose interceptor approach was because `X-Device-Fingerprint` only being set whenever there is value in `appState`.
With this approach, will the `X-Device-Fingerprint` header be clear when `appState.deviceFingerprint` has been clear.

src/v2/client/startLoginFlow.js
```diff
@@ -37,49 +38,64 @@ export async function startLoginFlow(settings) {
...
+      // 2. start the login again in order to introspect on settings.stateHandle
+      return startLoginFlow(settings);
+    }
   }
 
   // Use stateToken from options
   const stateHandle = settings.get('stateToken');
   if (stateHandle) {
-    return introspect(settings, stateHandle);
+    return authClient.idx.introspect({ stateHandle, version: idxApiVersion });
```
haishengwu-okta, line 98: is `version` mandatory to pass to `authClient.idx` ? otherwise we can omit it so that easy to deprecated `appState. apiVersion`? or even remove `appState.apiVersion` in this PR?

test/app/src/configArea.ts
```diff
@@ -95,7 +95,7 @@ export default class ConfigArea {
   }
 
   addEventListeners(): void {
-    this.configForm.addEventListener('change', (event) => {
```
haishengwu-okta, line 98: nit: enable ESLINT to error out for unused variable?

yarn.lock
```diff
@@ -2787,10 +2787,10 @@
     webcrypto-shim "^0.1.5"
     xhr2 "0.1.3"
 
-"@okta/okta-idx-js@0.22.0":
-  version "0.22.0"
-  resolved "https://registry.yarnpkg.com/@okta/okta-idx-js/-/okta-idx-js-0.22.0.tgz#c5dbebd155be2c8a8e2bbceb2e92701b6ba59349"
-  integrity sha512-SehJVwQI51xWrB1YaMn6UB54cOzhrM7z5k2xOI62I3B7qOQ3uFOZJvRCJnBcZ6TEcQuqJ3/6q93h6n8omw0tVQ==
+"@okta/okta-idx-js@0.24.0":
```
haishengwu-okta, line 2790: is auth-js depending on idx-js ?

src/v2/BaseLoginRouter.js
aarongranick-okta: The header will not be cleared. This is a change in behavior, and the test is updated. I discussed with @jmelberg-okta and seems there should not be any negative effect to keeping the header set.

src/v2/client/startLoginFlow.js
aarongranick-okta: Yeah, I think we can remove `apiVersion` in this PR. If we do need to change the version in the future this will happen internally in auth-js, I don't think we will be passing this on from a widget option.

test/app/src/configArea.ts
aarongranick-okta: sure, we can enable more lint on the test app. It's pretty new so can use some more love.

yarn.lock
aarongranick-okta: yes. Eventually auth-js will consume idx-js and then idx-js will be retired. This widget refactor is a prerequisite for that effort.

src/models/Settings.js
```diff
@@ -173,7 +173,7 @@ export default Model.extend({
     piv: ['object', false, {}],
 
     //Email verify callback
-    stateTokenExternalId: 'string'
+    otp: 'string'
```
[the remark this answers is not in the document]
aarongranick-okta: We handle the `otp` as a query parameter (it may or may not have been clicked on from an email, customers do have some power to deliver the OTP code in alternate ways).  The app needs to read the `otp` parameter from the URL and then pass it to the widget to continue the flow. Made sense to me to have the config option match the query parameter name so there is no confusion on how to wire it up.

jest.config.js
```diff
@@ -26,7 +27,7 @@ module.exports = {
     '^handlebars$': 'handlebars/runtime',
 
     // auth-js has a browser and server version. we want the browser version
-    '^@okta/okta-auth-js$': '<rootDir>/node_modules/@okta/okta-auth-js/dist/okta-auth-js.umd.js',
+    // '^@okta/okta-auth-js$': '<rootDir>/node_modules/@okta/okta-auth-js/esm',
 
     // idx-js uses cross-fetch. Force it to use the browser version so our spies work
     '^cross-fetch$': '<rootDir>/node_modules/cross-fetch/dist/browser-ponyfill.js',
```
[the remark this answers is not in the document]
aarongranick-okta: Yes this line can be removed, the UMD is the default/correct bundle but there was an error that was causing it to load the commonJS code which is why this line existed. I left in the commented line so we can test with ESM. I can adjust the comment to indicate this

test/app/src/configArea.ts
aarongranick-okta: It was coming as a warning in the test app, whereas this is an error with widget source. Seems like a quirk with Typescript: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unused-vars.md

I'll get it fixed so it is consistently returning an error

shuowu-okta (approved):

scripts/setup.sh
```diff
@@ -3,7 +3,7 @@
 # Install required node version
 export REGISTRY_REPO="npm-topic"
 export REGISTRY="${ARTIFACTORY_URL}/api/npm/${REGISTRY_REPO}"
-setup_service node v12.13.0
+setup_service node v12.20.0
```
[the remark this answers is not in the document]
aarongranick-okta: we don't actually need it. I'll remove it if I rebase again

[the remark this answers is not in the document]
aarongranick-okta: removed in last rebase

PR #2140, opened 2021-08-09
chore: Update Email Consent Text & Wording

worldleelavivat-okta:
## Description:

- Update Email Consent Text & Wording

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:
Before:
Screen Shot 2021-08-09 at 3 42 21 PM
Screen Shot 2021-08-09 at 3 42 27 PM

After:
Screen Shot 2021-08-10 at 10 53 40 AM
Screen Shot 2021-08-09 at 3 50 30 PM

### Issue:

- [OKTA-402915](https://oktainc.atlassian.net/browse/OKTA-402915)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -1048,9 +1048,9 @@ oie.consent.scopes.enduser.title = would like to:
 oie.consent.enduser.accept.label = Yes, it's me
 oie.consent.enduser.deny.label = No, it's not me
 oie.consent.enduser.title = Did you just try to sign in?
-oie.consent.enduser.email.allow.description = To continue, return to the device or window where you requested the email link. You may close this window at any time.
-oie.consent.enduser.deny.description = You may close this window at any time.
-oie.return.to.original.tab = Please return to the original tab.
+oie.consent.enduser.email.verified.title = Email verified!
+oie.return.to.original.tab = Please <span style="font-weight:bold">return</span> to your original screen to continue.
```
jmelberg-okta, line 1052: Did UX propose this? Wrapping `return` with text emphasis without a click handler seems like it'll confuse users.

I'd suggest making `return` a link or removing the `span` element.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -1048,9 +1048,9 @@ oie.consent.scopes.enduser.title = would like to:
...
+oie.close.this.window = Close this window anytime.
```
jmelberg-okta, line 1053: Please note that changing key names here will make existing translations for these keys **obsolete**. As such, all languages outside of `en` will fall-back to English instead of being translated.

Given this has already rolled out to new orgs, this is a **breaking change**.

packages/@okta/i18n/src/properties/login.properties
worldleelavivat-okta: UX Proposal: https://oktainc.atlassian.net/browse/OKTA-402915?focusedCommentId=2046134

`return` is just being bolded not a link since it is telling user to return to the original tab not navigating them from the current window.

packages/@okta/i18n/src/properties/login.properties
worldleelavivat-okta: I see. I will keep the key to be the same then and update the value instead. Would the translation get updated eventually? Or is it better to introduce new key while keep the current key as well?

jmelberg-okta: Yep - translations are usually delivered for our **monthly** releases.

If you reuse the existing key, worst case is customers will see **outdated** translations. In most cases, this is generally acceptable versus falling-back to English. For this reason, we usually advise on key reuse instead of renames.

jmelberg-okta: Looks good from the i18n resourcing side! 🚀

vincentvoong-okta (approved):
LGTM

sherizada-okta (approved):

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -1048,9 +1048,10 @@ oie.consent.scopes.enduser.title = would like to:
 oie.consent.enduser.accept.label = Yes, it's me
 oie.consent.enduser.deny.label = No, it's not me
 oie.consent.enduser.title = Did you just try to sign in?
-oie.consent.enduser.email.allow.description = To continue, return to the device or window where you requested the email link. You may close this window at any time.
-oie.consent.enduser.deny.description = You may close this window at any time.
-oie.return.to.original.tab = Please return to the original tab.
+oie.consent.enduser.email.allow.title = Success! Return to the original tab or window
```
jmelberg-okta, line 1051: Looks like we reverted this back to a new key.

jmelberg-okta (changes requested):

packages/@okta/i18n/src/properties/login.properties
worldleelavivat-okta: New key is required for `title` of the page.

worldleelavivat-okta: this change will go with monthly release PR so i think it should be okay to add new key.

jmelberg-okta (approved):
🚢

PR #2142, opened 2021-08-09
fix: error when enrolling OV in mobile on OIE

magizh-okta:
* in OIE during OV enrollment on mobile, we call switchForm to load the
'select-enrollment-channel' the issue was that switchForm was called even before the render of
initial form 'enroll-poll' thus both views were displayed.

* Moving to postRender also doesnt help since forms are added in postRender of FormController.
The add logic in baseView takes an extra eventloop to render the Form hence adding a timeout ot 0.

RESOLVES: OKTA-418833

**before**
image

**after**
Screen Shot 2021-08-09 at 4 30 04 PM

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-418833](https://oktainc.atlassian.net/browse/OKTA-418833)

### DIFF
--- a/src/v2/view-builder/views/ov/EnrollPollOktaVerifyView.js
+++ b/src/v2/view-builder/views/ov/EnrollPollOktaVerifyView.js
@@ -35,10 +35,6 @@ const Body = BaseForm.extend(Object.assign(
     noButtonBar: true,
     initialize() {
       BaseForm.prototype.initialize.apply(this, arguments);
-      if ((BrowserFeatures.isAndroid() || BrowserFeatures.isIOS()) &
-        this.options.appState.get('currentAuthenticator').contextualData.selectedChannel === 'qrcode') {
-        this.options.appState.trigger('switchForm', RemediationForms.SELECT_ENROLLMENT_CHANNEL);
-      }
       this.listenTo(this.model, 'error', this.stopPolling);
       this.startPolling();
     },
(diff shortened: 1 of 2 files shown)


src/v2/view-builder/views/ov/EnrollPollOktaVerifyView.js
```diff
@@ -71,6 +67,16 @@ const Body = BaseForm.extend(Object.assign(
 
       }
     },
+    postRender() {
```
mauriciocastillosilva-okta, line 70: there is an `afterRender` event that is triggered in FormController, would that help instead of adding a `setTimout` here?

src/v2/view-builder/views/ov/EnrollPollOktaVerifyView.js
```diff
@@ -71,6 +67,16 @@ const Body = BaseForm.extend(Object.assign(
...
+      // Using setTimeout of 0 because, the view is added in postRender of formController and not the initialize,
+      // and hence by the time swithForm is called, current view is not rendered and availabe on controller $el.
+      setTimeout(() => {
+        if ((BrowserFeatures.isAndroid() || BrowserFeatures.isIOS()) &
+          this.options.appState.get('currentAuthenticator').contextualData.selectedChannel === 'qrcode') {
+          this.options.appState.trigger('switchForm', RemediationForms.SELECT_ENROLLMENT_CHANNEL);
+        }
+      }, 0);
+    },
     getUISchema() {
```
mauriciocastillosilva-okta, line 80: should this return empty if we are in Android or iOs? Since we don't really need to render the QR code

src/v2/view-builder/views/ov/EnrollPollOktaVerifyView.js
magizh-okta: but for is not aware of existence of the controller and hence cannot listen to the event.

src/v2/view-builder/views/ov/EnrollPollOktaVerifyView.js
magizh-okta: @mauriciocastillosilva-okta  turns out the change that overrides the basecontroller render caused the issue. once that was reverted the issue got resolved. I created a ticket to just add a test to catch the regression in future. 
https://github.com/okta/okta-signin-widget/commit/018a85a76ce6c5ec270a333e6d6d028ec8183b6f#diff-563f8e94ab62f12655040f13d4aa05bf9111fecde04b01c9f39d29c2ff03467d

PR #2158, opened 2021-08-18
Display resend messages based on remediation in responses

kamaldeepcheema-okta:
## Description:
- For our existing SIW "Resend Views" for various factors, we previously hardcoded the message and display behavior.
- Now we totally drive based on the remediation in the responses. The backend now send the messages in the `messages` context, from which we extract the message and display right away. The backend will be the one driving when/what to display. 

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/77295104/129973252-6b511042-d637-4a43-94d6-c4223827fcae.mov

https://user-images.githubusercontent.com/77295104/129973275-6fcd1e0a-1d2e-43ae-962b-9c5536ec8a2d.mov

### Issue:

- [OKTA-412691](https://oktainc.atlassian.net/browse/OKTA-412691)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -312,6 +312,7 @@ email.button.send = Send me the code
 email.button.resend = Send again
 email.code.label = Verification code
 email.code.not.received = Haven\'t received an email?
+idx.email.code.not.received = Haven\'t received an email?
```
jmelberg-okta, line 315: Why do we need a second key? If possible, let's try to reuse `email.code.not.received` (and our other properties) and [map them accordingly](https://github.com/okta/okta-signin-widget/blob/5ae7f2bec639af5a7c58b1c0db8d82c8e358489d/src/v2/ion/i18nTransformer.js#L56).

pradeepdewda-okta: i think because backend is sending a new key

src/v2/view-builder/internals/BaseForm.js
```diff
@@ -128,9 +129,25 @@ export default Form.extend({
     // render messages as text
     const messagesObjs = this.options.appState.get('messages');
     if (messagesObjs?.value.length) {
-      const content = messagesObjs.value.map((messagesObj) => {
-        return messagesObj.message;
-      });
+
+      const content = messagesObjs.value
+        .filter(messagesObj => {
+          const isWarningMessage = messagesObj?.class === MESSAGE_CLASS.WARN;
```
pradeepdewda-okta, line 135: if there are multiple warnings is it going to show them all ? I think UX would like to just show the first warning widget receives.

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -1,42 +1,36 @@
...
+          type: 'warning',
+        }));
+      }
     },
 
-    showCalloutWithDelay() {
-      this.showMeTimeout = _.delay(() => {
-        this.$el.removeClass('hide');
-      }, SHOW_RESEND_TIMEOUT);
+    handelResendLink() {
```
pradeepdewda-okta, line 32: good that we are able to remove `setTImeout` perfect.. 👍

packages/@okta/i18n/src/properties/login.properties
jmelberg-okta: The backend can always send a different key, the Widget just needs to map this to a v1 key in our `i18nTransformer`. Best practice is to reuse existing translations whenever the flows are the same, especially when the values are identical.

pradeepdewda-okta: Agree. backend is not yet merged, we can ask them to use the existing key i.e `email.code.not.received`

src/v2/view-builder/internals/BaseForm.js
kamaldeepcheema-okta: This is a good point, I didn't account for this. So we just want to display the first warning we come across correct?

packages/@okta/i18n/src/properties/login.properties
kamaldeepcheema-okta: Looking at the backend PR here: https://github.com/okta/okta-core/pull/55383/files#diff-4ece32bc1f9d5ffcaf775429ce7813a08e7ec799bc1f74f2875ad43a0962ff2cR1037 - I think they use an existing pattern of starting their keys with the "idx" prefix so it might not make sense to ask them to change it just for the UI. 

In our case here, having the i18n mapping won't be enough either because in https://github.com/okta/okta-signin-widget/pull/2158/files#diff-ef082e5c8e18a4aa6fe3c57416549703855d9bbea222b747cb5e846b541185c6R17 we check the message key against the response we get from the server and the key in there won't use the mapping value.

jmelberg-okta: We have a convention for this already, using the [`getI18nKey` utility](https://github.com/okta/okta-signin-widget/blob/56930aa682bf8e0aee294b5a1c8588c5b0d4c6a8/src/v2/ion/i18nTransformer.js#L249-L269) to pick-up the mapped value.

Our goal is to reduce the number of breaking changes for customers migrating from classic to OIE. Specifying the map to a classic key allows them to continue using their customizations.

kamaldeepcheema-okta: Yeah that's a good point. I've updated the solution now so that it does re-use the v1 key. Please have a look again

pradeepdewda-okta:
@kamaldeepcheema-okta there is a new conversation brewing up for its implementation on backend. https://oktainc.atlassian.net/browse/OKTA-408316 So we need to hold off this PR.

kamaldeepcheema-okta:
Closing this PR as the plan has changed for this feature, so it will require re-implementation

PR #2159, opened 2021-08-19
Adding support for remember my username in SIW

kamaldeepcheema-okta:
## Description:
- Spec and details: https://docs.google.com/document/d/17ru7dSGKYgWO94czm3T75JNcxP8_AGfn-2AAbc5ADb8/edit
- This feature replaces the v1 "Remember me" feature and simplifies it to just remember the username/identifier in OIE and using that to pre-fill the identifier field.
- This functionality is gated behind this Admin config: https://github.com/okta/okta-core/pull/55782
- When this is on, upon successful authentication, we save the identifier in the browser Cookie, and the next time the identifier form load it uses this value to pre-fill in the input.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/77295104/130121616-f05d2245-57c9-48da-afa8-3e5222025fdb.mov

### Issue:

- [OKTA-414139](https://oktainc.atlassian.net/browse/OKTA-414139)


src/v2/controllers/FormController.js
```diff
@@ -283,4 +287,30 @@ export default Controller.extend({
...
+  /**
+    * When "Remember My Username" is enabled, we save the identifier in a cookie
+    * so that the next time the user visits the SIW, the identifier field can be 
+    * pre-filled with this value.
+   */
+  updateIdentifierCookie: function(model) {
+    const formName = model.get('formName');
+
+    // Only update the cookie when the user has successfully identified themselves (i.e. if we're on the
+    // challenge-authenticator form) to avoid incorrect/uneccessary updates.
```
pradeepdewda-okta, line 299: can you confirm with backend that this is always the case? that.. we present `challenge` view always only after authentication. Is there any corner case ?

src/v2/controllers/FormController.js
```diff
@@ -283,4 +287,30 @@ export default Controller.extend({
...
+    if (formName !== FORMS.CHALLENGE_AUTHENTICATOR) {
```
sherizada-okta, line 300: Is this always the next step after a user identifies? A user could have disabled mfa?

sherizada-okta: Would be easier to depend on the identifier form, in case other flows get added too.

src/v2/controllers/FormController.js
kamaldeepcheema-okta: Yeah from my understanding this is the case, but I would like to confirm with @iancorcoran-okta as well. Ian the intention here is to update the cookie only after successful authentication and here I was relying on the successful response when showing the challenge-authenticator form. Is this approach correct?

src/v2/controllers/FormController.js
kamaldeepcheema-okta: The intention here is to only update the cookie upon successful authentication, relying on the identifier doesn't work when we're using an identifier first flow. I'm confirming this approach with Ian on this thread: https://github.com/okta/okta-signin-widget/pull/2159#discussion_r693078144

src/v2/controllers/FormController.js
iancorcoran-okta: To make sure I'm understanding, you're asserting that the identifier is valid iff you get a success response from `challenge-authenticator`? I don't think that works... e.g. if an app only requires 1FA and user authenticates via IdP,  or may be some cases around user enumeration

kamaldeepcheema-okta: @iancorcoran-okta Yes essentially on the client side I need to know that successful authentication has happened, and I was relying on the above which seems incorrect now. The question is I guess how can I determine that on the client?

test/unit/spec/v2/view-builder/views/IdentifierView_spec.js
```diff
@@ -146,4 +150,62 @@ describe('v2/view-builder/views/IdentifierView', function() {
     expect(testContext.view.model.get('identifier')).toEqual('testUsername');
     expect(testContext.view.$el.find('.o-form-input-name-identifier input').val()).toEqual('testUsername');
   });  
+
+  it('view updates model and view correctly with rememberMe feature ON', function() {
```
sherizada-okta, line 154: "fills in identifier form with username from cookie when rememberMe feature is enabled" ?

src/v2/controllers/FormController.js
sherizada-okta: This is reminding me of the `processCreds` problem, where you need the API to return the identifier on identify with password remediation success?

kamaldeepcheema-okta: Note: moved this conversation to this slack thread: https://okta.slack.com/archives/CCA77QVL2/p1629477508430300

kamaldeepcheema-okta: UPDATE: as per the slack thread above, the most reliable way is to check for the `success` or `successWithInteractionCode` objects in the response. The implementation has been updated now

test/testcafe/spec/IdentifyWithRememberUsername_spec.js
```diff
@@ -0,0 +1,95 @@
...
+  .onRequestTo('http://localhost:3000/idp/idx/challenge/answer')
+  .respond(xhrErrorIdentify, 403);  
+
+const identifyMock = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(xhrIdentify)
+  .onRequestTo('http://localhost:3000/idp/idx/identify')
+  .respond(xhrPassword)
+  .onRequestTo('http://localhost:3000/idp/idx/challenge/answer')
+  .respond(xhrSuccess);
```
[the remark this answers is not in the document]
kamaldeepcheema-okta: updated the test now

test/unit/spec/v2/view-builder/views/IdentifierView_spec.js
```diff
@@ -151,7 +151,7 @@ describe('v2/view-builder/views/IdentifierView', function() {
     expect(testContext.view.$el.find('.o-form-input-name-identifier input').val()).toEqual('testUsername');
   });  
 
-  it('view updates model and view correctly with rememberMe feature ON', function() {
+  it('fills in identifier form with username from cookie when rememberMe feature is enabled', function() {
```
pradeepdewda-okta, line 150: `pre-fill identifier form ....`

test/unit/spec/v2/view-builder/views/IdentifierView_spec.js
```diff
@@ -179,7 +179,7 @@ describe('v2/view-builder/views/IdentifierView', function() {
     expect(testContext.view.$el.find('.o-form-input-name-identifier input').attr('autocomplete')).toEqual('identifier');
   });
 
-  it('view updates model and view correctly with rememberMe feature OFF', function() {
+  it('does not fill in identifier form with username from cookie when rememberMe feature is disabled', function() {
```
pradeepdewda-okta, line 177: `does not pre-fill identifier form ... `

test/testcafe/spec/IdentifyWithRememberUsername_spec.js
```diff
@@ -93,3 +143,97 @@ test.requestHooks(identifyRequestLogger, identifyMock)('should remember username
...
...zMiLCJlbmMiOiJBMjU2R0NNIiwiYWxnIjoiZGlyIn0..4z7WWUdd0LzIJLxz.GmOyeMJ5XtS7kZipFanQxaMd2rblpEeaWv8U5IfaJv5F2V1otwft4q1tVo3yqbedhBjN-nO5qk6qR-0Op34lmecwwmHeRyzYbhrFiLZaR88nhCFMblP8Bo5d3Opl5gkX02e0FQL-Osorxvml0XYbuO7GdTH5EkIv0Q7h0Dq7L__h_uFLies8AkJZWIDQh25RlqcEjToHCvVW31A_NJJ1Vf5c8GFuyb9LsJ9kUpZikpK6C72GPrU_LGrIW8VBT4l24dDEre4J8XvTNO7fdVDiq-H7BEeIjaY06q1zqlVLWwqOHoKpGNQ0NMhBIXB0ZZC57Me9pFI5GLIUwRDUIm1vw3t_mHJDVIcCJe9kmt29tTccZ8Zo0N3q5bSiwMoNHPLxZSOrbx-bf4fniorNH5ypJnke7pc2Q3DFmqfPrB7CE1REjAyCKBHYDAfVexYCkMfCl0E8oMFJinnLbynb7Bqvbp_DqL8h0pNIoXUF4KTTsuKQg8yCCqBhkajxlvh9G7L3Sf76o4B2itB7ldeqXzAE9H60yqhIKEZPNOHUgRC2SkWkWlH6NIaNWQ2Bi2CjnL9YvUuQmO-dpf08KeCgwfVmT4GBTGfTkXwy3pBitacCqEREen2j2iUH9mhi2LOOFaGLh0TXslcBgkGuht6P7gyH2JN6yFInQyIp33xQsqYg7nqOZG1LCrQSqoviTfI72-AC6b7tju8YEn1P0nXGbSzlCztSXl2pa95tr4L5pyX8fNydKYMTLeHEnmNtXlRB6wQYP1ljf4Tzgus7O0etyJs75znsXHZ42znxlEKGhTo3ucFe3CI-vsHF1FDDj2DVeWl21zVOTehTbBaemoD1ekD5F8OHS7SrK9Bw7PTa-lpyls1OxvE_Wsco4_eGbax_DoPm6DbCwj8hWzb5wLEs6TClZKoUJeV1MSVB3OgGBZ3AGzhwfeG0sGi5DnUpAeKqgP6IN8kziNRDmW3YE0qIY2mLs7nI438RTu__6bg1E6SH1QHMNucNbmoDR6VDIUmlYc0xEpygH6PBVqiPD64MnD73_D9IinVNzqW7KQzAvuFFQW_LGDfjuh1D-oTs1gi1wWDylibjxdJabveoJ10NHgeb6SaYHg.kf5iTnjNKsKqhz0iE5K_Yw',
```
pradeepdewda-okta, line 232: you can change this to use a smaller string.. `eyJ6aXAiOiJERUYiLCJhbGlhcyI6ImVuY3J5cHRpb25rZXkiLCJ2ZXIiOiIxIiwib2lkIjoiMD`

src/v2/BaseLoginRouter.js
```diff
@@ -245,6 +249,32 @@ export default Router.extend({
...
+  /**
+    * When "Remember My Username" is enabled, we save the identifier in a cookie
+    * so that the next time the user visits the SIW, the identifier field can be 
+    * pre-filled with this value.
+   */
+  updateIdentifierCookie: function(idxResponse) {
+    // Only update the cookie when the user has successfully authenticated themselves 
+    // to avoid incorrect/uneccessary updates. This is done by checking the server response
+    // and seeing if either the 'success' or 'successWithInteractionCode' objects are present.
+    if (!idxResponse?.rawIdxState?.success && !idxResponse?.rawIdxState?.successWithInteractionCode) {
```
pradeepdewda-okta, line 261: just for the sake of readability create another function

`isAuthenticationSucceeded()`

src/v2/BaseLoginRouter.js
```diff
@@ -84,6 +85,9 @@ export default Router.extend({
   },
 
   handleIdxResponseSuccess(idxResponse) {
+    // Check if cookie needs to be updated
+    this.updateIdentifierCookie(idxResponse);
```
pradeepdewda-okta, line 89: did you check further remediations (next screens) after authentication.. do they contains `dxResponse?.rawIdxState?.success`  ??

src/v2/BaseLoginRouter.js
```diff
@@ -245,6 +249,32 @@ export default Router.extend({
     this.controller.render();
   },
 
+  /**
+    * When "Remember My Username" is enabled, we save the identifier in a cookie
+    * so that the next time the user visits the SIW, the identifier field can be 
+    * pre-filled with this value.
+   */
+  updateIdentifierCookie: function(idxResponse) {
```
pradeepdewda-okta, line 257: every idx response will call `updateIdentifierCookie` even when it is not required.

test/unit/spec/v2/view-builder/views/IdentifierView_spec.js
kamaldeepcheema-okta: I'll update this

test/testcafe/spec/IdentifyWithRememberUsername_spec.js
kamaldeepcheema-okta: I just tried locally and the smaller string does not work. Are you sure about this?

src/v2/BaseLoginRouter.js
kamaldeepcheema-okta: I'll add this in.

src/v2/BaseLoginRouter.js
kamaldeepcheema-okta: It shouldn't matter because we're only interested in the time when the user successfully completes authentication, hence only updating the cookie at that time only. From my local tests in my OIE org, I can see it only returns those objects at the time of authentication only.

src/v2/BaseLoginRouter.js
kamaldeepcheema-okta: The very first step in that method checks for successful authentication so it will return early, but I can move that check outside the method instead for clarity.

test/testcafe/spec/IdentifyWithRememberUsername_spec.js
pradeepdewda-okta: you might need to change in the mock as well.. is it coming from an existing mock ?

pradeepdewda-okta: I would say do not check for `stateHandle`.. after JWE stateHandle can change..

kamaldeepcheema-okta: Yeah I was thinking about the same thing, not to check for the stateHandle, which is coming from the mock (and used in various other tests). I'll update this.

test/testcafe/spec/Identify_spec.js
```diff
@@ -310,7 +310,10 @@ test.requestHooks(identifyRequestLogger, identifyMockWithFingerprintError)('shou
 test.requestHooks(identifyRequestLogger, baseIdentifyMock)('should pre-populate identifier field with username config', async t => {
   const identityPage = await setup(t);
   await rerenderWidget({
-    username: 'myTestUsername@okta.com'
+    username: 'myTestUsername@okta.com',
+    features: { // Ensure remember my username feature is off so it doesn't override config
+      rememberMe: false,
```
sherizada-okta, line 315: You don't need this because it's undefined by default which is falsey, and updating this test means the existing functionality has changed, which it hasn't?

test/testcafe/spec/IdentifyWithRememberUsername_spec.js
```diff
@@ -0,0 +1,239 @@
...
+  });
+
+  // Ensure identifier field is pre-filled with saved username cookie
+  await identityPage.navigateToPage();
+  const identifier = identityPage.getIdentifierValue();
+  await t.expect(identifier).eql('testUser@okta.com');
+});
```
sherizada-okta, line 237: A test for `features.rememberMe: true` and `username: 'this_should_not_prefill_identifier'`

test/testcafe/spec/IdentifyWithRememberUsername_spec.js
```diff
@@ -0,0 +1,239 @@
...
+  await identityPage.navigateToPage(options || {});
+  return identityPage;
+}
+
+test.requestHooks(identifyRequestLogger, identifyWithError)('identifer first flow - should NOT remember username after failed authentication', async t => {
+  const identityPage = await setup(t);
+  await rerenderWidget({
+    features: {
+      rememberMe: true,
+    }
```
sherizada-okta, line 82: a test for `'features.rememberMe': true`, if this is supposed to support this configuration key

sherizada-okta (commented):
Can you update the docs with this new flag [here](https://github.com/okta/okta-signin-widget/blob/master/docs/interaction_code_flow.md#feature-flags)

test/testcafe/spec/IdentifyWithRememberUsername_spec.js
sherizada-okta: ^ essentially testing [this block](https://github.com/okta/okta-signin-widget/pull/2159/files#diff-dcaedbe027b0f2da495c93e3f2f0a1c1d6895a7863e99f729b704b014d731f1eR32-R36)

test/testcafe/spec/Identify_spec.js
kamaldeepcheema-okta: https://github.com/okta/okta-signin-widget#feature-flags - we're leveraging the same flag that was there before which by default is true, it's also set to true as part of our default .widgetrc file.

kamaldeepcheema-okta:
> Can you update the docs with this new flag [here](https://github.com/okta/okta-signin-widget/blob/master/docs/interaction_code_flow.md#feature-flags)

This is not a new flag, and is already doc'ed in https://github.com/okta/okta-signin-widget#feature-flags. The admin UI config option that drives this flag is new, but not this flag.

test/testcafe/spec/IdentifyWithRememberUsername_spec.js
kamaldeepcheema-okta: I don't follow here, we're essentially turning the feature on here already. The SIW has code to flatten the object settings structure so that will ultimately become "features.rememberMe", but it's not the intention of this spec to test that (not sure if that's what you're asking)

sherizada-okta:
> > Can you update the docs with this new flag [here](https://github.com/okta/okta-signin-widget/blob/master/docs/interaction_code_flow.md#feature-flags)
> 
> This is not a new flag, and is already doc'ed in https://github.com/okta/okta-signin-widget#feature-flags. The admin UI config option that drives this flag is new, but not this flag.

Right, the main SIW readme documents classic/v1 configuration options, but this ff is now supported in OIE which is documented at the link I posted. It's buried, but the OIE readme is also referenced on the main readme [here](https://github.com/okta/okta-signin-widget#interaction-code-flow).

kamaldeepcheema-okta:
> > > Can you update the docs with this new flag [here](https://github.com/okta/okta-signin-widget/blob/master/docs/interaction_code_flow.md#feature-flags)
> > 
> > 
> > This is not a new flag, and is already doc'ed in https://github.com/okta/okta-signin-widget#feature-flags. The admin UI config option that drives this flag is new, but not this flag.
> 
> Right, the main SIW readme documents classic/v1 configuration options, but this ff is now supported in OIE which is documented at the link I posted. It's buried, but the OIE readme is also referenced on the main readme [here](https://github.com/okta/okta-signin-widget#interaction-code-flow).

I see, that makes more sense. I've updated that file with this flag now.

test/testcafe/spec/Identify_spec.js
sherizada-okta: ah didn't know it defaulted to true

test/testcafe/spec/IdentifyWithRememberUsername_spec.js
kamaldeepcheema-okta: I've added a test for this block now

sherizada-okta (approved):

pradeepdewda-okta (approved):
This is lit now 🔥

PR #2170, opened 2021-08-24
fix: webauthn verification on OIE

magizh-okta:
* Part of challenge data, specifically the extension was not passed to
credentials.get.  This caused issue with verifying u2f keys that were originally enrolled
when u2f factor was enabled in v1 and later on migrated to webauthn and finallly webauthn on OIE.

RESOLVES: OKTA-422158

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

### DIFF
--- a/src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
+++ b/src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
@@ -1,4 +1,4 @@
-import { loc, createButton, createCallout } from 'okta';
+import { _, loc, createButton, createCallout } from 'okta';
 import { BaseForm } from '../../internals';
 import BaseAuthenticatorView from '../../components/BaseAuthenticatorView';
 import CryptoUtil from '../../../../util/CryptoUtil';
@@ -75,11 +75,11 @@ const Body = BaseForm.extend({
         });
       }
     });
-    const options = {
+    const challengeData = authenticatorData.contextualData.challengeData;
+    const options = _.extend({}, challengeData, {
       allowCredentials,
-      userVerification: authenticatorData.contextualData.challengeData.userVerification,
-      challenge: CryptoUtil.strToBin(authenticatorData.contextualData.challengeData.challenge),
-    };
+      challenge: CryptoUtil.strToBin(challengeData.challenge),
+    });
     navigator.credentials.get({
       publicKey: options,
       signal: this.webauthnAbortController.signal
--- a/test/unit/spec/v2/view-builder/views/ChallengeWebauthnView_spec.js
+++ b/test/unit/spec/v2/view-builder/views/ChallengeWebauthnView_spec.js
@@ -174,6 +174,9 @@ describe('v2/view-builder/views/webauthn/ChallengeWebauthnView', function() {
               ),
             },
           ],
+          extensions: {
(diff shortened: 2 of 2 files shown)


mpeng-okta (approved):
🌟 
Will test tomorrow!

magizh-okta:
@mauriciocastillosilva-okta  @mpeng-okta  tested with this SIW changes and this fixes the issue.

mauriciocastillosilva-okta (approved):
looks good!

src/v2/view-builder/views/webauthn/ChallengeWebauthnView.js
```diff
@@ -75,11 +75,11 @@ const Body = BaseForm.extend({
         });
       }
     });
-    const options = {
+    const challengeData = authenticatorData.contextualData.challengeData;
+    const options = _.extend({}, challengeData, {
       allowCredentials,
-      userVerification: authenticatorData.contextualData.challengeData.userVerification,
```
srinivasanagandla-okta, line 80: Why are we removing `userVerification` attribute? Did not quite understand this.

magizh-okta: @srinivasanagandla-okta  the userverificaiton is already in challengeData which we are extending from and we are not overriding it unlike the challenge.

srinivasanagandla-okta: So what is the net new change then since challenge attribute seems to be the same? Sorry for my JS ignorance.

mpeng-okta: NP! Basically, before we had an object `options` which we passed into the WebAuthn API. It looked like this:
```
{
  allowCredentials: [...],
  userVerification: "someString", // equal to authenticatorData.contextualData.challengeData.userVerification
  challenge: "challengeNonce" // equal to CryptoUtil.strToBin(authenticatorData.contextualData.challengeData.challenge)
}
```

**The issue**: we needed to pass the WebAuthn-relevant `extensions` object over into the final `options` object. To achieve this, we're using `challengeData`, which already contains that `extensions` object (but was going unused). `challengeData` looks like this:
```
{
  userVerification: "someString",
  challenge: "someNonce",
  extensions: { "appid": "https://org.okta1.com }
}
```

So, we're using the [underscore.js library's extend method](https://underscorejs.org/#extend): `_.extend({}, challengeData, {...});`; it takes an empty object `{}` and extends it by using the entire above `challengeData` object, then the `allowCredentials` + `challenge` values in the third object argument, and sets them into the `{}` object.[1]

So, with this new change, we:
1. Still add `allowCredentials` into the final `options` object
2. There's no need to specify `userVerification` because it's part of the `challengeData`, which is going into `options`
3. We still do still post-process the `challenge` nonce (put it into a binary data encoring format)
4. Similar to `userVerification`, there's no need to specify `extensions` now because it's part of `challengeData`

Let me know if that makes sense!

[1]Since it's copied by source, technically we could omit the `{}` object, but this way we just have consistent code w/ v1: https://github.com/okta/okta-signin-widget/blob/587acb9220137d324ab4d556b9dbf32c9e5cc8cc/src/VerifyWebauthnController.js#L97-L100

PR #2172, opened 2021-08-24
chore: make Okta Verify MFA challenge UI extends from BaseFormWithPolling

jhe-okta:
## Description:
Make the Okta Verify MFA challenge view extends from the BaseFormWithPolling so that it gets access to the polling functions. 

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Reviewers:
@magizh-okta

### Issue:

- [OKTA-422561](https://oktainc.atlassian.net/browse/OKTA-422561)

### DIFF
--- a/src/v2/view-builder/views/ov/ChallengeOktaVerifyFastPassView.js
+++ b/src/v2/view-builder/views/ov/ChallengeOktaVerifyFastPassView.js
@@ -1,9 +1,8 @@
 import { $ } from 'okta';
-import { BaseForm } from '../../internals';
+import { BaseFormWithPolling } from '../../internals';
 import Logger from '../../../../util/Logger';
 import { CHALLENGE_TIMEOUT } from '../../utils/Constants';
 import BrowserFeatures from '../../../../util/BrowserFeatures';
-import polling from '../shared/polling';
 import {doChallenge} from '../../utils/ChallengeViewUtil';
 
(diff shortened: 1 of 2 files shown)


src/v2/view-builder/views/ov/ChallengeOktaVerifyFastPassView.js
```diff
@@ -40,7 +39,7 @@ const Body = BaseForm.extend(Object.assign(
     },
 
     remove() {
-      BaseForm.prototype.remove.apply(this, arguments);
+      BaseFormWithPolling.prototype.remove.apply(this, arguments);
```
magizh-okta, line 42: nit : Wondering if we can move the remove to BaseFormWithPolling as well ?

magizh-okta (approved):
small nit after that 🚢

src/v2/view-builder/views/ov/ChallengeOktaVerifyFastPassView.js
jhe-okta: That's the ultimate goal. Currently the `DeviceChallengePollView` and `ChallengeOktaVerifyFastPassView` are mostly the same with small diff such as the `remove` function. I am going to make incremental changes and eventually have them share the common part.

PR #2178, opened 2021-08-27
Updating Resend Views to use a global timer

kamaldeepcheema-okta:
## Description:
- Updating Resend Views to use a global timer so that even if a SIW re-render occurs, at most the user will wait for 30s until they see the resend link.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/77295104/131181926-a509415a-6809-4559-af7c-e71495c9e8d5.mov

### Issue:

- [OKTA-412691](https://oktainc.atlassian.net/browse/OKTA-412691)

### DIFF
--- a/src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
+++ b/src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
@@ -1,11 +1,11 @@
-import { loc, View, createCallout, _ } from 'okta';
+import { loc, createCallout } from 'okta';
 import { BaseForm } from '../../internals';
 import email from '../shared/email';
 import BaseAuthenticatorView from '../../components/BaseAuthenticatorView';
-import { SHOW_RESEND_TIMEOUT } from '../../utils/Constants';
+import BaseResendView from '../shared/BaseResendView';
 import BaseFormWithPolling from '../../internals/BaseFormWithPolling';
 
-const ResendView = View.extend(
+const ResendView = BaseResendView.extend(
   {
     className: 'hide resend-email-view',
     events: {
@@ -26,23 +26,8 @@ const ResendView = View.extend(
       if (!this.$el.hasClass('hide')) {
         this.$el.addClass('hide');
       }
-      this.showCalloutWithDelay();
+      this.showCalloutAfterTimeout();
     },
-
-    postRender() {
-      this.showCalloutWithDelay();
-    },
-
-    showCalloutWithDelay() {
-      this.showMeTimeout = _.delay(() => {
-        this.$el.removeClass('hide');
-      }, SHOW_RESEND_TIMEOUT);
-    },
-
-    remove() {
-      View.prototype.remove.apply(this, arguments);
-      clearTimeout(this.showMeTimeout);
-    }
   },
 );
 
--- a/src/v2/client/sessionStorageHelper.js
+++ b/src/v2/client/sessionStorageHelper.js
@@ -12,6 +12,7 @@
 
 const STATE_HANDLE_SESSION_STORAGE_KEY = 'osw-oie-state-handle';
 const LAST_INITIATED_LOGIN_URL_SESSION_STORAGE_KEY = 'osw-oie-last-initiated-login-url';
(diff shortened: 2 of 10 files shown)


src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
```diff
@@ -34,9 +35,23 @@ const ResendView = View.extend(
...
+      const timeStamp = sessionStorageHelper.getResendTimestamp();
+      if (!timeStamp) {
+        sessionStorageHelper.setResendTimestamp(Date.now());
+      }
+
+      // We keep track of a 'global' timestamp in sessionStorage because if the SIW does a re-render,
+      // we don't want to force the user to wait another 30s again to see the resend link. With this
+      // the user will wait AT MOST 30s until they see the resend link.  
+      const start = sessionStorageHelper.getResendTimestamp();
+      this.showMeTimeout = setInterval(() => {
```
sherizada-okta, line 47: You probably don't want to create more than one setInterval timer?

src/v2/view-builder/views/ov/ResendView.js
```diff
@@ -31,9 +32,23 @@ export default View.extend({
...
+    // the user will wait AT MOST 30s until they see the resend link.
+    const start = sessionStorageHelper.getResendTimestamp();
+    this.showMeTimeout = setInterval(() => {
+      const now = Date.now();
+      if (now - start >= SHOW_RESEND_TIMEOUT) {
+        this.$el.removeClass('hide');
+        clearInterval(this.showMeTimeout);
+        sessionStorageHelper.removeResendTimestamp();
+      }      
+    }, 250);
```
sherizada-okta, line 51: `showCalloutWithDelay` could go into a base view that these other views extend from

src/v2/view-builder/views/email/BaseAuthenticatorEmailView.js
kamaldeepcheema-okta: @sherizada-okta  can you clarify what the issue here is? I'm not sure what you're referring to

pradeepdewda-okta:
@kamaldeepcheema-okta if the warning is already appeared and user refreshes.. then would we show the warning after refresh right away once very first timer has elapsed ?

src/v2/view-builder/views/ov/ResendView.js
kamaldeepcheema-okta: This is done now.

kamaldeepcheema-okta:
> @kamaldeepcheema-okta if the warning is already appeared and user refreshes.. then would we show the warning after refresh right away once very first timer has elapsed ?

@pradeepdewda-okta Yes that's right. I've signed off this behavior with UX. When I was thinking about this, adding that capability would have introduced more complexity around ensure the state is cleared correctly (i.e. there's some edge cases in making sure we don't just show warning right away when we don't mean to). The gain here was small so hence why the UX contact and I opted for this behavior.

src/v2/view-builder/views/shared/ResendViewBase.js
```diff
@@ -0,0 +1,35 @@
+import { View } from 'okta';
```
pradeepdewda-okta, line 1: Name it `BaseResendView`

src/v2/view-builder/views/ov/ResendView.js
```diff
@@ -1,8 +1,8 @@
-import { View, createCallout, _ } from 'okta';
+import { createCallout } from 'okta';
```
pradeepdewda-okta, line 1: We should change the name of this file now.. OVResendView

kamaldeepcheema-okta: done

src/v2/view-builder/views/shared/ResendViewBase.js
kamaldeepcheema-okta: done

pradeepdewda-okta (approved):

PR #2182, opened 2021-09-01
feat: hooks

aarongranick-okta:
## Description:

async functions can block processing and execute custom logic before or after form render (in OIE).

Hooks functions can be defined and passed via widget config, or at runtime using new methods `before()` and `after()`

New method `getUser` is added to facilitate needed logic for OKTA-422364

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-425549](https://oktainc.atlassian.net/browse/OKTA-425549)

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ See the [Usage Guide](#usage-guide) for more information on how to get started u
   - [on](#on)
   - [off](#off)
   - [authClient](#authclient)
+  - [before](#before)
+  - [after](#after)
 - [Configuration](#configuration)
   - [OIDC Applications](#oidc-applications)
   - [Basic config options](#basic-config-options)
@@ -87,6 +89,7 @@ See the [Usage Guide](#usage-guide) for more information on how to get started u
   - [Smart Card IdP](#smart-card-idp)
   - [Bootstrapping from a recovery token](#bootstrapping-from-a-recovery-token)
   - [Feature flags](#feature-flags)
+  - [Hooks](#hooks)
 - [Events](#events)
   - [ready](#ready)
   - [afterError](#aftererror)
@@ -627,6 +630,50 @@ var config = {
 
 var signIn = new OktaSignIn(config);
 // signIn.authClient.options.clientId === '{yourClientId}'
+```
+
+### before
+
+> **Note**: This function is only supported when using the Okta Identity Engine
+
+Adds an asynchronous hook function which will execute before a view is rendered.
+
+```javascript
+var config = {
+  baseUrl: 'https://{yourOktaDomain}',
+  authParams: {
+    issuer: 'https://{yourOktaDomain}/oauth2/default',
+    clientId: '{yourClientId}'  
+  },
+  useInteractionCodeFlow: true
+};
+var signIn = new OktaSignIn(config);
+signIn.before('success-redirect', async () => {
+  // custom logic can go here. when the function resolves, execution will continue.
+});
+
+```
+
+### after
+
+> **Note**: This function is only supported when using the Okta Identity Engine
+
+Adds an asynchronous hook function which will execute after a view is rendered.
+
+```javascript
+var config = {
+  baseUrl: 'https://{yourOktaDomain}',
+  authParams: {
+    issuer: 'https://{yourOktaDomain}/oauth2/default',
+    clientId: '{yourClientId}'  
+  },
+  useInteractionCodeFlow: true
+};
+var signIn = new OktaSignIn(config);
+signIn.after('identify', async () => {
+  // custom logic can go here. when the function resolves, execution will continue.
+});
+
 ```
 
 ## Configuration
@@ -1464,6 +1511,44 @@ features: {
 - **features.showPasswordToggleOnSignInPage** - End users can now toggle visibility of their password on the Okta Sign-In page, allowing end users to check their password before they click Sign In. This helps prevent account lock outs caused by end users exceeding your org's permitted number of failed sign-in attempts. Note that passwords are visible for 30 seconds and then hidden automatically. Defaults to `false`.
 
 - **features.scrollOnError** - By default, errors will be scrolled into view. Set to `false` to disable this behavior.
+
+### Hooks
+
+> **Note**: Hooks are only supported when using the Okta Identity Engine
+
+Asynchronous callbacks can be invoked before or after a specific view is rendered. Hook callbacks block processing to run custom logic. Nomal execution will resume after the Promise returned from the callback function resolves.
+
+```javascript
+// Hooks can be added via config
+const config = {
+  hooks: {
+    'identify': {
+      after: [
+        async function afterIdentify() {
+          // custom logic goes here
+        }
+      ]
+    },
+    'success-redirect': {
+      before: [
+        async function afterIdentify() {
+          // custom logic goes here
+        }
+      ]
+    }
+  }
+};
+
+// Hooks can also be added at runtime
+signIn.before('success-redirect', async () => {
+  // custom logic goes here
+});
+
+signIn.after('identify', async () => {
+  // custom logic goes here
+});
+
+```
 
 ## Events
 
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@
 [Custom Authorization Server]: https://developer.okta.com/docs/guides/customize-authz-server/overview/
 [Authorization Server]: https://developer.okta.com/docs/concepts/auth-servers/#which-authorization-server-should-you-use
 [Social Login]: https://developer.okta.com/docs/concepts/social-login/
+[Identity Engine]: https://developer.okta.com/docs/guides/oie-intro/
 <!-- end links -->
 
 <!-- omit in toc -->
@@ -40,6 +41,7 @@ The widget is used on Okta's default signin page to start an Okta SSO session an
 See the [Usage Guide](#usage-guide) for more information on how to get started using the Sign-in Widget.
 
 <!-- TOC is generated using Markdown All in One -->
+- [Okta Identity Engine](#okta-identity-engine)
 - [Related SDKs](#related-sdks)
   - [Javascript](#javascript)
   - [Java](#java)
@@ -101,6 +103,10 @@ See the [Usage Guide](#usage-guide) for more information on how to get started u
   - [Utilizing Pseudo-loc](#utilizing-pseudo-loc)
 - [Browser support](#browser-support)
 - [Contributing](#contributing)
+
+## Okta Identity Engine
+
+The Okta [Identity Engine][] (OIE) is a platform service that allows enterprises to build more flexible access experiences that are tailored to their organizational needs. The Okta Sign-in Widget supports OIE in all [usage](#usage-guide) scenarios.
 
 ## Related SDKs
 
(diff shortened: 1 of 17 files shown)


src/widget/OktaSignIn.js
```diff
@@ -94,6 +111,8 @@ var OktaSignIn = (function() {
       hide,
       show,
       remove,
+      before,
```
haishengwu-okta, line 114: Why would you think of 2 ways to add hooks rather only via config?
Are you anticipating hooks may be added on the fly?

haishengwu-okta (commented):
👍

src/widget/OktaSignIn.js
aarongranick-okta: I was thinking more of the case where config may be reconstituted from a string like an environment variable or yaml file, and the developer does not want to include any functions. Current logic (success function) is added at runtime, not config.

haishengwu-okta: I'd leaning toward single way to do the job rather being flexible until developer explicitly asks. just by `0.02`.

src/v2/BaseLoginRouter.js
```diff
@@ -35,6 +35,7 @@ import {
 import transformIdxResponse from './ion/transformIdxResponse';
 import { FORMS } from './ion/RemediationConstants';
 import CookieUtil from 'util/CookieUtil';
+import { doHooksAfter, doHooksBefore } from '../util/Hooks';
```
pradeepdewda-okta, line 38: does that mean execute ? `executeHooksAfter`

src/widget/OktaSignIn.js
pradeepdewda-okta: I would prefer to have the flexibility inbuilt rather than later when someone ask about it.. also it doesn't seem much and nicely meld with existing `hide` `show` functions

src/v2/BaseLoginRouter.js
aarongranick-okta: Yes, I like "execute" better.

src/v2/models/AppState.js
```diff
@@ -241,33 +243,47 @@ export default Model.extend({
...
-    }
+    const currentFormName = this.getFormName(transformedResponse);
 
     this.clearAppStateCache();
 
     // set new app state properties
     this.set(transformedResponse);
 
+    const hook = hooks?.getHook(currentFormName); // may be undefined
+    await executeHooksBefore(hook);
```
pradeepdewda-okta, line 259: Instead of widget always execute the hooks.. should we relay this control of execution to the individual flows which expect hooks to be executed.. So when these flows renders screens, they can instruct widget to execute the flow `preRender` or `postRender` ?

src/v2/models/AppState.js
```diff
@@ -241,33 +243,47 @@ export default Model.extend({
...
+
     // make sure change `currentFormName` is last step.
     // change `currentFormName` will re-render FormController,
     // which may depend on other derived properties hence
     // those derived properties must be re-computed before
     // re-rendering controller.
     this.set({ currentFormName });
-  }
+
+    await executeHooksAfter(hook);
```
pradeepdewda-okta, line 268: should we add a check that execute only if there is any hook exists..

src/v2/BaseLoginRouter.js
```diff
@@ -66,6 +66,7 @@ export default Router.extend({
       // and then the open tooltip will lose focus and close.
     });
 
+    this.hooks = options.hooks;
```
pradeepdewda-okta, line 69: since hooks are coming from `config` we should put it on `this.settings.hooks` instead..

src/widget/OktaSignIn.js
```diff
@@ -157,7 +187,7 @@ var OktaSignIn = (function() {
         Router.prototype.Events.on.apply(this, onArgs);
       }
     });
-    _.extend(this, getProperties(authClient, Router, options));
+    _.extend(this, getProperties(authClient, hooks, Router, options));
```
pradeepdewda-okta, line 190: can be passed a part of `options` ? `options.hooks = hooks`

src/widget/OktaSignIn.js
```diff
@@ -7,21 +7,22 @@ import buildRenderOptions from 'widget/buildRenderOptions';
...
-  function getProperties(authClient, Router, widgetOptions = {}) {
+  function getProperties(authClient, hooks, Router, widgetOptions = {}) {
     // Returns a promise that will resolve on success or reject on error
     function render(renderOptions, successFn, errorFn) {
       if (router) {
         throw new Error('An instance of the widget has already been rendered. Call remove() first.');
       }
 
-      const res = createRouter(Router, widgetOptions, renderOptions, authClient, successFn, errorFn);
+      const res = createRouter(Router, widgetOptions, renderOptions, authClient, successFn, errorFn, hooks);
```
pradeepdewda-okta, line 25: hooks are basically widget options. It seems it can be part of `widgetOptions` instead of adding a new argument

src/v2/BaseLoginRouter.js
aarongranick-okta: hooks *can* be defined in config, but they can also be set at runtime either before or after render. Since settings is created by the router (after render), it could not hold hook definitions added at runtime before render. For this reason I broke it apart into its own model.

src/widget/OktaSignIn.js
aarongranick-okta: widgetOptions is the widget config passed to constructor. Although hooks can be defined here, they can also be added at runtime.

src/widget/OktaSignIn.js
aarongranick-okta: options.hooks (set in widget config) is used to create the Hooks model which is available before or after render.

src/v2/models/AppState.js
aarongranick-okta: My expectation is that a common use of this feature will be for adding tracking or logging, so it makes sense to have it active on all views.

src/v2/models/AppState.js
aarongranick-okta: I have included a unit test to verify that passing an undefined hook will not error, and will be a no-op.

README.md
```diff
@@ -629,6 +632,50 @@ var signIn = new OktaSignIn(config);
...
+var signIn = new OktaSignIn(config);
+signIn.before('success-redirect', async () => {
+  // custom logic can go here. when the function resolves, execution will continue.
+});
+
+```
+
+### after
+
+> **Note**: This function is only supported when using the Okta Identity Engine
```
sherizada-okta, line 659: Can you link to that documentation, maybe the setup section?

docs/interaction_code_flow.md
```diff
@@ -45,6 +45,7 @@
   - [Feature flags](#feature-flags)
     - [features.showPasswordToggleOnSignInPage](#featuresshowpasswordtoggleonsigninpage)
     - [features.hideSignOutLinkInMFA](#featureshidesignoutlinkinmfa)
+    - [features.rememberMe](#featuresrememberme)
```
sherizada-okta, line 48: thanks!

README.md
```diff
@@ -629,6 +632,50 @@ var signIn = new OktaSignIn(config);
 // signIn.authClient.options.clientId === '{yourClientId}'
 ```
 
+### before
```
pradeepdewda-okta, line 635: Can we add more description about what hooks are for widget.. ? Basically adding following info..

1. Hooks can be added either on run time or via configs as below examples
2. Hooks can be used for ? eg. logging/tracking
3. Hooks blocks processing before/after a login step..

src/v2/BaseLoginRouter.js
```diff
@@ -125,7 +126,7 @@ export default Router.extend({
     // transform response
     const ionResponse = transformIdxResponse(this.settings, idxResponse, lastResponse);
 
-    this.appState.setIonResponse(ionResponse);
+    await this.appState.setIonResponse(ionResponse, this.hooks);
```
pradeepdewda-okta, line 129: It would be much cleaner if somehow hooks can be available under `ionResponse` object.. possible if we can pass this in `transformIdxResponse` instead?..

aarongranick-okta: why would it be cleaner if the hooks are part of the ionResponse? They are not related to the response, they are a configured part of the client state.

aarongranick-okta: if anything, they could be part of appState, but again because they need to be configured before or after render (when app state is created), I put them in their own model.

pradeepdewda-okta: I wanted to have `hooks` as part settings or options or something instead of standalone.. But it seems it's not possible because of how we want to execute and assign them..

README.md
```diff
@@ -1465,6 +1512,44 @@ features: {
 
 - **features.scrollOnError** - By default, errors will be scrolled into view. Set to `false` to disable this behavior.
 
+### Hooks
+
+> **Note**: Hooks are only supported when using the Okta Identity Engine
+
+Asynchronous callbacks can be invoked before or after a specific view is rendered. Hook callbacks block processing to run custom logic. Nomal execution will resume after the Promise returned from the callback function resolves.
```
pradeepdewda-okta, line 1519: typo: Nomal

test/unit/spec/Hooks_spec.js
```diff
@@ -0,0 +1,206 @@
...
+    });
+
+    describe('executeHooksAfter', () => {
+      it('does not throw if no hook was passed', () => {
+        expect(executeHooksAfter).not.toThrow();
+      });
+      it('does not throw if a hook with no before section was passed', () => {
+        expect(executeHooksAfter.bind(null, {})).not.toThrow();
+      });
+      it('executes afterr functions in order', async () => {
```
pradeepdewda-okta, line 182: typo: afterr

pradeepdewda-okta (approved):
Overall PR looks good to me.

src/v2/BaseLoginRouter.js
aarongranick-okta: It could be done with a bigger refactor, which I considered and even started, but I decided I did not want to mess with how Settings and AppState are constructed with this PR, to avoid possibility of creating regressions. If we constructed either of these objects before render they could be a good home for this feature. Settings was my first choice, AppState could also make sense (although it gets cleared regularly).  The main reason Settings is created late is because options from the constructor are merged with options passed to renderEl - a refactor would have to handle merging in renderEl options late. I'm not sure we have sufficient unit tests to be confident in making this change now. I can create a new ticket for a Settings refactor to include hooks, renderEl options, and unit tests.

aarongranick-okta: OKTA-432376

README.md
```diff
@@ -1514,9 +1522,9 @@ features: {
 
 ### Hooks
 
-> **Note**: Hooks are only supported when using the Okta Identity Engine
+> **Note**: Hooks are only supported when using the [Okta Identity Engine](#okta-identity-engine)
 
-Asynchronous callbacks can be invoked before or after a specific view is rendered. Hook callbacks block processing to run custom logic. Nomal execution will resume after the Promise returned from the callback function resolves.
+Asynchronous callbacks can be invoked before or after a specific view is rendered. Hooks can be used to add custom logic such as tracking, logging, or additional user input. Normal execution is blocked while the hooks is executing and will resume after the Promise returned from the hook function resolves. Hooks can be added via config, as shown below, or at runtime using the [before](#before) or [after](#after) methods. The full list of views can be found in [RemediationConstants.js](https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/RemediationConstants.js#L19).
```
pradeepdewda-okta, line 1527: 💯

PR #2191, opened 2021-09-08
Prevent back link in FastPass loopback view from showing multiple times.

jhe-okta:
## Description:

In loopback polling, when an error is responded, the back link is updated. In a rare case, loopback pollings respond with errors consecutively. To prevent a new back link to be added each time, remove the old back link first. 

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

back-link-before

back-link-after

### Reviewers:
@yannongli-okta 

### Issue:

- [OKTA-422726](https://oktainc.atlassian.net/browse/OKTA-422726)

### DIFF
--- a/test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
+++ b/test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
@@ -2,6 +2,7 @@ import { RequestLogger, RequestMock } from 'testcafe';
 import DeviceChallengePollPageObject from '../framework/page-objects/DeviceChallengePollPageObject';
 import identifyWithDeviceProbingLoopback from '../../../playground/mocks/data/idp/idx/identify-with-device-probing-loopback';
 import error from '../../../playground/mocks/data/idp/idx/error-403-access-denied';
+import { Constants } from '../framework/shared';
 
 const logger = RequestLogger(/introspect|probe|challenge/, { logRequestBody: true, stringifyRequestBody: true });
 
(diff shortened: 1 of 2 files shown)


pradeepdewda-okta (approved):

yannongli-okta (approved):
thanks for the change! probably add few words on how to reproduce it in the playground?

test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
```diff
@@ -10,8 +11,11 @@ const mock = RequestMock()
   .respond(identifyWithDeviceProbingLoopback)
   .onRequestTo(/\/idp\/idx\/authenticators\/poll/)
   .respond((req, res) => {
-    res.statusCode = '403';
-    res.setBody(error);
+    return new Promise((resolve) => setTimeout(function() {
+      res.statusCode = '403';
+      res.setBody(error);
+      resolve(res);
+    }, Constants.TESTCAFE_DEFAULT_AJAX_WAIT + 2000));
```
aarongranick-okta, line 18: why does this response need to be delayed?

pradeepdewda-okta: Probably to mimic the poll

aarongranick-okta: We should not have such delays in our tests. Could this be done better with a unit test that mocks time?

pradeepdewda-okta: While i also had concern regarding these delays.. but we need to test login flows and couldn't find any solutions yet for testcafe to mimic these delays.. Unit test won't be sufficient.

PR #2196, opened 2021-09-09
fix: Respect message type in introspect.

jhe-okta:
## Description:

In the identifier view, when a warning message is displayed, when the next error message is shown, the warning message is not removed. The fix includes:
- remove all the errors and warnings on clear form error
- have identifier view respect the message type in idx response when showing the callout

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Before:
warning-error-before

After:
warning-error-after

### Reviewers:
@pradeepdewda-okta 

### Issue:

- [OKTA-422719](https://oktainc.atlassian.net/browse/OKTA-422719)

### DIFF
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -250,6 +250,18 @@ const emailVerificationMocks = {
 
 
 // ===== IDX
+const fastpassUnassignedApp = {
+  '/idp/idx/introspect': [
+    'error-400-user-not-assigned',
+  ],
+  '/idp/idx/authenticators/okta-verify/launch': [
+    'error-400-user-not-assigned-2',
+  ],
+  '/idp/idx/identify': [
+    'error-400-user-not-assigned-3',
+  ],
+};
+
 // device probe: Windows authenticator with loopback server
 const windowAuthnLoopback = {
   '/idp/idx/introspect': [
--- a/src/v2/view-builder/internals/BaseForm.js
+++ b/src/v2/view-builder/internals/BaseForm.js
@@ -45,7 +45,9 @@ export default Form.extend({
   },
 
   handleClearFormError() {
-    if (this.$('.o-form-error-container').hasClass('o-form-has-errors')) {
+    const formErrorContainer = this.$('.o-form-error-container');
+    formErrorContainer.empty();
+    if (formErrorContainer.hasClass('o-form-has-errors')) {
       this.clearErrors();
     }
   },
--- a/src/v2/view-builder/views/IdentifierView.js
+++ b/src/v2/view-builder/views/IdentifierView.js
@@ -91,20 +91,15 @@ const Body = BaseForm.extend({
   },
 
   showMessages() {
-    /**
-     * Renders a warning callout for unknown user flow
-     * Note: Anytime we get back `messages` object along with identify view
-     * we would render it as a warning callout
-     * */
     const messagesObj = this.options.appState.get('messages');
     if (messagesObj?.value.length
         && this.options.appState.get('currentFormName') === 'identify') {
-      const content = messagesObj.value[0].message;
+      const displayMessageObj = messagesObj.value[0];
       const messageCallout = createCallout({
-        content: content,
-        type: 'warning',
+        content: displayMessageObj.message,
+        type: (displayMessageObj.class || '').toLowerCase(),
       });
-      this.add(messageCallout, '.o-form-error-container');
+      this.introspectMessage = this.add(messageCallout, '.o-form-error-container').last();
     }
   },
   /**
--- a/test/testcafe/spec/IdentifyUnknownUser_spec.js
+++ b/test/testcafe/spec/IdentifyUnknownUser_spec.js
@@ -1,6 +1,7 @@
 import IdentityPageObject from '../framework/page-objects/IdentityPageObject';
 import { checkConsoleMessages } from '../framework/shared';
 import unknownUser from '../../../playground/mocks/data/idp/idx/identify-unknown-user';
+import notAssignedApp from '../../../playground/mocks/data/idp/idx/error-400-user-not-assigned';
 import registeredUser from '../../../playground/mocks/data/idp/idx/authenticator-verification-select-authenticator.json';
 import identify from '../../../playground/mocks/data/idp/idx/identify';
 import { RequestMock } from 'testcafe';
@@ -13,8 +14,13 @@ const mock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/identify')
   .respond(registeredUser);
 
-fixture('Identify Unknown User')
-  .requestHooks(mock);
+const unassignedApplinkMock = RequestMock()
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(unknownUser)
+  .onRequestTo('http://localhost:3000/idp/idx/identify')
+  .respond(notAssignedApp);
+
+fixture('Identify Unknown User');
 
 async function setup(t) {
   const identityPage = new IdentityPageObject(t);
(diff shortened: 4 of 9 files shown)


src/v2/view-builder/internals/BaseForm.js
```diff
@@ -45,9 +45,8 @@ export default Form.extend({
   },
 
   handleClearFormError() {
-    if (this.$('.o-form-error-container').hasClass('o-form-has-errors')) {
-      this.clearErrors();
-    }
+    this.clearErrors();
+    this.introspectMessage && this.introspectMessage.remove();
```
pradeepdewda-okta, line 49: baseForm should not be aware of view specific actions..
can you try here instead 
```
    this.$el.find('.okta-form-infobox-error').remove();
```

playground/mocks/config/responseConfig.js
```diff
@@ -250,6 +250,18 @@ const emailVerificationMocks = {
 
 
 // ===== IDX
+const fastpassUnassignedApp = {
+  '/idp/idx/introspect': [
+    'error-400-user-not-assigned',
+  ],
+  '/idp/idx/authenticators/okta-verify/launch': [
+    'error-400-user-not-assigned-2',
```
pradeepdewda-okta, line 258: whats' the difference in these 3 data files ?

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -92,19 +92,21 @@ const Body = BaseForm.extend({
...
+     * The callout type is warning unless the message states differently
      * Note: Anytime we get back `messages` object along with identify view
      * we would render it as a warning callout
      * */
     const messagesObj = this.options.appState.get('messages');
     if (messagesObj?.value.length
         && this.options.appState.get('currentFormName') === 'identify') {
-      const content = messagesObj.value[0].message;
+      const displayMessageObj = messagesObj.value[0];
+      const messageType = (displayMessageObj.class || 'warning').toLowerCase();
```
pradeepdewda-okta, line 104: i think we should get rid of `warning` callout.. there is no use case to show warning message in this view.. just render the `error` callOut like any other view that overrides it.

src/v2/view-builder/internals/BaseForm.js
jhe-okta: Yes, but the error callout has a div wrap around the `.okta-form-infobox-error`. So removing `.okta-form-infobox-error` only does not clean up all the doms.

playground/mocks/config/responseConfig.js
jhe-okta: Just the time is different. Adding the mock and flow just to make sure that a new responds is returned and UI is rerendered.

src/v2/view-builder/internals/BaseForm.js
jhe-okta: Or I can rename `introspectMessage` to `calloutMessage` so it is less specific to the flow. What do you think?

playground/mocks/config/responseConfig.js
pradeepdewda-okta: if we can keep the same mock file.. would there be any issue ?

src/v2/view-builder/internals/BaseForm.js
pradeepdewda-okta: I was wondering this should have removed any existing message when user hit `save`.. 
https://github.com/okta/okta-signin-widget/pull/2196/files#diff-75984ad57a18502463b8e54df919333e998b41678372612bc528336229069c5bR58


Isn't the issue was that overridden `showMessages` function.. what happen if you remove that function ?

src/v2/view-builder/views/IdentifierView.js
jhe-okta: I don't know enough whether warning should be removed. The change here is just to respect whatever the idx api message type is. So it should work for both warning and error cases.

playground/mocks/config/responseConfig.js
jhe-okta: What's the issue?

src/v2/view-builder/internals/BaseForm.js
jhe-okta: Hitting `save` removes the existing errors. 

Removing `showMessages` from `IdentifierView` does not work because the fall back method from `BaseForm` shows the message in plain text, instead of in a callout format.

playground/mocks/config/responseConfig.js
pradeepdewda-okta: I mean why create multiple files for `error-400-user-not-assigned`

src/v2/view-builder/views/IdentifierView.js
pradeepdewda-okta: there is no `WARN` type in backend.. it's only `INFO` and `ERROR`.. the comment clarifies that it was something hardcoded.

src/v2/view-builder/internals/BaseForm.js
pradeepdewda-okta: Cool. so i feel the right solution to just override the showMessages in IdentifierView with a function that renders error callout.. We can safely remove any `warning` types because there is no `warn` in backend..

playground/mocks/config/responseConfig.js
jhe-okta: It is to mimic the flow where 1. an app link for an unassigned user is loaded, then 2. the user click OV button or submit username form.

pradeepdewda-okta: but we can use the same mock file right ?

jhe-okta: If it is the same mock file, we cannot guarantee the UI has the same rerender logic as with different mocks.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -90,21 +90,21 @@ const Body = BaseForm.extend({
     }
   },
 
+  handleClearFormError() {
```
pradeepdewda-okta, line 93: instead of overriding `handleClearFormError` add `this.$el.find('.o-form-error-container').empty();` to the `handleClearFormError` for `BaseForm` before `if` statement

test/testcafe/spec/IdentifyUnknownUser_spec.js
```diff
@@ -34,6 +34,7 @@ test('should show messages callout for unknown user', async t => {
   await identityPage.clickNextButton();
   await t.expect(identityPage.getUnknownUserCalloutContent())
     .eql('Unable to sign in');
+  await t.expect(identityPage.hasUnknownUserErrorCallout()).eql(true);
```
pradeepdewda-okta, line 37: can you add a check that there is only one error message.. and not two.

pradeepdewda-okta (approved):
add one check in test before merge.

test/testcafe/spec/IdentifyUnknownUser_spec.js
jhe-okta: Done. https://github.com/okta/okta-signin-widget/pull/2196/files#diff-d85a7b24825cc1d5d0fb184084c66b2d1f7607c14befff9cbd8ac383a8edf716R48

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -91,20 +91,15 @@ const Body = BaseForm.extend({
...
-      const content = messagesObj.value[0].message;
+      const displayMessageObj = messagesObj.value[0];
       const messageCallout = createCallout({
-        content: content,
-        type: 'warning',
+        content: displayMessageObj.message,
+        type: (displayMessageObj.class || '').toLowerCase(),
       });
-      this.add(messageCallout, '.o-form-error-container');
+      this.introspectMessage = this.add(messageCallout, '.o-form-error-container').last();
```
pradeepdewda-okta, line 102: sorry forgot this.. we can remove this assignment now

PR #2204, opened 2021-09-16
Don't show widget until loaded

denysoblohin-okta:
## Description:

Don't show widget in initial state (blank line w/ or w/o logo) until is ready and rendered. 

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

Before:

https://user-images.githubusercontent.com/72614880/133684081-3249a56f-33ac-4fec-9bb4-27a87e27417e.mov

After:

https://user-images.githubusercontent.com/72614880/133683195-c48492fe-a81d-40e8-8051-a910eaf0edb9.mov

### Reviewers:

@aarongranick-okta 

### Issue:

- [OKTA-416911](https://oktainc.atlassian.net/browse/OKTA-416911)

### Additional notes

1) To test changes locally uncomment this:
https://github.com/okta/okta-signin-widget/blob/8d9bbd5ff0e257e1391b3047750e02ee9b180c7c/playground/mocks/config/templateHelper.js#L41

2) An additional enhancement to show loading spinner for slow internet connection (not in scope of this task, could be in future) was partially implemented for loading of i18n file:
https://github.com/okta/okta-signin-widget/blob/8d9bbd5ff0e257e1391b3047750e02ee9b180c7c/src/util/LanguageUtil.js#L9-L12
Note that the loading spinner will not be shown after this PR (because whole widget will be hidden until all needed data has been loaded - well-known, instrospect, interact, language file).

### DIFF
--- a/src/v2/BaseLoginRouter.js
+++ b/src/v2/BaseLoginRouter.js
@@ -219,14 +219,6 @@ export default Router.extend({
       this.el = this.header.render().getContentEl();
     }
 
-    setTimeout(() => {
-      const hasRendered = this.controller && this.controller.constructor === Controller && this.controller.rendered();
-      if (!this.header.isLoadingBeacon() && !hasRendered) {
-        this.show();
-        this.header.setLoadingBeacon(true);
-      }
-    }, 1000);
-
     // If we need to load a language (or apply custom i18n overrides), do
     // this now and re-run render after it's finished.
     if (!Bundles.isLoaded(this.settings.get('languageCode'))) {
@@ -256,8 +248,6 @@ export default Router.extend({
 
       ColorsUtil.addStyle(colors);
     }
-
-    this.header.removeLoadingBeacon();
 
     // render Controller
     this.unload();
--- a/src/v2/BaseLoginRouter.js
+++ b/src/v2/BaseLoginRouter.js
@@ -79,6 +79,8 @@ export default Router.extend({
       settings: this.settings,
     });
 
+    this.hide();
+
     configIdxJsClient(this.appState);
     this.listenTo(this.appState, 'updateAppState', this.handleUpdateAppState);
     this.listenTo(this.appState, 'remediationError', this.handleIdxResponseFailure);
@@ -252,6 +254,8 @@ export default Router.extend({
     this.listenTo(this.controller, 'all', this.trigger);
 
     this.controller.render();
+
+    this.show();
   },
 
   /**
--- a/src/v2/BaseLoginRouter.js
+++ b/src/v2/BaseLoginRouter.js
@@ -81,13 +81,6 @@ export default Router.extend({
 
     // Hide until unitial render
     this.hide();
-    // Show if loading state was set to true before initial render
-    // (eg. because of long load of language)
-    this.listenTo(this.appState, 'loading', function(isLoading) {
-      if (isLoading) {
-        this.show();
-      }
-    });
 
     configIdxJsClient(this.appState);
     this.listenTo(this.appState, 'updateAppState', this.handleUpdateAppState);
--- a/src/v2/BaseLoginRouter.js
+++ b/src/v2/BaseLoginRouter.js
@@ -205,7 +205,7 @@ export default Router.extend({
     // }
   },
 
-  /* eslint max-statements: [2, 21] */
+  /* eslint max-statements: [2, 22] */
   render: async function(Controller, options = {}) {
     // If url changes then widget assumes that user's intention was to initiate a new login flow,
     // so clear stored token to use the latest token.
--- a/src/v2/BaseLoginRouter.js
+++ b/src/v2/BaseLoginRouter.js
@@ -79,7 +79,15 @@ export default Router.extend({
       settings: this.settings,
     });
 
+    // Hide until unitial render
     this.hide();
+    // Show if loading state was set to true before initial render
+    // (eg. because of long load of language)
+    this.listenTo(this.appState, 'loading', function(isLoading) {
+      if (isLoading) {
+        this.show();
+      }
+    });
 
     configIdxJsClient(this.appState);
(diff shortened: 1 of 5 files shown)


pradeepdewda-okta:
@denysoblohin-okta the better way to approach is to show something instead of let user staring the white space... If network is not good then it takes time to load and if user don't see anything, it also creates bad experience..

Let's do something like this and provide a better solution to this problem. 

Screen Shot 2021-09-21 at 4 36 15 PM
Screen Shot 2021-09-21 at 4 36 48 PM

sherizada-okta (commented):
Agree with @pradeepdewda-okta , we either need a loading state or even better, load it faster

denysoblohin-okta:
@sherizada-okta Showing loading spinner is out of scope of task OKTA-416911. The goal for now is just hold showing the widget until it’s ready to render.

deekoder:
@pradeepdewda-okta - this is an important point you are making. Its not that we are going to ignore showing a loading animation. But right now the issue is that it shows empty blank boxes instead of a spinner. There are 2 buckets of loading experiences. 1. When the loading is fast enough, we do not need any animation. 2. When the loading is not fast enough, then we need a visual way to inform the user that we will be loading the widget as soon as we are able to. This PR is to address #1. A separate PR will handle #2. HTH.

src/v2/BaseLoginRouter.js
```diff
@@ -238,6 +241,9 @@ export default Router.extend({
       ColorsUtil.addStyle(colors);
     }
 
+    // Show before initial render
+    this.show();
```
pradeepdewda-okta, line 245: If render fails due to some error, then there is no feedback now to enduser ? They will just see the white screen ?

test/unit/spec/LoginRouter_spec.js
```diff
@@ -2244,4 +2246,22 @@ Expect.describe('LoginRouter', function() {
       });
     });
   });
+
+  itp('should not be visible until initial render', async function() {
```
pradeepdewda-okta, line 2250: add a test for non happy path..

test/unit/spec/v2/WidgetRouter_spec.js
```diff
@@ -0,0 +1,66 @@
+import { _ } from 'okta';
+import WidgetRouter from 'v2/WidgetRouter';
+import FormController from 'v2/controllers/FormController';
+import $sandbox from 'sandbox';
+import createAuthClient from 'widget/createAuthClient';
+import XHRIdentifyWithPassword
+  from '../../../../playground/mocks/data/idp/idx/identify-with-password.json';
+
+describe('v2/WidgetRouter', function() {
```
pradeepdewda-okta, line 9: thanks for adding this.. 👍

test/unit/spec/v2/WidgetRouter_spec.js
```diff
@@ -0,0 +1,66 @@
...
+
+  beforeEach(function() {
+    testContext = {};
+  });
+
+  afterEach(function() {
+    $sandbox.empty();
+  });
+
+  it('should not be visible until initial render', async function() {
```
pradeepdewda-okta, line 59: what would happen on error ?

pradeepdewda-okta:
@denysoblohin-okta can you also add a video that show rendering on slower network connection... and also a video for a network error case..

pradeepdewda-okta:
> @pradeepdewda-okta - this is an important point you are making. It's not that we are going to ignore showing a loading animation. But right now the issue is that it shows empty blank boxes instead of a spinner. There are 2 buckets of loading experiences. 1. When the loading is fast enough, we do not need any animation. 2. When the loading is not fast enough, then we need a visual way to inform the user that we will be loading the widget as soon as we are able to. This PR is to address #1. A separate PR will handle #2. HTH.

I get the point, while i also think just adding a placeholder should not take too much effort. If there is a super urgency on this to go in, i would be fine with this change. Although as per video, we are showing a blank screen for 2 secs..

src/v2/BaseLoginRouter.js
denysoblohin-okta: No, user will see error, bacause `startLoginFlow` is being caught
https://github.com/okta/okta-signin-widget/blob/cb805a3dda39ffd61366180f6a6006aedce16b11/src/v2/BaseLoginRouter.js#L227-L228

Loading language is not handling rejections
https://github.com/okta/okta-signin-widget/blob/cb805a3dda39ffd61366180f6a6006aedce16b11/src/v2/BaseLoginRouter.js#L217
https://github.com/okta/okta-signin-widget/blob/cb805a3dda39ffd61366180f6a6006aedce16b11/src/util/LanguageUtil.js#L21
In case of error for loading language before change user will see infinite spinner, not he will see nothing

test/unit/spec/LoginRouter_spec.js
denysoblohin-okta: Looks like there is no proper error handling for v1 in case of error in `fetchInitialData ` - controller will not be rendered
https://github.com/okta/okta-signin-widget/blob/551c766ae892bb62f5b5c2ee32b2908fb9cc7128/src/util/BaseLoginRouter.js#L275-L277

denysoblohin-okta:
@pradeepdewda-okta 
> can you also add a video that show rendering on slower network connection... and also a video for a network error case..
Video 2 in PR description is done by simulating slow network.
Video for error case:

https://user-images.githubusercontent.com/72614880/134747498-15b45e40-ce76-4e74-af93-e4e854427ac3.mov

aarongranick-okta (approved):

test/unit/spec/LoginRouter_spec.js
aarongranick-okta: I have created a new issue OKTA-432776 for adding the error handling and related tests

PR #2214, opened 2021-09-21
Skip custom idp factor verify button when SKIP_IDP_FACTOR_VERIFICATION_BUTTON FF is on

mikiasmohamed-okta:
## Description:

Skip custom idp factor verify button when SKIP_IDP_FACTOR_VERIFICATION_BUTTON FF is on

When the SKIP_IDP_FACTOR_VERIFICATION_BUTTON feature flag is turned on
it hides the 'verify' button of the custom idp verification dialog and
automatically redirects the user to the idp service provider.

Resolves: OKTA-429138

## PR Checklist

- [ ] Added unit tests?
- [ ] Added e2e tests

### Reviewers:
@mauriciocastillosilva-okta 

### Issue:
- [OKTA-429138](https://oktainc.atlassian.net/browse/OKTA-429138)


pradeepdewda-okta (approved):
The change looks good to me. Needs to be tested in okta-core as well as, new selenium test should be added in okta-core for the flow (new FF)

@jmelberg-okta can you also take a look at this PR

jmelberg-okta (commented):
Can we please validate this new v1 behavior with tests?

src/VerifyCustomFactorController.js
```diff
@@ -31,10 +31,20 @@ export default FormController.extend({
...
+        this.set('factorType', 'claims_provider');
+        this.save();
+      }
     },
 
     save: function() {
       const rememberDevice = !!this.get('rememberDevice');
+      const self = this;
+      if (this.settings.get('features.skipIdpFactorVerificationBtn')) {
+        this.trigger('request');
```
mauriciocastillosilva-okta, line 46: why do we need to trigger a request? If I am following this right, when we automatically call save in `initialize` the request will be made as if we had clicked the button (if that is not true, then maybe we need to move that code to `postRender`?

src/VerifyCustomFactorController.js
```diff
@@ -57,6 +67,9 @@ export default FormController.extend({
             }
           })
           .catch(function(err) {
+            if (self.settings.get('features.skipIdpFactorVerificationBtn')) {
+              self.trigger('error', self, err);
```
mauriciocastillosilva-okta, line 71: same question here, why do we need the extra trigger? If we throw an error, the form should show the error already. Is that not true?

src/VerifyCustomFactorController.js
```diff
@@ -71,7 +84,11 @@ export default FormController.extend({
     });
     const vendorName = factor.get('vendorName');
     const saveText = loc('mfa.challenge.verify', 'login');
-    const subtitle = loc('verify.customFactor.subtitle', 'login', [vendorName]);
+    let subtitle = loc('verify.customFactor.subtitle', 'login', [vendorName]);
+
+    if (this.settings.get('features.skipIdpFactorVerificationBtn')) {
+      subtitle = loc('verify.customFactor.subtitle.redirect', 'login', [vendorName]);
```
mauriciocastillosilva-okta, line 90: I actually think we need both "subtitles", the reason is, when we redirect we show "redirecting to..." but if there is an error, we should show the button and the subtitle saying "Clicking below will redirect to ..." so the user can try again if needed (you can see on the screenshot you have on this PR that the error message reads "review the form and make corrections", which is not ideal for this screen, but in any case we should give the user a way to retry. Feel free to confirm with UX)

mauriciocastillosilva-okta (changes requested):
can we write some tests, maybe on [MfaVerify_spec.js](https://github.com/okta/okta-signin-widget/blob/master/test/unit/spec/MfaVerify_spec.js#L4388)?

src/VerifyCustomFactorController.js
mikiasmohamed-okta: @mauriciocastillosilva-okta those (request and trigger) are modelEvents I added to start and stop the loading spinner.

When request is triggered `_startVerification ` is called and when `error` is triggered `_stopVerification `. This is a pattern employed in other controllers of the signin-widget as well.

mauriciocastillosilva-okta: yeah, I see you added them, but I don't think we need them. Take a look at [BaseLoginModel](https://github.com/okta/okta-signin-widget/blob/master/src/models/BaseLoginModel.js#L59), in `manageTransaction` there is already a catch for any error and we trigger `error` there, so this will be a duplicate event we are triggering here

src/VerifyCustomFactorController.js
```diff
@@ -108,6 +134,19 @@ export default FormController.extend({
     };
   },
 
+  __lastFailedChallengeFactorData__: null,
```
jmelberg-okta, line 137: Any reason we can't just rely on the presence of appState's `lastFailedChallengeFactorData`?

src/models/Settings.js
```diff
@@ -86,6 +86,7 @@ export default Model.extend({
     'features.multiOptionalFactorEnroll': ['boolean', true, false],
     'features.deviceFingerprinting': ['boolean', false, false],
     'features.hideSignOutLinkInMFA': ['boolean', false, false],
+    'features.skipIdpFactorVerificationBtn': ['boolean', false, false],
```
jmelberg-okta, line 89: Do we need to document this under our [features](https://github.com/okta/okta-signin-widget/#feature-flags)?

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,263 @@
...
+
+      itp('does not redirect to third party automatically', function() {
+        spyOn(SharedUtil, 'redirect');
+        return setupClaimsProviderFactorWithIntrospect({ 'features.skipIdpFactorVerificationBtn': false }, null)
+          .then(function(test) {
+            test.setNextResponse([resChallengeClaimsProvider, resSuccess]);
+            test.form.initialize();
+            expect(SharedUtil.redirect).not.toHaveBeenCalledWith(
+              'http://rain.okta1.com:1802/sso/idps/idpId?stateToken=testStateToken'
+            );
```
jmelberg-okta, line 258: We don't expect a redirect to occur at all, right? If that is the case, let's change this to:

```suggestion
            expect(SharedUtil.redirect).not.toHaveBeenCalled();
```

or

```suggestion
            expect(SharedUtil.redirect).toHaveBeenCalledTimes(0);
```

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,263 @@
...
+  function expectSubtitleToBe(test, desiredSubtitle) {
+    expect(test.form.subtitleText()).toBe(desiredSubtitle);
+  }
+
+  async function setup(res, selectedFactorProps, settings, lastFailedChallengeFactorData, languagesResponse, startRouter) {
+    const setNextResponse = Util.mockAjax();
+    const baseUrl = 'https://foo.com';
+    const authClient = createAuthClient({ issuer: baseUrl, transformErrorXHR: LoginUtil.transformErrorXHR });
+    const successSpy = jasmine.createSpy('success');
+    const afterErrorHandler = jasmine.createSpy('afterErrorHandler');
```
jmelberg-okta, line 67: Are we using jest to run these tests? If so, let's please use Jest objects instead of Jasmine:

```suggestion
    const successSpy = jest.fn();
    const afterErrorHandler = jest.fn();
```

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,263 @@
...
+      itp('shows the right auto redirect subtitle when lastFailedChallengeFactorData has value', function() {
+        return setupClaimsProviderFactorWithIntrospect({ 'features.skipIdpFactorVerificationBtn': true }, {
+          errorMessage: 'lastFailedChallengeFactorData error message',
+        }).then(function(test) {
+          expectSubtitleToBe(test, 'Clicking below will redirect to verification with IDP factor');
+        });
+      });
+
+      itp('redirects automatically to third party when lastFailedChallengeFactorData is null', function() {
+        spyOn(SharedUtil, 'redirect');
```
jmelberg-okta, line 172: Similar to comment above - please use Jest objects instead of Jasmine:

```suggestion
        jest.spyOn(SharedUtil, 'redirect');
```

jmelberg-okta (approved):
Thanks for adding a test! Couple more questions.

Approving from the i18n/string side, so please wait for another codeowner to approve before merging.

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    setNextResponse(res);
+    if (languagesResponse) {
+      setNextResponse(languagesResponse);
+    }
+    router.refreshAuthState('dummy-token');
+    await Expect.waitForMfaVerify();
+
+    router.appState.set('lastFailedChallengeFactorData', lastFailedChallengeFactorData);
```
mauriciocastillosilva-okta, line 85: do we need to explicitly set it? If the response is a failure, then this should be set for you (happens in [RouterUtil](https://github.com/okta/okta-signin-widget/blob/master/src/util/RouterUtil.js#L186))

src/models/Settings.js
mikiasmohamed-okta: @mohitganotra-okta can you please confirm?

src/VerifyCustomFactorController.js
mikiasmohamed-okta: @mauriciocastillosilva-okta @jmelberg-okta 
I was not able to directly access `this.options.appState.get('lastFailedChallengeFactorData')` inside postRender

test/unit/spec/VerifyCustomFactor_spec.js
mikiasmohamed-okta: I had to set it explicitly.

src/VerifyCustomFactorController.js
```diff
@@ -108,6 +134,19 @@ export default FormController.extend({
...
+
+  postRender() {
+    if (this.settings.get('features.skipIdpFactorVerificationBtn')) {
+      this.$('.o-form-button-bar').hide();
+      this.$('.okta-waiting-spinner').show();
+      if (this.__lastFailedChallengeFactorData__) {
```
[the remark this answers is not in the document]
mikiasmohamed-okta: I have moved those toggles into postRender as per your suggestion with the following two conditions:

- this.options.appState.get('lastFailedChallengeFactorData') is not directly available in postRender. I had to keep the __lastFailedChallengeFactorData__ which is set in the `initialize`
- i am listening to any errors on the model inside the initialize function. when the FF is on

test/unit/spec/VerifyCustomFactor_spec.js
mikiasmohamed-okta: @jmelberg-okta can you resolve this comment please.

test/unit/spec/VerifyCustomFactor_spec.js
mauriciocastillosilva-okta: talked offline about this. Unit tests need this explicitly since it is set in RouterUtil, but these tests do not go through that util file.

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,282 @@
...
+        errorMessage: 'Verify failed.'
+      };
+
+      let settings = {};
+
+      beforeEach(() => {
+        settings = { 'features.skipIdpFactorVerificationBtn': true };
+      });
+
+      itp('hides the button bar when', function() {
```
mauriciocastillosilva-okta, line 144: just "hides the button bar". There is an extra "when" at the end of the sentence that we can remove

test/unit/helpers/dom/VerifyCustomFactorForm.js
```diff
@@ -0,0 +1,22 @@
...
+  buttonBar: function() {
+    return this.$('.o-form-button-bar');
+  },
+
+  hasErrorBox: function() {
+    return this.el('o-form-error-container').find('.infobox-error').length > 0;
+  },
+
+  hasSpinner() {
+    return this.$('.okta-waiting-spinner');
```
mauriciocastillosilva-okta, line 16: this should return a boolean, so should be 
```
return this.$('.okta-waiting-spinner').length > 0;
```
otherwise we can rename the function to be `spinner` or `loadingSpinner` or something like that
(I vote for returning a boolean)

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,282 @@
...
+      form: forms,
+      beacon: beacon,
+      ac: authClient,
+      setNextResponse: setNextResponse,
+      successSpy: successSpy,
+      afterErrorHandler: afterErrorHandler,
+    };
+  }
+
+  const setupClaimsProviderFactorWithIntrospect = _.partial(setup, resAllFactors, {
```
mauriciocastillosilva-okta, line 127: nit: this spect is just for Clamis Provider, so we can remove this function and integrate the logic to `setup()`.

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,282 @@
...
+          })
+          .then(function() {
+            expect(SharedUtil.redirect).toHaveBeenCalledWith(
+              'http://rain.okta1.com:1802/sso/idps/idpId?stateToken=testStateToken'
+            );
+          });
+      });
+    });
+
+    Expect.describe('when SKIP_IDP_FACTOR_VERIFICATION_BUTTON FF is off', function() {
```
mauriciocastillosilva-okta, line 199: let's change this to "when features.skipIdpFactorVerificationBtn is false"

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,268 @@
...
+            );
+          });
+      });
+
+      itp('does not redirect to third party automatically', function() {
+        jest.spyOn(SharedUtil, 'redirect');
+        return setup(settings)
+          .then(function(test) {
+            test.setNextResponse([resChallengeClaimsProvider, resSuccess]);
+            test.form.initialize();
```
mauriciocastillosilva-okta, line 232: as mentioned in a previous comment, we shouldn't be calling `initialize` explicitly, that function is called when the controller is rendered.

src/VerifyCustomFactorController.js
```diff
@@ -108,6 +133,18 @@ export default FormController.extend({
     };
   },
 
+  postRender() {
+    if (this.settings.get('features.skipIdpFactorVerificationBtn')) {
+      if (this.options.appState.get('lastFailedChallengeFactorData')) {
```
mauriciocastillosilva-okta, line 138: nit: I don't think this `if` is needed as is. In `formChildren` above, you add the spinner only if `skipIdpFactorVerificationBtn=true` and `!lastFailedChallengeFactorData`. So here in postRender we only need to hide the button bar and show the spinner of those 2 conditions are met, otherwise, there are no changes to what we currently have
```
if (this.settings.get('features.skipIdpFactorVerificationBtn') && !this.options.appState.get('lastFailedChallengeFactorData')) {
  this.$('.o-form-button-bar').hide();
  this.$('.okta-waiting-spinner').show();
}
```
right?

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,268 @@
...
+        xhr: resError,
+      },
+    ]);
+  }
+
+  function expectSubtitleToBe(test, desiredSubtitle) {
+    expect(test.form.subtitleText()).toBe(desiredSubtitle);
+  }
+
+  async function setup(settings, lastFailedChallengeFactorData, languagesResponse , startRouter) {
```
mauriciocastillosilva-okta, line 62: I don't see any tests passing `languagesResponse` or `startRouter`, so we can probably get rid of them

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,268 @@
...
+      itp('hides the button bar', function() {
+        return setup(settings)
+          .then(function(test) {
+            Expect.isNotVisible(test.form.buttonBar());
+          });
+      });
+
+      itp('shows the button bar when lastFailedChallengeFactorData is not null', function() {
+        return setup(settings ,lastFailedChallengeFactorData).then(function(test) {
+          expect(test.form.$('.o-form-button-bar').hasClass('hide')).toBe(false);
```
mauriciocastillosilva-okta, line 139: nit: we can use `test.form.buttonBar()` instead of `test.form.$('.o-form-button-bar')`

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,268 @@
...
+        return setup(settings, lastFailedChallengeFactorData).then(function(test) {
+          expectSubtitleToBe(test, 'Clicking below will redirect to verification with IDP factor');
+        });
+      });
+
+      itp('redirects automatically to third party', function() {
+        spyOn(SharedUtil, 'redirect');
+        return setup(settings)
+          .then(function(test) {
+            test.setNextResponse([resChallengeClaimsProvider, resSuccess]);
```
mauriciocastillosilva-okta, line 173: I don't think you need this line. The `setup` function gets all the way to challenge claims provider, and the redirect happens automatically, which will call `SharedUtil.redirect` already. This line has no effect on the test

test/unit/spec/VerifyCustomFactor_spec.js
```diff
@@ -0,0 +1,268 @@
...
+              'http://rain.okta1.com:1802/sso/idps/idpId?stateToken=testStateToken'
+            );
+          });
+      });
+
+      itp('does not redirect to third party automatically', function() {
+        jest.spyOn(SharedUtil, 'redirect');
+        return setup(settings)
+          .then(function(test) {
+            test.setNextResponse([resChallengeClaimsProvider, resSuccess]);
```
mauriciocastillosilva-okta, line 231: I don't think this line is needed. What we are trying to test is that we are in the challenge claims factor view, and we didn't redirect. So we don't need to queue responses if there are no requests made

PR #2219, opened 2021-09-23
Update enroll profile update view set secondEmail to empty string by …

pradeepdewda-okta:
## Description:
Update enroll profile view to set secondEmail to empty string by default. So that when the widget form contains all optional fields and user do not enter anything and click `finish` button instead of `skip`, the request payload should contains secondEmail property as empty.

This allows backend to identify that user skipped setting up secondEmail and now default to empty.

1. add test
2. fix mocks

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Screen Shot 2021-09-22 at 2 50 45 PM

### Issue:

- [OKTA-430124](https://oktainc.atlassian.net/browse/OKTA-430124)

### DIFF
--- a/playground/mocks/data/idp/idx/enroll-profile-update-all-optional-params.json
+++ b/playground/mocks/data/idp/idx/enroll-profile-update-all-optional-params.json
@@ -11,7 +11,7 @@
               "create-form"
            ],
            "name":"profile-update",
-           "href":"http://idx.okta1.com:1802/idp/idx/enroll/new",
+           "href":"http://localhost:3000/idp/idx/enroll/new",
            "method":"POST",
            "produces":"application/ion+json; okta-version=1.0.0",
            "value":[
@@ -37,7 +37,8 @@
                           "desc":"Use a second email to recover your account in case you become locked out. This email must be verified after set up.",
                           "required":false,
                           "minLength":1,
-                          "maxLength":100
+                          "maxLength":100,
+                          "descKey": "secondary.email.desc"
                        }
                     ]
                  }
@@ -96,7 +97,7 @@
         "create-form"
      ],
      "name":"cancel",
-     "href":"http://idx.okta1.com:1802/idp/idx/cancel",
+     "href":"http://localhost:3000/idp/idx/cancel",
      "method":"POST",
      "produces":"application/ion+json; okta-version=1.0.0",
      "value":[
--- a/src/v2/view-builder/views/EnrollProfileUpdateView.js
+++ b/src/v2/view-builder/views/EnrollProfileUpdateView.js
@@ -51,5 +51,16 @@ const Footer = BaseFooter.extend({
 
 export default BaseView.extend({
   Body,
-  Footer
+  Footer,
+
+  postRender() {
+    BaseView.prototype.postRender.apply(this, arguments);
+    /**
+     * As per requirement of this flow set secondEmail default to empty string, if exists in remediation
+     * ideally server should have passed default string in remediation
+     */
+    if (this.options.appState.getSchemaByName('userProfile.secondEmail')) {
+      this.model.set('userProfile.secondEmail', '');
+    }
+  }
 });
--- a/test/testcafe/spec/EnrollProfileUpdateView_spec.js
+++ b/test/testcafe/spec/EnrollProfileUpdateView_spec.js
@@ -4,6 +4,7 @@ import EnrollProfileUpdateViewPageObject from '../framework/page-objects/EnrollP
 import xhrIdentify from '../../../playground/mocks/data/idp/idx/identify-with-password';
 import xhrEnrollProfileUpdate from '../../../playground/mocks/data/idp/idx/enroll-profile-update-params';
 import xhrEnrollProfileUpdateAllOptional from '../../../playground/mocks/data/idp/idx/enroll-profile-update-all-optional-params';
+import xhrEnrollProfileUpdateSuccess from '../../../playground/mocks/data/idp/idx/success-with-app-user';
 
 const xhrEnrollProfileUpdateMock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/introspect')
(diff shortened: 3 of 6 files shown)


src/v2/view-builder/views/EnrollProfileUpdateView.js
```diff
@@ -51,5 +51,12 @@ const Footer = BaseFooter.extend({
...
 export default BaseView.extend({
   Body,
-  Footer
+  Footer,
+
+  postRender() {
+    BaseView.prototype.postRender.apply(this, arguments);
+
+    // As per requirement of this flow set secondEmail default to empty string
+    this.model.set('userProfile.secondEmail', '');
```
haoyuli-okta, line 60: if no input on the secondary email attribute from profile enrollment flow, and we click the finish/submit button, then we set an empty string for secondEmail key, correct?

pradeepdewda-okta: It's always set to empty first.. if user input something and then click finish, then input value would be used.

src/v2/view-builder/views/EnrollProfileUpdateView.js
```diff
@@ -51,5 +51,19 @@ const Footer = BaseFooter.extend({
...
+
+  postRender() {
+    BaseView.prototype.postRender.apply(this, arguments);
+    const uiSchema = this.options.currentViewState.uiSchema;
+    /**
+     * As per requirement of this flow set secondEmail default to empty string
+     * ideally server should have passed default string in remediation
+     */
+    uiSchema.forEach(input => {
+      if (input.name === 'userProfile.secondEmail') {
```
sherizada-okta, line 64: What's the value of secondEmail at this point, before setting it to an empty string?

sherizada-okta: Will secondEmail always be sent by the server?

src/v2/view-builder/views/EnrollProfileUpdateView.js
```diff
@@ -51,5 +51,19 @@ const Footer = BaseFooter.extend({
...
+        this.model.set('userProfile.secondEmail', '');
+      }
+    });
```
sherizada-okta, line 67:
```suggestion
    const hasSecondEmail = uiSchema.find(({ name })=> name === 'userProfile.secondEmail');
    if (hasSecondEmail) this.model.set('userProfile.secondEmail', '');
```

test/testcafe/spec/EnrollProfileUpdateView_spec.js
```diff
@@ -65,4 +74,25 @@ test.requestHooks(requestLogger, xhrEnrollProfileUpdateAllOptionalMock)('should
...
+  await enrollProfileUpdatePage.clickFinishButton();
+  const req = requestLogger.requests[0].request;
+  await t.expect(req.url).eql('http://localhost:3000/idp/idx/enroll/new');
+  const reqBody = JSON.parse(req.body);
+  await t.expect(reqBody).eql({
+    userProfile: {
+      secondEmail: '',
+    },
+    stateHandle: '02m5bP1_PZfgeox0qFcNeh5Zr71MB3gs3tw7kCsL2G',
+  });
```
sherizada-okta, line 96: Test the request body when there are required fields?

src/v2/view-builder/views/EnrollProfileUpdateView.js
pradeepdewda-okta: So this flow has a special case for widget (first form that has all fields as optional) where backend expect `secondEmail` to be sent as `empty` if user clicked on `finish` button instead of `skip`.. Before setting it to the empty the `secondEmail` doesn't exist and hence the only value sent back to the server is `stateHandle`..

>Will secondEmail always be sent by and to the server?I

If it exists in remediation, then yes.

src/v2/view-builder/views/EnrollProfileUpdateView.js
pradeepdewda-okta: I am writing a prototype func `getSchemaByName`.. so that it can be used by any views.

test/testcafe/spec/EnrollProfileUpdateView_spec.js
pradeepdewda-okta: where there are required field it's the default behavior and it always exists in payload like `stateHandle`.. No need to test that IMO

src/v2/view-builder/views/EnrollProfileUpdateView.js
kamaldeepcheema-okta: Maybe an obvious question but just want to be sure, is there a scenario where the server might set a value for this and we override it to an empty string here?

pradeepdewda-okta: @kamaldeepcheema-okta that won't happen, since this is an enrollment flow which means prompt for enrollment if not already enrolled.. Enrolled user will not see `secondEmail` field..

kamaldeepcheema-okta: I thought so, thanks for confirming!

kamaldeepcheema-okta (approved):
Lgtm pending the `getSchemaByName` change that you will add.

pradeepdewda-okta:
> Lgtm pending the `getSchemaByName` change that you will add.

Done.

test/unit/spec/v2/models/AppState_spec.js
```diff
@@ -5,6 +5,7 @@ import XHRAuthenticatorChallengOktaVerify
 import XHRAuthenticatorEnrollOktaVerify
   from '../../../../../playground/mocks/data/idp/idx/authenticator-enroll-ov-sms.json';
 import { FORMS_FOR_VERIFICATION, FORMS_WITHOUT_SIGNOUT } from 'v2/ion/RemediationConstants';
+import { jest } from '@jest/globals';
```
sherizada-okta: This is a new pattern, did you editor do this?

pradeepdewda-okta: yeah that could have been the problem.. Reverted that and reran on bacon

playground/mocks/data/idp/idx/enroll-profile-update-all-optional-params.json
```diff
@@ -37,7 +37,8 @@
                           "desc":"Use a second email to recover your account in case you become locked out. This email must be verified after set up.",
                           "required":false,
                           "minLength":1,
-                          "maxLength":100
+                          "maxLength":100,
+                          "descKey": "secondary.email.desc"
```
sherizada-okta, line 41: Does `descKey` get used? Never seen it before

pradeepdewda-okta: Yeah that would be the part of other jira ticket i have.. It has been added in the endpoint, but not yet used in widget.

sherizada-okta: I think you're adding it to reduce mock drift, just curious if the UI uses it

sherizada-okta (approved):

PR #2241, opened 2021-10-07
Added support for IDP Discovery for OAuth2 Device Flow

larsjohansen-okta:
## Description:

- Added support for IDP Discovery for OAuth2 Device Flow

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Reviewers:

@nakuljoshi-okta @haishengwu-okta 

### Issue:

- [OKTA-434226](https://oktainc.atlassian.net/browse/OKTA-434226)

### DIFF
--- a/src/ForceIDPDiscoveryController.js
+++ b/src/ForceIDPDiscoveryController.js
@@ -1,0 +1,60 @@
+/*!
+ * Copyright (c) 2015-2017, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import Spinner from 'views/shared/Spinner';
+import BaseLoginController from 'util/BaseLoginController';
+import Util from 'util/Util';
+
+export default BaseLoginController.extend({
+  className: 'force-idp-discovery',
+
+  View: Spinner,
+
+  Model: {},
+
+  initialize: function() {
+    const OKTA_IDP_TYPE = 'OKTA';
+    const RESOURCE = 'okta:acct:';
+
+    const options = this.options;
+    const lastAuthResponse = options.appState.get('lastAuthResponse');
+    const stateToken = lastAuthResponse && lastAuthResponse?.stateToken;
+
+    const webfingerArgs = {
+      resource: RESOURCE,
+      requestContext: stateToken,
+    };
+
+    options.appState.settings.authClient
+      .webfinger(webfingerArgs)
+      .then(res => {
+        if (res?.links && res.links[0]) {
+          if (res.links[0].properties['okta:idp:type'] !== OKTA_IDP_TYPE && res.links[0].href) {
+            const redirectFn = res.links[0].href.includes('OKTA_INVALID_SESSION_REPOST%3Dtrue')
+              ? Util.redirectWithFormGet.bind(Util)
+              : this.settings.get('redirectUtilFn');
+            //override redirectFn to only use Util.redirectWithFormGet if OKTA_INVALID_SESSION_REPOST is included
+            //it will be encoded since it will be included in the encoded fromURI
+
+            redirectFn(res.links[0].href);
+            return;
+          }
+        }
+        options.appState.trigger('navigate', '');
+      })
+      .catch(() => {
+        // TODO: OKTA-436775 Show error instead of navigating to default route
+        options.appState.trigger('navigate', '');
+      });
+  },
+
+});
--- a/src/LoginRouter.js
+++ b/src/LoginRouter.js
@@ -42,6 +42,7 @@ import EnrollmentLinkSentController from 'EnrollmentLinkSentController';
 import EnterPasscodePushFlowController from 'EnterPasscodePushFlowController';
 import ForgotPasswordController from 'ForgotPasswordController';
 import IDPDiscoveryController from 'IDPDiscoveryController';
+import ForceIDPDiscoveryController from 'ForceIDPDiscoveryController';
 import ManualSetupPushController from 'ManualSetupPushController';
 import ManualSetupTotpController from 'ManualSetupTotpController';
 import MfaVerifyController from 'MfaVerifyController';
@@ -139,6 +140,7 @@ export default BaseLoginRouter.extend({
     'signin/enroll-user': 'enrollUser',
     'signin/device-activate-complete': 'deviceActivateComplete',
     'signin/device-activate': 'deviceActivate',
+    'signin/idp-discovery-check': 'idpDiscoveryCheck',
     '*wildcard': 'defaultAuth',
   },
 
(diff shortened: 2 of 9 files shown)


src/ForceIDPDiscoveryController.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default BaseLoginController.extend({
+  className: 'force-idp-discovery',
+
+  View: Spinner,
+
+  Model: {},
+
+  initialize: function() {
+    let options = this.options;
```
nakuljoshi-okta, line 25: can we make this `const`?

src/models/AppState.js
```diff
@@ -761,6 +761,15 @@ export default Model.extend({
         return res._embedded.deviceActivationStatus;
       },
     },
+    forceIdpDiscovery: {
+      deps: ['lastAuthResponse'],
+      fn: function(res) {
+        if (!res._embedded) {
+          return null;
```
nakuljoshi-okta, line 768: return `false`

src/ForceIDPDiscoveryController.js
```diff
@@ -0,0 +1,57 @@
...
+    const lastAuthResponse = options.appState.get('lastAuthResponse');
+    const stateToken = lastAuthResponse && lastAuthResponse?.stateToken;
+
+    const webfingerArgs = {
+      resource: 'okta:acct:undefined',
```
nakuljoshi-okta, line 30: nit:
just to confirm `undefined` means there is no user name right? is `resource` a required parameter? may be we just don't pass it? 
I mean I am okay with special case string, just that `undefined` sounds weird as an input since its special keyword in JS.

playground/mocks/data/api/v1/authn/forced-idp-discovery-linkedin-idp.json
```diff
@@ -0,0 +1,16 @@
+{
+  "subject": "okta:acct:undefined",
+  "links": [
+    {
+      "rel": "okta:idp",
+      "href": "http://localhost:3000/sso/idps/0oa4onxsxfUDwUb8u0g4?stateToken=00CazfnZW7EeVlIJoByKsUAdvMEcBvnN39We9ji5QL&login_hint=undefined#",
```
nakuljoshi-okta, line 6: we might wanna keep the `stateToken` value consistent across all mocks just so that its clear the same state token is getting passed in the flow.

src/models/AppState.js
pradeepdewda-okta: how about check for happy path first?
```
if (res._embedded) {
return res._embedded.forceIdpDiscovery;
}
return false;
```

src/ForceIDPDiscoveryController.js
```diff
@@ -0,0 +1,57 @@
...
+            //it will be encoded since it will be included in the encoded fromURI
+
+            redirectFn(res.links[0].href);
+          } else {
+            options.appState.trigger('navigate', '');
+          }
+        }
+      })
+      .catch((error) => {
+        console.log(error);
```
pradeepdewda-okta, line 53: Is this the error we want to show to the enduser ?

src/ForceIDPDiscoveryController.js
larsjohansen-okta: I use this because it's the same value as you'd submit if you didn't type in a username in the regular idp discovery flow (from the model):
```
    const username = this.settings.transformUsername(this.get('username'), Enums.IDP_DISCOVERY);
    const remember = this.get('remember');
    const lastUsername = this.get('lastUsername');
    const resource = 'okta:acct:' + username;
 ```

It is a required parameter on the backend, so we do need to send something.

After doing some investigation on the backend, it basically treats this as a user that has the username `undefined`. I think we should instead send a blank username (which is always invalid) as opposed to the `undefined` string, that way it won't match with any users which is what we want in this particular case.

File: src/ForceIDPDiscoveryController.js [hunk 6]
```diff
@@ -0,0 +1,57 @@
...
+  Model: {},
+
+  initialize: function() {
+    let options = this.options;
+    const lastAuthResponse = options.appState.get('lastAuthResponse');
+    const stateToken = lastAuthResponse && lastAuthResponse?.stateToken;
+
+    const webfingerArgs = {
+      resource: 'okta:acct:undefined',
+      requestContext: stateToken,
```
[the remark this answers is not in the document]
larsjohansen-okta: It cannot be undefined if we're ending up in this controller, it's guaranteed by the backend.

src/ForceIDPDiscoveryController.js
larsjohansen-okta: Whoopsie, this was leftover from testing manually. I believe in an error case we should reroute to the default route as well. Will fix.

src/models/AppState.js
larsjohansen-okta: Fixed.

src/ForceIDPDiscoveryController.js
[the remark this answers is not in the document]
larsjohansen-okta: yeah, this code originated from a IdpDiscovery model, in which case this might not be guaranteed, but in this controller it will be. I can remove the `?` part if you find that less confusing.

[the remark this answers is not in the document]
pradeepdewda-okta: should be okay.

src/LoginRouter.js
```diff
@@ -158,6 +160,10 @@ export default BaseLoginRouter.extend({
     'verifyPIV',
   ],
 
+  idpDiscoveryCheck: function() {
```
haishengwu-okta, line 163: nit:
move after `defaultAuth`

src/models/AppState.js
haishengwu-okta: nit:
```
return !! res._embedded && res._embedded.forceIdpDiscovery;
```

test/testcafe/framework/page-objects-v1/DeviceCodeActivatePageObject.js
```diff
@@ -48,6 +48,10 @@ export default class DeviceCodeActivatePageObject extends BasePageObject {
     return this.form.findFormFieldInput('username').visible;
   }
 
+  isPasswordFieldVisible() {
```
haishengwu-okta, line 51: looks like you need additional page object, e.g.
- `PrimaryAuthPageObject`
- `IdPDiscoveryPageObject`

playground/mocks/data/api/v1/authn/forced-idp-discovery-linkedin-idp.json
```diff
@@ -0,0 +1,16 @@
+{
+  "subject": "okta:acct:undefined",
```
nakuljoshi-okta, line 2: nit:
looks like this subject is not checked anywhere, but to keep it consistent change it to `okta:acct:`.

src/ForceIDPDiscoveryController.js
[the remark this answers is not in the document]
larsjohansen-okta: Fixed this, it's redirecting to the default route in case of any errors here.

playground/mocks/data/api/v1/authn/forced-idp-discovery-linkedin-idp.json
```diff
@@ -0,0 +1,16 @@
+{
```
haishengwu-okta, line 1: Just curious why those 2 responses structure are different (e.g. no `stateToken`, `_links` is array instead of object.

test/testcafe/framework/page-objects-v1/DeviceCodeActivatePageObject.js
haishengwu-okta: AFAIK, PageObject pattern is meant to create a Class per page.
Those method related to userName, password implies user already lands on different page (idp-discovery, primary auth etc) hence shall create separated page object.

src/ForceIDPDiscoveryController.js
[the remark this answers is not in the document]
haishengwu-okta: Just double check, will error message be propagated (i.e. show errors after redirect to default route) ?
Otherwise, user will be confused what's happening ?

PR #2248, opened 2021-10-12
add missing i18n for secondary email label

pradeepdewda-okta:
## Description:
add missing i18n for secondary email label

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-436730](https://oktainc.atlassian.net/browse/OKTA-436730)

### DIFF
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -825,7 +825,7 @@ const secondaryEmail = {
   '/idp/idx/identify': [
     'enroll-profile-update-params',
     'enroll-profile-update-all-optional-params',
-    'error-enroll-profile-update-params'
+    'enroll-profile-update-params'
   ],
   '/idp/idx/skip': [
     'success-with-app-user'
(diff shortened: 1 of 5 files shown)


playground/mocks/config/responseConfig.js
```diff
@@ -825,7 +825,7 @@ const secondaryEmail = {
   '/idp/idx/identify': [
     'enroll-profile-update-params',
     'enroll-profile-update-all-optional-params',
-    'error-enroll-profile-update-params'
+    'enroll-profile-update-params'
```
sherizada-okta, line 828: just fixing the name? or should a mock be added/removed?

jmelberg-okta: @pradeepdewda-okta - Given most of these form fields are dynamic, we should probably defer all translations to the API layer.

IMO, it would be a bit strange to only support translations for `secondaryEmail`, but none of the others.

PR #2252, opened 2021-10-13
feat: Shows all OAuth2 JIT failures

jackward-okta:
## Description:

As we expand into more IdPs, we're seeing a lot of variations in the properties IdPs support. This makes it more difficult for admins to set up their mappings and new user JITing. Making all the JIT errors visible in the sign-in widget will make these configuration issues more obvious.

We were already showing JIT errors in these cases:

- if you were using a redirect auth (click the IdP button, then we redirect you to the IdP) we showed all JIT errors
- if you were using a pop-up window auth (click the IdP button, then we open a pop-up with IdP login) we showed JIT errors for bad usernames, since they had the `access_denied` error code

This expands our error displaying in the pop-up case to include JIT errors for bad emails and other Okta required profile properties. In order to deliver a user-friendly experience, we replace the error messages from Okta with the generic message:
> Unable to sign in, please try another sign in method.

This is because users have no control over the IdP's configuration in Okta. The best remediation they can take is to try signing in another way.

The error we trigger using `Util.triggerAfterError` contains the original, descriptive error message in case any handlers require a more specific description.

## PR Checklist

- [x] Added unit tests?
- [ ] Added e2e tests

### Screenshot/Video:

image

### Reviewers:

- App Plat.
- Sign-in Widget reviewers

### Issue:

- [OKTA-433673](https://oktainc.atlassian.net/browse/OKTA-433673)


jessicaallen-okta:
hi @pradeepdewda-okta , good call on these error messages not being end-user friendly. Not blocking this fix, but in the future, we should have better patterns for what belongs in the end-user facing message and what belongs in a browser console log (or system log) for admin/dev troubleshooting. 

Generally, the end-user message should be straight forward _and_ offer a path to remediation. If there is no self-service remediation possible, it should direct them to their admin – for example:
"There was an error signing you in, please contact your system administrator to proceed" and the details of the JIT failure should go to the admin in a log.

colekurashige-okta:
@pradeepdewda-okta 
@jessicaallen-okta 

I'm taking over for Jack for this PR. I talked with @johnbaldo-okta and we decided that it would make more sense to modify the error messages on the signin widget side of things. So I've added `error.jit_failure` to the labels.

We look for the `jit_failure_*` error codes and transform the response displayed to the user into a user-friendly error message (`error.jit_failure`). We still preserve the original error message when triggering the error in code afterwards. So if I've done things right, we get a best of both worlds:
- Users don't receive confusing messages.
- Admins implementing the SIW get more descriptive error messages, if they want to handle them.

colekurashige-okta:
I'm not sure why the CI is failing. I just ran the `testcafe` suite locally and it seemed to work fine. Any ideas?

ethanhsiung-okta (approved):
LGTM!

jackward-okta (commented):
`:approved:` 💯

test/unit/spec/OAuth2Util_spec.js
```diff
@@ -58,6 +61,42 @@ describe('util/OAuth2Util', function() {
...
     });
 
+    it.each([
+      'access_denied',
+      'jit_failure_missing_fields',
+      'jit_failure_invalid_login_format',
+      'jit_failure_values_not_match_pattern',
+      'jit_failure_values_too_long',
+      'jit_failure_invalid_locale',
+    ])('shows SDK errors of certain types', function(errCode, done) {
```
jmelberg-okta, line 71: nit: To easily identify the error code tested, let's include it in the test description:

```suggestion
    ])('shows generic error for errorCode "%s"', function(errCode, done) {
```

test/unit/spec/OAuth2Util_spec.js
```diff
@@ -58,6 +61,42 @@ describe('util/OAuth2Util', function() {
...
+      const errorMessage = 'Auth SDK error';
+      const authException = new AuthSdkError(errorMessage);
+      authException.errorCode = errCode;
+
+      spyOn(authClient.token, 'getWithPopup').and.callFake(function() {
```
jmelberg-okta, line 76: nit: Not a blocker, but we should try to use [Jest Objects](https://jestjs.io/docs/jest-object) instead of Jasmine. The difference is subtle, but the method signatures are different:

```suggestion
      jest.spyOn(authClient.token, 'getWithPopup').mockImplementation(() => {
```

jmelberg-okta (approved):
Couple small nits - nothing blocking!

PR #2266, opened 2021-10-20
Okta 430932: missing password requirements

lesterchoi-okta:
## Description:

Need some feedback on this fix. AFAICT, there are 3 options to address the a11y issue outlined in <https://oktainc.atlassian.net/browse/OKTA-430932>. The satisfied/unsatisfied status of requirements should be audible for screen-reader users to fulfill (by adjusting their input).

* Option A: add `role="status"` or `role="alert"` to `<p>` tag
* Option B: add "sr-only" css class to prepended `<span>` with localized string indicating "unsatisfied" or similar.
* Option C: show password hints before input field (see https://incl.ca/show-hide-password-accessibility-and-password-hints-tutorial/)

Option A is best because it avoids the need for updated localizations and achieves the same effect as Option B but in a more semantic/standardized way, but I can't seem to produce the expected speech in NVDA on Windows.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-430932](https://oktainc.atlassian.net/browse/OKTA-430932)

### DIFF
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -842,5 +842,5 @@ const selfServiceRegistration = {
 };
 
 module.exports = {
-  mocks: selfServiceRegistration,
+  mocks: idx,
 };
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -835,6 +835,12 @@ const secondaryEmail = {
   ]
 };
 
+const selfServiceRegistration = {
+  '/api/v1/registration/form': [
+    'registration-form'
+  ]
+};
+
 module.exports = {
-  mocks: idx,
+  mocks: selfServiceRegistration,
 };
--- a/playground/mocks/spec-okta-api/api/v1/registration/form.js
+++ b/playground/mocks/spec-okta-api/api/v1/registration/form.js
@@ -1,0 +1,76 @@
+/* eslint-disable @okta/okta/no-unlocalized-text */
+const templateHelper = require('../../../../config/templateHelper');
+
+module.exports = [
+  templateHelper({
+    path: '/api/v1/registration/form',
+    method: 'GET',
+    template: JSON.stringify({
+      policyId: 'regyb5XILfdDeLzm10g3',
+      lastUpdate: 1634687026000,
+      profileSchema: {
+        properties: {
+          firstName: {
+            type: 'string',
+            title: 'First name',
+            minLength: 1,
+            maxLength: 50,
+            default: 'string'
+          },
+          lastName: {
+            type: 'string',
+            title: 'Last name',
+            minLength: 1,
+            maxLength: 50,
+            default: 'string'
+          },
+          password: {
+            type: 'string',
+            title: 'Password',
+            allOf: [
+              {
+                description: 'At least 8 character(s)',
+                minLength: 8
+              },
+              {
+                description: 'At least 1 number(s)',
+                format: '/[\\d]+/'
+              },
+              {
+                description: 'At least 1 lowercase letter(s)',
+                format: '/[a-z]+/'
+              },
+              {
+                description: 'At least 1 uppercase letter(s)',
+                format: '/[A-Z]+/'
+              },
+              {
+                description: 'Does not contain part of username',
+                format: '/^[#/userName]/'
+              }
+            ],
+            default: 'Password'
+          },
+          email: {
+            type: 'email',
+            title: 'Email',
+            format: 'email',
+            default: 'Email'
+          }
+        },
+        required: [
+          'email',
+          'password',
+          'firstName',
+          'lastName'
+        ],
+        fieldOrder: [
+          'email',
+          'password',
+          'firstName',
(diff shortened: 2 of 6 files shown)


playground/mocks/config/responseConfig.js
```diff
@@ -835,6 +835,12 @@ const secondaryEmail = {
...
 
+const selfServiceRegistration = {
+  '/api/v1/registration/form': [
+    'registration-form'
+  ]
+};
+
 module.exports = {
-  mocks: idx,
+  mocks: selfServiceRegistration,
```
shawyu-okta, line 845: May need to revert this change. I assume you edited for local testing but I think `idx` mocks have been the default scenario configured.

src/util/RegistrationFormFactory.js
```diff
@@ -113,16 +113,31 @@ const checkSubSchemas = function(fieldName, model, subSchemas, showError) {
...
       if (showError) {
         ele.find('p span').removeClass('confirm-16');
         ele.find('p span').addClass('error error-16-small');
         ele.addClass('subschema-error subschema-unsatisfied');
+
+        // set role="status" on <p>
+        ele.find('p').attr('role', 'status');
+        
+        // ele.find('p .sr-only').length === 0 && ele.find('p').prepend('<span class="sr-only" >Error</span>');
+        // ele.find('.sr-only').length === 0 && ele.prepend('<span class="sr-only">Error</span>');
```
shawyu-okta, line 140: Flagging for removal if the commented code is unused in the final PR

src/util/RegistrationFormFactory.js
```diff
@@ -113,16 +113,31 @@ const checkSubSchemas = function(fieldName, model, subSchemas, showError) {
     } else {
       ele.children('p').addClass('default-schema');
     }
+
+    // clear role="status" on <p>
+    ele.children('p').removeAttr('role');
+
+    // reset errors
```
shawyu-okta, line 120: 💯  for some comment annotations in this code

webpack.playground.config.js
```diff
@@ -63,7 +65,9 @@ module.exports = {
     ],
     historyApiFallback: true,
     headers: {
-      'Content-Security-Policy': `script-src http://localhost:${DEV_SERVER_PORT}`
+      'Content-Security-Policy': [
+        `script-src http://${HOST}:${DEV_SERVER_PORT}`
+      ].join(';')
```
shawyu-okta, line 70: Does changing this string construction make a difference or is it just for future devs who need to add more CSP items?

src/util/RegistrationFormFactory.js
```diff
@@ -113,16 +113,31 @@ const checkSubSchemas = function(fieldName, model, subSchemas, showError) {
     } else {
       ele.children('p').addClass('default-schema');
     }
+
+    // clear role="status" on <p>
+    ele.children('p').removeAttr('role');
```
shawyu-okta, line 118: Is there a way to remove the role in the DOM by default rather than having to do it in JS?

playground/mocks/config/responseConfig.js
lkchoi: Yeah that won't get merged. I usually leave a note to self DONKTMERGE. Just forgot this time

src/util/RegistrationFormFactory.js
lkchoi: Those are the other options. B and C left them for now to easily try them out. I'll remove the code we don't use

src/util/RegistrationFormFactory.js
```diff
@@ -113,16 +113,35 @@ const checkSubSchemas = function(fieldName, model, subSchemas, showError) {
     } else {
       ele.children('p').addClass('default-schema');
     }
+
+    // clear aria role and live-region for re-validation
+    ele.children('p')
+      .removeAttr('role')
+      .removeAttr('aria-live');
```
shawyu-okta, line 120: Possibly worth seeing if this can live inside the `// validate` block but I think it's fine here too.

lesterchoi-okta: true. i think it's safer outside so we know it's back to a "default" state before applying the role and aria-live attributes again in the failed condition

playground/mocks/spec-okta-api/api/v1/registration/form.js
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable @okta/okta/no-unlocalized-text */
+const templateHelper = require('../../../../config/templateHelper');
+
+module.exports = [
+  templateHelper({
+    path: '/api/v1/registration/form',
+    method: 'GET',
+    template: JSON.stringify({
```
haishengwu-okta, line 8: Hi @lesterchoi-okta , I assume specify `template` isn't required (or actually be abundant) ?

lesterchoi-okta: @haishengwu-okta yes, think you're right. i'm going to add a cleanup task and follow up right after release

PR #2235, opened 2021-10-05
Email Magic Link: Addition of Terminal View for OTP information

justinabrokwah-okta:
## Description:
- Creating a new view for the email magic link flow when the link is clicked in a different browser, should display information about the login
#### Figma mockup (the goal): 
image

###### Oct. 21, 2021 @ 2:33PM
image

- Sample Response:
`{
   "version":"1.0.0",
   "messages":{
      "type":"array",
      "value":[
         {
            "message":"Enter the OTP where you where authenticating.",
            "i18n":{
               "key":"idx.enter.otp.in.original.tab"
            },
            "class":"INFO"
         }
      ]
   },
   "currentAuthenticator":{
      "type":"object",
      "value":{
         "contextualData":{
            "otp":"702255"
         },
         "type":"email",
         "key":"okta_email",
         "id":"aut5gn8p1c0jGB5nO0g4",
         "displayName":"Email",
         "methods":[
            {
               "type":"email"
            }
         ]
      }
   },
   "client":{
      "type":"object",
      "value":{
         "broswer":"FIREFOX",
         "os":"Mac OS X",
         "geolocation":"Invalid Location"
      }
   },
   "app":{
      "type":"object",
      "value":{
         "name":"oidc_client",
         "label":"my 3rd magic link spa",
         "id":"0oa5u4b54akg0cVYw0g4"
      }
   }
}`

## PR Checklist

- [x] Have you verified the basic functionality for this change?

- [ ] Added unit tests?

- [ ] Added e2e tests

- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?

### Screenshot/Video:
- Demo Video (w/out geolocation): https://okta.box.com/s/tuzg0kokd95jdz0c7yeddco0io6ygw4e
- Demo Video (w/ geolocation, set locally in backend): https://okta.box.com/s/vtpe26zby3fb76kt990t0wo9xt4vbmc9
- Demo Video (w/ permutations of geolocation input {city, state, country}, done through backend): https://okta.box.com/s/ovkwkfxj5h3acwywm22mxzwgejlqwf0k

### Issue:

- [OKTA-433992](https://oktainc.atlassian.net/browse/OKTA-433992)

PR #2281, opened 2021-10-29
Refactor showMessages in BaseForm

pradeepdewda-okta:
## Description:

Refactor BaseForm's implementation of showMessages.

1. remove redundant code from multiple views
2. enhance showMessages to accept options or callout view itself
3. override showMessages in views where necessary

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-441843](https://oktainc.atlassian.net/browse/OKTA-441843)

### DIFF
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -15,6 +15,8 @@ const idx = {
 
   '/idp/idx/introspect': [
     'identify',
+    // 'authenticator-enroll-ov-qr-enable-biometrics',
+    // 'authenticator-verification-okta-verify-push',
     // 'error-401-invalid-otp-passcode',
     // 'error-with-failure-redirect',
     // 'error-feature-not-enabled',
@@ -145,6 +147,9 @@ const idx = {
     // 'authenticator-verification-okta-verify-push',
     // 'authenticator-verification-custom-app-push',
     // 'authenticator-verification-custom-app-push-reject',
+    // 'authenticator-enroll-ov-sms-enable-biometrics',
+    // 'okta-verify-version-upgrade',
+    // 'okta-verify-uv-verify-enable-biometrics'
   ],
   '/idp/idx/challenge': [
     // 'authenticator-verification-webauthn',
@@ -496,6 +501,15 @@ const ovTotpError = {
   '/idp/idx/challenge/answer': [
     'error-okta-verify-totp',
     // 'error-okta-verify-uv-totp-verify-enable-biometrics',
+  ],
+};
+
+const totpEnableBiometrics = {
+  '/idp/idx/introspect': [
+    'authenticator-verification-okta-verify-totp'
+  ],
+  '/idp/idx/challenge/answer': [
+    'error-okta-verify-uv-totp-verify-enable-biometrics',
   ],
 };
 
--- a/src/v2/controllers/FormController.js
+++ b/src/v2/controllers/FormController.js
@@ -247,8 +247,18 @@ export default Controller.extend({
     return modelJSON;
   },
 
+  /**
+   * @param model current form model
+   * @param error any errors after user action
+   * @param form current form
+   * Handle errors that get displayed right after any user action. After such form errors widget doesn't
+   * reload or re-render, but updates the AppSate with latest remediation.
+   */
   showFormErrors(model, error, form) {
-    let errorObj, idxStateError;
+    /* eslint max-statements: [2, 22] */
+    let errorObj;
+    let idxStateError;
+    let showErrorBanner = true;
     model.trigger('clearFormError');
     
     if (!error) {
@@ -269,14 +279,14 @@ export default Controller.extend({
       Util.logConsoleError(error);
       errorObj = { responseJSON: { errorSummary: loc('error.unsupported.response', 'login')}};
     }
-    const showErrorBanner = !form?.isErrorCalloutCustomized(errorObj);
+
+    if(_.isFunction(form?.showCustomFormErrorCallout)) {
+      showErrorBanner = !form.showCustomFormErrorCallout(errorObj);
+    }
+
     // show error before updating app state.
     model.trigger('error', model, errorObj, showErrorBanner);
     idxStateError = Object.assign({}, idxStateError, {hasFormError: true});
-
-    if (!showErrorBanner) {
-      form.showCustomErrorCallout(errorObj);
-    }
 
     // TODO OKTA-408410: Widget should update the state on every new response. It should NOT do selective update.
     // For eg 429 rate-limit errors, we have to skip updating idx state, because error response is not an idx response.
--- a/src/v2/view-builder/internals/BaseForm.js
+++ b/src/v2/view-builder/internals/BaseForm.js
@@ -1,4 +1,4 @@
-import { _, Form, loc, internal } from 'okta';
+import { _, Form, loc, internal, createCallout, View } from 'okta';
 import FormInputFactory from './FormInputFactory';
 
 const { FormUtil } = internal.views.forms.helpers;
@@ -113,37 +113,39 @@ export default Form.extend({
   },
 
   /*
-   * Whether an form error message should be rendered as the default BaseForm error banner or not
-   * if this function returns true, BaseForm won't render the error banner after form submit. Instead,
-   * the form needs to override showCustomErrorCallout with the logic to render that error message
-   */
-  isErrorCalloutCustomized() {
-    return false;
-  },
+  * Views should override this function to render custom error callouts for invalid form actions.
+  * Should return true when callout is customized
+  */
+  showCustomFormErrorCallout: null,
 
   /*
-   * Renders an error message (if any) after form is submitted
-   * This should be overriden if isErrorCalloutCustomized() returns true
-   */
-  showCustomErrorCallout() {
-    this.$('.o-form-error-container').addClass('o-form-has-errors');
-  },
-
-  /*
-   * Renders the contents of messages object (if any) during initialize
+   * Renders the contents of messages object (if any, on error) during initialize
    * This function is called during Form.initialize, and will display
-   * messages when the form renders.
-   * Note: any errors happening after the form is rendered won't use this function,
-   * this is only used during initialize
+   * messages when the form reloads.
+   * Note: Any user action related form errors handled by FormController.showFormErrors
    */
-  showMessages() {
-    // render messages as text
-    const messagesObjs = this.options.appState.get('messages');
-    if (messagesObjs?.value.length) {
-      const content = messagesObjs.value.map((messagesObj) => {
-        return messagesObj.message;
+  showMessages(options) {
+    const messages = this.options.appState.get('messages') || {};
+    const errContainer = '.o-form-error-container';
+    if (Array.isArray(messages.value) && !(options instanceof View)) {
+      this.add('<div class="ion-messages-container"></div>', errContainer);
+      messages.value.forEach(obj => {
+        if(!obj?.class || obj.class === 'INFO') {
+          // add message as plain text
+          this.add(`<p>${obj.message}</p>`, '.ion-messages-container');
+        } else {
+          // add error callout with custom options
+          options = Object.assign(_.pick(obj, 'message', 'class'), options);
+          this.add(createCallout({
+            content: options.message,
+            type: (options.class || '').toLowerCase(),
+            title: options.title ? options.title : ''
+          }), errContainer);
+        }
       });
-      this.add(`<div class="ion-messages-container">${content.join(' ')}</div>`, '.o-form-error-container');
+    } else if (options instanceof View) {
+      // if callee is showCustomFormErrorCallout. show custom error views
+      this.add(options, errContainer);
     }
   },
 });
--- a/src/v2/view-builder/views/IdentifierView.js
+++ b/src/v2/view-builder/views/IdentifierView.js
@@ -1,4 +1,4 @@
-import { loc, createCallout } from 'okta';
+import { loc } from 'okta';
 import { FORMS as RemediationForms } from '../../ion/RemediationConstants';
 import { BaseForm, BaseView, createIdpButtons, createCustomButtons } from '../internals';
 import DeviceFingerprinting from '../utils/DeviceFingerprinting';
@@ -94,18 +94,6 @@ const Body = BaseForm.extend({
     }
   },
 
-  showMessages() {
-    const messagesObj = this.options.appState.get('messages');
-    if (messagesObj?.value.length
-        && this.options.appState.get('currentFormName') === 'identify') {
-      const displayMessageObj = messagesObj.value[0];
-      const messageCallout = createCallout({
-        content: displayMessageObj.message,
-        type: (displayMessageObj.class || '').toLowerCase(),
-      });
-      this.introspectMessage = this.add(messageCallout, '.o-form-error-container').last();
-    }
-  },
   /**
    * Update UI schemas for customization from .widgetrc.js or Admin Customization settings page.
    * @returns Array
--- a/src/v2/view-builder/views/TerminalView.js
+++ b/src/v2/view-builder/views/TerminalView.js
@@ -1,4 +1,4 @@
-import { createCallout, loc } from 'okta';
+import { loc } from 'okta';
 import { BaseForm, BaseFooter, BaseView } from '../internals';
 import { getBackToSignInLink, getSkipSetupLink, getReloadPageButtonLink } from '../utils/LinksUtil';
 import EmailAuthenticatorHeader from '../components/EmailAuthenticatorHeader';
@@ -121,31 +121,16 @@ const Body = BaseForm.extend({
       messagesObjs.value.push({ message: loc('oie.return.to.original.tab', 'login')});
     } else if (this.options.appState.containsMessageWithI18nKey('tooManyRequests')) {
       description = loc('oie.tooManyRequests', 'login');
+    } else if (this.options.appState.containsMessageWithI18nKey(RETURN_LINK_EXPIRED_KEY)) {
+      messagesObjs.value[0].class = 'ERROR';
     }
 
     if (description && Array.isArray(messagesObjs?.value)) {
       messagesObjs.value[0].message = description;
     }
 
-    if (messagesObjs && Array.isArray(messagesObjs.value)) {
-      this.add('<div class="ion-messages-container"></div>', '.o-form-error-container');
-
-      messagesObjs.value
-        .forEach(messagesObj => {
-          const msg = messagesObj.message;
-          if (messagesObj.class === 'ERROR' || messagesObj.i18n?.key === RETURN_LINK_EXPIRED_KEY) {
-            this.add(createCallout({
-              content: msg,
-              type: 'error',
-            }), {
-              selector: '.o-form-error-container',
-              prepend: true,
-            });
-          } else {
-            this.add(`<p>${msg}</p>`, '.ion-messages-container');
-          }
-        });
-    }
+    this.options.appState.set('messages', messagesObjs);
+    BaseForm.prototype.showMessages.call(this);
   },
 
 });
--- a/src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
+++ b/src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
@@ -18,41 +18,25 @@ const Body = BaseForm.extend(Object.assign(
     },
 
     showMessages() {
-      // override showMessages to display error in cases like reject push
-      // or timeout. Borrowed this logic from TerminalView.
-      const messagesObjs = this.options.appState.get('messages');
-      if (messagesObjs && Array.isArray(messagesObjs.value)) {
-        this.add('<div class="ion-messages-container"></div>', '.o-form-error-container');
-
-        messagesObjs.value.forEach(messagesObj => {
-          const msg = messagesObj.message;
-          if (messagesObj?.class === 'ERROR') {
-            const options = {
-              content: msg,
-              type: 'error',
-            };
-            if (this.options.appState.containsMessageWithI18nKey(OV_NMC_FORCE_UPGRAGE_SERVER_KEY)) {
-              // add a title for OV force upgrade
-              options.title = loc('oie.numberchallenge.force.upgrade.title', 'login');
-            } else if (this.options.appState.containsMessageWithI18nKey(OV_UV_ENABLE_BIOMETRIC_SERVER_KEY)) {
-              // add a title for OV enable biometrics message during verification
-              options.content = null;
-              options.className = 'okta-verify-uv-callout-content';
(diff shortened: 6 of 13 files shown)


playground/mocks/config/responseConfig.js
```diff
@@ -145,6 +147,9 @@ const idx = {
     // 'authenticator-verification-okta-verify-push',
     // 'authenticator-verification-custom-app-push',
     // 'authenticator-verification-custom-app-push-reject',
+    // 'authenticator-enroll-ov-sms-enable-biometrics',
+    // 'okta-verify-version-upgrade',
+    'okta-verify-uv-verify-enable-biometrics'
```
sherizada-okta, line 152: This should be the new default?

src/v2/view-builder/internals/BaseForm.js
```diff
@@ -130,20 +130,30 @@ export default Form.extend({
...
+    const messages = this.options.appState.get('messages') || {};
+    if (Array.isArray(messages.value)) {
+      messages.value.forEach(obj => {
+        if(!obj?.class || obj.class === 'INFO') {
+          this.add(`<div class="ion-messages-container">${obj.message}</div>`, '.o-form-error-container');
+        } else {
+          if(options instanceof View) {
+            this.add(options, '.o-form-error-container');
+          } else {
+            obj = Object.assign(obj, options);
```
sherizada-okta, line 148: Why do you want to add options to the obj?

src/v2/view-builder/internals/BaseForm.js
```diff
@@ -130,20 +130,30 @@ export default Form.extend({
...
-    const messagesObjs = this.options.appState.get('messages');
-    if (messagesObjs?.value.length) {
-      const content = messagesObjs.value.map((messagesObj) => {
-        return messagesObj.message;
+  showMessages(options) {
+    const messages = this.options.appState.get('messages') || {};
+    if (Array.isArray(messages.value)) {
+      messages.value.forEach(obj => {
+        if(!obj?.class || obj.class === 'INFO') {
+          this.add(`<div class="ion-messages-container">${obj.message}</div>`, '.o-form-error-container');
```
sherizada-okta, line 143: The original code wrapped the message in a p tag

playground/mocks/config/responseConfig.js
pradeepdewda-okta: there such no defaults for the mocks.. but i will change.

src/v2/view-builder/internals/BaseForm.js
pradeepdewda-okta: consuming views can add callout options like `title`.. few views did that.

src/v2/view-builder/internals/BaseForm.js
pradeepdewda-okta: does it affect anything ?

src/v2/view-builder/internals/BaseForm.js
pradeepdewda-okta: i will rather do this.. thanks for pointing that out.
```
            options = Object.assign(options, _.pick(obj, 'message', 'class'));
```

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
```diff
@@ -18,41 +18,25 @@ const Body = BaseForm.extend(Object.assign(
...
+        options.className = 'okta-verify-uv-callout-content';
+        options.title = loc('oie.authenticator.app.method.push.verify.enable.biometrics.title', 'login');
+        options.subtitle = loc('oie.authenticator.app.method.push.verify.enable.biometrics.description', 'login');
+        options.type = 'error';
+        options.bullets = [
+          loc('oie.authenticator.app.method.push.verify.enable.biometrics.point1', 'login'),
+          loc('oie.authenticator.app.method.push.verify.enable.biometrics.point2', 'login'),
+          loc('oie.authenticator.app.method.push.verify.enable.biometrics.point3', 'login')
+        ];
+        options = createCallout(options);
```
haishengwu-okta, line 37: Why do you need to create `createCallout` instead of just pass `options`?

src/v2/view-builder/internals/BaseForm.js
```diff
@@ -113,37 +113,53 @@ export default Form.extend({
...
   },
 
   /*
-   * Renders an error message (if any) after form is submitted
-   * This should be overriden if isErrorCalloutCustomized() returns true
+   * Views should override this function to render custom error callouts after any user actions.
+   * This should be overriden if isFormErrorCalloutCustomized() returns true
    */
-  showCustomErrorCallout() {
-    this.$('.o-form-error-container').addClass('o-form-has-errors');
```
haishengwu-okta, line 129: - Remind me why this line can be removed?
- Sound wired if overriding this method depends on value of another method? Do you think we can combine both?

    - set `showCustomErrorCallout` to be `_.identity` or null`

    - In `FormController.showFormErrors`

```
if isFunction(showCustomFormErrorCallout) then 
  form.showCustomFormErrorCallout();
```

haishengwu-okta (commented):
Thank you so much for working on this debt!!

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
pradeepdewda-okta: Gives ability to pass any view not just callout.. baseForm callout has only bare minimum properties.. content, type and title

src/v2/view-builder/internals/BaseForm.js
pradeepdewda-okta: you think we can get rid of `isFormErrorCalloutCustomized` ? and pass the logic to identify which message is customize and which are not in `showCustomFormErrorCallout` ?.. I think it make sense if thats what you were suggesting..

haishengwu-okta: yes , get rid of `isFormErrorCalloutCustomized` which equaling to the exists of function `showCustomFormErrorCallout`.

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
haishengwu-okta: I meant 
instead of doing this 
```
   if(options instanceof View) {
             // add custom callout or any view
             this.add(options, errContainer);
           } else {
             // add error callout with custom options 
             options = Object.assign(_.pick(obj, 'message', 'class'), options);
             this.add(createCallout(...))
           }
```

can we do

```
const callout = createCallout(createCallout(_.defaults(options), {title:'', type:option.class})
this.add(callout)
```

The reason I'm asking is because looks like we didn't have use case where child class want to pass any view.
If you're thinking `in future, maybe child class want to pass any view`, I'd suggest to leave it for future given it's small refactoring.

pradeepdewda-okta: @haishengwu-okta Yes, I wanted to provide an ability to render any view as terminal state.. We currently have couple of flows that implements their own `templates` to render terminal views that doesn't have any callouts.. 

for eg take a look at this.. 
https://github.com/okta/okta-signin-widget/pull/2235#pullrequestreview-792043622

src/v2/controllers/FormController.js
```diff
@@ -276,17 +277,16 @@ export default Controller.extend({
...
-    // custom form error callout
-    if (!showErrorBanner) {
+    if(_.isFunction(form?.showCustomFormErrorCallout)) {
       form.showCustomFormErrorCallout(errorObj);
+      this.showErrorBanner = !!form.showErrorBanner;
     }
 
+    // show error before updating app state.
+    model.trigger('error', model, errorObj, this.showErrorBanner);
+    idxStateError = Object.assign({}, idxStateError, {hasFormError: true});
```
haishengwu-okta: Would this work without using instance level properties?

```
let showErrorBanner = false;

if(_.isFunction(form?.showCustomFormErrorCallout)) {
        form.showCustomFormErrorCallout(errorObj);
        showErrorBanner = true;
}

model.trigger('error', model, errorObj, showErrorBanner);
```

The other way we could think of is set `showCustomFormErrorCallout = null`  by default, then

```
let showErrorBanner = _.isFunction(form?.showCustomFormErrorCallout);

if(showErrorBanner) {
        form.showCustomFormErrorCallout(errorObj);
}
```

src/v2/controllers/FormController.js
```diff
@@ -269,15 +277,16 @@ export default Controller.extend({
...
-    const showErrorBanner = !form?.isErrorCalloutCustomized(errorObj);
-    // show error before updating app state.
-    model.trigger('error', model, errorObj, showErrorBanner);
-    idxStateError = Object.assign({}, idxStateError, {hasFormError: true});
 
-    if (!showErrorBanner) {
-      form.showCustomErrorCallout(errorObj);
+    if(_.isFunction(form?.showCustomFormErrorCallout)) {
+      form.showCustomFormErrorCallout(errorObj);
+      this.showErrorBanner = !!form.showErrorBanner;
```
mauriciocastillosilva-okta, line 283: shouldn't this be `false` if we just rendered a custom form error in the line above? Otherwise we will end up with more error callouts than what we need

src/v2/view-builder/internals/BaseForm.js
```diff
@@ -111,39 +111,45 @@ export default Form.extend({
...
-    if (messagesObjs?.value.length) {
-      const content = messagesObjs.value.map((messagesObj) => {
-        return messagesObj.message;
+  showMessages(options) {
+    const messages = this.options.appState.get('messages') || {};
+    const errContainer = '.o-form-error-container';
+    if (Array.isArray(messages.value)) {
+      this.add('<div class="ion-messages-container"></div>', errContainer);
+      messages.value.forEach(obj => {
+        if(!obj?.class || obj.class === 'INFO') {
```
mauriciocastillosilva-okta, line 132: why are we including `info` as well here? We have a callout type `info` that we can use if we now have this `class` attribute in the message. Your changes below already will take care of that

src/v2/view-builder/internals/BaseForm.js
```diff
@@ -111,39 +111,45 @@ export default Form.extend({
...
+          // add message as text
+          this.add(`<p>${obj.message}</p>`, '.ion-messages-container');
```
mauriciocastillosilva-okta, line 134: before we were concatenating with spaces, now we have a paragraph for each error. Can you share a screenshot of before and after? I am curious to see which errors message we had here, and why spaces was working

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -94,18 +94,6 @@ const Body = BaseForm.extend({
     }
   },
 
-  showMessages() {
-    const messagesObj = this.options.appState.get('messages');
-    if (messagesObj?.value.length
-        && this.options.appState.get('currentFormName') === 'identify') {
-      const displayMessageObj = messagesObj.value[0];
-      const messageCallout = createCallout({
```
mauriciocastillosilva-okta, line 102: before for this view we were rendering a callout all the time, with the new changes it will either render a callout or just a `<p>`, depending if the `messagesObj.value[0].class` is defined. This is why I was asking in the BaseForm for some screenshots, to better see what is changing

src/v2/view-builder/internals/BaseForm.js
```diff
@@ -111,39 +111,45 @@ export default Form.extend({
...
+        } else {
+          if(options instanceof View) {
+            // add custom callout or any view
+            this.add(options, errContainer);
```
mauriciocastillosilva-okta, line 138: this means that for each of the messages items we will add the same `options` view? If the `options` argument is a View, maybe we don't need to render one per `messages` items? 

If I am following this right, let's say `options` is a `View`, and `messages` have 2 "ERROR" items, we will not render both errors, but instead we will render the `View` twice...

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
```diff
@@ -18,41 +18,25 @@ const Body = BaseForm.extend(Object.assign(
...
       }
+      BaseForm.prototype.showMessages.call(this, options);
```
mauriciocastillosilva-okta, line 39: Before, for each of the message we were evaluating `containsMessageWithI18nKey` and deciding what the content of the callout was
With this change we are passing `options` based on whatever the first message is. I don't think we will ever get the a messages object with these 2 errors together, but still this is prone to not render the right message. I made a comment in BaseForm that indicates where this is rendering the same view for all messages

src/v2/view-builder/views/ov/EnrollPollOktaVerifyView.js
```diff
@@ -45,33 +45,17 @@ const Body = BaseForm.extend(Object.assign(
...
+      const calloutOptions = {};
+      if (this.options.appState.containsMessageWithI18nKey(OV_FORCE_FIPS_COMPLIANCE_UPGRAGE_KEY_IOS) ||
+        this.options.appState.containsMessageWithI18nKey(OV_FORCE_FIPS_COMPLIANCE_UPGRAGE_KEY_NON_IOS)) {
+        // add a title for ov force upgrade
+        calloutOptions.title = loc('oie.okta_verify.enroll.force.upgrade.title', 'login');
+      } else if (this.options.appState.containsMessageWithI18nKey(OV_QR_ENROLL_ENABLE_BIOMETRICS_KEY)) {
+        // add a title for OV enable biometrics message during enrollment
+        calloutOptions.title = loc('oie.authenticator.app.method.push.enroll.enable.biometrics.title', 'login');
       }
+      BaseForm.prototype.showMessages.call(this, calloutOptions);
```
mauriciocastillosilva-okta, line 58: similar comment here. I don't think we have a flow where both errors above (`OV_FORCE_FIPS_COMPLIANCE_UPGRAGE_KEY_IOS` and `OV_QR_ENROLL_ENABLE_BIOMETRICS_KEY`) come at the same time, but if that happens we will end up sending to showMessage the incorrect callout options

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
```diff
@@ -16,29 +16,27 @@ const Body = BaseForm.extend(Object.assign(
...
+          className: 'okta-verify-uv-callout-content',
+          title: loc('oie.authenticator.app.method.push.verify.enable.biometrics.title', 'login'),
+          subtitle: loc('oie.authenticator.app.method.push.verify.enable.biometrics.description', 'login'),
+          bullets: [
+            loc('oie.authenticator.app.method.push.verify.enable.biometrics.point1', 'login'),
+            loc('oie.authenticator.app.method.push.verify.enable.biometrics.point2', 'login'),
+            loc('oie.authenticator.app.method.push.verify.enable.biometrics.point3', 'login')
+          ],
+        };
+        BaseForm.prototype.showMessages.call(this, createCallout(options));
```
mauriciocastillosilva-okta, line 34: Does this mean that whoever overrides `showCustomFormErrorCallout` has to explicitly call `BaseForm.prototype.showMessages`?

src/v2/controllers/FormController.js
pradeepdewda-okta: Done

src/v2/controllers/FormController.js
pradeepdewda-okta: I implemented differently in latest commit.. but `!!` returns false for `false` and `undefined`

src/v2/view-builder/internals/BaseForm.js
pradeepdewda-okta: anything that is not `ERROR` or `WARN` show up as plain text messages.. We don't show `INFO` callouts in widget.. but we do have messages tagged as `INFO`..

src/v2/view-builder/internals/BaseForm.js
pradeepdewda-okta: I don't see any existing error messages that has multiple `errors` so i guess it was just a fall back that never hit..

src/v2/view-builder/internals/BaseForm.js
pradeepdewda-okta: The idea in widget is to show just one error if there are multiple error in the same IDX response.. Once the user resolve that, then show the next error if it still there.. According to UX we should not overwhelm users with many errors at a time..

But overtime, this `showMessages` got some hacky implementation in `TerminalView` ([here](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/TerminalView.js#L118)) which adds more messages to `message` object.. to renders them as just plain text.. That is why i have to keep `forEach`..  But ideally it will always be the `top` error..


>but instead we will render the View twice...
There is a rare possibility.. But again there should be just one error message in the IDX api response. Let me know if you have example of IDX API response that send `two` error messages.

pradeepdewda-okta: >but instead we will render the View twice...

Actually i removed it that. Now `else if` will take care of add view just once.

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
pradeepdewda-okta: replied here https://github.com/okta/okta-signin-widget/pull/2281#discussion_r741309579

src/v2/view-builder/views/ov/EnrollPollOktaVerifyView.js
pradeepdewda-okta: replied here https://github.com/okta/okta-signin-widget/pull/2281#discussion_r741309579

src/v2/view-builder/views/ov/ChallengeOktaVerifyTotpView.js
pradeepdewda-okta: Fixed it..

src/v2/view-builder/views/IdentifierView.js
pradeepdewda-okta: I have to take a leap of faith here BUT it's a part of fix.. 

All the messages tagged as `ERROR` or `WARN` should have a callout.. and the messages that are not tagged or have a class `INFO` then it should be shown as `plain text message`... 

If an IDX api not adding a `class` that it expect widget to render as error.. then we should fix the API.

I sifted through `testcafes` i don't see any example for `IdentifierView` that was showing `error callout` for messages when `class` was undefined

src/v2/controllers/FormController.js
```diff
@@ -247,8 +247,15 @@ export default Controller.extend({
...
+  /**
+   * @param model current form model
+   * @param error any errors after user action
+   * @param form current form
+   * Handle errors that get displayed right after any user action. After such form errors widget doesn't
+   * reload or re-render, but updates the AppSate with latest remediation.
+   */
   showFormErrors(model, error, form) {
-    let errorObj, idxStateError;
+    let errorObj, idxStateError, showErrorBanner = true;
```
haishengwu-okta, line 258: nit:
I always encourage to use one variable declaration per line. :)
More on personal preference - easier to read.

haishengwu-okta (approved):

src/v2/controllers/FormController.js
pradeepdewda-okta: Done.

src/v2/view-builder/views/IdentifierView.js
mauriciocastillosilva-okta: Ok, we can take the leap of faith given you said the API messages must come with a `class` value

src/v2/view-builder/views/TerminalView.js
```diff
@@ -121,31 +121,16 @@ const Body = BaseForm.extend({
       messagesObjs.value.push({ message: loc('oie.return.to.original.tab', 'login')});
     } else if (this.options.appState.containsMessageWithI18nKey('tooManyRequests')) {
       description = loc('oie.tooManyRequests', 'login');
+    } else if (this.options.appState.containsMessageWithI18nKey(RETURN_LINK_EXPIRED_KEY)) {
+      messagesObjs.value[0].class = 'ERROR';
```
mauriciocastillosilva-okta, line 125: how do you know it is the first message? Before we were adding an error callout if the message had `RETURN_LINK_EXPIRED_KEY` as the key, but we would show other messages if they were available.

I think we need to know the index of this error and change that specific one. Most of our errors today are an array of 1 element, but I don't think that is part of the contract, so we shouldn't assume it. At the end, if `messages` is an array, it can contain any number of items

src/v2/view-builder/internals/BaseForm.js
mauriciocastillosilva-okta: I don't have an example unfortunately on the top of my head, but if `messages` is an array, we should not assume it is always length=1, but instead we should iterate through all the items and display them... 
With that said I agree with not showing 2 callouts, that would look bad, but we should not ignore the second error if it comes. Perhaps we can show them as `bullets` of the same callout? Can be a refactor in a different PR if that makes more sense

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
```diff
@@ -18,41 +18,25 @@ const Body = BaseForm.extend(Object.assign(
     },
 
     showMessages() {
```
mauriciocastillosilva-okta, line 20: should this be `showCustomFormErrorCallout` instead?

src/v2/view-builder/views/ov/EnrollPollOktaVerifyView.js
```diff
@@ -45,33 +45,17 @@ const Body = BaseForm.extend(Object.assign(
       }
     },
     showMessages() {
```
mauriciocastillosilva-okta, line 47: same question here, this seems to be a good use case for `showCustomFormErrorCallout` instead

src/v2/view-builder/views/ov/ChallengeOktaVerifyResendPushView.js
pradeepdewda-okta: Actually `showCustomFormErrorCallout` is for errors after any user action (on form submit) when widget doesn't re-render and hence couldn't call `showMessages` (because it get called only on form.initialize). 

This is why i think you introduced this new function that adds ability to customize default courage form error callout..

src/v2/view-builder/views/TerminalView.js
pradeepdewda-okta: I would create a backend jira for this.. The flow expects this to be `ERROR` but api is returning this error as `INFO`.. This needs to be fixed on backend side, so that i can remove this one off hack from widget.. 

https://github.com/okta/okta-core/blob/eed2411d033788729198661374245ae3d9562728/components/platformx/idx/api/src/main/java/com/okta/api/mediation/services/auth/TerminalStateTokenHelper.java#L121

https://oktainc.atlassian.net/browse/OKTA-443718

PR #2307, opened 2021-11-18
add title when device/account invalidated okta-443983

anatoliiboiarshchykov-okta:
## Description:
- add a title when device/account invalidated
- add test cafe case

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/80226149/142483762-a17512df-5acc-4753-984c-a6f6d25a43e2.mov

### Reviewers:
@haishengwu-okta 
@pradeepdewda-okta 
@okta/device-platform 

### Issue:
- [OKTA-443983](https://oktainc.atlassian.net/browse/OKTA-443983)


src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -145,6 +147,22 @@ const Body = BaseFormWithPolling.extend(
...
+      const errorSummary = error.responseJSON.errorSummary;
+      let options = {
+        type: 'error',
+        className: 'okta-verify-uv-callout-content',
+        subtitle: errorSummary,
+      };
+
+      if (errorSummary.includes(INVALIDATED_ACCOUNT_DEVICE_ERR_MSG)) {
+        options.title = loc('user.fail.verifyIdentity', 'login');
+      }
```
jmelberg-okta, line 160: It is very unlikely this will always be an English error message. Can you please rely on an `errorCode` instead?

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -9,6 +9,8 @@ import { doChallenge } from '../../utils/ChallengeViewUtil';
 import OktaVerifyAuthenticatorHeader from '../../components/OktaVerifyAuthenticatorHeader';
 import { getSignOutLink } from '../../utils/LinksUtil';
 
+const INVALIDATED_ACCOUNT_DEVICE_ERR_MSG = 'Your device or account was invalidated';
```
pradeepdewda-okta, line 12: In the idx error response there must be a `i18nKey` that should exists.. that can be used to identify the error instead of using error message itself.

src/v2/view-builder/views/device/DeviceChallengePollView.js
pradeepdewda-okta: +1

anatoliiboiarshchykov-okta: Updated

src/v2/view-builder/views/device/DeviceChallengePollView.js
anatoliiboiarshchykov-okta: Updated

anatoliiboiarshchykov-okta:
Regarding @jhe-okta questions
**Is the change only for identifier flow? Do we need to apply it to the MFA flow in ChallengeOktaVerifyFastPassView.js?** - not sure, this fix is only to specific case from the ticket description. But I think it can be easily added if needed
**What @jmelberg-okta says about the error message. We should have a more generic way to identify the error.** - updated
**Is this a bug fix? Do you need it to go in a weekly release or merge to master for monthly release?** Since we are late for 19th November SIW CF, I think we can go any way. But we need to be sure changes to core are merged alrady

playground/mocks/data/idp/idx/error-poll-400-device-account-invalid.json
```diff
@@ -0,0 +1,45 @@
...
+  "stateHandle":"eyJ6aXAiOiJERUYiLCJhb",
+  "expiresAt":"2021-11-18T20:16:13.000Z",
+  "intent":"LOGIN",
+  "messages":{
+    "type":"array",
+    "value":[
+      {
+        "message":"Your device or account was invalidated. If this is unexpected, contact your administrator for help.",
+        "i18n": {
+          "key": "core.auth.factor.signedNonce.error.deletedDevice"
```
jmelberg-okta, line 12: This key isn't listed in our `login.properties` or being transformed in [i18nTransformer.js](https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/i18nTransformer.js), so the error message will likely default to the `message` value here, resulting in English leaks.

I may be wrong, but it looks like you're trying to surface an error message different from the server. I'd suggest using the [i18nTransformer.js](https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/i18nTransformer.js)'s `I18N_OVERRIDE_MAPPINGS` object to handle this use case. When the Widget sees an error with the key `core.auth.factor.signedNonce.error.deletedDevice`, it can remap it to the expected `user.fail.verifyIdentity` message.

test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
```diff
@@ -34,32 +35,73 @@ const mock = RequestMock()
...
+const deviceInvalidatedErrorMsg = RequestMock()
+  .onRequestTo(/\/idp\/idx\/introspect/)
+  .respond(identifyWithDeviceProbingLoopback)
+  .onRequestTo(/\/idp\/idx\/authenticators\/poll/)
+  .respond((req, res) => {
+    return new Promise((resolve) => setTimeout(function() {
+      res.statusCode = '400';
+      res.setBody(errorDeviceInvalid);
+      resolve(res);
+    }, Constants.TESTCAFE_DEFAULT_AJAX_WAIT + 2000));
```
jmelberg-okta, line 47: I'm not that familiar with how polling works, but do we need this mandatory two second (plus `TESTCAFE_DEFAULT_AJAX_WAIT` **3** second) delay? Via mocking, I'd hope we could avoid unnecessary wait times.

@jhe-okta or @pradeepdewda-okta - Any ideas/suggestions?

test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
```diff
@@ -34,32 +35,73 @@ const mock = RequestMock()
...
+  })
+  .onRequestTo(/2000\/probe/)
+  .respond(null, 200, {
+    'access-control-allow-origin': '*',
```
jmelberg-okta, line 51: Curious - do we return this wildcard for the endpoint or is this mapped to the same origin?

packages/@okta/i18n/src/properties/login_ok_PL.properties
```diff
@@ -750,7 +751,7 @@ oie.authenticator.oktaverify.method.totp.verify.enable.biometrics.point1 = 》Ý
...
 oie.authenticator.oktaverify.method.fastpass.verify.enable.biometrics.mobile = 》Ýöûŕ ŕéšþöñšé ŵåš ŕéçéîṽéð، ƀûţ ýöûŕ öŕĝåñîžåţîöñ ŕéǫûîŕéš ƀîöɱéţŕîçš· Ṁåķé šûŕé ýöûŕ ðéṽîçé šûþþöŕţš ƀîöɱéţŕîçš، Öķţå Ṽéŕîƒý îš ûþ‐ţö‐ðåţé åñð ƀîöɱéţŕîçš åŕé éñåƀļéð ƒöŕ ýöûŕ åççöûñţ îñ Öķţå Ṽéŕîƒý، ţĥéñ ţŕý åĝåîñ· Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ《
-oie.authenticator.oktaverify.method.fastpass.verify.enable.biometrics.desktop = 》Ýöûŕ ŕéšþöñšé ŵåš ŕéçéîṽéð، ƀûţ ýöûŕ öŕĝåñîžåţîöñ ŕéǫûîŕéš ƀîöɱéţŕîçš· Ṁåķé šûŕé ýöûŕ ðéṽîçé šûþþöŕţš ƀîöɱéţŕîçš، Öķţå Ṽéŕîƒý îš ûþ‐ţö‐ðåţé، ƀîöɱéţŕîçš åŕé éñåƀļéð ƒöŕ ýöûŕ åççöûñţ îñ Öķţå Ṽéŕîƒý åñð ýöûŕ ðéṽîçé´š ƀîöɱéţŕîç šéñšöŕš åŕé åççéšîƀļé، ţĥéñ ţŕý åĝåîñ· ฐโ⾼ئ䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ《
+oie.authenticator.oktaverify.method.fastpass.verify.enable.biometrics.desktop = 》Ýöûŕ ŕéšþöñšé ŵåš ŕéçéîṽéð، ƀûţ ýöûŕ öŕĝåñîžåţîöñ ŕéǫûîŕéš ƀîöɱéţŕîçš· Ṁåķé šûŕé ýöûŕ ðéṽîçé šûþþöŕţš ƀîöɱéţŕîçš، Öķţå Ṽéŕîƒý îš ûþ‐ţö‐ðåţé، ƀîöɱéţŕîçš åŕé éñåƀļéð ƒöŕ ýöûŕ åççöûñţ îñ Öķţå Ṽéŕîƒý åñð ýöûŕ ðéṽîçé´š ƀîöɱéţŕîç šéñšöŕš åŕé åççéššîƀļé، ţĥéñ ţŕý åĝåîñ· Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ《
```
pradeepdewda-okta, line 754: add this key in `login.properties` as well..

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -145,6 +147,24 @@ const Body = BaseFormWithPolling.extend(
...
 
+    showCustomFormErrorCallout(error) {
+      const errorSummary = error.responseJSON.errorSummary;
+      let options = {
+        type: 'error',
+        className: 'okta-verify-uv-callout-content',
+        subtitle: errorSummary,
+      };
+
+      const containsSignedNonceError = error.responseJSON.errorSummaryKeys
```
pradeepdewda-okta, line 158: Is this error response is not standard IDX response ?

test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
pradeepdewda-okta: I think he is trying to mock actual `polling`.. There is no effective way in testcafe to mock `setTimeout` or `setInterval` as far as i know.

test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
```diff
@@ -34,32 +35,73 @@ const mock = RequestMock()
...
+
+test
+  .requestHooks(logger, deviceInvalidatedErrorMsg)('add title when device or account is invalidated', async t => {
+    const deviceChallengePollPageObject = await setup(t);
+    await t.expect(deviceChallengePollPageObject.getHeader()).eql('Verifying your identity');
+    await t.expect(deviceChallengePollPageObject.getFooterCancelPollingLink().innerText).eql('Cancel and take me to sign in');
+    await t.expect(deviceChallengePollPageObject.getFooterSignOutLink().length).eql(0);
+    await t.expect(deviceChallengePollPageObject.form.getErrorBoxText()).eql(
+      'Couldn’t verify your identity\n\n' + errorDeviceInvalid.messages.value[0].message);
+    await t.expect(deviceChallengePollPageObject.getSpinner().getStyleProperty('display')).eql('none');
```
pradeepdewda-okta, line 104: IMO you don't have to repeat what has been already verified by other tests.. Keep this test to strictly check for what you added..

playground/mocks/data/idp/idx/error-poll-400-device-account-invalid.json
anatoliiboiarshchykov-okta: Updated

test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
anatoliiboiarshchykov-okta: We need some time to do polling here

src/v2/view-builder/views/device/DeviceChallengePollView.js
anatoliiboiarshchykov-okta: updated

packages/@okta/i18n/src/properties/login_ok_PL.properties
anatoliiboiarshchykov-okta: updated

anatoliiboiarshchykov-okta: Looks like it was already [there](https://github.com/okta/okta-signin-widget/blob/c26adba7c4fd68233c50edc1a38a651b886fefb4/packages/%40okta/i18n/src/properties/login.properties#L1025), as result it was added when I generated the PL translations

test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
anatoliiboiarshchykov-okta: It is mapped to the same origin

jhe-okta:
> * OKTA-443983

@stevennguyen-okta can you please review as well?

stevennguyen-okta (approved):
Title seen in the video looks good

stevennguyen-okta:
Forgot to clarify: this title should be added to all the respective FastPass polling screens for desktop, iOS, and Android as well, across Identify as well as Assurance. So off the top of my head, I would expect this lifecycle error to have a title for these screens:

1. OV desktop custom URI polling - identifier
2. iOS UL polling - identifier
3. Android app link polling - identifier
4. OV desktop custom URI polling - assurance
5. iOS UL polling - assurance
6. Android app link polling - assurance

Just in case this change was added only to desktop polling identifier since I can't tell

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -145,6 +147,23 @@ const Body = BaseFormWithPolling.extend(
...
 
+    showCustomFormErrorCallout(error) {
+      let options = {
+        type: 'error',
+        className: 'okta-verify-uv-callout-content',
+        subtitle: error.responseJSON.errorSummary,
+      };
+
+      const containsSignedNonceError = error.responseJSON.errorSummaryKeys
+        .some((key) => key.includes(SIGNED_NONCE_ERROR_KEY));
```
pradeepdewda-okta, line 158: can there be multiple keys in the response ?

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -815,7 +815,15 @@ deviceTrust.universalLink.fallback.getOktaMobile.title = Get Okta Mobile
 # {0} and {1} are beginning and closing HTML elements.
 deviceTrust.universalLink.fallback.getOktaMobile.subtitle = Go to the {0}AppStore{1}, {0}search{1} for {0}Okta Mobile{1} and tap on {0}GET{1} Okta Mobile. Once installed, sign in to Okta Mobile and follow the instructions to secure your device.
 
+# Fastpass
+core.auth.factor.signedNonce.error.invalidEnrollment=Your device or account was invalidated. Re-enroll your account on Okta Verify and try again.
+core.auth.factor.signedNonce.error.invalidFactor=Your device or account was invalidated. Re-enroll your account on Okta Verify and try again.
```
pradeepdewda-okta, line 820: if this and above message is same.. we can just use one common key. No need to create two keys for duplicate strings.

anatoliiboiarshchykov-okta: Actually, I have copied it from [the core](https://github.com/okta/okta-core/blob/0901ed2c7503633ad66e26e597fac5ebfabfd2a2/resources/src/main/resources/messages-untranslated.properties#L960). Because that core can return those 2 fault codes.
Please advise is it strongly preferred to update the core to use a single error code?

src/v2/view-builder/views/device/DeviceChallengePollView.js
anatoliiboiarshchykov-okta: Yes, it is possible. That is why we are checking to have at least one match.
Also, in the case of multiple i18n codes. The message will be composed of localised i18n code messages, 
${loc(i18nKeys[0])} + '. ' + ${loc(i18nKeys[1])} + '. ' + ${loc(i18nKeys[n])}

this logic can be found [here](https://github.com/okta/okta-signin-widget/blob/f6b6246e537f0980acca98fa5d5421399fc30dbc/src/v2/ion/IonResponseHelper.js#L114)

anatoliiboiarshchykov-okta:
> Forgot to clarify: this title should be added to all the respective FastPass polling screens for desktop, iOS, and Android as well, across Identify as well as Assurance. So off the top of my head, I would expect this lifecycle error to have a title for these screens:
> 
> 1. OV desktop custom URI polling - identifier
> 2. iOS UL polling - identifier
> 3. Android app link polling - identifier
> 4. OV desktop custom URI polling - assurance
> 5. iOS UL polling - assurance
> 6. Android app link polling - assurance
> 
> Just in case this change was added only to desktop polling identifier since I can't tell

I verified some of those cases. Also, I have added a [comment](https://oktainc.atlassian.net/browse/OKTA-443983?focusedCommentId=2248963) for QA in ticket to check it for sure

packages/@okta/i18n/src/properties/login.properties
jmelberg-okta: You can add one of these keys and the value of the other to our remapping tool in [i18nTransformer](https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/i18nTransformer.js#L56).

Better - consider renaming both of these i18n keys to make them more straightforward to consumers:

```suggestion
fastpass.signedNonce.error = Your device or account was invalidated. Re-enroll your account on Okta Verify and try again.
```

Then, in the `i18nTransformer` we can map both of the API keys to our new `fastpass.signedNonce.error` key.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -815,7 +815,15 @@ deviceTrust.universalLink.fallback.getOktaMobile.title = Get Okta Mobile
 # {0} and {1} are beginning and closing HTML elements.
 deviceTrust.universalLink.fallback.getOktaMobile.subtitle = Go to the {0}AppStore{1}, {0}search{1} for {0}Okta Mobile{1} and tap on {0}GET{1} Okta Mobile. Once installed, sign in to Okta Mobile and follow the instructions to secure your device.
 
+# Fastpass
+core.auth.factor.signedNonce.error.invalidEnrollment=Your device or account was invalidated. Re-enroll your account on Okta Verify and try again.
```
jmelberg-okta, line 819: Double checking, is "**on** Okta Verify" grammatically correct? I'd assume `with` or `in` is better suited here.

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -815,7 +815,15 @@ deviceTrust.universalLink.fallback.getOktaMobile.title = Get Okta Mobile
...
+core.auth.factor.signedNonce.error.invalidFactor=Your device or account was invalidated. Re-enroll your account on Okta Verify and try again.
+core.auth.factor.signedNonce.error.invalidDevice=Your device or account was invalidated. If this is unexpected, contact your administrator for help.
+core.auth.factor.signedNonce.error.invalidDeviceStatus=Your device or account was invalidated. If this is unexpected, contact your administrator for help.
```
jmelberg-okta, line 822: In what scenarios is an invalidated account **expected** for users attempting to authenticate? Seems like we could leverage the `core.auth.factor.signedNonce.error.deletedDevice` key for each of these error cases.

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -145,6 +147,23 @@ const Body = BaseFormWithPolling.extend(
       });
     },
 
+    showCustomFormErrorCallout(error) {
+      let options = {
```
jmelberg-okta, line 151: nit: Should be able to treat this as a `const` as we're only modifying object props.

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -9,6 +9,8 @@ import { doChallenge } from '../../utils/ChallengeViewUtil';
 import OktaVerifyAuthenticatorHeader from '../../components/OktaVerifyAuthenticatorHeader';
 import { getSignOutLink } from '../../utils/LinksUtil';
 
+const SIGNED_NONCE_ERROR_KEY = 'auth.factor.signedNonce.error';
```
jmelberg-okta, line 12: nit: Doesn't need to be a top-level constant. Let's move this within the scope of the function using it.

test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
jmelberg-okta: Can we create a ticket to investigate this?

Ideally, we should avoid any test execution delays, like this **5 second wait**. I'd hope after more investigation we can solve this via mocking and/or ensuring our tests are running in parallel so the burden isn't as high for our developers.

jmelberg-okta (approved):
Couple of suggestions for reducing the number of i18n keys used + some wording changes. Once we get agreement there 👉🏼  🚢

packages/@okta/i18n/src/properties/login.properties
pradeepdewda-okta: +1 ^^

You can keep just one string in login.properties and map these different keys in `i18nTransformer`.. No need to change backend.

test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
pradeepdewda-okta: There is one i filled a while ago.. 

https://oktainc.atlassian.net/browse/OKTA-430216

pradeepdewda-okta: one more similar ticket - https://oktainc.atlassian.net/browse/OKTA-365670

src/v2/view-builder/views/device/DeviceChallengePollView.js
anatoliiboiarshchykov-okta: updated

src/v2/view-builder/views/device/DeviceChallengePollView.js
anatoliiboiarshchykov-okta: updated

packages/@okta/i18n/src/properties/login.properties
anatoliiboiarshchykov-okta: updated

packages/@okta/i18n/src/properties/login.properties
anatoliiboiarshchykov-okta: I will check if this can be done, It will require some core changes

packages/@okta/i18n/src/properties/login.properties
anatoliiboiarshchykov-okta: Reached out to @stevennguyen-okta  This one will be verified with writers and changed in a separate ticket if needed.

packages/@okta/i18n/src/properties/login.properties
jhe-okta: line 819 and 820 has the same text, we only need one

jhe-okta (approved):
One change on the properties file then ship.

packages/@okta/i18n/src/properties/login.properties
anatoliiboiarshchykov-okta: updated

pradeepdewda-okta (approved):
Looks good

PR #2315, opened 2021-12-03
Add send push notification

indirathirumalaimurugan-okta:
Design doc: https://oktawiki.atlassian.net/wiki/spaces/~50021507/pages/2360314998/Enterprise+Auth+-+OIE+-+Auto+send+push+notification

1. Add 'Send push automatically'  checkbox to the existing OV push view. It should be checked by default. User can uncheck if need be.

Screen Shot 2021-12-02 at 10 23 31 PM

2. If 'Send push automatically' is unchecked on the previous login, then on the next login user should be taken to a view with 'Send Push' button and 'Send push automatically' checkbox thats is unchecked. 'Send Push' should take to the last screen as above.

Screen Shot 2021-12-07 at 12 30 27 AM

## PR Checklist

- [ ] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:

Usecase 1 (When user is presented with authenticators  list and push is selected) : 
https://okta.box.com/s/tznk759nzdji59jbk1o4fn5xp5435m5o
Usecase 2 (when user has previously chosen to verify with push notification): 
https://okta.box.com/s/vqn8ubekkxuhr447bf0jsayeuugytbcv

### Issue:

- [OKTA-443683](https://oktainc.atlassian.net/browse/OKTA-443683)


assets/sass/common/shared/o-forms/_o-form.scss
```diff
@@ -980,7 +980,7 @@ only screen and (min-resolution: 2dppx) {
 
 /* disabled state */
 .o-form-disabled,
-.o-form-saving .o-form-input > *:not(.o-form-disabled) {
+.o-form-saving .o-form-input > *:not(.o-form-disabled):not(.o-form-input-name-autoChallenge) {
```
pradeepdewda-okta, line 983: I would avoid changing the css globally to address any specific view.. Is there any other way ?

playground/mocks/data/idp/idx/login-with-username-password.json
```diff
@@ -0,0 +1,113 @@
...
...pwfmAKqFtboxnmEtROY7YiDzxjvwMEwhtS2T1Wtf0QpTYlC2C-qOdEN48ZuJxAjxrrJMojm549zbR6eaeRX-HiSfsE2-noeZxuqED15OK9BrPlx7EZ2XgePdD1p_XKf0khJfvGh6Q5Mxc8tbhsFnN6UCFRNNg3ZMDasYLv3YyC48C-294v4zlpyYkY6YJFJ37Tb2KXVhiu5UvFfuJKcVfo0SYZM9O_CZzTTj1tJEXkiJ-BBQfuEnIzoNghNnkPk2iQILgHlYdHQ1R-zLMzEmSJYLDv5JZH2O2ghkJ7SVrroip00ovsygwJkapx_oQU9P6ALciLaC3_sO63ZHL37fg4d86dLxOGVnPpZlOJqbohAPglYsg0gCNSYVrBtZUInYKNJJNYxVUzQCV0ZqNlHUwYv5SulGO2EkuLGki5Uo3X8Ptbb3680Key6njJRL7NO8Su9wS_D3rw8R4fD3lAK9N8zorVTwYlQszSNmaKOttkD1u2Tw1hna1vJC1DUCFOj66UMbG_-xE1LhKUlaTxoLF1YYVMAdiV7CKrzFAh4reaUL_gqrDDp_besFS0oP0aYSPpLAUQnMUGctXfZkFlSdW0gs5kfUFv1plI5Pc6IUK5yIGPolqdDF4c90oYniuJK0f5lcGMe5oDbHVdW8n4jY8gUZ0AwzaDIO4wCZsMQmchdufUx4INtih660Rxzg0runddgnM3W18QplKJjf63D-B6Q_AsdjkphUa4J_uDLVVrJqmEsN7IpQVk-Nsl2cFjV80SmJDMAwIgLu4tEWAehH2qNyoFsnrhJucew2fzsZtADd7ga1r0nLapMFqSKl5wUaW7iUkLAy7HZV4bE6jInA4foAnlW-m28sLlGZTvi8EQEFyqQW2XqtmSeKd-328xEXQqNve-L-WjItHgCBg1kdTXRcE_ZtBAVcURKBvXkyMZSFnUTlzEFbuowQBUkoydTOLVfkXcEaS-ovUryIG1w3NdgjCw7k0S2GA8nYkDQUD2x7wKRTfcomf75rY8T-Zo5l54r6tQg2xv8Tm3NYyShaZLiPhiB6Ez8aGg0HuvZzuoN2u-p3A7uUKTN7tS8dJpeq-MdKoPCp9tToPuae_fcBf7PpQw15GS5wKJCerhc6BlZgkR9NayaV5bCI9HQhBbKr3BZzhIkVY5jXAh73d_Bj_bx.e_A6AQEUcYi7hB6V6_IxFQ",
```
pradeepdewda-okta, line 3: you can keep smaller string for stateHandle..

playground/mocks/data/idp/idx/login-with-username-password.json
```diff
@@ -0,0 +1,113 @@
...
```
pradeepdewda-okta, line 44: same here

pradeepdewda-okta:
@indirathirumalaimurugan-okta nice work !! 👍

src/v2/view-builder/views/shared/ChallengePushView.js
```diff
@@ -47,6 +47,9 @@ const Body = BaseForm.extend(Object.assign(
       const className = this.isOV() ?
         'okta-verify-push-challenge' : ' custom-app-push-challenge';
       this.$el.addClass(className);
+      // Move checkbox below the button
+      const checkbox = this.$el.find('.o-form-fieldset');
```
pradeepdewda-okta, line 51: wondering why this is happening.. i see the same code in `postRender` ChallengeAuthenticatorDataOktaVerifyView.js  as well

assets/sass/common/shared/o-forms/_o-form.scss
indirathirumalaimurugan-okta: I tried keeping the css in `_okta-verify.scss` but this global css overrides it still.

src/v2/view-builder/views/shared/ChallengePushView.js
indirathirumalaimurugan-okta: In both views, by default this line https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/internals/BaseForm.js#L40 adds the input before adding any button and other components to the UI. So had to adjust the positioning in postRender. not sure if there is a better way to do it.

playground/mocks/data/idp/idx/login-with-username-password.json
indirathirumalaimurugan-okta: guess it can be any string

assets/sass/common/shared/o-forms/_o-form.scss
pradeepdewda-okta: `autoChallenge`  is view specific field and should not be added to global css

indirathirumalaimurugan-okta: make sense. I am looking at other options

indirathirumalaimurugan-okta: Have fixed this. please look.

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyView.js
```diff
@@ -7,7 +7,42 @@ import { Body as SelectAuthenticatorVerifyViewBody } from '../SelectAuthenticato
 import { AUTHENTICATOR_KEY } from '../../../ion/RemediationConstants';
 
 const Body = SelectAuthenticatorVerifyViewBody.extend({
+  initialize() {
```
pradeepdewda-okta, line 10: Q: The `Send Push` view seems to be an existing view.. can we just reuse `ChallengePushVIew.js` ?  Why do we need to modify this view.. ?

pradeepdewda-okta: there is no polling involved for this flow ?

src/v2/view-builder/views/shared/ChallengePushView.js
pradeepdewda-okta: Nice catch. This use case we never had before where we have to add any input after the button.. even `keep me signed is` is also above `next` or `login` button.. 

But we need to address it properly, instead of this hack.

We should update our BaseForm in a way that it can choose where to add a input.. before a button or after a button..

Default : existing.. always all inputs before the button
Custom: add inputs (some inputs) after the button

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyView.js
indirathirumalaimurugan-okta: `ChallengePushView` is shown when the user selects 'Get a push notification' from the auth list (usecase 1). 
`Send push` is a new view (not existing now) thats shown only when the user has previously opted for ov push authentication by  unchecking the checkbox ('Send push automatically') (usecase 2). 
They both use different remediation form and `ChallengePushView` cannot be reused.

`ChallengeAuthenticatorDataOktaVerifyView` is being modified because this and the new 'Send push' view is backed up by the same remediation form with slight difference.

src/v2/view-builder/views/shared/ChallengePushView.js
indirathirumalaimurugan-okta: I will look into this

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyView.js
pradeepdewda-okta: Got it.. thanks for clarifying.. If possible just create another view instead of updating `ChallengeAuthenticatorDataOktaVerifyView`

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyView.js
```diff
@@ -7,7 +7,42 @@ import { Body as SelectAuthenticatorVerifyViewBody } from '../SelectAuthenticato
...
+      // Move checkbox below the button
+      const checkbox = this.$el.find('.o-form-fieldset');
+      checkbox.length && this.$el.find('.o-form-fieldset-container').append(checkbox);
+    }
+  },
+
   getUISchema() {
+    if (this.isPushOnlyFlow()) {
+      // Prevent from displaying radio buttons on the UI
+      const uiSchemas = BaseForm.prototype.getUISchema.apply(this, arguments);
```
mauriciocastillosilva-okta, line 43: nit: since it is the same logic and lines below, maybe we can do 
```
// Change the UI schema to not display radios here.
const uiSchemas = BaseForm.prototype.getUISchema.apply(this, arguments);
const methodsSchema = uiSchemas.find(schema => schema.name === 'authenticator.methodType');
if (this.isPushOnlyFlow()) {
  // Only one item, so no need to sort
  return methodsSchema;
}
```
Although this might change if we extend `BaseForm` per my previous comment

playground/mocks/data/idp/idx/challenge-with-push-notification.json
```diff
@@ -0,0 +1,268 @@
+{
```
mauriciocastillosilva-okta, line 1: can we rename this file to be similar to the other `challenge-poll` remediation files? Something like `authenticator-verification-okta-verify-push-autoChallenge.json`?
And maybe some suffix like that for the other files that are already similar to the files we currently have. It seems that the only difference is the `autoChallenge` bit, but the rest of the forms are the same

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyView.js
```diff
@@ -7,7 +7,42 @@ import { Body as SelectAuthenticatorVerifyViewBody } from '../SelectAuthenticato
 import { AUTHENTICATOR_KEY } from '../../../ion/RemediationConstants';
 
 const Body = SelectAuthenticatorVerifyViewBody.extend({
```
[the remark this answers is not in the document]
indirathirumalaimurugan-okta: As per suggestion, I will be creating a separate view for the 'Send Push' use case. That view will extend from BaseForm directly. I am working on it will update the PR soon.

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyView.js
indirathirumalaimurugan-okta: this piece of code will also be moved to the new view. Also, its needed to filter out `authenticator.methodType` otherwise radio button would be displayed on the UI for 'Get a push notification' type.

src/v2/view-builder/views/shared/ChallengePushView.js
```diff
@@ -43,7 +47,17 @@ const Body = BaseForm.extend(Object.assign(
       );
     },
 
+    render() {
+      BaseForm.prototype.render.apply(this, arguments);
+      // Move checkbox below the button
```
pradeepdewda-okta, line 52: Add more to the comment --
```
Move checkbox below the button.
Checkbox is rendered by BaseForm using remediation response and hence always gets added by default above buttons.
```

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyPushOnlyView.js
```diff
@@ -0,0 +1,50 @@
...
+    // Move checkbox below the button
+    const checkbox = this.$el.find('.o-form-fieldset');
+    checkbox.length && this.$el.find('.o-form-fieldset-container').append(checkbox);
+  },
+
+  getUISchema() {
+    const uiSchemas = BaseForm.prototype.getUISchema.apply(this, arguments);
+    return uiSchemas.filter(schema => schema.name !== 'authenticator.methodType');
+  },
+  noButtonBar: true,
```
pradeepdewda-okta, line 40: put this on top above initialize

src/v2/view-builder/views/ov/ChallengeAuthenticatorOktaVerifyView.js
```diff
@@ -1,23 +1,23 @@
...
+      this.Body = ChallengeAuthenticatorDataOktaVerifyPushOnlyView;
     } else {
-      this.Body = ChallengeAuthenticatorOktaVerifyView;
+      this.Body = ChallengeAuthenticatorDataOktaVerifyView;
     }
   },
 
   isPushOnlyFlow() {
-    const uiSchema = this.options.currentViewState.uiSchema;
+    const uiSchema = this.options?.currentViewState?.uiSchema;
```
pradeepdewda-okta, line 18: were you not able to use `authenticatorMethodType` of AppState to get the methodType ?

pradeepdewda-okta: OR this.model.get('authenticator.methodType') ??

test/testcafe/spec/ChallengeOktaVerifyPush_spec.js
```diff
@@ -58,6 +73,69 @@ test
...
+
+    // Verify links
+    await t.expect(await challengeOktaVerifyPushPageObject.switchAuthenticatorLinkExists()).ok();
+    await t.expect(challengeOktaVerifyPushPageObject.getSwitchAuthenticatorLinkText()).eql('Verify with something else');
+    await t.expect(await challengeOktaVerifyPushPageObject.signoutLinkExists()).ok();
+    await t.expect(challengeOktaVerifyPushPageObject.getSignoutLinkText()).eql('Back to sign in');
+  });
+
+test
+  .requestHooks(logger, sendPushWaitMock)('should call polling API and checkbox should be clickable after polling started', async t => {
```
pradeepdewda-okta, line 110: Earlier you mentioned there is no polling for this view?

indirathirumalaimurugan-okta: There are 2 views. Existing one `ChallengePushView` where just the checkbox is added has polling. I was talking about the new view `ChallengeAuthenticatorDataOktaVerifyPushOnlyView` (Send push flow) that has no polling.

src/v2/view-builder/views/ov/ChallengeAuthenticatorOktaVerifyView.js
indirathirumalaimurugan-okta: Using `this.options.appState.get('authenticatorMethodType')`  returns the value `push` but here its required to get the possible options under the method type 'push'. this is how the uiSchema object looks and the goal is to find out whether there is just one option and its of value `push`. Options could have multiple values. based on what it has, we are deciding on the view. Said that, do you think there is a better way to achieve the same?

```{
    "name": "authenticator.methodType",
    "required": true,
    "options": [
        {
            "value": "push",
            "label": "Get a push notification"
        },
        {
            "value": "totp",
            "label": "Enter a code"
        },
        {
            "value": "signed_nonce",
            "label": "Use Okta FastPass"
        }
    ],
    "value": "push",
    "label-top": true,
    "data-se": "o-form-fieldset-authenticator.methodType",
    "type": "radio"
}
```

pradeepdewda-okta: See if this method is helpful to count the authenticator options ? `getRemediationAuthenticationOptions`

indirathirumalaimurugan-okta: Discussed offline. `getRemediationAuthenticationOptions` method does not help here as the api response structure is different. Would be using `getSchemaByName` instead and from that can get the `options`.

src/v2/view-builder/views/ov/ChallengeAuthenticatorOktaVerifyView.js
```diff
@@ -0,0 +1,22 @@
...
+    if (this.isPushOnlyFlow()) {
+      this.Body = ChallengeAuthenticatorDataOktaVerifyPushOnlyView;
+    } else {
+      this.Body = ChallengeAuthenticatorDataOktaVerifyView;
+    }
+  },
+
+  isPushOnlyFlow() {
+    const methodType = this.options.appState.getSchemaByName('authenticator.methodType');
+    const methodTypeOptions = methodType?.options;
```
pradeepdewda-okta, line 19: can get rid of methodTypeOptions as well..

pradeepdewda-okta (approved):
Looks good to me..

🥇

playground/mocks/data/idp/idx/authenticator-verification-factor-list.json
```diff
@@ -0,0 +1,195 @@
+{
```
mauriciocastillosilva-okta, line 1: I think we can use https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/data/idp/idx/authenticator-verification-select-authenticator-without-signed-nonce.json instead of adding this new json file. The difference is Email instead of Password, but for the purposes of this file, what we really care about is that it has Okta Verify push

playground/mocks/data/idp/idx/authenticator-verification-okta-verify-send-push-autoChallenge.json
```diff
@@ -0,0 +1,305 @@
+{
```
mauriciocastillosilva-okta, line 1: and following my other comments, this should be "authenticator-verification-data-okta-verify-push.json" (similar to the other file with `authenticator-verification-data` remediation form, but this one does not have `autoChallenge` on

playground/mocks/data/idp/idx/identify-with-username-password.json
```diff
@@ -0,0 +1,113 @@
+{
```
mauriciocastillosilva-okta, line 1: We can use https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/data/idp/idx/identify-with-password.json, it is the same (I think), so no need to create a new file

src/v2/view-builder/ViewFactory.js
```diff
@@ -230,7 +230,7 @@ const VIEWS_MAPPING = {
   },
   [RemediationForms.AUTHENTICATOR_VERIFICATION_DATA]: {
     [AUTHENTICATOR_KEY.PHONE]: ChallengeAuthenticatorDataPhoneView,
-    [AUTHENTICATOR_KEY.OV]: ChallengeAuthenticatorDataOktaVerifyView,
+    [AUTHENTICATOR_KEY.OV]: ChallengeAuthenticatorOktaVerifyView,
```
mauriciocastillosilva-okta, line 233: ideally so we follow the name conventions that it seems we already have, this new file should have part of the remediation form in the name "ChallengeAuthenticatorData*", so can we rename it to either "ChallengeAuthenticatorDataOktaVerifyBaseView" or "ChallengeAuthenticatorDataOktaVerifyMasterView"?

or if you make the change for the other comments I made about exporting Body in the other 2 files, then this can actually just be named "ChallengeAuthenticatorDataOktaVerifyView" (as it is right now)

src/v2/view-builder/views/ov/ChallengeAuthenticatorOktaVerifyView.js
```diff
@@ -0,0 +1,22 @@
+import BaseAuthenticatorView from '../../components/BaseAuthenticatorView';
+import { 
+  Body as ChallengeAuthenticatorDataOktaVerifyPushOnlyView
+} from './ChallengeAuthenticatorDataOktaVerifyPushOnlyView';
+import { Body as ChallengeAuthenticatorDataOktaVerifyView } from './ChallengeAuthenticatorDataOktaVerifyView';
+
+export default BaseAuthenticatorView.extend({
```
mauriciocastillosilva-okta, line 7: this file needs the `BaseFooter` too, right?
```
Footer: BaseFooter,
```

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyView.js
```diff
@@ -46,7 +46,12 @@ const Body = SelectAuthenticatorVerifyViewBody.extend({
   }
 });
 
-export default BaseAuthenticatorView.extend({
+const AuthenticatorView = BaseAuthenticatorView.extend({
```
mauriciocastillosilva-okta, line 49: seems that we don't really need to create a View here, so maybe we can just export the Body instead?
```
export default Body;
```
and we could even rename this to "ChallengeAuthenticatorDataOktaVerifyBody.js"

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyPushOnlyView.js
```diff
@@ -0,0 +1,53 @@
+import { createButton, loc } from 'okta';
+import { BaseForm } from '../../internals';
+import BaseAuthenticatorView from '../../components/BaseAuthenticatorView';
+
+const Body = BaseForm.extend(Object.assign({
+  noButtonBar: true,
+
+  initialize() {
+    BaseForm.prototype.initialize.apply(this, arguments);
+    this.addView();
```
mauriciocastillosilva-okta, line 10: nit: we don't really need another function, we coult just add the inputs here.

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyPushOnlyView.js
```diff
@@ -0,0 +1,53 @@
...
+    checkbox.length && this.$el.find('.o-form-fieldset-container').append(checkbox);
+  },
+
+  getUISchema() {
+    const uiSchemas = BaseForm.prototype.getUISchema.apply(this, arguments);
+    return uiSchemas.filter(schema => schema.name !== 'authenticator.methodType');
+  },
+}));
+
+const AuthenticatorView = BaseAuthenticatorView.extend({
```
mauriciocastillosilva-okta, line 46: similar to my other comment, seems that we don't really need to create a View here, so exporting the Body instead should do the trick
```
export default Body;
```
and we could even rename this to "ChallengeAuthenticatorDataOktaVerifyPushOnlyBody.js"

src/v2/view-builder/views/shared/ChallengePushView.js
```diff
@@ -25,6 +25,10 @@ const Body = BaseForm.extend(Object.assign(
 
     initialize() {
       BaseForm.prototype.initialize.apply(this, arguments);
+      // 'hasSavingState' would be true when the auth key is custom_app.
+      // Setting it to false when auth key is okta_verify.
+      // So that 'o-form-saving' css class is not added while polling and checkbox remains enabled. 
+      this.hasSavingState = !this.isOV();
```
mauriciocastillosilva-okta, line 31: should this then be
```
if (this.isOV()) {
  this.hasSavingState = false;
}
```
otherwise we would be overriding the value in flows where this is not OV, without knowing what it was before

Also, should we also check if `autoChallenge` is present? Otherwise we don't need to change this right? so maybe something like
```
if (this.isOVWithAutoChallenge() {
  this.hasSavingState = false;
}
```

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyPushOnlyView.js
```diff
@@ -0,0 +1,53 @@
...
+      },
+    }));
+    this.add(
+      `<span class='accessibility-text' role='alert'>${loc('oie.okta_verify.sendPushButton', 'login')}</span>`,
+    );
+  },
+
+  render() {
+    BaseForm.prototype.render.apply(this, arguments);
+    this.$el.addClass('okta-verify-push-challenge');
```
mauriciocastillosilva-okta, line 32: I think we can add the class directly to the form with `className: 'okta-verify-push-challenge',` (probably around like 6 or so, right below `BaseForm.extends`

test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
```diff
@@ -46,4 +49,29 @@ export default class ChallengeOktaVerifyPushPageObject extends ChallengeFactorPa
     return this.form.getElement(FORM_INFOBOX_WARNING);
   }
 
+  subtitleExists() {
```
mauriciocastillosilva-okta, line 52: nit: `hasSubtitle()` (and we can probably add this to the parent PageObject, not sure if any other test needs it, but seems that it is a good candidate to be reused

test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
```diff
@@ -46,4 +49,29 @@ export default class ChallengeOktaVerifyPushPageObject extends ChallengeFactorPa
...
+
+  getAutoChallengeCheckboxLabel(selector) {
+    return this.form.getElement(selector);
+  }
+
+  async clickAutoChallengeCheckbox(selector) {
+    await this.t.click(this.form.getElement(selector));
+  }
+
+  async  formWithClassExist() {
```
mauriciocastillosilva-okta, line 68: let's rename this to `isOktaVerifyPushChallengeForm()`, so it is a little more explicit what we are testing

test/testcafe/spec/ChallengeAuthenticatorOktaVerifyPushOnly_spec.js
```diff
@@ -0,0 +1,88 @@
...
+
+async function setup(t) {
+  const challengeOktaVerifyPushPageObject = new ChallengeOktaVerifyPushPageObject(t);
+  await challengeOktaVerifyPushPageObject.navigateToPage();
+  return challengeOktaVerifyPushPageObject;
+}
+
+test.requestHooks(mockOktaVerifySendPushOnly)('should load view with a button and a checkbox when push is the only method type', async t => {
+  const AUTO_CHALLENGE_CHECKBOX_SELECTOR = '[name="authenticator.autoChallenge"]';
+  const AUTO_CHALLENGE_CHECKBOX_LABEL_SELECTOR = '[data-se-for-name="authenticator.autoChallenge"]';
```
mauriciocastillosilva-okta, line 37: we should move these 2 constants to the page object, that way the page object "knows" how to find or click autoChallenge checkbox, and then we don't need to define them in all the tests specs

test/testcafe/spec/ChallengeOktaVerifyPush_spec.js
```diff
@@ -5,8 +5,12 @@ import { checkConsoleMessages } from '../framework/shared';
 
 import pushPoll from '../../../playground/mocks/data/idp/idx/authenticator-verification-okta-verify-push';
 import success from '../../../playground/mocks/data/idp/idx/success';
+import sendPushPoll from '../../../playground/mocks/data/idp/idx/authenticator-verification-okta-verify-push-autoChallenge';
```
mauriciocastillosilva-okta, line 8: nit: rename to `pushPollAutoChallenge`

playground/mocks/data/idp/idx/authenticator-verification-okta-verify-factor-list.json
```diff
@@ -0,0 +1,244 @@
+{
```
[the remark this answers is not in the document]
indirathirumalaimurugan-okta: `autochallenge` is not required for this view. So this can be deleted and found an existing json file (`authenticator-verification-data-ov-only-without-device-known`) that can be used.

test/testcafe/spec/ChallengeOktaVerifyPush_spec.js
```diff
@@ -25,6 +29,16 @@ const pushWaitMock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/challenge/poll')
   .respond(pushPoll);
 
+const sendPushMock = RequestMock()
```
mauriciocastillosilva-okta, line 32: nit: change the names of this and the next constant to reflect they have autoChallenge. Maybe `pushAutoChallengeMock` and `pushWaitAutoChallengeMock`?

mauriciocastillosilva-okta (commented):
test reviewed. 
Overall looks great, I left some general comments, but the logic looks good to me

test/testcafe/spec/ChallengeAuthenticatorOktaVerifyPushOnly_spec.js
indirathirumalaimurugan-okta: the value of these constants differs based on the view. And test spec of both the views share the same page object. So the page object has function that takes this selector as arguments. to prevent from creating 4 functions I kept the constants in the test file itself. does it make sense?

test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
indirathirumalaimurugan-okta: have kept the name as `subtitleExists` as thats being followed for similar 'exist' check. and moved it to `BasePageObject`

src/v2/view-builder/views/shared/ChallengePushView.js
indirathirumalaimurugan-okta: This view shows 'Get a push notification' view that has this button disabled. Going forward the view will also have the checkbox. In this specific case I dont think checking for `isOVWithAutoChallenge` is required. My understanding is there cannot be any other ov factor in this page

indirathirumalaimurugan-okta: the same view is used when the key is 'okta_verify' or 'custom_app'. Currently `hasSavingState` is true always. With this change we are forcing `hasSavingState` to be true only when key is other than 'okta_verify'. Its same as having it under `if` condition as suggested. If you still feel otherwise, can change it.

src/v2/view-builder/views/ov/ChallengeAuthenticatorOktaVerifyView.js
indirathirumalaimurugan-okta: Only the `ChallengeAuthenticatorDataOktaVerifyBody` needs to override `Footer: BaseFooter` as it does not need to show the link for 'Verify with something else` 
`ChallengeAuthenticatorDataOktaVerifyPushOnlyBody` view uses the `AuthenticatorFooter` to show 'Verify with something else`

test/testcafe/spec/ChallengeAuthenticatorOktaVerifyPushOnly_spec.js
mauriciocastillosilva-okta: oh... can we use another selector? I see you are using `[data-se="o-form-fieldset-autoChallenge"]` in the View, would that work?

playground/mocks/data/idp/idx/authenticator-verification-okta-verify-send-push-autoChallenge.json
indirathirumalaimurugan-okta: have renamed this to `authenticator-verification-data-okta-verify-push-autoChallenge-off` and 
renamed `authenticator-verification-okta-verify-push-autoChallenge` to `authenticator-verification-okta-verify-push-autoChallenge-on`

test/testcafe/spec/ChallengeAuthenticatorOktaVerifyPushOnly_spec.js
indirathirumalaimurugan-okta: even they differ. Its automatically set based on the remediation info. And that info differs for these views. 
For `ChallengeAuthenticatorDataOktaVerifyPushOnlyBody` checkbox container has attribute as `[data-se="o-form-fieldset-authenticator.autoChallenge"]`

For `ChallengePushView` checkbox container has attribute as `[data-se="o-form-fieldset-autoChallenge"]`

src/v2/view-builder/views/ov/ChallengeAuthenticatorOktaVerifyView.js
indirathirumalaimurugan-okta: Didnt get it. could you explain?

pradeepdewda-okta:
@mauriciocastillosilva-okta @indirathirumalaimurugan-okta Naming files ending with `Body` doesn't make much sense to me.. There are no other existing pattern like that..

These are still the Views (form views) -- Let's not create another abstraction to confuse developer in identifying differences between Body and View.. 

I would suggest to keep the file name ends with View..

test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
```diff
@@ -46,4 +48,25 @@ export default class ChallengeOktaVerifyPushPageObject extends ChallengeFactorPa
     return this.form.getElement(FORM_INFOBOX_WARNING);
   }
 
+  async autoChallengeInputExists(selector) {
```
mauriciocastillosilva-okta, line 51: nit: I was thinking about this a little more, and what we could do is add a selector like `$('[name$="autoChallenge"]` (attribute ends with) here, so we wouldn't need to pass the selector. Maybe give it a try?
```
return this.form.elementExist('[name$="autoChallenge"]');
```

mauriciocastillosilva-okta (commented):
just 2 minor comments, rest looks great!

src/v2/view-builder/views/ov/ChallengeAuthenticatorDataOktaVerifyPushOnlyFormView.js
```diff
@@ -0,0 +1,41 @@
+import { createButton, loc } from 'okta';
+import { BaseForm } from '../../internals';
+
+const Body = BaseForm.extend(Object.assign({
+  className: 'okta-verify-push-challenge',
+  noButtonBar: true,
```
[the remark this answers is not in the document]
indirathirumalaimurugan-okta: Pradeep and I spoke about this. He had also mentioned it in slack thread. If we go with default button, positioning checkbox below the button in `postRender` doesn't work. That is because the default button is added via `save()` function which is called after `postRender()`.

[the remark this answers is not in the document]
mauriciocastillosilva-okta: Interesting. What we have is good, but if we ever change the classnames or styles of the primary button in a form, we might not get those changes in this view. I see we are adding the right classnames, so it should be ok, but maybe we can look into it in this PR or a future one

mauriciocastillosilva-okta (approved):

assets/sass/v2/_okta-verify.scss
```diff
@@ -56,3 +56,21 @@
...
   }
 }
+
+.okta-verify-send-push-form {
+  .o-form-button-bar {
+    padding-bottom: 10px;
+  }
+}
+
+.okta-verify-push-challenge {
```
mauriciocastillosilva-okta, line 66: do you still need this class? Seems that after the last commit this might no longer be needed...

indirathirumalaimurugan-okta: this is still used to align the checkbox on this view
Screen Shot 2021-12-06 at 1 31 29 PM

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -1000,6 +1000,7 @@ oie.okta_verify.totp.enterCodeText = Enter code from Okta Verify app
 oie.okta_verify.push.title = Get a push notification
 oie.okta_verify.push.sent = Push notification sent
 oie.okta_verify.push.resend = Resend push notification
+oie.okta_verify.sendPushButton = Send push
```
jmelberg-okta, line 1003: Calling this out as an FYI, but this breaks existing customizations for customers who have altered the existing [`oktaverify.send`](https://github.com/okta/okta-signin-widget/blob/c7aaf9159fdf6af7a2102d485016d7817e50f670/packages/%40okta/i18n/src/properties/login.properties#L570) key.

I'd suggest confirming with product (and reach out to DevEx who maintains the Changelog for this repo) that customers migrating will need to update their customization keys.

jmelberg-okta (approved):

PR #2365, opened 2022-01-25
fix: trigger error when primaryauth fails after idp discovery with passwordless

mauriciocastillosilva-okta:
## Description:

- We create a different model in the IdPDiscoveryController, so the errors are not been triggered properly

## PR Checklist

- [x] Added unit tests?
- [ ] Added e2e tests

### Screenshot/Video:

|Before|After|
|---|---|
| siw-idpdiscovery-error | siw-idpdiscovery-fix | 

### Issue:

- [OKTA-455199](https://oktainc.atlassian.net/browse/OKTA-455199)

### DIFF
--- a/src/IDPDiscoveryController.js
+++ b/src/IDPDiscoveryController.js
@@ -103,6 +103,10 @@ export default PrimaryAuthController.extend({
       );
 
       // Events to set the transaction attributes on the app state.
+      this.listenTo(primaryAuthModel, 'error', function(src, errObj) {
+        this.toggleButtonState(false);
+        this.model.trigger('error', this.model, errObj);
+      });
       this.addModelListeners(primaryAuthModel);
       // Make the primary auth request
       primaryAuthModel.save();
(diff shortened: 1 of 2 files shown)


aarongranick-okta (approved):

src/IDPDiscoveryController.js
```diff
@@ -103,6 +103,10 @@ export default PrimaryAuthController.extend({
       );
 
       // Events to set the transaction attributes on the app state.
+      this.listenTo(primaryAuthModel, 'error', function(src, errObj) {
+        this.toggleButtonState(false);
+        this.model.trigger('error', this.model, errObj);
+      });
       this.addModelListeners(primaryAuthModel);
```
haishengwu-okta, line 110: Looks like the primary reason was we forgot to propagate certain events from `primaryAuthModel` to `this.model` which has listened to many events already.
Not for this PR but I wonder if we'd like to try a follow up refactoring that simply bubble up events (errors, setTransaction etc) would be good enough without even invoke `addModelListeners`.

haishengwu-okta (approved):

src/IDPDiscoveryController.js
mauriciocastillosilva-okta: I was thinking the same, either bubble up events somehow or override `this.model` instead of creating a new model here. I didn't want to do it now since this is urgent and we would need more testing if we refactor more. I created https://oktainc.atlassian.net/browse/OKTA-463593 to follow up, take a look @haishengwu-okta and add any more details if I missed something

PR #1782, opened 2021-03-26
Implement route for loading register page

haishengwu-okta:
## Description:

- Add route `/signin/register` and show registration form directly
- Show terminal error if registration is not supported.

## PR Checklist

- [ ] Have you verified the basic functionality for this change?
- [ ] Added unit tests?
- [ ] Added e2e tests
- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?

### Screenshot/Video:
Screen Shot 2021-03-31 at 1 45 21 PM
Screen Shot 2021-03-31 at 1 45 00 PM

### Issue:

- okta-380941

PR #2368, opened 2022-01-26
removes route logic for /signin/register on V2

aarongranick-okta:
## Description:
Removes logic based on the browser location in V2. This logic interferes with proper implementation of this feature server-side.

Embedded widgets can use the `flow` configuration option to achieve the same result.

### Issue:

- [OKTA-462003](https://oktainc.atlassian.net/browse/OKTA-462003)


haishengwu-okta (approved):
Probably can also clean the properties? https://github.com/okta/okta-signin-widget/pull/1782/files#diff-391ef791333e16bfeed711da3def133ca2471a6ebbe1e917ecc61330a1530fe3

pradeepdewda-okta (approved):
neat

aarongranick-okta:
> Probably can also clean the properties? https://github.com/okta/okta-signin-widget/pull/1782/files#diff-391ef791333e16bfeed711da3def133ca2471a6ebbe1e917ecc61330a1530fe3

I am thinking we want to leave this key because the backend will still want to use the widget to display an error in the case that registration is not enabled. As long as backend uses the previous key `oie.registration.is.not.enabled` it should have translations.

moushmibanerjee-okta (approved):
LGTM.

PR #2372, opened 2022-01-27
OKTA-461362 Apply biometrics error in SSO extension view

yannongli-okta:
## Description:
- Previously, the biometrics error was showed in loopback/custom uri/universal link/app links. SSO extension uses a different view so we need to apply the change to it separately.
- Added test cafe tests for all binding methods.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
- e2e demo: https://okta.box.com/s/7mfgl7q9wappfs2pf8ogpczt3bpq9z3z
- playground screenshot:
Screen Shot 2022-01-26 at 5 13 35 PM

### Reviewers:
@okta/device-platform @jhe-okta

### Issue:

- [OKTA-461362](https://oktainc.atlassian.net/browse/OKTA-461362)

### DIFF
--- a/playground/mocks/spec-okta-api/idp/idx/index.js
+++ b/playground/mocks/spec-okta-api/idp/idx/index.js
@@ -1,5 +1,9 @@
 const templateHelper = require('../../../config/templateHelper');
 const cancelTransaction = require('../../../data/idp/idx/identify-with-no-sso-extension');
+const biometricsErrorMobile = 
+  require('../../../data/idp/idx/error-okta-verify-uv-fastpass-verify-enable-biometrics-mobile');
+const biometricsErrorDesktop = 
+  require('../../../data/idp/idx/error-okta-verify-uv-fastpass-verify-enable-biometrics-desktop');
 
 const idx = [
   '/idp/idx',
@@ -47,7 +51,7 @@ const ssoExtension = [
       next();
     },
     // TODO: find a way to improve this, now the mock config is not always on responseConfig
-    template: cancelTransaction,
+    template: cancelTransaction, // To test biometrics error, change it to biometricsErrorMobile/biometricsErrorDesktop
   })
 ];
 
--- a/playground/mocks/spec-okta-api/idp/idx/index.js
+++ b/playground/mocks/spec-okta-api/idp/idx/index.js
@@ -1,9 +1,9 @@
 const templateHelper = require('../../../config/templateHelper');
 const cancelTransaction = require('../../../data/idp/idx/identify-with-no-sso-extension');
-const biometricsErrorMobile = 
-  require('../../../data/idp/idx/error-okta-verify-uv-fastpass-verify-enable-biometrics-mobile');
-const biometricsErrorDesktop = 
-  require('../../../data/idp/idx/error-okta-verify-uv-fastpass-verify-enable-biometrics-desktop');
+// const biometricsErrorMobile = 
+//   require('../../../data/idp/idx/error-okta-verify-uv-fastpass-verify-enable-biometrics-mobile');
+// const biometricsErrorDesktop = 
+//   require('../../../data/idp/idx/error-okta-verify-uv-fastpass-verify-enable-biometrics-desktop');
 
 const idx = [
   '/idp/idx',
--- a/src/v2/controllers/FormController.js
+++ b/src/v2/controllers/FormController.js
@@ -14,6 +14,8 @@ import ViewFactory from '../view-builder/ViewFactory';
 import IonResponseHelper from '../ion/IonResponseHelper';
 import { getV1ClassName } from '../ion/ViewClassNamesFactory';
 import { FORMS, TERMINAL_FORMS, FORM_NAME_TO_OPERATION_MAP } from '../ion/RemediationConstants';
+import { OV_UV_ENABLE_BIOMETRICS_FASTPASS_DESKTOP, 
+  OV_UV_ENABLE_BIOMETRICS_FASTPASS_MOBILE } from '../view-builder/utils/Constants';
 import Util from '../../util/Util';
 import sessionStorageHelper from '../client/sessionStorageHelper';
 import { clearTransactionMeta } from '../client';
@@ -216,12 +218,14 @@ export default Controller.extend({
           onSuccess();
         }
       }).catch(error => {
-        if (error.stepUp) {
+        if (error.stepUp && !this.isBiometricsErrorFlow(error)) {
           // Okta server responds 401 status code with WWW-Authenticate header and new remediation
           // so that the iOS/MacOS credential SSO extension (Okta Verify) can intercept
           // the response reaches here when Okta Verify is not installed
           // we need to return an idx object so that
           // the SIW can proceed to the next step without showing error
+          // Also, if the received error is biometrics required error,
+          // we will step into below else block to show custom error messages
           this.handleIdxResponse(error);
         } else {
           this.showFormErrors(model, error, this.formView.form);
@@ -230,6 +234,17 @@ export default Controller.extend({
       .finally(() => {
         this.toggleFormButtonState(false);
       });
+  },
+
+  isBiometricsErrorFlow(error) {
+    if (!error?.context?.messages) {
+      return false;
+    }
+
+    const messagesObjs = error.context.messages.value;
+    const biometricsErrorKeys = [OV_UV_ENABLE_BIOMETRICS_FASTPASS_DESKTOP, OV_UV_ENABLE_BIOMETRICS_FASTPASS_MOBILE];
+    return messagesObjs && Array.isArray(messagesObjs)
+    && messagesObjs.some(messagesObj => _.contains(biometricsErrorKeys, messagesObj.i18n?.key));
   },
 
   transformIdentifier(formName, model) {
--- a/src/v2/view-builder/utils/ChallengeViewUtil.js
+++ b/src/v2/view-builder/utils/ChallengeViewUtil.js
@@ -13,7 +13,9 @@ import { loc, View, createButton, _ } from 'okta';
 import hbs from 'handlebars-inline-precompile';
 import Enums from '../../../util/Enums';
 import Util from '../../../util/Util';
-import { FASTPASS_FALLBACK_SPINNER_TIMEOUT, IDENTIFIER_FLOW } from '../utils/Constants';
+import { FASTPASS_FALLBACK_SPINNER_TIMEOUT, IDENTIFIER_FLOW, 
+  OV_UV_ENABLE_BIOMETRICS_FASTPASS_DESKTOP, 
+  OV_UV_ENABLE_BIOMETRICS_FASTPASS_MOBILE  } from '../utils/Constants';
 
 export function appendLoginHint(deviceChallengeUrl, loginHint) {
   if (deviceChallengeUrl && loginHint) {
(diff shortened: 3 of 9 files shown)


src/v2/controllers/FormController.js
```diff
@@ -232,6 +236,17 @@ export default Controller.extend({
...
 
+  isBiometricsErrorFlow(error) {
+    if (!error?.context?.messages) {
+      return false;
+    }
+
+    const messagesObjs = error.context.messages.value;
+    const biometricsErrorKeys = [OV_UV_ENABLE_BIOMETRICS_FASTPASS_DESKTOP, OV_UV_ENABLE_BIOMETRICS_FASTPASS_MOBILE];
+    return messagesObjs && Array.isArray(messagesObjs)
+    && messagesObjs.some(messagesObj => _.contains(biometricsErrorKeys, messagesObj.i18n?.key));
```
jhe-okta, line 247: It does not seem reliable to depend on an i18n key from the response. Is there other signals like error type that you can check?

yannongli-okta: > It does not seem reliable to depend on an i18n key from the response. Is there other signals like error type that you can check?

@jhe-okta here is the sample API response. I can add a class check of `ERROR` in this method.

jhe-okta: @yannongli-okta can you add the link?

yannongli-okta: @jhe-okta sorry here it is https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/data/idp/idx/error-okta-verify-uv-fastpass-verify-enable-biometrics-desktop.json#L10

src/v2/controllers/FormController.js
```diff
@@ -216,12 +218,14 @@ export default Controller.extend({
           onSuccess();
         }
       }).catch(error => {
-        if (error.stepUp) {
+        if (error.stepUp && !this.isBiometricsErrorFlow(error)) {
```
pradeepdewda-okta, line 221: we should not do this `!this.isBiometricsErrorFlow(error)` ... View specific code should remain in Views.

src/v2/controllers/FormController.js
pradeepdewda-okta: Good catch @jhe-okta .. we should avoid view specific logic from FormController..

pradeepdewda-okta: Instead use `showCustomFormErrorCallout` attribute 

search in the repo and see how it is being used.. 

for eg take a look at `ChallengeOktaVerifyFastPassView.js`

pradeepdewda-okta: Ignore earlier comments.. Is that error not coming as IDX response ?

src/v2/view-builder/views/TerminalView.js
```diff
@@ -132,6 +134,10 @@ const Body = BaseForm.extend({
     } else if (this.options.appState.containsMessageWithI18nKey(IDX_RETURN_LINK_OTP_ONLY)) {
       this.add(OTPInformationTerminalView);
       hasCustomView = true;
+    } else if (this.options.appState.containsMessageStartingWithI18nKey(OV_UV_ENABLE_BIOMETRICS_FASTPASS)) {
+      const options = getBiometricsErrorOptions(messagesObjs, true);
```
pradeepdewda-okta, line 138: can you extract this logic out and put it into `ChallengeViewUtil` and just import the view that you want to add like `OTPInformationTerminalView`

playground/mocks/spec-okta-api/idp/idx/index.js
```diff
@@ -1,5 +1,9 @@
 const templateHelper = require('../../../config/templateHelper');
 const cancelTransaction = require('../../../data/idp/idx/identify-with-no-sso-extension');
+// const biometricsErrorMobile =
```
pradeepdewda-okta, line 3: remove

playground/mocks/spec-okta-api/idp/idx/index.js
```diff
@@ -47,7 +51,7 @@ const ssoExtension = [
       next();
     },
     // TODO: find a way to improve this, now the mock config is not always on responseConfig
-    template: cancelTransaction,
+    template: cancelTransaction, // To test biometrics error, change it to biometricsErrorMobile/biometricsErrorDesktop
```
pradeepdewda-okta, line 54: add the comment as a new line on top

src/v2/controllers/FormController.js
```diff
@@ -222,6 +222,8 @@ export default Controller.extend({
           // the response reaches here when Okta Verify is not installed
           // we need to return an idx object so that
           // the SIW can proceed to the next step without showing error
+          // Also, if the received error is biometrics required error,
```
pradeepdewda-okta, line 225: remove

src/v2/view-builder/views/ov/ChallengeOktaVerifyFastPassView.js
```diff
@@ -130,36 +125,14 @@ const Body = BaseFormWithPolling.extend(Object.assign(
...
-          className: 'okta-verify-uv-callout-content',
-          title: loc('oie.authenticator.oktaverify.method.fastpass.verify.enable.biometrics.title', 'login'),
-          subtitle: loc('oie.authenticator.oktaverify.method.fastpass.verify.enable.biometrics.description', 'login'),
-          bullets: bulletPoints,
-        };
-        this.showMessages(createCallout(options));
-        return true;
+      const options = getBiometricsErrorOptions(error, false);
+      
+      if (_.isEmpty(options)) {
```
pradeepdewda-okta, line 130: in which case it will be empty ?

src/v2/view-builder/views/TerminalView.js
yannongli-okta: @pradeepdewda-okta `getBiometricsErrorOptions ` is also being used in https://github.com/okta/okta-signin-widget/pull/2372/files#diff-c1cf614d130adc2fe0fe8b6d85d6f33b9777027a281f901b701cd97e2ae3cea8R128 as form error. I don't see a way to implement it as a separate view and make it inconsistent across different usage.

playground/mocks/spec-okta-api/idp/idx/index.js
yannongli-okta: @pradeepdewda-okta this change is related to below template change. I would like to inform later developers how to test the flow in playground. Is there any existing way that i can make it work without adding a comment?

src/v2/view-builder/views/ov/ChallengeOktaVerifyFastPassView.js
yannongli-okta: it shouldn't be empty as of current usages. This is more like a safety check for future usage.

pradeepdewda-okta: you can remove this..

playground/mocks/spec-okta-api/idp/idx/index.js
pradeepdewda-okta: you can put right above the `template` variable and say use this file for other two errors..
```
error-okta-verify-uv-fastpass-verify-enable-biometrics-mobile
error-okta-verify-uv-fastpass-verify-enable-biometrics-desktop
```

src/v2/view-builder/views/TerminalView.js
pradeepdewda-okta: @yannongli-okta what's the issue if you create a new view here ?

yannongli-okta: @pradeepdewda-okta so after creating this view (let's call it `OktaFastPassBiometricsErrorView`), it will be used in:
1. TerminalView for SSO extension flow, as `this.add(OktaFastPassBiometricsErrorView)` under `showMessage` method.
2. ChallengeOktaVerifyFastPassView for all other flows, as `this.add(OktaFastPassBiometricsErrorView)` under `showCustomFormErrorCallout`
Is above what you recommended? I'm not sure if this is feasible technically hence asked.

yannongli-okta: also, since we are consuming different type of objects for SSO extension and other methods, how can we get the actual error for view template data? I saw OTPInformationTerminalView is using appState but `showCustomFormErrorCallout` is relying on the `error` object passed from above layer of code.

pradeepdewda-okta: did you try passing the data in your view when you add it.. 

 `this.add(new OktaFastPassBiometricsErrorView( { data })`

pradeepdewda-okta: @jhe-okta should help you from here..

src/v2/view-builder/views/ov/ChallengeOktaVerifyFastPassView.js
yannongli-okta: Correction: if the error is not biometrics error, options will be empty so we need to keep this change.

yannongli-okta:
@pradeepdewda-okta @jhe-okta Pls take a look again

src/v2/view-builder/utils/ChallengeViewUtil.js
```diff
@@ -186,3 +186,11 @@ export function getBiometricsErrorOptions(error, isMessageObj) {
 
   return options;
 }
+
+export const OktaFastPassBiometricsErrorView = View.extend({
```
pradeepdewda-okta, line 190: OktaFastPassBiometricsErrorView change to `OktaFastPassBiometricsTerminalView`

pradeepdewda-okta (approved):
Looks perfect now.. thanks for making the changes.!! 🏅

jhe-okta (approved):

src/v2/view-builder/utils/ChallengeViewUtil.js
```diff
@@ -142,3 +144,45 @@ export function doChallenge(view, fromView) {
...
+  }
+
+  const isBiometricsRequiredMobile = errorSummaryKeys 
+      && errorSummaryKeys.includes(OV_UV_ENABLE_BIOMETRICS_FASTPASS_MOBILE);
+  const isBiometricsRequiredDesktop = errorSummaryKeys 
+      && errorSummaryKeys.includes(OV_UV_ENABLE_BIOMETRICS_FASTPASS_DESKTOP);
+  let options = [];
+
+  if (!isBiometricsRequiredMobile && !isBiometricsRequiredDesktop) {
+    return  options;
```
jhe-okta, line 163: nit: one space only in between

jhe-okta (approved):
Looks good to me. Please add some comments to clarify the difference between the /verify api's 401 and 400 response, as well as `ChallengeOktaVerifyView` vs `DeviceChallengePollView` and why only the former needs to show custom error.

yannongli-okta:
In Credential SSO Extension flow, there are following scenarios:
1. If the /verify endpoint returns a 401, we know that SSO extension **is not installed/not working** and thus we need to step into https://github.com/okta/okta-signin-widget/blob/master/src/v2/controllers/FormController.js#L218 so that no errors will be shown and we can step into next step/next view.
2. If the /verify endpoint returns a 400 (or some other status code), that means that SSO extension is working but there are other errors happening. In these case, we do want to show the error in the view and it's `ChallengeOktaVerifySSOExtensionView.js` for biometrics required error
For this change, since the error only happens at authentication/assurance step, we won't make change to `SSOExtensionView` or `DeviceChallengePollView`.

PR #2380, opened 2022-02-01
fix: clicking "Back to sign in" on "/signin/forgot-password"

sherizada-okta:
## Description:
Fixes clicking "Back to sign in" on the "Reset your password" view at the "/signin/forgot-password" route.

When "Back to sign in" is clicked and the window path is "/signin/forgot-password", clear the path so the router routes to the sign in form.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-458745](https://oktainc.atlassian.net/browse/OKTA-458745)

### DIFF
--- a/src/v2/controllers/FormController.js
+++ b/src/v2/controllers/FormController.js
@@ -153,12 +153,15 @@ export default Controller.extend({
       // TODO: OKTA-243167 what's the approach to show spinner indicating API in flight?
       actionFn()
         .then((resp) => {
-          if (actionPath === 'cancel' && window.location.pathname === '/signin/forgot-password') {
-            window.location = window.location.origin;
-          } else if (actionPath === 'cancel' && this.options.settings.get('useInteractionCodeFlow')) {
-            // In this case we need to restart login flow and recreate transaction meta
-            // that will be used in interactionCodeFlow function
-            this.options.appState.trigger('restartLoginFlow');
+          const { settings, appState } = this.options;
+          if (actionPath === 'cancel') {
+            if (settings.get('useInteractionCodeFlow')) {
+              // embedded
+              appState.trigger('restartLoginFlow');
+            } else {
+              // federated
+              window.location.assign(window.location.origin);
+            }
           } else {
             this.handleIdxResponse(resp);
           }
--- a/src/v2/controllers/FormController.js
+++ b/src/v2/controllers/FormController.js
@@ -153,7 +153,9 @@ export default Controller.extend({
       // TODO: OKTA-243167 what's the approach to show spinner indicating API in flight?
       actionFn()
         .then((resp) => {
-          if (actionPath === 'cancel' && this.options.settings.get('useInteractionCodeFlow')) {
+          if (actionPath === 'cancel' && window.location.pathname === '/signin/forgot-password') {
+            window.location = window.location.origin;
+          } else if (actionPath === 'cancel' && this.options.settings.get('useInteractionCodeFlow')) {
             // In this case we need to restart login flow and recreate transaction meta
             // that will be used in interactionCodeFlow function
             this.options.appState.trigger('restartLoginFlow');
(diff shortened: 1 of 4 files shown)


src/v2/controllers/FormController.js
```diff
@@ -153,7 +153,9 @@ export default Controller.extend({
       // TODO: OKTA-243167 what's the approach to show spinner indicating API in flight?
       actionFn()
         .then((resp) => {
-          if (actionPath === 'cancel' && this.options.settings.get('useInteractionCodeFlow')) {
+          if (actionPath === 'cancel' && window.location.pathname === '/signin/forgot-password') {
```
aarongranick-okta, line 156: Maybe this should just be based on `useInteractionCodeFlow` ? Then `Back to Sign in` will go to the origin even if this flow is running on a different URL (like authorize endpoint).

```
const { settings, appState } = options;
if (actionPath === 'cancel') }
  if (settings.get('useInteractionCodeFlow')) {
      // embedded
      appState.trigger('restartLoginFlow');
  } else {
     // federated
      window.location.assign(window.location.origin);
  }
}

```

test/testcafe/spec/IdentifyWithPassword_spec.js
```diff
@@ -153,16 +155,25 @@ test.requestHooks(identifyWithPasswordMock)('should add sub labels for Username
...
   await t.expect(page.form.getTitle()).eql('Reset your password');
   await t.expect(page.form.getErrorBoxText()).eql('Forgot password is not enabled for this organization.');
 });
+
+test.requestHooks(identifyWithPasswordMock)('should navigate back when "Back to sign in" clicked on /signin/forgot-password', async t => {
+  const page = new IdentityRecoverPageObject(t);
+  await page.navigateToPage();
+  await t.expect(page.form.getTitle()).eql('Reset your password');
+  await page.clickSignOutLink();
+  await t.expect(page.form.getTitle()).eql('Sign In');
```
aarongranick-okta, line 178: add an assertion for the browser URL

aarongranick-okta (approved):
just a couple comments

aarongranick-okta:
Rebase against `6.0` for weekly patch release

src/v2/controllers/FormController.js
sherizada-okta: @aarongranick-okta then every 'cancel' will be a page reload

pradeepdewda-okta, line 156: Would this be removed once backend change is merged as you indicated in Jira ? if yes then put a TODO here..

aarongranick-okta: Isn't that what we need to get a new stateToken

src/v2/controllers/FormController.js
```diff
@@ -153,10 +153,15 @@ export default Controller.extend({
       // TODO: OKTA-243167 what's the approach to show spinner indicating API in flight?
       actionFn()
         .then((resp) => {
-          if (actionPath === 'cancel' && this.options.settings.get('useInteractionCodeFlow')) {
-            // In this case we need to restart login flow and recreate transaction meta
-            // that will be used in interactionCodeFlow function
-            this.options.appState.trigger('restartLoginFlow');
+          const { settings, appState } = this.options;
+          if (actionPath === 'cancel') {
```
pradeepdewda-okta, line 157: on `cancel` we never hit `this.handleIdxResponse` now.. seems risky change to me. Just verify once in bacon first.

sherizada-okta: Yea, and it would clear any url parameters which may be needed.. I pushed primarily so I could test this in okta-core.

sherizada-okta: SIW will call what `handleIdxResponse` does when we cancel because of the page reload

PR #2412, opened 2022-02-28
Resolves: removing optional attributes from the model

vicentevillegas-okta:
## Description:
This change is to fix [bug](https://oktainc.atlassian.net/browse/OKTA-470927), the problem is that when having optional attributes if we add a value then we remove that value, the model will send the attribute as part of the request. As part of the default behavior we shouldn't send empty attributes as part of the request. This change will remove empty optional attributes from the model.

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Reviewers:
@pradeepdewda-okta 
@okta/ciamx 

### Issue:
- [OKTA-470927](https://oktainc.atlassian.net/browse/OKTA-470927)


src/v2/view-builder/views/EnrollProfileView.js
```diff
@@ -29,6 +29,16 @@ const Body = BaseForm.extend({
 
   },
   saveForm() {
+    const uiSchema = this.getUISchema();
+    _.each(this.model.attributes, (value, name) => {
+      if (_.isEmpty(value)) {
+        const uiSchemaProperty = uiSchema.find(schema => schema.name === name);
+        if (!_.isUndefined(uiSchemaProperty) && !uiSchemaProperty.required) {
```
pradeepdewda-okta, line 36: instead of deleting all required empty attributes, let's make it specific to which element you must delete if it is empty..

Also, we can make a generalized function in `AppState.js` and use it in this view... 

```
cleanModelState(attr) {
  // sanitize or delete here, attr = []
}
```

what to do you think ?

vicentevillegas-okta: Hello @pradeepdewda-okta, 

I like your idea about making a generalized function in `AppState.js`.

Regarding the first comment, I am deleting all non required attributes (optional attributes) from the model. At the beginning I thought this problem was exclusive for attributes rendered as dropdown, however after a lot of tests and checking other existing bugs I figure out that this problem is happening for attributes rendered as Text box too.

What I can do is to add an additional check so this deletion can happen only for dropdown and textbox.

src/v2/view-builder/views/EnrollProfileView.js
```diff
@@ -29,15 +29,8 @@ const Body = BaseForm.extend({
...
-    _.each(this.model.attributes, (value, name) => {
-      if (_.isEmpty(value)) {
-        const uiSchemaProperty = uiSchema.find(schema => schema.name === name);
-        if (!_.isUndefined(uiSchemaProperty) && !uiSchemaProperty.required) {
-          delete this.model.attributes[name];
-        }
-      }
-    });
+    // cleaning up model from empty optional attributes
+    this.options.appState.cleanModelState(this.model.attributes);
```
pradeepdewda-okta, line 33: you need not to pass attributes here.. in AppState, it should be available automatically.

src/v2/models/AppState.js
```diff
@@ -328,6 +328,24 @@ export default Model.extend({
...
+   */
+  cleanModelState(attributes) {
+    const currentViewState = this.getCurrentViewState();
+    if(currentViewState) {
+      const uiSchema = currentViewState.uiSchema;
+      _.each(attributes, (value, name) => {
+        if (_.isEmpty(value)) {
+          const uiSchemaProperty = uiSchema.find(schema => schema.name === name);
+          if (!_.isUndefined(uiSchemaProperty) && !uiSchemaProperty.required &&
+            (uiSchemaProperty.type === 'select' || uiSchemaProperty.type === 'text')) {
```
pradeepdewda-okta, line 344: I feel ui specific deletion should not be done.. It should be strictly based on `model`. So let's remove 
`(uiSchemaProperty.type === 'select' || uiSchemaProperty.type === 'text')`

and keep it to delete all non-required empty but defined attributes..

At least now we have a function that clearly says what it does. We can extend its functionality in future

pradeepdewda-okta: you can ignore above... just change the function name to `cleanViewModelState`

src/v2/models/AppState.js
```diff
@@ -328,6 +328,24 @@ export default Model.extend({
     const isCurrentAuthenticatorEmail = this.get('currentAuthenticatorEnrollment')?.key === AUTHENTICATOR_KEY.EMAIL;
 
     return isSameExceptMessages && isChallengeAuthenticator && isCurrentAuthenticatorEmail;
-  }
+  },
 
+  /**
+   * It will remove empty optional attributes from the model for attributes rendered as dropdown and text box.
```
pradeepdewda-okta, line 334: It will remove empty optional attributes from the view model of current view state.... `

pradeepdewda-okta:
@vicentevillegas-okta also add `profile-enrollment-string-fields-options` to the responseConfig.js mock list

src/v2/view-builder/views/EnrollProfileView.js
pradeepdewda-okta: you can ignore above... just change the function name to cleanViewModelState

pradeepdewda-okta:
@vicentevillegas-okta you can try following to override toJSON instead..

```
    ModelClass = ModelClass.extend({
      toJSON: () => {
        const modelJSON = Model.prototype.toJSON.call(this, arguments);
        console.log(this);
        // remove unwanted data from modelJSON
        return modelJSON;
      }
    });
   ```

File: test/testcafe/spec/ProfileEnrollmentWithStringOptions_spec.js [hunk 5]
```diff
@@ -50,4 +53,29 @@ test.requestHooks(requestLogger, ProfileEnrollmentSignUpWithStringFieldsMock)('s
...
+});
+
+test.requestHooks(requestLogger, ProfileEnrollmentSignUpWithStringFieldsMock)('should submit form when all optional fields are empty', async t => {
+  const profileEnrollmentString = new ProfileEnrollmentStringOptionsViewPageObject(t);
+  const identityPage = await setup(t);
+  await identityPage.clickSignUpLink();
+
+  await profileEnrollmentString.fillEmailField('test.carlos@mycompany.com');
+  await profileEnrollmentString.fillFirstNameField('Test Carlos');
+  await profileEnrollmentString.fillLastNameField('Test');
```
pradeepdewda-okta, line 65: can we add a field here with empty value ?

src/v2/view-builder/views/EnrollProfileView.js
```diff
@@ -61,8 +61,20 @@ export default BaseView.extend({
...
     const currentSchema = JSON.parse(JSON.stringify((currentViewState.uiSchema)));
+    const appState = this.options.appState;
+    let ModelClass = BaseView.prototype.createModelClass.apply(this, arguments, appState);
+
+    ModelClass = ModelClass.extend({
+      toJSON: function() {
+        const modelJSON = Model.prototype.toJSON.call(this, arguments, appState);
+        // remove unwanted data from modelJSON
+        if(modelJSON.userProfile) {
+          appState.cleanViewModelState(modelJSON.userProfile);
```
pradeepdewda-okta, line 73: we don't need `cleanViewModelState` anymore if we are overriding `toJSON`.. we can simply just delete empty optional attributes.

if we are adding `userProfile` in `cleanViewModelState` then this function is not generalized and not serving the purpose of keeping it in AppState.js

test/testcafe/spec/ProfileEnrollmentWithStringOptions_spec.js
vicentevillegas-okta: I was re-checking my test and trying to see what else I was missing... but you found it!! you have a very good eye for details, thank you so much @pradeepdewda-okta

src/v2/view-builder/views/EnrollProfileView.js
```diff
@@ -61,8 +61,28 @@ export default BaseView.extend({
...
   Footer,
   createModelClass(currentViewState, optionUiSchemaConfig, settings) {
-    let ModelClass = BaseView.prototype.createModelClass.apply(this, arguments);
     const currentSchema = JSON.parse(JSON.stringify((currentViewState.uiSchema)));
+    let ModelClass = BaseView.prototype.createModelClass.apply(this, arguments, currentViewState);
+
+    ModelClass = ModelClass.extend({
+      toJSON: function() {
+        const modelJSON = Model.prototype.toJSON.call(this, arguments, currentViewState);
+        // remove unwanted data from modelJSON
```
pradeepdewda-okta, line 70: change the comment to `delete optional attributes if they are empty and not required`

pradeepdewda-okta (approved):
after few nits.. 🚢

vicentevillegas-okta:
image

fangshi-okta (approved):

justinabrokwah-okta (approved):
Looks great 🚀

PR #2413, opened 2022-03-01
fs-okta-472423-progressive-profiling-country-code-siw

fangshi-okta:
## Description:

1. if get display type as `select_country_code`, render the field as a dropdown with predefined list
2. else render as a text box for country code

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Reviewers:
@vicentevillegas-okta 

### Issue:

- [OKTA-472423](https://oktainc.atlassian.net/browse/OKTA-472423)

### DIFF
--- a/src/v2/ion/ui-schema/ion-string-handler.js
+++ b/src/v2/ion/ui-schema/ion-string-handler.js
@@ -64,6 +64,8 @@ const populateUISchemaForDisplay = (uiSchema, ionField) => {
     uiSchema.wide = true;
     //it will create a placeholder for dropdowns, by default it will show 'Select an Option'
     uiSchema.options = Object.assign({'': ''}, ionOptionsToUiOptions(display.options));
+  } else if (display.inputType === 'select_country_code') {
+    Object.assign(uiSchema, countryUISchema);
   }
 };
 
@@ -86,10 +88,6 @@ const createUiSchemaForString = (ionFormField, remediationForm, transformedResp,
 
   if (ionFormField.hint === HINTS.CAPTCHA) {
     Object.assign(uiSchema, getCaptchaUiSchema());
-  }
-
-  if(ionFormField.name === 'userProfile.countryCode'){
-    Object.assign(uiSchema, countryUISchema);
   }
 
   if(ionFormField.name === 'userProfile.timezone'){
--- a/test/testcafe/spec/EnrollProfileView_spec.js
+++ b/test/testcafe/spec/EnrollProfileView_spec.js
@@ -79,9 +79,11 @@ test.requestHooks(requestLogger, EnrollProfileSignUpWithAdditionalFieldsMock)('s
   await identityPage.clickSignUpLink();
 
   requestLogger.clear();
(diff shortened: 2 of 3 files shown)


src/v2/ion/ui-schema/ion-string-handler.js
```diff
@@ -88,10 +90,6 @@ const createUiSchemaForString = (ionFormField, remediationForm, transformedResp,
     Object.assign(uiSchema, getCaptchaUiSchema());
   }
 
-  if(ionFormField.name === 'userProfile.countryCode'){
```
pradeepdewda-okta, line 91: why to remove this from `createUiSchemaForString` function ?

fangshi-okta: for our new design, we allow admin to select a dropdown list with predefined values or text box display for all country code data type attribute.

test/testcafe/spec/EnrollProfileView_spec.js
```diff
@@ -79,9 +79,11 @@ test.requestHooks(requestLogger, EnrollProfileSignUpWithAdditionalFieldsMock)('s
   await identityPage.clickSignUpLink();
 
   requestLogger.clear();
+  await t.expect(await enrollProfilePage.getFormFieldLabel('userProfile.country')).eql('Country');
```
pradeepdewda-okta, line 82: also add a test that checks for `textbox` as well ?

fangshi-okta:
> @fangshi-okta are we defaulting timezone to UTC ?

We did not set the default value for timezone. so it picked the first value from the list.

test/testcafe/spec/EnrollProfileView_spec.js
fangshi-okta: test added. Thanks

vicentevillegas-okta:
> > @fangshi-okta are we defaulting timezone to UTC ?
> 
> We did not set the default value for timezone. so it picked the first value from the list. We might revisit timezone data type if we have any new design for it in the future. Thanks

@fangshi-okta what will happen if Country code or timezone are defined as `optional` in the enrollment form? I think for this case we are always storing the first value from the list (as you mentioned), so it behaves like a required attribute. I mean, the final user has not the ability to not select a value, the value is always selected by default. (from my personal point of view this is not a good user experience, I think we should be consistence and fix it)

fangshi-okta:
> > > @fangshi-okta are we defaulting timezone to UTC ?
> > 
> > 
> > We did not set the default value for timezone. so it picked the first value from the list. We might revisit timezone data type if we have any new design for it in the future. Thanks
> 
> @fangshi-okta what will happen if Country code or timezone are defined as `optional` in the enrollment form? I think for this case we are always storing the first value from the list (as you mentioned), so it behaves like a required attribute. I mean, the final user has not the ability to not select a value, the value is always selected by default. (from my personal point of view this is not a good user experience, I think we should be consistence and fix it)

The default value was suggested by PM. but you are right that user can not leave it as empty even though it is an optional. we can bring up this to PM office hour and to see if we want to make the change. And I will create a follow up ticket for that if needed?

pradeepdewda-okta (approved):

PR #2422, opened 2022-03-08
Resolves: OKTA-476273 - Fixing SIW hanging scrollbar when we have a dropdown at the very end

vicentevillegas-okta:
## Description:
When having a dropdown attribute to the very end of the SIW and clicking it, the SIW main container shows a vertical scrollbar additionally to the scrollbar shown by the page body. 

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Normal devices, left image shows the bug and the right image shows after fixing the bug.
SIW_5

Small devices, left image shows the bug and the right image shows after fixing the bug.
SIW_6

### Reviewers:
@okta/ciamx 

### Issue:
- [OKTA-476273](https://oktainc.atlassian.net/browse/OKTA-476273)

### DIFF
--- a/assets/sass/_container.scss
+++ b/assets/sass/_container.scss
@@ -8,7 +8,7 @@
   background-color: $secondary-bg-color;
   color: $medium-text-color;
   position: relative;
-  overflow: auto;
+  overflow: hidden;
 
   // Auth-container styles
   border-radius: 3px;


justinabrokwah-okta:
> When having a dropdown attribute to the very end of the SIW and clicking it, the SIW main container shows a vertical scrollbar additionally to the scrollbar shown by the page body.
If the dropdown attribute isn't at the end of the SIW, is the behaviour still the same? (The scrollbar being present)

jakebacher2-okta:
tagging @narayun-okta for design verification

vicentevillegas-okta:
> > When having a dropdown attribute to the very end of the SIW and clicking it, the SIW main container shows a vertical scrollbar additionally to the scrollbar shown by the page body.
> > If the dropdown attribute isn't at the end of the SIW, is the behaviour still the same? (The scrollbar being present)

As far as I have seeing and tested, the behavior will be still the same (The scrollbar is not present)

jakebacher2-okta:
Have you tested this with dropdown that comes from an enum with many values? The fix definitely works for the country code, but we use a different dropdown for that

vicentevillegas-okta:
> Have you tested this with dropdown that comes from an enum with many values? The fix definitely works for the country code, but we use a different dropdown for that

You are right! they are different dropdowns. This is an excellent catch @jakebacher2-okta, I am just seeing that for those dropdowns the fix is not working :fist_right:

Thank you so much!!

vicentevillegas-okta:
> > Have you tested this with dropdown that comes from an enum with many values? The fix definitely works for the country code, but we use a different dropdown for that
> 
> You are right! they are different dropdowns. This is an excellent catch @jakebacher2-okta, I am just seeing that for those dropdowns the fix is not working 🤜
> 
> Thank you so much!!

Wait a sec... I tested this scenario without the fix :disappointed: ... it is working fine with this type of dropdown too.

jakebacher2-okta (approved):
LGTM but i think you still need a codeowner to take a look

justinabrokwah-okta (approved):

lesterchoi-okta (approved):

PR #2423, opened 2022-03-09
fs-477714-progressive-profiling-string-with-format

fangshi-okta:
## Description:

- if get display type as select and format is country-code, render the field as a dropdown with predefined list
- else render as a text box 

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

Screen Shot 2022-03-01 at 1 13 17 PM

Screen Shot 2022-03-09 at 9 03 31 AM

### Reviewers:

@okta/ciamx 

### Issue:

- [OKTA-477714](https://oktainc.atlassian.net/browse/OKTA-477714)

### DIFF
--- a/playground/mocks/data/idp/idx/enroll-profile-new-additional-fields.json
+++ b/playground/mocks/data/idp/idx/enroll-profile-new-additional-fields.json
@@ -44,7 +44,8 @@
                       "value":{
                         "type":"object",
                         "value":{
-                          "inputType":"select_country_code"
+                          "format": "country-code",
+                          "inputType":"select"
                         }
                       }
                     }
@@ -61,6 +62,7 @@
                       "value":{
                         "type":"object",
                         "value":{
+                          "format": "country-code",
                           "inputType":"text"
                         }
                       }
--- a/src/v2/ion/ui-schema/ion-string-handler.js
+++ b/src/v2/ion/ui-schema/ion-string-handler.js
@@ -13,7 +13,7 @@
 /* eslint max-depth: [2, 3] */
 /* eslint complexity: [2, 20] */
 import { loc } from 'okta';
-import { HINTS } from '../RemediationConstants';
+import { HINTS, ATTR_FORMAT } from '../RemediationConstants';
 import CountryUtil from '../../../util/CountryUtil';
 import TimeZone from '../../view-builder/utils/TimeZone';
 
@@ -55,7 +55,7 @@ const populateUISchemaForDisplay = (uiSchema, ionField) => {
     uiSchema.options = display.options;
   } else if (display.inputType === 'select') {
     uiSchema.wide = true;
-    if (display.format && display.format === 'country-code') {
+    if (display.format && display.format === ATTR_FORMAT.COUNTRY_CODE) {
       uiSchema.options = CountryUtil.getCountryCode();
       uiSchema.value = 'US';
     } else {
--- a/src/v2/ion/ui-schema/ion-string-handler.js
+++ b/src/v2/ion/ui-schema/ion-string-handler.js
@@ -60,12 +60,12 @@ const populateUISchemaForDisplay = (uiSchema, ionField) => {
   uiSchema.type = display.inputType;
   if (display.inputType === 'radio') {
     uiSchema.options = display.options;
+  } else if (display.inputType === 'select' && display.format === 'country-code') {
+    Object.assign(uiSchema, countryUISchema);
   } else if (display.inputType === 'select') {
     uiSchema.wide = true;
     //it will create a placeholder for dropdowns, by default it will show 'Select an Option'
     uiSchema.options = Object.assign({'': ''}, ionOptionsToUiOptions(display.options));
-  } else if (display.inputType === 'select_country_code') {
-    Object.assign(uiSchema, countryUISchema);
   }
 };
 
--- a/src/v2/ion/ui-schema/ion-string-handler.js
+++ b/src/v2/ion/ui-schema/ion-string-handler.js
@@ -55,7 +55,7 @@ const populateUISchemaForDisplay = (uiSchema, ionField) => {
     uiSchema.options = display.options;
   } else if (display.inputType === 'select') {
     uiSchema.wide = true;
-    if (display.format && display.format === ATTR_FORMAT.COUNTRY_CODE) {
+    if (display.format === ATTR_FORMAT.COUNTRY_CODE) {
       uiSchema.options = CountryUtil.getCountryCode();
       uiSchema.value = 'US';
     } else {
--- a/src/v2/ion/ui-schema/ion-string-handler.js
+++ b/src/v2/ion/ui-schema/ion-string-handler.js
@@ -60,12 +60,14 @@ const populateUISchemaForDisplay = (uiSchema, ionField) => {
   uiSchema.type = display.inputType;
   if (display.inputType === 'radio') {
     uiSchema.options = display.options;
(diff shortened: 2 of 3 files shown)


playground/mocks/data/idp/idx/enroll-profile-new-additional-fields.json
```diff
@@ -61,6 +62,7 @@
                       "value":{
                         "type":"object",
                         "value":{
+                          "format": "country-code",
                           "inputType":"text"
```
justinabrokwah-okta, line 66: Just curious, why are we allowing text as an input for country-code?

src/v2/ion/ui-schema/ion-string-handler.js
```diff
@@ -60,12 +60,12 @@ const populateUISchemaForDisplay = (uiSchema, ionField) => {
   uiSchema.type = display.inputType;
   if (display.inputType === 'radio') {
     uiSchema.options = display.options;
+  } else if (display.inputType === 'select' && display.format === 'country-code') {
```
vicentevillegas-okta, line 63: nit: I think the format wont be present for all attributes, wondering if we should make sure it is not empty, perhaps something like this:
`_.isNotEmpty(display.format)`

playground/mocks/data/idp/idx/enroll-profile-new-additional-fields.json
fangshi-okta: Good question. because we need to store admin's settings. every time after admin choose `text` display, next time, when admin comes back to the settings, the `text` option can be remembered. Previously, we do not add it as it is not needed in SIW and just an default value in ui. but for country code, it is one of the radio option in ui.

fangshi-okta:
> Can you add a screenshot of you opening the dropdown?

screenshot added. thanks

src/v2/ion/ui-schema/ion-string-handler.js
fangshi-okta: updated. Good call

src/v2/ion/ui-schema/ion-string-handler.js
```diff
@@ -60,12 +60,12 @@ const populateUISchemaForDisplay = (uiSchema, ionField) => {
   uiSchema.type = display.inputType;
   if (display.inputType === 'radio') {
     uiSchema.options = display.options;
+  } else if (display.inputType === 'select' && !_.isEmpty(display.format) && display.format === 'country-code') {
```
pradeepdewda-okta, line 63: if it's a `select` why we need another `else if` here ?

playground/mocks/data/idp/idx/enroll-profile-new-additional-fields.json
justinabrokwah-okta: Oh okay, so it's another option we support then. Got it!

src/v2/ion/ui-schema/ion-string-handler.js
fangshi-okta: updated. thanks

src/v2/ion/ui-schema/ion-string-handler.js
```diff
@@ -61,11 +61,13 @@ const populateUISchemaForDisplay = (uiSchema, ionField) => {
   if (display.inputType === 'radio') {
     uiSchema.options = display.options;
   } else if (display.inputType === 'select') {
-    uiSchema.wide = true;
-    //it will create a placeholder for dropdowns, by default it will show 'Select an Option'
-    uiSchema.options = Object.assign({'': ''}, ionOptionsToUiOptions(display.options));
-  } else if (display.inputType === 'select_country_code') {
-    Object.assign(uiSchema, countryUISchema);
+    if(!_.isEmpty(display.format) && display.format === 'country-code'){
+      Object.assign(uiSchema, countryUISchema);
```
vicentevillegas-okta, line 65: nit: wondering if we can do something like this
image

src/v2/ion/ui-schema/ion-string-handler.js
```diff
@@ -61,11 +61,13 @@ const populateUISchemaForDisplay = (uiSchema, ionField) => {
   if (display.inputType === 'radio') {
     uiSchema.options = display.options;
   } else if (display.inputType === 'select') {
-    uiSchema.wide = true;
-    //it will create a placeholder for dropdowns, by default it will show 'Select an Option'
-    uiSchema.options = Object.assign({'': ''}, ionOptionsToUiOptions(display.options));
-  } else if (display.inputType === 'select_country_code') {
-    Object.assign(uiSchema, countryUISchema);
+    if(!_.isEmpty(display.format) && display.format === 'country-code'){
```
pradeepdewda-okta, line 64: what's special about `countryUISchema` ? it still does have same `type` and `wide` attributes. The only change is `options` and `value`..

i think it's time we can eliminate `countryUISchema` and just update `options` and `value`

src/v2/ion/ui-schema/ion-string-handler.js
fangshi-okta: good suggestion. maybe we can implement this when we decide to do the same pattern for timezone and language code. thanks!

src/v2/ion/ui-schema/ion-string-handler.js
fangshi-okta: updated based on the suggestion. Thank you!

vicentevillegas-okta (approved):
I have tested locally okta-core + okta-ui + SIW, so these changes are working fine.

LGTM 👍 

I think you still need to get approval from a codeowner

Thank you @fangshi-okta for these changes!

src/v2/ion/ui-schema/ion-string-handler.js
```diff
@@ -62,10 +55,13 @@ const populateUISchemaForDisplay = (uiSchema, ionField) => {
     uiSchema.options = display.options;
   } else if (display.inputType === 'select') {
     uiSchema.wide = true;
-    //it will create a placeholder for dropdowns, by default it will show 'Select an Option'
-    uiSchema.options = Object.assign({'': ''}, ionOptionsToUiOptions(display.options));
-  } else if (display.inputType === 'select_country_code') {
-    Object.assign(uiSchema, countryUISchema);
+    if (display.format && display.format === 'country-code') {
```
pradeepdewda-okta, line 58: `display?.format === 'country-code'`

pradeepdewda-okta: add `country-code` as constant instead of a string.. Add it to `RemediationConstants.js`

fangshi-okta: Thanks for the suggestions! I think display value has been checked in `const display = ionField?.value?.value;` here I just want to check that `format` is not empty. country-code was added to RemediationConstants.js

pradeepdewda-okta: can `format` be empty ?

fangshi-okta: yeah only for `country-code` data type attribute, the format will contains value. any concerns?

pradeepdewda-okta: want to confirm `display.format === ''` is possible ? if not then `display?.format === 'country-code'` should be enough right ?

fangshi-okta: updated based on our discussion. Thanks!

pradeepdewda-okta (approved):

pradeepdewda-okta:
Add a unit test and then merge.

PR #2430, opened 2022-03-13
Differentiate user and loopback cancel

jhe-okta:
## Description:

Pass `triggeredByUser` value to the polling cancel request in order to differentiate loopback failure vs user cancelation.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

#### loopback fails, SIW cancels the polling
Screen Shot 2022-03-10 at 5 33 47 PM

#### user clicks on cancel and go back link
Screen Shot 2022-03-10 at 5 43 48 PM

### Reviewers:
@pradeepdewda-okta @santoshmale-okta 

### Issue:

- [OKTA-469749](https://oktainc.atlassian.net/browse/OKTA-469749)

### DIFF
--- a/src/v2/view-builder/utils/Constants.js
+++ b/src/v2/view-builder/utils/Constants.js
@@ -11,4 +11,4 @@ export const OV_UV_ENABLE_BIOMETRICS_FASTPASS_DESKTOP
     = 'oie.authenticator.oktaverify.method.fastpass.verify.enable.biometrics.desktop';
 export const OV_UV_ENABLE_BIOMETRICS_FASTPASS_MOBILE 
     = 'oie.authenticator.oktaverify.method.fastpass.verify.enable.biometrics.mobile';
-
+export const REQUEST_PARAM_LOOPBACK_CANCEL_TRIGGER = 'triggeredByUser';
--- a/src/v2/view-builder/views/device/DeviceChallengePollView.js
+++ b/src/v2/view-builder/views/device/DeviceChallengePollView.js
@@ -3,7 +3,8 @@ import {BaseFormWithPolling, BaseFooter, BaseView} from '../../internals';
 import Logger from '../../../../util/Logger';
 import BrowserFeatures from '../../../../util/BrowserFeatures';
 import Enums from '../../../../util/Enums';
-import { CANCEL_POLLING_ACTION, CHALLENGE_TIMEOUT, IDENTIFIER_FLOW } from '../../utils/Constants';
+import { CANCEL_POLLING_ACTION, CHALLENGE_TIMEOUT, IDENTIFIER_FLOW, REQUEST_PARAM_LOOPBACK_CANCEL_TRIGGER }
+  from '../../utils/Constants';
 import Link from '../../components/Link';
 import { doChallenge } from '../../utils/ChallengeViewUtil';
 import OktaVerifyAuthenticatorHeader from '../../components/OktaVerifyAuthenticatorHeader';
@@ -15,6 +16,12 @@ const request = (opts) => {
     contentType: 'application/json',
   }, opts);
   return $.ajax(ajaxOptions);
+};
+
+const cancelPolling = (appState, triggeredByUser = false) => {
+  const actionParams = {};
+  actionParams[REQUEST_PARAM_LOOPBACK_CANCEL_TRIGGER] = triggeredByUser;
+  appState.trigger('invokeAction', CANCEL_POLLING_ACTION, actionParams);
 };
 
 const Body = BaseFormWithPolling.extend(
@@ -102,7 +109,7 @@ const Body = BaseFormWithPolling.extend(
         countFailedPorts++;
         if (countFailedPorts === ports.length) {
           Logger.error('No available ports. Loopback server failed and polling is cancelled.');
-          this.options.appState.trigger('invokeAction', CANCEL_POLLING_ACTION);
+          cancelPolling(this.options.appState, false);
         }
       };
 
@@ -187,7 +194,9 @@ const Footer = BaseFooter.extend({
         options: {
           name: 'cancel-authenticator-challenge',
           label: loc('loopback.polling.cancel.link', 'login'),
-          actionPath: CANCEL_POLLING_ACTION,
+          clickHandler: () => {
+            cancelPolling(this.options.appState, true);
+          },
         }
       }).last();
     }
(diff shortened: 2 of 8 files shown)


src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -17,6 +18,12 @@ const request = (opts) => {
   return $.ajax(ajaxOptions);
 };
 
+const cancelPolling = (appState, triggeredByUser = false) => {
```
pradeepdewda-okta, line 21: make `cancelPolling` part of `polling.js`.. it has access to `this.option.Appstate`... 

`actionParams` can be part as args.. 
```
cancelPolling(actionParams) {
this.stopPolling(); // check if this is needed?? to stopPoll before trigger cancel 

this.options.appState.trigger('invokeAction', 'cancel', actionParams);
}
```

jhe-okta: @pradeepdewda-okta , `stopPolling` is to stop/remove the ongoing `polling` request on the page, `cancelPolling` is a different API which responds with a different remediation. When this new remediation is returned, the current UI will be removed and `stopPolling` will be invoked. I know the naming is confusing, `cancelPolling` is simply named after its API URL. We can name it something else, such as `stopCurrentAuthentication`. 
In regarding of moving `cancelPolling` to polling.js, I can think about that and potentially address it in a separate JIRA. We have another [JIRA](https://oktainc.atlassian.net/browse/OKTA-390965) for refactoring that's targeted for next week.

pradeepdewda-okta: >When this new remediation is returned, the current UI will be removed and stopPolling will be invoked.

Nice that make sense.

It seems like a polling action.. so we can rename as `invokePollingActionWithParams`  and move it to `polling.js` (in your next PR)

pradeepdewda-okta (approved):

src/v2/view-builder/views/device/DeviceChallengePollView.js
jhe-okta: > > When this new remediation is returned, the current UI will be removed and stopPolling will be invoked.
> 
> Nice that make sense.
> 
> It seems like a polling action.. so we can rename as `invokePollingActionWithParams` and move it to `polling.js` (in your next PR)

Let me give more thought about the naming. Thanks.

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -17,6 +23,12 @@ const request = (opts) => {
   return $.ajax(ajaxOptions);
 };
 
+const cancelPolling = (appState, cancelReason) => {
```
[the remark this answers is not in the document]
jhe-okta: @pradeepdewda-okta renamed it to `cancelPollingWithParams`, as the action we trigger in the function is specific for canceling polling, we just pass in different reasons.

jhe-okta:
@pradeepdewda-okta please review the 2nd commit. The changes are
- instead of passing in `triggeredByUser` with boolean value in the cancel polling request body, when @santoshmale-okta and I did the testing, we identified a 3rd case to track. So now we pass in `reason` to indicate the cause of the loopback cancellation. 
- update the `DeviceChallengePollView.js`(identify flow) to keep it more consistent with how loopback works in `ChallengeOktaVerifyFastPassView.js` (assurance flow). There're still some small differences, but they should not affect the functionality.
Screen Shot 2022-03-16 at 3 52 51 PM
Screen Shot 2022-03-16 at 4 06 26 PM
Screen Shot 2022-03-16 at 3 56 35 PM

src/v2/view-builder/utils/Constants.js
```diff
@@ -11,4 +11,10 @@ export const OV_UV_ENABLE_BIOMETRICS_FASTPASS_DESKTOP
     = 'oie.authenticator.oktaverify.method.fastpass.verify.enable.biometrics.desktop';
 export const OV_UV_ENABLE_BIOMETRICS_FASTPASS_MOBILE 
     = 'oie.authenticator.oktaverify.method.fastpass.verify.enable.biometrics.mobile';
-
+export const REQUEST_PARAM_LOOPBACK_CANCEL_TRIGGER = 'triggeredByUser';
```
anatoliiboiarshchykov-okta, line 14: Can this one be removed, because of AUTHENTICATION_CANCEL_REASONS.USER_CANCELED ?

anatoliiboiarshchykov-okta (approved):

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -98,33 +110,21 @@ const Body = BaseFormWithPolling.extend(
...
-            return this.probingXhr.done(() => {
-              foundPort = true;
-              // log in as soon as challenge request is finished
-              return this.trigger('save', this.model);
-            }).fail((xhr) => {
-              Logger.error(`OV challenge response with HTTP code ${xhr.status} ${xhr.responseText}`);
-              if (xhr.statusText === 'timeout') {
+            return onPortFound()
+              .done(() => {
                 foundPort = true;
```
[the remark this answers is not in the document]
jhe-okta: Ok, thanks @santoshmale-okta for the context. The change in `DeviceChallengePollView` is correct. We should make `ChallengeOktaVerifyFastPassView` constant with it during the refactoring. The context is that there could be multiple profile in the same device, if the unexpected profile responds to the probing call first, loopback may fail later on. So we want to keep probing to find an open port and do challenge with the found port until challenge succeeds or all the ports are exhausted.

src/v2/view-builder/views/device/DeviceChallengePollView.js
```diff
@@ -187,7 +191,9 @@ const Footer = BaseFooter.extend({
         options: {
           name: 'cancel-authenticator-challenge',
           label: loc('loopback.polling.cancel.link', 'login'),
-          actionPath: CANCEL_POLLING_ACTION,
+          clickHandler: () => {
```
yannongli-okta, line 194: my understanding on why we don't have USER_CANCELLED at authentication step:
1. if user clicks Verify with something else, there will be no api call and user just fallback to factor selection page based on previous remediaton `select-authenticator-authenticate`
2. if user clicks cancel in OV, it's OV that will send the request to server so SIW can't do anything until they receive next polling response
If so, pls add a comment here to state the diff so that ppl know this is intentional

PR #2432, opened 2022-03-16
feat: Add usage text to each authenticator listed during enrollment

indirathirumalaimurugan-okta:
## Description:

Add usage related text such as "Used for access/recovery" to each authenticator during enrollment

Note: Changes look good on okta-core. Bacon run on okta-core went through fine with this alpha version.

## PR Checklist

- [ ] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:

**Required step**

Before

Screen Shot 2022-03-17 at 5 22 57 PM

After

Screen Shot 2022-03-17 at 5 17 04 PM

**Optional step**

Before

Optional_before

After

Screen Shot 2022-03-17 at 5 16 02 PM

If for any reason an authenticator have 'allowedFor' set to anything other than 'sso' or 'recovery' or 'any' then the usage text is hidden. This situation is not possible but this is how UI handles just in case

Screen Shot 2022-03-17 at 5 15 41 PM

### Issue:

- [OKTA-477018](https://oktainc.atlassian.net/browse/OKTA-477018)


test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
```diff
@@ -62,87 +67,140 @@ test.requestHooks(mockEnrollAuthenticatorPassword)('should load select authentic
...
+  await t.expect(selectFactorPage.getFactorDescriptionByIndex(2))
+    .eql('Okta Verify is an authenticator app, installed on your phone, used to prove your identity');
+  await t.expect(await selectFactorPage.factorUsageTextExistsByIndex(2)).eql(true);
+  await t.expect(selectFactorPage.getFactorUsageTextByIndex(2)).eql('Used for access and recovery');
+
+  await t.expect(selectFactorPage.getFactorLabelByIndex(3)).eql('Security Question');
+  await t.expect(selectFactorPage.getFactorIconClassByIndex(3)).contains('mfa-okta-security-question');
+  await t.expect(selectFactorPage.getFactorSelectButtonByIndex(3)).eql('Set up');
+  await t.expect(selectFactorPage.getFactorSelectButtonDataSeByIndex(3)).eql('security_question');
+  await t.expect(selectFactorPage.getFactorDescriptionByIndex(3)).eql('Choose a security question and answer that will be used for signing in');
```
pradeepdewda-okta, line 200: also check for `getFactorUsageTextByIndex` ?

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -1137,9 +1137,13 @@ oie.select.authenticators.enroll.subtitle.custom = Security methods help protect
 
 oie.setup.required = Set up required
 oie.setup.optional = Set up optional
-oie.optional.authenticator.button.title = Finish
+oie.optional.authenticator.button.title = Set up later
 oie.verify.authenticator.button.text = Select
 oie.enroll.authenticator.button.text = Set up
+oie.enroll.authenticator.usage.text.access.recovery = Used for access and recovery
+oie.enroll.authenticator.usage.text.recovery = Used for recovery
```
pradeepdewda-okta, line 1144: For enduser it may not be intuitive to understand what is `recovery` or `access` means while signing up or logging in.  @vishwasrikaanth-okta

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
```diff
@@ -42,7 +44,35 @@ const AuthenticatorRow = View.extend({
...
+      case 'any':
+        authenticatorUsageText = loc('oie.enroll.authenticator.usage.text.access.recovery', 'login');
+        break;
+      case 'recovery':
+        authenticatorUsageText = loc('oie.enroll.authenticator.usage.text.recovery', 'login');
+        break;
+      case 'sso':
+        authenticatorUsageText = loc('oie.enroll.authenticator.usage.text.access', 'login');
+        break;
+      default:
```
praveendurai-okta, line 63: do we need a default if there are no default action to be performed?

test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
indirathirumalaimurugan-okta: Security Question authenticator has 'allowedFor' set to 'none' in the remediation form. If 'allowedFor' has anything other than 'sso' or 'recovery' or 'any' then the `.authenticator-usage-text` element itself won't exist. so not checking for its text content here.

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
indirathirumalaimurugan-okta: I don't think we need it. Removed.

packages/@okta/i18n/src/properties/login.properties
indirathirumalaimurugan-okta: concern has been discussed offline and decided to leave the text as is.

praveendurai-okta (approved):
lgtm

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
```diff
@@ -16,19 +16,21 @@ import hbs from 'handlebars-inline-precompile';
...
+    <div class="authenticator-icon-container">
+      <div class="factor-icon authenticator-icon {{iconClassName}}"></div>
+    </div>
+    <div class="authenticator-description">
+      <h3 class="authenticator-label no-translate">{{label}}</h3>
+      {{#if description}}
+        <p class="authenticator-description--text">{{description}}</p>
+      {{/if}}
+      {{#if authenticatorUsageText}}
+        <p class="authenticator-usage-text">{{authenticatorUsageText}}</p>
```
pradeepdewda-okta, line 29: we can use the existing courage `explain` class ?

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
```diff
@@ -42,7 +44,33 @@ const AuthenticatorRow = View.extend({
...
     this.$el.addClass('authenticator-row-min');
-  }
+  },
+  getTemplateData() {
+    let authenticatorUsageText;
+    const allowedFor = this.model.get('relatesTo').allowedFor;
+
+    if (allowedFor) {
+      switch (allowedFor) {
+      case 'any':
```
pradeepdewda-okta, line 54: make them const ?

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
```diff
@@ -42,7 +44,33 @@ const AuthenticatorRow = View.extend({
...
+        break;
+      case 'sso':
+        authenticatorUsageText = loc('oie.enroll.authenticator.usage.text.access', 'login');
+        break;
+      }
+    }
+
+    return {
+      authenticatorUsageText,
+      iconClassName: this.model.get('iconClassName'),
```
pradeepdewda-okta, line 68: if these are coming from model then they should be auto available as `options`

indirathirumalaimurugan-okta: do you mean `this.options.model.get('iconClassName')`? I don't see the values directly under `options`. does it make a difference whether I get the value as `this.model.get.. or this.options.model.get`

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
indirathirumalaimurugan-okta: is this the right spot to put such constants `src/v2/view-builder/utils/Constants.js`

playground/mocks/data/idp/idx/authenticator-enroll-select-authenticator-with-usage-info.json
```diff
@@ -0,0 +1,219 @@
...
+        "name": "stateHandle",
+        "required": true,
+        "value": "02CqFbzJ_zMGCqXut-1CNXfafiTkh9wGlbFqi9Xupt",
+        "visible": false,
+        "mutable": false
+      }
+    ]
+  }
+}
+
```
mauriciocastillosilva-okta, line 219: nit: there are 2 extra lines at the end, with one should be good

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
indirathirumalaimurugan-okta: the text requires specific color and other css props. adding `explain` or `explain-text` does not add any css values.

assets/sass/v2/_select-authenticator.scss
```diff
@@ -67,6 +67,13 @@
   display: flex;
   align-items: center;
 
+  .authenticator-usage-text {
+    color: $light-text-color;
+    font-size: 13px;
```
mauriciocastillosilva-okta, line 72: do we really need all these styles? 
* `font-size` should be inherited, so it is the same as the authenticator description
* `line-height`: same situation, should be inherited to the same
* `margin-top`: this we might or might not need. I think it looks good without an extra few pixels, so we should remove it. Can you run it by UX please?
* `color`: this we definitely need

mauriciocastillosilva-okta (commented):
just one minor comment, otherwise looks great!

playground/mocks/data/idp/idx/authenticator-enroll-select-authenticator-with-usage-info.json
indirathirumalaimurugan-okta: there were extra spaces on the line 219. removed it.

assets/sass/v2/_select-authenticator.scss
indirathirumalaimurugan-okta: UX is ok with having just the color.

mauriciocastillosilva-okta (approved):

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
pradeepdewda-okta: yeah.. why not ?

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
pradeepdewda-okta: I meant you need not to create the object for existing values again. they should be available in the template to use as is.

For ex.. just extend the object instead.? 

```
    const data = View.prototype.getTemplateData.apply(this, arguments);
    
    //extend data with new attributes 
    return data
```

test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
pradeepdewda-okta: so lets check for that element doesn't exist.. This can cover a not-so-happy path

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
pradeepdewda-okta: would it make sense now to use inbuilt class ? since now we don't have any other values.

indirathirumalaimurugan-okta: we are still passing the color `#a7a7a7`. could you be specific on what inbuilt class you are referring to?

test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
indirathirumalaimurugan-okta: its being checked at line line #201

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
pradeepdewda-okta: The className was not top of my ahead. But here you go, i search for it. 

https://github.com/okta/okta-signin-widget/blob/master/assets/sass/okta-theme.scss#L54

indirathirumalaimurugan-okta: `$medium-text-color` is set to `#777` but UX wants `#a7a7a7`

TianGan-okta (approved):

pradeepdewda-okta (approved):
🥇

PR #2434, opened 2022-03-17
Okta 477746 svg support custom app push

praveendurai-okta:
## Description:

INCREMENTAL DEVELOPMENT

Add SVG Logo support which will be introduced in Custom App Authenticator. Logos should look exactly same as other authenticator in Factor Listing, Enroll and Challenge Screens.

Custom App Will be appended in screens based on the API response, and we do have FF in backend to avoid that for all orgs. 

## PR Checklist

- [ ] Added unit tests?

- [X] Added e2e tests

### Screenshot/Video:

Listing Page

Screen Shot 2022-03-22 at 3 07 10 PM

Setup/Enroll Page
Custom App will be hidden from Setup page. Backend work is in progress. For now it will look like this. 

Screen Shot 2022-03-22 at 3 12 10 PM

Challenge Page

Screen Shot 2022-03-22 at 3 07 16 PM

### Reviewers:

### Issue:

- [OKTA-477746](https://oktainc.atlassian.net/browse/OKTA-477746)


assets/sass/v2/_select-authenticator.scss
```diff
@@ -100,6 +100,23 @@
...
+    width: 38px;
+    height: 38px;
+    border: 1px solid #ccc;
+    border-radius: 50%;
+    .custom-logo {
+      background-repeat: no-repeat;
+      background-position: 50% 50%;
+      background-size: cover;
+      width: 75%;
+      height: 75%;
```
alexdahl-okta, line 116: 👋 Based on screenshots, it looks like something about the scaling implementation causes inconsistent outcomes between the list view and the challenge view? (The bird has less padding in the challenge view.)

Could we also stress test for forward compatibility of the SVG asset? The first-party graphics, sizing, spacing, background color, and customer theming of authenticator icons will change over time... ideally, an asset a customer provides today would be durable across those changes. I _think_ it does make sense for the widget to do padding or scaling somehow, and not have the customer bake padding into their asset. 🤔 But not sure, curious if @lesterchoi-okta has any thoughts for future proofing?

praveendurai-okta: The bird has less padding in challenge view because we had to extend the SVG icon end to end without adding too much padding programatically. for example, if admin needs to upload colored background SVG something like below, they would not see too much padding around.
Screen Shot 2022-03-21 at 3 43 46 PM
 
The bird SVG we are using in this example is a draft version. May be it needs more padding in itself. We do have preview feature in admin screen, where they can upload different kind of logos to see if it fits perfectly for our usage. 

We also had a discussion regarding this in [slack](https://okta.slack.com/archives/C01DR8G2QBE/p1647892844770659?thread_ts=1647886715.796429&cid=C01DR8G2QBE)

playground/mocks/data/idp/idx/authenticator-verification-custom-app-push-reject.json
```diff
@@ -178,7 +178,7 @@
       "key":"custom_app",
       "id":"aut198w4v0f8dr8gT0g4",
       "displayName":"Custom Push",
-      "logo":"https://tc1static.oktacdn.com/assets/img/logos/custompushLogo.svg",
```
indirathirumalaimurugan-okta, line 181: is this `logo` prop no longer used anywhere?

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
```diff
@@ -16,19 +16,26 @@ import hbs from 'handlebars-inline-precompile';
...
-          {{#if description}}\
-            <p class="authenticator-description--text">{{description}}</p>\
-          {{/if}}\
-          <div class="authenticator-button" {{#if buttonDataSeAttr}}data-se="{{buttonDataSeAttr}}"{{/if}}></div>\
-        </div>\
-      '),
+  template: hbs`
+        <div class="authenticator-icon-container">
+        {{#if logoUri}}
+          <div class="custom-auth-icon">
```
indirathirumalaimurugan-okta, line 22: can `factor-icon` and `authenticator-icon` css class be used here as well? it may already have the needed css values.

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
```diff
@@ -16,19 +16,26 @@ import hbs from 'handlebars-inline-precompile';
...
+            <img class="custom-logo" src="{{logoUri}}" 
+              alt="{{i18n code="oie.auth.logo.alt" bundle="login"}}"/>
+          </div>
+          {{else}}
+          <div class="factor-icon authenticator-icon {{iconClassName}}">
+          {{/if}}
+          </div>
```
indirathirumalaimurugan-okta, line 29: nit: indentation is little off here

playground/mocks/data/idp/idx/authenticator-verification-custom-app-push-reject.json
praveendurai-okta: No it is not. SIW Custom push work was partly done long back and it was not used anywhere. during recent development we have decided to change the prop to `logoUri`

src/v2/view-builder/utils/AuthenticatorUtil.js
```diff
@@ -25,6 +25,7 @@ const getButtonDataSeAttr = function(authenticator) {
 };
 
 /* eslint complexity: [0, 0] */
+// eslint-disable-next-line max-statements
```
indirathirumalaimurugan-okta, line 28: is it needed to disable this lint rule?

src/v2/view-builder/utils/AuthenticatorUtil.js
```diff
@@ -181,12 +182,16 @@ const getAuthenticatorData = function(authenticator, isVerifyAuthenticator) {
...
+      description: '', // TODO: Yet to decide on the description OKTA-480733
       buttonDataSeAttr: getButtonDataSeAttr(authenticator),
     });
     break;
   }
+
+  }
+  // Add logo URI, Logo URI must take precedence than icon class name
+  if(authenticator?.relatesTo?.logoUri) {
+    authenticatorData.logoUri = authenticator.relatesTo.logoUri;
```
indirathirumalaimurugan-okta, line 194: if `logoUri` is specific to `custom_app` can this line be moved within that case check `case AUTHENTICATOR_KEY.CUSTOM_APP`

src/v2/view-builder/views/shared/ChallengePushView.js
```diff
@@ -104,9 +104,17 @@ const Body = BaseFormWithPolling.extend(Object.assign(
...
 
+    isCustomApp() {
+      return this.options.appState.get('authenticatorKey') === AUTHENTICATOR_KEY.CUSTOM_APP;
+    },
+
     isOVWithAutoChallenge() {
       return this.isOV() && this.options.appState.getSchemaByName('autoChallenge');
     },
+
+    isCustomAppAutoChallenge(){
```
indirathirumalaimurugan-okta, line 115: nit: isCustomAppAutoChallenge -> isCustomAppWithAutoChallenge

src/v2/view-builder/utils/AuthenticatorUtil.js
praveendurai-okta: I do not want to restrict it only for `custom_app`. in future, if any auth have `logoUri` prop then it can take precedence.

indirathirumalaimurugan-okta: nit: there are 15 auth keys and this if condition would be hit for all of those even though its relevant just to custom_app. I find it unnecessary unless there is any plan to add logo uri for other authenticators in near future.

src/v2/view-builder/components/HeaderBeacon.js
```diff
@@ -14,19 +14,28 @@ import hbs from 'handlebars-inline-precompile';
...
-    </div >\
-  '),
+  template: hbs`
+    <div data-type="beacon-container" class="beacon-container">
+      <div class="beacon-blank auth-beacon">
+        <div class="beacon-blank js-blank-beacon-border auth-beacon-border"></div>
+      </div>
+      {{#if logoUri}}
+      <div class="bg-helper auth-beacon auth-beacon-factor custom-logo" data-se="factor-beacon" 
+        style="background-image: url('{{logoUri}}')">
```
mauriciocastillosilva-okta, line 24: I see you're using a `background-image` here instead of an `` tag like in the other files. Any specific reason we can't do the same in all files, either use `background-image` or the image tag? I am guessing there is something special about the Beacon that we need this background-image here, so curious what it is

test/testcafe/spec/ChallengeOktaVerifyPush_spec.js
```diff
@@ -57,9 +59,11 @@ test
     const pageTitle = challengeOktaVerifyPushPageObject.getFormTitle();
     const pushBtn = challengeOktaVerifyPushPageObject.getPushButton();
     const a11ySpan = challengeOktaVerifyPushPageObject.getA11ySpan();
+    const logoClass = challengeOktaVerifyPushPageObject.getBeaconClass();
     await t.expect(pushBtn.textContent).contains('Push notification sent');
     await t.expect(a11ySpan.textContent).contains('Push notification sent');
     await t.expect(pushBtn.hasClass('link-button-disabled')).ok();
+    await t.expect(logoClass).contains(BEACON_CLASS);
```
mauriciocastillosilva-okta, line 66: I see in the challenge auto-push spec you compare to a string directly `custom-logo`, and here you compare to a constant. I think you should be consistent for both. Direct string is fine with me, but if you want a constant, let's name it something more specific to what it is `BEACON_CLASS` is very generic for the Beacon, it is not for challenge ov push

src/v2/view-builder/components/HeaderBeacon.js
praveendurai-okta: Beacon logo is bigger than previous ones and need to support any kind of SVGs from small to large ones. Hence used `background-image` as it gives more control on positioning the image also i can reuse the existing css classes. 
used `` in other places because it is semantically correct, screen readers compatible and no much customization or positioning is needed as it is very small size.

As per today's discussion, we concluded that No padding/spacing should be added from SIW. I will go ahead and replace the other occurrences of `` to `background-image` so that i can reuse existing classes, and have more control on customization if needed in future. Will add `role` and `aria-label` to support Screen readers.

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
praveendurai-okta: i had to use `` earlier so i cannot really reuse existing classes. Now that requirement has changed, i have used `background-image` with existing classes

src/v2/view-builder/utils/AuthenticatorUtil.js
praveendurai-okta: moved inside switch.

assets/sass/v2/_select-authenticator.scss
```diff
@@ -104,6 +104,20 @@
     background-size: 38px 38px;
   }
 
+  .custom-auth-icon {
```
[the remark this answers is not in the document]
praveendurai-okta: this is not valid anymore. i have removed those styles.

test/testcafe/spec/ChallengeOktaVerifyPush_spec.js
praveendurai-okta: i have use string directly now.

assets/sass/v2/_select-authenticator.scss
praveendurai-okta: For now we are NOT going to add any padding from SIW Widget as per our UX suggestion. As this is Incremental development, we are happy to update it if UX feels otherwise. 
cc. @mauriciocastillosilva-okta

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
```diff
@@ -19,7 +19,14 @@ const AuthenticatorRow = View.extend({
   className: 'authenticator-row clearfix',
   template: hbs`
     <div class="authenticator-icon-container">
-      <div class="factor-icon authenticator-icon {{iconClassName}}"></div>
+      {{#if logoUri}}
+        <div class="factor-icon authenticator-icon custom-app-logo" role="img"
```
indirathirumalaimurugan-okta, line 23: is this `custom-app-logo` defined in any scss file? can the style `background-image: url('{{logoUri}}')` be defined within `custom-app-logo` so the inline style can be removed?

praveendurai-okta: `logoUri` is a dynamic value which we will be getting from API response. hence adding it as inline styles dynamically. `custom-app-logo` is used in test cases to differentiate it from other `{{iconClassName}}`. Also this is Increment development, we can add more styles to this class when needed.

test/testcafe/spec/ChallengeCustomAppPush_spec.js
```diff
@@ -124,6 +169,51 @@ test
...
+
+    const { request: {
+      method: answerRequestMethod2,
+      url: answerRequestUrl2,
+    }
+    } = logger.requests[1];    
+    await t.expect(answerRequestMethod2).eql('post');
+    await t.expect(answerRequestUrl2).eql('http://localhost:3000/idp/idx/authenticators/poll');
+
+    // unselect checkbox on click to ensure its still accessbile while polling
```
indirathirumalaimurugan-okta, line 212: nit: accessbile -> accessible

test/testcafe/spec/ChallengeOktaVerifyPush_spec.js
```diff
@@ -7,6 +7,7 @@ import pushPoll from '../../../playground/mocks/data/idp/idx/authenticator-verif
 import success from '../../../playground/mocks/data/idp/idx/success';
 import pushPollAutoChallenge from '../../../playground/mocks/data/idp/idx/authenticator-verification-okta-verify-push-autoChallenge-on';
 
+
```
indirathirumalaimurugan-okta, line 10: nit: are there 2 line breaks here?

src/v2/view-builder/views/consent/AdminConsentViewHeader.js
```diff
@@ -12,7 +12,7 @@ const AdminConsentViewHeader = View.extend({
     <a href="{{clientURI}}" class="client-logo-link" target="_blank">
     {{/if}}
     {{#if customLogo}}
-      <img class="client-logo custom-logo" src="{{customLogo}}" alt="{{altText}}" aria-hidden="true" />
+      <img class="client-logo custom-app-logo" src="{{customLogo}}" alt="{{altText}}" aria-hidden="true" />
```
mauriciocastillosilva-okta, line 15: I don't think you need to change this file, this is for a different view

src/views/shared/templates/consentLogoHeaderTemplate.js
```diff
@@ -5,7 +5,7 @@ const consentLogoHeaderTemplate = hbs`{{#if clientURI}}
 {{/if}}
 {{#if customLogo}}
   <img
-    class="client-logo custom-logo"
+    class="client-logo custom-app-logo"
```
mauriciocastillosilva-okta, line 8: same here, I don't think consent views and templates need changes

test/testcafe/spec/ChallengeCustomAppPush_spec.js
```diff
@@ -67,6 +77,41 @@ test
...
+
+    const pageTitle = challengeCustomAppPushPageObject.getPageTitle();
+    const pushBtn = challengeCustomAppPushPageObject.getPushButton();
+    const a11ySpan = challengeCustomAppPushPageObject.getA11ySpan();
+    const checkboxLabel = challengeCustomAppPushPageObject.getAutoChallengeCheckboxLabel();
+    const logoClass = challengeCustomAppPushPageObject.getBeaconClass();
+    await t.expect(pushBtn.textContent).contains('Push notification sent');
+    await t.expect(a11ySpan.textContent).contains('Push notification sent');
+    await t.expect(pushBtn.hasClass('link-button-disabled')).ok();
+    await t.expect(logoClass).contains('custom-app-logo');
```
mauriciocastillosilva-okta, line 98: can we also check this is not true when `logoUri` is not present?

playground/mocks/data/idp/idx/authenticator-verification-custom-app-push-enroll.json
```diff
@@ -0,0 +1,251 @@
+{
```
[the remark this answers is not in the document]
mauriciocastillosilva-okta: Actually, is this file used?

mauriciocastillosilva-okta (changes requested):

src/v2/view-builder/views/consent/AdminConsentViewHeader.js
praveendurai-okta: Yes Good catch.. i thought i reverted this but somehow it slipped.

test/testcafe/spec/ChallengeCustomAppPush_spec.js
praveendurai-okta: if `logoUri` is not present, it will be considered as non-custom app . we do have a check for okta-verify [here](https://github.com/okta/okta-signin-widget/blob/0b46d2d18fc60d7ca8f811876f77d0d35f765f2f/test/testcafe/spec/ChallengeOktaVerifyPush_spec.js#L65) Isn't that sufficient ?

playground/mocks/data/idp/idx/authenticator-verification-custom-app-push-enroll.json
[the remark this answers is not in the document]
praveendurai-okta: I have named this file to match with other similar mocks for custom app. Do you want to rename other files too ?
[authenticator-verification-custom-app-push.json](https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/data/idp/idx/authenticator-verification-custom-app-push.json)
[authenticator-verification-custom-app-push-reject.json](https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/data/idp/idx/authenticator-verification-custom-app-push-reject.json)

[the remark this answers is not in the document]
mauriciocastillosilva-okta: Those files are not the same as this one, as I mentioned above the naming convention is following the remediation form present in the mock response. In this file the remediation form is `select-authenticator-enroll` (line 11), so this file should be have the prefix of `authenticator-enroll-`

The other file you are adding in this PR is for a challenge verification, so the name is correct.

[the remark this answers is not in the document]
praveendurai-okta: I see. it makes sense now to have prefix of remediation form name. thanks for the info. I have updated it.

src/v2/view-builder/utils/AuthenticatorUtil.js
praveendurai-okta: This is not needed as i moved some of the statements inside switch hence removed it.

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -581,3 +603,12 @@ test.requestHooks(mockSelectAuthenticatorForRecovery)('should not show custom fa
...
   await t.expect(await pageObject.factorPageHelpLinksExists()).notOk();
 });
+
+test.requestHooks(mockChallengeCustomApp)('should navigate to Custom App challenge page', async t => {
+  const selectFactorPage = await setup(t);
+  await t.expect(selectFactorPage.getIdentifier()).eql('testUser@okta.com');
+  await t.expect(selectFactorPage.getFormTitle()).eql('Verify it\'s you with a security method');
+  selectFactorPage.selectFactorByIndex(9);
+  const challengeFactorPage = new ChallengeFactorPageObject(t);
+  await t.expect(challengeFactorPage.getFormTitle()).eql('Verify with Custom Push');
```
indirathirumalaimurugan-okta, line 613: nit: may be also check for the presence of custom app logo

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -1129,6 +1130,10 @@ oie.yubikey.challenge.title = Verify with YubiKey
 oie.yubikey.description = Insert the YubiKey into a USB port and tap it to generate a verification code.
 oie.yubikey.passcode.label = Insert then tap your YubiKey
 
+## Custom App Authenticator
+# {0} refers to the display name of the authenticator "My Custom Auth"
```
indirathirumalaimurugan-okta, line 1134: nit: I understand it as "My Custom Auth" is the authenticator name but I believe it can be anything. Can it be changed to `# {0} refers to the display name of the authenticator like "My Custom Auth"`

test/testcafe/spec/ChallengeCustomAppPush_spec.js
```diff
@@ -124,6 +169,51 @@ test
...
+    await t.wait(1000); // 8 sec total elapsed
+    await t.expect(logger.count(() => true)).eql(2);
+
+    const { request: {
+      method: answerRequestMethod2,
+      url: answerRequestUrl2,
+    }
+    } = logger.requests[1];    
+    await t.expect(answerRequestMethod2).eql('post');
+    await t.expect(answerRequestUrl2).eql('http://localhost:3000/idp/idx/authenticators/poll');
```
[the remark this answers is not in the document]
praveendurai-okta: yeah i got your point but I believe the intention here is to check if Poll is happening at regular interval and not only once. do you still think we need to cut down 2nd call ?

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
praveendurai-okta: app logo and other details in the page are already verified [here](https://github.com/okta/okta-signin-widget/blob/33291ef94d29bbc9543a0b3fe8b8aa36f1549b54/test/testcafe/spec/ChallengeCustomAppPush_spec.js#L81). The scope of this test is just to ensure the page navigates correctly.

packages/@okta/i18n/src/properties/login.properties
mauriciocastillosilva-okta: good point, I think we can even be just generic about this, without a name. It still gives the context to the translators
`# {0} is the display name of the authenticator`
We already use this around [line 490](https://github.com/okta/okta-signin-widget/blob/33291ef94d29bbc9543a0b3fe8b8aa36f1549b54/packages/%40okta/i18n/src/properties/login.properties#L494)

mauriciocastillosilva-okta (approved):
just one comment in the properties file. After fixing that, looks great ⛵

praveendurai-okta:
have tested this version in okta-core and it is [green](https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&sha=6380f6143c677e553475e0b279bdf46b4e1847da&page=1&pageSize=6)

indirathirumalaimurugan-okta (approved):
lgtm

PR #2435, opened 2022-03-17
fix: open custom help links on the new tab

TianGan-okta:
## Description:
open custom help links on the new tab

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
open_in_new_tab

### Reviewers:

### Issue:

- [OKTA-479148](https://oktainc.atlassian.net/browse/OKTA-479148)

### DIFF
--- a/src/v2/view-builder/components/Link.js
+++ b/src/v2/view-builder/components/Link.js
@@ -23,7 +23,8 @@ const Link = View.extend({
     let href = this.options.href || '#';
     return {
       'data-se': this.options.name,
-      href: href
+      href: href,
+      target: this.options.target,
     };
   },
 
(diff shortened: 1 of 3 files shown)


pradeepdewda-okta (approved):

src/v2/view-builder/components/Link.js
```diff
@@ -23,7 +23,8 @@ const Link = View.extend({
     let href = this.options.href || '#';
     return {
       'data-se': this.options.name,
-      href: href
+      href: href,
+      target: this.options.target,
```
indirathirumalaimurugan-okta, line 27: in case when `this.options.target` doesn't exist, should it default to `_self`? `target: this.options.target || '_self'`

TianGan-okta: I don't think we need to do this because it will not add the target to the element if there is no value ({...target: undefined}), which will open the window self, we don't need to explicitly add '_self'.

indirathirumalaimurugan-okta (approved):
lgtm

PR #2446, opened 2022-03-25
OKTA-462156 l10n error on password expiry warning

lesterchoi-okta:
### Bug:
Customer reports seeing `L10N_ERROR` instead of the correct warning message about an expiring password. This bug only occurs when the widget is on a custom URL domain (vs. Okta-hosted).

### Expected:
Widget displays the message "When your password expires you will be locked out of your {company_name} account."

### Actual:
Widget displays the message: "L10N_ERROR[idx.password.expiring.message]"

### Cause:
An unexpected i18n param in the IDX API response for introspect causes the widget to display  on customer-hosted widget.

Okta-hosted returns:
https://github.com/okta/okta-signin-widget/blob/b9d2e70e911168f1da95b7bb14ae9cc15121fc52/playground/mocks/data/idp/idx/authenticator-expiry-warning-password.json#L150-L164

Customer-hosted returns:
https://github.com/okta/okta-signin-widget/blob/b9d2e70e911168f1da95b7bb14ae9cc15121fc52/playground/mocks/data/idp/idx/authenticator-expiry-warning-password-with-brand-name.json#L150-L164

l10n:
https://github.com/okta/okta-signin-widget/blob/b9d2e70e911168f1da95b7bb14ae9cc15121fc52/packages/%40okta/i18n/src/properties/login.properties#L898

### Fix:
- ~~Option 1: update the API to use a new i18n key that expects a param for company name + send "Okta" as default param~~
- **Option 2: update the API to remove the company name param, and re-word message.**
- ~~Option 3: update the SIW (or i18n lib?) to handle the case conditionally~~

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

localhost_3000_

### Issue:

- [OKTA-462156](https://oktainc.atlassian.net/browse/OKTA-462156)

### DIFF
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -18,7 +18,7 @@ const idx = {
   ],
 
   '/idp/idx/introspect': [
-    'identify',
+    // 'identify',
     // 'error-identify-multiple-errors',
     // 'authenticator-enroll-ov-qr-enable-biometrics',
     // 'authenticator-verification-okta-verify-push',
@@ -66,7 +66,8 @@ const idx = {
     // 'authenticator-expired-password',
     // 'authenticator-expired-password-no-complexity',
     // 'authenticator-expired-password-with-enrollment-authenticator',
-    // 'authenticator-expiry-warning-password',
+    'authenticator-expiry-warning-password',
+    'authenticator-expiry-warning-password-with-brand-name',
     // 'device-code-activate',
     // 'enroll-profile',
     // 'enroll-profile-new',
--- a/playground/mocks/data/idp/idx/authenticator-expiry-warning-password-with-brand-name.json
+++ b/playground/mocks/data/idp/idx/authenticator-expiry-warning-password-with-brand-name.json
@@ -1,0 +1,165 @@
+{
+  "stateHandle":"022P5Fd8jBy3b77XEdFCqnjz__5wQxksRfrAS4z6wP",
+  "version":"1.0.0",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "name": "@okta/okta-signin-widget",
   "description": "The Okta Sign-In Widget",
-  "version": "6.2.0",
+  "version": "6.3.0",
```
aarongranick-okta, line 5: this should not be here. If you want it based on 6.2, you must rebase interactively and remove unrelated commits from master

playground/mocks/config/responseConfig.js
```diff
@@ -18,7 +18,7 @@ const idx = {
   ],
 
   '/idp/idx/introspect': [
-    'identify',
+    // 'identify',
```
aarongranick-okta, line 21: remember to revert this change before you open PR for review

package.json
lesterchoi-okta: Yeah i see the other unrelated. I'll rebase and update

playground/mocks/config/responseConfig.js
lesterchoi-okta: Will do. Enabled just the relevant responses for easy bug repro

lesterchoi-okta:
Closing. Fix recommendation is to ask customer to pass `brandName` to prevent error caused by unexpected API param in IDX response.
https://oktainc.atlassian.net/browse/OKTA-462156?focusedCommentId=2487219

PR #2455, opened 2022-04-08
Publishes type declarations

aarongranick-okta:
## Description:
- Builds types for courage bundle
- Converts some core objects to Typescript
- Adds type declarations to published module
- Includes test for type declarations
- Adds lint to validate imports in type declarations

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

- Did you verify the change by running downstream monolith artifacts? (YES)

### Issue:

- [OKTA-390113](https://oktainc.atlassian.net/browse/OKTA-390113)


packages/@okta/courage-dist/types/courage/framework/Collection.d.ts
```diff
@@ -0,0 +1,43 @@
+import Backbone, { _Result } from 'backbone';
```
haishengwu-okta, line 1: do we need to check in all those types ?

packages/@okta/courage-for-signin-widget/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
+    "allowJs": true,
+    "checkJs": false,
+    "skipLibCheck": true,
+    "paths": {
+      "framework/*": ["node_modules/@okta/courage/src/framework/*"],
+      "vendor/*": ["node_modules/@okta/courage/src/vendor/*"],
+      "handlebars-inline-precompile": ["../../../scripts/buildtools/tsc/handlebars-inline-precompile"],
+      "ConfirmationDialog": ["buildtools/ConfirmationDialog.ts"],
```
haishengwu-okta, line 17: this probably wont work any more as I change the `ConfirmationDialog` to relative path in courage... :(
maybe patch the courage code like copy the empty module to `node_module/courage/...` ?

src/models/Settings.ts
```diff
@@ -0,0 +1,581 @@
...
+    cache: true,
+  }
+};
+
+type SettingsProps = typeof local & typeof derived;
+
+export default class Settings extends Model {
+  authClient:OktaAuth;
+
+  get<A extends Backbone._StringKey<SettingsProps>>(attributeName: A): any {
```
haishengwu-okta, line 361: 👍 👍

src/types/api.ts
```diff
@@ -0,0 +1,161 @@
...
+  showSignInAndRedirect(options: RenderOptions): Promise<void>;
+  renderEl(
+    options: RenderOptions,
+    success?: RenderSuccessCallback,
+    error?: RenderErrorCallback
+  ): Promise<RenderResult>;
+
+  getUser(): void
+}
+export interface OktaSignInConstructor {
```
haishengwu-okta, line 35: Since OktaSignIn is already a `class`. why would still need an interface and constructor type?
- will be able to generate class type from the `class OktaSignIn` implementation?
- for developer, they property can do `const siw : OktaSignIn = new OktaSignIn({...})`?

haishengwu-okta (commented):
THIS IS AWESOME !!

packages/@okta/courage-for-signin-widget/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "outDir": "./target",
+    "sourceMap": true,
+    "resolveJsonModule": true,
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
+    "allowJs": true,
+    "checkJs": false,
+    "skipLibCheck": true,
+    "paths": {
+      "framework/*": ["node_modules/@okta/courage/src/framework/*"],
```
haishengwu-okta, line 14: this is probably also not needed?

packages/@okta/courage-dist/types/courage/framework/Collection.d.ts
aarongranick-okta: we do need to check in the courage types, since only we will have the original source and be able to build it.

packages/@okta/courage-for-signin-widget/tsconfig.json
aarongranick-okta: thanks for the heads up, I'll see what I can do

packages/@okta/courage-for-signin-widget/webpack.config.js
```diff
@@ -57,31 +88,18 @@ const webpackConfig = {
       'ConfirmationDialog': EMPTY,
 
       'vendor': SHARED_JS + '/vendor',
+
+      'backbone': `${SHARED_JS}/vendor/lib/backbone.js`
```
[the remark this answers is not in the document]
aarongranick-okta: backbone is not installed in node_modules, it is coming in from courage. I'm not completely sure how it was working before: https://github.com/okta/okta-signin-widget/pull/2292/files

src/v2/BaseLoginRouter.ts
```diff
@@ -272,27 +288,32 @@ export default Router.extend({
...
     this.stopListening(this.settings);
     Bundles.remove();
     Backbone.history.stop();
-  },
-});
+  }
+}
+
+// Add "Events" to prototype for compatibility with existing code
+BaseLoginRouter.prototype.Events = Backbone.Events;
```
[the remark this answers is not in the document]
aarongranick-okta: This is for compatibility with existing code which is calling BaseLoginRouter.prototype.Events.  Class members (fields) aer not put on the prototype automatically in TS classes. (although this can be done with a babel transform which converts to CJS). We are planning to expose widget src as ESM so we need it to be compatible without transpile to CJS

src/widget/OktaSignIn.ts
```diff
@@ -0,0 +1,214 @@
...
+    });
+  }
+
+  getUser(): any {
+    return this.router?.appState?.getUser();
+  }
+  
+  // Events API
+
+  on(...args: any[]): void {
```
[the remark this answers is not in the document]
aarongranick-okta: yeah good suggestion

packages/@okta/courage-for-signin-widget/tsconfig.json
aarongranick-okta: I was able to replace it using webpack resolve

packages/@okta/courage-for-signin-widget/.eslintignore
```diff
@@ -0,0 +1 @@
+src/courage
```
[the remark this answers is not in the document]
aarongranick-okta: courage source is symlinked to hoist it out of node_modules and create a clean relative path. Only Okta engineers will be able to build/symlink the courage src. ts-transform-paths will not transform any paths under node_modules

packages/@okta/courage-for-signin-widget/webpack.config.js
```diff
@@ -37,6 +67,7 @@ const webpackConfig = {
   },
   externals: EXTERNAL_PATHS,
   resolve: {
+    extensions: ['.js', '.ts'],
     alias: {
 
       // jsons is from StringUtil
```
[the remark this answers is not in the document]
aarongranick-okta: We are building from raw courage source (not the webpack bundle), so it includes .ts files.

packages/@okta/courage-for-signin-widget/webpack.config.js
```diff
@@ -57,31 +88,18 @@ const webpackConfig = {
       'ConfirmationDialog': EMPTY,
```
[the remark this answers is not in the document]
aarongranick-okta: have adjusted this

playground/main.ts
```diff
@@ -1,24 +1,48 @@
-/* global OktaSignIn */
 /* eslint no-console: 0 */
 
-import signinWidgetOptions from '../.widgetrc.js';
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const signinWidgetOptions = require('../.widgetrc.js'); // commonJS module
```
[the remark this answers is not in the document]
aarongranick-okta: everyone would need to change their widgetrc to do `export default` instead of `module.exports`. Babel doesn't mind either way, but TS is pickier about wanting a "default" export for ES modules.

src/types/api.ts
aarongranick-okta: One downside of the class is that because it is both a type and value, there can be conflicts if you have two different versions of the object, if the newer version has added a property it won't be assignable to the old type. One way this could easily happen is because we allow the `authClient` to be passed in directly. We defend by declaring that authClient is using the OktaAuthInterface (the shape of AuthJS), not OktaAuth (the exact class we imported). The App may be using a different version of AuthJS than the widget was built against.  Or it is loading a different version widget than it had types for. In this case an app can do something like:

```
const OktaSignIn: OktaSignInConstructor = require('@okta/okta-signin-widget');
const signIn: OktaSignInAPI = new OktaSignIn(...options);
```

By using interfaces instead of classes, it is looser about the shape of the object that is returned from require() and new() - it only has to conform with the public API.

src/widget/OktaSignIn.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  after(formName: string, callbackFn: HookFunction): void {
+    this.hooks.mergeHook(formName, {
+      after: [
+        callbackFn
+      ]
+    });
+  }
+
+  getUser(): any {
```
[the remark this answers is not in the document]
aarongranick-okta: This was added to support the account chooser https://login.okta.com/

packages/@okta/courage-for-signin-widget/webpack.config.js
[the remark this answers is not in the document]
aarongranick-okta: I am now using backbone-min from node_modules, shaved about 40K off the bundle size! Also added source-map loader so should be a good debugging experience

packages/@okta/courage-for-signin-widget/webpack.config.js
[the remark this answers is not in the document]
haishengwu-okta: I meant the jsons alias.  IIRC JSONS had been removed from Courage already.

scripts/buildtools/commands/verify-package.js
```diff
@@ -33,7 +33,10 @@ exports.handler = function() {
     'dist/labels/json/login_ru.json',
     'dist/sass/_fonts.scss',
     'dist/font/okticon.ttf',
-    'dist/font/okticon.woff'
+    'dist/font/okticon.woff',
+    'types/index.d.ts',
+    'types/generated/src/index.d.ts',
+    'types/generated/packages/@okta/courage-dist/types/courage/framework/Model.d.ts'
```
haishengwu-okta, line 39: is this just a sanity check that `courage for widget` types has been generated?

src/index.ts
```diff
@@ -0,0 +1,7 @@
+// Entry point for Typescript build
+import OktaSignIn from './widget/OktaSignIn';
+
+export default OktaSignIn;
+export { OktaSignIn };
```
haishengwu-okta, line 5: why would need to do both?
isn't `export default OktaSignIn` good enough?

src/types/options.ts
```diff
@@ -0,0 +1,276 @@
...
+export interface WidgetOptions extends Partial<Pick<OktaAuthOptions,
+  'issuer' | 
+  'clientId' | 
+  'redirectUri' |
+  'state' | 
+  'scopes' |
+  'codeChallenge' |
+  'codeChallengeMethod' |
+  'flow'
+>> {
```
shuowu, line 20: merge conflict?

src/types/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './api';
+export * from './errors';
+export * from './registration';
```
shuowu, line 3: Does external devs need types from this module?

Gruntfile.js
```diff
@@ -28,6 +29,19 @@ module.exports = function(grunt) {
...
+      'types': {
+        files: [
+          {
+            expand: true,
+            cwd: COURAGE_TYPES,
+            src: [
+              '**/*.d.ts'
+            ],
+            dest: `types/generated/${COURAGE_TYPES}`
+          }
```
haishengwu-okta, line 41: why need to copy courage types to top level.

scripts/buildtools/commands/verify-package.js
aarongranick-okta: yes, verify that:
- types were copied to dist
- widget generated types exist
- courage generated types exist

src/index.ts
aarongranick-okta: When we are using named exports for code, we will want to do both. This is getting a little bit ahead, I will remove it to avoid confusion, since `import { OktaSignIn }` would be a type-only import at this time.

src/types/options.ts
aarongranick-okta: Nope, it's just a generic inside a generic.  `Partial<Pick<X>>`

src/types/errors.ts
```diff
@@ -0,0 +1,16 @@
+
+export interface FieldError {
+  errorSummary: string;
+  reason?: string;
+  location?: string;
+  locationType?: string;
+  domain?: string;
+}
+
+export interface ApiError {
```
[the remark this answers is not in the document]
aarongranick-okta: It does look a lot like the `ApiError` interface from AuthJS. I'll take a closer look and see if it makes sense to combine

src/types/index.ts
aarongranick-okta: Sure. See usage in playground and test app.

Gruntfile.js
aarongranick-okta: courage is not available as an installable module so all necessary courage types are published in a folder underneath the widget's type exports

src/types/errors.ts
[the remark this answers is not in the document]
aarongranick-okta: Difference is AuthJS defines `errorCauses` as `string[]`, but SIW has definition that shows `errorCauses` contains objects (`FieldError` type). I think SIW definition may be more correct here, we may want to update the AuthJS side (and add some testing with real error mocks)

src/v2/BaseLoginRouter.ts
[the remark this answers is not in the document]
aarongranick-okta: I think there probably can be some refactor/simplification around the existing code, but this PR is already so large. I am trying to reduce any chance of breaking something. So I'm mostly only changing things that are necessary for TS type exports. A few files I converted to TS/classes to verify that the courage types are working as expected (ran into issues converting some Views to classes, so rolled that back). There will definitely be follow up PRs to continue TS conversion and refine types, I think we can tackle this Router events thing in one of those PRs.

src/types/errors.ts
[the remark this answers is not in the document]
shuowu: oh, right, I also noticed this issue recently in myaccount PR.. 
how about a `//TODO` comment here for future update?

shuowu-okta (approved):

Gruntfile.js
haishengwu-okta: For my education, cannot we generate a single `index.d.ts` which includes all (like `backbone/index.d.ts`. although it may not be generated by hand crafted)

src/types/errors.ts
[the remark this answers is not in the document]
aarongranick-okta: Created an issue to fix/refine types (in okta-auth-js): OKTA-489428

playground/main.ts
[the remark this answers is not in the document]
haishengwu-okta: I wonder if we can make this break change at next major release or so. just for consistency (all ESM)

packages/@okta/courage-for-signin-widget/webpack.config.js
```diff
@@ -54,35 +82,28 @@ const webpackConfig = {
       'vendor/plugins/jquery.simplemodal': EMPTY,
 
       // util/BaseRouter -> ConfirmationDialog
-      'ConfirmationDialog': EMPTY,
+      '../views/components/ConfirmationDialog': EMPTY,
```
haishengwu-okta, line 85: so can `buildtools/ConfirmationDialog.ts` be deleted?

haishengwu-okta (approved):
🚀 🚀 🚀

packages/@okta/courage-for-signin-widget/webpack.config.js
aarongranick-okta: yes

PR #2460, opened 2022-04-11
OKTA-481752 a11y fix

praveendurai-okta:
## Description:
`Select your device type` page is skipped when used only keyboard ie "Tab" & "Enter" key.
Root cause- we do have submit button which gets triggered for every "Enter" on any input.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

Yes it will fail in monolith. I will fix it once i get approval for this PR.

### Screenshot/Video:

https://user-images.githubusercontent.com/94395085/162770252-2f9a9231-ec02-4473-8698-2c5c1f33427b.mp4

### Reviewers:

### Issue:

- [OKTA-481752](https://oktainc.atlassian.net/browse/OKTA-481752)

### DIFF
--- a/src/EnrollTotpController.js
+++ b/src/EnrollTotpController.js
@@ -115,7 +115,7 @@ const EnrollTotpControllerEnrollTotpController = FormController.extend({
           attributes: {
             'data-type': 'save'
           },
-          className: 'button button-primary button-wide default-custom-button',
+          className: 'button button-primary default-custom-button',
           click: function() {
             this.model.save();
           },
--- a/src/EnrollTotpController.js
+++ b/src/EnrollTotpController.js
@@ -89,7 +89,6 @@ const EnrollTotpControllerEnrollTotpController = FormController.extend({
 
       return loc('enroll.totp.title', 'login', [factorName]);
     },
-    autoSave: true,
     noButtonBar: true,
     attributes: { 'data-se': 'step-device-type' },
 
@@ -111,9 +110,15 @@ const EnrollTotpControllerEnrollTotpController = FormController.extend({
           View: EnrollTotpControllerAppDownloadInstructionsView,
           showWhen: showWhenDeviceTypeSelected,
         }),
-        FormType.Toolbar({
-          noCancelButton: true,
-          save: loc('oform.next', 'login'),
+        FormType.Button({
+          title: loc('oform.next', 'login'),
+          attributes: {
+            'data-type': 'save'
+          },
+          className: 'button button-primary button-wide default-custom-button',
+          click: function() {
+            this.model.save();
+          },
           showWhen: showWhenDeviceTypeSelected,
         }),
       ];
--- a/assets/sass/modules/_accessibility.scss
+++ b/assets/sass/modules/_accessibility.scss
@@ -10,7 +10,9 @@
 }
 
 .focused-input,
-.link.help:focus {
+.link.help:focus,
+input[type="text"]:focus,
+input[type="radio"]:focus + label {
   box-shadow: 0 0 8px rgba(81, 203, 238, 1);
 }
 


src/EnrollTotpController.js
```diff
@@ -111,9 +110,15 @@ const EnrollTotpControllerEnrollTotpController = FormController.extend({
           View: EnrollTotpControllerAppDownloadInstructionsView,
           showWhen: showWhenDeviceTypeSelected,
         }),
-        FormType.Toolbar({
-          noCancelButton: true,
-          save: loc('oform.next', 'login'),
```
indirathirumalaimurugan-okta, line 116: Q: this triggers the submit operation on any input type other than 'submit' as well? and the current fix is to make the submit operation work only for when attribute is 'save'

src/EnrollTotpController.js
```diff
@@ -111,9 +110,15 @@ const EnrollTotpControllerEnrollTotpController = FormController.extend({
...
+        FormType.Button({
+          title: loc('oform.next', 'login'),
+          attributes: {
+            'data-type': 'save'
+          },
+          className: 'button button-primary default-custom-button',
+          click: function() {
+            this.model.save();
+          },
```
indirathirumalaimurugan-okta, line 121: is there a test?

src/EnrollTotpController.js
praveendurai-okta: Yes

src/EnrollTotpController.js
praveendurai-okta: Both submit button and newly added button have same attribute and hence no change needed in specs. existing spec works good.

indirathirumalaimurugan-okta (approved):
lgtm

src/EnrollTotpController.js
```diff
@@ -111,9 +110,15 @@ const EnrollTotpControllerEnrollTotpController = FormController.extend({
           View: EnrollTotpControllerAppDownloadInstructionsView,
           showWhen: showWhenDeviceTypeSelected,
         }),
-        FormType.Toolbar({
-          noCancelButton: true,
```
pradeepdewda-okta, line 115: getting rid of this after the fix ?

src/EnrollTotpController.js
```diff
@@ -111,9 +110,15 @@ const EnrollTotpControllerEnrollTotpController = FormController.extend({
...
-          save: loc('oform.next', 'login'),
+        FormType.Button({
+          title: loc('oform.next', 'login'),
+          attributes: {
+            'data-type': 'save'
+          },
+          className: 'button button-primary default-custom-button',
+          click: function() {
+            this.model.save();
```
mauriciocastillosilva-okta, line 120: this seems to me is doing the same as the `autoSave` (calling model.save when the form is saved). 
Can you expand on how this solves the issue?

praveendurai-okta: it does same except for `model.save` will be called only on click of this button. If we have` autoSave = true` and toolbar `submit` button, model.save is called behind the scene for every `enter` key event on every `input` element.

praveendurai-okta: https://user-images.githubusercontent.com/94395085/162847455-7ca7a891-1d88-4e0e-83ff-eb3c1e20cd7d.mp4

I have explained the issue here.

mauriciocastillosilva-okta: I think we should make the selection of either iOS or Android mandatory, so the form cannot be submitted without the user selecting an option. This can be done in addition to what you have, but it might also solve the issue on its own.

mauriciocastillosilva-okta: Talked offline about this. It seems that we have extra processing for Radio events (keyup specifically) and that is triggering a form save. So the changes in this PR will avoid that. Let's just add a comment in the code to remember why we had to do this instead of using autoSave

mauriciocastillosilva-okta (approved):

praveendurai-okta:
Merging this PR to okta-core will fail few selenium tests. Fix can be found here `OKTA-481752-selenium-fix-test`

PR #2471, opened 2022-04-17
refines downstream build

aarongranick-okta:
## Description:
Refines downstream build process:

- verify package includes check for beta AuthJS version, will fail if using non published version and will log custom message
- verify package is separated from publish.sh, allows testing of artifact even if built using a beta AuthJS version
- beta check is skipped on downstream branches (build will pass)
- changes order of bacon tasks so publish is at the top
- adds lint to scripts folder
- restores check on user agent (was merged out in previous PR).

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-468777](https://oktainc.atlassian.net/browse/OKTA-468777)

### DIFF
--- a/scripts/buildtools/commands/verify-package.js
+++ b/scripts/buildtools/commands/verify-package.js
@@ -2,7 +2,7 @@ exports.command = 'verify-package';
 exports.describe = 'Verifies that the NPM package has the correct format';
 
 function verifyAuthJSVersion() {
-  if (/^d16t-okta-auth-js-.*/.test(process.env['BRANCH'])) {
+  if (/^d16t-okta-auth-js-.*/.test(process.env.BRANCH)) {
     console.log('Skipping verification of okta-auth-js version for downstream artifact build');
     return;
   }
--- a/scripts/buildtools/commands/verify-package.js
+++ b/scripts/buildtools/commands/verify-package.js
@@ -1,17 +1,32 @@
 exports.command = 'verify-package';
 exports.describe = 'Verifies that the NPM package has the correct format';
-exports.handler = function() {
-  const expect = require('expect');
-  const package = require('../../../package.json');
-  const report = require('../../../test-reports/pack-report.json');
 
-  expect(package.version).toBeTruthy();
+function verifyAuthJSVersion() {
+  if (/^d16t-okta-auth-js-.*/.test(process.env['BRANCH'])) {
+    console.log('Skipping verification of okta-auth-js version for downstream artifact build');
+    return;
+  }
+
+  const version = require('../../../node_modules/@okta/okta-auth-js/package.json').version;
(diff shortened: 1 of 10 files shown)


scripts/buildtools/commands/verify-package.js
```diff
@@ -1,17 +1,32 @@
 exports.command = 'verify-package';
 exports.describe = 'Verifies that the NPM package has the correct format';
-exports.handler = function() {
+
+function verifyAuthJSVersion() {
```
shuowu-okta, line 4: Future refactor idea: we can turn this script into a plugin of whatever bundler we will eventually stay with. Currently, we have build tools in this repo and how they are connected together is not very straight forward, it would be great to have one bundler as the source of truth, then other process can be built around it as loaders or plugins.

Seems like there is a way to link webpack and grunt together: https://webpack.js.org/guides/integrations/#grunt

shuowu-okta (approved):

scripts/update-se-drivers.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
```
lesterchoi-okta, line 1: Couldn't you just split the array?

lesterchoi-okta (approved):

scripts/buildtools/commands/verify-package.js
aarongranick-okta: The `scripts/buildtools/commands` pattern is used in other repos in Okta, including okta-ui. I think the pattern is mean to solve the problem you describe, to have one place that links together the build tools.

We can switch to rollup, but I'm not sure how that would change the verifications we are doing here, which are on the final package, not any individual bundle.

aarongranick-okta: There is a valid argument against loaders/plugins which is that it makes it harder to migrate to a different build tool. If/when we switch to rollup we will need to find/build replacements for every webpack plugin. I think we should be thoughtful with our use of plugins, unless it is directly related to the specific build process and needs to be "in stream", we may be better off with a series of simple scripts which can be run independently.

scripts/update-se-drivers.js
aarongranick-okta: There are 3 lines in this script which exceed max-length. With this PR I am not attempting to modify this script beyond minimum necessary to pass lint, which was added in this PR.

scripts/buildtools/commands/verify-package.js
shuowu: Yeah, as long as we have a pattern to follow to manage all script together, then we are in good place. My main concern is currently we have different build scripts, grunt, webpack, verify-package, locate in different places, it increases the difficult of understanding it and future upgrade.

PR #2477, opened 2022-04-20
adding timeout for device polling OKTA-450250

anatoliiboiarshchykov-okta:
## Description:
- adding timeout for device polling
- Add testcafe test case

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-450250](https://oktainc.atlassian.net/browse/OKTA-450250)


YangChen-okta:
@pradeepdewda-okta @haishengwu-okta it has been a while for this PR, can you help review, thanks!

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
```diff
@@ -36,6 +42,7 @@ const Body = BaseFormWithPolling.extend({
     this.listenTo(this.model, 'error', this.onPollingFail);
     this.doChallenge();
     this.startPolling();
+    setTimeout(this.stopPolling.bind(this), DEFAULT_POLLING_TIMEOUT);
```
YangChen-okta, line 45: Don't we need to cancel this stopPolling function if polling succeeds?

anatoliiboiarshchykov-okta: I do not think so. If polling is succeded, then the user is redirected and this timer will be cleaned

aarongranick-okta: If polling succeeds, then this remediation will be satisfied and widget will proceed to next step in flow - which may or may not be redirect. There should be logic to cleanup this timeout.

anatoliiboiarshchykov-okta: added cleanup

test/unit/spec/v2/view-builder/internals/BaseOktaVerifyChallengeView_spec.js
```diff
@@ -0,0 +1,68 @@
...
+    const deviceChallenge = {
+      domain: 'testDomain',
+      ports: [1111]
+    };
+    jest.spyOn($, 'ajax').mockReturnValue($.Deferred().resolve());
+    jest.spyOn(window, 'clearTimeout');
+
+    init();
+    testContext.view.doLoopback(deviceChallenge);
+    await wait(500);
```
aarongranick-okta, line 62: Instead of wait(), can you use `jest.runAllTimers()` ?

shuowu-okta:
Do we really need the timeout for OV polling? Should the widget render terminal view directly when session timeouts?

anatoliiboiarshchykov-okta:
> Do we really need the timeout for OV polling? Should the widget render terminal view directly when session timeouts?

Since we are recreating transaction polling will last forever. @YangChen-okta please keep me honest here

test/unit/spec/v2/view-builder/internals/BaseOktaVerifyChallengeView_spec.js
anatoliiboiarshchykov-okta: I am using wait here not because of timer, but because of doLoopback function, since it needs some time to run all promise calls

YangChen-okta:
> > Do we really need the timeout for OV polling? Should the widget render terminal view directly when session timeouts?
> 
> Since we are recreating transaction polling will last forever. @YangChen-okta please keep me honest here

@shuowu-okta @anatoliiboiarshchykov-okta Other than the recreating transaction issue, the poll doesn't stop even when error status is returned, https://oktainc.atlassian.net/browse/OKTA-523910
It doesn't have to be the timeout solution to stop polling, any other suggestion @shuowu-okta?

shuowu-okta:
@YangChen-okta widget should only start a new polling request when a new response is available, and the error response should include no polling related remediation. So technically polling should be stopped automatically when session expired error is detected. 

This infinite pooling issue seems like loop has been created when error happen. Imo, the fix should be finding the cause of loop and try to avoid that.

cc: @anatoliiboiarshchykov-okta

test/unit/spec/v2/view-builder/internals/BaseOktaVerifyChallengeView_spec.js
aarongranick-okta: We'd much rather have wait *for condition* instead of wait *for time*.
In this case, it looks like you are expecting clearTimeout to be called at the end of the sequence.
Try:
```
await Expect.waitForSpyCall(window.clearTimeout)
```
instead of waiting 500ms, it will poll in a loop (every 20ms) until the expected condition (clearTimeout has been called) is truthy

PR #2495, opened 2022-04-29
refactor v1 unit tests to "v1" folder

aarongranick-okta:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:


test/unit/legacy-tests.js
```diff
@@ -25,7 +25,7 @@ module.exports = [
   'EnrollYubikey_spec.js',
   'ForgotPassword_spec.js',
   'IDPDiscovery_spec',
```
haishengwu-okta, line 27: Do you need to update path for other specs? or refactor the karma test config a bit so you dont have to change every entry here?

aarongranick-okta: These are run through a simple regex so it does not matter what their path, as long as it's not ambiguous. LoginRouter needs a more specific path because it contains the name of another test (BaseLoginRouter)

PR #2509, opened 2022-05-09
fix: dependencies for TS apps

aarongranick-okta:
## Description:
Puts dependencies needed for Typescript applications in the "dependencies" section (instead of "devDependencies").

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-496970](https://oktainc.atlassian.net/browse/OKTA-496970)


haishengwu-okta (approved):

scripts/verify-package.sh
```diff
@@ -28,4 +28,14 @@ then
   exit ${PUBLISH_TYPE_AND_RESULT_DIR_BUT_ALWAYS_FAIL}
 fi
 
+# Move node_modules out of the way to validate the dist package dependencies
+mv node_modules node_modules2
```
shuowu-okta, line 32: is it possible to run `yarn` inside dist, so it disregard the node_modules in workspace? I think as long as dist is not a module in workspace, it should work.

shuowu-okta (approved):

scripts/verify-package.sh
aarongranick-okta: Nope. NPM, Yarn and TSC will keep walking upwards looking for a folder called node_modules to resolve the dependency. They will even keep walking above the workspace.

aarongranick-okta: Running `yarn` in dist would install the dependencies, but would not help us detect this case where a workspace dev dependency should be a regular dependency, because the module will be installed and findable in node_modules

aarongranick-okta: https://nodejs.org/api/modules.html#loading-from-node_modules-folders

lesterchoi-okta (approved):

PR #2516, opened 2022-05-12
added error handler

praveendurai-okta:
## Description:

Auth-js throws an error when polling is stopped manually. In this case, user is moving to next screen which is a valid scenario and hence no need to show any error to user. 

For now the errors are ignored. We will be doing a cleanup based on the outcome of [OKTA-324849](https://oktainc.atlassian.net/browse/OKTA-324849)

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
For Before Video refer the the attachment in the [bug](https://oktainc.atlassian.net/browse/OKTA-482298).

After fix:

https://user-images.githubusercontent.com/94395085/168682582-ffb4cc51-805a-4a4d-8cd6-66b6e0edfbfa.mov

### Reviewers:
@indirathirumalaimurugan-okta @aarongranick-okta 

### Issue:

- [OKTA-482298](https://oktainc.atlassian.net/browse/OKTA-482298)

### DIFF
--- a/src/v1/controllers/BarcodePushController.js
+++ b/src/v1/controllers/BarcodePushController.js
@@ -54,10 +54,9 @@ export default FormController.extend({
   Footer: Footer,
 
   initialize: function() {
-    this.pollForEnrollment().catch((error)=>{
-      if (!(error instanceof AuthPollStopError)) {
-        this.trigger('error', this, error);
-      }
+    this.pollForEnrollment().catch(()=>{
+      // Ignoring the errors for now. 
+      // clean up will be done based on OKTA-324849
     });
   },
 
--- a/src/v1/controllers/BarcodePushController.js
+++ b/src/v1/controllers/BarcodePushController.js
@@ -10,7 +10,6 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { AuthPollStopError } from '@okta/okta-auth-js';
 import { loc } from 'okta';
 import FactorUtil from 'util/FactorUtil';
 import FormController from 'v1/util/FormController';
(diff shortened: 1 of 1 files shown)


src/v1/controllers/BarcodePushController.js
```diff
@@ -53,7 +54,11 @@ export default FormController.extend({
   Footer: Footer,
 
   initialize: function() {
-    this.pollForEnrollment();
+    this.pollForEnrollment().catch((error)=>{
+      if (!(error instanceof AuthPollStopError)) {
```
indirathirumalaimurugan-okta, line 58: would the instance be other than `AuthPollStopError` when the polling is stopped due to some other means and throwing an error is expected?

aarongranick-okta:
No tests?

src/v1/controllers/BarcodePushController.js
praveendurai-okta: When polling is stopped the only error thrown is `AuthPollStopError`. The other error `AuthSdkError` thrown by [this](https://github.com/okta/okta-auth-js/blob/3f6c3365c419760e652a7fb2a29e2a0575b0d647/lib/tx/poll.ts#L92 ) seems to be a fatal one. So i guess we need to handle it and show it.

praveendurai-okta: @aarongranick-okta What do you think of above? should we handle the other errors?

aarongranick-okta: Isn't catching it and then calling `trigger('error'...` handling it? It should bubble up and be displayed somehow? This is something we can could confirm with a test.

praveendurai-okta: For now the errors are ignored. as discussed, we will be doing a cleanup based on the outcome of OKTA-324849

indirathirumalaimurugan-okta (approved):

aarongranick-okta (approved):

PR #2543, opened 2022-05-26
fixes language override error

spacewar75:
when set to const can't override the locale. We got error in our Okta preview environment.


denysoblohin-okta:
Please explain what error this PR fixes.
What error do you get? Please provide logs or screenshot.

spacewar75:
Our Oktapreview environment is now picking up a different settings.ts file so can't reproduce the error.
image

PR #2544, opened 2022-05-26
fixes dutch/portuguese issue

jaredperreault-okta:
## Description:
SIW language detection logic uses `navigator.languages`, a browser API which returns a _read only_ array of user preferred languages. An attempt is made to manipulate this array (indirectly), but doing so causes an error to be throw

```javascript
Uncaught TypeError: Cannot assign to read only property '0' of object '[object Array]'
```

This PR fixes this issue by cloning the array returned from `navigator.languages`, so it is no longer read-only when consumed.

Issue was introduced by OKTA-491150 changes

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/90656038/170540506-30f11306-5508-4f43-b2b5-7c6f800c49f9.mov

### Issue:
- [OKTA-501779](https://oktainc.atlassian.net/browse/OKTA-501779)

### DIFF
--- a/test/unit/spec/UtilBrowserFeatures_spec.js
+++ b/test/unit/spec/UtilBrowserFeatures_spec.js
@@ -8,7 +8,7 @@ describe('util/BrowserFeatures', () => {
       let result = null;
       try {
         result = BrowserFeatures.getUserLanguages();
-        result[0] = 'foo'
+        result[0] = 'foo';
       }
       catch (err) {
         errorThrown = true;
--- a/test/unit/spec/UtilBrowserFeatures_spec.js
+++ b/test/unit/spec/UtilBrowserFeatures_spec.js
@@ -1,0 +1,20 @@
+import BrowserFeatures from 'util/BrowserFeatures';
+
+describe('util/BrowserFeatures', () => {
+  describe('getUserLanguages', () => {
+    it('returns a non-read-only array on Chrome/Firefox', () => {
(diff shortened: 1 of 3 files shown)


test/unit/spec/UtilBrowserFeatures_spec.js
```diff
@@ -0,0 +1,20 @@
...
+      let errorThrown = false;
+      let result = null;
+      try {
+        result = BrowserFeatures.getUserLanguages();
+        result[0] = 'foo'
+      }
+      catch (err) {
+        errorThrown = true;
+      }
+      expect(errorThrown).toBe(false);
```
shuowu-okta, line 16: how about test the happy path instead, like property of the return value can be changed, which describes the expected behaviour.

shuowu-okta (approved):

test/unit/spec/UtilBrowserFeatures_spec.js
jaredperreault-okta: https://github.com/okta/okta-signin-widget/pull/2544/files/efa8d43212bfdd965bdd6d3ba3d1179e907c0e19#diff-4f8e33ace31dfd3c70ac41d51bc10cccb6a149623fae160bd11dcd7a9dfdd199R11

This does modify a field of the return value - aka happy path

shuowu-okta: I mean try ... catch may not be needed, test fails if error is thrown.

aarongranick-okta (approved):

PR #2546, opened 2022-05-26
Add synthetic imports for Angular apps

mraible:
## Description:

I get a ton of errors when upgrading the Okta CLI’s Angular widget sample to use Sign-In Widget 6.3.3. Here’s a sampling:

```
Error: node_modules/@okta/okta-signin-widget/types/packages/@okta/courage-dist/types/courage/util/handlebars/helper-i18n.d.ts:1:8 - error TS1259: Module '"handlebars"' can only be default-imported using the 'allowSyntheticDefaultImports' flag

1 import Handlebars from 'handlebars';
```

### Reviewers:

@aarongranick-okta 

### Issue:

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -28,6 +28,20 @@ The existing `.widgetrc` file used to configure the Widget has been **removed**.
 
 ## Migrating from 5.x to 6.x
 
+### Enable synthetic imports for TypeScript apps
+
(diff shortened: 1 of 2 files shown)


aarongranick-okta:
@mraible Thanks for the report. You are correct that synthetic imports is required due to transitive dependencies on commonJS modules. We will add a section on Typescript usage to the README.

aarongranick-okta:
internal ref: OKTA-502013

PR #2550, opened 2022-06-03
Remove the background image from the custom-checkbox label class

not-mike-smith:
## Description:
The background image in the label of the `custom-checkbox` css class is showing up in our css outside of the Okta widget, and it does not seem clear that the image should be there in the first place since it seems to be an example of different possible checkbox states.  This PR is to remove that image from the label so it doesn't affect the css of projects which include this project.

## PR Checklist

- [✓] Have you verified the basic functionality for this change?

- [ ] Added unit tests?

- [ ] Added e2e tests

No
### Screenshot/Video:

Image of checkbox with the image:
chechbox-bg-image

Image of corrected checkbox label class rendering:
chechbox-no-bg-image

### Issue:

I didn't create an issue, just the PR

### DIFF
--- a/assets/sass/modules/_forms.scss
+++ b/assets/sass/modules/_forms.scss
@@ -223,7 +223,7 @@
 
 .custom-checkbox {
   label {
-    background-image: url('../img/ui/forms/checkbox-sign-in-widget.png');
+    background-image: none;
   }
 
   label.focus {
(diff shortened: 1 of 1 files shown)


jaredperreault-okta:
Thanks for the submission! We will review this and get back to you!

Internal Ref: OKTA-504410

PR #2552, opened 2022-06-07
fix: do not exclude courage from babel

aarongranick-okta:
## Description:

An issue was found on IE11 caused by untranspiled code from the courage library. This fixes the issue by including courage in the babel transpilation.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-504824](https://oktainc.atlassian.net/browse/OKTA-504824)


aarongranick-okta:
> How about also add https://github.com/amilajack/eslint-plugin-compat against the js bundle to prevent similar issues in the future?

This is a good idea. I've created an issue to follow up and add the extra lint task. OKTA-505055

haishengwu-okta (approved):

PR #2553, opened 2022-06-07
chore: fix dt setup script

shuowu-okta:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -14,6 +14,12 @@ setup_service yarn 1.21.1 /etc/pki/tls/certs/ca-bundle.crt
 
 cd ${OKTA_HOME}/${REPO}
 
+if ! yarn install --ignore-scripts ; then
+  echo "yarn install failed! Exiting..."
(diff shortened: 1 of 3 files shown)


scripts/setup.sh
```diff
@@ -14,6 +14,12 @@ setup_service yarn 1.21.1 /etc/pki/tls/certs/ca-bundle.crt
 
 cd ${OKTA_HOME}/${REPO}
 
+if ! yarn install --ignore-scripts ; then
```
jaredperreault-okta, line 17: I think you'll want to put a `yarn build` call at the end of this file. `yarn install` triggers build via `prepare`, however `--ignore-scripts` will skip that part. Also a build as to happen _after_ the upstream auth-js artifact is installed, but also independent of the upsteam logic

reference: https://github.com/okta/okta-react/pull/235/files#diff-ce435f38623d032d25670f2c63a430dfd1a02db661d10c1ef031cbef2f0f0ef0R83

PR #2571, opened 2022-06-16
fix: remove invalid values from self service registration request

jonathanhansen-okta:
## Description:
- Self-Service Registration (classic) is using null values in the registration request for some optional fields that are not populated in the registration form. This causes a NPE in okta-core
   - Add logic which removes any null, undefined, or empty string fields from the request

- [Green Monolith Bacon](https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&branch=OKTA-495844-inline-hook-bug&page=1&pageSize=6&sha=b27a53cadd5382af8b7657b71d17a7eef5f497e7&tab=main)

## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

- Did you verify the change by running downstream monolith artifacts? (**YES**)

### Screenshot/Video:
* [Video - Before](https://okta.app.box.com/file/972244347977)
* [Video - After](https://okta.app.box.com/file/972244442523)

### Reviewers:
@vicentevillegas-okta @justinabrokwah-okta @okta/ciamx 

### Issue:

- [OKTA-495844](https://oktainc.atlassian.net/browse/OKTA-495844)

### DIFF
--- a/src/v1/controllers/RegistrationController.js
+++ b/src/v1/controllers/RegistrationController.js
@@ -94,6 +94,8 @@ export default BaseLoginController.extend({
   registerUser: function(postData) {
     const self = this;
 
+    Object.keys(postData).forEach((k) =>
+      (_.isNull(postData[k]) || _.isUndefined(postData[k]) || postData[k] === '') && delete postData[k]);
     this.model.attributes = postData;
     // Model.save returns a jqXHR
     Backbone.Model.prototype.save
--- a/test/unit/spec/v1/Registration_spec.js
+++ b/test/unit/spec/v1/Registration_spec.js
@@ -1037,4 +1037,100 @@ Expect.describe('Registration', function() {
         });
     });
   });
+
+  const sanitizeSetting = function() {
+    const preSubmitSpy = jasmine.createSpy('preSubmitSpy');
+    return {
+      registration: {
+        preSubmit: function(postData, onSuccess, onFailure) {
+          preSubmitSpy(postData, onSuccess, onFailure);
+          onSuccess(postData);
+        },
+        postSubmit: jasmine.createSpy('postSubmitSpy'),
+      },
+    };
+  };
+
+  Expect.describe('sanitizes invalid request values', function() {
+    itp('null value', function() {
+      const setting = sanitizeSetting();
+      return setup(setting).then(function(test) {
+        Util.resetAjaxRequests();
+        test.form.setUserName('test@example.com');
+        test.form.setPassword('Abcd1234');
+        test.form.setFirstname('firstName');
+        test.form.setLastname(null);
+        test.form.submit();
+        const model = test.router.controller.model;
+
+        spyOn(Backbone.Model.prototype, 'save').and.returnValue($.Deferred().resolve());
+        model.save();
+        Util.callAllTimeouts();
+        expect(test.router.controller.model.get('userName')).toBe('test@example.com');
+        expect(test.router.controller.model.get('firstName')).toBe('firstName');
+        expect(test.router.controller.model.has('lastName')).toBe(false);
+        expect(setting.registration.postSubmit).toHaveBeenCalled();
+      });
+    });
+    itp('undefined value', function() {
+      const setting = sanitizeSetting();
+      return setup(setting).then(function(test) {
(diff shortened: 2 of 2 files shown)


src/v1/controllers/RegistrationController.js
```diff
@@ -94,6 +94,7 @@ export default BaseLoginController.extend({
   registerUser: function(postData) {
     const self = this;
 
+    Object.keys(postData).forEach((k) => postData[k] === null && delete postData[k]);
```
vicentevillegas-okta, line 97: Should we assert for `undefined` too?
`Object.keys(postData).forEach((k) => (_.isNull(postData[k]) || _.isUndefined(postData[k]) && delete postData[k]);`

Please add a test for this.

vicentevillegas-okta (changes requested):
it looks good to me, just a small comment.

src/v1/controllers/RegistrationController.js
vicentevillegas-okta: This is fixing a bug, and it means we do not have a test for this specific scenario, otherwise we should have catched this before. Should we add an e2e test too?

Please add a video/screenshot for the fix.

One more thing, since this is fixing a bug, I think you will need to base your branch on the latest release branch I think it is `6.5` . One more thing, you will need to run downstream test on monolith.

vicentevillegas-okta: I agree with you, e2e will be too much 👍

justinabrokwah-okta: +1 to changing your target branch to `6.4` since this is a bug fix! Also agree with checking for undefined and adding a unit test

vicentevillegas-okta: @jonathanhansen-okta I have one more question, is that possible that the `postData[k]` could be empty string? is this something we need to validate too? I am just curios

jonathanhansen-okta: Addressed all your comments
- added check for undefined and empty
- added unit tests
- added video for fix
- added monolith bacon test results
- changed target branch to 6.4

test/unit/spec/v1/Registration_spec.js
```diff
@@ -1037,4 +1037,126 @@ Expect.describe('Registration', function() {
...
+      const preSubmitSpy = jasmine.createSpy('preSubmitSpy');
+      const setting = {
+        registration: {
+          preSubmit: function(postData, onSuccess, onFailure) {
+            preSubmitSpy(postData, onSuccess, onFailure);
+            onSuccess(postData);
+          },
+          postSubmit: jasmine.createSpy('postSubmitSpy'),
+        },
+      };
```
justinabrokwah-okta, line 1051: Since this segment is reused a few times, is it possible for us to extrapolate it?

test/unit/spec/v1/Registration_spec.js
```diff
@@ -1037,4 +1037,126 @@ Expect.describe('Registration', function() {
...
+
+      return setup(setting).then(function(test) {
+        Util.resetAjaxRequests();
+        test.form.setUserName('test@example.com');
+        test.form.setPassword('Abcd1234');
+        test.form.setFirstname('firstName');
+        test.form.setLastname('lastName');
```
justinabrokwah-okta, line 1148: Curious, would it be possible here to add a boolean field to the form and verify the logic works on it too? Just asking since that's where the bug originated from

justinabrokwah-okta: Discussed offline, resolving !

test/unit/spec/v1/Registration_spec.js
jonathanhansen-okta: Done

justinabrokwah-okta (approved):
LGTM

src/v1/controllers/RegistrationController.js
vicentevillegas-okta: Thank you so much @jonathanhansen-okta, great work!!

vicentevillegas-okta (approved):
LGTM 👍 
make sure to ask your review at `#eng-sign-in-widget`

PR #2588, opened 2022-07-05
enables IE11 e2e testing

jaredperreault-okta:
## Description:
Enables IE11 e2e browser testing of the SIW via SauceLabs

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:
- OKTA-504526

### DIFF
--- a/scripts/e2e-saucelabs.sh
+++ b/scripts/e2e-saucelabs.sh
@@ -12,14 +12,12 @@ get_vault_secret_key devex/sauce-labs accessKey SAUCE_ACCESS_KEY
 export TEST_SUITE_TYPE="junit"
 export TEST_RESULT_FILE_DIR="${REPO}/build2/reports/junit"
 
-# This file contains all the env vars we need for e2e tests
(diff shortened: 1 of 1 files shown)


vijetmahabaleshwar-okta (approved):

test/app/.babelrc
```diff
@@ -1,13 +1,10 @@
...
-          "esmodules": true
-        }
-      }
-    ],
-    "@babel/preset-typescript"
+    "@babel/preset-typescript",
+    "@babel/preset-env"
+  ],
+  "plugins": [
+    "@babel/plugin-transform-runtime",
```
aarongranick-okta, line 7: Can we convert this file to `babel.config.js` and add the `transform-runtime` plugin conditionally based on an environment variable? This plugin makes debugging very difficult, it is much better for our DevEx if we can exclude it by default and only include it where needed (like IE11)

aarongranick-okta (approved):

PR #2597, opened 2022-07-12
feat(IdentifierView): show custom access denied error message

sherizada-okta:
## Description:
- new i18n key is gated on backend by ff
- renders custom error message callout:
1. identify view
Screen Shot 2022-07-25 at 9 39 18 AM

2. terminal view
Screen Shot 2022-07-25 at 9 38 50 AM

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

- [x] Did you verify the change by running downstream monolith artifacts? (YES | NO | UNSURE)

### Issue:

- [OKTA-514055](https://oktainc.atlassian.net/browse/OKTA-514055)

### DIFF
--- a/playground/mocks/data/idp/idx/error-identify-access-denied-custom-message.json
+++ b/playground/mocks/data/idp/idx/error-identify-access-denied-custom-message.json
@@ -1,0 +1,19 @@
+{
+  "stateHandle": "01OCl7uyAUC4CUqHsObI9bvFiq01cRFgbnpJQ1bz82",
+  "version": "1.0.0",
+  "expiresAt": "2020-05-18T20:42:23.000Z",
+
+  "intent": "LOGIN",
+  "messages": {
+    "type": "array",
+    "value": [
+      {
+        "message": "You do not have <a href='#'>permission</a> to perform the requested action.",
+        "i18n": {
+          "key": "security.access_denied_custom_message"
+        },
+        "class": "ERROR"
+      }
+    ]
+  }
+}
--- a/playground/mocks/data/idp/idx/error-identify-access-denied-custom-message.json
+++ b/playground/mocks/data/idp/idx/error-identify-access-denied-custom-message.json
@@ -8,7 +8,11 @@
     "type": "array",
     "value": [
       {
-        "message": "You do not have <a href='#'>permission</a> to perform the requested action.",
+        "message": "You do not have permission to perform the requested action.",
+        "links": [
+          {"label": "Help link 1", "url": "https://www.okta.com/"},
+          {"label": "Help link 2", "url": "https://www.okta.com/help?page=1"}
+        ],
         "i18n": {
           "key": "security.access_denied_custom_message"
         },
--- a/src/v2/controllers/FormController.ts
+++ b/src/v2/controllers/FormController.ts
@@ -350,10 +350,8 @@ export default Controller.extend({
       errorObj = { responseJSON: { errorSummary: loc('error.unsupported.response', 'login')}};
     }
 
-    this.options.appState.set('messages', idxStateError.messages)
-
     if(_.isFunction(form?.showCustomFormErrorCallout)) {
-      showErrorBanner = !form.showCustomFormErrorCallout(errorObj);
+      showErrorBanner = !form.showCustomFormErrorCallout(errorObj, idxStateError.messages);
     }
 
     // show error before updating app state.
--- a/src/v2/controllers/FormController.ts
+++ b/src/v2/controllers/FormController.ts
@@ -350,6 +350,8 @@ export default Controller.extend({
       errorObj = { responseJSON: { errorSummary: loc('error.unsupported.response', 'login')}};
     }
 
+    this.options.appState.set('messages', idxStateError.messages)
+
     if(_.isFunction(form?.showCustomFormErrorCallout)) {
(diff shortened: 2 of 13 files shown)


src/v2/view-builder/views/IdentifierView.js
```diff
@@ -160,6 +164,21 @@ const Body = BaseForm.extend({
...
+
+    const { errorSummary } = error.responseJSON;
+    const options = {
+      type: 'error',
+      content: new HtmlView({ message: errorSummary }),
+    };
+
+    this.showMessages(createCallout(options));
+    return true;
+  },
```
kamaldeepcheema-okta, line 180: curious, where is this new function invoked? I don't see any references of it in this PR.

kamaldeepcheema-okta (approved):

sherizada-okta:
> Can you include a video of the change and merge this into `master` rather than `6.5`. This seems like a feature rather than a bug fix

The i18n key is behind a feature flag in okta-core

aarongranick-okta (approved):

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -160,6 +164,28 @@ const Body = BaseForm.extend({
     return newSchemas;
   },
 
+  showCustomFormErrorCallout(error) {
+    const {
+      errorSummaryKeys,
+      errorSummary,
+    } = error.responseJSON;
+
+    const isPossibleXss = errorSummary.includes('script') || errorSummary.includes('data');
```
lesterchoi-okta, line 173: Extract to function and improve truthy conditions. This implementation may cause false positives/negatives.

lesterchoi-okta (changes requested):
Consider finding a lib/tool that performs XSS prevention either via escaping or detection

src/v2/controllers/FormController.ts
```diff
@@ -349,6 +349,8 @@ export default Controller.extend({
       errorObj = { responseJSON: { errorSummary: loc('error.unsupported.response', 'login')}};
     }
 
+    this.options.appState.set('messages', idxStateError.messages)
```
sherizada-okta, line 352: I need access to the original messages object to retrieve some custom data. See mock response [here](https://github.com/okta/okta-signin-widget/pull/2597/commits/b5deef5bd25891189024c6ba855db92499daea60#diff-cf39648fe1b124ef6de42e035db9fbe3d107450fd14c082441ad0100e2c021ccR12).

sherizada-okta: Alternatively, could pass `messages` to `showCustomFormErrorCallout` on [line 355](https://github.com/okta/okta-signin-widget/pull/2597/commits/b5deef5bd25891189024c6ba855db92499daea60#diff-72a0e868d27f4dfd26d36bcba1d860aa25999401fa00788f3ac694304043dd4dR355)

lesterchoi-okta (changes requested):
I think this'll cause an english leak as is?
Could you run the `detect-english-leak` suite?

playground/mocks/data/idp/idx/error-identify-access-denied-custom-message.json
```diff
@@ -0,0 +1,23 @@
...
+
+  "intent": "LOGIN",
+  "messages": {
+    "type": "array",
+    "value": [
+      {
+        "message": "You do not have permission to perform the requested action.",
+        "links": [
+          {"label": "Help link 1", "url": "https://www.okta.com/"},
+          {"label": "Help link 2", "url": "https://www.okta.com/help?page=1"}
```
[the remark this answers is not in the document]
sherizada-okta: It can't be localized because it's a custom string. The user can put whatever they want as these labels.

src/v2/view-builder/views/IdentifierView.js
```diff
@@ -160,6 +164,25 @@ const Body = BaseForm.extend({
...
+      return false;
+    }
+
+    const { errorSummary } = error.responseJSON;
+
+    const options = {
+      type: 'error',
+      content: new CustomAccessDeniedErrorMessage({
+        message: errorSummary,
+        links: messages[0]?.links,
```
lesterchoi-okta, line 178: There are cases where the [IDX response will contain more than one message](https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/data/idp/idx/errors-authenticator-verification-okta-verify-reject-push.json#L8-L38), will that ever happen with a custom access denied error message?

lesterchoi-okta (approved):

src/v2/view-builder/views/IdentifierView.js
lesterchoi-okta: Approved PR so you're not stuck, but would consider whether the multi-message response scenario could apply. If the custom error message is not the first message in the response, it'll also cause an incorrect render (raw html, again)

sherizada-okta: I'll grab the message with the right key

playground/mocks/data/idp/idx/error-identify-access-denied-custom-message.json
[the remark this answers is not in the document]
sherizada-okta: How are they? I don't know of any custom strings that come from the API like this. We have them for the helpLinks below the login form, but those come from Settings.

PR #2612, opened 2022-07-20
OKTA-514042 client side validation

shuowu-okta:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-514042](https://oktainc.atlassian.net/browse/OKTA-514042)


src/v3/src/mocks/response/idp/idx/introspect/default.json
```diff
@@ -17,7 +17,8 @@
         "value": [
           {
             "name": "identifier",
-            "label": "Username"
+            "label": "Username",
+            "required": true
```
shuowu-okta, line 21: update to latest idx response

src/v3/src/transformer/email/emailVerificationTransformer.ts
```diff
@@ -67,10 +67,8 @@ export const transformEmailVerification: IdxStepTransformer = (transaction, form
     scope: `#/properties/${ButtonType.SUBMIT}`,
     options: {
       type: ButtonType.SUBMIT,
-      idxMethodParams: {
-        authenticator: {
-          methodType: methodTypeElement.options.inputMeta.options?.[0].value as string,
-        },
+      actionParams: {
+        'authenticator.methodType': methodTypeElement.options.inputMeta.options?.[0].value as string,
```
shuowu-okta, line 71: flatten params for easier lookup in validation step

shuowu-okta: forced in types

src/v3/src/transformer/field/transform.ts
```diff
@@ -52,8 +52,27 @@ export const transformStepInputs: StepTransformer = (step?: NextStep): FormBag =
...
+          validate(data) {
+            const isValid = !!data[name];
+            return isValid ? undefined : {
+              i18n: {
+                key: 'model.validation.field.blank',
+              },
+            };
+          },
+        };
+      }
```
glenfannin-okta, line 74: Instead of introducing `dataSchema`, can you build up the schema property that already exists? 

It would actually be extremely simple, since this is only handling `required` properties, you can build it up like so: 

```
if (required && mutable !== false) {
   acc.schema.properties[name] = {
      type: input.type
   }
   acc.schema.required.push(name);
}
```

Then, you can purely validate based on the required properties (using a simple function). But this will adhere to JsonSchema architecture and make it extremely easy to drop in AJV when we are ready.

src/v3/src/components/Form/Form.tsx
```diff
@@ -49,17 +53,53 @@ const getSubmitButtonSchema: any = (uischema: UISchemaLayout, stepperStepIndex:
...
+            const message = elementSchema.validate({
+              ...data,
+              ...params,
+            });
+            if (message) {
+              acc[name] = message;
+            }
+          }
+          return acc;
+        }, {});
```
glenfannin-okta, line 88: See my comment [here](https://github.com/okta/okta-signin-widget/pull/2612/files#diff-5909dabad92ba072e785e92903b1509f10099132412da3cd25f3c2537c38fc41R64) that will simplify this validation function. Ideally, we should just need to pass in the `data` and the `schema` from `FormBag` and allow it to validate it based on the rules we have set in place (in this case only need to look at the required property).

src/v3/src/types/schema.ts
```diff
@@ -23,6 +23,15 @@ import { FunctionComponent } from 'preact';
 import { PasswordSettings } from './password';
 import { UserInfo } from './userInfo';
 
+export type FormBag = {
+  schema: Record<string, unknown>;
+  uischema: UISchemaLayout;
+  data: Record<string, unknown>;
+  // temp schema bag to handle client validation and form submission
+  dataSchema: Record<string, DataSchema>;
```
glenfannin-okta, line 31: From my previous comments we can just use schema and forego introducing a new custom field and type to keep things simple and easily scalable.

src/v3/src/util/flattenInputs.ts
```diff
@@ -12,19 +12,34 @@
 
 import { Input } from '@okta/okta-auth-js';
 
-export const flattenInputs = (input: Input, name = ''): Input[] => {
-  const res: Input[] = [];
-  const { value } = input;
+export const flattenInputs = (input: Input): Input[] => {
```
glenfannin-okta, line 15: The comments are great, it would be nice if there was a unit test for this that displayed the various scenarios (calling flattenInputs on credentials object, on identifier field, on fields that are mutable but not visible, etc).

src/v3/src/types/schema.ts
shuowu-okta: cc: @lesterchoi-okta

src/v3/src/components/Form/Form.tsx
shuowu-okta: let's not over refactor for now before reaching to a proper design of client side validation, again, it's a temp solution for now, and we will want it to be simple.

shuowu-okta: If we are not using AJV eventually, I would prefer a schema driven design with proper validators functions in a separate module, like `isRequiredValidator`, `minLengthValidator`, so schema can match to proper validators to generate the eventual messages. But since we still have some uncertainty and consider the effort to get there for current stage, let's focus on a half way solution for now. By half way I mean, it's on the same direction, and it requires minimal change to get to an ideal solution.

glenfannin-okta: I think it was mentioned that we want to use AJV as its a standard moving forward because other teams are already adopting JsonSchema. 

Can you take a look [here](https://github.com/okta/okta-signin-widget/pull/2600/files#diff-ddc62c2eea3f73a53b09fe54d85aaf4bff5d173f9f35d3f03bb5bda1499dd277R60-R66) and [here](https://github.com/okta/okta-signin-widget/pull/2600/files#diff-5909dabad92ba072e785e92903b1509f10099132412da3cd25f3c2537c38fc41R40-R54) I already experimented what it would take to implement the jsonschema and AJV. It requires very little changes but most importantly its standardized. 

The original ticket for this was for a bug I was working on, I can continue that and include the necessary changes. 

@lesterchoi-okta can chime in if I am off base but I thought AJV w/ JsonSchema was the ultimate end goal (in a perfect world :D )?

shuowu-okta: @glenfannin-okta feel free to set up a quick meeting and include engineers that you think are required if further discussion is needed. As I mentioned, this is a temp solution to unblock bug bash, and I think the team is ok with this.

glenfannin-okta (approved):

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -40,20 +41,30 @@ const InputPassword: UISchemaElementComponent<{
...
+        }}
+      />
+      {error && (
+        <FormHelperText
+          data-se={`${name}-error`}
+          error
+        >
+          {error}
+        </FormHelperText>
+      )}
```
shawyu-okta, line 65: The `PasswordInput` should end up handling the error directly. We can remove this additional piece one it's fixed. Add a comment with this ticket number: OKTA-514722

src/v3/src/components/Spinner/Spinner.tsx
```diff
@@ -30,6 +30,7 @@ const Spinner: FunctionComponent<{
       flexDirection="column"
       justifyContent="center"
       alignItems="center"
+      data-se="o-spinner"
```
shawyu-okta, line 33: Should the data-se go on the Box or on `CircularProgress` directly?

src/v3/src/transformer/phone/phoneEnrollmentTransformer.test.ts
```diff
@@ -67,7 +60,10 @@ describe('PhoneEnrollmentTransformer Tests', () => {
     expect(updatedFormBag).toMatchSnapshot();
   });
 
-  it('should validate phoneNumber format', () => {
+  // TODO: re-visit when add phone number client side validation
+  // it's not safe to use one client side regext to validate, since phone number standard varies by regions
+  // before a good hint is available from idx response, it would be safe to just depend on backend validation
+  it.skip('should validate phoneNumber format', () => {
```
shawyu-okta, line 66: What is the v2 behavior? If we currently attempt to do client side phone validation, we still should attempt to do so in v3 but let the backend do final validation.

src/v3/test/integration/util/createAuthClient.ts
```diff
@@ -10,6 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
shawyu-okta, line 13: nit: preference for using this disable per-line. i counted 5 instances in the whole file, so it's not too bad

src/v3/src/components/Spinner/Spinner.tsx
shuowu-okta: this one was added to verify loading state in jest, not for testcafe parity. I am ok with either way.

src/v3/src/transformer/phone/phoneEnrollmentTransformer.test.ts
shuowu-okta: v2 only do client side validation based on `required` attr

src/v3/test/integration/util/createAuthClient.ts
shuowu-okta: used here in file scope mostly because this is only a test util function

src/v3/test/integration/enroll-profile.test.tsx
```diff
@@ -66,4 +66,44 @@ describe('enroll-profile', () => {
...
+
+    await user.click(submitButton);
+    const firstNameError = await findByTestId('userProfile.firstName-error');
+    expect(firstNameError.textContent).toEqual('This field cannot be left blank');
+    const lastNameError = await findByTestId('userProfile.lastName-error');
+    expect(lastNameError.textContent).toEqual('This field cannot be left blank');
+    const emailError = await findByTestId('userProfile.email-error');
+    expect(emailError.textContent).toEqual('This field cannot be left blank');
+
+    expect(authClient.options.httpRequestClient).not.toHaveBeenCalledWith(
```
shawyu-okta, line 88: nit: is it possible to change assertion to simply check that the client was not called?

shuowu-okta: it is called by other request, like /interact, so we need to be specific here

shawyu-okta: Ah ok, I had a feeling that was the case.

I still find this assertion brittle since it is quite easy for a request to pass this, so if there actually is a request fired but we missed matching a single field correctly below, the test will still pass. Can we write a more robust assertion to check this?

shuowu-okta: not sure how, we can investigate in later stage, have a test like this is better than nothing

src/v3/src/components/Spinner/Spinner.tsx
shawyu-okta: I thought you added this for a testcafe test. If this is only for jest, then update the test rather than add a `data-se` if possible.

src/v3/src/transformer/phone/phoneEnrollmentTransformer.test.ts
shawyu-okta: If that's the case, we can delete client-side validation code for phone and remove this test

src/v3/src/components/Spinner/Spinner.tsx
shuowu-okta: we will need a selector as id, that's what suggested by testing library. also, what does an extra data-se hurt?

src/v3/src/components/InputPassword/InputPassword.tsx
shuowu-okta: tests should be above to catch this issue when mui/odyssey package upgrade, that should be easier to catch than a jira comment

src/v3/src/components/Spinner/Spinner.tsx
shawyu-okta: > also, what does an extra data-se hurt?

The added `data-se` isn't necessary. Why add it?

We can use `ByLabelText` https://testing-library.com/docs/queries/bylabeltext

RTL docs recommend using `ByTestId` as the last resort, so if we have an option to query by something else we should make an effort to do so: https://testing-library.com/docs/queries/about/#priority

shuowu-okta: does a spinner have a label text?

src/v3/test/integration/enroll-profile.test.tsx
shawyu-okta: I did a quick test and it turns out this assertion actually passes:
```js
    expect(authClient.options.httpRequestClient).not.toHaveBeenCalled();
```
It appears that the request client is not ever called, at least for this particular case. You can see it by inspecting `authClient.options.httpRequestClient.mock.calls` in the test.

src/v3/src/components/Spinner/Spinner.tsx
shawyu-okta: Yes, `aria-label` is counted, see the RTL docs linked above.

src/v3/test/integration/enroll-profile.test.tsx
```diff
@@ -66,4 +66,26 @@ describe('enroll-profile', () => {
...
+      'POST',
+      'https://oie-8425965.oktapreview.com/idp/idx/enroll/new',
+      expect.anything(),
```
shawyu-okta, line 88: 👌

PR #2614, opened 2022-07-20
OKTA-508875-add v3 CI tasks to bacon

shuowu-okta:
## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-516689](https://oktainc.atlassian.net/browse/OKTA-516689)


scripts/e2e-v3.sh
```diff
@@ -0,0 +1,29 @@
...
+export TEST_SUITE_TYPE="jsunit"
+export TEST_RESULT_FILE_DIR=${REPO}/src/v3/build2/reports/e2e
+echo $TEST_SUITE_TYPE > $TEST_SUITE_TYPE_FILE
+echo $TEST_RESULT_FILE_DIR > $TEST_RESULT_FILE_DIR_FILE
+
+source ${OKTA_HOME}/${REPO}/scripts/setup.sh
+
+cd ${OKTA_HOME}/${REPO}
+
+setup_service google-chrome-stable 91.0.4472.77-1
```
lesterchoi-okta, line 12: note: v2 uses `setup_service google-chrome-stable 83.0.4103.61-1`

src/v3/src/mocks/scenario/development.ts
```diff
@@ -13,7 +13,7 @@
 import { scenario } from '../registry';
 
 const getRouteUrl = (url: string) => {
-  const parts = url.split('/idp/idx/', 1);
+  const parts = url.split('/idp/idx/');
```
lesterchoi-okta, line 16: hopefully there's no difference

lesterchoi-okta (approved):

src/v3/src/mocks/scenario/development.ts
shuowu-okta: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#parameters

when limit is set as 1, it only keeps i item in array, need to change it back to fix e2e failures

scripts/e2e-v3.sh
shuowu-okta: Yea, also noticed that, I feel it's most probably not get updated. Also feel using newer version might be better choice, since it exposes latest browser changes to tests, like cookie related changes. I think we can discuss more on this in office hour next week.

jaredperreault-okta (approved):

PR #2623, opened 2022-07-22
OKTA-516853 - transformation layer refactor setup 

shuowu-okta:
## Description:

Changes:
* created a new folder (transformernew) with all the refactor code
* added a switch logic in widget/index.ts to turn on new transformation code based on remediation name (IDX_STEP)
* the target for this PR is to have green bacon for Identify step

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-516853](https://oktainc.atlassian.net/browse/OKTA-516853)

### DIFF
--- a/src/v3/src/transformernew/attribute/transform.test.ts
+++ b/src/v3/src/transformernew/attribute/transform.test.ts
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2022-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { Input } from '@okta/okta-auth-js';
+
+import { transformField } from './transform';
+
+describe('Attributes transformer', () => {
+  it('should return uischema object with options.attributes.autocomplete === "username" if formfield.name === "identifier" is present in ion object', () => {
+    const formfield: Input = { name: 'identifier' };
+
+    const result = { attributes: { autocomplete: 'username' } };
+
+    expect(transformField(formfield)).toEqual(result);
+  });
+
+  it('should return uischema object with options.attributes.autocomplete === "current-password" if formfield.name === "password" is present in ion object', () => {
+    const formfield: Input = { name: 'password' };
+
+    const result = { attributes: { autocomplete: 'current-password' } };
+
+    expect(transformField(formfield)).toEqual(result);
+  });
+
+  it('should return uischema object with options.attributes.autocomplete === "current-password" if formfield.name === "newPassword" is present in ion object', () => {
+    const formfield: Input = { name: 'newPassword' };
+
+    const result = { attributes: { autocomplete: 'current-password' } };
+
+    expect(transformField(formfield)).toEqual(result);
+  });
+
+  it('should return uischema object with options.attributes.autocomplete === "current-password" if formfield.name === "passcode" & formfield.secret === true is present in ion object', () => {
+    const formfield: Input = { name: 'credentials.passcode', secret: true };
+
+    const result = { attributes: { autocomplete: 'current-password' } };
+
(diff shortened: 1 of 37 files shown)


src/v3/src/transformernew/attribute/transform.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    const result = { attributes: { autocomplete: 'current-password' } };
+
+    expect(transformField(formfield)).toEqual(result);
+  });
+
+  it('should return uischema object with options.attributes.autocomplete === "current-password" if formfield.name === "newPassword" is present in ion object', () => {
+    const formfield: Input = { name: 'newPassword' };
+
+    const result = { attributes: { autocomplete: 'current-password' } };
```
lesterchoi-okta, line 37: shouldn't autocomplete be `"new-password"`?

src/v3/src/transformernew/attribute/transform.ts
```diff
@@ -0,0 +1,82 @@
...
+  InputAttributes, 
+  Nullable,
+  FormBag,
+  FieldElement,
+} from '../../types';
+
+const autocompleteValueMap = new Map<string, AutoCompleteValue>([
+  ['identifier', 'username'],
+  ['password', 'current-password'],
+  ['newPassword', 'current-password'],
```
lesterchoi-okta, line 25:
```suggestion
  ['newPassword', 'new-password'],
```

src/v3/src/transformernew/attribute/transform.ts
```diff
@@ -0,0 +1,82 @@
...
+  inputName: string,
+): string | undefined => {
+  let isMatch;
+  map.forEach((value: AutoCompleteValue, key: string) => {
+    if (inputName.match(key)?.length) {
+      isMatch = key;
+    }
+  });
+  return isMatch;
+};
```
lesterchoi-okta, line 45: Reverse lookup? Why is this needed?

src/v3/src/transformernew/attribute/transform.ts
```diff
@@ -0,0 +1,82 @@
...
+  const { options: { inputMeta: input } } = field;
+  // passcode name is shared with password + totp code types, only differ by secret
+  if (input.name === 'credentials.passcode' && input.secret) {
+    return autocompleteValueMap.get('password') ?? null;
+  }
+  const key = getKeyFromMap(autocompleteValueMap, input.name);
+  if (key) {
+    return autocompleteValueMap.get(key) ?? null;
+  }
+  return null;
```
lesterchoi-okta, line 57:
```suggestion
  return autocompleteValueMap.get(input.name) ?? null;
```

src/v3/src/transformernew/attribute/transform.test.ts
shuowu-okta: @glenfannin-okta can you confirm?

src/v3/src/transformernew/attribute/transform.ts
shuowu-okta: @glenfannin-okta ^

src/v3/src/transformernew/attribute/transform.test.ts
glenfannin-okta: Actually, this value does not matter, `newPassword` doesnt seem to be an actual value now that we are using generic remediator version of auth-js. Seems to always be `credentials.passcode` no matter the flow. So this line can be removed completely. We actually update the autocomplete value in the custom transformer because there is no way to decipher what it should be at this level.

src/v3/src/transformernew/attribute/transform.ts
glenfannin-okta: So, this was done because the keys that are passed to this function are flattened into a string. 

So for example phoneNumber will be presented as `authenticator.phoneNumber` or password will be presented as `credentials.passcode`, otp will either come as `credentials.passcode` or `totp`, etc. 

However, I don't know of every format phoneNumber can come, because it is present when registering as well as when authenticating as well as when setting up a new phone for an authenticator. 

Because of these variables, I just rely on the last part of the field name. So to do that, I have to perform a match on the key of the map (which is what this function does). 

If you are confident the field names are standardized and there are no variances then we can get rid of it but I was not confident in making that assumption which is why I created this lookup function.

src/v3/src/transformernew/attribute/transform.ts
glenfannin-okta: If this is changed will need to update the keys of the map to include all possible variations of field names (in their flattened format), i.e. `credentials.passcode`, `authenticator.phoneNumber`, `authenticator.firstName`, etc. and be wary of duplicate keys.

PR #2625, opened 2022-07-22
fix: Fixes a UI overlapping when changing the language

nikolaystoilov-okta:
Fixed the issue when changing the language will result is some cases with overlapped UI elements

Resolves: OKTA-507211

## Description:
The problem was related when the promise to remove the beacon was interleaved with loading event and the parts to remove the beacon was executed after the beacon was set to loading.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
Before my change:
Screen Shot 2022-07-22 at 4 51 05 PM
https://okta.enterprise.slack.com/files/U7KV2GCN8/F03QH13UQKG/ui_overlap.mov

Here is a link for the same test after my change:
https://okta.enterprise.slack.com/files/U7KV2GCN8/F03QQ61US2J/after_change.mov
### Issue:

- [OKTA-507211](https://oktainc.atlassian.net/browse/OKTA-507211)

### DIFF
--- a/src/v1/views/shared/Header.ts
+++ b/src/v1/views/shared/Header.ts
@@ -39,39 +39,9 @@ function addBeacon(view, NextBeacon, selector, options) {
   view.currentBeacon = view.first();
 }
 
-function typeOfTransition(currentBeacon, NextBeacon, options) {
-  if (!currentBeacon && !NextBeacon) {
-    return 'none';
-  }
-  // Show Loading beacon
-  if (!currentBeacon && options.loading) {
-    return 'load';
-  }
-  // Swap/Hide Loading beacon
-  if (currentBeacon && isLoadingBeacon(currentBeacon)) {
-    return NextBeacon ? 'swap' : 'unload';
-  }
-  if (currentBeacon && currentBeacon.equals(NextBeacon, options)) {
-    return 'same';
-  }
-  if (!currentBeacon && NextBeacon) {
-    return 'add';
-  }
-  if (currentBeacon && !NextBeacon) {
-    return 'remove';
-  }
-  if (currentBeacon instanceof NextBeacon) {
-    return 'fade';
-  }
-  // If none of the above
-  // then we are changing the type of beacon
-  // ex. from SecurityBeacon to FactorBeacon
-  return 'swap';
-}
-
 export default class Header extends View {
   currentBeacon;
-  
+
   preinitialize(...args) {
     this.currentBeacon = null;
     /* eslint-disable @okta/okta/no-unlocalized-text-in-templates */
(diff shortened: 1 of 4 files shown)


jaredperreault-okta:
Can you add a video/screenshot and add some tests to prevent regressions?

nikolaystoilov-okta:
@jaredperreault-okta I am not sure what test to add. Can you suggest what will be able validate the change?
There is no new UI element, there are the same but will not overlap. Before my change I was able to reproduce the change every ~5 language changes. After my change I tried 30+ times and was not able to reproduce it.

jaredperreault-okta (approved):

src/v1/views/shared/Header.ts
```diff
@@ -104,74 +74,74 @@ export default class Header extends View {
...
-    case 'swap':
-      return Animations.swapBeacons({
-        $el: container,
-        swap: () => {
-          const isLoading = isLoadingBeacon(this.currentBeacon);
-
-          // Order of these calls is important for -
-          // loader --> security/factor beacon swap.
-          removeBeacon(this);
-          if (isLoading) {
```
mauriciocastillosilva-okta, line 157: is this the fix? That we now remove the classes and we don't need to check if the beacon is loading?

src/v1/views/shared/Header.ts
```diff
@@ -185,14 +155,7 @@ export default class Header extends View {
 
   // Hide the beacon on primary auth failure. On primary auth success, setBeacon does this job.
   removeLoadingBeacon() {
```
mauriciocastillosilva-okta, line 157: this function is also called in by the listener on line 72 (event: `removeLoading`)
With this changes, this function no longer calls `Animations.implode`. Can we test that when that listener triggers, the beacon still looks good?

test/testcafe/spec/v1/MfaOnlySmsChallenge_spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+stupportedLanguages.forEach( (lang) => {
+  test
+    .requestHooks(mock)('Show SMS challenge for MFA only setup "' + lang + '" locale', async t => {
+      const pageObject = await setup(t, {
+        language: lang
+      });
+
+      // wait for all elements to render
+      await t.wait(5000);
```
mauriciocastillosilva-okta, line 74: This test will take a long time given it waits 5 secs and we are testing all languages.
- Do we really need to test all languages? I don't see any difference in the code in this PR for one lang vs another
- Can we get away with a unit test? This is v1, and we have a few specs for mfa required and challenge

src/v1/views/shared/Header.ts
nikolaystoilov-okta: I think that it will. 
`this.setBeacon(null, null)` will evaluate transition to `swap` or `unload`. 
in case of `unload` we will execute `Animations.implode(container)` on line # 135
in case of `swap` we will execute `Animations.swapBeacons` which internally will execute `implode` and `explode`

test/testcafe/spec/v1/MfaOnlySmsChallenge_spec.js
nikolaystoilov-okta: This is demo that changing the language do not overlap the elements ... It is not related to specific language but for language transition. I can reduce the number leave just 2 languages  ...

src/v1/views/shared/Header.ts
nikolaystoilov-okta: Yes, we don't evaluate the state 2 times (if we are in `swap` case the beacon `isLoading`) and the changes in `removeLoadingBeacon` to move the actual update to happen in `unload` case.

test/testcafe/spec/v1/MfaOnlySmsChallenge_spec.js
mauriciocastillosilva-okta: Got it. If we can, lets run 3-5 languages then

src/v1/views/shared/Header.ts
mauriciocastillosilva-okta: I see. Ok, sounds good. Can we just test it manually please? I think we have tests, but just want to double check

test/testcafe/spec/v1/MfaOnlySmsChallenge_spec.js
mauriciocastillosilva-okta: Also, what I was asking about unit tests, I see we have a bunch of MFA_REQUIRED mocks for unit tests here: https://github.com/okta/okta-signin-widget/tree/master/test/unit/helpers/xhr, any chance we can use those? Not a big deal adding another mock, but if we have one already let's reuse it (Might not work for an e2e test, but it will for a unit test)

nikolaystoilov-okta: This is mocking the response from ADFS challenge. I am not sure any of the existing mock will be similar

PR #2641, opened 2022-07-27
Maps brand colors to override MUI-odyssey theme

shawyu-okta:
## PR Checklist

- [x] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-514615](https://oktainc.atlassian.net/browse/OKTA-514615)

### DIFF
--- a/src/v3/src/components/Widget/index.tsx
+++ b/src/v3/src/components/Widget/index.tsx
@@ -240,9 +240,6 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [idxTransaction, bootstrap]);
 
-  // TODO: OKTA-517723 temporary override until odyssey-react-mui theme borderRadius value is fixed
-  odysseyTheme.shape.borderRadius = 4;
-
   return (
     <WidgetContextProvider value={{
       authClient,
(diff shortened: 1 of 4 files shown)


lesterchoi-okta (commented):
Does this need tests?
Looks trivial now but we can make sure the mapping works if we have to update the `brand` object.

shawyu-okta:
> Does this need tests?
> Looks trivial now but we can make sure the mapping works if we have to update the brand object.

I didn't add tests since the previous mapping didn't have tests (we did test the "derive colors" function) but I agree that having some tests would help in the scenario you pointed out. I'll add some simple assertions for that.

shuowu-okta:
maybe migrate theming test from v3/testcafe to jest, and add additional tests there if it's not too much work

lesterchoi-okta (approved):
Leaving the testing decision up to you. Maybe we can add it when ODS theming stabilizes

shawyu-okta:
Migrating the e2e test for theming is a good idea - I looked and it seems quite doable, but we would still have to keep it skipped since the components within the widget aren't reflecting the theme in MUI yet, so even if I migrated it they would still be skipped. Because of that I want to hold off migrating the test until I can see it pass.

lesterchoi-okta:
Can you create a story for that?

shawyu-okta:
@lesterchoi-okta yup, i have added a comment to the JIRA in the code as well: https://oktainc.atlassian.net/browse/OKTA-519086

PR #2643, opened 2022-07-28
OKTA-518866 Upgrade v3 auth-js

shuowu-okta:
## Description:

* upgrade authjs in v3 package.json

* remove databag preset code in transformer/field

* update all button element schema to include step (required)

* add payload test for select-authenticator-*

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-518866](https://oktainc.atlassian.net/browse/OKTA-518866)


src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
```diff
@@ -22,43 +22,28 @@ import { PASSWORD_REQUIREMENT_VALIDATION_DELAY_MS } from '../../constants';
...
     passwordMatchingKey,
     uischema.elements as FieldElement[],
   ) as FieldElement;
+  if (!passwordElement) {
+    passwordMatchingKey = 'credentials.newPassword';
+    passwordElement = getUIElementWithName(
+      passwordMatchingKey,
+      uischema.elements as FieldElement[],
+    ) as FieldElement;
+  }
```
glenfannin-okta, line 46: What is this doing? I am not sure if credentials.newPassword is real in the Generic Remediator implementation, so this would break functionality I believe.

src/v3/test/integration/identify-with-password.test.tsx
```diff
@@ -50,7 +50,6 @@ describe('identify-with-password', () => {
             credentials: {
               passcode: 'fake-password',
             },
-            rememberMe: false,
```
glenfannin-okta, line 53: Any reason this was removed?

src/v3/src/transformer/field/transform.ts
```diff
@@ -73,17 +73,6 @@ export const transformStepInputs: StepTransformer = (step?: NextStep): FormBag =
...
-      // TODO: logic here can be removed once authjs can handle auto proceed without hints from the downstream
-      const { options: { type } } = uischema as FieldElement;
-      let defaultValue;
-      if (type === 'string') {
-        defaultValue = '';
-      }
-      if (type === 'boolean') {
-        defaultValue = false;
-      }
-      acc.data[name] = defaultValue;
```
glenfannin-okta, line 86: This fixed the data leftover issue from flow to flow, does that still work with this change?

glenfannin-okta (approved):
Only real concern here is with the `src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts` because that can negatively impact the enroll password / reset password flow

src/v3/src/transformer/field/transform.ts
shuowu-okta: all tests passed

src/v3/test/integration/identify-with-password.test.tsx
shuowu-okta: ok to ignore it, backend will handle default value for optional field

src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
shuowu-okta: `credentials.newPassword` is not a thing based on OIE protocol, it will not break anything

lesterchoi-okta (approved):

src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
```diff
@@ -23,43 +23,28 @@ import { getUserInfo, loc } from '../../util';
...
   if (passwordElement) {
```
shawyu-okta, line 48: could this be turned into `if/else`?

lesterchoi-okta (approved):
`codegen` adds ~15-20s before each of the commands because of `&&`. Fine for CI but a bit annoying when you're working locally and know you've already generated the files.

src/v3/src/transformer/terminal/transformTerminalTransaction.ts
```diff
@@ -84,6 +84,7 @@ const appendViewLinks = (
     type: 'Link',
     options: {
       label: loc('goback', 'login'),
+      step: 'cancel',
```
glenfannin-okta, line 87: I dont believe `cancel` is a step, it is an `action` right? So if it is passed in as a step wouldnt it be rejected with an error that this step does not exist?

src/v3/src/transformer/terminal/transformTerminalTransaction.ts
```diff
@@ -116,6 +117,7 @@ const appendViewLinks = (
             actions: [{ name: skipStep?.name, params: rest }],
           });
         }),
+        step: transaction.nextStep!.name,
```
glenfannin-okta, line 120: I do not believe nextStep will be present here, this is a terminal scenario, so more likely than not there is no nextStep in transaction.

PR #2658, opened 2022-08-02
OKTA-488781 Password enroll/expire/reset with confirmation component implementation

glenfannin-okta:
## Description:
Purpose of this PR is to implement the custom Password with confirmation component.

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/97472729/182455720-2caebfc4-e9f3-4631-b14c-4d7ce2b52540.mov

### Issue:

- [OKTA-488781](https://oktainc.atlassian.net/browse/OKTA-488781)

### DIFF
--- a/src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
+++ b/src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
@@ -20,6 +20,7 @@ import { getMessage } from '../../../../v2/ion/i18nTransformer';
 import { useWidgetContext } from '../../contexts';
 import {
   ChangeEvent,
+  DataSchema,
   FormBag,
   PasswordWithConfirmationElement,
   UISchemaElementComponent,
@@ -33,6 +34,7 @@ const PasswordWithConfirmation: UISchemaElementComponent<{
   const {
     options: {
       inputMeta: {
+        name: newPwName,
         // @ts-ignore expose type from auth-js
         messages: newPasswordMessages = {},
       },
@@ -40,10 +42,10 @@ const PasswordWithConfirmation: UISchemaElementComponent<{
   } = newPasswordElement;
 
   const {
-    name: confirmPwName,
     label: confirmPwLabel,
     options: {
       inputMeta: {
+        name: confirmPwName,
         // @ts-ignore expose type from auth-js
         messages = {},
       },
@@ -57,10 +59,10 @@ const PasswordWithConfirmation: UISchemaElementComponent<{
 
   // Overrides default validate function for password field
   const validate = useCallback((data: FormBag['data']) => {
-    const newPw = data[newPasswordElement.name];
+    const newPw = data[newPwName];
     const errorMessages: Partial<IdxMessage & { name?: string }>[] = [];
     if (!newPw) {
-      errorMessages.push({ name: newPasswordElement.name, i18n: { key: 'model.validation.field.blank' } });
+      errorMessages.push({ name: newPwName, i18n: { key: 'model.validation.field.blank' } });
     }
 
     if (!confirmPassword) {
@@ -68,10 +70,9 @@ const PasswordWithConfirmation: UISchemaElementComponent<{
     } else if (confirmPassword !== newPw) {
       errorMessages.push({ name: confirmPwName, i18n: { key: 'password.error.match' } });
     }
-    console.log('errorMessages from validation:', errorMessages);
 
     return errorMessages.length ? errorMessages : undefined;
-  }, [confirmPassword, newPasswordElement.name, confirmPwName]);
+  }, [confirmPassword, newPwName, confirmPwName]);
 
   const handleConfirmPasswordChange = (e: ChangeEvent<HTMLInputElement>) => {
     setConfirmPassword(e.currentTarget.value);
@@ -80,15 +81,15 @@ const PasswordWithConfirmation: UISchemaElementComponent<{
   useEffect(() => {
     // update validate function to prevent submission w/o confirm pw data
     // when server messages are set, we must reset the validate function
-    dataSchemaRef.current![newPasswordElement.name].validate = validate;
-  }, [validate, confirmPasswordError, newPasswordElement.name, dataSchemaRef, newPasswordMessages]);
+    (dataSchemaRef.current![newPwName] as DataSchema).validate = validate;
+  }, [validate, newPwName, dataSchemaRef, confirmPasswordError, newPasswordMessages]);
 
   return (
     <Box>
       <Box marginBottom={4}>
         <InputPassword uischema={newPasswordElement} />
       </Box>
-      <Box marginBottom={4}>
+      <Box>
         <PasswordInput
           label={confirmPwLabel}
           name={confirmPwName}
--- a/src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
+++ b/src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
@@ -53,6 +53,8 @@ const PasswordWithConfirmation: UISchemaElementComponent<{
     },
   } = confirmPasswordElement;
 
+  // Updating dataSchema validate fn requires a slight delay, need to wait for completion to show component
+  const [isReady, setIsReady] = useState<boolean>(false);
   const [confirmPassword, setConfirmPassword] = useState<string>();
   const { dataSchemaRef } = useWidgetContext();
   const confirmPasswordError = messages?.value?.[0] && getMessage(messages.value[0]);
@@ -79,12 +81,14 @@ const PasswordWithConfirmation: UISchemaElementComponent<{
   };
 
   useEffect(() => {
+    setIsReady(false);
     // update validate function to prevent submission w/o confirm pw data
     // when server messages are set, we must reset the validate function
     (dataSchemaRef.current![newPwName] as DataSchema).validate = validate;
+    setIsReady(true);
   }, [validate, newPwName, dataSchemaRef, confirmPasswordError, newPasswordMessages]);
 
-  return (
+  return isReady ? (
     <Box>
       <Box marginBottom={4}>
         <InputPassword uischema={newPasswordElement} />
@@ -104,17 +108,17 @@ const PasswordWithConfirmation: UISchemaElementComponent<{
           }}
         />
         {!!confirmPasswordError && (
-          <FormHelperText
-            ariaDescribedBy={confirmPwName}
-            data-se={`${confirmPwName}-error`}
-            error
-          >
-            {confirmPasswordError}
-          </FormHelperText>
+        <FormHelperText
+          ariaDescribedBy={confirmPwName}
+          data-se={`${confirmPwName}-error`}
+          error
+        >
+          {confirmPasswordError}
+        </FormHelperText>
         )}
       </Box>
     </Box>
-  );
+  ) : null;
 };
 
 export default PasswordWithConfirmation;
--- a/src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
+++ b/src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
@@ -27,15 +27,10 @@ import {
 } from '../../types';
 import { getLabelName } from '../helpers';
 
-enum PasswordType {
-  NEW,
-  CONFIRM,
-}
-
 const PasswordWithConfirmation: UISchemaElementComponent<{
   uischema: PasswordWithConfirmationElement
 }> = ({ uischema }) => {
-  const { input } = uischema.options;
+  const { inputMeta } = uischema.options;
   const {
     name: newPwName,
     label: newPwLabel,
@@ -46,11 +41,11 @@ const PasswordWithConfirmation: UISchemaElementComponent<{
       },
       attributes,
     },
-  } = input;
+  } = inputMeta;
 
   const { dataSchemaRef } = useWidgetContext();
-  const value = useValue(input);
-  const onChangeHandler = useOnChange(input);
+  const value = useValue(inputMeta);
+  const onChangeHandler = useOnChange(inputMeta);
   // Must use this flag to determine which field contains error
   const [hasNewPwError, setHasNewPwError] = useState<boolean>(false);
   const [isNewPwTouched, setIsNewPwTouched] = useState<boolean>(false);
@@ -64,7 +59,7 @@ const PasswordWithConfirmation: UISchemaElementComponent<{
 
   // Overrides default validate function for password field
   const validate = useCallback((data: FormBag['data']) => {
-    const newPw = data[input.name];
+    const newPw = data[inputMeta.name];
     const errorMessages: Partial<IdxMessage>[] = [];
     if (!newPw) {
       setIsNewPwTouched(false);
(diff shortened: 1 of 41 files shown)


src/v3/src/util/resetMessagesToInputs.ts
```diff
@@ -14,7 +14,7 @@ import { IdxMessage, Input } from '@okta/okta-auth-js';
 
 export function resetMessagesToInputs(
   inputs: Input[],
-  messages: Record<string, Partial<IdxMessage>>,
+  messagesByField: Record<string, Partial<IdxMessage>[]>,
```
shuowu-okta, line 17: renamed due to naming collision?

src/v3/test/integration/authenticator-expired-password.test.tsx
```diff
@@ -68,4 +61,99 @@ describe('authenticator-expired-password', () => {
...
+    await findByText(/Your password has expired/);
+    await findByText(/Password requirements/);
+
+    const submitButton = await findByTestId('#/properties/submit');
+    const newPasswordEle = await findByTestId('credentials.passcode') as HTMLInputElement;
+
+    await user.type(newPasswordEle, 'superSecretP@ssword12');
+
+    await user.click(submitButton);
+    await findByText(/This field cannot be left blank/);
```
shuowu-okta, line 79: be specific this field message comes from confirmPassword field

src/v3/test/integration/authenticator-expired-password.test.tsx
```diff
@@ -68,4 +61,99 @@ describe('authenticator-expired-password', () => {
...
+
+    await findByText(/Your password has expired/);
+    await findByText(/Password requirements/);
+
+    const submitButton = await findByTestId('#/properties/submit');
+    const confirmPasswordEle = await findByTestId('credentials.confirmPassword') as HTMLInputElement;
+
+    await user.type(confirmPasswordEle, 'abc123');
+    await user.click(submitButton);
+    await findByText(/This field cannot be left blank/);
```
shuowu-okta, line 101: be specific this field message comes from new password field

src/v3/test/integration/authenticator-expired-password.test.tsx
```diff
@@ -68,4 +61,99 @@ describe('authenticator-expired-password', () => {
...
+    const submitButton = await findByTestId('#/properties/submit');
+    const newPasswordEle = await findByTestId('credentials.passcode') as HTMLInputElement;
+    const confirmPasswordEle = await findByTestId('credentials.confirmPassword') as HTMLInputElement;
+
+    const password = 'superSecretP@ssword12';
+    await user.type(newPasswordEle, password);
+
+    await user.type(confirmPasswordEle, 'abc123');
+    await user.click(submitButton);
+    await findByText(/New passwords must match/);
```
shuowu-okta, line 127: be specific this field message comes from confirmPassword field

src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
```diff
@@ -54,6 +54,18 @@ export const transformEnrollPasswordAuthenticator: IdxStepTransformer = ({
...
+  uischema.elements = removeUIElementWithName(
+    passwordFieldName,
+    uischema.elements,
+  );
+
+  const passwordWithConfirmationElement: PasswordWithConfirmationElement = {
+    type: 'PasswordWithConfirmation',
+    label: loc('oie.password.confirmPasswordLabel', 'login'),
+    options: {
+      input: passwordElement,
```
shuowu-okta, line 66: how about following the `inputMeta` pattern for FieldElement?

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
```diff
@@ -0,0 +1,170 @@
...
+      errorMessages.push({ i18n: { key: 'model.validation.field.blank' } });
+    } else if (confirmPassword !== newPw) {
+      setIsTouched(false);
+      errorMessages.push({ i18n: { key: 'password.error.match' } });
+    }
+
+    return errorMessages.length ? errorMessages : undefined;
+  }, [confirmPassword, input.name, setHasNewPwError, setIsTouched, setIsNewPwTouched]);
+
+  const handlePasswordChange = (e: ChangeEvent<HTMLInputElement>, type: PasswordType) => {
```
shuowu-okta, line 88: split this handler function into two separate ones and attach each to the corresponding field

src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
```diff
@@ -54,6 +54,18 @@ export const transformEnrollPasswordAuthenticator: IdxStepTransformer = ({
...
     };
   }
+  uischema.elements = removeUIElementWithName(
+    passwordFieldName,
+    uischema.elements,
+  );
+
+  const passwordWithConfirmationElement: PasswordWithConfirmationElement = {
+    type: 'PasswordWithConfirmation',
+    label: loc('oie.password.confirmPasswordLabel', 'login'),
```
shawyu-okta, line 64: not a required change: but wondering if we ever have an instance where we need the flexibility to pass the label for this element type. if not, we can consider encapsulating this label

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
```diff
@@ -0,0 +1,170 @@
...
+      attributes,
+    },
+  } = input;
+
+  const { dataSchemaRef } = useWidgetContext();
+  const value = useValue(input);
+  const onChangeHandler = useOnChange(input);
+  // Must use this flag to determine which field contains error
+  const [hasNewPwError, setHasNewPwError] = useState<boolean>(false);
+  const [isNewPwTouched, setIsNewPwTouched] = useState<boolean>(false);
```
shuowu-okta, line 56: can we focus this PR on form submission validation only, and handle this state in the following PR?

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
```diff
@@ -0,0 +1,170 @@
...
+    // when server messages are set, we must reset the validate function
+    dataSchemaRef.current![input.name].validate = validate;
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [validate, newPasswordError, confirmPasswordError, input.name]);
+
+  return (
+    <Box>
+      <Box marginBottom={4}>
+        <PasswordInput
+          label={getLabelName(newPwLabel!)}
```
shawyu-okta, line 117: related to my other comment about the label being encapsulated -- if it were encapsulated it would mean we could avoid this type assertion as well

shawyu-okta (approved):
minor comments

shawyu-okta:
I noticed this case in the video, is this expected? Seems weird that it reports the first password field as blank when it isn't (I'd expect it to say passwords don't match or only say the field is blank for the second input)
image

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
```diff
@@ -0,0 +1,170 @@
...
+    // update validate function to prevent submission w/o confirm pw data
+    // when server messages are set, we must reset the validate function
+    dataSchemaRef.current![input.name].validate = validate;
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [validate, newPasswordError, confirmPasswordError, input.name]);
+
+  return (
+    <Box>
+      <Box marginBottom={4}>
+        <PasswordInput
```
shuowu-okta, line 116: You can have two uischema states, one for password, and one for confirmPassword, and pass each uischema to its InputPassword component. 

Also make InputPassword can accept `value` props to override `const value = useValue(uischema)`, so you can pass state down to that component.

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
```diff
@@ -0,0 +1,170 @@
...
+    if (type === PasswordType.NEW) {
+      setIsNewPwTouched(true);
+      onChangeHandler(e.currentTarget.value);
+    } else {
+      setIsTouched(true);
+      setConfirmPassword(e.currentTarget.value);
+    }
+  };
+
+  const handlePasswordBlur = (type: PasswordType) => {
```
shuowu-okta, line 98: can be two functions, and let's handle it in the following PR, so the feature comes in with tests

shuowu-okta (commented):
Thanks for separating this change out!

glenfannin-okta:
> I noticed this case in the video, is this expected? Seems weird that it reports the first password field as blank when it isn't (I'd expect it to say passwords don't match or only say the field is blank for the second input) image

This is expected in the current state of the widget, I reported this issue here: https://oktainc.atlassian.net/browse/OKTA-518428 it is caused by this because when submitting, it clears the data bag, but doesn't reflect that change in our inputs.

src/v3/src/util/resetMessagesToInputs.ts
glenfannin-okta: Yes and no, this prop looks like this 
```
{ 
   "identifier": [{ i18n: { key: 'some.key'}}],
   "credentials.passcode": [{ i18n: { key: 'some.key'}}]
}
```
The name of the prop being "messages" isnt exactly reflective of what it stores. This is one reason why I renamed it, the second reason is because we actually grab the messages from the record below, and I wanted to use that variable name.

src/v3/test/integration/authenticator-expired-password.test.tsx
glenfannin-okta: in this case, there can only be one error message, if there are multiple with this text jest will throw an error.

shuowu-okta: but it's not clear which field this error message belongs to, right?

src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
glenfannin-okta: Do you mean pass the `label` as a prop to `options` instead of here? 

I can do that, I just re-used the `label` attr that was available for the confirm password field to prevent duplicating props in the schema.

shawyu-okta: I mean if `label` can live hardcoded inside the `PasswordWithConfirmation` component

glenfannin-okta: Ohhh, the direction that was told to me was that the components should not use `loc` or any translation. That the components should always receive translated text.

src/v3/test/integration/authenticator-expired-password.test.tsx
shawyu-okta: agreed @shuowu-okta I think it would be good to scope this `find*` selector within the specific input we care about rather than the global `screen` selectors

shawyu-okta:
> This is expected in the current state of the widget, I reported this issue here: https://oktainc.atlassian.net/browse/OKTA-518428 it is caused by this because when submitting, it clears the data bag, but doesn't reflect that change in our inputs.

thanks for the confirming!

src/v3/test/integration/authenticator-expired-password.test.tsx
glenfannin-okta: The test description does, but I will update. :)

shuowu-okta: It's not the same, think about if this error message appears in new password field, test will still pass, but that will be a bug in widget

glenfannin-okta: Yep, forgot this is an integration test and should guard against other issues like this. Updated.

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
glenfannin-okta: The other Pr is already open (and I am waiting for this to merge to accept and resolve conflicts), I think you are looking to have a HOC somewhat contain this type of state which I can continue looking into after this gets merged and resolve them together. 

But this is to at least match parity for now and drive this feature to completion and refactor (if needed) in the other PR that is open.

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
glenfannin-okta: Other than the label and name, what benefit do you get out of creating a separate schema element that is not from the transaction, that will only be used in this component?

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
shuowu-okta: this will end up with not tested code get merged in, let's be focus in this PR for submission validation only

src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
glenfannin-okta: I made this update, but, wouldnt it be confusing to someone that looks at `inputMeta` in the `FieldElement` and see its signature, then come here and expect that same signature but to realize it is actually a `FieldElement`?

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
glenfannin-okta: I can add a regression test to prevent this.

Also, will stage and commit my unit test.

src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
shuowu-okta: this one might be a special case,  maybe we can add `password` and `confirmPassword` to contain uischemas for both fields, or you can derive internal uischema states. Not sure which one works better

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
shuowu-okta: then what's the purpose of splitting the changes, it's getting together into another mega PR

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
shuowu-okta: DRY

shuowu-okta: if the component is hard to be reused, which most probably means that component is not well organized

glenfannin-okta: Are you saying to do something like this? 

```
export interface PasswordWithConfirmationElement extends UISchemaElement {
  type: 'PasswordWithConfirmation';
  options: {
    password: FieldElement,
    confirmPassword: FieldElement,
  };
}
```

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
glenfannin-okta: I will remove it and handle it in the next PR, but i will create a test and skip it to make sure it gets covered.

src/v3/test/integration/authenticator-expired-password.test.tsx
```diff
@@ -68,4 +61,108 @@ describe('authenticator-expired-password', () => {
...
+    expect(confirmPasswordError.innerHTML).toBe('New passwords must match');
+    expect(authClient.options.httpRequestClient).not.toHaveBeenCalledWith(
+      'POST',
+      'https://oie-4695462.oktapreview.com/idp/idx/challenge/answer',
+      expect.anything(),
+    );
+  });
+
+  // TODO: validate function is not being overwritten here, but works in real org, why?
+  it.skip('should not make network request without completing any password fields', async () => {
```
glenfannin-okta, line 143: An odd issue here is happening (only in test) after rebasing with 0.0, the validate function is not being overwritten. When I manually test, it works just fine and I can see both errors, but in this test it fails because the function isnt overwritten. I debugged through it and I can see the validate function being re-set but still doesnt get executed. Still debugging, but wanted to push up the other changes for re-review in the meantime.

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
```diff
@@ -0,0 +1,87 @@
...
+    }));
+  };
+
+  return (
+    <Box>
+      <Box marginBottom={4}>
+        <InputPassword uischema={newPasswordElement} />
+      </Box>
+      <Box>
+        <PasswordInput
```
shuowu, line 60: nit: you can also InputPasswprd to accept value and onChange props to override the default behavior, then it can be reused here

src/v3/src/components/Widget/index.tsx
```diff
@@ -76,6 +76,7 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
   } = widgetProps;
 
   const [data, setData] = useState<FormBag['data']>({});
+  const [additionalData, setAdditionalData] = useState<FormBag['data']>({});
```
shuowu, line 79: you can add comment here, like data not for form submission

shuowu (approved):

PR #2660, opened 2022-08-02
refactor: centralize test attributes transformation

shuowu-okta:
## Description:

* centralize test attributes transformation

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-516853](https://oktainc.atlassian.net/browse/OKTA-516853)

### DIFF
--- a/src/v3/src/components/Form/renderers.tsx
+++ b/src/v3/src/components/Form/renderers.tsx
@@ -96,7 +96,14 @@ export default [
     renderer: WebAuthNSubmitButton,
   },
   {
-    tester: ({ name }: FieldElement) => name?.endsWith('phoneNumber'),
+    tester: (uischema: FieldElement) => {
+      const {
+        options: {
+          inputMeta: { name } = {},
+        } = {},
+      } = uischema;
+      return name?.endsWith('phoneNumber');
+    },
     renderer: PhoneAuthenticator,
   },
   {
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -19,9 +19,9 @@ import { getImmutableData, toNestedObject } from '../util';
 
 type OnSubmitHandlerOptions = {
   includeData?: boolean;
-  actionFn?: NextStep['action'];
   params?: Record<string, unknown>;
   step: string;
+  isActionStep?: boolean;
   stepToRender?: string;
 };
 
@@ -40,12 +40,16 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
 
   return useCallback(async (options: OnSubmitHandlerOptions) => {
     const {
-      actionFn, params, includeData, step, stepToRender,
+      params, 
+      includeData, 
+      step, 
+      isActionStep,
+      stepToRender,
     } = options;
 
     const immutableData = getImmutableData(currTransaction!, step);
 
-    const fn = actionFn || authClient.idx.proceed;
+    const fn = authClient.idx.proceed;
 
     let payload: IdxActionParams = {};
     if (includeData) {
@@ -54,7 +58,14 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     if (params) {
       payload = merge(payload, params);
     }
-    payload = { ...payload, step };
+    if (isActionStep) {
+      payload = {
+        ...payload,
+        actions: [{ name: step, params }],
+      };
+    } else {
+      payload = { ...payload, step };
+    }
     payload = merge(payload, immutableData);
     payload = toNestedObject(payload);
     if (currTransaction!.context.stateHandle) {
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { IdxActionParams, NextStep } from '@okta/okta-auth-js';
+import { IdxActionParams } from '@okta/okta-auth-js';
 import { merge, omit } from 'lodash';
 import { useCallback } from 'preact/hooks';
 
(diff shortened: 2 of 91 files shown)


src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -54,7 +58,14 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     if (params) {
       payload = merge(payload, params);
     }
-    payload = { ...payload, step };
+    if (isActionStep) {
+      payload = {
+        ...payload,
+        actions: [{ name: step, params }],
+      };
```
glenfannin-okta, line 65: This will break all functionality on rain/monolith

shuowu-okta: have you verified it locally, in the end of this function, we have a step to add stateToken to params

glenfannin-okta: Oh I see it, was hidden. Ty!

src/v3/src/components/Form/renderers.tsx
```diff
@@ -96,7 +96,14 @@ export default [
...
   {
-    tester: ({ name }: FieldElement) => name?.endsWith('phoneNumber'),
+    tester: (uischema: FieldElement) => {
+      const {
+        options: {
+          inputMeta: { name } = {},
+        } = {},
+      } = uischema;
+      return name?.endsWith('phoneNumber');
+    },
```
glenfannin-okta, line 106: What prompted using the name from `inputMeta` vs the top level name prop?

shawyu-okta:
I had thought this might be more generic, so it would not only handle a single `data-se` attribute, but rather allow any arbitrary testing attributes into the components for more flexibility.

shuowu-okta:
@shawyu-okta Firstly, this PR  is a setup, so it can be easily extended when more test attributes come, like when the team work on selector parity changes, this PR won't run against parity tests, so by no means it can handle all scenarios. 
And, I feel right now we only have `data-se` as the test attr, and `data-type=save` for submit.

src/v3/src/transformer/terminal/transformTerminalTransaction.ts
```diff
@@ -85,18 +85,7 @@ const appendViewLinks = (
...
-      // @ts-ignore OKTA-512706 temporary until auth-js applies this fix
-      action: cancelStep?.action && ((params?: IdxActionParams) => {
-        const { stateHandle, ...rest } = params ?? {};
-        return widgetProps.authClient?.idx.proceed({
-          // @ts-ignore stateHandle can be undefined
-          stateHandle,
-          actions: [{ name: 'cancel', params: rest }],
-        });
-      }),
+      href: cancelStep ? undefined : (baseUrl || '/'),
```
shawyu-okta, line 88: just confirming that we don't need to check `cancelStep.action` and only existence of `cancelStep` is enough here

shuowu-okta: Step action fn cannot compatible with okta hosted scenario very well (due to missing stateToken), v3 will not depend on that function any more.
To your question, it won't be needed, as action is required in cancel step, so only check step will be enough.

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -54,7 +58,14 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     if (params) {
       payload = merge(payload, params);
     }
-    payload = { ...payload, step };
+    if (isActionStep) {
+      payload = {
+        ...payload,
```
glenfannin-okta, line 63: It looks like `payload` includes `data` shouldn't that be included in the `params` part of the action? IIRC the only thing expected at the base of the object is stateHandle when there are actions, I thought everything else went into the action object (name and params for that action).

src/v3/src/components/Form/renderers.tsx
shuowu-okta: remove duplicated values

src/v3/src/hooks/useOnSubmit.ts
shuowu-okta: upstream transformation logic controls that

glenfannin-okta: Sorry, i am referring to this file

```
if (includeData) {
      payload = merge(payload, omit(data, OMITTED_PROPERTIES));
    }
    if (params) {
      payload = merge(payload, params);
    }
```
Data is added to payload, then params are merged to payload but for actions, only params are passed. but payload contains field level data. Was this intentional? IIRC, any corresponding data should also be passed to params for the action. Am I mistaken?

shuowu-okta: that what I mean by controlled by upstream, `includeData` is passed down from transformation layer -> ui component -> hook

src/v3/src/transformer/testAttribute/transform.ts
```diff
@@ -0,0 +1,97 @@
...
+      const { type, options } = element as ButtonElement | LinkElement;
+      return (type === 'Button' && (options as ButtonElement['options']).type === ButtonType.BUTTON)
+        || type === 'Link';
+    },
+  });
+
+  return formbag;
+};
+
+export const transformTestAttribute: TransformStepFn = (formbag) => flow(
```
shawyu-okta, line 93: Are there any unit tests that cover this transform?

shuowu-okta: Planning to add tests in later stage, currently transformer tests are disabled in jest config to avoid noise from existing tests.
I can make some config change if you have strong feeling to include tests in this PR.

glenfannin-okta: I actually fixed all the transformer tests in the last push to 0.0, so you could really just re-enable them and they should not fail (outside of the few cases that were updated in this branch).

shuowu-okta: those custom transformer will be changed in later stage, don't want to bother with touching all those files for now

shawyu-okta (approved):
adding tests sooner would be nice but i won't block this PR on them if there is a followup plan to add them.

glenfannin-okta (approved):

PR #2662, opened 2022-08-03
Fs okta 520681 progressive profiling set required true for boolean optional attributes checkbox

fangshi-okta:
## Description:
set required as true, so an optional checkbox will have default value as false

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:
https://okta.box.com/s/lxo24gnhn7775668kdi83h7lazg22r4f

### Reviewers:
@okta/ciamx 

### Issue:

- [OKTA-520681](https://oktainc.atlassian.net/browse/OKTA-520681)


src/v2/ion/ui-schema/ion-boolean-handler.js
```diff
@@ -32,12 +32,13 @@ const getRadioUiSchema = ({ label, required, options }) => ({
   sublabel: required? null : loc('oie.form.field.optional', 'login'),
 });
 
-const getCheckboxUiSchemaWithDefaultValue = ({ label, type, required }) => ({
+const getCheckboxUiSchemaWithDefaultValue = ({ label, type }) => ({
```
lesterchoi-okta, line 35: test? already covered? alternatively you could keep the param and set default value to `true`

fangshi-okta: yeah there is a test covered this new uiSchema. I just did an end to end test to confirm that it works as expected. shown as the screen record in box link. Thanks!

src/v2/ion/ui-schema/ion-boolean-handler.js
```diff
@@ -32,12 +32,13 @@ const getRadioUiSchema = ({ label, required, options }) => ({
...
   placeholder: label,
   label: false,
   modelType: type,
   type: 'checkbox',
-  required: required,
+  // set required true so default value can be passed to optional attributes as well
```
justinabrokwah-okta, line 40: If `required` is set to false, what happens?

justinabrokwah-okta (commented):
If you can add a little bit more context to the PR description, I think it'll help to explain why you're making this change

src/v2/ion/ui-schema/ion-boolean-handler.js
fangshi-okta: null value will be passed to the backend instead of default value. since the value is not required which means it can be empty. so after setting it as true, it will force default value as false if it is null.

justinabrokwah-okta: Got it, so it needs to be required so a T/F value is passed through 👌

justinabrokwah-okta: All good on test coverage too?

fangshi-okta: right. 
```  await t.expect(await enrollProfilePage.getCheckboxValue('userProfile.required')).eql(false);
  await t.expect(await enrollProfilePage.getCheckboxValue('userProfile.optional')).eql(false);

fangshi-okta: right. 
```  await t.expect(await enrollProfilePage.getCheckboxValue('userProfile.required')).eql(false);
  await t.expect(await enrollProfilePage.getCheckboxValue('userProfile.optional')).eql(false);

justinabrokwah-okta (approved):
LGTM!

lesterchoi-okta (approved):
Only bugfixes are allowed in weekly release

1. confirm this is a bugfix and update the Jira to reflect that? 
2. set values in the Bug and RelReview field before merging

vicentevillegas-okta (approved):
LGTM 👍

PR #2664, opened 2022-08-03
OKTA-516589 Added auth list item focus outline style

leemchale-okta:
## Description:

Fixes the issue of no focus indicator for authenticator buttons.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

Screen Shot 2022-08-02 at 3 53 54 PM

### Issue:

- [OKTA-516589](https://oktainc.atlassian.net/browse/OKTA-516589)

### DIFF
--- a/src/v3/src/components/AuthenticatorButton/AuthenticatorButton.theme.ts
+++ b/src/v3/src/components/AuthenticatorButton/AuthenticatorButton.theme.ts
@@ -15,4 +15,8 @@ import type { ThemeReducer } from '@okta/odyssey-react-theme';
 export const theme: ThemeReducer = (t) => ({
   ColorHover: t.ColorPrimaryDark,
   ColorPrimaryBase: t.ColorPrimaryBase,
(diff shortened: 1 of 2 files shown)


shawyu-okta (approved):
This looks good for our beta release - we will ultimately migrate off of this to using MUI theme to accomplish the same task

shawyu-okta:
Actually, @leemchale-okta curious if we should go with more of the same style as the `:focus` state where the border of the unfocused element is just darkened instead of the border around the whole thing
image

leemchale-okta:
> Actually, @leemchale-okta curious if we should go with more of the same style as the `:focus` state where the border of the unfocused element is just darkened instead of the border around the whole thing image

Yeah I thought about that as well but I showed it to @alexdahl-okta he said it looks good so I think its okay but it can easily be changed.

shawyu-okta:
Ah ok great! Thanks for confirming with Alex. :shipit:

PR #2665, opened 2022-08-03
OKTA-486151 update link list style

leemchale-okta:
## Description:

Converts link list buttons to use the Link component and its styling.

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Screenshot/Video:

Screen Shot 2022-08-03 at 2 12 54 PM

### Issue:

- [OKTA-486151](https://oktainc.atlassian.net/browse/OKTA-486151)

### DIFF
--- a/src/v3/src/components/Link/Link.tsx
+++ b/src/v3/src/components/Link/Link.tsx
@@ -24,6 +24,7 @@ const Link: UISchemaElementComponent<{
     options: {
       label,
       href,
+      dataSe,
       actionParams,
       isActionStep,
       step,
@@ -49,6 +50,8 @@ const Link: UISchemaElementComponent<{
             // eslint-disable-next-line no-script-url
             href="javascript:void(0)"
             onClick={onClick}
+            // eslint-disable-next-line react/jsx-props-no-spreading
+            {...(dataSe && { 'data-se': dataSe } )}
           >
             {getLabelName(label)}
           </LinkMui>
--- a/src/v3/src/transformer/button/getButtonControls.ts
+++ b/src/v3/src/transformer/button/getButtonControls.ts
@@ -19,6 +19,7 @@ import { IDX_STEP } from '../../constants';
 import {
   ButtonElement,
   ButtonType,
+  LinkElement,
   UISchemaElement,
 } from '../../types';
 import { loc } from '../../util';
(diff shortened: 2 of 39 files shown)


src/v3/src/transformer/button/getButtonControls.ts
```diff
@@ -70,13 +71,10 @@ export const getButtonControls = (
     ?? getButtonStep('currentAuthenticatorEnrollment-recover');
   if (config.stepWithForgotPassword && forgotPasswordStep) {
     const { name } = forgotPasswordStep;
-    const forgotPassword: ButtonElement = {
-      type: 'Button',
-      label: loc('forgotpassword', 'login'),
+    const forgotPassword: LinkElement = {
+      type: 'Link',
```
shawyu-okta, line 75: Wonder if we should rename this transformer or move these to another new transformer

shuowu-okta: on the list, will be in next few PRs

src/v3/src/components/Link/Link.tsx
```diff
@@ -36,23 +40,31 @@ const Link: UISchemaElementComponent<{
...
       step,
+      stepToRender,
     });
   };
 
+  const hasAction = actionFn !== undefined || step !== undefined;
+
   return (
-    <Box marginTop={4}>
+    <Box marginTop={2}>
```
shawyu-okta, line 53: How does this change affect other views that currently use this component?

shuowu-okta: hmm, interesting, not sure when this one is touched, probably will change it back

src/v3/src/types/schema.ts
```diff
@@ -248,10 +248,9 @@ export interface PasswordRequirementsElement extends UISchemaElement {
 
 export interface LinkElement extends UISchemaElement {
   type: 'Link';
-  options: ActionOptions & {
+  options: Omit<ButtonElement['options'], 'type'> & {
```
shawyu-okta, line 251: Perhaps we should just define our own type at this point rather than copy from `ButtonElement`?

src/v3/src/components/Link/Link.tsx
leemchale-okta: Actually this is a mistake I meant to change it back to 4

src/v3/src/components/Link/Link.tsx
```diff
@@ -36,23 +40,31 @@ const Link: UISchemaElementComponent<{
...
-          </LinkMui>
-        ) : <LinkMui href={href}>{getLabelName(label)}</LinkMui>
+        <LinkMui
+          data-se={uischema.scope}
+          // eslint-disable-next-line no-script-url
+          href={hasAction ? 'javascript:void(0)' : href}
+          // eslint-disable-next-line react/jsx-props-no-spreading
+          {...(hasAction && { onClick } )}
+          // eslint-disable-next-line react/jsx-props-no-spreading
+          {...(dataType && { 'data-type': dataType } )}
```
shawyu-okta, line 62: Is this necessary for the `Link` type? I know it was used on `Button` - but maybe still needed for tests.

shawyu-okta (approved):
just one last question but otherwise lgtm

src/v3/src/components/Link/Link.tsx
leemchale-okta: Yeah i was trying to keep it in parity to be safe for testing, but I think I can trim some unnecessary ones.

shawyu-okta (approved):

PR #2673, opened 2022-08-04
OKTA-509025 : Initial implementation for Select authenticator enroll Spec

glenfannin-okta:
## Description:
Implementation for SelectAuthenticatorForEnroll_spec for v3 parity.

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

### Issue:

- [OKTA-509025](https://oktainc.atlassian.net/browse/OKTA-509025)

### DIFF
--- a/src/v3/src/components/InformationalText/InformationalText.tsx
+++ b/src/v3/src/components/InformationalText/InformationalText.tsx
@@ -10,8 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { Box } from '@mui/material';
-import { Text } from '@okta/odyssey-react';
+import { Box, Typography } from '@mui/material';
 import { h } from 'preact';
 
 import { DescriptionElement, UISchemaElementComponent } from '../../types';
@@ -28,7 +27,12 @@ const InformationalText: UISchemaElementComponent<{
       display="flex"
       justifyContent="flex-start"
     >
-      <Text as="p">{content}</Text>
+      <Typography
+        paragraph
+        data-se="o-form-explain"
+      >
+        {content}
+      </Typography>
     </Box>
   );
 };
--- a/src/v3/src/transformer/messages/transform.ts
+++ b/src/v3/src/transformer/messages/transform.ts
@@ -96,6 +96,7 @@ export const transformMessages: TransformStepFnWithOptions = ({ transaction }) =
       contentType: 'string',
       class: message.class ?? 'INFO',
       message: message.message,
+      dataSe: `infobox-${(message.class ?? 'INFO').toLowerCase()}`,
     },
   } as InfoboxElement));
 
--- a/src/v3/src/transformer/messages/transform.ts
+++ b/src/v3/src/transformer/messages/transform.ts
@@ -100,7 +100,7 @@ export const transformMessages: TransformStepFnWithOptions = ({ transaction }) =
         message: message.message,
         dataSe: `infobox-${messageClass.toLowerCase()}`,
       },
-    } as InfoboxElement)
+    } as InfoboxElement);
   });
 
   return formBag;
--- a/src/v3/src/transformer/messages/transform.ts
+++ b/src/v3/src/transformer/messages/transform.ts
@@ -90,15 +90,18 @@ export const transformMessages: TransformStepFnWithOptions = ({ transaction }) =
     return formBag;
   }
 
-  messages.forEach((message) => uischema.elements.unshift({
-    type: 'InfoBox',
-    options: {
-      contentType: 'string',
-      class: message.class ?? 'INFO',
-      message: message.message,
-      dataSe: `infobox-${(message.class ?? 'INFO').toLowerCase()}`,
-    },
-  } as InfoboxElement));
+  messages.forEach((message) => {
+    const messageClass = message.class ?? 'INFO';
+    uischema.elements.unshift({
+      type: 'InfoBox',
+      options: {
+        contentType: 'string',
+        class: messageClass,
+        message: message.message,
+        dataSe: `infobox-${messageClass.toLowerCase()}`,
+      },
+    } as InfoboxElement)
+  });
 
   return formBag;
 };
--- a/src/v3/src/transformer/selectAuthenticator/utils.ts
+++ b/src/v3/src/transformer/selectAuthenticator/utils.ts
@@ -190,7 +190,7 @@ const formatAuthenticatorOptions = (
     const id = getOptionValue(option.value as Input[], 'id')?.value;
     const methodType = getOptionValue(option.value as Input[], 'methodType')?.value;
     const enrollmentId = getOptionValue(option.value as Input[], 'enrollmentId')?.value;
-    const DATASE_WITH_METHOD_TYPE = [
+    const AUTHENTICATORS_WITH_METHOD_TYPE = [
       AUTHENTICATOR_KEY.ON_PREM,
       AUTHENTICATOR_KEY.RSA,
     ];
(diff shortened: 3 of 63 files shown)


test/testcafe/framework/page-objects/components/BaseFormObjectV3.js
```diff
@@ -0,0 +1,16 @@
+import BaseFormObject from './BaseFormObject';
+import { Selector } from 'testcafe';
+
+const FORM_SELECTOR = '[data-se="form"]';
+const FORM_INFOBOX_ERROR_SELECTOR = '[data-se="message"] [data-se="infobox-error"]';
+
+export default class BaseFormObjectV3 extends BaseFormObject {
+  constructor(t, index = 0) {
+    super(t);
+    this.el = Selector(FORM_SELECTOR).nth(index);
```
shuowu-okta, line 10: is this `nth` fn considered an anti-pattern?

glenfannin-okta: I pulled this from @lesterchoi-okta PR https://github.com/okta/okta-signin-widget/pull/2648/files#diff-9b67ece18bbf878941e07ad5952c5de98abcf533d0e4bdde0ca8f58ba57d13c3R11

test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
```diff
@@ -227,19 +242,40 @@ test.requestHooks(requestLogger, mockEnrollAuthenticatorPassword)('select passwo
...
-  const req2 = requestLogger.requests[1].request;
-  await t.expect(req2.url).eql('http://localhost:3000/idp/idx/credential/enroll');
-  await t.expect(req2.method).eql('post');
-  await t.expect(req2.body).eql('{"authenticator":{"id":"autwa6eD9o02iBbtv0g3"},"stateHandle":"02CqFbzJ_zMGCqXut-1CNXfafiTkh9wGlbFqi9Xupt"}');
-
-  const req3 = requestLogger.requests[2].request;
-  await t.expect(req3.url).eql('http://localhost:3000/idp/idx/credential/enroll');
-  await t.expect(req3.method).eql('post');
-  await t.expect(req3.body).eql('{"authenticator":{"id":"autwa6eD9o02iBbtv0g3"},"stateHandle":"02CqFbzJ_zMGCqXut-1CNXfafiTkh9wGlbFqi9Xupt"}');
+  if (process.env.OKTA_SIW_V3) {
```
shuowu-okta, line 245: do we really need to change the test spec? this does not seems like a parity test

glenfannin-okta: This is what I discussed in the thread yesterday. v3 makes an extra request that v2 does not make when switching authenticators.

test/testcafe/framework/page-objects/components/BaseFormObjectV3.js
shawyu-okta: even if this isn't the ideal pattern, let's go with the quickest selector pattern if we spend any more time on spec conversion. we will need to return after beta to convert more specs and spend the time then to be more thoughtful

src/v3/src/components/InformationalText/InformationalText.tsx
```diff
@@ -10,8 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { Box } from '@mui/material';
-import { Text } from '@okta/odyssey-react';
```
shawyu-okta, line 14: nice, thank for migrating

src/v3/src/transformer/messages/transform.ts
```diff
@@ -96,6 +96,7 @@ export const transformMessages: TransformStepFnWithOptions = ({ transaction }) =
       contentType: 'string',
       class: message.class ?? 'INFO',
       message: message.message,
+      dataSe: `infobox-${(message.class ?? 'INFO').toLowerCase()}`,
```
shawyu-okta, line 99: consider extracting the value/logic of `message.class ?? 'INFO'` to a variable since the value of properties `dataSe` and `class` should be tightly coupled

src/v3/src/transformer/terminal/transformTerminalMessages.ts
```diff
@@ -47,6 +47,7 @@ const appendMessageElements = (uischema: UISchemaLayout, messages: IdxMessage[])
           message: message.message,
           class: message.class ?? 'WARNING',
           contentType: 'string',
+          dataSe: `infobox-${(message.class ?? 'WARNING').toLowerCase()}`,
```
shawyu-okta, line 50: same comment here about `message.class`

test/testcafe/framework/page-objects/FactorEnrollPasswordPageObjectV3.js
```diff
@@ -0,0 +1,24 @@
...
+const SWITCH_AUTHENTICATOR_LINK = '[data-se="switchAuthenticator"]';
+
+/**
+ * This page object will be used by 
+ * password enrollment
+ * password expiry
+ * password will expire soon 
+ * admin initiated password reset scenarios.
+ * 
+ * TODO: Rename this to have AuthenticatorPasswordPageObject when Factor is cleaned up.
```
shawyu-okta, line 14: can you expand this comment further so we know what "Factor is cleaned up" means? Or file a JIRA in tech debt and link it to this

src/v3/src/transformer/selectAuthenticator/utils.ts
```diff
@@ -159,6 +163,24 @@ const getAuthenticatorDescription = (
   return undefined;
 };
 
+const getUsageDescription = (option: IdxOption): string | undefined => {
+  // @ts-ignore IdxAuthenticator missing allowedFor property
+  const { allowedFor } = option.relatesTo;
+  if (!allowedFor) {
+    return undefined;
+  }
```
shawyu-okta, line 171: i don't think this guard is necessary since the `switch` case will fall through to `default` which is the same

test/testcafe/framework/page-objects/components/BaseFormObjectV3.js
shuowu: is this line needed in this PR, not see any page object initialized with index

test/testcafe/framework/page-objects/FactorEnrollPasswordPageObjectV3.js
glenfannin-okta: Actually this was copied from Lester's PR 😅  I will remove this.

shawyu-okta (approved):

test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
```diff
@@ -236,10 +253,12 @@ test.requestHooks(requestLogger, mockEnrollAuthenticatorPassword)('select passwo
   await t.expect(req2.method).eql('post');
   await t.expect(req2.body).eql('{"authenticator":{"id":"autwa6eD9o02iBbtv0g3"},"stateHandle":"02CqFbzJ_zMGCqXut-1CNXfafiTkh9wGlbFqi9Xupt"}');
 
-  const req3 = requestLogger.requests[2].request;
-  await t.expect(req3.url).eql('http://localhost:3000/idp/idx/credential/enroll');
-  await t.expect(req3.method).eql('post');
-  await t.expect(req3.body).eql('{"authenticator":{"id":"autwa6eD9o02iBbtv0g3"},"stateHandle":"02CqFbzJ_zMGCqXut-1CNXfafiTkh9wGlbFqi9Xupt"}');
+  if (!process.env.OKTA_SIW_V3) {
```
shuowu-okta, line 256: why v3 not making this request? or just the request order is changed?

shuowu-okta: please also add a comment to explain why this spec should not happen in v3

glenfannin-okta: I will add a comment, also adding to the parity test wiki. 

But the reason is because v2 only makes 3 requests none of which are for the "back to authenticator list" click (v2 does not make a request for this click). v3 makes 4 requests, where the "back to authenticator list" is included, req3 for v3 is the back to authenticator list which only includes the stateHandle in the payload so this would fail on v3.

shuowu-okta: ic, switch authenticator is a request in the middle, so it's about the ordering

PR #2677, opened 2022-08-08
feat: client validation for sq enrollment

shuowu-okta:
## Description:

* adds fieldsToValidate array in dataSchema to support stepper fields validation
  * by default it contains all fields in non-stepper form
  * this field can be configurable in when step changes (stepper form)
* add 'callback' option for StepperRadio component to update schema while step changing 

Screen Shot 2022-08-08 at 2 20 24 PM

## PR Checklist

- [ ] Added unit tests?

- [ ] Added e2e tests

### Issue:

- [OKTA-516694](https://oktainc.atlassian.net/browse/OKTA-516694)

### DIFF
--- a/src/v3/src/components/Form/Form.tsx
+++ b/src/v3/src/components/Form/Form.tsx
@@ -18,7 +18,7 @@ import { useCallback } from 'preact/hooks';
 import { useWidgetContext } from '../../contexts';
 import { useOnSubmit } from '../../hooks';
 import {
-  ActionOptions, DataSchema, SubmitEvent, UISchemaLayout,
+  DataSchema, SubmitEvent, UISchemaLayout,
 } from '../../types';
 import { loc, resetMessagesToInputs } from '../../util';
 import Layout from './Layout';
@@ -41,7 +41,14 @@ const Form: FunctionComponent<{
     e.preventDefault();
     setMessage(undefined);
 
-    const { actionParams: params, step } = dataSchemaRef.current!.submit as ActionOptions;
+    const {
+      submit: {
+        actionParams: params,
+        step,
+        includeImmutableData,
+      },
+      fieldsToValidate,
+    } = dataSchemaRef.current!;
 
     // client side validation - only validate for fields in nextStep
     const { nextStep } = currTransaction!;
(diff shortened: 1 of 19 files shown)


src/v3/src/types/.eslintrc.js
```diff
@@ -0,0 +1,17 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+module.exports = {
+  rules: {
+    'import/no-cycle': 'off',
```
shawyu-okta, line 15: How difficult is it to resolve this? We will want to eventually untangle the types to avoid dependency cycles

src/v3/src/components/Form/Form.tsx
```diff
@@ -50,7 +57,7 @@ const Form: FunctionComponent<{
         .reduce((acc: Record<string, Partial<IdxMessage & { name?: string }>[]>, curr) => {
           const name = curr[0];
           const elementSchema = curr[1] as DataSchema;
-          if (typeof elementSchema.validate === 'function') {
+          if (fieldsToValidate.includes(name) && typeof elementSchema.validate === 'function') {
```
shawyu-okta, line 60: This seems like we are narrowing the number of times where we call the validation function, is that correct?

src/v3/src/types/.eslintrc.js
shuowu-okta: I don't really feel it matters for types, that why I added the extension file under types dir.

according to how to avoid this lint warning, we will need to put types into one giant file, which makes me feel type category weigh more than lint rule for this scenario

src/v3/src/components/Form/Form.tsx
shuowu-okta: it's introduced to support form with stepper schema, so when step switches, it updates the fields that need to be validate in that specific step. See security question enrollment schema for example.

src/v3/src/types/.eslintrc.js
shawyu-okta: That seems reasonable - I even found this issue `https://github.com/import-js/eslint-plugin-import/issues/1453` that makes a similar argument. I think we had trouble with using the `import type` syntax but at some point we can try migrating to it when we have time. Seems like that syntax solves the error without needing the lint ignore.

shawyu-okta (approved):

PR #2682, opened 2022-08-08
OKTA-519488 : Implement customized key functionality

glenfannin-okta:
## Description:
The purpose of this PR is to implement the customized i18n key implementation for identifier and password fields including error messages.

I did not add an e2e in v3 because the loc function isn't using the overwriting value in jest, however, this is covered in v2 IdentifyWithPassword_spec. You can see the functionality in the video below.

## PR Checklist

- [x] Added unit tests?

- [x] Added e2e tests

### Screenshot/Video:

https://user-images.githubusercontent.com/97472729/183489876-73b39308-31f8-49f9-ab82-3a29588d2365.mov

### Issue:

- [OKTA-519488](https://oktainc.atlassian.net/browse/OKTA-519488)


src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
```diff
@@ -39,22 +37,23 @@ const PasswordRequirements: UISchemaElementComponent<{
...
-    fieldKey,
     header,
     id,
     requirements,
     settings,
     userInfo,
     validationDelayMs,
   } = uischema.options as PasswordRequirementsElement['options'];
-  const password = get(data, fieldKey);
+  const password = 'credentials.newPassword' in data
```
shuowu-okta, line 47: why depend on data bag in transformer layer, those preset data logics have been removed. 

also seems like we have testing gap for this scenario

src/v3/src/transformer/identify/transformIdentify.ts
```diff
@@ -11,19 +11,81 @@
...
   IdxStepTransformer,
   TitleElement,
   UISchemaElement,
+  WidgetProps,
 } from '../../types';
 import { getUsernameCookie, loc } from '../../util';
+import { isCustomizedI18nKey } from '../i18n';
 import { getUIElementWithName, removeUIElementWithName } from '../utils';
 
+const addCustomizedErrorKeys = (
```
shuowu-okta, line 28: can all i18n related logic be moved to i18n transformer?

src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
glenfannin-okta: > why depend on data bag in transformer layer, those preset data logics have been removed.
do you mean in places where we are setting a default value, like identifier? 

> also seems like we have testing gap for this scenario
I will update the password expired test to capture the requirement validation hook, the reason I didnt do it initially was because I wasnt sure if we were going to remove that valdiation from the app, I know we talked about it during bug bash review but wanted to wait until they decided, but I will add it and we can just remove the test if that happens.

shuowu-okta: > do you mean in places where we are setting a default value, like identifier?

yep, those logic have been removed. and seems like you are doing refactor to remove fieldKey in schema, right? then maybe add a todo comment, and revisit once decision is made.

glenfannin-okta: Ah, makes sense, I will do that and link the bug ticket. 

How are we supposed to set default data from the transformer? There are a few scenarios where we need to do that, in this case, for setting the idenfier into the input field when it is available in the cookie?

shuowu-okta: that should not be changed, still touch formbag.data directly. but other transformation logic should depend on transaction object instead of the data bag

glenfannin-okta: Ahh, ok great! The changes in here are not new, I refactored a couple to stop passing data keys to components but that is is, it was previously using that functionality. I did this with hopes of having a better way of telling a component that it needs to watch on data changes of another field. 

The transformers that are using data to set a default value (like transformIdentify) seems to be standard practice, it is just pulling the value from the cookie (if it exists) and setting the default identifier. The issue was we were setting `defaultOption` but that doesn't get read or used in components, so I updated here to set it directly on the data bag for it to work correctly.

src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
```diff
@@ -39,22 +37,25 @@ const PasswordRequirements: UISchemaElementComponent<{
...
     id,
     requirements,
     settings,
     userInfo,
     validationDelayMs,
   } = uischema.options as PasswordRequirementsElement['options'];
-  const password = get(data, fieldKey);
+  // TODO: OKTA-521321 - Depending on the outcome of this ticket, we may be able to remove this
+  // If it is kept, need to add integration test for validation
+  const password = 'credentials.newPassword' in data
```
shuowu-okta, line 49: curious can you get password here during debugging? data should be an empty object when reach here

src/v3/src/transformer/i18n/transform.ts
```diff
@@ -73,6 +83,65 @@ const updateLabel = (transaction: IdxTransaction, element: FieldElement): void =
...
+    { field: 'identifier', key: 'error.username.required' },
+    { field: 'credentials.passcode', key: 'error.password.required' },
+  ].forEach((obj) => {
+    const ele = getUIElementWithName(
+      obj.field,
+      elements as UISchemaElement[],
+    ) as FieldElement;
+    if (ele && isCustomizedI18nKey(obj.key, widgetProps)) {
+      // eslint-disable-next-line no-param-reassign
+      dataSchema[obj.field] = {
```
shuowu-okta, line 107: i18n override should not touch validation logic, try find an approach to consume the overriding translation in validate function directly

src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
glenfannin-okta: 100% can, this is how the icons on the password requirements currently updates (changes from ❌  to ✅ ) using this same thing, the only difference here is I stopped passing this key from transformer.

src/v3/src/transformer/i18n/transform.ts
```diff
@@ -73,6 +83,65 @@ const updateLabel = (transaction: IdxTransaction, element: FieldElement): void =
...
+            message: loc(obj.key, 'login'),
+            i18n: { key: obj.key },
+          }];
+        },
+      };
+    }
+  });
+};
+
+const transformElementSubLabels = (
```
shuowu-okta, line 120: since you are touching logic here, can we move directly to `translations[]` approach instead of keep adding custom field in schema? if it's too much effort, we can start with identify form first

shuowu-okta: centralized translation step does not mean manage each custom translation in one file, but figure out a general pattern that can be applied to most/all custom steps.

src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
shuowu-okta: ohh, this one is under components dir, missed that

src/v3/src/components/PhoneAuthenticator/PhoneAuthenticator.tsx
```diff
@@ -69,19 +74,20 @@ const PhoneAuthenticator: UISchemaElementComponent<{
...
-          label={getTranslation(translations, 'extension')}
-          value={extension}
-          autocomplete="tel-extension"
-          onChange={(e: ChangeEvent<HTMLInputElement>) => {
+          id="phoneExtension"
+          onChange={(e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
             setExtension(e.currentTarget.value);
           }}
+          inputProps={{
+            'data-se': 'extension',
```
shawyu-okta, line 88: Should we try to get this from the transformer layer?

shawyu-okta: Same for the next input in this file

src/v3/src/transformer/identify/transformIdentify.ts
```diff
@@ -53,24 +59,14 @@ export const transformIdentify: IdxStepTransformer = ({ formBag, widgetProps, tr
...
-        defaultOption: username,
-      };
+      data.identifier = username;
     } else if (features?.rememberMe && features?.rememberMyUsernameOnOIE) {
       const usernameCookie = getUsernameCookie();
-      identifierElement.options = {
-        ...identifierElement.options,
-        defaultOption: usernameCookie ?? undefined,
-      };
+      data.identifier = usernameCookie ?? undefined;
```
shawyu-okta, line 69: Is the `??` necessary if the right side is going to be `undefined` anyways?

glenfannin-okta: Good catch, this is useless,  updating, ty!

src/v3/src/components/PhoneAuthenticator/PhoneAuthenticator.tsx
glenfannin-okta: Do you mean to like create adhoc FieldElements for Extension & Country? Would those then be attached to `options` of the `(*).phoneNumber` FieldElement?

shawyu-okta: Oh I see this isn't as simple as some of the other places where we can easily spread `data-se` in... Yeah would be trickier. We can leave hardcoded values in here if trying to do the spreading from transformer values adds a lot of complexity.

src/v3/src/transformer/identify/transformIdentify.ts
```diff
@@ -53,24 +59,14 @@ export const transformIdentify: IdxStepTransformer = ({ formBag, widgetProps, tr
...
-
   // add username/identifier from config if provided
   if (identifierElement) {
+    // add username/identifier from config if provided
     if (username) {
-      identifierElement.options = {
-        ...identifierElement.options,
-        defaultOption: username,
-      };
+      data.identifier = username;
```
shawyu-okta, line 66: Does this solve https://oktainc.atlassian.net/browse/OKTA-521302 ?

glenfannin-okta: Yessir it does! I will pull it in to my lane

shuowu-okta: Do you mind to fix that issue in a separate PR, it would be great to fix bug with a test (integration level).

glenfannin-okta: Created a PR with tests: https://github.com/okta/okta-signin-widget/pull/2699

src/v3/test/integration/__snapshots__/authenticator-enroll-data-phone.test.tsx.snap
```diff
@@ -1469,31 +1469,47 @@ exports[`authenticator-enroll-data-phone should render form 1`] = `
...
+                              <legend
+                                class="emotion-46"
+                              >
+                                <span
+                                  class="notranslate"
+                                >
+                                  ​
+                                </span>
+                              </legend>
+                            </fieldset>
```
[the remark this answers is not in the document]
glenfannin-okta: This seems to be a placeholder for MUI's default field label, see screenshot from MUI's website: 
image

src/v3/test/integration/__snapshots__/okta-verify-sms-channel-enrollment.test.tsx.snap
```diff
@@ -1348,31 +1348,47 @@ exports[`okta-verify-sms-channel-enrollment should render sms channel form and s
...
-                            </span>
-                          </label>
-                          <div
-                            class="ods-10aW0m"
+                            <p
+                              class="MuiTypography-root MuiTypography-body1 emotion-22"
+                            >
+                              +1
+                            </p>
+                          </span>
```
glenfannin-okta, line 1368: So it seems like the input adornment only controls the outermost element, but the prefix is still wrapped with the paragraph tag (we cant control that part it seems).

src/v3/test/integration/__snapshots__/authenticator-enroll-data-phone.test.tsx.snap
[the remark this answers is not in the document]
lesterchoi-okta: ugh. ok that's annoying but not worth fixing at the moment

src/v3/test/integration/__snapshots__/okta-verify-sms-channel-enrollment.test.tsx.snap
lesterchoi-okta: that's ok. it was a nit and we haven't verified that it violates a11y; just a hunch

lesterchoi-okta (approved):
LGTM

PR #2686, opened 2022-08-09
OKTA-488781 : Implementation for live (onChange) field validation updates

glenfannin-okta:
## Description:
The purpose of this PR is to handle live field change validation updates. This fixes the issue where when submitting a blank form, field level validation messages stay attached to the field. This also corrects the issue of clearing a required field, not triggering an error until the user submits.

The commented tests are just temporary until I get PR feedback to avoid unncessary change cycles to implementation and tests. 

WIP: Adding integration tests for live field change validation.

## PR Checklist

### Issue:

- [OKTA-488781](https://oktainc.atlassian.net/browse/OKTA-488781)

### Screenshot/Video:

https://user-images.githubusercontent.com/97472729/183540144-b001efda-e78e-449a-85ba-bce01575bb59.mov


src/v3/src/components/Form/renderers.tsx
```diff
@@ -136,7 +136,7 @@ export default [
           type, options, name, secret,
         } = {} as Input,
       },
-    }: InputTextElement) => ((type === 'string' || defaultType === 'string') && !options && !secret)
```
glenfannin-okta, line 139: This type was only used here and in component, no transformer used it so it is unnecessary, which is why I converted back to use the standard `FieldElement`

src/v3/src/components/PhoneAuthenticator/PhoneAuthenticator.test.tsx
```diff
@@ -64,7 +64,8 @@ jest.mock('../../hooks', () => ({
   useOnChange: () => mockHandleFunction,
 }));
 
-describe('PhoneAuthenticator tests', () => {
+// TODO: Re-enable before merging
+describe.skip('PhoneAuthenticator tests', () => {
```
glenfannin-okta, line 68: Ignore this, awaiting feedback to re-enable tests.

src/v3/src/components/PasswordWithConfirmation/PasswordWithConfirmation.tsx
```diff
@@ -57,28 +42,11 @@ const PasswordWithConfirmation: UISchemaElementComponent<{
...
-          onChange={handleConfirmPasswordChange}
-          fullWidth
-          inputProps={{
-            'data-se': name,
-            ...attributes,
-          }}
+        <InputPassword
+          uischema={confirmPasswordElement}
+          // @ts-ignore
+          customChangeHandler={handleConfirmPasswordChange}
```
shuowu-okta, line 48: nit: customChangeHandler -> onChange

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -67,4 +84,4 @@ const InputPassword: UISchemaElementComponent<{
   );
 };
 
-export default InputPassword;
+export default withFormValidationState(InputPassword);
```
shuowu-okta, line 87: nit: try replace hoc with hook, should be a bit better performance and consistent coding style

src/v3/src/transformer/i18n/transform.ts
```diff
@@ -142,6 +142,27 @@ const transformElementSubLabels = (
...
 
+const transformPhoneValidationError = (formBag: FormBag): void => {
+  const { dataSchema, uischema: { elements } } = formBag;
+  const phoneElement = elements.find((element) => (element as FieldElement)
+    .options.inputMeta.name?.endsWith('phoneNumber')) as FieldElement;
+
+  if (phoneElement) {
+    const fieldName = phoneElement.options.inputMeta.name;
+    dataSchema[fieldName] = {
+      validate(data) {
```
shuowu-okta, line 153: validate fn should not appear in i18n step

src/v3/src/components/hocs/withFormValidationState.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const { uischema } = props;
+    const {
+      options: {
+        inputMeta: {
+          // @ts-ignore expose type from auth-js
+          messages = {},
+        },
+      },
+    } = uischema;
+    const submissionError = messages?.value?.[0] && getMessage(messages.value[0]);
```
shuowu-okta, line 44: rename to `errorFromSchema` maybe, component should not know how error comes from, so it can handle it generically

src/v3/src/components/hocs/withFormValidationState.tsx
```diff
@@ -0,0 +1,67 @@
...
+    // For server side errors, need to reset the touched value
+    useEffect(() => {
+      setTouched(false);
+    }, [submissionError]);
+
+    const combinedProps = {
+      ...props,
+      touched,
+      setTouched,
+      error: touched ? error : submissionError,
```
shuowu-okta, line 58: does `touched` state really matter? how about just initial error state with info from uischema, then update it with later events.

src/v3/src/components/hocs/getDisplayName.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { FieldElement, RendererComponent } from '../../types';
+
+export const getDisplayName = (
```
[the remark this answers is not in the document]
glenfannin-okta: Yes, for debugging: https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -13,30 +13,43 @@
...
+}> = ({
+  uischema,
+  setTouched,
+  error,
+  setError,
+  onValidateHandler,
+  onChange,
+}) => {
   const value = useValue(uischema);
   const onChangeHandler = useOnChange(uischema);
```
shawyu-okta, line 37: Is there any way to change the signature of either `useOnChange` or `onChange` to be the same (i.e. both `ChangeEvent` or both `string`) so that this can be:
```js
const onChangeHandler = typeof onChange === 'function' ? onChange : useOnChange(uischema);
```

src/v3/src/transformer/field/transform.ts
```diff
@@ -92,7 +92,11 @@ export const transformStepInputs = (
       if (required && mutable !== false && type !== 'object') {
         acc.dataSchema[name] = {
           validate(data) {
-            const isValid = !!data[name];
+            // Phone number fields contain +[countryCode] characters by default
```
shawyu-okta, line 95: I'm having a hard time understanding this comment easily, can you clarify?

src/v3/src/hooks/useValue.ts
```diff
@@ -16,7 +16,7 @@ import { FieldElement } from '../types';
 export const useValue = (uischema: FieldElement): unknown => {
   const { name } = uischema.options.inputMeta;
 
-  const { data } = useWidgetContext();
+  const { data, additionalData } = useWidgetContext();
 
-  return data[name];
+  return { ...additionalData, ...data }[name];
```
shawyu-okta, line 21: What is `additionalData` and why isn't it in `data`?

src/v3/src/transformer/field/transform.ts
glenfannin-okta: W.r.t Phone Numbers, we cannot just blindly assert that the field is set, the reason for this is because phone numbers will always have a `+` and country code, set as the default value no matter what (this is set in PhoneAuthenticator.tsx). If we just assert on the fact that it is empty or not for validation, it will never block the user from leaving an empty field and will allow them to submit without entering a phone number in the phone number input field. 

Want me to update the comment in the code?

src/v3/src/hooks/useValue.ts
glenfannin-okta: Additional data (right now) really just contains the `confirmPassword` data, it was recommended to create this state prop for fields that should not be submitted to auth-js. I added it here as to not repeat the useValue logic just for this field. It wont cause cause because data and additional data should always contain unique keys.

src/v3/src/transformer/field/transform.ts
shawyu-okta: Ah, okay, that's what I suspected but I wanted to be sure. I think a comment that's more like "contain the '+' and country code characters by default" is clearer.

As a followup, we assert only `> 2` but if the country code is longer, won't this fail to properly prevent submitting an empty input field? e.g. U.K. is `+44` which is 3 characters.

src/v3/src/components/InputPassword/InputPassword.tsx
glenfannin-okta: I really like this approach but unless I do something like the below, it is not allowed, see: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level
Specifically: `Don’t call Hooks inside loops, conditions, or nested functions.`

```
const changeHandler = useOnChange(uischema);
const onChangeHandler = typeof onChange === 'function' ? onChange : changeHandler;
```

glenfannin-okta: I assumed the above would be acceptable so I pushed it, if not just let me know and I will revert.

src/v3/src/transformer/field/transform.ts
glenfannin-okta: Oh yea, I think I need some regex here to cover all country code scenarios 😮‍💨 Will update that.

shuowu: wait, iirc, we discussed to not do client side validation for phone number for now, and only follow regex pattern from backend when it's available in the response.

shuowu: and changes should not happen in the generic field transformation step

glenfannin-okta: This is only for `empty` field validation. You can't just blindly check the existence of data in the phone number field because by default the phone number will always have `+1` or `+[whatever country code is selected]` so it will always pass the validation check.

shuowu: then reassign dataschema for phone transformer, not here

glenfannin-okta: Regardless of what flow the phone number shows up in, this validation check rule should always apply, so doesn't that mean it belongs here? This logic is for setting up the validation rule for 'required' fields, when phone number is a required field, it just had different logic to validate whether it is set or not. Just curious?

shuowu: generic logic should not know anything about phone number, if there is a special scenario, handle it in that specific transformer, or dataSchema step.

glenfannin-okta: 👍 Ok, I will move to dataSchema.

shuowu: with handling it in specific transformer, I think you can just get the country code and remove it from data, so no regex should be needed. bad regex blocks user from finishing the flow

glenfannin-okta: Even in custom transformers, we wont know what part of the phone number is the country code as it is all just appended as one single string. 

`+2162` vs `+1216` vs `+202`

How will the transformer know what part of this string is a country code?

I could use the countryCodeUtil to perform a match on the string to extract the country code, and then assert on the remaining part of the string to determine if a value is set or not. Is that what you meant?

shuowu: possible to update dataSchema for phonenumber field to include countryCode while it changes, then pass it as options to validate function. then you should have enough info to proceed with required validation.

src/v3/src/components/InputPassword/InputPassword.tsx
shawyu-okta: Oh right, ok I forgot about `rules of hooks`. The change looks good to me.

src/v3/src/transformer/field/transform.ts
shuowu: maybe it's not a bad idea to separate this feature into another PR, and keep this pr focused

glenfannin-okta: Ok, this is going to require some complex logic, I will remove this change and create a separate PR specifically for phone number validation.

PR #2692, opened 2022-08-09
Re-enable `e2e-v3` suite for merges in CI

shawyu-okta:
## PR Checklist

### Issue:

- [OKTA-518087](https://oktainc.atlassian.net/browse/OKTA-518087)

### DIFF
--- a/src/v3/package.json
+++ b/src/v3/package.json
@@ -37,7 +37,7 @@
     "build": "preact build --no-prerender --no-sw --dest ../../dist/dist/js",
     "clean-svg": "svgo -f src/img/authCoinIcons",
     "codegen": "yarn workspace @okta/okta-signin-widget codegen",
-    "dev": "preact watch --no-prerender --no-sw --port 3000",
+    "dev": "yarn codegen && preact watch --no-prerender --no-sw --port 3000",
     "lint": "yarn codegen && eslint .",
     "lint:report": "scripts/buildtools/lint-report.sh",
     "serve": "sirv ../../dist/dist/js --port 3000",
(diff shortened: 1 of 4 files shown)


src/v3/package.json
```diff
@@ -37,7 +37,7 @@
     "build": "preact build --no-prerender --no-sw --dest ../../dist/dist/js",
     "clean-svg": "svgo -f src/img/authCoinIcons",
     "codegen": "yarn workspace @okta/okta-signin-widget codegen",
-    "dev": "preact watch --no-prerender --no-sw --port 3000",
+    "dev": "yarn codegen && preact watch --no-prerender --no-sw --port 3000",
```
shuowu-okta, line 40: how about adding codegen to test:e2e?

lesterchoi-okta: It's `dev` that actually breaks if `codegen` isn't run first, right? This seems like the right place IMO

lesterchoi-okta (approved):

src/v3/package.json
shawyu-okta: I was trying to decide the best place to put this. I went with this since the e2e test script runs the `dev` script, and if you do a fresh clone of the repo, trying to run `dev` will fail. I feel that this covers both situations well. Any objection?

shuowu-okta: codegen takes time, and most probably we will need to run it once and it can be used for a long term, but we may restart local dev server several time. and I am ok with the current place, just feel it can become a bit annoying when restart happens

shuowu-okta (approved):

PR #2694, opened 2022-08-10
chore: expose root mocks as workspace

shuowu-okta:
## Description:

This PR can be taken as the first step to using root mocks in jest integration tests, so both v2 and v3 follows the same mocks.

Changes:

* exposes root mocks as yarn workspace, so it can be referred as a isolated module from v3
* update `enroll-profile` jest integration test with new mock from @okta/mocks

## PR Checklist

### Issue:

- [OKTA-522967](https://oktainc.atlassian.net/browse/OKTA-522967)


lesterchoi-okta (changes requested):
This should be a change to webpack configs for alias module resolution

shuowu-okta:
@lesterchoi-okta why webpack config matters? this is for jest, and I don't see issue with making mocks as a workspace package.

lesterchoi-okta:
> @lesterchoi-okta why webpack config matters? this is for jest, and I don't see issue with making mocks as a workspace package.

Then it should be a change to `jest.config.js` `moduleNameMapper`, right? Why do you need to add another workspace? There are no dependencies in the package.json and this change appears just to make the import path simpler.

shuowu-okta:
mostly to avoid the complex relative path pattern, and I do feel this is the proper pattern to use external modules (outside of v3), if we config moduleMapper, code still will be changed when folder changes, but one isolated workspace can just avoid all future changes. If you search in the repo, you will find @okta/env is used the same way. It's actually the real power of using monorepo.

lesterchoi-okta (approved):
discussed offline. using workspace pattern is fine since there is precedent, e.g., `@okta/env`

PR #2699, opened 2022-08-11
OKTA-513224 : Adding implementation for pre-populating identifier

glenfannin-okta:
## Description:
The purpose of this PR is to auto-populate the identifier into the username field when set in widget config props or stored in `ln` cookie to match parity with v2.

## PR Checklist

### Issue:

- [OKTA-513224](https://oktainc.atlassian.net/browse/OKTA-513224)

### Screenshot/Video:

https://user-images.githubusercontent.com/97472729/184170219-099e6e1b-ae19-4fc4-b987-57ac036a584f.mov

https://user-images.githubusercontent.com/97472729/184170243-2bc18140-088b-45cf-b820-188fc33cbdb9.mov


shuowu-okta (approved):

src/v3/test/integration/identify-with-password.test.tsx
```diff
@@ -15,6 +15,9 @@ import { setup } from './util';
 
 import mockResponse from '../../src/mocks/response/idp/idx/introspect/default.json';
 
+const mockUsername = 'testuser@okta1.com';
+jest.mock('js-cookie', () => ({ get: () => mockUsername }));
```
shawyu-okta, line 19: suggestion: we only mock these values on the tests where they are relevant

shawyu-okta (approved):

src/v3/test/integration/identify-with-password.test.tsx
glenfannin-okta: I initially tried that but for some reason the mock did not take, it only triggered when I added it to the top of the test. I have seen this happen before with other tests so i thought it was a standard thing.

glenfannin-okta: Maybe I need to follow [these steps ](https://stackoverflow.com/questions/71470603/jest-mocks-only-work-at-top-of-file-never-inside-describe-or-it-blocks)to get the mock to work inside the `it`, I will give it a try.

shuowu:
In the future refactor, identity transformer and identifyWithPassword transformer should be separated based on current authenticator.

src/v3/test/integration/identify-with-password.test.tsx
glenfannin-okta: Ok so those steps did not work, and looking into their documentation it looks like jest works this way for ES module imports: https://jestjs.io/docs/manual-mocks#using-with-es-module-imports

However, it says using `jest.doMock` should work but trying it does not work. It seems like the automatic hoisting is causing this. I ended up using `spyOn` instead of mock to keep it scoped to the actual test using it.

PR #2703, opened 2022-08-11
OKTA-523580: set version and commit sha

lesterchoi-okta:
## Description:

Set version and commit sha for SIW_NEXT Selenium3 test to use.

- set `data-version` and `data-commit` on widget root component
- set static properties `__version` and `__commit` on `OktaSignIn` class

## PR Checklist

- [x] Added unit tests? YES, but skipped due to type errors

- [ ] Added e2e tests? NO

- Did you verify the change by running downstream monolith artifacts? PENDING

### Issue:

- [OKTA-523580](https://oktainc.atlassian.net/browse/OKTA-523580)


src/v3/src/components/AuthContainer/AuthContainer.tsx
```diff
@@ -14,6 +14,7 @@ import { Box, useMediaQuery } from '@mui/material';
 import classNames from 'classnames';
 import { FunctionComponent, h } from 'preact';
 
+import { version } from '../../../package.json';
```
shawyu-okta, line 17: Would it be better to read this off the global or an env variable so we don't bundle package.json?

shawyu-okta (approved):

shuowu-okta:
We have already exposed commit hash in request header, can that be reused in selenium?

lesterchoi-okta:
> We have already exposed commit hash in request header, can that be reused in selenium?

More trouble that it's worth: we'd have to wait for . This is for a very basic smoke test just to check that v3 is mounted and rendered.

shuowu-okta (approved):

PR #2704, opened 2022-08-11
refactor: i18n transformer

shuowu-okta:
## PR Checklist

### Issue:

### DIFF
--- a/src/v3/src/transformer/i18n/transform.ts
+++ b/src/v3/src/transformer/i18n/transform.ts
@@ -10,66 +10,47 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { IdxTransaction, Input } from '@okta/okta-auth-js';
+import { IdxTransaction } from '@okta/okta-auth-js';
 
 import { getI18NParams, getI18NValue, getMessage } from '../../../../v2/ion/i18nTransformer';
-import { AUTHENTICATOR_KEY } from '../../constants';
 import {
+  AuthenticatorButtonElement,
   FieldElement,
   FormBag,
   TransformStepFnWithOptions, UISchemaElement,
 } from '../../types';
 import { loc } from '../../util';
-import { getOptionValue } from '../selectAuthenticator/utils';
+import { traverseLayout } from '../util';
 
-const geti18nPath = (fieldName: string, stepName: string, authenticatorKey: string): string => {
-  const authenticatorKeyPath = authenticatorKey
-    ? `.${authenticatorKey}`
-    : '';
-  return `${stepName}${authenticatorKeyPath}.${fieldName}`;
-};
-
-const updateLabel = (transaction: IdxTransaction, element: FieldElement): void => {
+const getI18nOptions = (transaction: IdxTransaction, element: UISchemaElement) => {
   const { nextStep } = transaction;
-  const { relatesTo, name: stepName = '' } = nextStep || {};
+  const { relatesTo, name: stepName = '' } = nextStep!;
   const authenticatorKey = relatesTo?.value?.key || '';
-  const i18nPath = geti18nPath(element.options.inputMeta.name, stepName, authenticatorKey);
+  const fieldName = (element as FieldElement).options?.inputMeta?.name;
 
-  if (element.label) {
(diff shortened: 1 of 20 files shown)


src/v3/src/transformer/i18n/transform.ts
```diff
@@ -10,75 +10,58 @@
...
+    path = authenticatorKey
+      ? `${stepName}.${authenticatorKey}.${fieldName}`
+      : `${stepName}.${fieldName}`;
+  } else if (element.type === 'AuthenticatorButton') {
+    const methodType = (
+      element as AuthenticatorButtonElement
+    ).options.authenticator?.methods?.[0]?.type;
+    path = methodType
+      ? `${stepName}.authenticator.${(element as AuthenticatorButtonElement).options.key}.${methodType}`
+      : `${stepName}.authenticator.${(element as AuthenticatorButtonElement).options.key}`;
```
glenfannin-okta, line 43: I believe something with this logic may be causing the Email Authenticator (and other) label to not be translated properly. See attached screenshots: 
Screen Shot 2022-08-12 at 11 08 40 AM
Screen Shot 2022-08-12 at 11 10 01 AM

 
Screen Shot 2022-08-12 at 11 10 08 AM

glenfannin-okta:
Here are some sample current state screenshots from 0.0 branch for the authenticator list view: 
Screen Shot 2022-08-12 at 11 04 03 AM
Screen Shot 2022-08-12 at 11 13 25 AM
Screen Shot 2022-08-12 at 11 13 34 AM
Screen Shot 2022-08-12 at 11 13 41 AM

src/v3/src/transformer/i18n/transform.ts
shuowu-okta: should be fine for beta, as it only target to English, right now loc fn is called everywhere, so it's not easy to add tests in jest to capture english leaks, but it will be handled once i18n transformer is fully refactored.

curious is it quick to add english leaks tests to v3, or will it slow development? since I remember we discussed it not top priority for beta.

glenfannin-okta: We'd definitely want to create an English leaks test, but it will probably be after we get through the bugs first since I think those are highest priority right now?

glenfannin-okta: > I also feel we should remove fallback logic in translation, which causes the leaks.

That may be a question to pose to product, if we dont fall back, nothing will display, and that could be considered a worse experience (because at least some browsers can perform their own translation).

shuowu-okta: not really, no default value means it fails in the first layer tests, feature even will not be shipped

glenfannin-okta: Ohhh, I see what you mean. I thought you meant something else.

shuowu-okta: and that's also why we will want to
1) move all translation logic into one transformation step
2) only use translations field in component

I am doing refactor part2 now, hopefully I can finish most of that by the EOD

glenfannin-okta, line 43: This is the part that causes the issue with the authenticator labels. As you can see in the map, only a few authenticators uses the method Type in the mapping (only OV custom_app), but the methodType is available on more than that from the ion response. 

I think this should be updated to fix that so we wont have to go back and research why it doesn't work later (it should be a fairly straightforward fix).

https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/i18nTransformer.js#L66-L91

shuowu-okta: will be handled in my next PR, still we will need tests, otherwise it can regress at any time

glenfannin-okta: Agree, there is already an ok-pl test in v2 that I thought we would re-use. I tried loading the language in jest but it didnt seem to work.

shuowu-okta: don't worry about that, there will be a generic in jest layer to run against all mocks (only against transformation layer), but not sure when will it be added

glenfannin-okta (approved):

PR #2711, opened 2022-08-15
OKTA-521364 : Fix for field level data being carried over between transactions

glenfannin-okta:
## Description:
This PR fixes the issue that surfaces when navigating between pages that re-use the same component (i.e. Navigating from password page and security question verification page). 

## PR Checklist

### Issue:

- [OKTA-521364](https://oktainc.atlassian.net/browse/OKTA-521364)

### DIFF
--- a/src/v3/src/components/Form/Layout.tsx
+++ b/src/v3/src/components/Form/Layout.tsx
@@ -78,11 +78,12 @@ const Layout: FunctionComponent<LayoutProps> = ({ uischema }) => {
 
           const uischemaElement = (element as UISchemaElement);
           const fieldElement = (uischemaElement as FieldElement);
-          const key = fieldElement.options?.inputMeta?.name && fieldElement.label ? `${fieldElement.options?.inputMeta?.name}_${fieldElement.label}` : elementKey;
+          const key = fieldElement.options?.inputMeta?.name && fieldElement.label
+            ? `${fieldElement.options?.inputMeta?.name}_${fieldElement.label}`
+            : elementKey;
           const Component = renderer.renderer as UISchemaElementComponent;
           return (
             <Box
-              // key={elementKey}
               key={key}
               // eslint-disable-next-line react/jsx-props-no-spreading
               {...(!(uischemaElement).noMargin && { marginBottom: 4 })}
--- a/src/v3/src/components/Form/Layout.tsx
+++ b/src/v3/src/components/Form/Layout.tsx
@@ -14,6 +14,7 @@ import { Box } from '@mui/material';
 import { FunctionComponent, h } from 'preact';
 
 import {
+  FieldElement,
   StepperLayout,
   UISchemaElement,
   UISchemaElementComponent,
@@ -75,14 +76,18 @@ const Layout: FunctionComponent<LayoutProps> = ({ uischema }) => {
             return null;
           }
 
+          const uischemaElement = (element as UISchemaElement);
+          const fieldElement = (uischemaElement as FieldElement);
+          const key = fieldElement.options?.inputMeta?.name && fieldElement.label ? `${fieldElement.options?.inputMeta?.name}_${fieldElement.label}` : elementKey;
           const Component = renderer.renderer as UISchemaElementComponent;
           return (
             <Box
-              key={elementKey}
+              // key={elementKey}
+              key={key}
               // eslint-disable-next-line react/jsx-props-no-spreading
-              {...(!(element as UISchemaElement).noMargin && { marginBottom: 4 })}
+              {...(!(uischemaElement).noMargin && { marginBottom: 4 })}
             >
-              <Component uischema={element as UISchemaElement} />
+              <Component uischema={uischemaElement} />
             </Box>
           );
         })
--- a/src/v3/src/components/Form/Layout.tsx
+++ b/src/v3/src/components/Form/Layout.tsx
@@ -28,9 +28,10 @@ import Stepper from './Stepper';
 
 type LayoutProps = {
   uischema: UISchemaLayout;
+  stepName?: string;
 };
 
-const Layout: FunctionComponent<LayoutProps> = ({ uischema }) => {
+const Layout: FunctionComponent<LayoutProps> = ({ uischema, stepName }) => {
   const { type, elements } = uischema;
 
   const isHorizontalLayout = type === UISchemaLayoutType.HORIZONTAL;
@@ -44,7 +45,8 @@ const Layout: FunctionComponent<LayoutProps> = ({ uischema }) => {
     >
       {
         elements.map((element, index) => {
-          const elementKey = `${element.type}_${index}`;
+          const stepNameKeyPart = stepName ? `_${stepName}_` : '';
+          const elementKey = `${element.type}${stepNameKeyPart}${index}`;
 
           if (element.type === UISchemaLayoutType.STEPPER) {
             return (
@@ -61,6 +63,7 @@ const Layout: FunctionComponent<LayoutProps> = ({ uischema }) => {
               <Layout
                 key={elementKey}
                 uischema={element as UISchemaLayout}
+                stepName={stepName}
               />
             );
           }
@@ -78,8 +81,8 @@ const Layout: FunctionComponent<LayoutProps> = ({ uischema }) => {
 
           const uischemaElement = (element as UISchemaElement);
           const fieldElement = (uischemaElement as FieldElement);
-          const key = fieldElement.options?.inputMeta?.name && fieldElement.label
-            ? `${fieldElement.options?.inputMeta?.name}_${fieldElement.label}`
+          const key = fieldElement.options?.inputMeta?.name && stepNameKeyPart
+            ? `${fieldElement.options?.inputMeta?.name}${stepNameKeyPart}`
             : elementKey;
           const Component = renderer.renderer as UISchemaElementComponent;
           return (
--- a/src/v3/src/components/Form/Layout.tsx
+++ b/src/v3/src/components/Form/Layout.tsx
@@ -45,8 +45,7 @@ const Layout: FunctionComponent<LayoutProps> = ({ uischema, stepName }) => {
     >
       {
         elements.map((element, index) => {
-          const stepNameKeyPart = stepName ? `_${stepName}_` : '';
-          const elementKey = `${element.type}${stepNameKeyPart}${index}`;
+          const elementKey = [element.type, stepName, index].join('_');
 
           if (element.type === UISchemaLayoutType.STEPPER) {
             return (
@@ -81,8 +80,8 @@ const Layout: FunctionComponent<LayoutProps> = ({ uischema, stepName }) => {
 
           const uischemaElement = (element as UISchemaElement);
           const fieldElement = (uischemaElement as FieldElement);
-          const key = fieldElement.options?.inputMeta?.name && stepNameKeyPart
-            ? `${fieldElement.options?.inputMeta?.name}${stepNameKeyPart}`
+          const key = fieldElement.options?.inputMeta?.name && stepName
+            ? `${fieldElement.options?.inputMeta?.name}_${stepName}`
             : elementKey;
           const Component = renderer.renderer as UISchemaElementComponent;
           return (
--- a/src/v3/src/components/InformationalText/InformationalText.tsx
+++ b/src/v3/src/components/InformationalText/InformationalText.tsx
@@ -20,7 +20,7 @@ const InformationalText: UISchemaElementComponent<{
 }> = ({
   uischema,
 }) => {
-  const { content, variant } = uischema.options;
+  const { content } = uischema.options;
 
   return (
     <Box
@@ -30,8 +30,6 @@ const InformationalText: UISchemaElementComponent<{
       <Typography
         paragraph
         data-se="o-form-explain"
-        // eslint-disable-next-line react/jsx-props-no-spreading
-        {...(variant && { variant })}
       >
         {content}
       </Typography>
--- a/src/v3/src/components/InputPassword/InputPassword.tsx
+++ b/src/v3/src/components/InputPassword/InputPassword.tsx
@@ -33,7 +33,6 @@ const InputPassword: UISchemaElementComponent<{
       messages = {},
       name,
     },
-    hideLabel,
   } = uischema.options;
   const error = messages?.value?.[0] && getMessage(messages.value[0]);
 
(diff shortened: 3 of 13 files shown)


lesterchoi-okta:
Looks like you need a rebase after my merge

src/v3/src/components/InformationalText/InformationalText.tsx
```diff
@@ -30,8 +30,6 @@ const InformationalText: UISchemaElementComponent<{
       <Typography
         paragraph
         data-se="o-form-explain"
-        // eslint-disable-next-line react/jsx-props-no-spreading
-        {...(variant && { variant })}
```
glenfannin-okta, line 34: Variant is useless here because this component is used as a paragraph. It was being used for H* elements but we already have a dedicated component for that `Heading`. But ultimately did not need this.

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -44,8 +43,7 @@ const InputPassword: UISchemaElementComponent<{
   return (
     <Box>
       <PasswordInput
-        // eslint-disable-next-line react/jsx-props-no-spreading
-        {...(!hideLabel && { label })}
```
glenfannin-okta, line 48: Trying to hide the label was causing issues. it was only used for security question verification but it was using a paragraph tag as the label. This doesnt match parity and did not pass a11y checks. We need to use the built in label with this component.

src/v3/src/transformer/field/transform.ts
```diff
@@ -13,27 +13,27 @@
...
-  FormBag, TransformStepFnWithOptions, UISchemaElement,
+  FieldElement,
+  FormBag, TransformStepFnWithOptions,
 } from '../../types';
 import { flattenInputs } from '../../util';
 import { transformer as attributesTransformer } from './attributes';
 import { transformer as typeTransformer } from './type';
 
-const mapUiElement = (input: Input): UISchemaElement => {
-  const { name, label } = input;
```
glenfannin-okta, line 23: `name` was not being used in transformers so removed it here.

src/v3/src/transformer/layout/securityQuestion/securityQuestionEnroll.ts
```diff
@@ -46,14 +46,15 @@ export const transformSecurityQuestionEnroll: IdxStepTransformer = ({ transactio
   const { uischema, data, dataSchema } = formBag;
 
   const predefinedQuestionOptions = inputs?.[0]?.options?.filter(({ value }) => !(value as Input[])?.some(({ name }) => name === 'question'));
+  const customQuestionOptions = inputs?.[0]?.options?.filter(({ value }) => (value as Input[])?.some(({ name }) => name === 'question'));
```
glenfannin-okta, line 49: We were re-using the predefined questions elements for custom answer field. This causes a problem when errors are set because when an error is set on the custom element, it will not surface to the UI since the predefined answer element is used. (This fixes this issue).

src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
```diff
@@ -35,20 +34,15 @@ export const transformSecurityQuestionVerify: IdxStepTransformer = ({ transactio
...
   };
 
-  const questionElement: DescriptionElement = {
-    type: 'Description',
-    noMargin: true,
-    options: {
-      content: relatesTo?.value?.profile?.question as string,
-      variant: 'h6',
-    },
-  };
```
glenfannin-okta, line 45: This was using a paragraph tag as a label for the security question verify field. This should use the actual label to resolve a11y issues and match parity with v2.

src/v3/src/components/InputPassword/InputPassword.tsx
shuowu-okta: which a11y check it fails?

shuowu-okta: btw, it's allowed to hide label for input
https://www.w3.org/WAI/tutorials/forms/labels/#hiding-label-text

glenfannin-okta: If we were using MUI's raw component, absolutely (there is a dedicated attr for that) but this is ODY's component we are using and it does not allow it. You can even see in the snapshot there was still a label attached to the DOM with no value with the `for` attached to the credentials.answer field. 

https://github.com/okta/odyssey/blob/develop/packages/odyssey-react-mui/src/components/PasswordInput/PasswordInput.tsx

shuowu-okta: we will want the form make sense for the end user first, it's not about which comp library we are using, if there is an issue, then we should report and fix it

shuowu-okta:
Curious which line fixes the issue?

src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
```diff
@@ -35,20 +34,15 @@ export const transformSecurityQuestionVerify: IdxStepTransformer = ({ transactio
...
+  const answerElement: FieldElement = getUIElementWithName(
     'credentials.answer',
     uischema.elements as UISchemaElement[],
-  ) as InputTextElement;
-  answerElement.options.hideLabel = true;
+  ) as FieldElement;
+  answerElement.translations = [{
+    name: 'label',
+    i18nKey: '',
+    value: relatesTo?.value?.profile?.question as string,
```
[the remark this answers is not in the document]
glenfannin-okta: I'm not totally sure, we can talk to Alex about this, but for parity, this is what current v2 looks like

image
image

glenfannin-okta:
> Curious which line fixes the issue?

This: https://github.com/okta/okta-signin-widget/pull/2711/files#diff-b84e9e4e678a95194e21610a7921b240718e595e29e5ff9c9cbcb6477d2ff017R79-R87

shuowu-okta:
the key?

glenfannin-okta:
> the key?

Yep, the issue was that the elements were not being unmounted and re-mounted properly when being re-used between views/transactions. React thought since the element key was the same, it should not unmount the component causing the data to remain in the field.

src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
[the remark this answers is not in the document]
shuowu-okta: ok, I didn't notice that in v2. if that's the case let's follow v2 first, and we can change v2 afterwards even if it's an issue

src/v3/src/components/Form/Layout.tsx
```diff
@@ -75,14 +76,19 @@ const Layout: FunctionComponent<LayoutProps> = ({ uischema }) => {
             return null;
           }
 
+          const uischemaElement = (element as UISchemaElement);
+          const fieldElement = (uischemaElement as FieldElement);
+          const key = fieldElement.options?.inputMeta?.name && fieldElement.label
+            ? `${fieldElement.options?.inputMeta?.name}_${fieldElement.label}`
```
shuowu-okta, line 82: maybe bring in stepName (remediation name)  as a prop from From can be more robust

glenfannin-okta: Yea, so I was talking to Lester about this,I settled on this small change to prevent this PR from becoming too large, I am creating a ticket to resolve it in a more scalable and complete fashion, it requires heavy testing with regards to flows (OV, Email, and security question enroll) to ensure nothing is regressed.

shuowu-okta: pass a prop from the component one level up should not be a big change, and if it's only be added as a key prefix, I don't really think it will regress anything

lesterchoi-okta: Good idea if step-name will make the key unique (vs. transaction id as discussed on our call).

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -32,7 +32,6 @@ const InputPassword: UISchemaElementComponent<{
       messages = {},
       name,
     },
-    hideLabel,
```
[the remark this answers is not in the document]
glenfannin-okta: This was only added and used for one flow (security question verify). Initially I think it was thought that the question should display as a Header and the answer field should be displayed without a label. But comparing to current state of v2. The security question is displayed as a label attached to the to answer input field. 

Additionally, I removed it because if we want to really hide the label for a password component, we will not be able to use the ODY component, we will need to use generic MUI components because the ODY component does not allow the label to be hidden, it is baked into their component, and by not passing a label, it just creates an empty label element in the dom. 

We don't have a need right now to hide the label for any flow, but I figured if we do, we can introduce logic at that point (which doesn't use the ODY component since it isnt compatible).

src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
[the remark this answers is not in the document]
lesterchoi-okta: > does label a input with a question make sense? I feel if screen reader reads a full question for as answer input confuses end user

I don't think it's a problem to use a question as the label, MDN example: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

src/v3/test/integration/__snapshots__/authenticator-verification-security-question.test.tsx.snap
```diff
@@ -108,34 +94,35 @@ exports[`authenticator-verification-security-question renders form 1`] = `
                       class="MuiBox-root emotion-4"
                     >
                       <label
-                        class="MuiInputLabel-root MuiFormLabel-root MuiFormLabel-colorPrimary emotion-20"
+                        class="MuiInputLabel-root MuiFormLabel-root MuiFormLabel-colorPrimary emotion-17"
```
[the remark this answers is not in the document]
glenfannin-okta: :( Unfortunately yes, I have just been running tests locally to see what fails and why, if it is a simple emotion class update I just force update the snapshot. This one has a valid change though, for the security question label.

src/v3/src/components/Form/Layout.tsx
```diff
@@ -43,7 +45,8 @@ const Layout: FunctionComponent<LayoutProps> = ({ uischema }) => {
     >
       {
         elements.map((element, index) => {
-          const elementKey = `${element.type}_${index}`;
+          const stepNameKeyPart = stepName ? `_${stepName}_` : '';
+          const elementKey = `${element.type}${stepNameKeyPart}${index}`;
```
lesterchoi-okta, line 49: 1. Is the underscore delimiter just an arbitrary choice? AFAICT this is going to result in something like `VerticalLayout_identify-with-password_0`
2. Do we still need the index? Index shouldn't be used in the key prop if possible... eslint would complain if the key was constructed inside the JSX. I guess the problem is that it's possible/likely we'll have more than one of `VerticalLayout` or `HorizontalLayout`... we can address it separately.
3. nit:
```suggestion
          const elementKey = [element.type, stepName, index].join('_');
```

src/v3/src/components/InputPassword/InputPassword.tsx
lesterchoi-okta: We should also ask Odyssey to make the label more flexible, but we can stick with the question as the label for now since it meets parity and doesn't seem to violate a11y requirements.

src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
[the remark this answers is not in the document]
lesterchoi-okta: Why is there a non-breaking space after the question (`&nbsp;`)? It doesn't look like it's present in the l10n file

src/v3/test/integration/__snapshots__/authenticator-verification-security-question.test.tsx.snap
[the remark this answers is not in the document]
lesterchoi-okta: Also strange the `&nbsp;` doesn't appear in the snapshot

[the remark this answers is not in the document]
lesterchoi-okta: [OKTA-524480](https://oktainc.atlassian.net/browse/OKTA-524480)

lesterchoi-okta (approved):

src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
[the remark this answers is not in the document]
glenfannin-okta: Looks like this is hardcoded by courage: https://github.com/okta/okta-signin-widget/blob/989988cf4c4f80a63a1c1deb6791e24fd161cf73/packages/%40okta/courage-dist/esm/src/courage/views/forms/helpers/InputLabel.js#L136

We don't have this issue in v3.

src/v3/test/integration/__snapshots__/authenticator-verification-security-question.test.tsx.snap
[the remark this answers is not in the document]
glenfannin-okta: Thanks for that ticket! that will help tremendously. 

Also, the `&nbsp;` screenshot is from v2 widget. V3 does not have that, the blank space is added by courage: https://github.com/okta/okta-signin-widget/blob/989988cf4c4f80a63a1c1deb6791e24fd161cf73/packages/%40okta/courage-dist/esm/src/courage/views/forms/helpers/InputLabel.js#L136

src/v3/src/components/Form/Layout.tsx
glenfannin-okta: 1. yep is it, I know it doesnt matter in the grand scheme of things (as long as it is unique) but do we care if it is delimited or not?
2. We sort of still need the index because of components rendered that do not have a name but same type, for example a view that has multiple informational text elements will throw a console error of duplicate keys since the types and step name will be the same. I still think we need some sort of unique identifier that can be set on each element that can be used at this layout layer when constructing a key to be certain it is unique.
3. Thank you 🙏

src/v3/test/integration/__snapshots__/authenticator-verification-security-question.test.tsx.snap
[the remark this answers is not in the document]
lesterchoi-okta: ah. gotcha... wonder why it adds that

src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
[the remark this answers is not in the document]
shuowu-okta: but from screen reader perspective, it might be confusing to read question for an answer input, maybe it just confuses me

PR #2713, opened 2022-08-15
OKTA-523716 : Fix for global error messages persisting between views

leemchale-okta:
## Description:

Fixes the issue of global error messages persisting between views.

## PR Checklist

### Issue:

- [OKTA-523716](https://oktainc.atlassian.net/browse/OKTA-523716)

### DIFF
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -51,6 +51,7 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
       stepToRender,
     } = options;
 
+    // TODO: Revisit and refactor this function as it is a dupe of handleError fn in Widget/index.tsx
     const handleError = (error: unknown) => {
       // TODO: handle error based on types
       // AuthApiError is one of the potential error that can be thrown here
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { IdxActionParams } from '@okta/okta-auth-js';
+import { AuthApiError, IdxActionParams } from '@okta/okta-auth-js';
 import { omit } from 'lodash';
 import merge from 'lodash/merge';
 import { useCallback } from 'preact/hooks';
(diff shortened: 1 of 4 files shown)


src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -47,6 +50,20 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
...
+      // AuthApiError is one of the potential error that can be thrown here
+      // We will want to expose development stage errors from auth-js and file jiras against it
+      setAuthApiError(error as AuthApiError);
+      console.error(error);
+      // error event
+      events?.afterError?.({
+        stepName: currTransaction?.nextStep?.name,
+      }, error);
+      return null;
+    };
```
glenfannin-okta, line 65: Can you make this a function in the `/util` path, and re-use this here and in widget/index.tsx?

glenfannin-okta (commented):
Overall this looks good, can you add an integration test for verify this scenario (just to ensure we do not regress on it in future iterations/changes).

src/v3/src/hooks/useOnSubmit.ts
leemchale-okta: yup will do

shuowu-okta: maybe a hook if you want to generalize the logic, since it has component state handling included

leemchale-okta: Ended up not being able to make a hook because im getting the lint error: 
`React Hook "useOnError" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function`.  The custom useOnError hook cant be used in the widget index because it doesnt have access to the context yet.

shuowu: i see, it's a hook already, personally I don't think this handleError fn will be called in other places, since useOnSubmit is  the only centralized place to trigger idx.proceed. For other calls, like idx.start, you probably will want to handle errors differently

glenfannin-okta: That same function is used in the `bootstrap` callback and the `resume` callback: https://github.com/okta/okta-signin-widget/blob/0.0/src/v3/src/components/Widget/index.tsx#L173-L202

That is why I recommended extracting to a function, are you ok with this function being duplicated in both places?

shuowu-okta: in this case, it can be wrapped in `useCallback` and shared via context. util dir is used to collect pure functions, so it can be used/tested isolatedly

glenfannin-okta (approved):
We can go forward with this and revisit consolidating the two handle error functions later

src/v3/src/hooks/useOnSubmit.ts
glenfannin-okta: Cool, going to create a ticket for this to follow up and implement this logic

glenfannin-okta (approved):

leemchale-okta:
@lesterchoi-okta nope not anymore i am merging now

lesterchoi-okta:
> @lesterchoi-okta nope not anymore i am merging now

Wanted to squeeze this into another deployment to demo cell if possible since it's highly visible and may squash a couple of bug tickets if QA can confirm

leemchale-okta:
> 

Ah I see okay sounds good it has been merged now

PR #2715, opened 2022-08-16
fix(identify): re-render if idx.unknown.user message

sherizada-okta:
## Description:
- the bug: Identify form was stuck in saving mode if the user entered the same identifier twice in a row because form does not re-render if previous and current API responses are identical
- the fix: re-render if API response contains a message with "idx.unknown.user" key

## PR Checklist

### Issue:

- [OKTA-520158](https://oktainc.atlassian.net/browse/OKTA-520158)

### Screenshot/Video:
Bug:

https://user-images.githubusercontent.com/71431120/184957055-4cadbe5f-9bff-4934-a148-0cc91d316d13.mov

Fix:

https://user-images.githubusercontent.com/71431120/184957287-e3c4ba56-187e-4c9e-bac5-fa57dfbb57cb.mov


test/unit/spec/v2/models/AppState_spec.js
```diff
@@ -187,6 +187,53 @@ describe('v2/models/AppState', function() {
       expect(this.appState.shouldReRenderView(transformedResponse)).toBe(false);
     });
 
+    it('rerender view should be true if idx message contains idx.unknown.user', () => {
```
lesterchoi-okta, line 190:
```suggestion
    it('should re-render the view if the idx message contains idx.unknown.user', () => {
```

src/v2/models/AppState.ts
```diff
@@ -314,26 +315,31 @@ export default class AppState extends Model {
...
+         */
+        reRender = true;
+      } else if (FORMS_WITH_STATIC_BACK_LINK.includes(this.get('currentFormName'))) {
+        /**
+         * returns true: We want to force reRender if you go back to selection screen from challenge or enroll screen
+         * and re-select the same authenticator for challenge. In this case also new response will be identical
+         * to the old response.
+         */
+        reRender = true;
+      } else if (this.containsMessageWithI18nKey(UNKNOWN_USER_I8N_KEY)) {
```
lesterchoi-okta, line 333: Is this the only new case that needs to be re-rendered? i.e., is the `idx.unknown.user` response actually special or is it just the one that was noticed/reported?

aarongranick-okta (approved):

lesterchoi-okta (commented):
I would check to see if matching on `idx.unknown.user` i18n key is the right condition. Have you verified that this doesn't happen on other screens with similar steps-to-reproduce?

src/v2/models/AppState.ts
sherizada-okta: It's special: https://okta.slack.com/archives/CCA77QVL2/p1660695611393009.
ftr, more info on this PR for future devs: https://okta.slack.com/archives/CAZH2MWEL/p1660675971972929

PR #2716, opened 2022-08-17
OKTA-523413 : Initial implementation for phone number empty field

glenfannin-okta:
## Description:
The purpose of this PR is to implement validation for phone number fields (these fields are special because the country code and possibly extension are appended, and at the time of validation, you do not have the context as to different parts of the string).

## PR Checklist

### Issue:

- [OKTA-523413](https://oktainc.atlassian.net/browse/OKTA-523413)

### Screenshot/Video:

https://user-images.githubusercontent.com/97472729/185013039-065ea07e-5974-4d04-a41a-69057f86e463.mov


src/v3/src/transformer/dataSchema/addPhoneValidation.ts
```diff
@@ -0,0 +1,75 @@
...
+    const endIndex = code.length + 1;
+    return phoneNumber.substring(1, endIndex) === code;
+  });
+  if (!countryCode) {
+    return true;
+  }
+  const [phoneWithCountryCode] = phoneNumber.split('x');
+  const phoneNumberOnly = phoneWithCountryCode.substring(countryCode.length + 1);
+  return !!phoneNumberOnly;
+};
```
glenfannin-okta, line 50: @lesterchoi-okta I know we discussed possibly just using the number of characters to determine if the validation message should be thrown. However, after trying that, I noticed for countries with a short country code, the `empty field` error message would display even though the field was not empty. 

So I reverted back to this approach (assuming that invalid error would not be acceptable). 

Wanted to get your thoughts on either approach.

shuowu: possible to track country code from user interaction (component), then use it in validate fn?

lesterchoi-okta (approved):

PR #2727, opened 2022-08-18
OKTA-523460 : Initial implementation to fix security question enroll flow

glenfannin-okta:
## Description:
This PR fixes various issues with the Security question enroll flow: 
1) When you select the `create your own security question` option, trigger an error (incorrect number of characters) and submit with a new value for the answer, go back to Predefined question, trigger an error, then correct the error and submit the new answer. It sends `question` in the payload, which results in a bad request error. 

2) When you select `create your own security question` option, trigger an error (incorrect number of characters) and submit again with correct number of characters, the `questionKey` is set to the predefined default value (which results in an invalid request and bad user experience because they submitted a custom question/answer but the payload is default predefined with an answer that doesn't match).

3) When submitting a request that results in an error, the previous value is in the field, but when you click submit again, a client side error is thrown the the field is missing values.

## PR Checklist

### Issue:

- [OKTA-523460](https://oktainc.atlassian.net/browse/OKTA-523460)


src/v3/test/integration/authenticator-enroll-security-question.test.tsx
```diff
@@ -103,6 +106,266 @@ describe('authenticator-enroll-security-question', () => {
...
+            credentials: {
+              questionKey: 'disliked_food',
+              answer,
+            },
+            stateHandle: 'fake-stateHandle',
+          }),
+          headers: {
+            Accept: 'application/json; okta-version=1.0.0',
+            'Content-Type': 'application/json',
+            'X-Okta-User-Agent-Extended': 'okta-auth-js/9.9.9',
```
lesterchoi-okta, line 138: what's this 9.9.9?

src/v3/src/types/schema.ts
```diff
@@ -134,12 +135,11 @@ export function isUISchemaLayoutType(type: string): boolean {
 
 export interface FieldElement extends UISchemaElement {
   type: 'Field';
-  // TODO: only use limited field as i18n field
```
lesterchoi-okta, line 137: Still needed?

src/v3/test/integration/authenticator-enroll-security-question.test.tsx
```diff
@@ -103,6 +106,266 @@ describe('authenticator-enroll-security-question', () => {
...
+      const answerEle = await findByTestId('credentials.answer') as HTMLInputElement;
+
+      const answer = 'pi';
+      await user.type(answerEle, answer);
+      expect(answerEle.value).toEqual(answer);
+
+      await user.click(submitButton);
+      expect(authClient.options.httpRequestClient).toHaveBeenCalledWith(
+        'POST',
+        'https://oie-4695462.oktapreview.com/idp/idx/challenge/answer',
```
lesterchoi-okta, line 126: `http://localhost:3000` hostname

src/v3/test/integration/authenticator-enroll-security-question.test.tsx
```diff
@@ -103,6 +106,266 @@ describe('authenticator-enroll-security-question', () => {
...
+    it('should show field level character count error message on multiple attempts to submit with invalid character count', async () => {
+      const {
+        user, authClient, findByText, findByTestId,
+      } = await setup({ mockResponse: predefinedQuestionAnswerErrorResponse });
+
+      await findByText(/Set up security question/);
+      const submitButton = await findByText('Verify', { selector: 'button' });
+      const answerEle = await findByTestId('credentials.answer') as HTMLInputElement;
+
+      const answer = 'pi';
```
lesterchoi-okta, line 158: This testing pattern smells a little iffy. We're setting the mock response to return an error regardless of the input, i.e., even if this test set input to "superlonganswer" the mock response would've been the same error... We should re-assess how we test these validations. Maybe skip the input part of the test altogether since it's misleading

src/v3/test/integration/authenticator-enroll-security-question.test.tsx
```diff
@@ -103,6 +106,266 @@ describe('authenticator-enroll-security-question', () => {
...
+            stateHandle: 'fake-stateHandle',
+          }),
+          headers: {
+            Accept: 'application/json; okta-version=1.0.0',
+            'Content-Type': 'application/json',
+            'X-Okta-User-Agent-Extended': 'okta-auth-js/9.9.9',
+          },
+          withCredentials: true,
+        },
+      );
```
lesterchoi-okta, line 203: Let's add a tech-debt/todo to extract/DRY this boilerplate to a utility function.

src/v3/test/integration/__snapshots__/authenticator-verification-email.test.tsx.snap
```diff
@@ -620,7 +620,6 @@ exports[`authenticator-verification-email renders correct form renders the otp c
                           id="credentials.passcode"
                           name="credentials.passcode"
                           type="text"
-                          value=""
```
lesterchoi-okta, line 623: Is this expected?

src/v3/src/components/Form/Layout.tsx
```diff
@@ -45,7 +51,7 @@ const Layout: FunctionComponent<LayoutProps> = ({ uischema, stepName }) => {
     >
       {
         elements.map((element, index) => {
-          const elementKey = [element.type, stepName, index].join('_');
+          const elementKey = getElementKey(element, index);
```
lesterchoi-okta, line 54: I notice `stepName` was removed from `LayoutProps`. Is it not useful as part of the `key`?

src/v3/src/mocks/scenario/securityquestion-enroll-mfa.ts
```diff
@@ -56,12 +56,28 @@ scenario('securityquestion-enroll-mfa', (rest) => ([
...
+    let response = null;
+    let responseStatus;
+    if (answer?.length < 4) {
+      if (questionKey === 'custom') {
+        response = (await import('../response/idp/idx/challenge/answer/enroll-security-question-custom-with-error.json')).default;
+      } else {
+        response = (await import('../response/idp/idx/challenge/answer/enroll-security-question-predefined-with-error.json')).default;
+      }
+      // responseStatus = 401;
+      responseStatus = 200; // error from auth-js causes breakage when setting 401 status on this flow
```
lesterchoi-okta, line 73: Ah, I found out a little bit more about why this is a 200 instead of 401. Will DM you with link to relevant Slack thread

src/v3/src/transformer/layout/securityQuestion/securityQuestionEnroll.ts
```diff
@@ -120,43 +124,61 @@ export const transformSecurityQuestionEnroll: IdxStepTransformer = ({ transactio
...
+      dataSchemaRef.current!.submit = predefinedSubmitButton.options;
+      dataSchemaRef.current!.fieldsToValidate = ['credentials.answer'];
+
+      // reset default selected question
+      setData((prev) => ({
+        ...prev,
+        [QUESTION_KEY_INPUT_NAME]: preSelectedQuestion,
+      }));
+
+      // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
lesterchoi-okta, line 141: Oof, looks like these dependencies are pretty tangled. No easy way to define things ahead of time to avoid this I'm guessing?

src/v3/src/transformer/uischema/transform.ts
```diff
@@ -0,0 +1,35 @@
...
+      // eslint-disable-next-line no-param-reassign
+      element.key = element.key ? `${name}_${element.key}` : name;
+    },
+  });
+  return formbag;
+};
+
+export const transformUISchema: TransformStepFnWithOptions = (options) => (formbag) => flow(
+  addKeyToElement(options),
+)(formbag);
```
lesterchoi-okta, line 35: Do you need to use `flow` here if it's only one function?

lesterchoi-okta (approved):
Approving to unblock, but seeing this makes me feel there are inherent weaknesses in the types for the transformer layer and some bad patterns in the integration testing. Let's re-assess after getting through the remaining bugs.

src/v3/test/integration/authenticator-enroll-security-question.test.tsx
glenfannin-okta: This was from the newly created mock file, I will change this to match the other mock files.

src/v3/src/types/schema.ts
glenfannin-okta: Nope this isn't really needed anymore, we have the `translations` filed in `UISchemaElement` so that solves the purpose of this TODO.

src/v3/src/components/Form/Layout.tsx
glenfannin-okta: Ah, `stepName` is still used, but I had to move where it was being populated. When it was set as a prop to `Layout.tsx` it was causing a weird re-rendering issue when the new transaction gets populated. This issue impacted the WebAuthN flow causing it to display the browser prompt after already signing in. 

I created a new transformer that will append the stepName to the key before it hits layout. This way, if/when we make updates in the future (i.e. if we want to generate some unique GUID per transaction) we can easily append it to each element.

src/v3/src/transformer/layout/securityQuestion/securityQuestionEnroll.ts
glenfannin-okta: Actually I was able to figure out a way, will push up with my next commit.

src/v3/src/transformer/uischema/transform.ts
glenfannin-okta: I don't really, I only did it this way to match the other transformers but this is overkill, we can add flow if we create more like this in this file.

src/v3/test/integration/__snapshots__/authenticator-verification-email.test.tsx.snap
glenfannin-okta: Ah yes, what was happening before this PR was when you would enter an invalid OTP code, it would reload the view but would clear the input field (even though visibly, the text was still left in the field). This was invalid and did not match parity (also was the root of our issues with the "field cannot be blank" error. Now since we reload the view with the same text (when there is an error) the field is not artificially cleared allowing the user to amend the original text/value.

src/v3/test/integration/authenticator-enroll-security-question.test.tsx
glenfannin-okta: Actually, sorry, this value is used in every integration test, it is set this way when the mock authClient is created. You can see it is explicitly set here: https://github.com/okta/okta-signin-widget/blob/0.0/src/v3/test/integration/util/createAuthClient.ts#L153-L156

src/v3/test/integration/authenticator-enroll-security-question.test.tsx
glenfannin-okta: I ended up creating a mockRequest client, performing a check on the answer length (ensuring it is >= 4 characters), if it is not, we return the error response, otherwise, we return an empty response. This will make the test make more sense since it will not fail if the number of characters are >= 4

shuowu-okta: if it's server side validation, then the test can be setup with the error response directly, as long as the client can render correctly with the response it should be in good state, and it should not really care how the error response comes.

glenfannin-okta: Yep it is a server side validation, and that is exactly what I did, the test was a little weird because initially I wasn't using the mock request client, and just always responding with the character limit error for this test. To make it make more sense to a third party, I think using the mock request client and explicitly checking the num of characters in the answer payload will help (at least that is my hope :D ).

shuowu-okta: I mean the error part can sit in its own test, so one response renders one form, and we test it with one test suite. in this way, we can avoid overlaps cross tests

glenfannin-okta: That is how this is set up, this specific test is toggling between Stepper forms (because of an issue that was found and fixed to prevent future regression). (This specific test is dedicated to just one error response and one form)

shuowu-okta: The client does not need to mock any server side response. So the first one tests if request can be made, as long as no client side validation logic, it should end with hitting the mocked http client. Any other steps are making assumptions, and it may introduce inconsistency with server side state.

The the second test verifies if client can render correctly with the error response, and if buttons in the form can perform request/s correctly.

PR #2737, opened 2022-08-23
make useAutoFocus hook generic

lesterchoi-okta:
Make useAutoFocus hook generic, return immutable Ref


lesterchoi-okta (commented):
@leemchale-okta see https://dev.to/wojciechmatuszewski/mutable-and-immutable-useref-semantics-with-react-typescript-30c9

leemchale-okta (approved):

PR #2744, opened 2022-08-24
OKTA-527165 : Link missclick fix

leemchale-okta:
## Description:

This PR fixes the issue of miss clicking links due to field errors pushing the link away from the mouse cursor.

## PR Checklist

### Issue:

- [OKTA-527165](https://oktainc.atlassian.net/browse/OKTA-527165)


glenfannin-okta:
If we are going with the `mousedown` approach, there is another that needs it `TextWithHtml` component.

lesterchoi-okta (approved):
Is there any reasonable way to unit test this?

leemchale-okta:
not sure if you could unit test this since jest is simulating clicks programmatically?  @glenfannin-okta what do you think?  Maybe we can just check make sure the onBlur event does not fire when mouseDown on a link?

src/v3/src/components/TextWithHtml/TextWithHtml.tsx
```diff
@@ -32,6 +33,7 @@ const TextWithHtml: UISchemaElementComponent<{
   } = uischema.options;
   const onSubmitHandler = useOnSubmit();
 
+  const onMouseDown: ClickHandler = (e) => { e.preventDefault(); console.log('mouse down'); };
```
glenfannin-okta, line 36:
```suggestion
  const onMouseDown: ClickHandler = (e) => { e.preventDefault(); };
```

glenfannin-okta:
> not sure if you could unit test this since jest is simulating clicks programmatically? @glenfannin-okta what do you think? Maybe we can just check make sure the onBlur event does not fire when mouseDown on a link?

I think the practical test for this would be an integration test that lands on a page with the first field auto-focused, then immediately clicking a link (i.e. forgot password or sign up) that should pass without an issue. 

I dont think we have a specific test set up like this (because most of our integration tests loads the target page first - so it doesnt have a need to navigate by clicking a specific link)

glenfannin-okta (approved):

PR #2745, opened 2022-08-24
OKTA-527227 : Implementation to fix back to sign in link on account unlock success page

glenfannin-okta:
## Description:

This PR fixes the back to sign in link on account unlock page as well as all Terminal pages that contains the link. The `cancel` link is an action and has to be treated as such when sending the request to auth-js, otherwise we will receive an error and the navigation will not happen. 

## PR Checklist

### Issue:

- [OKTA-527227](https://oktainc.atlassian.net/browse/OKTA-527227)

### DIFF
--- a/src/v3/src/transformer/terminal/transformTerminalTransaction.ts
+++ b/src/v3/src/transformer/terminal/transformTerminalTransaction.ts
@@ -42,6 +42,7 @@ import {
   loc,
   removeUsernameCookie,
   setUsernameCookie,
+  shouldShowCancelLink,
 } from '../../util';
 import { redirectTransformer } from '../redirect';
 import { createForm } from '../utils';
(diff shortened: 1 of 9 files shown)


src/v3/src/util/shouldShowCancelLink.ts
```diff
@@ -0,0 +1,19 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { OktaWidgetFeatures } from '../types';
+
+export const shouldShowCancelLink = (features?: OktaWidgetFeatures): boolean => {
+  const { hideSignOutLinkInMFA, mfaOnlyFlow } = features || {};
+
+  return !(hideSignOutLinkInMFA || mfaOnlyFlow);
+};
```
glenfannin-okta, line 19: I realized we need this check otherwise we will display the cancel link on pages/flows that should not have them. See: https://github.com/okta/okta-signin-widget/blob/0.0/src/v2/view-builder/internals/BaseFooter.js#L59-L61

src/v3/src/transformer/terminal/transformTerminalTransaction.ts
```diff
@@ -77,13 +78,15 @@ const appendViewLinks = (
...
-  const { baseUrl } = widgetProps;
+  const { baseUrl, features } = widgetProps;
+  const isCancelAvailable = shouldShowCancelLink(features);
   const cancelStep = transaction?.availableSteps?.find(({ name }) => name === 'cancel');
   const cancelLink: LinkElement = {
     type: 'Link',
     options: {
       label: loc('goback', 'login'),
       step: 'cancel',
+      isActionStep: true,
```
glenfannin-okta, line 89: This is the cruz of the fix, since this was not being set, we were sending the request as a step instead of as an action, and auth-js failed when clicking.

lesterchoi-okta (approved):
Only suggestion/thought would be to calculate this once on init and set it on the `features` or `options` object with some convention to indicate it's a **derived** value, maybe `__shouldShowCancelLink` or something like that. But not necessary for this PR, just a tech debt item if we find ourselves deriving more values from features

PR #2752, opened 2022-08-29
OKTA-521284 - Updating "Back to sign in" link logic on Terminal pages to match v2

glenfannin-okta:
## Description:

The purpose of this PR is to fix the "Back to sign in" link on various terminal pages.

## PR Checklist

### Issue:

- [OKTA-521284](https://oktainc.atlassian.net/browse/OKTA-521284)


src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -51,6 +53,15 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
       stepToRender,
     } = options;
 
+    let { stateHandle } = currTransaction!.context;
+
+    if (step === 'cancel') {
+      authClient?.transactionManager.clear({ clearIdxResponse: false });
```
glenfannin-okta, line 59: I added this line to match what v2 is currently doing when it encounters a `cancel` action: https://github.com/okta/okta-signin-widget/blob/master/src/v2/controllers/FormController.ts#L165-L177

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -51,6 +53,15 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
...
+      if (previousTransaction) {
+        stateHandle = previousTransaction.context.stateHandle;
+      }
```
glenfannin-okta, line 62: To explain why this fixes the issue: 

This is ultimately an auth-js issue, and we are adding in a temporary workaround to prevent broken links (this issue exists in v2 as well). 
[This](https://github.com/okta/okta-auth-js/blob/master/lib/idx/run.ts#L130) is the entrypoint in auth-js. If we follow the flow, you can see it eventually hits [this](https://github.com/okta/okta-auth-js/blob/master/lib/idx/introspect.ts#L30) line, which finally lands [here](https://github.com/okta/okta-auth-js/blob/master/lib/TransactionManager.ts#L340). This is problematic because it is expecting the stateHandle to match the stateHandle from the last request. However, we are sending the stateHandle from the current transaction (which was updated on the terminal response). When they do not match, this line returns null and causes `introspect` to be called instead of the `cancel` action. Calling introspect after a failure, just results in the same failure, causing this link to appear broken. The only way the user can escape this is by refreshing the page.

glenfannin-okta: Ticket opened to auth-js team: https://oktainc.atlassian.net/browse/OKTA-528448

shawyu-okta: Attach a short explanation and ticket to the top of this block for context and also to discourage anyone from adding code inside of this logic for any reason

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -91,12 +102,13 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     }
     payload = toNestedObject(payload);
     if (currTransaction!.context.stateHandle) {
-      payload.stateHandle = currTransaction!.context.stateHandle;
+      payload.stateHandle = stateHandle;
     }
```
shawyu-okta, line 106: If we're going to be able to remove the logic above, would it be easier to leave this untouched even if it's a bit more verbose? i.e. would be nice to be able to just remove the earlier code when the auth-js issue is resolved without having to undo this change.

shawyu-okta (approved):
small comments but overall lgtm

src/v3/src/hooks/useOnSubmit.ts
glenfannin-okta: Totally, it would require me to move that if block below this if block (to overwrite the stateHandle that is set in the payload). But as you mentioned, once that code can be removed, it can easily be deleted without modifying this line as well. 

Is that what you meant?

shawyu-okta: Oh I forgot that you'd have to move the block down. If it doesn't affect the functionality nor cause any additional complication in the code then I think that would be desirable. If it makes things difficult then it's not a big deal at all to leave things as-is.

src/v3/src/components/Widget/index.tsx
```diff
@@ -82,6 +82,7 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
   });
   const [message, setMessage] = useState<IdxMessage | undefined>();
   const [idxTransaction, setIdxTransaction] = useState<IdxTransaction | undefined>();
+  const [previousTransaction, setPreviousTransaction] = useState<IdxTransaction | undefined>();
```
shuowu-okta, line 85: you don't need this state, see line 88

glenfannin-okta: Oh yea! Great catch, I actually tried to use this initially, I think it may have something to do with when it gets updated but this previous transaction did not have the correct stateHandle. The stateHandle in this object matched that from the current transaction, which is why I created a new state object. 

When I saw that I didnt debug more into why, mostly because I didnt want to make any changes that would negatively impact the code that is using this ref.

glenfannin-okta: Here is what I was talking about: 
image

glenfannin-okta: Ref is the existing ref that was created, current transaction is the current transaction, and prev is the new state object I created. 

My fear was changing something with that prev transaction ref and causing breakage for whatever was using it. Especially if this logic is going to go away after a bug fix, I didn't think it would be worth it to put it a bunch of effort into that.

shuowu-okta: the last one in your log (probably what you tracked in state), does not look like a stateHandle, but more like a token instead, what you get if you decode it?

glenfannin-okta: This is what I get when I decode it, it is directly from the `stateHandle` property that is in the transaction context. 

It is 100% the difference of which transaction is being used. The ref uses the last successful transaction whereas this state object uses the "currentTransaction". Maybe the name is confusing. I think I should just create a stateProp called `savedStateHandle` or `stateHandle` or something like that to avoid confusion. 

image

glenfannin-okta: You can see here, this is the transaction we are using the stateHandle from for this workaround: https://github.com/okta/okta-signin-widget/pull/2752/files#diff-114b8aded3d172ffc719a9452df5b38cfe8f9e82362d922bc0ca4f7f4440005aR111

glenfannin-okta: Where as the ref uses the last successful transaction. So maybe the best idea is to just save the stateHandle as a state prop and not the entire transaction.

shuowu-okta: ohh, this one is for the terminal state issue, authjs clears transaction related data, I don't think a workaround can be applied in widget side. I was suggesting https://github.com/okta/okta-signin-widget/blob/21b98876539d6c48a7108a44de337027085f41d7/src/v2/view-builder/utils/LinksUtil.js#L121

glenfannin-okta: Ahh, so the above solution does work, see the integration test, (albeit not the greatest). I took at look at that `getBackToSignInLink` but ultimately didnt do it because in this scenario, the cancel action exists on the transaction. How will we know when to use it? Or are you saying to always use this logic regardless if `cancel` action exists or not?

shuowu-okta: As I said before, it's how v2 works, and authjs currently has issue for this scenario. So I would suggest to follow v2 instead of hacking another solution.

glenfannin-okta: Yep, I recall :D but this is how v2 handles it today: https://github.com/okta/okta-signin-widget/blob/21b98876539d6c48a7108a44de337027085f41d7/src/v2/view-builder/views/TerminalView.js#L158-L164

This only happens if the cancel object doesnt exist, in this scenario, the cancel action is present. So are you saying to always use this logic even if the cancel action is present?

glenfannin-okta: I am not against doing this, I just want to make sure this doesnt cause any other issues by not using the cancel action when it is expected to be used.

shuowu: v3 should work if v2 works, which does not send cancel request in the current version (based on what i read from code)

glenfannin-okta: V2 is broken in this scenario (and other scenarios like this) as well (same exact issue, it calls introspect instead of cancel) 😅

glenfannin-okta: The only fix would be to always use this `getBackToSignInLink` logic in v3 until a fix can be applied to auth-js. Or what I have above with the stateHandle. 

Which method is preferred?

src/v3/src/util/settingsUtils.ts
```diff
@@ -73,3 +73,25 @@ export const getLanguageCode = (widgetProps: WidgetProps): LanguageCode => {
 
   return (supportedLanguages[supportedPos] ?? config.defaultLanguage) as LanguageCode;
 };
+
+export const getBackToSignInUri = (widgetProps: WidgetProps): string | undefined => {
+  const { backToSignInLink, signOutLink } = widgetProps;
+  return backToSignInLink || signOutLink;
+};
+
+export const getBaseUrl = (widgetProps: WidgetProps): string | undefined => {
```
glenfannin-okta, line 82: For reference: https://github.com/okta/okta-signin-widget/blob/21b98876539d6c48a7108a44de337027085f41d7/src/models/Settings.ts#L393-L404

src/v3/src/util/settingsUtils.ts
```diff
@@ -73,3 +73,25 @@ export const getLanguageCode = (widgetProps: WidgetProps): LanguageCode => {
 
   return (supportedLanguages[supportedPos] ?? config.defaultLanguage) as LanguageCode;
 };
+
+export const getBackToSignInUri = (widgetProps: WidgetProps): string | undefined => {
```
glenfannin-okta, line 77: For reference: https://github.com/okta/okta-signin-widget/blob/21b98876539d6c48a7108a44de337027085f41d7/src/models/Settings.ts#L182-L186

src/v3/test/integration/error-authenticator-enrollment-not-allowed.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+describe('error-authenticator-enrollment-not-allowed', () => {
+  it('should render form', async () => {
+    const { container, findByText } = await setup({ mockResponse });
+    await findByText(/Authenticator enrollment is not allowed at this time./);
+    expect(container).toMatchSnapshot();
+  });
+
+  // TODO: OKTA-528448 : Re-enable once this auth-js bug is fixed
+  it.skip('should send correct payload when clicking back to sign in link', async () => {
```
glenfannin-okta, line 25: Skipping this test until auth-js fix can be implemented.

src/v3/src/transformer/terminal/transformTerminalTransaction.ts
```diff
@@ -237,6 +241,10 @@ export const transformTerminalTransaction = (
 
   const { messages } = transaction;
 
+  if (containsMessageKey(TERMINAL_KEY.SESSION_EXPIRED, messages)) {
+    authClient?.transactionManager.clear();
+  }
```
glenfannin-okta, line 246: For reference: https://github.com/okta/okta-signin-widget/blob/21b98876539d6c48a7108a44de337027085f41d7/src/v2/controllers/FormController.ts#L367-L370

src/v3/src/types/schema.ts
```diff
@@ -277,6 +277,7 @@ export interface LinkElement extends UISchemaElement {
     label: string;
     href?: string;
     dataSe?: string;
+    onClick?: () => Promise<void>;
```
shawyu-okta, line 280: is the return type necessary? i wonder if it could be `() => unknown`

src/v3/src/components/Widget/index.tsx
```diff
@@ -100,6 +100,37 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
...
+    return null;
+  };
+
+  const bootstrap = useCallback(async () => {
+    try {
+      const transaction = await authClient.idx.start({
+        stateHandle: stateToken,
+      });
+
+      setAuthApiError(undefined);
```
shawyu-okta, line 122: can we use `null` instead?

shawyu-okta (approved):

PR #2760, opened 2022-08-31
OKTA-529389 : Preserve user provided data during error scenarios

glenfannin-okta:
## Description:

Purpose of this PR is to preserve user entered data when there are error scenarios. This prevents the widget from prompting for user data when the field already has information in the input fields (during error scenarios).

I did not add an automation test here because we already had one that covered this exact scenario from another ticket, see: https://github.com/okta/okta-signin-widget/blob/0.0/src/v3/test/integration/identify-with-password-error-flow.test.tsx#L52

This change covers all scenarios instead of just the username/identifier scenario.

## PR Checklist

### Issue:

- [OKTA-529389](https://oktainc.atlassian.net/browse/OKTA-529389)

### DIFF
--- a/src/v3/src/transformer/identify/transformIdentify.ts
+++ b/src/v3/src/transformer/identify/transformIdentify.ts
@@ -25,7 +25,6 @@ export const transformIdentify: IdxStepTransformer = ({
   formBag,
   widgetProps,
   transaction,
-  isClientTransaction,
 }) => {
   const { features, username } = widgetProps;
   const { uischema, data } = formBag;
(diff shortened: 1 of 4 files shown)


src/v3/src/transformer/identify/transformIdentify.ts
```diff
@@ -34,7 +33,7 @@ export const transformIdentify: IdxStepTransformer = ({
     'identifier',
     uischema.elements as UISchemaElement[],
   ) as FieldElement;
-  if (identifierElement && !isClientTransaction) {
```
shawyu-okta, line 37: what is the significance of removing this check?

glenfannin-okta: Originally, this change made it so that we did not set the identifier in the `data` prop of FormBag. What happens is that this allows the `setData((prev) => { ...prev, ...formBag.data })` to work because `identifier` never existed in the data object in the first place. 

With the change I made, this condition is no longer necessary, because now, even if data contains identifier and is set by default in the transformer (i.e. in the case of the cookie) during error transactions, it will be overwritten by what was previously in the data bag (before the error happened, basically whatever the user entered before the error).

shawyu-okta (approved):

PR #2762, opened 2022-09-01
OKTA-527004 : Error link focus fix 

leemchale-okta:
## Description:

This PR fixes the issue of the `Back to sign in` link being not in focus when the session expired error appears.

## PR Checklist

### Issue:

- [OKTA-527004](https://oktainc.atlassian.net/browse/OKTA-527004)

### DIFF
--- a/src/v3/src/transformer/layout/email/emailChallengeTransformer.ts
+++ b/src/v3/src/transformer/layout/email/emailChallengeTransformer.ts
@@ -53,6 +53,9 @@ export const transformEmailChallenge: IdxStepTransformer = ({ transaction, formB
     'credentials.passcode',
     uischema.elements as UISchemaElement[],
   );
+  if (passcodeElement) {
+    passcodeElement.focus = true;
+  }
 
   const redactedEmailAddress = nextStep.relatesTo?.value?.profile?.email;
   const maginLinkText = redactedEmailAddress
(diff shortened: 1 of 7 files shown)


src/v3/src/transformer/layout/email/emailChallengeTransformer.ts
```diff
@@ -53,6 +53,9 @@ export const transformEmailChallenge: IdxStepTransformer = ({ transaction, formB
     'credentials.passcode',
     uischema.elements as UISchemaElement[],
   );
+  if (passcodeElement) {
+    passcodeElement.focus = true;
+  }
```
glenfannin-okta, line 58: Why was this needed? Shouldn't it be found as the first element in the transformer?

glenfannin-okta (commented):
Can you also add a test to [this](https://github.com/okta/okta-signin-widget/blob/0.0/src/v3/test/integration/error-session-expired.test.tsx#L18) spec and just asserting that the back to sign in link is focused.

src/v3/src/transformer/layout/email/emailChallengeTransformer.ts
leemchale-okta: this was needed because the Enter a code button flow and the Enter code input flow are combined in the same transformer in a stepper layout.  So only the Enter code button is focused because it is the first element.  So after you press the Enter a code button in order for the Enter code input to be focused, I had to manually add focus to that input component.

PR #2764, opened 2022-09-01
OKTA-516622 : Global error message implementation

glenfannin-okta:
## Description:

This PR is to implement the global error message to be displayed when there are transaction errors. Additionally, this makes the reminder prompt unmount when there are transaction errors (to prevent the global error and reminder prompt from being displayed at the same time).

## PR Checklist

### Issue:

- [OKTA-516622](https://oktainc.atlassian.net/browse/OKTA-516622)

### Screenshot/Video:

https://user-images.githubusercontent.com/97472729/188007637-df5c14da-0b22-472d-8254-841c48783b2d.mov

### DIFF
--- a/src/v3/src/components/Widget/index.tsx
+++ b/src/v3/src/components/Widget/index.tsx
@@ -224,11 +224,16 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
      * i.e. unlocking an account and immeidiately being challenged with a different authenticator
      * when that is the case, we need to check the authenticator key to
      * determine if they are the same (as you shouldn't be challenged with the same authenticator)
+     * But if for some reason the keys are the same between them, we perform a last ditch check
+     * against the current authenticator's ID, which should always be unique between challenges
     */
     const pollingTxAuthKey = getAuthenticatorKey(pollingTransaction);
     const currentTxAuthKey = getAuthenticatorKey(idxTransaction);
+    const pollingTxAuthId = pollingTransaction.context.currentAuthenticator?.value?.id;
+    const currentTxAuthId = idxTransaction.context.currentAuthenticator?.value?.id;
     if (pollingTransaction.nextStep?.name !== idxTransaction.nextStep?.name
-        || pollingTxAuthKey !== currentTxAuthKey) {
+        || pollingTxAuthKey !== currentTxAuthKey
+        || pollingTxAuthId !== currentTxAuthId) {
       setIdxTransaction(pollingTransaction);
     }
   // eslint-disable-next-line react-hooks/exhaustive-deps
--- a/src/v3/src/components/Widget/index.tsx
+++ b/src/v3/src/components/Widget/index.tsx
@@ -20,7 +20,6 @@ import {
   IdxStatus,
   IdxTransaction,
 } from '@okta/okta-auth-js';
-import isEqual from 'lodash/isEqual';
 import { FunctionComponent, h } from 'preact';
 import {
   useCallback,
@@ -223,19 +222,12 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     /**
      * polling transaction name can sometimes be the same as the current transaction's name
      * i.e. unlocking an account and immeidiately being challenged with a different authenticator
-     * when that is the case, we need to compare inputs to determine if it is the same transaction
-     * lastly, inputs could also be the same (i.e. email otp to password) in that case we finally
-     * check the authenticator key to determine if they are the same
-     * (as you shouldnt be challenged with the same authenticator)
+     * when that is the case, we need to check the authenticator key to
+     * determine if they are the same (as you shouldn't be challenged with the same authenticator)
     */
-    const pollingTxAndCurrentTxInputsDiffer = !isEqual(
-      pollingTransaction.nextStep?.inputs,
-      idxTransaction.nextStep?.inputs,
-    );
     const pollingTxAuthKey = getAuthenticatorKey(pollingTransaction);
     const currentTxAuthKey = getAuthenticatorKey(idxTransaction);
     if (pollingTransaction.nextStep?.name !== idxTransaction.nextStep?.name
-        || pollingTxAndCurrentTxInputsDiffer
         || pollingTxAuthKey !== currentTxAuthKey) {
       setIdxTransaction(pollingTransaction);
     }
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -10,13 +10,14 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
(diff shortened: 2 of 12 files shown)


src/v3/src/components/Widget/index.tsx
```diff
@@ -223,19 +222,12 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
...
+     * determine if they are the same (as you shouldn't be challenged with the same authenticator)
     */
-    const pollingTxAndCurrentTxInputsDiffer = !isEqual(
-      pollingTransaction.nextStep?.inputs,
-      idxTransaction.nextStep?.inputs,
-    );
     const pollingTxAuthKey = getAuthenticatorKey(pollingTransaction);
     const currentTxAuthKey = getAuthenticatorKey(idxTransaction);
     if (pollingTransaction.nextStep?.name !== idxTransaction.nextStep?.name
-        || pollingTxAndCurrentTxInputsDiffer
```
glenfannin-okta, line 238: I had to remove this because when there are field level errors, the `isEqual` function always returns false (even when it should be true. In this specific scenario, there isnt a need to check the inputs since we are also checking authenticator keys. So I removed this to check the name and authenticator keys only.

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -100,8 +101,21 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     try {
       const newTransaction = await fn(payload);
       setIdxTransaction(newTransaction);
-      const isClientTransaction = newTransaction.nextStep?.name === currTransaction?.nextStep?.name;
+      const isClientTransaction = !(newTransaction.requestDidSucceed ?? true);
       setIsClientTransaction(isClientTransaction);
+      if (isClientTransaction
+          && !newTransaction.messages?.length
```
glenfannin-okta, line 107: Messages from the transaction are added to the ui schema during the normal transformation flow, so here we need to avoid displaying the generic global error message and the transaction errors.

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -100,8 +101,21 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     try {
       const newTransaction = await fn(payload);
       setIdxTransaction(newTransaction);
-      const isClientTransaction = newTransaction.nextStep?.name === currTransaction?.nextStep?.name;
+      const isClientTransaction = !(newTransaction.requestDidSucceed ?? true);
```
glenfannin-okta, line 104: This needed to change because we cant assume the current transaction and the new transaction will only have the same name when there are errors. Looking at v2, they use this requestDidSucceed field for determining when to show the errors.

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -100,8 +101,21 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     try {
       const newTransaction = await fn(payload);
       setIdxTransaction(newTransaction);
-      const isClientTransaction = newTransaction.nextStep?.name === currTransaction?.nextStep?.name;
```
glenfannin-okta, line 103: This needed to change because we cant assume the current transaction and the new transaction will only have the same name when there are errors. Looking at v2, they use this requestDidSucceed field for determining when to show the errors.

src/v3/src/mocks/scenario/securityquestion-enroll-mfa.ts
```diff
@@ -61,13 +61,15 @@ scenario('securityquestion-enroll-mfa', (rest) => ([
...
       response = (await import('../response/idp/idx/challenge/answer/enroll-security-question-with-character-limit-error.json')).default;
+      responseStatus = 400;
     } else {
       response = (await import('../response/oauth2/default/v1/token/default.json')).default;
     }
     return res(
-      ctx.status(200),
+      ctx.status(responseStatus),
       ctx.json(response),
     );
```
shawyu-okta, line 74: suggest that we move the `return`s to inside the `if/else` so the `responseStatus` variable is no longer needed.

src/v3/src/util/generateUUID.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+// eslint-disable-next-line max-len
+export const generateUUID = (): string => Date.now().toString(36)
+  + Math.random().toString(36).substring(2);
```
shawyu-okta, line 15: is a UUID explicitly necessary? If so, then we should use a real UUID generator library. If we only need a random string, then we should rename this.

shawyu-okta (commented):
Does the reminder re-appear after the usual timeout when there is already an error message showing? (in other words, the reverse order of events as the video shows)

glenfannin-okta:
> Does the reminder re-appear after the usual timeout when there is already an error message showing? (in other words, the reverse order of events as the video shows)

Yes it does actually, so once I wait for the prompt to be displayed, I enter an invalid code, it removes the reminder prompt and displays the global error message. However, if I sit on this page for another 30 seconds, the reminder prompt is displayed again, but the global error is not removed. 

When I compare this flow to v2, the reminder prompt is not displayed again after the initial error (I am assuming this is a bug in v2 because it also stops polling - which should not happen). 

So there could be a potential for both to be displayed at the same time if they but I feel like this may be warranted? What do you think? If we absolutely do not want both to be displayed under any circumstance, I would just need to add the set function for the global message to the reminder prompt component and clear the message when the reminder prompt is displayed.

shawyu-okta:
> So there could be a potential for both to be displayed at the same time, but I feel like this may be warranted? 

I agree both _should_ be displayed at the same time under some circumstances so this is the right behavior to me

src/v3/test/integration/authenticator-verification-email.test.tsx
```diff
@@ -124,13 +125,20 @@ describe('authenticator-verification-email', () => {
...
       // render invalid otp message
       const codeEle = await findByTestId('credentials.passcode') as HTMLInputElement;
       const submitButton = await findByText('Verify', { selector: 'button' });
       const verificationCode = '123456';
       await user.type(codeEle, verificationCode);
       await user.click(submitButton);
       await findByText('Invalid code. Try again.');
+      // After an error, verify that the Reminder prompt is removed in lieu of the global error
+      expect(queryByText(/Haven't received an email?/)).toBeNull();
+      await findByText(/We found some errors./);
```
shawyu-okta, line 141: Can you also update this test file to test that the reminder prompt re-appears again given enough time after the error?

shawyu-okta (approved):

PR #2768, opened 2022-09-07
OKTA-525579 : Fixing content overflow issue

glenfannin-okta:
## Description:

Purpose of this PR is to fix the content overflow issue that is caused by text that has no breaks which can cause the content to overflow beyond the widget's container. 

## PR Checklist

### Issue:

- [OKTA-525579](https://oktainc.atlassian.net/browse/OKTA-525579)

### Screenshot/Video:
Before: 
image

After: 
image

### DIFF
--- a/src/v3/src/components/Form/Form.tsx
+++ b/src/v3/src/components/Form/Form.tsx
@@ -78,7 +78,7 @@ const Form: FunctionComponent<{
       onSubmit={handleSubmit}
       className="o-form" // FIXME update page objects using .o-form selectors
       data-se="form"
-      style={{ maxWidth: '100%', overflowWrap: 'anywhere' }}
+      style={{ maxWidth: '100%', wordBreak: 'break-word' }}
     >
       <Layout uischema={uischema} />
     </form>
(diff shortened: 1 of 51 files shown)


src/v3/src/components/Form/Form.tsx
```diff
@@ -78,7 +78,7 @@ const Form: FunctionComponent<{
       onSubmit={handleSubmit}
       className="o-form" // FIXME update page objects using .o-form selectors
       data-se="form"
-      style={{ maxWidth: '100%' }}
+      style={{ maxWidth: '100%', wordBreak: 'break-word' }}
```
shawyu-okta, line 81: Can you look into these MDN docs? It looks like `word-break: break-word` is deprecated but there's another relevant property called `overflow-wrap`:
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

glenfannin-okta: 🤦 I did not know this was deprecated (TIL), looking at the docs overflow-wrap is definitely the property to use, I am now going to question every prop that I have used before to make sure it is still viable 😅 

TY!

shawyu-okta (approved):

PR #2776, opened 2022-09-12
OKTA-516592 : Phone enrollment one option fix

leemchale-okta:
## Description:

This PR fixes the issue where Phone enrollment shows a radio button when there is only one option.

## PR Checklist

### Issue:

- [OKTA-516592](https://oktainc.atlassian.net/browse/OKTA-516592)

### Screenshot/Video:
image


src/v3/src/transformer/phone/phoneEnrollmentTransformer.ts
```diff
@@ -76,37 +77,62 @@ export const transformPhoneEnrollment: IdxStepTransformer = ({ formBag, transact
...
+      ],
+    },
+    voice: {
+      type: UISchemaLayoutType.VERTICAL,
+      elements: [
+        voiceInfoTextElement,
+        phoneNumberElement,
+        voiceStepSubmitButton,
+      ],
+    },
```
glenfannin-okta, line 96: `uischema.elements` looks to already be a UISchemaLayout, it seems like you can just append these elements directly to uischema.elements instead of adding a new layout.

src/v3/src/transformer/phone/phoneEnrollmentTransformer.ts
```diff
@@ -76,37 +77,62 @@ export const transformPhoneEnrollment: IdxStepTransformer = ({ formBag, transact
...
 
   uischema.elements = [
     titleElement,
-    stepper,
   ];
 
+  const phoneMethodOptions = methodTypeElement!.options!.inputMeta.options!;
+  if (phoneMethodOptions.length === 1) {
+    const methodType = phoneMethodOptions[0].value.toString();
+    uischema.elements.push(phoneOptionLayouts[methodType]);
```
glenfannin-okta, line 129: In addition to this, you should also append the `actionParams` to the applicable button, setting the `authenticator.methodType` as the first methodOption value.

src/v3/src/transformer/phone/phoneEnrollmentTransformer.ts
```diff
@@ -76,37 +77,62 @@ export const transformPhoneEnrollment: IdxStepTransformer = ({ formBag, transact
...
+  } else {
+    uischema.elements.push(phoneOptionLayouts.both);
```
glenfannin-okta, line 131: In this condition, you can move line 42 here:
`data['authenticator.methodType'] = methodTypeElement!.options!.inputMeta.options![0].value;`

glenfannin-okta (commented):
Does an integration test already exist where we test the payload with 1 option vs 2?

src/v3/src/transformer/phone/phoneEnrollmentTransformer.ts
leemchale-okta: @glenfannin-okta do the `actionParams` need to be set in the case where both phone options are present?

glenfannin-okta: Nope, when both are present no action params should be added (since its controlled by the radio button selection)

src/v3/src/transformer/phone/phoneEnrollmentTransformer.ts
leemchale-okta: It looks like this line needs to before outside before the condition, or else it wont show the extension input in the case where voice is the only option

glenfannin-okta: Ohhh that is right, the phone authenticator needs that value for the extension 

Ok, I'm sorry, scratch that and the button action params I discussed above. I think we should be safe by just setting the value directly in data (I dont think it will be overwritten).

src/v3/src/transformer/phone/phoneEnrollmentTransformer.ts
```diff
@@ -76,37 +77,62 @@ export const transformPhoneEnrollment: IdxStepTransformer = ({ formBag, transact
...
+  }
+
   // set default dataSchema
-  dataSchema.submit = smsStepSubmitButton.options;
+  dataSchema.submit = phoneMethodOptions[0].value === 'sms' ? smsStepSubmitButton.options : voiceStepSubmitButton.options;
```
glenfannin-okta, line 135: Since we wont add action params, you can revert this (since the options to both buttons are the same)

src/v3/src/transformer/phone/phoneEnrollmentTransformer.ts
leemchale-okta: okay got it no prob!

glenfannin-okta (approved):

src/v3/src/transformer/phone/phoneEnrollmentTransformer.ts
```diff
@@ -102,9 +114,20 @@ export const transformPhoneEnrollment: IdxStepTransformer = ({ formBag, transact
...
+  const firstOptionMethod = phoneMethodOptions[0].value;
+  if (phoneMethodOptions.length === 1) {
+    if (firstOptionMethod === 'sms') {
+      uischema.elements = uischema.elements.concat(smsOptionElements);
+    } else {
+      uischema.elements = uischema.elements.concat(voiceOptionElements);
+    }
+  } else {
+    uischema.elements.push(bothOptionStepper);
+  }
```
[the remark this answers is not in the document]
shawyu-okta: To take this even further, optionally we could inline the array of elements (i.e. smsOptionElements and voiceOptionElements) inside the conditionals rather than defining them in the upper scope.

src/v3/test/integration/__snapshots__/authenticator-email-verification-data.test.tsx.snap
```diff
@@ -134,7 +134,7 @@ exports[`authenticator-email-verification-data should render form 1`] = `
                   class="MuiBox-root emotion-11"
                 >
                   <button
-                    class="MuiButton-root MuiButton-primary MuiButton-primaryPrimary MuiButton-sizeMedium MuiButton-primarySizeMedium MuiButton-disableElevation MuiButton-fullWidth MuiButtonBase-root  emotion-18"
+                    class="MuiButtonBase-root MuiButton-root MuiButton-primary MuiButton-primaryPrimary MuiButton-sizeMedium MuiButton-primarySizeMedium MuiButton-disableElevation MuiButton-fullWidth emotion-18"
```
shawyu-okta, line 137: Curious why this reordered itself

shawyu-okta (approved):

PR #2779, opened 2022-09-13
OktaSignIn a11y plugin

lesterchoi-okta:
## Description:

Plugin to improve the accessibility (a11y) of the Okta sign in widget to comply with WCAG2.1AA (https://www.w3.org/TR/WCAG21/) standards and Section 508 (https://www.section508.gov/).

## PR Checklist

### Issue:

- [OKTA-524494](https://oktainc.atlassian.net/browse/OKTA-524494)

### DIFF
--- a/playground/main.ts
+++ b/playground/main.ts
@@ -3,7 +3,6 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const signinWidgetOptions = require('../.widgetrc.js'); // commonJS module
 
-// import OktaSignIn from '../src/widget/OktaSignIn';
 import {
   OktaSignInConstructor,
   OktaSignInAPI,
@@ -12,10 +11,14 @@ import {
   RenderResultSuccessNonOIDCSession,
 } from '../src/types';
 import { assertNoEnglishLeaks } from '../playground/LocaleUtils';
+
 declare global {
   interface Window {
     // added by widget CDN bundle
     OktaSignIn: OktaSignInConstructor;
+    
+    // from <script src="/js/okta-plugin-a11y.js">
+    OktaPluginA11y: (widget: OktaSignInAPI, options: Record<string, unknown>) => void;
 
     // added in this file
     getWidgetInstance: () => OktaSignInAPI;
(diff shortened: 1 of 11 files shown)


playground/main.ts
```diff
@@ -3,7 +3,8 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const signinWidgetOptions = require('../.widgetrc.js'); // commonJS module
 
-// import OktaSignIn from '../src/widget/OktaSignIn';
+import { OktaA11yPlugin } from '../src/plugins/OktaA11yPlugin';
```
shuowu-okta, line 6: Can playground access this plugin via global OktaPlugins instead of the src dir? maybe add another script tag in index.html

webpack.release.config.js
```diff
@@ -53,4 +55,16 @@ var devConfig = config('okta-sign-in.js', 'development');
 devConfig.plugins = plugins({ isProduction: false, analyzerFile: 'okta-sign-in.analyzer' });
 usePolyfill(devConfig);
 
-module.exports = [entryConfig, noPolyfillConfig, cdnConfig, noJqueryConfig, devConfig];
+// 6. plugins: a11y
+var pluginConfig = config('okta-plugins.js', 'production');
```
shuowu-okta, line 59: how about name it as `okta-plugins.beta.js` so we have the freedom to make changes?

shuowu-okta (approved):

playground/main.ts
lesterchoi-okta: Good idea, I'll give it a try

webpack.release.config.js
lesterchoi-okta: Is there another way we can make clear it's experimental? Maybe a big disclaimer in the header? Changing the URL later will mean updating usage

shuowu-okta: I feel it's ok to leave it in the filename/url, devs also need to make changes anyway if they are using a npm package with beta tag.
what the main use case for this plugin file? if it's mostly for monolith, then we still have the full control of the url.

lesterchoi-okta: I primarily had the monolith scenario in mind during implementation, but AFAICT there's no reason it couldn't be used outside.

PR #2798, opened 2022-09-20
OKTA-516607 : Widget vertical position height fix

leemchale-okta:
## Description:

This PR fixes the issue of the Sign in widget appearing too low on the page.

## PR Checklist

### Issue:

- [OKTA-516607](https://oktainc.atlassian.net/browse/OKTA-516607)

### Screenshot/Video:
image

### DIFF
--- a/src/v3/src/components/AuthContainer/AuthContainer.tsx
+++ b/src/v3/src/components/AuthContainer/AuthContainer.tsx
@@ -11,11 +11,13 @@
  */
 
 import { Box, useMediaQuery } from '@mui/material';
+import classNames from 'classnames';
 import { FunctionComponent, h } from 'preact';
 
 import style from './style.css';
 
 const AuthContainer: FunctionComponent = ({ children }) => {
+  const classes = classNames('auth-container', 'main-container', style.mainViewContainer);
   const isMobileWidth = useMediaQuery('screen and (max-width: 391px)');
   return (
     <Box
@@ -24,7 +26,7 @@ const AuthContainer: FunctionComponent = ({ children }) => {
       display="flex"
       justifyContent="center"
       alignItems="center"
-      className={style.mainViewContainer}
+      className={classes}
       data-version={VERSION}
       data-commit={COMMITHASH}
     >
--- a/src/v3/src/components/AuthContainer/AuthContainer.tsx
+++ b/src/v3/src/components/AuthContainer/AuthContainer.tsx
@@ -11,13 +11,11 @@
  */
 
 import { Box, useMediaQuery } from '@mui/material';
-import classNames from 'classnames';
 import { FunctionComponent, h } from 'preact';
 
 import style from './style.css';
 
 const AuthContainer: FunctionComponent = ({ children }) => {
-  const classes = classNames('auth-container', 'main-container', style.mainViewContainer);
   const isMobileWidth = useMediaQuery('screen and (max-width: 391px)');
   return (
     <Box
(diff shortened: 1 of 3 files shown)


src/v3/src/components/AuthContainer/AuthContainer.tsx
```diff
@@ -11,13 +11,11 @@
  */
 
 import { Box, useMediaQuery } from '@mui/material';
-import classNames from 'classnames';
 import { FunctionComponent, h } from 'preact';
 
 import style from './style.css';
 
 const AuthContainer: FunctionComponent = ({ children }) => {
-  const classes = classNames('auth-container', 'main-container', style.mainViewContainer);
```
glenfannin-okta, line 20: I recall this was originally added to help preserve some of the classes that exists in v2 (i.e. if an org is hooking onto the `main-container` class name, to perform custom css changes, this was done to prevent breaking their customizations). Also, I can't recall if the existing v2 tests require one of these classes, if none of these are a concern any longer then I think it is ok to remove, otherwise, we may need to keep it.

Was removing these additional classes needed to make this work?

leemchale-okta: Actually it was not necessary to remove them to make it work, it looked like the widget container was not inheriting any styles from those classes and my thinking was we were going to extract them anyway.  But on second thought I agree I think they still may be needed and I will add them back in.

glenfannin-okta (approved):

src/v3/src/components/Widget/index.tsx
```diff
@@ -286,7 +286,8 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     >
       {/* Note that we need two theme providers until we fully migrate to odyssey-mui */}
       <MuiThemeProvider theme={mapMuiThemeFromBrand(brandColors)}>
-        <ScopedCssBaseline>
+        {/* the style is to allow the widget to inherit the parent's bg color */}
+        <ScopedCssBaseline sx={{ backgroundColor: 'inherit' }}>
```
shawyu-okta, line 290: just curious, this can't be done in the css of `AuthContainer`?

glenfannin-okta: When I tried it on the auth container, the background color of the body was still not being inherited to the widget (I initially tried to set it there).

shawyu-okta: Curious if this behavior is what we want -- Is there a discussion about this somewhere? Or an example screenshot of how this behaves?

glenfannin-okta: So it was something I noticed when discussing the height issue with Lee. If you look at the current widget, they set the bg color in Monolith in the containing div 

image


If we mimic this with V3, you will notice the background color does not get applied because the MuiScopedCssBaseline div set the bg color to white. 
image

image

Not sure if the beta users will customize a background color or not but ultimately this is something I think we will need to support because if the widget is embedded in a company's website/page (custom hosted scenarios), the widget will set the containing bg color to white when it should inherit whatever div it is placed in). IIRC, we should not control colors/styling outside of the container, whereas this MuiScopedCssBaseline div is by setting the bg to white.

leemchale-okta: I believe this is in parity with V2 @glenfannin-okta is that correct?

shawyu-okta: Oh I see, thank you for the explanation - I was thinking this was affecting the inner container of the widget. Agreed that the larger surrounding container should not set its own bgcolor and instead should inherit the surrounding container

shawyu-okta (approved):

PR #2800, opened 2022-09-21
OKTA-533535 : Disable submit button on loading

leemchale-okta:
## Description:

Implements the feature of disabling the submit button while loading.  This fixes the issue of the user being able to send multiple api requests by repeatedly clicking on the submit button.

## PR Checklist

### Issue:

- [OKTA-533535](https://oktainc.atlassian.net/browse/OKTA-533535)

### Screenshot/Video:
image


src/v3/src/components/Button/Button.tsx
```diff
@@ -73,6 +77,8 @@ const Button: UISchemaElementComponent<{
       // eslint-disable-next-line react/jsx-props-no-spreading
       {...(ariaLabel && { 'aria-label': ariaLabel } )}
     >
+      {loading && <CircularProgress />}
+      {loading && ' '}
```
glenfannin-okta, line 81: I think you can use MUI's loading button component instead, see: https://mui.com/material-ui/react-button/#loading-button

leemchale-okta: The loading button is in the @mui/lab package which looks like an experimental package I think?  Do we want to add that package?

glenfannin-okta: Oh I didn't notice that, yeah since this is in the lab and since the lab can introduce breaking changes, let's not use that (unless someone else feels differently). 

Would it be possible to update the css for the progress indicator that is added to the button? I was thinking making it aligned similar to MUI's loading button, or if that is too much, can the button text be removed and just display the spinner? 

Additionally, if it is simple enough, can the color of the spinner in the button be changed to match MUI's loading button? 

This is just so it looks a little better visually. 

image

glenfannin-okta:
Oh yea, I think this should point to 0.0 branch since this was a P1.

src/v3/src/components/Button/Button.tsx
shawyu-okta, line 81: Can these be combined into a single conditional?

shawyu-okta (commented):
Is the screenshot how it currently looks when the loading indicator is showing? It seems slightly misaligned.

It also seems like there's a bunch of other changes in this PR, probably due to retargeting the 0.0 branch - maybe a rebase is needed?

leemchale-okta:
screenshot has been updated

leemchale-okta:
when I try and rebase with 0.0 it says it is already up to date

shawyu-okta:
@leemchale-okta I think you'll need to rebase interactively and remove the first three commits that were picked up from the 0.1 target

shawyu-okta:
> screenshot has been updated

This looks much better -- I just realized that this seems to realign the text off-center when the spinner appears, but the Figma shows the text remaining in the same centered position when the spinner appears. Is there any way to solve this? If not, check with @alexdahl-okta to see if this slightly different treatment is acceptable. I'm concerned about the amount of movement in the button between loading/non-loading states.

leemchale-okta:
added a margin to the right of the text to compensate the offset from the circle progress component

alexdahl-okta:
Sweet! ✨

Yeah, ideally there'd be minimal movement of the button's label.
Adding a margin to compensate feels promising!

I expect Odyssey to implement an opinionated button loading state eventually, so don't want to invest _too_ much in custom treatment.

If we need to consider alternatives down the road, we probably have a variety of options... a couple ideas for posterity:

- transition the spinner dimensions up from 0, so that the label's movement is smoother
- place the spinner at the leading or trailing edge of the button, with the label always centered

src/v3/test/integration/identify-with-password.test.tsx
```diff
@@ -327,4 +327,23 @@ describe('identify-with-password', () => {
...
+    const usernameEl = await findByTestId('identifier') as HTMLInputElement;
+    const passwordEl = await findByTestId('credentials.passcode') as HTMLInputElement;
+    const submitButton = await findByText('Sign in', { selector: 'button' });
+
+    await user.type(usernameEl, 'testuser@okta.com');
+    expect(usernameEl.value).toEqual('testuser@okta.com');
+    await user.type(passwordEl, 'fake-password');
+    expect(passwordEl.value).toEqual('fake-password');
+    await user.click(submitButton);
+    await user.click(submitButton);
```
shawyu-okta, line 345: I think there might be an actual double click event. If there is, we should use that

shawyu-okta (approved):
Just one last comment about clicking in the test.

I noticed we're only disabling the button. Was there a concrete outcome to the discussion regarding if we should totally disable the whole form or not?

PR #2806, opened 2022-09-23
OKTA-534967 : Authenticator button description fix

glenfannin-okta:
## Description:

The purpose of this PR is to fix an issue with the description of Authenticator buttons when there are different authenticators (sharing the same key but an alternate description is required). 

i.e. When someone sets up a phone authenticator with 2 different phone numbers, previously the logic just retrieved the description for the first found phone number entry instead of it being distinct based on each individual authenticator.

## PR Checklist

### Issue:

- [OKTA-534967](https://oktainc.atlassian.net/browse/OKTA-534967)

### Screenshot/Video:
image

### DIFF
--- a/src/v3/src/transformer/selectAuthenticator/utils.ts
+++ b/src/v3/src/transformer/selectAuthenticator/utils.ts
@@ -21,13 +21,6 @@ import {
 import { ActionParams, AuthenticatorButtonElement, ButtonType } from '../../types';
 import { loc } from '../../util';
 
-const getAuthenticatorOption = (
-  options: IdxOption[],
-  authenticatorKey: string,
-): IdxOption | undefined => options?.find(
-  ({ relatesTo }) => relatesTo?.key === authenticatorKey,
-);
-
 export const getOptionValue = (
   inputs: Input[],
   key: string,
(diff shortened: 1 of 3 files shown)


src/v3/src/transformer/selectAuthenticator/utils.ts
```diff
@@ -214,7 +190,7 @@ const formatAuthenticatorOptions = (
           ? loc('oie.enroll.authenticator.button.text', 'login')
           : loc('oie.verify.authenticator.button.text', 'login'),
         description: getAuthenticatorDescription(
-          options,
```
shawyu-okta, line 217: Nice and simple solution -- Was there any reason we were passing the whole array of options before? It seems pretty obvious now that we should only care about the single option rather than get the whole array, so wondering if there was some sort of utility or plans for this method to do more.

shawyu-okta (approved):

src/v3/src/transformer/selectAuthenticator/utils.ts
glenfannin-okta: I can't think of a legitimate reason for passing the entire options array (because that just complicates things and forces you to search the array for the corresponding option, which is already known since the caller is iterating over options). 

The only thing I can think of is that with OV, you need to modify the options array (for totp and push method types) but that is handled in a completely separate isolated function that doesn't overlap with this one, so maybe they were combined at one point then refactored 🤔 (its the only logical explanation of why this was written this way lol)

src/v3/test/integration/__snapshots__/authenticator-verification-select-authenticator.test.tsx.snap
```diff
@@ -593,7 +593,7 @@ exports[`authenticator-verification-select-authenticator renders form 1`] = `
                         class="MuiTypography-root MuiTypography-body1 MuiTypography-paragraph emotion-54"
                         data-se="authenticator-button-description"
                       >
-                        +1 XXX-XXX-5309
+                        +1 XXX-XXX-5310
```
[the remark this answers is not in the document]
glenfannin-okta: This integration test only asserts on the snapshot (which lists about 13 authenticator options), but I think adding the unit test is a little more valuable with this sort of thing. This scenario wasn't covered in the unit test and without really inspecting the mock file and understanding how the description text is added to the UI it is easy to miss (which is exactly what happened in the scenario). There were already multiple phone numbers/authenticators in this mock file and the UI displayed the same phone number for both.

shawyu-okta (approved):

PR #2813, opened 2022-09-28
OKTA-534909 : Disable form Elements when request in progress

glenfannin-okta:
## Description:

The purpose of this PR is to implement the UI changes to disable form elements when an API request is in progress. 

To match figma designs, I had to switch from using react-odyssey Select component to the MUI select component. 

## PR Checklist

### Issue:

- [OKTA-534909](https://oktainc.atlassian.net/browse/OKTA-534909)

Screen Shot 2022-09-28 at 11 50 08 AM
Screen Shot 2022-09-28 at 11 50 18 AM
Screen Shot 2022-09-28 at 11 53 00 AM

### DIFF
--- a/src/v3/test/integration/identify-with-password-error-flow.test.tsx
+++ b/src/v3/test/integration/identify-with-password-error-flow.test.tsx
@@ -91,7 +91,7 @@ describe('identify-with-password-error-flow', () => {
     // TODO: OKTA-526754 - Update this assertion once merged back into 0.1
     expect(authClient.options.httpRequestClient).toHaveBeenCalledWith(
       'POST',
-      'https://oie-8425965.oktapreview.com/idp/idx/identify',
+      'http://localhost:3000/idp/idx/identify',
       {
         data: JSON.stringify({
           identifier: username,


src/v3/src/components/Button/Button.tsx
```diff
@@ -87,7 +87,12 @@ const Button: UISchemaElementComponent<{
             // compensate the offset from the CircularProgress component
             marginRight="23px"
           >
-            <CircularProgress sx={{ color: 'white' }} />
+            <CircularProgress
+              // TODO: OKTA-518793 - replace english string with key once created
+              aria-label="Loading..."
+              aria-valuetext="Loading..."
```
glenfannin-okta, line 93: This is needed to prevent A11y failure.

glenfannin-okta:
cc: @alexdahl-okta

src/v3/src/transformer/layout/securityQuestion/securityQuestionEnroll.ts
```diff
@@ -214,7 +214,7 @@ export const transformSecurityQuestionEnroll: IdxStepTransformer = ({ transactio
 
   // update default dataSchema
   dataSchema.submit = predefinedSubmitButton.options;
-  dataSchema.fieldsToValidate = ['credentials.answer'];
+  dataSchema.fieldsToValidate = ['credentials.answer', QUESTION_KEY_INPUT_NAME];
```
glenfannin-okta, line 217: Since the predefined question dropdown has an "empty" option and since it is a required field, we have to add it to the array of fields that need client side validation.

shawyu-okta (commented):
Would this work if we set the `disabled` value on the `form` at the top level instead of setting it on each input?

glenfannin-okta:
> Would this work if we set the disabled value on the form at the top level instead of setting it on each input?

I did try this but it did not work (I believe this prop disables the form from submitting to the `action` but there isnt much documentation on this prop w.r.t the form element).

image

Additionally, Lee tried to implement the `disabled` fieldset (wrapping all elements) (see: https://stackoverflow.com/questions/3507958/how-can-i-make-an-entire-html-form-readonly), however, it didnt apply the appropriate styling to each element which would have required him to still make changes at each component/element anyway :(

src/v3/src/transformer/layout/securityQuestion/securityQuestionEnroll.ts
shawyu-okta: This is so the user has to pick something other than the "empty" option?

src/v3/src/components/Select/Select.tsx
```diff
@@ -10,68 +10,102 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { Box } from '@mui/material';
-import { NativeSelect } from '@okta/odyssey-react';
```
shawyu-okta, line 14: nice, one more component migrated! 🙌

src/v3/src/components/Button/Button.tsx
shawyu-okta, line 93: Was there a reason we can't put the new translations into the properties file now?

shawyu-okta (approved):

src/v3/src/transformer/layout/securityQuestion/securityQuestionEnroll.ts
glenfannin-okta: Yessir exactly! Previously, if they selected the empty option, it would allow submission (with the empty value) and fail (instead of preventing it with the client-side error).

src/v3/src/components/Button/Button.tsx
glenfannin-okta: Lee actually has that ticket and is currently working on it with Product and design. There is a process to get new translations added, mostly just need to agree on the Copy text and then it can be added to master, then the service should pick it up in the next release and generate the text in the other languages.

PR #2831, opened 2022-10-03
OKTA-518979 : terminal view TestCafe Parity fix

leemchale-okta:
## Description:

This PR fixes the terminal view spec test for testcafe parity.

## PR Checklist

### Issue:

- [OKTA-518979](https://oktainc.atlassian.net/browse/OKTA-518979)


src/v3/src/components/Link/Link.tsx
```diff
@@ -69,6 +69,8 @@ const Link: UISchemaElementComponent<{
         <LinkMui
           href={href}
           ref={focusRef}
+          // eslint-disable-next-line react/jsx-props-no-spreading
+          {...(dataSe && { 'data-se': dataSe } )}
```
lesterchoi-okta, line 73: What happens if you don't check whether dataSe is set? My guess is the prop would be ignored?

lesterchoi-okta (approved):
LGTM. This fixes the spec with no test changes?

test/testcafe/spec/TerminalView_spec.js
```diff
@@ -140,8 +141,11 @@ async function setup(t) {
   test
     .requestHooks(mock)(testTitle, async t => {
       const terminalViewPage = await setup(t);
-      await t.expect(await terminalViewPage.goBackLinkExists()).notOk();
```
leemchale-okta, line 143: I did have to make a small change here

src/v3/src/components/Link/Link.tsx
leemchale-okta: yes i believe react ignores attributes that are undefined

test/testcafe/spec/TerminalView_spec.js
leemchale-okta: for some reason i am failing parity-v3 test on bacon, but I am passing locally

leemchale-okta:
for some reason i am failing parity-v3 test on bacon, but I am passing locally

test/testcafe/spec/TerminalView_spec.js
glenfannin-okta: In v2, are these links the same (as far as the text they display)? is the only difference between the two the classes attached to the links?

src/v3/src/components/Link/Link.tsx
glenfannin-okta: I think if you dont check the truthy-ness of the variable, it will still set it as a prop on the element, but without a value

so it would look something like 

```
<a href="..." data-se>Blah</a>
```

Correct me if I am wrong though but I _thought_ that is what happens 🤔

shawyu-okta: https://codesandbox.io/s/sweet-shape-d15j7z?file=/src/App.js

test/testcafe/spec/TerminalView_spec.js
leemchale-okta: Yes the text they display are the same just the classes are different

Screen Shot 2022-10-05 at 11 15 32 AM
e5-1907-416f-ae43-c5f6200038cc.png)

src/v3/src/components/Link/Link.tsx
glenfannin-okta: Sure enough it ignores it 🤦 

Ty! So really we could clean all of these up and get rid of that condition (at a later time, but good to know! TY!)

test/testcafe/spec/TerminalView_spec.js
leemchale-okta: yeah turns out the text for those links are the same.  But the data-se's are different.
Screen Shot 2022-10-05 at 12 20 05 PM
Screen Shot 2022-10-05 at 11 34 41 AM

test/testcafe/spec/TerminalView_spec.js
```diff
@@ -140,8 +141,11 @@ async function setup(t) {
...
+
       await t.expect(await terminalViewPage.signoutLinkExists()).ok();
+      if(!process.env.OKTA_SIW_V3) {
```
shawyu-okta, line 146: Env variables in NodeJS are always strings, so this probably won't work how we're expecting. Can you try copying over the approach I have in my other PR https://github.com/okta/okta-signin-widget/pull/2835 using the `userVariables` in TestCafe config? Would ensure we don't have to remember to set the env variable when running outside of a script, too.

test/testcafe/spec/TerminalView_spec.js
```diff
@@ -61,10 +62,10 @@ const terminalUnlockAccountFailedPermissionsMock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/introspect')
   .respond(terminalUnlockAccountFailedPermissions);
 
-fixture('Terminal view');
+fixture('Terminal view').meta('v3', true);
 
 async function setup(t) {
-  const terminalPageObject = new TerminalPageObject(t);
+  const terminalPageObject = process.env.OKTA_SIW_V3 ? new TerminalPageObjectV3(t) : new TerminalPageObject(t);
```
shawyu-okta, line 68: Sorry this is a bit of a late comment - we can update separately so as not to block this PR since I only just worked out the new approach:

We are trying to put any ternaries that choose between v3 and v2 selectors as low and close to the selector itself as possible. This means putting a ternary in the PageObject itself, rather than switching between different PageObjects.

shawyu-okta (approved):

shawyu-okta (approved):

PR #2833, opened 2022-10-03
OKTA-537520 : Hide errors on submit

leemchale-okta:
## Description:

This PR hides api errors while the widget is loading to signal the user of widget activity.

## PR Checklist

### Issue:

- [OKTA-537520](https://oktainc.atlassian.net/browse/OKTA-537520)

### Screenshot/Video:

https://user-images.githubusercontent.com/107433508/193702535-557b7e83-b092-48ee-8927-2369f78d23ca.mov


glenfannin-okta (commented):
LGTM, curious to get @shawyu-okta thoughts

leemchale-okta:
> Did we explore the option of clearing errors out `onSubmit` of the form rather than simply hiding the alert box during loading state?

We did explore that option but it seems the only way was to use `setIdxTransaction` directly on the transaction to remove the errors which may not be safe to do?  It also causes the form to re-render which could cause problems

src/v3/src/components/InfoBox/InfoBox.tsx
```diff
@@ -32,7 +37,7 @@ const InfoBox: UISchemaElementComponent<{
     },
   } = uischema;
 
-  return (
+  return !loading ? (
```
shawyu-okta, line 40: nit: invert the ternary so you can write this without the `!`

shawyu-okta (approved):

PR #2835, opened 2022-10-04
Enable Identify spec Testcafe tests for v3

shawyu-okta:
## PR Checklist

### Issue:

- [OKTA-518979](https://oktainc.atlassian.net/browse/OKTA-518979)

### DIFF
--- a/playground/mocks/data/idp/idx/identify.json
+++ b/playground/mocks/data/idp/idx/identify.json
@@ -18,7 +18,8 @@
             "value":[
               {
                   "name":"identifier",
-                  "label":"Username"
+                  "label":"Username",
+                  "required": true
               },
               {
                "name":"rememberMe",
--- a/src/v3/src/OktaSignIn/index.ts
+++ b/src/v3/src/OktaSignIn/index.ts
@@ -69,9 +69,12 @@ export default class OktaSignIn {
     [key in OktaWidgetEventType]: OktaWidgetEventHandler
   } | Record<string, never>;
 
+  el: string | null;
+
   constructor(options: WidgetProps) {
     this.version = VERSION;
     this.options = options;
+    this.el = null;
 
     this.events = {};
 
@@ -139,6 +142,7 @@ export default class OktaSignIn {
     onError?: RenderErrorCallback,
   ): Promise<RenderResult> {
     const { el } = options;
+    this.el = el;
 
     return new Promise<RenderResult>((resolve, reject) => {
       const onSuccessWrapper = (data: JsonObject): void => {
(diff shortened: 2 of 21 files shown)


playground/mocks/data/idp/idx/identify.json
```diff
@@ -18,7 +18,8 @@
             "value":[
               {
                   "name":"identifier",
-                  "label":"Username"
+                  "label":"Username",
+                  "required": true
```
shawyu-okta, line 22: Pretty sure this was a bad mock since identifier is a required field to proceed. Not sure why it's like this, maybe this was the actual API response way in the past and v2 widget has hardcoded logic to say it's actually required. v3 totally trusts the API response, so we need to have the "required" field set as true otherwise it will allow the user to click Submit without this field filled.

src/v3/src/OktaSignIn/index.ts
```diff
@@ -214,8 +218,17 @@ export default class OktaSignIn {
...
+    const target = typeof this.el === 'string'
+      ? document.querySelector(this.el)
+      : this.el;
+
+    if (target) {
+      render(null, target);
+
+      this.el = null;
+    }
+  }
```
shawyu-okta, line 231: This is needed for the test `rerender` functionality to work correct since it seems preact won't remount the widget on top of itself.

test/testcafe/framework/page-objects/IdentityPageObject.js
```diff
@@ -110,14 +110,6 @@ export default class IdentityPageObject extends BasePageObject {
...
   }
 
-  /**
-   * @deprecated
-   * @see hasIdentifierErrorMessage
-   */
-  hasIdentifierError() {
-    return this.form.hasTextBoxError('identifier');
-  }
-
```
shawyu-okta, line 120: I went through and just finished deprecating this method. We have duplicate assertions with this deprecated version side-by-side with the replacement one.

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -18,20 +18,6 @@ export default class BaseFormObject {
...
-    // clear exists text
-    await this.t
-      .selectText(element)
-      .pressKey('delete');
-
-    // type new text
-    if (text) {
-      await this.t.typeText(element, text);
-    }
-  }
```
shawyu-okta, line 33: Wasn't removed, moved down a bit (also changed internally) in this file for better grouping.

glenfannin-okta (approved):
Overall lgtm, I love the semantic approach.

PR #2837, opened 2022-10-04
OKTA-537519 : Profile enrollment custom attribute implementation

glenfannin-okta:
## Description:
The purpose of this PR is to fix the Profile enrollment custom attribute transformation and profile update view. 

## PR Checklist

### Issue:

- [OKTA-537519](https://oktainc.atlassian.net/browse/OKTA-537519)

### Screenshot/Video:

After speaking to @alexdahl-okta he asked that we only show the asterisks on the profile enrollment page because it causes unnecessary noise on other pages (since those fields are always required. 

https://user-images.githubusercontent.com/97472729/194337717-518a9503-8bfd-4a4f-a28b-a8a485d85a3c.mov


src/v3/src/transformer/field/transform.ts
```diff
@@ -91,12 +91,18 @@ export const transformStepInputs = (
       const uischema = mapUiElement(input);
       acc.uischema.elements = [...acc.uischema.elements, uischema];
 
+      if (type === 'boolean' && required) {
+        acc.data[name] = Boolean(input.value);
+      }
```
glenfannin-okta, line 96: For required boolean inputs (checkboxes), we just set it as what is provided in value field, if there is nothing in value field, it will be false (unchecked) by default.

src/v3/src/transformer/i18n/util.ts
```diff
@@ -55,9 +55,9 @@ export const addLabelTranslationToFieldElement = (
     : `${stepName}.${fieldName}`;
   const i18nKey = getI18nKey(path);
   const params = getI18NParams(nextStep, authenticatorKey);
-  if (i18nKey) {
+  if (i18nKey || element.label) {
```
glenfannin-okta, line 58: Labels werent showing up because of this missing logic.

src/v3/src/transformer/uischema/transform.ts
```diff
@@ -55,9 +58,43 @@ export const setFocusOnFirstElement: TransformStepFn = (formbag) => {
...
+export const updateCustomFields: TransformStepFn = (formbag) => {
+  traverseLayout({
+    layout: formbag.uischema,
+    predicate: (el) => (el.type === 'Field'),
+    callback: (el) => {
+      const fieldElement = (el as FieldElement);
+      const { options: { inputMeta: { options } } } = fieldElement;
+      if (Array.isArray(options) && options[0]?.value) {
+        const [option] = options;
+        if (option.label === 'display') {
```
glenfannin-okta, line 70: See: https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/ui-schema/ion-string-handler.js#L50

src/v3/src/transformer/profile/transformEnrollProfile.ts
```diff
@@ -74,14 +85,26 @@ export const transformEnrollProfile: IdxStepTransformer = ({ transaction, formBa
     },
   };
 
+  if (isAlternateView) {
```
glenfannin-okta, line 88: see: https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/EnrollProfileView.js#L9-L23

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -50,7 +50,7 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
   return (
     <Box>
       <PasswordInput
-        label={label}
+        label={label + (required ? ' *' : '')}
```
glenfannin-okta, line 53: Since this is an ODY component, I had to add this asterisk, their component doesnt allow us to remove the label (to implement a custom label) and the ODS rule is that all fields are required by default, see: https://odyssey.okta.design/?path=/docs/mui-components-forms-text-field--password#optional

Not sure how else to work around this without ODS' assistance. 

cc: @lesterchoi-okta

src/v3/test/integration/enroll-profile-update.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  });
+
+  it('should send correct payload', async () => {
+    const {
+      authClient, user, findByTestId, findByText,
+    } = await setup({ mockResponse });
+
+    await findByText(/Sign in/);
+
+    const submitButton = await findByText('Submit', { selector: 'button' });
```
shawyu-okta, line 31: Can you try `findByRole('button', { name: 'Submit' });`? The order of the queries in the RTL docs is ordered from preferred to least preferred

shawyu-okta: Same above, you might be able to do `findByRole('heading', { level: 2 })` for the title text.

src/v3/src/transformer/field/transform.ts
```diff
@@ -91,12 +91,18 @@ export const transformStepInputs = (
...
+
       // add client validation for "required" field
       // do not validate immutable fields, they will always be added to payload programatically
       if (required && mutable !== false && type !== 'object') {
         acc.dataSchema[name] = {
           validate(data) {
-            const isValid = !!data[name];
+            // in the case of a required boolean input, just return true
+            // if the backend requires the checbox val to be true, error will be displayed
+            const isValid = typeof data[name] === 'boolean' ? true : !!data[name];
```
shawyu-okta, line 105: Is there any way to combine this logic with the next line using early return? e.g.
```js
// i'm thinking you might start with this but haven't thought through how to handle the right side of the ternary
if (typeof data[name] === 'boolean') return undefined;
```

src/v3/src/types/schema.ts
```diff
@@ -142,9 +142,13 @@ export function isUISchemaLayoutType(type: string): boolean {
 export interface FieldElement extends UISchemaElement {
   type: 'Field';
   key: string;
+  /**
+   * @description TODO: temporary solution for custom fields in profile enrollment
+   */
+  required?: boolean | undefined;
```
shawyu-okta, line 148: Which part is the temporary bit? And what is needed to remediate this?

src/v3/src/components/Select/Select.tsx
```diff
@@ -10,68 +10,105 @@
...
-            key="empty"
-          />].concat(
+          [
+            <option
+              aria-label="Select an Option"
+              value=""
+              key="empty"
+            >
+              {/* TODO: OKTA-518793 - need translation key for this string */}
+              Select an Option
```
shawyu-okta, line 86: Does the v2 widget not have this placeholder text at all? Also if we add this, make sure we address the `aria-label` part with translations.

src/v3/src/components/Checkbox/Checkbox.tsx
```diff
@@ -10,46 +10,76 @@
...
+        label={label as string}
+      />
+      <FormHelperText
+        id={`${name}-error`}
+        role="alert"
+        data-se={`${name}-error`}
+        error
+      >
+        {error}
+      </FormHelperText>
```
shawyu-okta, line 80: Does this need to be guarded by `error && (`?

src/v3/src/transformer/profile/transformEnrollProfile.ts
shawyu-okta: Can we try to file a dependency against the home team to fix this from the backend side?

src/v3/src/transformer/profile/transformEnrollProfile.ts
```diff
@@ -45,6 +54,13 @@ export const transformEnrollProfile: IdxStepTransformer = ({ transaction, formBa
         autocomplete: 'new-password',
       },
     };
+    // @ts-ignore expose type from auth-js
+    const passwordErrors = passwordElement.options.inputMeta.messages?.value;
+    if (passwordErrors?.length) {
+      const messages = updatePasswordRequirementsNotMetMessage(passwordErrors);
+      // @ts-ignore expose type from auth-js
+      passwordElement.options.inputMeta.messages.value = messages;
```
shawyu-okta, line 62: Is there a JIRA we can attach? If not, can we file one for the SDK team and add it here?

src/v3/src/transformer/field/transform.ts
glenfannin-okta: > I'm not totally clear why we need to set the value explicitly this way, can you explain a bit more or give an example of where this is necessary?

In the case of _required_ boolean (checkboxes), we need to set the value in the data bag as a boolean. If we don't and the user never checks (or checks and unchecks) the box, the value will be `undefined` in the data bag and once the submission is made, it will fail (since the boolean is required). The change here is to set the field with a default boolean value to prevent that error early on. 

> Why do we need to use the Boolean function here?

I used boolean because `input.value` is not always set, it can be undefined. So the idea was to set the value to what is available in the input by default, or if a value isnt set, make it `false` by default. 

Additionally, the type of `input.value` is `string` (even though it is truly a boolean when the type is a boolean) so this sort of gets around that.

src/v3/src/types/schema.ts
glenfannin-okta: Ah, so I added this `required` prop as a temp solution to denote the required vs optional props in the profile enrollment flow only. 

So just this `required` field (in this file) is the temp solution. In the transformer the function to update this prop is also apart of the temporary solution. 

I spoke with ODS on this, and the long term solution is after they wrap MUI components, they will provide a optional attribute that will display a label above the input (on the right side) similar to v2. 

Once we implement their wrapped components we can get rid of this and use the required prop directly apart of `inputMeta`

src/v3/src/components/Select/Select.tsx
glenfannin-okta: > Does the v2 widget not have this placeholder text at all? Also if we add this, make sure we address the aria-label part with translations.

The v2 widget has this text but it looks to be provided (by default) by courage and there is no translation key/string. I submitted a request to create the key and will update this and the aria-label in our i18n transformer (similar to others).

glenfannin-okta: > Also curious if option can be aria-labelledby an inner element rather than using explicit aria-label?

Do you mean something like: 
```suggestion
              <option
              aria-labelledby="emptyLabel"
              value=""
              key="empty"
            >
               <span id="emptyLabel">Select an Option</span>
            </option>
```

src/v3/src/transformer/profile/transformEnrollProfile.ts
glenfannin-okta: Done! Had a conversation with them last week about it. 

https://oktainc.atlassian.net/browse/OKTA-539666

https://okta.slack.com/archives/C9AJXB5GC/p1665149067212879

glenfannin-okta: Starting this discussion with the feature team: 

https://okta.slack.com/archives/C029NJQTTC5/p1665422645575769 

Will update with the ticket once I get a response from them.

glenfannin-okta: Spoke to the team, and created a ticket for them: 
https://oktainc.atlassian.net/browse/OKTA-540383

Discussion: 
https://okta.slack.com/archives/C029NJQTTC5/p1665422645575769

src/v3/src/transformer/uischema/transform.ts
```diff
@@ -55,9 +59,62 @@ export const setFocusOnFirstElement: TransformStepFn = (formbag) => {
...
+            fieldElement.options.inputMeta.options = undefined;
+          }
+        }
+      }
+    },
+  });
+  return formbag;
+};
+
+// TODO: temporary solution for custom fields in profile enrollment
```
shawyu-okta, line 95: is there a ticket that will allow us to remove this?

src/v3/src/components/Select/Select.tsx
shawyu-okta: Yeah exactly, just wondering first if labelledby/label is needed at all (i.e. is the `option` implicitly labelled by the child text?) and if it is needed then if `labelledby` is a good way to mark it rather than embedding the same text twice.

shawyu-okta: Yeah exactly, just wondering first if labelledby/label is needed at all (i.e. is the `option` implicitly labelled by the child text?) and if it is needed then if `labelledby` is a good way to mark it rather than embedding the same text twice.

src/v3/src/transformer/field/transform.ts
glenfannin-okta: Good point, in this scenario, if the type of the input is ever === `boolean` and the `value` is set, I have only ever seen it (the value) set as an actual boolean or undefined (either `true` or `false` or `undefined`). I have never seen a boolean type value set as a string, granted I have only seen the value set in one scenario (which is the OV push notification scenario). 

In the scenario of profile enrollment, I don't see a default value passed (and not sure if one can even be set from admin dashboard). 

Should I treat it as it it can be a string?

src/v3/src/components/PhoneAuthenticator/PhoneAuthenticator.tsx
```diff
@@ -136,27 +137,42 @@ const PhoneAuthenticator: UISchemaElementComponent<UISchemaElementComponentWithV
 
   const renderCountrySelect = () => (
     <Box marginBottom={4}>
-      <NativeSelect
+      <InputLabel
+        id="countryLabel"
+        required={required}
```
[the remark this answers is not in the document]
glenfannin-okta: Ah, [FormControlLabel](https://mui.com/material-ui/api/form-control-label/#main-content) doesn't offer the `required` prop, so I had to add the `FormControl` wrapper which does. This is also the recommended approach for checkboxes, Radios, and Switches. 

While you can use a FormControl on a select or any other component, since it isnt required and since the InputLabel accepts the `required`, I decided against adding that wrapper.

src/v3/src/transformer/uischema/transform.ts
glenfannin-okta: Will check with ODS team to get a jira for the wrapper work they are doing. 🙏

src/v3/src/components/PhoneAuthenticator/PhoneAuthenticator.tsx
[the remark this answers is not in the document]
shawyu-okta: Ok sounds good!

src/v3/src/transformer/field/transform.ts
shawyu-okta: Might be safer and avoid potential confusion in the future if we do an explicit check:
```suggestion
      if (type === 'boolean' && required) {
        acc.data[name] = input.value === true;
      }
```
I think this will work the same way as the `Boolean` way but also default to `false` if a string is passed as `input.value`

shawyu-okta (approved):
Address [this comment/question](https://github.com/okta/okta-signin-widget/pull/2837#discussion_r996012297) and a few minor tweaks remaining but overall lgtm

src/v3/src/transformer/field/transform.ts
glenfannin-okta: I had to make the change this way because the type defined in auth-js does not account for booleans (even though it can be boolean and will be in this scenario).

```
if (type === 'boolean' && required) {
   acc.data[name] = (input.value as unknown as boolean) === true;
}
```

shawyu-okta: Oh, does it not allow you to compare `unknown` with `boolean`? Seems like it would safe to allow this.

src/v3/src/components/Select/Select.tsx
glenfannin-okta: I was looking at MUI's example, they used the aria attribute, I realized now it was because they didnt have an actual label in the option tag (we also didnt before until I found this Select an Option default string in courage). But to answer the question, we do not need this 😓 

i removed it.

src/v3/src/transformer/field/transform.ts
glenfannin-okta: Oh yea, I can just make it `unknown`, I will update that.

PR #2853, opened 2022-10-11
Enable Success spec Testcafe tests for v3

shawyu-okta:
## PR Checklist

### Issue:

- [OKTA-518977](https://oktainc.atlassian.net/browse/OKTA-518977)

### DIFF
--- a/test/testcafe/framework/page-objects/BasePageObject.js
+++ b/test/testcafe/framework/page-objects/BasePageObject.js
@@ -9,7 +9,6 @@ const SWITCH_AUTHENTICATOR_LINK = '.auth-footer .js-switchAuthenticator';
 const ionMessagesSelector = '.ion-messages-container';
 const SUBTITLE_SELECTOR = '[data-se="o-form-explain"]';
 const FACTOR_PAGE_HELP_LINK = '[data-se="factorPageHelpLink"]';
-const FORM_SELECTOR = 'form';
 
 export default class BasePageObject {
   constructor(t) {
@@ -52,10 +51,6 @@ export default class BasePageObject {
       }, '');
     }
     await this.t.navigateTo(`http://localhost:3000${this.url}${qs}`);
-
-    if (queryParams?.render !== false) {
-      await this.t.expect(Selector(FORM_SELECTOR).exists).eql(true);
-    }
   }
 
   async preventRedirect(toUrls) {
--- a/test/testcafe/framework/page-objects/BasePageObject.js
+++ b/test/testcafe/framework/page-objects/BasePageObject.js
@@ -52,7 +52,10 @@ export default class BasePageObject {
       }, '');
     }
     await this.t.navigateTo(`http://localhost:3000${this.url}${qs}`);
-    await this.t.expect(Selector(FORM_SELECTOR).exists).eql(true);
+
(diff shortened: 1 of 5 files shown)


test/testcafe/spec/Success_spec.js
```diff
@@ -8,10 +8,17 @@ const mock = RequestMock()
...
   .respond(identify)
   .onRequestTo('http://localhost:3000/idp/idx/identify')
-  .respond(success);
+  .respond(success)
+  .onRequestTo(/^http:\/\/localhost:3000\/app\/UserHome.*/)
+  .respond(`
+        <h1 id="mock-user-dashboard-title">Mock User Dashboard</h1>
+        <h2>Query parameters</h2>
+        <a href="/">Back to Login</a>
+  `);
```
glenfannin-okta, line 17: Is this done to prevent the page redirection? Just wondering if this is necessary because another spec I worked on had a similar redirect (want to know if I should add this same response).

glenfannin-okta (approved):

test/testcafe/spec/Success_spec.js
shawyu-okta: This is actually to give the browser something to render upon redirection. The existing test relies on a Dyson mock but in v3 we don't have that server running so instead it's simpler to just inline it into the TestCafe mocks.

test/testcafe/framework/page-objects/BasePageObject.js
```diff
@@ -52,7 +52,10 @@ export default class BasePageObject {
       }, '');
     }
     await this.t.navigateTo(`http://localhost:3000${this.url}${qs}`);
-    await this.t.expect(Selector(FORM_SELECTOR).exists).eql(true);
+
+    if (queryParams?.render !== false) {
+      await this.t.expect(Selector(FORM_SELECTOR).exists).eql(true);
+    }
```
shawyu-okta, line 58: @leemchale-okta sometimes we actually don't want to render the widget for certain tests, so I'm adding a guard here to skip this assertion when that is the case

shawyu-okta: Checking explicitly for `!== false` since sometimes it's `undefined`

test/testcafe/spec/Success_spec.js
glenfannin-okta: Ah that makes sense. Ty!

test/testcafe/spec/TerminalView_spec.js
```diff
@@ -69,6 +71,8 @@ fixture('Terminal view').meta('v3', true);
 async function setup(t) {
   const terminalPageObject = userVariables.v3 ? new TerminalPageObjectV3(t) : new TerminalPageObject(t);
   await terminalPageObject.navigateToPage();
+  // ensure form has loaded
+  await t.expect(Selector('form').exists).eql(true);
```
shawyu-okta, line 75: @leemchale-okta FYI I moved this again to _just_ the TerminalView spec for now. We can re-evaluate if we should move it back up with better guards around it in the future if we see the form loading flakiness again in other specs.

leemchale-okta: okay sounds good

PR #2854, opened 2022-10-12
OKTA-518974 : Select Authenticator Verify spec parity test updates

glenfannin-okta:
## Description:

Select Authenticator for Verification & Enroll spec test updates for parity.  

## PR Checklist

### Issue:

- [OKTA-518974](https://oktainc.atlassian.net/browse/OKTA-518974)


test/testcafe/framework/page-objects/BasePageObject.js
```diff
@@ -1,11 +1,13 @@
...
+import { Selector, ClientFunction, userVariables } from 'testcafe';
+
+const BACK_TO_SIGNIN_LINK = '[data-se="cancel"]';
 
 const SIGNOUT_LINK = '.auth-footer .js-cancel';
 const GO_BACK_LINK = '.auth-footer .js-go-back';
 const SKIP_LINK = '.auth-footer .js-skip';
 const SKIP_SET_UP_LINK = '.auth-footer .js-skip-setup';
-const SWITCH_AUTHENTICATOR_LINK = '.auth-footer .js-switchAuthenticator';
+const SWITCH_AUTHENTICATOR_LINK = '[data-se="switchAuthenticator"]';
```
glenfannin-okta, line 10: Confirmed in LinkUtil.js (v2) that the switch authenticator links (in either enroll or verify) both include this data-se attribute.

test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
```diff
@@ -228,14 +213,12 @@ test.requestHooks(mockEnrollAuthenticatorPassword)('should navigate to password
   await t.expect(enrollPasswordPage.confirmPasswordFieldExists()).eql(true);
 });
 
-test.requestHooks(requestLogger, mockEnrollAuthenticatorPassword)('select password challenge page and hit switch authenticator and re-select password', async t => {
+test.meta('v3', false).requestHooks(requestLogger, mockEnrollAuthenticatorPassword)('select password challenge page and hit switch authenticator and re-select password', async t => {
```
glenfannin-okta, line 216: Had to disable this because when switching back to authenticator list makes an API request in v3 but does not in V2. Until we update that (which is apart of this ticket https://oktainc.atlassian.net/browse/OKTA-521014) we will need to disable this test for the parity execution.

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -298,22 +299,22 @@ test.requestHooks(mockChallengePassword)('should navigate to password challenge
 
   selectFactorPage.selectFactorByIndex(0);
   const challengeFactorPage = new ChallengeFactorPageObject(t);
-  await t.expect(challengeFactorPage.getPageTitle()).eql('Verify with your password');
+  await t.expect(challengeFactorPage.getFormTitle()).eql('Verify with your password');
 });
 
-test.requestHooks(requestLogger, mockChallengePassword)('select password challenge page and hit switch authenticator and re-select password', async t => {
+test.meta('v3', false).requestHooks(requestLogger, mockChallengePassword)('select password challenge page and hit switch authenticator and re-select password', async t => {
```
glenfannin-okta, line 305: Similar to Enroll, had to disable this because when switching back to authenticator list makes an API request in v3 but does not in V2. Until we update that (which is apart of this ticket https://oktainc.atlassian.net/browse/OKTA-521014) we will need to disable this test for the parity execution.

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -509,7 +510,7 @@ test.requestHooks(requestLogger, mockChallengeOVPush)('should navigate to okta v
   });
 });
 
-test.requestHooks(requestLogger, mockChallengeOVFastPass)('should navigate to okta verify fast pass page', async t => {
+test.meta('v3', false).requestHooks(requestLogger, mockChallengeOVFastPass)('should navigate to okta verify fast pass page', async t => {
```
glenfannin-okta, line 513: Fastpass not enabled in v3 yet.

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -535,7 +536,7 @@ test.requestHooks(requestLogger, mockChallengeOVFastPass)('should navigate to ok
   });
 });
 
-test.requestHooks(mockChallengeOnPremMFA)('should navigate to on prem mfa challenge page', async t => {
+test.meta('v3', false).requestHooks(mockChallengeOnPremMFA)('should navigate to on prem mfa challenge page', async t => {
```
glenfannin-okta, line 539: onPrem MFA not enabled in v3 yet.

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -544,7 +545,7 @@ test.requestHooks(mockChallengeOnPremMFA)('should navigate to on prem mfa challe
   await t.expect(challengeFactorPage.getPageTitle()).eql('Verify with Atko Custom On-prem');
 });
 
-test.requestHooks(mockChallengeRsa)('should navigate to RSA challenge page', async t => {
+test.meta('v3', false).requestHooks(mockChallengeRsa)('should navigate to RSA challenge page', async t => {
```
glenfannin-okta, line 548: RSA not enabled on v3 yet.

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -564,7 +565,7 @@ test.requestHooks(mockChallengeDuo)('should navigate to Duo challenge page', asy
   await t.expect(challengeFactorPage.getPageTitle()).eql('Verify with Duo Security');
 });
 
-test.requestHooks(mockChallengeCustomOTP)('should navigate to Custom OTP challenge page', async t => {
+test.meta('v3', false).requestHooks(mockChallengeCustomOTP)('should navigate to Custom OTP challenge page', async t => {
```
glenfannin-okta, line 568: Custom OTP not enabled on v3 yet.

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -574,7 +575,7 @@ test.requestHooks(mockChallengeCustomOTP)('should navigate to Custom OTP challen
   await t.expect(challengeFactorPage.getFormTitle()).eql('Verify with Atko Custom OTP Authenticator');
 });
 
-test.requestHooks(mockChallengePassword)('should show custom factor page link', async t => {
+test.meta('v3', false).requestHooks(mockChallengePassword)('should show custom factor page link', async t => {
```
glenfannin-okta, line 578: Custom factor not enabled on v3 yet.

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -590,7 +591,7 @@ test.requestHooks(mockChallengePassword)('should show custom factor page link',
   await t.expect(pageObject.getFactorPageHelpLink()).eql('https://acme.com/what-is-okta-autheticators');
 });
 
-test.requestHooks(mockSelectAuthenticatorForRecovery)('should not show custom factor page link', async t => {
+test.meta('v3', false).requestHooks(mockSelectAuthenticatorForRecovery)('should not show custom factor page link', async t => {
```
glenfannin-okta, line 594: Custom factor not enabled on v3 yet.

test/testcafe/framework/page-objects/SelectAuthenticatorPageObject.js
```diff
@@ -1,6 +1,16 @@
...
 import BasePageObject from './BasePageObject';
 
+const AUTHENTICATOR_BUTTON_SELECTOR = '[data-se="authenticator-button"]';
+const AUTHENTICATOR_BUTTON_LABEL_SELECTOR = '[data-se="authenticator-button-label"]';
+const AUTHENTICATOR_BUTTON_ICON_SELECTOR = '[data-se="authenticator-icon"] div';
+const AUTHENTICATOR_BUTTON_CTA_BUTTON_LABEL_SELECTOR = `${AUTHENTICATOR_BUTTON_SELECTOR} [data-se="cta-button-label"]`;
+const AUTHENTICATOR_BUTTON_CTA_BUTTON_SELECTOR = `${AUTHENTICATOR_BUTTON_SELECTOR} .cta-button`;
+const AUTHENTICATOR_BUTTON_DESCRIPTION_SELECTOR = '[data-se="authenticator-button-description"]';
+const AUTHENTICATOR_BUTTON_USAGE_TEXT_SELECTOR = '[data-se="authenticator-button-usage-text"]';
+const V3_CUSTOM_OTP_BUTTON_SELECTOR = `${AUTHENTICATOR_BUTTON_SELECTOR} [data-se="custom_otp"]`;
```
shawyu-okta, line 11: Can any of these be replaced with `getByRole('button', { name: 'button text' })`?

test/testcafe/framework/page-objects/BasePageObject.js
glenfannin-okta: Is it possible to use find by role when the name can be 1 of two possible options? 

In this case, the link text can either be `Return to authenticator list` or `Verify with something else` I can create two separate functions, but it will require me to update the test to use the correct function (since the text varies based on if you are enrolling or verifying).

test/testcafe/framework/page-objects/SelectAuthenticatorPageObject.js
glenfannin-okta: One issue I experienced with this is that, in v2, the authenticator button container is just a simple div (doesnt have a role of button) but in v3 we do have a role of button for the container.

What I thought about doing was using the `getByRole('button')` and using the index to assert proper order and elements within the button (since it contains more than just a label, i.e. logo, description, link, etc), but this would definitely require me to change the test. 

For example, here is one example of how the test asserts a specific authenticator:

```
await t.expect(selectFactorPage.getFactorLabelByIndex(0)).eql('Password');
  await t.expect(await selectFactorPage.factorDescriptionExistsByIndex(0)).eql(false);
  await t.expect(selectFactorPage.getFactorIconClassByIndex(0)).contains('mfa-okta-password');
  await t.expect(await selectFactorPage.factorCustomLogoExist(0)).eql(false);
  await t.expect(selectFactorPage.getFactorSelectButtonByIndex(0)).eql('Select');
  await t.expect(selectFactorPage.getFactorSelectButtonDataSeByIndex(0)).eql('okta_password');
  ``` 
  
How much autonomy do we have to update the test?

File: test/testcafe/framework/page-objects/BasePageObject.js [hunk 11]
```diff
@@ -100,17 +102,22 @@ export default class BasePageObject {
     return this.form.getElement(ionMessagesSelector).innerText;
   }
 
+  getSignoutLinkElement() {
+    const selector = userVariables.v3 ? BACK_TO_SIGNIN_LINK : SIGNOUT_LINK;
+    return Selector(selector);
```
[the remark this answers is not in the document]
shawyu-okta: Yeah, so that's actually the sort of redundancy we can eliminate when we encounter it and use a more direct/targeted selector that does some of the assertion work for us implicitly (since we are asking for the exact piece of text we don't need to also check if the text is that text we asked for, as you pointed out).

Would be great if we have a `getLink(text)` method in BaseForm/BasePage and then the {*}Page classes could wrap those with `getSignoutLink` that calls `getLink("Signout Link Text")` or something like that.

It's up to you if you think deprecation + new method is the right move or if updating all the references in the tests to use a new method is right. I think the latter is best if there are only a few or zero instances of the method outside of the scope of this ticket.

test/testcafe/framework/page-objects/SelectAuthenticatorPageObject.js
glenfannin-okta: I ended up just adding the classes from v2 into v3 to make the spec pass without introducing new conditional statements in each function. 

The way this spec is set up, it seems almost impossible to grab these elements semantically and pass the spec while also maintaining how v2 is structured for this view. 

Wanted to get your input on this.

src/v3/src/transformer/selectAuthenticator/utils.ts
```diff
@@ -36,6 +36,42 @@ const getAuthenticatorDataSeVal = (authenticatorKey: string, methodType?: string
...
+    return authButtons;
+  }
+
+  const updatedAuthenticatorBtns = authButtons.filter(
+    (button) => button.options.actionParams?.['authenticator.methodType'] !== 'signed_nonce',
+  );
+
+  // Re-arrange fastpass in options based on deviceKnown
+  // If deviceKnown is set, set fastpass as the first option in the list
+  // otherwise, place it as the last item in the list of OV options
```
shawyu-okta, line 64: Ah interesting so there is actually some client-side ordering that is expected? Is this logic from v2?

glenfannin-okta: Oh yes! 

https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/ui-schema/ion-object-handler.js#L61

src/v3/src/transformer/selectAuthenticator/utils.ts
```diff
@@ -167,57 +203,80 @@ const formatAuthenticatorOptions = (
...
-    ];
-    const authenticator = option.relatesTo;
+): AuthenticatorButtonElement[] => {
+  const authenticatorOptionSet = new Set<string>();
+  return options
+    .filter((option: IdxOption) => {
+      if (isEnroll) {
+        return true;
+      }
+      // If webauthn enrollments > 1 just show one entry with a generic namne (first)
```
shawyu-okta, line 213: Isn't this a problem during verification, though? If you only show one, it might not match the device the user wants to verify with.

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -199,7 +203,8 @@ export default class BaseFormObject {
   }
 
   getAllErrorBoxTexts() {
-    return this.getInnerTexts(FORM_INFOBOX_ERROR);
+    const selector  = userVariables.v3 ? FORM_INFOBOX_ERROR_SELECTOR : FORM_INFOBOX_ERROR;
```
shawyu-okta, line 206: For v3, is is possible to do this via `getByRole('alert')` and then sub select on the `.infobox-error` class?

test/testcafe/framework/shared/index.js
```diff
@@ -84,6 +84,13 @@ export async function assertRequestMatches(loggedRequest, url, method, body) {
   }
 }
 
+// eslint-disable-next-line @okta/okta/no-unlocalized-text-in-templates
+export const OktaDashboardContent = `
```
shawyu-okta, line 88: nit: I think only classes with constructors should be exported as TitleCase
```suggestion
export const oktaDashboardContent = `
```

src/v3/src/transformer/selectAuthenticator/utils.ts
glenfannin-okta: I wondered the same thing. 

[This](https://github.com/okta/okta-signin-widget/blob/f3f40ed2c66db24abafbbfcb0a5d712ad23d35e8/src/v2/view-builder/internals/FormInputFactory.ts#L48-L59) is the logic this mimics. 

Let me reach out to Aaron/SIW team to understand how that works more.

test/testcafe/framework/page-objects/SelectAuthenticatorPageObject.js
```diff
@@ -20,20 +20,35 @@ export default class SelectFactorPageObject extends BasePageObject {
     super(t);
   }
 
+  getFactorButton() {
```
shawyu-okta, line 23:
```suggestion
  getFactorButtons() {
```

test/testcafe/framework/page-objects/SelectAuthenticatorPageObject.js
```diff
@@ -46,28 +61,38 @@ export default class SelectFactorPageObject extends BasePageObject {
...
   getFactorSelectButtonDataSeByIndex(index) {
     return this.form.getElement(factorSelectButtonDiv).nth(index).getAttribute('data-se');
   }
 
   async selectFactorByIndex(index) {
-    await this.t.click(this.form.getElement(factorSelectButtonSelector).nth(index));
+    await this.t.click(this.getFactorCTAButtonByIndex(index));
   }
 
   async skipOptionalEnrollment() {
```
shawyu-okta, line 80: How difficult would it be to rename this to be more reflective of the action it takes?
```suggestion
  async clickSetUpLaterButton() { // or something like this
```

test/testcafe/framework/page-objects/SelectAuthenticatorPageObject.js
```diff
@@ -46,28 +61,38 @@ export default class SelectFactorPageObject extends BasePageObject {
...
 
   getCustomSignOutLink() {
     return Selector(CUSTOM_SIGN_OUT_LINK_SELECTOR).getAttribute('href');
   }
 
   async clickCustomOTP() {
-    await this.t.click(this.form.getElement(CUSTOM_OTP_BUTTON_SELECTOR));
+    if (userVariables.v3) {
+      const button = this.form.getButton('Atko Custom OTP Authenticator');
+      await this.t.click(button);
```
shawyu-okta, line 92: Does this selector work in v2 too?

src/v3/src/transformer/selectAuthenticator/utils.ts
glenfannin-okta: Looks like it does work as we expect based on [this](https://okta.slack.com/archives/CAZH2MWEL/p1666210457353909) note from Vijet, but I also tested on v3 with this change and it also works. What I have noticed however, is that, it doesnt seem like this logic matters anymore (or maybe I haven't ran into the exact experience/scenario?). What I am seeing in both v2 and v3 is that when the authenticator options are passed to the Widgets from auth-js, even if we are enrolled in multiple webauthN factors it only ever gives one option back, and it is ALWAYS the correct one for the browser we are in. 

It seems like auth-js may be doing something on their end to determine the device we are on and surfaces the corresponding factor id to the widget. 
Screen Shot 2022-10-20 at 5 03 19 PM
Screen Shot 2022-10-20 at 5 14 22 PM

So now my question is, should I keep this logic, skip this test for v3 and remove this logic, or remove the logic in both v2 and v3 and remove the test?

shawyu-okta: To unblock this PR, we can keep the logic. I think we can followup with a tech-debt JIRA to research further and remove the logic later in both widgets with SDK team's input.

test/testcafe/framework/page-objects/SelectAuthenticatorPageObject.js
shawyu-okta: Ahh, right 😢

shawyu-okta (approved):

PR #2859, opened 2022-10-13
OKTA-518966 : Introspect parity spec test

glenfannin-okta:
## Description:

The purpose of this PR is to enable parity v2/v3 spec testing and implement necessary changes to ensure the v3 widget is in parity with v2.

## PR Checklist

### Issue:

- [OKTA-518966](https://oktainc.atlassian.net/browse/OKTA-518966)

### DIFF
--- a/src/v3/src/components/InfoBox/InfoBox.tsx
+++ b/src/v3/src/components/InfoBox/InfoBox.tsx
@@ -48,7 +48,8 @@ const InfoBox: UISchemaElementComponent<{
         variant="infobox"
         title={title}
         // eslint-disable-next-line react/jsx-props-no-spreading
-        {...(dataSe && { 'data-se': dataSe })}
+        {...({ 'data-se': dataSe })}
+        className={`infobox-${messageClass.toLowerCase()}`}
       >
         {message}
       </Alert>
--- a/src/v3/src/components/Widget/index.tsx
+++ b/src/v3/src/components/Widget/index.tsx
@@ -37,7 +37,6 @@ import { transformTerminalTransaction, transformUnhandledErrors } from '../../tr
 import { createForm } from '../../transformer/utils';
 import {
   FormBag,
-  MessageType,
   UISchemaLayout,
   UISchemaLayoutType,
   WidgetProps,
@@ -235,21 +234,7 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     if (!idxTransaction) {
       return;
     }
-
-    const { messages: newMessages = [] } = idxTransaction;
-
     events?.afterRender?.(getEventContext(idxTransaction));
-
-    // for multiple error messages
-    newMessages?.forEach((newMessage) => {
-      const { class: type, message: msg } = newMessage;
-      if (type === MessageType.ERROR) {
-        // error event
-        events?.afterError?.({
-          stepName: idxTransaction?.nextStep?.name,
-        }, { message: msg });
-      }
-    });
 
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [idxTransaction, bootstrap]);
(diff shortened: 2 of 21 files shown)


src/v3/src/components/Widget/index.tsx
```diff
@@ -237,22 +234,8 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
...
-    // for multiple error messages
-    newMessages?.forEach((newMessage) => {
-      const { class: type, message: msg } = newMessage;
-      if (type === MessageType.ERROR) {
-        // error event
-        events?.afterError?.({
-          stepName: idxTransaction?.nextStep?.name,
-        }, { message: msg });
-      }
-    });
```
glenfannin-okta, line 254: Compared to v2, these errors should only be triggered after user interaction (model events), so I moved this into the useOnSubmit hook.

src/v3/src/util/getEventContext.ts
```diff
@@ -14,15 +14,44 @@ import { IdxTransaction } from '@okta/okta-auth-js';
...
+    }
+    // no remediation or only skip remediation with messages for device enrollment state
+    // and the state is meant to be terminal state with different UI than the regular terminal view
+    // @ts-ignore deviceEnrollment missing from type
+    if (rawIdxState.deviceEnrollment) {
+      return IDX_STEP.DEVICE_ENROLLMENT_TERMINAL;
+    }
+  }
+  return formName;
+};
```
glenfannin-okta, line 46: The spec tested for proper formName values in the afterRender hook, I did some digging and found that we were missing some scenarios for applying the proper formName.

src/v3/src/components/InfoBox/InfoBox.tsx
```diff
@@ -48,7 +48,8 @@ const InfoBox: UISchemaElementComponent<{
         variant="infobox"
         title={title}
         // eslint-disable-next-line react/jsx-props-no-spreading
-        {...(dataSe && { 'data-se': dataSe })}
+        {...({ 'data-se': dataSe })}
+        className={`infobox-${messageClass.toLowerCase()}`}
```
shawyu-okta, line 52: I've been wondering if there's any way for us to differentiate between different alerts besides using class. Any ideas?

glenfannin-okta: The message type is structured like this:

```
{
   message: string;
   i18n: {
      key: string;
      params?: string[];
   };
   class: string;
```

Where class can be any of these: `ERROR`, `WARNING`, `INFO`

There are times however, where `INFO` is set when it actually needs to be displayed as an error (we determine that based on the i18n key). So not really sure how we can really determine what should be an error vs warning vs info message (if that was the question) without more information from the server side.

shawyu-okta: I was referring to the ability for testing-library to differentiate between different `alert` boxes. Right now, there isn't much to differentiate the various alerts unless you inspect their class or CSS attributes.

shawyu-okta (approved):

src/v3/src/components/InfoBox/InfoBox.tsx
glenfannin-okta: Ohhhhh, I'm sorry about that, so thinking and looking through the options available to `getByRole`, we could possibly use a hidden description element and using the aria-describedby property. This way, when we try to grab an error message by role, we can do so like this: 

```
getByRole('alert', { description: 'ERROR' });
```

Assuming the HTML looks like this 

```
<div role="alert" aria-describedby="error-desec">
   Some Error message
</div>
<span id="error-desec" aria-hidden="true">ERROR</span>
```

However, the only issue with this is that a screen reader will read out that `ERROR` description.

PR #2869, opened 2022-10-18
OKTA-518971 : ReEnrollAuthenticatorWarningPasswordView parity spec

glenfannin-okta:
## Description:

The purpose of this PR is to implement necessary changes for Re-Enroll authenticator password warning spec for parity testing.

## PR Checklist

### Issue:

- [OKTA-518971](https://oktainc.atlassian.net/browse/OKTA-518971)

### DIFF
--- a/test/testcafe/spec/ReEnrollAuthenticatorWarningPasswordView_spec.js
+++ b/test/testcafe/spec/ReEnrollAuthenticatorWarningPasswordView_spec.js
@@ -45,7 +45,8 @@ const mockChangePasswordNotAllowed = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/skip')
   .respond(xhrSuccess);
 
-fixture('Password Authenticator Expiry Warning');
+fixture('Password Authenticator Expiry Warning')
+  .meta('v3', true);
 
 async function setup(t) {
   const passwordExpiryWarningPage = new FactorEnrollPasswordPageObject(t);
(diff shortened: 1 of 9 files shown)


test/testcafe/spec/ReEnrollAuthenticatorWarningPasswordView_spec.js
```diff
@@ -65,21 +66,21 @@ async function setup(t) {
...
       await t.expect(passwordExpiryWarningPage.getFormTitle()).eql(formTitle);
-      await t.expect(passwordExpiryWarningPage.getSaveButtonLabel()).eql('Change Password');
+      await t.expect(passwordExpiryWarningPage.changePasswordButtonExists()).eql(true);
       await t.expect(passwordExpiryWarningPage.getRequirements()).contains('Password requirements:');
       await t.expect(passwordExpiryWarningPage.getRequirements()).contains('At least 8 characters');
       await t.expect(passwordExpiryWarningPage.getRequirements()).contains('An uppercase letter');
       await t.expect(passwordExpiryWarningPage.getRequirements()).contains('A lowercase letter');
       await t.expect(passwordExpiryWarningPage.getRequirements()).contains('A number');
       await t.expect(passwordExpiryWarningPage.getRequirements()).contains('No parts of your username');
       await t.expect(passwordExpiryWarningPage.getRequirements()).contains('Your password cannot be any of your last 4 passwords');
```
glenfannin-okta, line 80: I had to disable this test for v3 because we removed server side password requirements from the view. 

Should I instead add a conditional statement here to allow the other assertions to pass (I tested the locally and they do pass) @shawyu-okta

test/testcafe/spec/ReEnrollAuthenticatorWarningPasswordView_spec.js
```diff
@@ -130,14 +131,14 @@ test
     });
   });
 
-test
+test.meta('v3', false)
```
glenfannin-okta, line 134: I had to disable this test because it expects the previous transaction's elements to stay in the UI even though they are not included with this error message (it is a terminal transaction). 

I checked master and this test has not been changed. Should I make the changes in v3 to reserve that step's data to re-render on this transaction? @shawyu-okta

shawyu-okta: Let's file a JIRA for this and mark it here. We can followup with the change if needed separately

test/testcafe/spec/ReEnrollAuthenticatorWarningPasswordView_spec.js
shawyu-okta: Hmm good question - I think normally we don't want to change the test assertions dependent on which version of the widget due to our goal of parity but I think in this case we did make the call that this was not a parity feature we wanted, so I think that's OK if we mark it as such with a short explanation.

shawyu-okta (approved):

PR #2894, opened 2022-11-01
Check for Possible Dependency Confusion

denysoblohin-okta:
## Description:

There are several internal packages that are used only during build process and should be protected from potential Dependency Confusion attack (described [here](https://oktainc.atlassian.net/browse/OKTA-529256)):
- typingdna
- okta-i18n-bundles
- handlebars-inline-precompile
Also added to the list:
- duo
- qtip
- okta (corage-dist)

These packages should have `@okta/` namespace to prevent potential code injection from any package registry.

Changes:
- Added `@okta/` namespace prefix for packages listed above
- Added tests to `verify-package` script that tries to inject fake malicious modules with same names  and checks that such modules are NOT included in output bundle file and lock file.
- Added script `find-internal-packages` (uses [npm-check](https://github.com/dylang/npm-check)) to detect using of internal packages from source code. 
Current output:
```
Internal packages:
- not present in registry: @okta/handlebars-inline-precompile, @okta/duo, @okta/qtip, @okta/typingdna, @okta/qtip, @okta/okta-i18n-bundles
- present in registry: backbone
- not present in registry, installed in node_modules: @okta/okta
```
Throws error code 1 if found packages that are installed to "node_modules" without having `@okta` namespace

## PR Checklist

### Issue:

- [OKTA-529256](https://oktainc.atlassian.net/browse/OKTA-529256)

### Reviewers:
@aarongranick-okta


scripts/find-internal-packages.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+source $OKTA_HOME/$REPO/scripts/setup.sh
+
+if ! NODE_TLS_REJECT_UNAUTHORIZED=0 yarn find-internal-packages; then
```
jaredperreault-okta, line 5: why do we need to turn off ssl for this?

scripts/find-internal-packages.js
```diff
@@ -0,0 +1,153 @@
...
+  return {
+    ...pkg,
+    files
+  };
+};
+
+// Filter result of npm-check: include only internal packages
+const filterPackages = (pkg, result) => {
+  if (pkg.unused) {
+    // Ignore unused package
```
jaredperreault-okta, line 50: nit: add warning on unused package

scripts/find-internal-packages.js
```diff
@@ -0,0 +1,153 @@
...
+    result.hidden.push(pkg.moduleName);
+    return false;
+  }
+  if (aliases.includes(pkg.moduleName) && !pkg.regError && pkg.pkgError) {
+    // Internal package that IS present in registry but not installed in node_modules (backbone)
+    result.overridden.push(pkg.moduleName);
+    return false;
+  }
+  if (aliases.includes(pkg.moduleName) && pkg.regError && pkg.isInstalled) {
+    // Internal package that is not present in registry but is present in node_modules (@okta/okta)
```
jaredperreault-okta, line 68: nit: provide example of this case occurring

scripts/find-internal-packages.sh
denysoblohin-okta: Without this npm-check returns error `Registry error self signed certificate in certificate chain` for all npm packages.
Probably related to this: https://github.com/okta/okta-signin-widget/blob/master/scripts/setup.sh#L12-L13

jaredperreault-okta: Do we need the full feature set of `npm-check`? Maybe `npm outdated` can work for us?

https://github.com/dylang/npm-check/issues/11#issuecomment-294296207

aarongranick-okta (approved):
There is some really great work here! Unfortunately it breaks the courage-for-signin-widget build. I think some of this logic (particularly in handlebars-inline-precompile and okta-i18n-bundles) needs to be moved upstream to okta-ui/courage (or the courage-for-signin-widget build has to make the modifications at the source to produce the needed changes in courage-dist)

aarongranick-okta (changes requested):
There is some really great work here! Unfortunately it breaks the courage-for-signin-widget build. I think some of this logic (particularly in handlebars-inline-precompile and okta-i18n-bundles) needs to be moved upstream to okta-ui/courage (or the courage-for-signin-widget build has to make the modifications at the source to produce the needed changes in courage-dist)

scripts/find-internal-packages.sh
denysoblohin-okta: `npm outdated` just checks new versions of installed packages.
`npm-check` can analyse sources and detect packages that are referenced in sources but not present in NPM registry - this feature is very useful to detect internal packages, I haven't found alternatives to it

scripts/verify-package.sh
```diff
@@ -12,11 +12,78 @@ fi
 
 set -e
 
+# Internal packages to check for potential Dependency Confusion Attack (OKTA-529256)
+INTERNAL_PACKAGES_TO_FAKE=("@okta/typingdna" "@okta/handlebars-inline-precompile" "@okta/okta-i18n-bundles" "@okta/duo" "@okta/qtip")
+INTERNAL_PACKAGES=(${INTERNAL_PACKAGES_TO_FAKE[@]} "@okta/okta")
+inject_marker="f@@@@@@ke"
+
+# Inject fake packages with same names as ones that are used internally
+inject_fake_packages() {
```
aarongranick-okta, line 21: Please add a method to reverse this process and make sure that if we run this script locally it will reverse the changes even if the script fails. See the trap for "restore_node_modules" as an example.

aarongranick-okta:
Closing in favor of https://github.com/okta/okta-signin-widget/pull/2993

PR #2897, opened 2022-11-02
OKTA-539161 : Login redirect L10 error fix

leemchale-okta:
## Description:

This PR fixes the issue where an L10 error appears during the redirect when trying to login via bookmark app.

## PR Checklist

### Issue:

- [OKTA-539161](https://oktainc.atlassian.net/browse/OKTA-539161)

### Screenshot/Video:
**Before**: 
Screen Shot 2022-11-02 at 11 05 17 AM

**After**:
Screen Shot 2022-11-02 at 11 08 28 AM

### DIFF
--- a/src/v3/src/transformer/redirect/redirectTransformer.ts
+++ b/src/v3/src/transformer/redirect/redirectTransformer.ts
@@ -40,8 +40,6 @@ export const redirectTransformer = (
   const appInfo = getAppInfo(transaction);
   const userInfo = getUserInfo(transaction);
 
-  let titleContent = 'oie.success.text.signingIn';
-
   if (interstitialBeforeLoginRedirect === InterstitialRedirectView.DEFAULT) {
     uischema.elements.push({
       type: 'Spinner',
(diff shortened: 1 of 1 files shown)


src/v3/src/transformer/redirect/redirectTransformer.ts
```diff
@@ -63,6 +62,7 @@ export const redirectTransformer = (
   const { label, name } = appInfo;
   const appName = label ?? name;
```
glenfannin-okta, line 63: When you debug, can you break here and see if appInfo has any data or possibly check the transaction object and see if there is any information in the context that indicates the app name? 

I only ask because with this, I thought maybe the bookmark name would show up as the app name but maybe not? 🤔

leemchale-okta: appInfo did not have any data.  I looked at the transaction object and there is no app data like name or anything

leemchale-okta: image

glenfannin-okta: Awesome! So this is expected! Great! Thank you so much!

glenfannin-okta (approved):

src/v3/src/transformer/redirect/redirectTransformer.ts
```diff
@@ -76,7 +76,7 @@ export const redirectTransformer = (
 
   uischema.elements.unshift({
     type: 'Description',
-    options: { content: loc(titleContent, 'login', [contentParams]) },
+    options: { content: loc(titleContent, 'login', contentParams.length > 0 ? contentParams : undefined) },
```
shawyu-okta, line 79: Might it be simpler to rewrite some of this logic so that all of the work is enclosed in each `if/else`? Sort of like this:

```
if (condition1) {
  unshift a description element matching condition 1 onto uischema
} else if (condition2) {
  unshift a description element matching condition 2 onto uischema
} else {
  unshift a default description element onto uischema
}
```

shawyu-okta: If this works it would eliminate the `let titleContent` and also the `contentParams` issues

shawyu-okta (approved):

PR #2906, opened 2022-11-04
OKTA-547018 : PIV/CAC v3 implementation 

glenfannin-okta:
## Description:
The purpose of this PR is to add the initial changes for PIV/CAC which includes the button, divider and the action to switch the form in a synthetic way (since no network request can be made to switch the form).

This PR also includes changes for [OKTA-547019](https://oktainc.atlassian.net/browse/OKTA-547019) to complete the PIV/CAC implementation.

## PR Checklist

### Issue:

- [OKTA-547018](https://oktainc.atlassian.net/browse/OKTA-547018)

### Screenshot/Video:
Button implementation
image

Full sign in implementation: 

https://user-images.githubusercontent.com/97472729/202023704-dd8c003f-ca51-499e-b6cb-7d63c49bedcc.mov


src/v3/src/transformer/button/transformIDPButtons.ts
```diff
@@ -0,0 +1,88 @@
...
+  ButtonType,
+  DividerElement,
+  IWidgetContext,
+  TransformStepFnWithOptions,
+} from '../../types';
+import { loc } from '../../util';
+
+const PIV_TYPE = 'X509';
+
+// TODO: Implement CUSTOM IDP Buttons here
```
glenfannin-okta, line 25: The idea here is to group all IDP buttons in this transformer, the reason for this is because this is also how v2 groups them, PIV/CAC is just always placed as the first (when custom IDPs also exist). But since we have not implemented custom IDP buttons yet, I am marking this with a TODO, will create a ticket for custom IDP implementation and add here.

src/v3/src/transformer/button/transformIDPButtons.ts
```diff
@@ -0,0 +1,88 @@
...
+        );
+        setIdxTransaction({
+          ...transaction,
+          messages: [],
+          neededToProceed: pivRemediation,
+          availableSteps,
+          nextStep: { name: IDX_STEP.PIV_IDP },
+        });
+      },
+    },
```
glenfannin-okta, line 68: This is the crux of the change that allows us to synthetically switch forms (when you cant make a network request). The reason I am updating the remediation/availableSteps is so the default transformers cant add buttons/links that do not belong on the next view. 

In these scenarios the custom transformer will add all the buttons/links necessary (except for cancel - if available), so we just clean up the remediation array to only contain the steps we care about.

src/v3/src/components/AuthCoin/AuthCoinConfig.ts
```diff
@@ -112,6 +112,13 @@ const AuthCoinByAuthenticatorKey: Record<string, AuthCoinConfig> = {
     description: loc('factor.call', 'login'),
     iconClassName: 'mfa-okta-phone',
   },
+  [IDX_STEP.PIV_IDP]: {
+    icon: IDPIcon,
```
glenfannin-okta, line 116: Awaiting the real PIV/CAC AuthCoin Icon, this is used as a placeholder for now.

shawyu-okta (commented):
some initial comments

src/v3/test/integration/authenticator-piv-cac-verification.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  it('should render PIV/CAC view when clicking PIV button', async () => {
+    const {
+      container, user, findByRole, findByText,
+    } = await setup({ mockResponse });
+
+    const pivButton = await findByText('Sign in with PIV / CAC card', { selector: 'button' });
+
+    await user.click(pivButton);
```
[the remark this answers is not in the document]
glenfannin-okta: Yep, we totally can, I didnt initially because we wont be able to use the util function, but I will add the custom assertion here for the request. 🙏

src/v3/src/components/PIVButton/PIVButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const { options } = uischema;
+
+  const btnLabel = getTranslation(translations);
+
+  const { setMessage, loading } = useWidgetContext();
+  const [waiting, setWaiting] = useState<boolean>(false);
+
+  const showLoading = waiting || loading;
+
+  const initCertPrompt = async () => {
```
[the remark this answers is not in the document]
glenfannin-okta: Oops, this was leftover from my initial implementation, I thought I was going to be able to asyn/await but you can't. Will update.

src/v3/src/components/PIVButton/PIVButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+} from '../../types';
+import { getTranslation } from '../../util';
+
+const PIVButton: UISchemaElementComponent<{
+  uischema: PIVButtonElement
+}> = ({ uischema }) => {
+  const { translations = [] } = uischema;
+  const { options } = uischema;
+
+  const btnLabel = getTranslation(translations);
```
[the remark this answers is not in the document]
glenfannin-okta: Oh no, you are totally right, I am not sure of the original thought behind allowing to grab a translation without specifying the key, but I will update this and the function (and any others using it this way).

src/v3/src/components/AuthCoin/AuthCoinConfig.ts
```diff
@@ -112,6 +112,13 @@ const AuthCoinByAuthenticatorKey: Record<string, AuthCoinConfig> = {
     description: loc('factor.call', 'login'),
     iconClassName: 'mfa-okta-phone',
   },
+  [IDX_STEP.PIV_IDP]: {
```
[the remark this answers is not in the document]
glenfannin-okta: From my understanding, we need this to be different because the original `redirect-idp` remediation name is shared between PIV/CAC and custom IDPs. This was established in v2 as the work around for the PIV/CAC implementation to distinguish between the different types of `redirect-idp` remediations.

src/v3/test/integration/authenticator-piv-cac-verification.test.tsx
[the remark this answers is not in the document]
glenfannin-okta: Actually, a correction to this, we can't assert on the request because we are using a v2 util function that loads a hidden form on the page (with the action value being the MTLS endpoint) and submits it. There doesn't seem to be a way for me to spy this to verify. 

But I am adding a unit test to the PIVButton component to assert it triggers the click as expected.

src/v3/src/components/AuthCoin/AuthCoinConfig.ts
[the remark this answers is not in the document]
glenfannin-okta: Ohh, I think you mean just the actual constant being used (`AUTHENTICATOR_KEY` vs `IDX_STEP`). I don't _think_ it is an issue, originally, I didn't add PIV_IDP to the `AUTHENTICATOR_KEY` constant because I thought that would be confusing since it isn't actually an authenticator key.

src/v3/test/integration/authenticator-piv-cac-verification.test.tsx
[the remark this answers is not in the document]
shawyu-okta: Ah okay, perhaps this is something that is coverable in the testcafe suites since those have network spies.

src/v3/src/transformer/button/transformIDPButtons.ts
```diff
@@ -0,0 +1,97 @@
...
+    type: 'Divider',
+    options: { text: loc('socialauth.divider.text', 'login') },
+  };
+
+  if (isBottomDivider) {
+    const titleIndex = formbag.uischema.elements.findIndex((element) => element.type === 'Title');
+    const pivButtonPos = titleIndex !== -1 ? titleIndex + 1 : 0;
+    // Add button after title (if exists) otherwise add to top of array
+    formbag.uischema.elements.splice(pivButtonPos, 0, pivButton);
+    formbag.uischema.elements.splice(pivButtonPos + 1, 0, dividerElement);
```
shawyu-okta, line 85: Does the multi-value overload of `splice` work here?

```
splice(start, deleteCount, item1, item2, itemN)
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

src/v3/src/components/PIVButton/PIVButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [waiting, setWaiting] = useState<boolean>(false);
+
+  const showLoading = waiting || loading;
+
+  const initCertPrompt = () => {
+    setWaiting(true);
+    setMessage(undefined);
+
+    options.onClick();
+  };
```
shawyu-okta, line 44: This is named `initCertPrompt` but none of the enclosed code is clearly doing that. I realize that `options.onClick()` will actually be what does that, and will probably always be the same function, but still feels a little strange to me. Any thoughts? I'm thinking someone looking through this in the future will be confused at first.

glenfannin-okta: Totally agree with this, this doesn't make much sense in this context, what I ended up doing was moving the cert trigger directly into this `PIVButton` component. The reason I did this was because this component is pretty specific to PIV, I contemplated maybe re-using it for other IDP buttons but I haven't researched those enough to understand the difference, I figured, when we get to that implementation, we can make the decision and refactor this component along with adding the `onClick` to the interface if needed. But since this is the only thing using it now, I thought it made sense to add it directly to the button that uses. 

What do you think?

src/v3/src/transformer/button/transformIDPButtons.ts
glenfannin-okta: Works like a charm, ty!

src/v3/src/transformer/button/transformIDPButtons.ts
```diff
@@ -0,0 +1,105 @@
...
+    label: piv?.text || loc('piv.cac.card', 'login'),
+    options: {
+      type: ButtonType.BUTTON,
+      step: IDX_STEP.PIV_IDP,
+      dataSe: 'piv-card-button',
+      classes: `${piv?.className || ''} piv-button`,
+      variant: 'secondary',
+      onClick: (widgetContext: IWidgetContext) => {
+        // To render the PIV view, we have to use a remediation that is provided on initial load
+        // This remeidation doesnt allow a network request, so we have to update the transaction
```
shawyu-okta, line 57:
```suggestion
        // This remediation doesn't allow a network request, so we have to update the transaction
```

src/v3/src/components/PIVButton/PIVButton.tsx
shawyu-okta: I think that's okay for now - it does feel like this component can become more generic as one that always performs an action on load and then can redo that action on click.

src/v3/src/transformer/button/transformIDPButtons.ts
shawyu-okta: I think we want some better unit test coverage of this button transformer or perhaps just the onClick function of the Button if possible

shawyu-okta (approved):

PR #2913, opened 2022-11-08
OKTA-546292 : Yubikey OTP implementation 

leemchale-okta:
## Description:

This PR implements the Yubikey OTP authenticator in SIW V3.

## PR Checklist

### Issue:

- [OKTA-546292](https://oktainc.atlassian.net/browse/OKTA-546292)

### Screenshot/Video:
image
image
Screen Shot 2022-11-09 at 12 55 48 PM


test/testcafe/spec/AuthenticatorYubikeyView_spec.js
```diff
@@ -100,13 +101,13 @@ test
...
-    await pageObject.submit();
+    await pageObject.clickVerifyButton();
 
     pageObject.form.waitForErrorBox();
-    await t.expect(pageObject.form.getErrorBoxText()).eql('We found some errors. Please review the form and make corrections.');
+    await t.expect(pageObject.errorBoxTextExists()).eql(true);
   });
 
-test.requestHooks(verifyMock)('should show custom factor page link', async t => {
+test.meta('v3', false).requestHooks(verifyMock)('should show custom factor page link', async t => {
```
leemchale-okta, line 110: Disabled since v3 does not support this yet

src/v3/src/components/ImageWithText/ImageWithText.tsx
```diff
@@ -23,21 +23,27 @@ const ImageWithText: UISchemaElementComponent<{
...
 }> = ({ uischema }) => {
   const Icon = uischema.options.SVGIcon;
+  const { textContent } = uischema.options;
 
   return (
     <Box
       id={uischema.options.id}
       display="flex"
-      justifyContent="flex-start"
+      justifyContent={textContent ? 'flex-start' : 'center'}
```
leemchale-okta, line 32: image was not being centered properly

src/v3/src/util/getEventContext.ts
```diff
@@ -49,7 +50,7 @@ export const getEventContext = (transaction: IdxTransaction): EventContext => {
   const { context } = transaction;
   const authenticatorKey = context.currentAuthenticator?.value?.key
     || getAuthenticatorKey(transaction);
-  const methodType = transaction.context.currentAuthenticator?.value?.type;
+  const methodType = getAuthenticatorMethod(transaction);
```
leemchale-okta, line 53: methodType was not being extracted from the correct location

src/v3/src/util/getAuthenticatorMethod.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+import { IdxTransaction } from '@okta/okta-auth-js';
+
+export const getAuthenticatorMethod = (
+  transaction: IdxTransaction,
+): string | undefined => {
+  // currentAuthenticator is from enrollment flows and currentAuthenticatorEnrollment is from verify flows
+  // @ts-ignore TODO: Add currentAuthenticatorEnrollment to IdxContext type
+  const { context: { currentAuthenticator, currentAuthenticatorEnrollment } } = transaction;
```
leemchale-okta, line 20: currentAuthenticatorEnrollment was missing from IdxContext type

playground/mocks/config/responseConfig.js
```diff
@@ -62,6 +62,8 @@ const idx = {
     // 'authenticator-verification-select-authenticator',
     // 'authenticator-verification-select-authenticator-ov-m2',
     // 'authenticator-verification-webauthn',
+    // 'authenticator-verification-yubikey',
```
lesterchoi-okta, line 65: You shouldn't (typically) commit changes to the `responseConfig.js` since it affects all developers' default playground experience. This is a DX pain-point we plan to address in the future.

src/v3/src/components/ImageWithText/ImageWithText.tsx
lesterchoi-okta: Pass `uischema.options.alignment` to `ImageWithText` (default: `flex-start`) for now. Eventually hoping to convert to a markdown renderer

src/v3/src/img/yubikeyDemo.svg
```diff
@@ -0,0 +1,19 @@
+<svg width="91" height="85" viewBox="0 0 91 85" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="yubikeyDemo">
```
lesterchoi-okta, line 1: @alexdahl-okta This image isn't considered an "icon" is it? i.e., it shouldn't be theme-able, right?

src/v3/src/img/yubikeyDemo.svg
```diff
@@ -0,0 +1,19 @@
...
+<title id="yubikeyDemo">yubikey</title>
```
lesterchoi-okta, line 2: Could you stick a FIXME/TODO note here that the `<title>` needs to be i18n/l10n-friendly?

src/v3/src/img/yubikeyDemo.svg
```diff
@@ -0,0 +1,19 @@
...
+</g>
+<defs>
+<clipPath id="clip0_201_13130">
+<rect width="91" height="85" rx="4" fill="white"/>
```
lesterchoi-okta, line 16: Not sure this should be filled? @alexdahl-okta

src/v3/src/mocks/response/idp/idx/authenticator-enroll-yubikey.json
```diff
@@ -0,0 +1,200 @@
+{
```
lesterchoi-okta, line 1: re-use/update existing v2 mock?

https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/data/idp/idx/authenticator-enroll-yubikey.json

src/v3/src/mocks/response/idp/idx/authenticator-verification-yubikey.json
```diff
@@ -0,0 +1,177 @@
+{
```
lesterchoi-okta, line 1: re-use/update existing v2 mock?
https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/data/idp/idx/authenticator-verification-yubikey.json

lesterchoi-okta (changes requested):
- revert changes to `responseConfig.js`
- ping design re; svg

src/v3/src/img/yubikeyDemo.svg
alexdahl-okta: Good catch; should **not** be filled white.

I updated Figma, I think it should also be safe to just delete `fill="white"`

src/v3/src/transformer/yubikey/transformYubikeyOtpAuthenticator.ts
```diff
@@ -0,0 +1,88 @@
...
+  };
+
+  uischema.elements = [
+    titleElement,
+    descriptionElement,
+    passcodeElement,
+    submitButtonControl,
+  ];
+
+  if (nextStep!.name === IDX_STEP.ENROLL_AUTHENTICATOR) {
```
lesterchoi-okta, line 83: 👍 that's careful attention to detail. just confirmed the design

src/v3/src/util/getAuthenticatorMethod.ts
```diff
@@ -0,0 +1,25 @@
...
+export const getAuthenticatorMethod = (
+  transaction: IdxTransaction,
+): string | undefined => {
+  // currentAuthenticator is from enrollment flows and currentAuthenticatorEnrollment is from verify flows
+  // @ts-ignore OKTA-549893 TODO: Add currentAuthenticatorEnrollment to IdxContext type
+  const { context: { currentAuthenticator, currentAuthenticatorEnrollment } } = transaction;
+  const enrollMethods = currentAuthenticator?.value?.methods;
+  const verifyMethods = currentAuthenticatorEnrollment?.value?.methods;
+
+  return (enrollMethods && enrollMethods[0].type) || (verifyMethods && verifyMethods[0].type);
```
lesterchoi-okta, line 24: Are there -- or will there be -- cases where we want another index? e.g., `verifyMethods[1]`. First one that comes to mind is phone voice vs sms. Maybe the index is a param with default=0?

src/v3/test/integration/authenticator-enroll-yubikey-otp.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { createAuthJsPayloadArgs, setup } from './util';
+
+import mockResponse from '@okta/mocks/data/idp/idx/authenticator-enroll-yubikey.json';
```
lesterchoi-okta, line 15: nice

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
```diff
@@ -9,7 +10,22 @@ export default class YubiKeyAuthenticatorPageObject extends BasePageObject {
...
+    }
+    return this.clickVerifyButton();
+  }
+
+  clickVerifyButton() {
+    return this.form.clickButton('Verify');
+  }
+
+  errorBoxTextExists() {
+    return this.form.getByTextOnScreen('We found some errors. Please review the form and make corrections.').exists;
```
lesterchoi-okta, line 29: Is there a better -- more semantic -- way to grab the error box? This seems 
1. too closely tied to the translation
2. implicitly testing the message

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
```diff
@@ -9,7 +10,22 @@ export default class YubiKeyAuthenticatorPageObject extends BasePageObject {
     return this.form.setTextBoxValue(name, value);
   }
 
-  submit() {
-    return this.form.clickSaveButton();
+  async waitForYubikeyView() {
```
lesterchoi-okta, line 13: Why did you have to add this wait? Doesn't the navigate check console and wait for the form to render?

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -193,6 +193,10 @@ export default class BaseFormObject {
     return within(this.el).getByText(text, options);
   }
 
+  getByTextOnScreen(text, options = undefined) {
+    return screen.getByText(text, options);
+  }
+
```
lesterchoi-okta, line 199: Is this just to access `screen` var? Seems redundant...
And `options = undefined` looks unnecessary

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
```diff
@@ -9,7 +10,22 @@ export default class YubiKeyAuthenticatorPageObject extends BasePageObject {
...
+  clickEnrollButton() {
+    if(userVariables.v3) {
+      return this.form.clickButton('Set up');
+    }
+    return this.clickVerifyButton();
+  }
+
+  clickVerifyButton() {
+    return this.form.clickButton('Verify');
+  }
```
lesterchoi-okta, line 26: Just checking: this replaces `submit()` because the semantic selector needs to find by text, right?

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
```diff
@@ -65,13 +66,13 @@ test
     await t.expect(pageObject.getFormTitle()).eql('Set up YubiKey');
     
     // Do not fill out the form and submit
-    await pageObject.submit();
+    await pageObject.clickEnrollButton();
 
     pageObject.form.waitForErrorBox();
-    await t.expect(pageObject.form.getErrorBoxText()).eql('We found some errors. Please review the form and make corrections.');
+    await t.expect(pageObject.errorBoxTextExists()).eql(true);
```
lesterchoi-okta, line 72: I think the previous assertion was more explicit/clear. If there's no way to grab the error box without knowing the text, that's the real problem.

lesterchoi-okta (changes requested):
- My bad on the `ImageWithText` "alignment" prop name suggestion. Didn't notice it was "flex". It's fine for now
- A few undesirable spec changes to be reverted or addressed.

src/v3/src/components/ImageWithText/ImageWithText.tsx
```diff
@@ -23,21 +23,27 @@ const ImageWithText: UISchemaElementComponent<{
...
   const Icon = uischema.options.SVGIcon;
+  const { textContent, alignment = 'flex-start' } = uischema.options;
 
   return (
     <Box
       id={uischema.options.id}
       display="flex"
-      justifyContent="flex-start"
+      justifyContent={alignment}
       alignItems="center"
```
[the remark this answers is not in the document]
leemchale-okta: yeah i see what you mean now.  Okay sounds good.

src/v3/src/util/getAuthenticatorMethod.ts
leemchale-okta: yeah so I was using this code in v2 as a reference: [https://github.com/okta/okta-signin-widget/blob/master/src/v2/models/AppState.ts#L63]

We dont have any cases currently where we would need another index but i agree as a util function I think we would want it to be able to get another index since there will probably be cases for it in the future.  I like having the index as a param and will change it to that. 👍

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
lesterchoi-okta: There should be some aria attrs that you can use. If not we have to add them (separate bug/PR)

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
leemchale-okta: yeah the problem was there are two elements with role="alert", one being the error box and the other being the error element under the input field.  This was causing the findByRole method to not grab the error box correctly.  Ill see if I can find another way.

leemchale-okta: also the error box is not inside the `form`

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
leemchale-okta: the `checkConsoleMessages` tests were failing unless i specifically waited for the form to render
image

leemchale-okta: seems like the navigate doesnt wait for the form to fully render

test/testcafe/framework/page-objects/components/BaseFormObject.js
leemchale-okta: yeah I made this since the error box is outside the form but may not need it anymore

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
leemchale-okta: thats right it needs to find the button semantically by text.  Especially since in V3 the button text between the enroll and verify flow are different.

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
leemchale-okta: yeah i will see if I can find a better way

leemchale-okta: Yeah it looks like since the error box is not inside the form, the getErrorBox function cant grab it.  And I dont see any aria attrs that i can use

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
```diff
@@ -35,24 +35,24 @@ async function setup(t) {
...
 
+    await t.expect(pageObject.getFormTitle()).eql('Set up YubiKey');
+    await t.expect(pageObject.getFormSubtitle()).eql('Insert the YubiKey into a USB port and tap it to generate a verification code.');
+
     await checkConsoleMessages({
       controller: 'enroll-yubikey',
       formName: 'enroll-authenticator',
       authenticatorKey:'yubikey_token',
       methodType: 'otp'
     });
```
leemchale-okta, line 51: moved the checkConsoleMessages method lower to wait for the form to fully load

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
```diff
@@ -65,30 +65,30 @@ test
...
 
+    await t.expect(pageObject.getFormTitle()).eql('Verify with YubiKey');
+    await t.expect(pageObject.getFormSubtitle()).eql('Insert the YubiKey into a USB port and tap it to generate a verification code.');
+
     await checkConsoleMessages({
       controller: 'mfa-verify',
       formName: 'challenge-authenticator',
       authenticatorKey:'yubikey_token',
       methodType: 'otp'
     });
```
leemchale-okta, line 87: moved the checkConsoleMessages method lower to wait for the form to fully load

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
leemchale-okta: I moved the `checkConsoleMessages` method lower to wait for the form to fully load and eliminated the need for this `waitForYubikeyView` method

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
lesterchoi-okta: 1. Should the error box be inside the form?
2. Isn't this the page object?

I'm ok if this is following an existing pattern, but it might be revealing an a11y or util structure issue?

leemchale-okta: I checked in V2 and confirmed the error box there does reside inside the form element.  I found that in V3 we are using the InfoSection component for displaying client side errors such as empty input cases which this spec is testing for.  This InfoSection component is not currently placed inside the form element.  I think if we move this component to inside the form it should solve this problem.

leemchale-okta: then we should be able to use the original assertion.

src/v3/src/components/Form/Form.tsx
```diff
@@ -94,6 +96,7 @@ const Form: FunctionComponent<{
       data-se="o-form"
       style={{ maxWidth: '100%', wordBreak: 'break-word' }}
     >
+      <InfoSection message={message} />
```
leemchale-okta, line 99: Moved the `InfoSection` component here so client side errors will be inside the `form` element and match V2.

test/testcafe/framework/page-objects/YubiKeyAuthenticatorPageObject.js
leemchale-okta: Moved the `InfoSection` component here so client side errors will be inside the form element and match V2.

src/v3/src/components/Widget/index.tsx
```diff
@@ -291,7 +291,6 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
               />
               <AuthContent>
                 <IdentifierContainer />
-                <InfoSection message={message} />
                 {
                   uischema.elements.length > 0
                     ? <Form uischema={uischema as UISchemaLayout} />
```
lesterchoi-okta, line 296: If `<InfoSection>` moved inside the `<Form>`, don't you have to pass down the prop?

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -210,7 +210,7 @@ export default class BaseFormObject {
 
   getErrorBoxText() {
     if (userVariables.v3) {
-      return within(this.el).getByRole('alert').innerText;
+      return within(this.el).getAllByRole('alert').nth(0).innerText;
```
lesterchoi-okta, line 213: Is this the change that allows you to get the correct error box? Seems like `getByRole` would default to the first match

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
lesterchoi-okta: Is this happening because the `afterRender` was triggering prematurely? That should be fixed in the latest on `0.0`. If it's ported to `0.1` the order shouldn't need to be changed in this spec.

src/v3/src/components/Form/Form.tsx
lesterchoi-okta: See question on Widget component about passing through the `message` prop

lesterchoi-okta: ah nvm. it's coming from `useWidgetContext`

lesterchoi-okta (approved):

src/v3/src/types/schema.ts
```diff
@@ -311,7 +311,8 @@ export interface ImageWithTextElement extends UISchemaElement {
   options: {
     id: string;
     SVGIcon: FunctionComponent;
-    textContent: string;
+    textContent?: string;
+    alignment?: string;
```
shawyu-okta, line 315: We should consider updating this in a future PR to use a string enum

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
```diff
@@ -100,13 +100,13 @@ test
 
     await t.expect(pageObject.getFormTitle()).eql('Verify with YubiKey');
     
-    await pageObject.submit();
+    await pageObject.clickVerifyButton();
 
-    pageObject.form.waitForErrorBox();
+    await pageObject.form.waitForErrorBox();
     await t.expect(pageObject.form.getErrorBoxText()).eql('We found some errors. Please review the form and make corrections.');
```
glenfannin-okta, line 106: These two are actually redundant assertions, now I see why the change was made in the Base form object. We could just use the waitForErrorBox assertion since it uses the name of the alert. We can update this later.

glenfannin-okta (approved):

src/v3/src/types/schema.ts
lesterchoi-okta: It needs more TLC any way, the name "alignment" is  misleading since the display is "flex"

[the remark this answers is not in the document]
lesterchoi-okta: Looks like a +1 of the comment from Shaw

test/testcafe/framework/page-objects/components/BaseFormObject.js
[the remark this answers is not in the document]
lesterchoi-okta: Agreed. TODO

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
lesterchoi-okta: Agreed, redundant

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
```diff
@@ -35,24 +35,24 @@ async function setup(t) {
...
 }
 
-fixture('Enroll YubiKey Authenticator');
+fixture('Enroll YubiKey Authenticator').meta('v3', true);
 test
   .requestHooks(logger, enrollMock)('enroll with YubiKey authenticator', async t => {
     const pageObject = await setup(t);
 
+    await t.expect(pageObject.getFormTitle()).eql('Set up YubiKey');
+    await t.expect(pageObject.getFormSubtitle()).eql('Insert the YubiKey into a USB port and tap it to generate a verification code.');
```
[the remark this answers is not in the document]
glenfannin-okta: Ohh, that is caused by the function trying to assert on console logs before the form renders, we have a solution for this, it is just to perform an assertion on the existence of the `form` element. This resolves the issue, but no worries, since this also effectively does the same thing no reason to insert another useless assertion. Ty!

PR #2919, opened 2022-11-14
Fixed user enumeration problem in SSPR by removing wording that contains masked email

hanyuanzhang-okta:
## Description:
- Fixed user enumeration problem in SSPR by removing the if condition that displays the wording with masked email if emails exists.
- The problem happens when going to /signin/forgot-password directly with a valid user identifier, or when adding/removing a phone authenticator in settings

## PR Checklist

### Issue:

- [OKTA-525791](https://oktainc.atlassian.net/browse/OKTA-525791)

### Reviewers:
@okta/ciamx @jaredperreault-okta 
### Screenshot/Video:
Please note that there is a known SIW rendering issue that causes the extra spaces that is not related to this PR
NoMaskedEmailInAuthenticatorEnrollment https://okta.box.com/s/nyp77d2puq6sdy0mz3yhj82eiudis7e0
NoMaskedEmailInSSPR https://okta.box.com/s/oyme05nu5r18354j2xf7t48v4kyr7n0a

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&branch=d16t-okta-signin-widget-442c6d7-6375782a&page=1&pageSize=6&sha=d98454d0d6f180a484e64d525b7bc6edb79fcdd3&tab=main

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -946,11 +946,7 @@ oie.email.authenticator.description = Verify with a link or code sent to your em
 oie.email.mfa.title = Verify with your email
 oie.email.challenge.mfa.title = Get a verification email
 oie.email.verify.primaryButton = Send me an email
-# {0} is a user's email address
-oie.email.verify.subtitle.text.with.email = Send a verification email to <$1>{0}</$1> by clicking on "Send me an email".
 oie.email.verify.subtitle.text.without.email = Send a verification email by clicking on "Send me an email".
-# {0} is a user's email address
-oie.email.verify.alternate.magicLinkToEmailAddress = We sent an email to <$1>{0}</$1>. 
 oie.email.verify.alternate.magicLinkToYourEmail = We sent you a verification email. 
 oie.email.verify.alternate.instructions = Click the verification link in your email to continue or enter the code below.
 oie.email.verify.alternate.show.verificationCode.text = Enter a verification code instead
--- a/src/v2/view-builder/views/email/AuthenticatorEmailViewUtil.js
+++ b/src/v2/view-builder/views/email/AuthenticatorEmailViewUtil.js
@@ -8,19 +8,10 @@ const CheckYourEmailTitle = View.extend({
     'data-se': 'o-form-explain',
   },
   template: hbs`
-      {{#if email}}
-        {{i18n 
-            code="oie.email.verify.alternate.magicLinkToEmailAddress" 
-            bundle="login" 
-            arguments="email" 
-            $1="<span class='strong'>$1</span>"
-        }}
-      {{else}}
-        {{i18n 
-          code="oie.email.verify.alternate.magicLinkToYourEmail" 
-          bundle="login"
-        }}
-      {{/if}}
+      {{i18n 
+        code="oie.email.verify.alternate.magicLinkToYourEmail" 
+        bundle="login"
+      }}
       
       {{#if useEmailMagicLinkValue}}
         {{i18n 
--- a/src/v2/view-builder/views/email/ChallengeAuthenticatorDataEmailView.js
+++ b/src/v2/view-builder/views/email/ChallengeAuthenticatorDataEmailView.js
@@ -7,19 +7,10 @@ const BaseAuthenticatorEmailForm = BaseAuthenticatorView.prototype.Body;
 const SubtitleView = View.extend({
   template: hbs`
     <div class="okta-form-subtitle" data-se="o-form-explain">
-      {{#if email}}
-        {{i18n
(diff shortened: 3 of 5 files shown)


justinabrokwah-okta (commented):
LGTM! If possible can you add some screenshots or a video to verify the change please?

hanyuanzhang-okta:
> LGTM! If possible can you add some screenshots or a video to verify the change please?
Thanks for the quick review (still a draft PR :D). I was gonna ask SIW folks to review it and add video after my local is up (I wanted to wait until the known SIW rendering issue is merged).

lokeshhemrajani-okta (commented):
1. If `oie.email.verify.alternate.magicLinkToEmailAddress` and `oie.email.verify.subtitle.text.with.email` is not being used anywhere, then lets clean up from properties as well
2. I see `test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js` fixes the tests but I there are already other test case which are covering without email in the message. So, I would suggest cleanup tests rather than fixing it to avoid duplication.

hanyuanzhang-okta:
> 1. If `oie.email.verify.alternate.magicLinkToEmailAddress` and `oie.email.verify.subtitle.text.with.email` is not being used anywhere, then lets clean up from properties as well
> 2. I see `test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js` fixes the tests but there are already other test case which are covering without email in the message. So, I would suggest cleanup tests rather than fixing it to avoid duplication.

I will remove them from properties file, but I do not think the tests needs clean up.
It is true that multiple ones are asserting on the property, but they are testing different scenarios. For example, 3 tests in `ChallengeAuthenticatorEmail_spec` are asserting on `Send a verification email by clicking on "Send me an email"`: one is the normal path (has valid profile), the other two covers scenarios of empty profile and no profile.

justinabrokwah-okta (approved):
Nice work! 🚀

lokeshhemrajani-okta:
> > 1. If `oie.email.verify.alternate.magicLinkToEmailAddress` and `oie.email.verify.subtitle.text.with.email` is not being used anywhere, then lets clean up from properties as well
> > 2. I see `test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js` fixes the tests but there are already other test case which are covering without email in the message. So, I would suggest cleanup tests rather than fixing it to avoid duplication.
> 
> @lokeshhemrajani-okta I will remove them from properties file, but I do not think the tests needs clean up. It is true that multiple ones are asserting on the property, but they are testing different scenarios. For example, 3 tests in `ChallengeAuthenticatorEmail_spec` are asserting on `Send a verification email by clicking on "Send me an email"`: one is the normal path (has valid profile), the other two covers scenarios of empty profile and no profile.

Post your changes. `test.requestHooks(validOTPmock)('challenge email authenticator screen has right labels',` is same as `test.requestHooks(validOTPmockNoProfile)('challenge email authenticator screen has right labels when profile is null'`. So, you kind of dont need both of them because irrespective of profile present or not as we are not using anything from it now, you're displaying 1 message only. This applies to others as well.

hanyuanzhang-okta:
> 

It is true that the value in profile is no longer used, but they are still different scenarios (user identified vs user not identified), I think we should keep them.

hanyuanzhang-okta:
> > > 1. If `oie.email.verify.alternate.magicLinkToEmailAddress` and `oie.email.verify.subtitle.text.with.email` is not being used anywhere, then lets clean up from properties as well
> > > 2. I see `test/testcafe/spec/ChallengeAuthenticatorEmail_spec.js` fixes the tests but there are already other test case which are covering without email in the message. So, I would suggest cleanup tests rather than fixing it to avoid duplication.
> > 
> > 
> > @lokeshhemrajani-okta I will remove them from properties file, but I do not think the tests needs clean up. It is true that multiple ones are asserting on the property, but they are testing different scenarios. For example, 3 tests in `ChallengeAuthenticatorEmail_spec` are asserting on `Send a verification email by clicking on "Send me an email"`: one is the normal path (has valid profile), the other two covers scenarios of empty profile and no profile.
> 
> Post your changes. `test.requestHooks(validOTPmock)('challenge email authenticator screen has right labels',` is same as `test.requestHooks(validOTPmockNoProfile)('challenge email authenticator screen has right labels when profile is null'`. So, you kind of dont need both of them because irrespective of profile present or not as we are not using anything from it now, you're displaying 1 message only. This applies to others as well.

It is true that the value in profile is no longer used, but they are still different scenarios (user identified vs user not identified), I think we should keep them.

jaredperreault-okta (approved):

hanyuanzhang-okta:
Expected behavior updated, should fix this in backend. Declining this PR.

PR #2923, opened 2022-11-15
OKTA-509021 - Enroll password authenticator v3 parity spec

vijetmahabaleshwar-okta:
## Description:
- This PR adds v3 parity to the Enroll Password Authenticator Testcafe spec

## PR Checklist

### Issue:

- [OKTA-509021](https://oktainc.atlassian.net/browse/OKTA-509021)

### DIFF
--- a/playground/mocks/data/idp/idx/authenticator-enroll-password.json
+++ b/playground/mocks/data/idp/idx/authenticator-enroll-password.json
@@ -21,7 +21,8 @@
                 {
                   "name": "passcode",
                   "label": "Enter password",
-                  "secret": true
+                  "secret": true,
+                  "required": true
                 }
               ]
             }
--- a/playground/mocks/data/idp/idx/authenticator-enroll-password.json
+++ b/playground/mocks/data/idp/idx/authenticator-enroll-password.json
@@ -21,11 +21,11 @@
                 {
                   "name": "passcode",
                   "label": "Enter password",
-                  "secret": true,
-                  "required": true
+                  "secret": true
                 }
               ]
-            }
+            },
+            "required": true
           },
           {
             "name": "stateHandle",
--- a/test/testcafe/framework/page-objects/components/BaseFormObject.js
+++ b/test/testcafe/framework/page-objects/components/BaseFormObject.js
@@ -176,10 +176,11 @@ export default class BaseFormObject {
     await this.t.click(this.el.find(CANCEL_BUTTON_SELECTOR));
   }
 
-  getSaveButtonLabel() {
+  getSaveButtonLabel(name = 'Next') {
     // in v3 buttons dont have a value prop
     if (userVariables.v3) {
-      return within(this.el).getByRole('button').textContent;
+      const submitButton = this.getButton(name);
+      return submitButton.textContent;
     }
 
     return within(this.el).getByRole('button').value;
--- a/test/testcafe/framework/page-objects/components/BaseFormObject.js
+++ b/test/testcafe/framework/page-objects/components/BaseFormObject.js
@@ -5,11 +5,11 @@ const TERMINAL_CONTENT = '.o-form-error-container .ion-messages-container';
 const FORM_INFOBOX_ERROR = '[data-se="o-form-error-container"] .infobox-error';
 
 const CANCEL_BUTTON_SELECTOR = '[data-type="cancel"]';
+const SAVE_BUTTON_SELECTOR = '[data-type="save"]';
 
 const focusOnSubmitButton = () => {
   // client function is not able to refer any variables defined outside
-  const submitButton = '[data-type="save"]';
-  document.querySelector(submitButton).focus();
+  document.querySelector('[data-type="save"]').focus();
 };
 
 export default class BaseFormObject {
@@ -167,7 +167,7 @@ export default class BaseFormObject {
    * Clicks the button with type=save, regardless of the button value
    */
   async clickSaveButtonAsInput() {
-    const buttonToClick = this.el.find('[data-type="save"]');
+    const buttonToClick = this.el.find(SAVE_BUTTON_SELECTOR);
 
     await this.t.click(buttonToClick);
   }
@@ -176,11 +176,10 @@ export default class BaseFormObject {
     await this.t.click(this.el.find(CANCEL_BUTTON_SELECTOR));
   }
 
-  getSaveButtonLabel(name = 'Next') {
+  getSaveButtonLabel() {
     // in v3 buttons dont have a value prop
     if (userVariables.v3) {
-      const submitButton = this.getButton(name);
-      return submitButton.textContent;
+      return this.el.find(SAVE_BUTTON_SELECTOR).textContent;
     }
 
     return within(this.el).getByRole('button').value;
--- a/test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
+++ b/test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
@@ -1,4 +1,4 @@
-import { RequestMock } from 'testcafe';
+import { RequestMock, Selector } from 'testcafe';
 import FactorEnrollPasswordPageObject from '../framework/page-objects/FactorEnrollPasswordPageObject';
 import SuccessPageObject from '../framework/page-objects/SuccessPageObject';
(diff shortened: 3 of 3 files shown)


test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
```diff
@@ -98,8 +98,8 @@ test.requestHooks(successMock)('should have the correct reqiurements', async t =
   await t.expect(enrollPasswordPage.getRequirements()).contains('A symbol');
   await t.expect(enrollPasswordPage.getRequirements()).contains('Does not include your first name');
   await t.expect(enrollPasswordPage.getRequirements()).contains('Does not include your last name');
-  await t.expect(enrollPasswordPage.getRequirements()).contains('At least 2 hour(s) must have elapsed since you last changed your password');
+  // await t.expect(enrollPasswordPage.getRequirements()).contains('At least 2 hour(s) must have elapsed since you last changed your password');
```
vijetmahabaleshwar-okta, line 101: These 2 messages aren't available in the v3 widget for password authenticator enrollment page.
I'm not sure if these messages are from server through the response or rendered by the widget.

test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
```diff
@@ -59,7 +61,6 @@ test.requestHooks(successMock)('should have both password and confirmPassword fi
   await enrollPasswordPage.waitForErrorBox();
   await t.expect(enrollPasswordPage.hasPasswordError()).eql(false);
   await t.expect(enrollPasswordPage.getConfirmPasswordError()).eql('New passwords must match');
-  await t.expect(enrollPasswordPage.getErrorBoxText()).eql('We found some errors. Please review the form and make corrections.');
```
vijetmahabaleshwar-okta, line 62: This was a redundant assertion.
`await enrollPasswordPage.waitForErrorBox();` does the same check

test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
glenfannin-okta: In v3, UX & Product actually made a decision to _not_ display the server side messages because they can be sort of confusing. There is another ticket pending final work from UX that will re-introduce these server side messages but in a different context.

For now, I would actually recommend wrapping this in a condition (to not impact v2 tests but exclude it for v3), this is one of the rare times that we include conditional statements in the test, you can mimic what we did in this test: https://github.com/okta/okta-signin-widget/blob/0.1/test/testcafe/spec/ReEnrollAuthenticatorWarningPasswordView_spec.js#L84-L89

test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
```diff
@@ -98,8 +98,8 @@ test.requestHooks(successMock)('should have the correct reqiurements', async t =
   await t.expect(enrollPasswordPage.getRequirements()).contains('A symbol');
   await t.expect(enrollPasswordPage.getRequirements()).contains('Does not include your first name');
   await t.expect(enrollPasswordPage.getRequirements()).contains('Does not include your last name');
-  await t.expect(enrollPasswordPage.getRequirements()).contains('At least 2 hour(s) must have elapsed since you last changed your password');
+  // await t.expect(enrollPasswordPage.getRequirements()).contains('At least 2 hour(s) must have elapsed since you last changed your password');
   await t.expect(enrollPasswordPage.getRequirements()).contains('No parts of your username');
-  await t.expect(enrollPasswordPage.getRequirements()).contains('Your password cannot be any of your last 4 passwords');
+  // await t.expect(enrollPasswordPage.getRequirements()).contains('Your password cannot be any of your last 4 passwords');
```
glenfannin-okta, line 103: You can group this and the one above in the same condition.

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -176,10 +176,11 @@ export default class BaseFormObject {
     await this.t.click(this.el.find(CANCEL_BUTTON_SELECTOR));
   }
 
-  getSaveButtonLabel() {
+  getSaveButtonLabel(name = 'Next') {
     // in v3 buttons dont have a value prop
     if (userVariables.v3) {
-      return within(this.el).getByRole('button').textContent;
+      const submitButton = this.getButton(name);
+      return submitButton.textContent;
```
glenfannin-okta, line 183: Is this change needed? Since there should only be one button on the page and that button should be the "Next" button?

playground/mocks/data/idp/idx/authenticator-enroll-password.json
```diff
@@ -21,7 +21,8 @@
                 {
                   "name": "passcode",
                   "label": "Enter password",
-                  "secret": true
+                  "secret": true,
+                  "required": true
```
shawyu-okta, line 25: Can you double-check to ensure the `required` field appears here rather than a level up between line 18 and 19? I have seen it appear at both levels.

test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
shawyu-okta: Do we have a ticket tracking the re-introduction of this message with new UX? If so, we should include it in the comment explaining why we are skipping this assertion in v3.

test/testcafe/framework/page-objects/components/BaseFormObject.js
shawyu-okta: I do think in some situations there is more than one button on the page -- is that why you added the query for a specific button by its name @vijetmahabaleshwar-okta?

playground/mocks/data/idp/idx/authenticator-enroll-password.json
glenfannin-okta: Oh yea, looks like it should technically be at the `credentials` level. 

image

test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
glenfannin-okta: Yep! We have that ticket documented in the other test too. `OKTA-533383`

If you want to copy the exact comment that the link above has I think that is totally acceptable since they are pretty much equivalent.

test/testcafe/framework/page-objects/components/BaseFormObject.js
vijetmahabaleshwar-okta: That's right @shawyu-okta. 
On the password enrollment page, the "show/hide" password is a button. And the test fails since there are multiple buttons that match the selector. Had to do this change to get over that issue. @glenfannin-okta

vijetmahabaleshwar-okta:
Thanks for the comments, @shawyu-okta and @glenfannin-okta.
I have addressed your concerns in the new commit. Please review when possible.

glenfannin-okta (approved):

test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
```diff
@@ -18,11 +18,13 @@ const errorMock = RequestMock()
...
   .respond(xhrAuthenticatorEnrollPasswordError, 403);
 
-fixture('Authenticator Enroll Password');
+fixture('Authenticator Enroll Password')
+  .meta('v3', true);
 
 async function setup(t) {
   const enrollPasswordPage = new FactorEnrollPasswordPageObject(t);
   await enrollPasswordPage.navigateToPage();
+  await t.expect(Selector('form').exists).eql(true);
```
lesterchoi-okta, line 27: I think we do this a lot for v3. Feel free to move it into a base object if/when it's re-used.

lesterchoi-okta (approved):

test/testcafe/spec/EnrollAuthenticatorPasswordView_spec.js
shawyu-okta: @leemchale-okta tried to add this at the baseform level and found some tests failed since they don't actually show the form at all. Perhaps we can look at a more nuanced inclusion (e.g. flag in the constructor or utility method for setup in the baseform)

shawyu-okta (approved):

PR #2927, opened 2022-11-17
OKTA-509004 - Challenge password authenticator v3 parity spec

vijetmahabaleshwar-okta:
## PR Checklist

### Issue:

- [OKTA-509004](https://oktainc.atlassian.net/browse/OKTA-509004)

### DIFF
--- a/test/testcafe/framework/page-objects/components/BaseFormObject.js
+++ b/test/testcafe/framework/page-objects/components/BaseFormObject.js
@@ -206,7 +206,7 @@ export default class BaseFormObject {
 
   getErrorBoxCount() {
     if (userVariables.v3) {
-      return within(this.el).getAllByRole('alert').nth(0).count;
+      return within(this.el).getAllByRole('alert').count;
     }
 
     return this.el.find(FORM_INFOBOX_ERROR).count;
--- a/test/testcafe/framework/page-objects/components/BaseFormObject.js
+++ b/test/testcafe/framework/page-objects/components/BaseFormObject.js
@@ -205,6 +205,10 @@ export default class BaseFormObject {
   }
 
   getErrorBoxCount() {
+    if (userVariables.v3) {
+      return within(this.el).getAllByRole('alert').nth(0).count;
+    }
+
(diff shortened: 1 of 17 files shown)


test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -205,6 +205,10 @@ export default class BaseFormObject {
   }
 
   getErrorBoxCount() {
+    if (userVariables.v3) {
+      return within(this.el).getAllByRole('alert').nth(0).count;
```
glenfannin-okta, line 209: My only feedback for this is that, when you use `getAllByRole(...).nth(n)` it will always only give back the first item that was found (if any), so this function (for v3) will only ever give back 0 or 1. 

Could the test be updated to search for a specific error message instead (both for v2 and v3)?

glenfannin-okta (approved):

test/testcafe/framework/page-objects/components/BaseFormObject.js
vijetmahabaleshwar-okta: In the 2 specs that this method is used, one of them verifies that the error count is 0. 
The other one verifies that the error count is 1 and then also verifies the error message like this - 
```
  await t.expect(challengePasswordPage.form.getErrorBoxCount()).eql(1);
  await t.expect(challengePasswordPage.form.getErrorBoxText())
    .eql('Reset password is not allowed at this time. Please contact support for assistance.');
```
The error box count is redundant in the above case. But no harm in keeping it I guess? Wdty?

vijetmahabaleshwar-okta: Upon further examination, instead of using `getAllByRole(...).nth(n)`, I can use `getAllByRole('alert').count`.
This should give us the desired error count.

glenfannin-okta: Ahh, ok that makes much more sense in this case. Ty!

PR #2940, opened 2022-11-29
Adds transformer for launch-authenticator button

shawyu-okta:
## PR Checklist

### Issue:

- [OKTA-548960](https://oktainc.atlassian.net/browse/OKTA-548960)

### DIFF
--- a/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
+++ b/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
@@ -12,8 +12,8 @@
 
 import { Box, Button as ButtonMui } from '@mui/material';
 import { h } from 'preact';
-import { useOnSubmit } from '../../hooks';
 
+import { useOnSubmit } from '../../hooks';
 import {
   ClickHandler,
   LaunchAuthenticatorButtonElement,
@@ -58,7 +58,7 @@ const LaunchAuthenticatorButton: UISchemaElementComponent<{
       >
         <OktaVerifyIcon
           name="ov"
-          description='foo'
+          description="foo"
           width={24}
           height={24}
         />
--- a/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
+++ b/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
@@ -53,13 +53,14 @@ const LaunchAuthenticatorButton: UISchemaElementComponent<{
         justifyContent="center"
         alignContent="space-between"
         gap="5px"
-        marginY="-7px"
+        // keep the icon from stretching the button vertically
+        marginY="-3px"
       >
         <OktaVerifyIcon
           name="ov"
           description='foo'
-          width={32}
-          height={32}
+          width={24}
+          height={24}
         />
         {label}
       </Box>
--- a/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
+++ b/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2022-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { Box, Button as ButtonMui } from '@mui/material';
+import { h } from 'preact';
+import { useOnSubmit } from '../../hooks';
+
+import {
+  ClickHandler,
+  LaunchAuthenticatorButtonElement,
+  UISchemaElementComponent,
+} from '../../types';
+import { getTranslation } from '../../util';
+import { OktaVerifyIcon } from '../Icon';
+
+const LaunchAuthenticatorButton: UISchemaElementComponent<{
+  uischema: LaunchAuthenticatorButtonElement
+}> = ({ uischema }) => {
+  const onSubmitHandler = useOnSubmit();
+  const {
+    translations = [],
+    options: {
+      step,
+    },
+  } = uischema;
+
+  const handleClick: ClickHandler = async () => {
+    onSubmitHandler({
+      step,
+    });
+  };
+
+  const label = getTranslation(translations, 'label');
+
+  return (
+    <ButtonMui
+      variant="secondary"
+      fullWidth
+      onClick={handleClick}
+    >
+      <Box
+        display="flex"
+        alignItems="center"
+        justifyContent="center"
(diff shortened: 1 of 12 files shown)


src/v3/src/transformer/button/transformLaunchAuthenticatorButton.ts
```diff
@@ -0,0 +1,55 @@
...
+  const containsLaunchAuthenticator = remediations.some(
+    (remediation) => remediation.name === IDX_STEP.LAUNCH_AUTHENTICATOR,
+  );
+  if (!containsLaunchAuthenticator) {
+    return formBag;
+  }
+
+  const launchAuthenticatorButton: ButtonElement = {
+    type: 'Button',
+    label: loc('oktaVerify.button', 'login'),
```
shawyu-okta, line 34: I'm considering creating a new button just for this button since it's also used in the other two `launch-authenticator` views with varying behaviors. The impetus here is that the `Button` will need the Icon but no loading state, and the existing button is heavily laden with logic already, and I would have to decide how the button should behave if `loading=true` AND there is an Icon set. I think separating out into a new component (`LaunchAuthenticatorButton` perhaps) would sidestep this issue and also simplify the component.

Any thoughts?

src/v3/src/transformer/button/transformLaunchAuthenticatorButton.ts
```diff
@@ -0,0 +1,55 @@
...
+    options: {
+      type: ButtonType.BUTTON,
+      step: IDX_STEP.LAUNCH_AUTHENTICATOR,
+      variant: 'secondary',
```
shawyu-okta, line 38: We're going to have to figure out `onClick` behavior for this (or the separated component) so that it is able to both do a custom onClick behavior AND also still call the default onClick behavior (submitting the `action` in the `transaction`)

shawyu-okta: This might also be a lot easier with the separate, specialized button (see comment above)

src/v3/src/transformer/button/transformLaunchAuthenticatorButton.ts
```diff
@@ -0,0 +1,55 @@
...
+    type: 'Divider',
+    options: { text: loc('socialauth.divider.text', 'login') },
+  };
+
+  const titleIndex = formBag.uischema.elements.findIndex((element) => element.type === 'Title');
+  const pivButtonPos = titleIndex !== -1 ? titleIndex + 1 : 0;
+
+  // Add button after title (if exists) otherwise add to top of array
+  formBag.uischema.elements.splice(pivButtonPos, 0, launchAuthenticatorButton);
+  formBag.uischema.elements.splice(pivButtonPos + 1, 0, dividerElement);
```
glenfannin-okta, line 52: Can these be combined to utilize splice's item1, item2, itemN param? :D

glenfannin-okta (approved):

src/v3/src/transformer/button/transformLaunchAuthenticatorButton.ts
shawyu-okta: Haha, yes definitely 😄

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+        gap="5px"
+        // keep the icon from stretching the button vertically
+        marginTop="-4px"
+        marginBottom="-5px"
+      >
+        <OktaVerifyIcon
+          name="mfa-okta-verify"
+          description={iconDescription}
+          width={theme.spacing(5)}
+          height={theme.spacing(5)}
```
shawyu-okta, line 67: FYI probably going to drop these theme values and use explicit values for now

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+        display="flex"
+        alignItems="center"
+        justifyContent="center"
+        alignContent="space-between"
+        gap="5px"
+        // keep the icon from stretching the button vertically
+        marginTop="-4px"
+        marginBottom="-5px"
+      >
+        <OktaVerifyIcon
```
glenfannin-okta, line 63: Is it possible to use MUI's `startIcon` prop for icon placement on the button?

```
<ButtonMui
      variant="secondary"
      fullWidth
      startIcon={<DeleteIcon />}
      onClick={handleClick}
    >
    ...
```
See: https://mui.com/material-ui/react-button/#buttons-with-icons-and-label

glenfannin-okta: This may also alleviate the need for the `<Box` component within the button since MUI controls the icon placement.

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+    onSubmitHandler({
+      step,
+    });
+  };
+
+  const label = getTranslation(translations, 'label');
+  const iconDescription = getTranslation(translations, 'icon-description') || '';
+
+  return (
+    <ButtonMui
```
glenfannin-okta, line 48: Do you need to use the `loading` flag here so when the request is pending, the spinner shows on the button? 

If so, will also need to add the logic to the button to display the spinner (or at least disable the button) when loading = true.

glenfannin-okta (approved):

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
shawyu-okta: There isn't a spinner in the current gen2 widget, however if we do want a spinner for this newer widget we might need design input that we can implement in a separate PR. The reason is that we need to work out how to display the spinner in the loading state (i.e. do we show the spinner and the icon, hide the icon when spinner is there, or something else?)

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
shawyu-okta: I did try using the `startIcon` prop but it didn't layout the button properly (placed the icon and text on two lines). It looks like this is something to do with the Odyssey theme, since the issue isn't present if I remove the theme entirely. I will followup with Odyssey to see if this is a bug, but in the meantime will leave this as-is until we have a resolution.

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
glenfannin-okta: Ohh yes, that makes total sense, does v2 disable the button after clicking? Or is there no harm in double clicks?

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
glenfannin-okta: Makes total sense! Ty!

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
shawyu-okta: v2 does not disable the button after clicking, although we may actually want to do that. I don't think there's a ton of harm in double clicking however.

PR #2947, opened 2022-12-01
Enroll security question authenticator v3 parity

vijetmahabaleshwar-okta:
## PR Checklist

### Issue:

- [OKTA-509014](https://oktainc.atlassian.net/browse/OKTA-509014)

### DIFF
--- a/test/testcafe/framework/page-objects/EnrollSecurityQuestionPageObject.js
+++ b/test/testcafe/framework/page-objects/EnrollSecurityQuestionPageObject.js
@@ -1,3 +1,4 @@
+import { userVariables } from 'testcafe';
 import BasePageObject from './BasePageObject';
 
 const ANSWER_FIELD = 'credentials.answer';
@@ -10,14 +11,25 @@ export default class EnrollSecurityQuestionPageObject extends BasePageObject {
   }
 
   clickChooseSecurityQuestion() {
+    if (userVariables.v3) {
+      return this.form.selectRadioButtonOptionByValue('','predefined');
+    }
     return this.form.selectRadioButtonOption('sub_schema_local_credentials', 0);
   }
 
   clickCreateYouOwnQuestion() {
+    if (userVariables.v3) {
+      return this.form.selectRadioButtonOptionByValue('', 'custom');
+    }
     return this.form.selectRadioButtonOption('sub_schema_local_credentials', 1);
   }
 
   selectSecurityQuestion(index) {
+    // In v3 widget, "Choose a security question" is also a value in dropdown
+    // We need to skip that value, by increasing the index by 1
+    if (userVariables.v3) {
+      index = index + 1;
+    }
     return this.form.selectValueChozenDropdown(QUESTION_KEY_FIELD, index);
   }
 
@@ -42,7 +54,7 @@ export default class EnrollSecurityQuestionPageObject extends BasePageObject {
   }
 
   clickVerifyButton() {
-    return this.form.clickSaveButton();
+    return this.form.clickButton('Verify');
   }
 
 }
--- a/test/testcafe/framework/page-objects/components/BaseFormObject.js
+++ b/test/testcafe/framework/page-objects/components/BaseFormObject.js
@@ -337,6 +337,19 @@ export default class BaseFormObject {
   }
 
   async selectRadioButtonOptionByValue(fieldName, value) {
+    if (userVariables.v3) {
+      const radioOption = await this.el.find(`input[type="radio"][value="${value}"]`);
+      await this.t.click(radioOption); 
+
+      const label = this.el
+        .find(`input[type="radio"][value="${value}"]`)
+        .parent('span')
+        .parent('label')
+        .textContent;
+
+      return label; 
+    }
+    
     const radioOption = await this.findFormFieldInput(fieldName)
       .find(`input[value="${value}"] + .radio-label`);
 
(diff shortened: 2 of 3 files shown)


test/testcafe/framework/page-objects/EnrollSecurityQuestionPageObject.js
```diff
@@ -10,14 +11,25 @@ export default class EnrollSecurityQuestionPageObject extends BasePageObject {
   }
 
   clickChooseSecurityQuestion() {
+    if (userVariables.v3) {
+      return this.form.selectRadioButtonOptionByValue('','predefined');
+    }
     return this.form.selectRadioButtonOption('sub_schema_local_credentials', 0);
```
vijetmahabaleshwar-okta, line 17: Is there a better way to do this? 
In v3, we don't have a field name for radio options.
In v2, field name is used like this - `this.el.find(`[data-se="${fieldName}"]`);`
In v3, we can use just the value field (2nd param) - `this.el.find(`input[type="radio"][value="${value}"]`);`

I don't want to create another method that just takes 1 argument, since methods with same names are not allowed according to lint rules.

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -337,6 +337,19 @@ export default class BaseFormObject {
   }
 
   async selectRadioButtonOptionByValue(fieldName, value) {
+    if (userVariables.v3) {
+      const radioOption = await this.el.find(`input[type="radio"][value="${value}"]`);
+      await this.t.click(radioOption); 
+
```
vijetmahabaleshwar-okta, line 343: Getting the label text associated with a radio button is not very clean.
This is how it looks in the DOM - 
Screen Shot 2022-12-01 at 3 06 04 PM

Getting parent of parent works but is there a better way to do this?

test/testcafe/framework/page-objects/EnrollSecurityQuestionPageObject.js
```diff
@@ -10,14 +11,25 @@ export default class EnrollSecurityQuestionPageObject extends BasePageObject {
   }
 
   clickChooseSecurityQuestion() {
+    if (userVariables.v3) {
```
vijetmahabaleshwar-okta, line 14: Is there a better way to do this? i.e. not pass a dummy `` value as 1st argument.
In v3, we don't have a field name for radio options.
In v2, field name is used like this - this.el.find([data-se="${fieldName}"]`);`
In v3, we can use just the value field (2nd param) - `this.el.find(`input[type="radio"][value="${value}"]`);`

I don't want to create another method that just takes 1 argument, since methods with same names are not allowed according to lint rules.

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -337,6 +337,19 @@ export default class BaseFormObject {
...
+      const label = this.el
+        .find(`input[type="radio"][value="${value}"]`)
+        .parent('span')
+        .parent('label')
+        .textContent;
+
+      return label; 
+    }
```
vijetmahabaleshwar-okta, line 351: Is there a better way to do this? i.e. not pass a dummy `` value as 1st argument to selectRadioButtonOptionByValue
In v3, we don't have a field name for radio options.
In v2, field name is used like this -
`this.el.find([data-se="${fieldName}"]`);`

In v3, we can use just the value field (2nd param) -
`this.el.find(`input[type="radio"][value="${value}"]`);`

I don't want to create another method that just takes 1 argument, since methods with same names are not allowed according to lint rules.

test/testcafe/framework/page-objects/EnrollSecurityQuestionPageObject.js
```diff
@@ -10,14 +11,25 @@ export default class EnrollSecurityQuestionPageObject extends BasePageObject {
...
+      return this.form.selectRadioButtonOptionByValue('','predefined');
+    }
```
vijetmahabaleshwar-okta, line 16: Is there a better way to do this? i.e. not pass a dummy `` value as 1st argument to `selectRadioButtonOptionByValue`
In v3, we don't have a field name for radio options.
In v2, field name is used like this - 
```this.el.find([data-se="${fieldName}"]`);```

In v3, we can use just the value field (2nd param) - 
```this.el.find(`input[type="radio"][value="${value}"]`);```

I don't want to create another method that just takes 1 argument, since methods with same names are not allowed according to lint rules.

Comment (glenfannin-okta) [reply test/testcafe/framework/page-objects/EnrollSecurityQuestionPageObject.js:16]: If we added a name to the radio button, would you then be able to use this function that exists? `selectRadioButtonOption`

Comment (glenfannin-okta) [reply test/testcafe/framework/page-objects/EnrollSecurityQuestionPageObject.js:16]: If so, we can move forward with this and I can make the update in a separate PR to add the field name to the radio button.

File: test/testcafe/framework/page-objects/EnrollSecurityQuestionPageObject.js [hunk 6]
```diff
@@ -10,14 +11,25 @@ export default class EnrollSecurityQuestionPageObject extends BasePageObject {
...
     return this.form.selectRadioButtonOption('sub_schema_local_credentials', 1);
   }
 
   selectSecurityQuestion(index) {
+    // In v3 widget, "Choose a security question" is also a value in dropdown
+    // We need to skip that value, by increasing the index by 1
+    if (userVariables.v3) {
+      index = index + 1;
+    }
```
glenfannin-okta, line 32: I am a little confused by this, does v2 not also have the same options in the security question select box?

glenfannin-okta (approved):
If you want me to add the field name to the radio button, let me know and I can create a small ticket

PR #2951, opened 2022-12-02
OKTA-545802 : Dynamically adding aria-describedby values to interactive elements

glenfannin-okta:
## Description:

The purpose of this PR is to build and apply aria-describedby values to interactive elements to enable screen readers to read out descriptive text on the page to visually impaired users.

## PR Checklist

### Issue:

- [OKTA-545802](https://oktainc.atlassian.net/browse/OKTA-545802)


src/v3/src/transformer/layout/email/emailChallengeTransformer.ts
```diff
@@ -127,11 +128,13 @@ export const transformEmailChallenge: IdxStepTransformer = ({ transaction, formB
           titleElement,
           informationalText,
           showCodeStepperButton,
-        ],
+        ].map((ele: UISchemaElement) => ({ ...ele, viewIndex: 0 })),
```
glenfannin-okta, line 131: This adds a `viewIndex` value to the element so we can determine which view we are on when adding describedby values.

src/v3/src/components/InformationalText/InformationalText.tsx
```diff
@@ -28,6 +28,7 @@ const InformationalText: UISchemaElementComponent<{
       justifyContent="flex-start"
     >
       <Typography
+        id={uischema.id}
```
shawyu-okta, line 31: I'm a little bit concerned that users will start to rely on these `id`s to hook into the DOM. Is this `id` dynamic (as in has some randomness to it) or are they deterministic?

src/v3/src/util/idxUtils.ts
```diff
@@ -145,8 +146,12 @@ export const areTransactionsEqual = (
...
 
-  const tx1AuthId = tx1?.context.currentAuthenticator?.value?.id;
-  const tx2AuthId = tx2?.context.currentAuthenticator?.value?.id;
+  const tx1AuthId = typeof tx1 !== 'undefined'
+    ? getCurrentAuthenticator(tx1)?.value?.id
+    : undefined;
+  const tx2AuthId = typeof tx2 !== 'undefined'
+    ? getCurrentAuthenticator(tx2)?.value?.id
+    : undefined;
   if (tx1AuthId !== tx2AuthId) {
```
shawyu-okta, line 155: `undefined !== undefined` evaluates to `false` just making sure this is OK

src/v3/src/transformer/uischema/addIdToElement.ts
```diff
@@ -0,0 +1,41 @@
...
+      elementCount += 1;
+      let elementId = element.key
+        ? `${name}_${element.type}_${element.key}`
+        : `${name}_${element.type}`;
+      elementId += [
+        authId,
+        (typeof element.viewIndex !== 'undefined' ? (element.viewIndex + 1) : undefined),
+      ].filter(Boolean).map((str) => `_${str}`).join('');
+      // eslint-disable-next-line no-param-reassign
+      element.id = elementIdSet.has(elementId) ? `${elementId}_${elementCount}` : elementId;
```
shawyu-okta, line 36: Is there any downside to always appending the `elementCount` for simplicity? Perhaps it's nicer to not add it unnecessarily?

src/v3/src/transformer/uischema/buildAriaDescribedByValues.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  traverseLayout({
+    layout: formbag.uischema,
+    predicate: (el) => isInteractiveType(el.type),
+    callback: (el) => {
+      const descriptiveElementIds = descriptionElements
+        .filter(
+          (descrEle: UISchemaElement) => typeof descrEle.viewIndex === 'undefined'
+            || descrEle.viewIndex === el.viewIndex,
```
shawyu-okta, line 34: I think I understand why we have `viewIndex` but can you give a short explanation?

src/v3/src/util/idxUtils.ts
glenfannin-okta: Yessir, this is expected because if they are both `undefined` it will ultimately fall to the `return true` block. Each condition checks for a difference between the values, if that condition results in `true` it will return false for the overall function.

src/v3/src/transformer/uischema/addIdToElement.ts
glenfannin-okta: I updated this to always include the `elementCount`, it doesnt hurt anything and actually practically ensures that the ID will be unique.

src/v3/src/transformer/uischema/buildAriaDescribedByValues.ts
glenfannin-okta: I added the explanation as a ts doc to the variable in the interface. Is this ok or should I move it directly here? I chose the ts doc incase someone uses it in the future, they can quickly see the purpose of the variable. 
image

shawyu-okta (approved):

PR #2955, opened 2022-12-06
OKTA-552792 : Yubikey field data carryover fix

glenfannin-okta:
## Description:

The purpose of this PR is to prevent the field data carry over from one flow (Yubikey verification) to another (Challenge-authenticator w/ credentials.passcode field) by ensuring the React Element keys are unique between flows.

## PR Checklist

### Issue:

- [OKTA-552792](https://oktainc.atlassian.net/browse/OKTA-552792)

### DIFF
--- a/src/v3/src/mocks/response/idp/idx/identify/error-invalid-username.json
+++ b/src/v3/src/mocks/response/idp/idx/identify/error-invalid-username.json
@@ -125,7 +125,7 @@
       },
       "type": "password",
       "key": "okta_password",
-      "id": "aut2h3ffszv3me6O31d7",
+      "id": "aut1egri5t7iomJ3l1d7",
       "displayName": "Password",
       "methods": [{ "type": "password" }]
     }
--- a/src/v3/src/transformer/uischema/addKeyToElement.ts
+++ b/src/v3/src/transformer/uischema/addKeyToElement.ts
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2022-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import {
+  TransformStepFnWithOptions,
+} from '../../types';
+import {
+  generateRandomString,
+  getAuthenticatorKey,
+  getCurrentAuthenticator,
+} from '../../util';
+import { traverseLayout } from '../util';
+
+export const addKeyToElement: TransformStepFnWithOptions = ({ transaction }) => (formbag) => {
+  traverseLayout({
+    layout: formbag.uischema,
+    predicate: (element) => !!element.type,
+    callback: (element) => {
+      const { nextStep: { name } = {} } = transaction;
+      const authKey = getAuthenticatorKey(transaction);
+      const authId = getCurrentAuthenticator(transaction)?.value?.id;
+
+      // We need Reminder Elements to unmount from view when new transaction
+      // is set, this prevents this alert and error alerts from both displaying
+      // at the same time
+      if (element.type === 'Reminder') {
+        // eslint-disable-next-line no-param-reassign
+        element.key = `${name}_${generateRandomString()}`;
+      } else {
+        let elementKey = element.key
+          ? `${name}_${element.type}_${element.key}`
+          : `${name}_${element.type}`;
+        elementKey += [
+          authKey,
+          authId,
+        ].filter(Boolean).map((str) => `_${str}`).join('');
+        // eslint-disable-next-line no-param-reassign
+        element.key = elementKey;
+      }
+    },
+  });
+  return formbag;
+};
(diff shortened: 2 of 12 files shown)


src/v3/src/mocks/response/idp/idx/identify/error-invalid-username.json
```diff
@@ -125,7 +125,7 @@
       },
       "type": "password",
       "key": "okta_password",
-      "id": "aut2h3ffszv3me6O31d7",
+      "id": "aut1egri5t7iomJ3l1d7",
```
glenfannin-okta, line 128: Had to update this id to match the original response to pass flow scenarios (IDs do not change between flows).

src/v3/src/transformer/uischema/addKeyToElement.ts
```diff
@@ -0,0 +1,52 @@
...
+        // eslint-disable-next-line no-param-reassign
+        element.key = `${name}_${generateRandomString()}`;
+      } else {
+        let elementKey = element.key
+          ? `${name}_${element.type}_${element.key}`
+          : `${name}_${element.type}`;
+        elementKey += [
+          authKey,
+          authId,
+        ].filter(Boolean).map((str) => `_${str}`).join('');
```
glenfannin-okta, line 45: Similar to polling, auth flows can be determined to be unique based on their authenticator key and authenticator ID as you will not be presented with the same authenticator ID unless there was an error.

src/v3/src/transformer/uischema/createTextElementKeys.ts
```diff
@@ -0,0 +1,38 @@
...
+} from '../../types';
+import { traverseLayout } from '../util';
+
+const textElementTypes = new Set([
+  'Title',
+  'Heading',
+  'Description',
+]);
+
+export const createTextElementKeys: TransformStepFn = (formbag) => {
```
glenfannin-okta, line 27: I do this only on text type elements because if more than one is present on the page at once, using the text content apart of the key ensures its uniqueness.

Unless someone knows of a better way to do this without some external package.

src/v3/src/transformer/uischema/updateElementKeys.ts
```diff
@@ -0,0 +1,52 @@
...
+    predicate: (element) => !!element.type,
+    callback: (element) => {
+      const { nextStep: { name } = {} } = transaction;
+      const authKey = getAuthenticatorKey(transaction);
+      const authId = getCurrentAuthenticator(transaction)?.value?.id;
+
+      // We need Reminder Elements to unmount from view when new transaction
+      // is set, this prevents this alert and error alerts from both displaying
+      // at the same time
+      if (element.type === 'Reminder') {
```
shawyu-okta, line 35: Is there a constant that already exists for `Reminder`?

src/v3/src/transformer/uischema/updateElementKeys.test.ts
```diff
@@ -0,0 +1,246 @@
...
+    const randomString = '0987654321abc';
+    jest.spyOn(randomStringUtil, 'generateRandomString').mockReturnValue(randomString);
+    formBag.uischema.elements.unshift({ type: 'Reminder', options: { content: 'See errors below' } } as ReminderElement);
+
+    const updatedFormBag = updateElementKeys({ transaction, widgetProps, step: '' })(formBag);
+    const { elements } = <{ elements: UISchemaElement[] }>updatedFormBag.uischema;
+
+    expect(updatedFormBag).toMatchSnapshot();
+    expect(elements[0].key)
+      .toBe(`${IDX_STEP.CHALLENGE_AUTHENTICATOR}_${randomString}`);
```
shawyu-okta, line 244: What do you think about inlining string literals here in the tests instead of recreating the concatenation logic?

src/v3/test/integration/flow-yubikey-otp-to-password-verification.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import yubikeyVerificationResponse from '@okta/mocks/data/idp/idx/authenticator-verification-yubikey.json';
+
+import identifyWithPassword from '@okta/mocks/data/idp/idx/authenticator-verification-password.json';
+import { createAuthJsPayloadArgs, setup } from './util';
+
+describe('Flow transition from YubiKey verification to Password MFA', () => {
+  it('clears field data when transition from yubikey verification to password MFA', async () => {
```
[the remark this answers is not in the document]
glenfannin-okta: Oh yes, we have a few tests like this, 

https://github.com/okta/okta-signin-widget/blob/0.1/src/v3/test/integration/flow-password-authentication-security-question-authentication.test.tsx#L18

https://github.com/okta/okta-signin-widget/blob/0.1/src/v3/test/integration/flow-transition.test.tsx

src/v3/src/transformer/uischema/updateElementKeys.test.ts
glenfannin-okta: Totally can do that, I went back and forth in my head about doing it but decided against it, but from a test readability standpoint what you suggests makes more sense. 🙏

src/v3/src/transformer/uischema/updateElementKeys.ts
shawyu-okta: Oh, I see, yeah I think the literal is fine, I was just thinking in case we _did_ have a constant already we should use it.

src/v3/src/transformer/uischema/createTextElementKeys.ts
shawyu-okta: The only other way I can think of is to generate pretty much entirely random ids as keys

glenfannin-okta: I was considering it, but I thought Lester mentioned to avoid using random string/ids for the key. I know doing this on Fields can be problematic (because the components are unmounted and remounted when the key changes) but not so sure how problematic it would be with components that just render text (maybe we still want to avoid an unnecessary re-render?)

shawyu-okta: Yeah I think this approach is fine, I was just spitballing the only other way I could think of (not necessarily that it's a better approach)

shawyu-okta (approved):
I think the only outstanding comment is about replacing assertions with string literals

PR #2972, opened 2022-12-09
OKTA-547443 : Reformatting Authenticator Buttons into a List

glenfannin-okta:
## Description:

The purpose of this PR is to make the Authenticator buttons display within a List. 

## PR Checklist

### Issue:

- [OKTA-547443](https://oktainc.atlassian.net/browse/OKTA-547443)

### DIFF
--- a/src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
+++ b/src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
@@ -61,10 +61,10 @@ const AuthenticatorButton: UISchemaElementComponent<{
   const onValidationHandler = useOnSubmitValidation();
   const focusRef = useAutoFocus<HTMLButtonElement>(focus);
   const describedByIds = [
-    `${iconName}-ctaLabel`,
+    ariaDescribedBy,
     description && `${iconName}-description`,
     usageDescription && `${iconName}-usageDescription`,
-    ariaDescribedBy,
+    `${iconName}-ctaLabel`,
   ].filter(Boolean).join(' ');
 
   const onClick: ClickHandler = async () => {
--- a/src/v3/src/transformer/i18n/transformAuthenticatorButton.ts
+++ b/src/v3/src/transformer/i18n/transformAuthenticatorButton.ts
@@ -16,6 +16,7 @@ import {
 } from '../../../../v2/ion/i18nTransformer';
 import {
   AuthenticatorButtonElement,
+  AuthenticatorButtonListElement,
   TransformStepFnWithOptions,
 } from '../../types';
 import { traverseLayout } from '../util';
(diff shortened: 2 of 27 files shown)


src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -61,10 +61,10 @@ const AuthenticatorButton: UISchemaElementComponent<{
   const onValidationHandler = useOnSubmitValidation();
   const focusRef = useAutoFocus<HTMLButtonElement>(focus);
   const describedByIds = [
-    `${iconName}-ctaLabel`,
+    ariaDescribedBy,
     description && `${iconName}-description`,
     usageDescription && `${iconName}-usageDescription`,
-    ariaDescribedBy,
+    `${iconName}-ctaLabel`,
```
glenfannin-okta, line 67: I am re-ordering these because it makes more sense when the screenreader announces the texts this way.

lesterchoi-okta: Could you give an example of a before and after?

lesterchoi-okta: Could you give an example of a before and after?

src/v3/src/transformer/selectAuthenticator/transformSelectOVMethodVerify.ts
```diff
@@ -76,7 +77,11 @@ export const transformSelectOVMethodVerify: IdxStepTransformer = ({ transaction,
       'authenticator.methodType',
       uischema.elements as UISchemaElement[],
     );
-    uischema.elements = uischema.elements.concat(buttonElements);
+    const authenticatorListElement: AuthenticatorButtonListElement = {
+      type: 'AuthenticatorButtonList',
+      options: { buttons: buttonElements },
+    };
+    uischema.elements = uischema.elements.concat([authenticatorListElement]);
```
shawyu-okta, line 84: Is there any reason we want to use concat vs push? Only asking for the sake of consistency

src/v3/test/integration/__snapshots__/authenticator-apple-sso-extension.test.tsx.snap
```diff
@@ -41,7 +41,7 @@ exports[`authenticator-apple-sso-extension calls verify then cancels the transac
                     class="MuiBox-root emotion-7"
                   >
                     <h2
-                      class="MuiTypography-root MuiTypography-h3 emotion-8"
+                      class="MuiTypography-root MuiTypography-h4 emotion-8"
```
shawyu-okta, line 44: Are these MUI class changes intended?

src/v3/src/transformer/i18n/transformAuthenticatorButton.ts
```diff
@@ -28,40 +29,42 @@ export const transformAuthenticatorButton: TransformStepFnWithOptions = ({
...
 
-      // try get i18nKey without methodType
-      let i18nKey = getI18nKey(`${stepName}.authenticator.${(element as AuthenticatorButtonElement).options.key}`);
-      if (i18nKey) {
-        addTranslation({
-          element, name: 'label', i18nKey, params,
-        });
-        return;
-      }
+      (ele as AuthenticatorButtonListElement).options.buttons.forEach((element) => {
```
shawyu-okta, line 39: Can we rename the inner parameter `element` to something else to differentiate it from the `ele` from the callback function? I know you already renamed it to try to avoid the conflict but it is still a bit of a collision. `buttonElement` or similar might work?

src/v3/test/integration/__snapshots__/authenticator-apple-sso-extension.test.tsx.snap
glenfannin-okta: Yep these are intentional, what seemed to have happened is that this file was never updated after the Title variant change was made (from h3 to h4 by Alex). I am unsure as to how it wasnt caught until now though 😓

src/v3/src/transformer/selectAuthenticator/transformSelectOVMethodVerify.ts
glenfannin-okta: Oh yea, concat is not necessary here, just pushing the element is more than enough. Updating.

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
glenfannin-okta: Yessir! So let's take the example in this screenshot. Looking at the Phone Authenticator, the aria-describedby value is correctly set to 

```
select-authenticator-authenticate_Title_Verify_its_you_with_a_security_method_1 select-authenticator-authenticate_Description_Select_from_the_following_options_2 phone_number_4-description phone_number_4-ctaLabel
```

Which means a screenreader will first read out the button label, then the Title (Verify its you with a security method), then the subtitle (Select from the following options), then the button description (if any), then the call to action label (i.e. Select). 

image


Before this change, what would happen is it would read out the label, then the call to action text (in this case it would be "Select"), then the button description then finally the Title, subtitle, etc.

src/v3/src/transformer/selectAuthenticator/transformSelectOVMethodVerify.ts
shawyu-okta: Ok sounds good! There's a few other instances in other transformers as well

shawyu-okta (approved):

PR #2977, opened 2022-12-12
Fix csp issue in EnrollAuthenticatorGoogleAuthenticatorView

indirathirumalaimurugan-okta:
## Description:

Fix csp issue in EnrollAuthenticatorGoogleAuthenticatorView
Added test

Removed the attributes prop because`display: block` css is already being set on the next button via [_btns.scss](https://github.com/okta/okta-signin-widget/blob/master/assets/sass/modules/_btns.scss#L143). 

## PR Checklist

### Issue:

- [OKTA-555191](https://oktainc.atlassian.net/browse/OKTA-555191)

### Screenshot/Video:
Before
Screen Shot 2022-12-12 at 11 25 02 AM

After
Screen Shot 2022-12-12 at 11 22 56 AM

### DIFF
--- a/test/testcafe/framework/page-objects/EnrollGoogleAuthenticatorPageObject.js
+++ b/test/testcafe/framework/page-objects/EnrollGoogleAuthenticatorPageObject.js
@@ -58,7 +58,7 @@ export default class EnrollGoogleAuthenticatorPageObject extends BasePageObject 
     return this.form.getTextBoxErrorMessage(CODE_FIELD_NAME);
   }
 
-  getNextButtonClass() {
-    return this.form.getElement('.google-authenticator-next').getAttribute('class');
+  getNextButton() {
+    return this.form.getElement('.google-authenticator-next');
   }
 }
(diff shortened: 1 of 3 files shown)


test/testcafe/framework/page-objects/EnrollGoogleAuthenticatorPageObject.js
```diff
@@ -57,4 +57,8 @@ export default class EnrollGoogleAuthenticatorPageObject extends BasePageObject
   getCodeFieldError() {
     return this.form.getTextBoxErrorMessage(CODE_FIELD_NAME);
   }
+
+  getNextButtonClass() {
+    return this.form.getElement('.google-authenticator-next').getAttribute('class');
```
mauriciocastillosilva-okta, line 62: if it is just to check the visibility, I think you can do something like
```
isNextButtonVisible() {
  return this.form.getElement('.google-authenticator-next').visible;
}
```
If we need to check the specific value for `display`, maybe we should specifically check for that 
```
getNextButton() {
 this.form.getElement('.google-authenticator-next');
}

// in the test
expect(getNextButton().getStyleProperty('display')).eql('block');
```

mauriciocastillosilva-okta (commented):
some improvement for the test. Code changes look good!

shuowu-okta (approved):

PR #2979, opened 2022-12-13
Adds transformer for `device-challenge-poll` for `LOOPBACK`

shawyu-okta:
## Description:

Adds the device-challenge-poll transformer when the method is LOOPBACK.

The LoopbackProbe component contains nearly all of the imperative logic for this transformer.

We also will need to update this implementation once the auth-js library has either an exposed HTTP client or encapsulates the request logic to the localhost server within the library for this application to consume.

## PR Checklist

### Issue:

- [OKTA-548961](https://oktainc.atlassian.net/browse/OKTA-548961)

### DIFF
--- a/src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
+++ b/src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
@@ -10,10 +10,13 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+// Allow this shadow dependency from auth-js because it is temporary
+// eslint-disable-next-line import/no-extraneous-dependencies
+import crossFetch from 'cross-fetch';
 import { FunctionComponent } from 'preact';
 import { useEffect } from 'preact/hooks';
-import crossFetch from 'cross-fetch';
-import Logger from 'util/Logger';
+
+import Logger from '../../../../util/Logger';
 import { useWidgetContext } from '../../contexts';
 import { useOnSubmit } from '../../hooks';
 import { LoopbackProbeElement } from '../../types';
@@ -24,7 +27,7 @@ type RequestOptions = {
   timeout: number;
   method: 'GET' | 'POST';
   data?: string;
-}
+};
 
 /**
  * Temporary request client to be removed once this functionality is added
@@ -32,7 +35,9 @@ type RequestOptions = {
  * @see https://oktainc.atlassian.net/browse/OKTA-561852
  * @returns Promise<Response>
  */
-const makeRequest = async ({ url, timeout, method, data }: RequestOptions) => {
+const makeRequest = async ({
+  url, timeout, method, data,
+}: RequestOptions) => {
   const controller = new AbortController();
   const id = setTimeout(() => controller.abort(), timeout);
   const fetch = global.fetch || crossFetch;
@@ -46,7 +51,7 @@ const makeRequest = async ({ url, timeout, method, data }: RequestOptions) => {
   clearTimeout(id);
 
   return response;
-}
+};
 
 const LoopbackProbe: FunctionComponent<{ uischema: LoopbackProbeElement }> = () => {
   const onSubmitHandler = useOnSubmit();
@@ -55,9 +60,9 @@ const LoopbackProbe: FunctionComponent<{ uischema: LoopbackProbeElement }> = () 
   const relatesTo = idxTransaction?.nextStep?.relatesTo;
 
   // @ts-expect-error probeTimeoutMillis is not on IdxAuthenticator
-  const probeTimeoutMillis: number = typeof relatesTo?.value.probeTimeoutMillis === 'undefined' ?
+  const probeTimeoutMillis: number = typeof relatesTo?.value.probeTimeoutMillis === 'undefined'
     // @ts-expect-error probeTimeoutMillis is not on IdxAuthenticator
-    100 : relatesTo?.value.probeTimeoutMillis;
+    ? 100 : relatesTo?.value.probeTimeoutMillis;
   // @ts-expect-error ports is not on IdxAuthenticator
   const ports: number[] = relatesTo?.value.ports || [];
   // @ts-expect-error domain is not on IdxAuthenticator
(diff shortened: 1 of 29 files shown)


test/testcafe/framework/page-objects/TerminalPageObjectV3.js
```diff
@@ -1,15 +1,10 @@
...
 import TerminalPageObject from './TerminalPageObject';
 
-const BEACON_SELECTOR = '[data-se="factor-beacon"]';
 const CANCEL_LINK = '[data-se="cancel"]';
 
 export default class TerminalPageObjectV3 extends TerminalPageObject {
 
-  getBeaconClass() {
-    return Selector(BEACON_SELECTOR).getAttribute('class');
-  }
```
shawyu-okta, line 11: this just falls through to the BaseFormObject which has the same implementation

shawyu-okta:
Some TODOs:

* [ ] Add component tests for `LoopbackProbe.tsx`
* [ ] Add unit tests for new transformer

src/v3/src/transformer/oktaVerify/transformOktaVerifyChallengePoll.ts
```diff
@@ -119,6 +119,10 @@ export const transformOktaVerifyChallengePoll: IdxStepTransformer = (options) =>
       } as SpinnerElement);
     }
   } else if (selectedMethod.type === 'signed_nonce') {
+    if (relatesTo?.value.type === 'app') {
+      // selectedMethod.type === 'app' reflects a FastPass OV Loopback flow
```
shawyu-okta, line 123: TODO improve this

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
```diff
@@ -0,0 +1,172 @@
...
+}) => {
+  const onSubmitHandler = useOnSubmit();
+
+  const probeTimeoutMillis: number = typeof deviceChallengePayload.probeTimeoutMillis === 'undefined'
+    ? 100 : deviceChallengePayload.probeTimeoutMillis;
+  const ports: number[] = deviceChallengePayload.ports || [];
+  const { domain } = deviceChallengePayload;
+  const { challengeRequest } = deviceChallengePayload;
+
+  const cancelHandler = (params?: Record<string, unknown> | undefined) => {
```
glenfannin-okta, line 75: Just curious, is using the `ActionParams` type possible here? 
`const cancelHandler = (params?: ActionParams | undefined) => {`

Also, is `undefined` necessary if the parameter is optional?

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
```diff
@@ -0,0 +1,172 @@
...
+              // when challenge response with other error statuses,
+              // cancel polling and return
+              cancelHandler({
+                reason: 'OV_RETURNED_ERROR',
+                statusCode: challengeResponse.status,
+              });
+
+              break;
+            }
+            // no errors but this is not the port we're looking for
```
glenfannin-okta, line 134: ⭐ You deserve an Oktappreciate for the Star wars reference here 😅

glenfannin-okta (commented):
Overall lgtm. 

One additional question, will you need to modify `buildAuthCoinProps` in `src/util/idxUtils.ts` to ensure the authCoin is displayed or does it just work with the current implementation? 

It seems like it just works but was curious given how the transformer mapping is set up where it uses challenge methods instead of authenticator keys.

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
shawyu-okta: Good point, I think the `undefined` part was an artifact of a previous implementation. And I didn't know about the `ActionParams` type so that works great, thanks!

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
shawyu-okta: Haha I'm glad you noticed

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
```diff
@@ -0,0 +1,172 @@
...
+/**
+ * Temporary request client can be removed if this functionality is added
+ * to auth-js library.
+ * @see https://oktainc.atlassian.net/browse/OKTA-561852
+ * @returns Promise<Response>
+ */
+const makeRequest = async ({
+  url, timeout, method, data,
+}: RequestOptions) => {
+  const controller = new AbortController();
```
shawyu-okta, line 38: TODO; need to find an IE11 compatible way to handle abort, this is not available in IE11 and core-js polyfill does not have support for it currently either. Possibly use Promise.race? It still won't stop the request from completing but at least will keep the widget from hanging on long requests.

shawyu-okta: added in [3abcdbb](https://github.com/okta/okta-signin-widget/pull/2979/commits/3abcdbb4a0e4d106fd146e06a90238f7722dbb23)

shawyu-okta:
> One additional question, will you need to modify buildAuthCoinProps in src/util/idxUtils.ts to ensure the authCoin is displayed or does it just work with the current implementation?

> It seems like it just works but was curious given how the transformer mapping is set up since it uses challenge methods instead of authenticator keys.

Ah yeah so since `buildAuthCoinProps` uses `getAuthenticatorKey` it works since the latter method checks for `DEVICE_CHALLENGE_POLL` specifically and in that case returns the `challengeMethod`. Those values are mapped in the `AuthCoinConfig` file to pull the OV icon for all of the challenge methods so we show it correctly.

glenfannin-okta (approved):
Looks great! 

Just curious, will you add a test for the LookbackProbe component? Since it has quite a bit of logic in there? 
No big deal if not since you have the wiki that basically explains what this logic is doing, was just wondering.

shawyu-okta:
> Just curious, will you add a test for the LookbackProbe component? Since it has quite a bit of logic in there?
No big deal if not since you have the wiki that basically explains what this logic is doing, was just wondering.

Yeah, I want to add some tests before merging

glenfannin-okta (approved):

PR #2985, opened 2022-12-14
fix: ColorsUtil CSP issue

shuowu-okta:
## PR Checklist

### Issue:

### DIFF
--- a/assets/sass/okta-theme.scss
+++ b/assets/sass/okta-theme.scss
@@ -756,6 +756,7 @@
     border-color: $disabled-primary-button-bg-color;
     background-image: none;
     box-shadow: none;
+    opacity: 0.5;
 
     &:hover,
     &:focus,
--- a/src/util/ColorsUtil.js
+++ b/src/util/ColorsUtil.js
@@ -10,21 +10,8 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import Enums from 'util/Enums';
+import { $ } from 'okta';
 const fn = {};
-
-const template = function(colors) {
-  return `
-#okta-sign-in.auth-container .button-primary,
-#okta-sign-in.auth-container .button-primary:active,
-#okta-sign-in.auth-container .button-primary:focus { background: ${colors.brand}; }
-#okta-sign-in.auth-container .button-primary:hover { background: ${fn.lighten(colors.brand, 0.05)}; }
-#okta-sign-in.auth-container .button.button-primary.link-button-disabled {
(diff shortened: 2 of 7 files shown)


assets/sass/okta-theme.scss
```diff
@@ -756,6 +756,7 @@
     border-color: $disabled-primary-button-bg-color;
     background-image: none;
     box-shadow: none;
+    opacity: 0.5;
```
aarongranick-okta, line 759: seems like there may be a small difference here. The "opacity" rule added by the "template" function is only applied if `colors.brand` is set, but the CSS rule here would apply in all cases.

src/util/ColorsUtil.js
```diff
@@ -41,20 +28,16 @@ fn.lighten = function(hex, lum) {
...
-  main.appendChild(style);
-};
-
-fn.isLoaded = function() {
-  return !!document.getElementById(Enums.WIDGET_CONFIG_COLORS_ID);
+fn.addStyle = function($containerEl, colors) {
+  const $primaryButton = $containerEl.find('.button-primary');
+  $primaryButton
+    .css('background', colors.brand)
+    .on('mouseenter', function() {
```
aarongranick-okta, line 35: do we actually want "mouseover"?

src/util/ColorsUtil.js
```diff
@@ -41,20 +28,16 @@ fn.lighten = function(hex, lum) {
...
+      $(this).css('background', fn.lighten(colors.brand, 0.05));
+    })
+    .on('mouseleave', function() {
```
aarongranick-okta, line 38: same question, do we want "mouseout" ?

src/v2/controllers/FormController.ts
```diff
@@ -67,6 +68,14 @@ export default Controller.extend({
           currentViewState,
         }
       }).last();
+      
+      // Apply the custom colors after render
+      if (this.settings.get('colors.brand')) {
```
jaredperreault-okta, line 73: Do we no longer need the `!ColorUtils.isLoaded()` check?

src/v1/BaseLoginRouter.js
```diff
@@ -216,6 +207,15 @@ export default Router.extend({
 
         this.controller.render();
 
+        // Apply the custom colors after render
+        if (this.settings.get('colors.brand')) {
```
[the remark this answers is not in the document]
shuowu-okta: with this approach, not style block will be injected, then no need to check that any more

PR #2991, opened 2022-12-16
Enroll authenticator email v3 parity

vijetmahabaleshwar-okta:
## PR Checklist

### Issue:

- [OKTA-509017](https://oktainc.atlassian.net/browse/OKTA-509017)


test/testcafe/spec/EnrollAuthenticatorEmail_spec.js
```diff
@@ -158,9 +164,9 @@ test
 
     await t.expect(enrollEmailPageObject.form.getTitle()).eql('Verify with your email');
 
-    const emailAddress = xhrEnrollEmailWithEmailMagicLink.user.value.identifier;
+    // const emailAddress = xhrEnrollEmailWithEmailMagicLink.user.value.identifier;
     await t.expect(enrollEmailPageObject.getFormSubtitle())
-      .eql(`We sent an email to ${emailAddress}. Click the verification link in your email to continue or enter the code below.`);
```
vijetmahabaleshwar-okta, line 163: v3 widget doesn't display the email address, at least in the tests

test/testcafe/spec/EnrollAuthenticatorEmail_spec.js
```diff
@@ -143,9 +153,9 @@ test
 
     await t.expect(enrollEmailPageObject.form.getTitle()).eql('Verify with your email');
     
-    const emailAddress = xhrEnrollEmailWithoutEmailMagicLink.user.value.identifier;
+    // const emailAddress = xhrEnrollEmailWithoutEmailMagicLink.user.value.identifier;
```
shawyu-okta, line 156: Some commented code here - do we need to re-enable it for the v2 test?

test/testcafe/spec/EnrollAuthenticatorEmail_spec.js
shawyu-okta: Do we need to have different assertions for v2 vs v3?

test/testcafe/spec/EnrollAuthenticatorEmail_spec.js
```diff
@@ -293,20 +302,20 @@ test
...
-    await t.expect(enrollEmailPageObject.resendEmail.getText()).eql('Haven\'t received an email? Send again');
+    await t.expect(await enrollEmailPageObject.resendEmailExists()).eql(true);
+    await t.expect(enrollEmailPageObject.resendEmailText()).contains('Haven\'t received an email?');
 
     // 8 poll requests in 31 seconds and 1 resend request after click.
     await t.expect(logger.count(
       record => record.response.statusCode === 200 &&
         record.request.url.match(/poll/)
-    )).eql(8);
+    )).gte(8);
```
shawyu-okta, line 314: Are we polling more than 8 times in v3? If so, how many more times?

test/testcafe/spec/EnrollAuthenticatorEmail_spec.js
```diff
@@ -330,28 +339,29 @@ test
     await t.expect(firstRequestUrl).eql('http://localhost:3000/idp/idx/challenge/poll');
 
     jsonBody = JSON.parse(lastRequestBody);
-    await t.expect(jsonBody).eql({'stateHandle':'eyJ6aXAiOiJER'});
+    // in V3, it's { resend: true, stateHandle: 'eyJ6aXAiOiJER' }
+    await t.expect(jsonBody).contains({'stateHandle':'eyJ6aXAiOiJER'});
```
shawyu-okta, line 343: I think this is a reasonable change that doesn't necessarily require a comment to explain it

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
```diff
@@ -21,12 +22,57 @@ export default class EnrollAuthenticatorPhonePageObject extends BasePageObject {
...
+
+    const isHidden = await Selector(RESEND_EMAIL).hasClass('hide')
+    return !isHidden;
+  }
+
+  resendEmailText() {
+    if (userVariables.v3) {
+      return this.form.el.find('[role="alert"]').textContent;
+    }
+    else {
```
shawyu-okta, line 59: nit: else is not necessary here

test/testcafe/spec/EnrollAuthenticatorEmail_spec.js
vijetmahabaleshwar-okta: I can add that but I was trying to avoid having different assertions for v2 and v3. Maybe a parking lot discussion item?

test/testcafe/spec/EnrollAuthenticatorEmail_spec.js
vijetmahabaleshwar-okta: I did see it fail on odd occasions where v3 was polling > 8 times. I tested locally now and it seems to be passing with `.eql(8)` assertion. Will test a few more times to make sure it's stable and push the change.

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
vijetmahabaleshwar-okta: Done

test/testcafe/spec/EnrollAuthenticatorEmail_spec.js
vijetmahabaleshwar-okta: No harm in keeping the comment I guess.

test/testcafe/spec/EnrollAuthenticatorEmail_spec.js
vijetmahabaleshwar-okta: Changed it back to `.eql(8)` assertion. Will monitor the CI runs to make sure it's green. So far, it's stable locally.

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
```diff
@@ -21,12 +22,55 @@ export default class EnrollAuthenticatorPhonePageObject extends BasePageObject {
   }
 
   async enterCodeFromEmailLinkExists() {
-    const elCount = await Selector(ENTER_CODE_FROM_EMAIL).count;
-    return elCount === 1;
+    if (userVariables.v3) {
+      const elCount = await Selector('[type="button"]').count;
+      return elCount === 1;
+    } else {
```
shawyu-okta, line 28: nit: unnecessary `else`

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
```diff
@@ -21,12 +22,55 @@ export default class EnrollAuthenticatorPhonePageObject extends BasePageObject {
...
+
+    const isHidden = await Selector(RESEND_EMAIL).hasClass('hide')
+    return !isHidden;
+  }
+
+  resendEmailText() {
+    if (userVariables.v3) {
+      return this.form.el.find('[role="alert"]').textContent;
+    }
+      return this.form.el.find(RESEND_EMAIL).textContent;
```
shawyu-okta, line 59: It looks like there isn't linting enabled for `testcafe` directory fyi @lesterchoi-okta we might want to turn this on somewhere in a future piece of tech debt work

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
```diff
@@ -21,12 +22,55 @@ export default class EnrollAuthenticatorPhonePageObject extends BasePageObject {
...
+  }
+
+  async clickResendEmail() {
+    if (userVariables.v3) {
+      const resendEmail = this.form.getLink('Send again');
+      await this.t.click(resendEmail);
+    }
+    else {
+      const resendEmailLink = this.form.el.find(RESEND_LINK);
```
shawyu-okta, line 68: Are we able to use the same `getLink` code as in v3 to pick this link out? As long as both are an `a` tag with 'Send again' as the text it should work for both.

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
```diff
@@ -21,12 +22,55 @@ export default class EnrollAuthenticatorPhonePageObject extends BasePageObject {
   }
 
   async enterCodeFromEmailLinkExists() {
-    const elCount = await Selector(ENTER_CODE_FROM_EMAIL).count;
-    return elCount === 1;
+    if (userVariables.v3) {
+      const elCount = await Selector('[type="button"]').count;
```
shawyu-okta, line 26: This selector might be a little too brittle since it picks up all buttons on the page - Are you able to use something like
```ts
return this.form.getButton(ENTER_CODE_FROM_EMAIL_TEXT).exists;
```

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
```diff
@@ -21,12 +22,55 @@ export default class EnrollAuthenticatorPhonePageObject extends BasePageObject {
...
+    if (userVariables.v3) {
+      await this.form.clickButton(ENTER_CODE_FROM_EMAIL_TEXT);
+    } else {
+      await this.form.clickElement(ENTER_CODE_FROM_EMAIL_CLASS);
+    }
+  }
+
+  async resendEmailExists() {
+    if (userVariables.v3) {
+      return Selector('[role="alert"]').exists;
```
shawyu-okta, line 48: We should use one of these existing selectors (see `BaseFormObject`):
* `this.form.getErrorBoxCount`
* `this.form.getErrorBoxText`
* `this.form.hasErrorBox`

test/testcafe/spec/EnrollAuthenticatorEmail_spec.js
shawyu-okta: I think we could update the `enrollEmailPageObject` to have a new method that asserts different texts for v2 vs v3 like `hasEmailVerificationSubtitle(emailAddress)` so the branching isn't in the spec itself but rather still in the page object.

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
vijetmahabaleshwar-okta: Screen Shot 2023-01-03 at 9 55 54 AM
Turns out that `this.form.getButton` works for elements with `role=button`. But for this element, it only has `type=button` and there's no corresponding method. Is there any way we can add `role=button` for this element? @shawyu-okta

shawyu-okta: `button` elements have an implied `role=button` so `getButton` should work. Let me know if that's not the case

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
vijetmahabaleshwar-okta: `this.form.getLink` looks for `role="link"`. Even though the v2 element has an `a` tag, it doesn't have `role="link"` and `getLink` fails to pick it up

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
vijetmahabaleshwar-okta: For some reason, this element doesn't have the implied `role=button`. `getButton` doesn't work and fails in this particular case.

vijetmahabaleshwar-okta: This doesn't work as expected. 
`this.form.getButton(ENTER_CODE_FROM_EMAIL_TEXT).exists;` fails in case the element doesn't exist. This is because the `getButton` method looks for an element with `role=button` and when it doesn't exist, it throws an error saying element not found.

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -221,7 +221,7 @@ export default class BaseFormObject {
   }
 
   hasErrorBox() {
-    return within(this.el).getByRole('alert').exists;
+    return Selector('[role="alert"]').exists;
```
shawyu-okta, line 224: Is this change needed for the tests? We should prefer the `getByRole` selector whenever possible

shawyu-okta (approved):
Last comment, lgtm!

test/testcafe/framework/page-objects/components/BaseFormObject.js
vijetmahabaleshwar-okta: `within(this.el).getByRole('alert').exists;` never works when the alert doesn't exist. The `getByRole` requires the element to be present to work. So in case it doesn't exist, it throws an error.
So far, this was only used to check if the alert exists. But in this PR, I had to test for alert doesn't exist and this doesn't work.

shawyu-okta: Let's change it to `queryByRole`, should work then

PR #3017, opened 2023-01-09
OKTA-568301: a11y plugin improvements

lesterchoi-okta:
## Description:

Refactor of `OktaPluginA11y`

- updates page titles on `afterRender` based on `.okta-form-title`
- use `oktaSignIn.options.brandName` instead of `companyName` (type updated)
- use existing DOM elements instead of hardcoded strings for better l10n support
- move minor style changes to `.scss` instead of overriding in plugin
- replace checkbox / radio button sprites with assets from [OKTA-513274](https://oktainc.atlassian.net/browse/OKTA-513274)

## PR Checklist

### Issue:

- [OKTA-568301](https://oktainc.atlassian.net/browse/OKTA-568301)
- [OKTA-555277](https://oktainc.atlassian.net/browse/OKTA-555277)

### Screenshot/Video:

Fixed
screenshot 2023-01-09 at 17 46 53

Outstanding
screenshot 2023-01-10 at 11 13 02

Oustanding
screenshot 2023-01-10 at 09 39 17


src/types/api.ts
```diff
@@ -20,6 +20,7 @@ export interface RouterEventsAPI {
 export type RenderSuccessCallback = (res: RenderResult) => void;
 export type RenderErrorCallback = (err: RenderError) => void;
 export interface OktaSignInAPI extends HooksAPI, RouterEventsAPI {
+  readonly options: Pick<WidgetOptions, 'brandName'>;
```
lesterchoi-okta, line 23: type-only update

assets/sass/_variables.scss
```diff
@@ -24,7 +24,7 @@ $secondary-bg-color: #f9f9f9;
 // Text Color
 $header-text-color: #5e5e5e;
 $dark-text-color: #5e5e5e;
-$medium-text-color: #777;
+$medium-text-color: #767676;
```
lesterchoi-okta, line 27: `#777:#fff` → 4.47:1 FAIL
`#767676:#fff` → 4.54:1 PASS

assets/sass/common/shared/helpers/_variables.scss
```diff
@@ -4,7 +4,7 @@
 @use "sass:math";
 
 // Okta colors
-$primary-color: #007dc1;
+$primary-color: #007cc0;
```
lesterchoi-okta, line 7: `#007dc1:#fff` → 4.46:1 FAIL
`#007cc0:#fff` → 4.52:1 PASS

assets/css/okta-plugin-a11y.css
```diff
@@ -22,7 +28,7 @@ button {
   color: #222!important;
 }
 #okta-sign-in.auth-container.main-container {
-  color: #222!important;
+  color: #767676!important;
```
lesterchoi-okta, line 31: `#222` was too dark. The contrast between normal text and "subtle" text was not perceivable for sighted users

assets/css/okta-plugin-a11y.css
```diff
@@ -14,6 +14,12 @@ button {
   background: none;
 }
 
+a:link,
+a:visited,
+a:active {
```
lesterchoi-okta, line 19: Considered changing this in [assets/sass/_base.scss](https://github.com/okta/okta-signin-widget/blob/9897f7a238a04fd5e7b841ec01bbfa431de91bd7/assets/sass/_base.scss#L7) but needs designer verification

alexdahl-okta: Running this branch locally, looks like `_base.scss` is overriding this underline, and `text-decoration: none` is still in effect.

Philosophically, I support underlines on links. (Odyssey disagrees. But we're optimizing for different use cases, so divergence could be justifiable.) If we do introduce underlines, we should do so carefully.

For example, we'll need to make sure that buttons built as `<a>` don't underline their labels.

**Adding underlines to links is probably easiest?**

I believe [WCAG guidance](https://www.w3.org/WAI/WCAG21/Techniques/general/G183) is that while not preferred, lack of underlines is acceptable if the link color has 3:1 contrast against surrounding text.

The current color palette in the widget does not achieve that contrast.

The current Odyssey palette [does pass contrast requirements](https://webaim.org/resources/linkcontrastchecker/?fcolor=007BBD&bcolor=767676) against body text, but not for links against subtle text.

I'm open to changing our color use to the Odyssey palette instead of adding underlines, but that's probably harder. Widget links currently in the body color would need to change to the link color, and subtle text next to a link would need to change to the darker body color. There'd be a lot to audit, and I expect a lot of CSS to wrangle with.

assets/sass/_variables.scss
alexdahl-okta: I might have a preference for aligning with the current Odyssey colors for subtle and body text, instead of introducing more arbitrary shades. But not a big deal, probably no practical difference.

Subtle text: `#6E6E78`
Body text: `#1D1D21`

alexdahl-okta (commented):
👏 Excited for this!

I'm mostly talking philosophically, so feel free to ignore anything in favor of practicality.

Philosophically, I think these fixes should eventually get streamlined into "real" authoritative assets like `_base.scss` and shipped to everybody. But I also agree with our approach to start as an opt-in patch. So I'll defer to you on the details on what lives where for now!

lkchoi:
> 👏 Excited for this!
> 
> 
> 
> I'm mostly talking philosophically, so feel free to ignore anything in favor of practicality.
> 
> 
> 
> Philosophically, I think these fixes should eventually get streamlined into "real" authoritative assets like `_base.scss` and shipped to everybody. But I also agree with our approach to start as an opt-in patch. So I'll defer to you on the details on what lives where for now!

Agree. Tagged you because there's a violation that requires some design consideration.

Right now we distinguish normal text from a link via color only (black vs blue). WCAG 1.4.1 says color should not be the only way we convey this information.

Unfortunately, we use underline for other states like hover. Can you help design how we should distinguish among the various states of a link?

Extra complication is that some elements that **appear** to be buttons are actually `<a>` tags.

lkchoi:
@alexdahl-okta See the [3rd screenshot](https://user-images.githubusercontent.com/77294605/211605414-59f195f4-067b-4993-b793-066660a246e4.png) in description for context

.testcaferc.json
```diff
@@ -1,6 +1,17 @@
...
+      "module": "axe-core/axe.min.js"
+    }
+  ],
+  "concurrency": 4,
+  "src": [
+    "test/testcafe/spec/*_spec.js"
+  ],
+  "userVariables": {
+    "v3": false
+  }
```
shawyu-okta, line 16: Is this needed in branches from the mainline?

shawyu-okta (commented):
What is the process with regards to changing styles/sprites for all users? It would be a somewhat significant visual change, is that something we want to push out to all users?

src/plugins/OktaPluginA11y.ts
```diff
@@ -183,106 +9,47 @@ const isEqual = (a,b) => [...Object.keys(a)].every((k) => a[k] === b[k]);
...
+     * sibling element .authenticator-description conveys the same information.
+     * 
+     * https://www.w3.org/WAI/tutorials/images/decision-tree/
+     * https://www.w3.org/WAI/tutorials/images/decorative/
+     */
+    document.querySelectorAll('.authenticator-list').forEach((list) => {
+      list.querySelectorAll('.authenticator-row')
+        .forEach((row) => {
+          row.querySelector('.authenticator-icon')
+            ?.setAttribute('aria-hidden', '');
```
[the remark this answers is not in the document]
lesterchoi-okta: > What is the process with regards to changing styles/sprites for all users? It would be a somewhat significant visual change, is that something we want to push out to all users?

Not sure re; process. I'd group it with the other changes in the PR and say it's a set of noticeable visual changes that improves accessibility, i.e., bug fixes.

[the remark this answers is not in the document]
lesterchoi-okta: For the suggested code change, it's to match the method's type signature.

```ts
// from lib.dom.d.ts
setAttribute(qualifiedName: string, value: string): void;
```

assets/sass/v2/_identify.scss
```diff
@@ -73,7 +73,8 @@
       background: $white;
       padding: 0 15px;
       font-size: 13px;
-      color: $border-color-default;
+      color: $medium-text-color;
+      // color: $border-color-default;
```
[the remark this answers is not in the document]
lesterchoi-okta: I'll clean up

.testcaferc.json
lesterchoi-okta: Nope. Will clean up

src/plugins/OktaPluginA11y.ts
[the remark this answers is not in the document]
shawyu-okta: Okay, so would it be correct then to do this instead?
```suggestion
            ?.setAttribute('aria-hidden', 'true');
```

shawyu-okta (approved):

glenfannin-okta (approved):

PR #3019, opened 2023-01-11
OKTA-548963 : Support APP_LINK and UNIVERSAL_LINK in device-challenge-poll transformer

leemchale-okta:
## Description:

This PR adds support for `APP_LINK` and `UNIVERSAL_LINK` in the `device-challenge-poll` transformer.

## PR Checklist

### Issue:

- [OKTA-548963](https://oktainc.atlassian.net/browse/OKTA-548963)


src/v3/src/components/OpenOktaVerifyFPButton/OpenOktaVerifyFPButton.tsx
```diff
@@ -61,14 +64,25 @@ const OpenOktaVerifyFPButton: UISchemaElementComponent<{
...
+      },
     },
   };
 
   return (
     <React.Fragment>
       <Button uischema={buttonUiSchema} />
-      {href && (
+      {(href && challengeMethod === CHALLENGE_METHOD.CUSTOM_URI) && (
       <IFrame
```
leemchale-okta, line 86: @shawyu-okta  Just to clarify is the IFrame only shown for the CUSTOM_URI method?

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -12,60 +12,158 @@
...
+  };
 
-  uischema.elements.push({
-    type: 'Description',
-    contentType: 'subtitle',
-    options: { content: loc('customUri.required.content.download.title', 'login') },
-  } as DescriptionElement);
-
-  uischema.elements.push({
+  const cancelStep = transaction.nextStep?.name === IDX_STEP.DEVICE_CHALLENGE_POLL
```
leemchale-okta, line 90: polling cancel step is used

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -537,7 +544,9 @@ test
     await t.expect(deviceChallengePollPageObject.getAppLinkContent())
       .contains('If Okta Verify did not open automatically, tap Open Okta Verify.');
     await t.expect(deviceChallengePollPageObject.getPrimaryButtonText()).eql('Open Okta Verify');
-    await t.expect(deviceChallengePollPageObject.getFooterCancelPollingLink().exists).eql(false);
+    if(!userVariables.v3) {
+      await t.expect(deviceChallengePollPageObject.getFooterCancelPollingLink().exists).eql(false);
```
leemchale-okta, line 548: FooterCancelPollingLink and SIgnOutLink are different in v2

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -697,7 +708,7 @@ test
   });
 
 test
-  .meta('v3', false) // Not yet implemented in v3 (OKTA-548963)
+  .meta('v3', true)
```
shawyu-okta, line 711: instead of changing to `true` you can delete this line instead. same with the others above

test/testcafe/framework/page-objects/DeviceChallengePollPageObject.js
```diff
@@ -27,6 +27,9 @@ export default class DeviceChallengePollViewPageObject extends BasePageObject {
   }
 
   getAppLinkContent() {
+    if(userVariables.v3) {
+      return this.form.getByText('If Okta Verify did not open automatically, tap Open Okta Verify.').innerText;
+    }
```
shawyu-okta, line 32: This is a bit redundant, since the returned text will just be the text passed to `getByText`.

In these cases I like to try to update the method to be more like `hasAppLinkContent` and allow the `getByText` to throw when it's not there, or still leave it as `getAppLinkContent` but update the selection to `return queryByText('target text')` and then also update the test to assert that the element returned exists.

src/v3/src/components/AutoStepper/AutoStepper.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { useEffect } from 'preact/hooks';
+
+import { useStepperContext } from '../../contexts';
+import { AutoStepperElement, UISchemaElementComponent } from '../../types';
+
+const AutoStepper: UISchemaElementComponent<{
+  uischema: AutoStepperElement
+}> = ({ uischema }) => {
+  const { setStepIndex } = useStepperContext();
```
shawyu-okta, line 21: We should also leave a note here to remind users that this must be used within a stepper context provider

src/v3/src/transformer/layout/idxTransformerMapping.ts
```diff
@@ -178,11 +178,18 @@ const TransformerMap: {
...
+        showDefaultSubmit: false,
+        showDefaultCancel: false,
+      },
+    },
     [CHALLENGE_METHOD.CUSTOM_URI]: {
       transform: transformOktaVerifyDeviceChallengePoll,
       buttonConfig: {
         showDefaultSubmit: false,
-        showDefaultCancel: true,
+        showDefaultCancel: false,
```
shawyu-okta, line 192: Ah, did I get this one wrong? I guess we actually need to add the cancel on our own in a custom way.

test/testcafe/spec/DeviceChallengePollView_spec.js
glenfannin-okta: When you say they are different, are they completely different links? Or is it just a naming difference and functionality is the same, can this conditional be added directly to the function `getFooterCancelPollingLink`?

src/v3/src/components/AutoStepper/AutoStepper.tsx
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { useEffect } from 'preact/hooks';
+
+import { useStepperContext } from '../../contexts';
+import { AutoStepperElement, UISchemaElementComponent } from '../../types';
+
+const AutoStepper: UISchemaElementComponent<{
```
glenfannin-okta, line 18: Overall, this is fine and works, but it is set up for a very specific scenario. If we wanted to we could make this a little more generic by creating a `callback` property in the AutoStepperElement interface, and using that callback in the useEffect of this component. This way, the transformers can define what this "Navigator" Component  should do, this also would fall more in line with the new name suggestion of this Component and interface. 

The callback function can accept the Stepper context as a parameter (to allow the transformer to do what it needs to do), and this component can pass the context into the callback. 

Not sure if this is overkill but I just want to avoid future duplication of this sort of thing and give future devs a path forward to do this sort of thing.

glenfannin-okta: Would love to hear your thoughts @shawyu-okta maybe this is just overkill.

shawyu-okta: @glenfannin-okta can you give an example of how that would look? I'm wondering what the parameter to the callback would be.

test/testcafe/spec/DeviceChallengePollView_spec.js
leemchale-okta: Looks like just the names are different but the functionality is the same in v2.  In v3 they both have the name '`Back to sign in`' so I think this conditional needs to be here in the spec test to check if one exists but not the other for v2 since `getFooterCancelPollingLink` also gets the same Back to sign in link in v3

src/v3/src/components/AutoStepper/AutoStepper.tsx
glenfannin-okta: Yessir, something like this; 

Interface: 
```
export interface StepperNavigatorElement {
  type: 'StepperNavigator',
  options: {
     callback: (stepperContext: IStepperContext) => void;
  }
} 
```

Component:
```
const StepperNavigator: UISchemaElementComponent<{
  uischema: StepperNavigatorElement
}> = ({ uischema }) => {
  const stepperContext = useStepperContext();
  const {
    options: {
      callback,
    },
  } = uischema;

  useEffect(() => {
    callback(stepperContext);
  }, []);

  return null;
};

export default StepperNavigator;
```

Then finally, the transformer:
```
...
if (challengeMethod === CHALLENGE_METHOD.APP_LINK
      && prevTransaction?.nextStep?.name === IDX_STEP.IDENTIFY) {
    const stepperNavigatorElement: StepperNavigatorElement = {
      type: 'StepperNavigator',
      options: {
        callback: (stepperContext: IStepperContext) => {
           const { setStepIndex } = stepperContext;
           setTimeout(() => setStepIndex!(1), FASTPASS_FALLBACK_SPINNER_TIMEOUT);
        },
      },
    };

    uischema.elements.push({
      type: UISchemaLayoutType.STEPPER,
      elements: [
        {
          type: UISchemaLayoutType.VERTICAL,
          elements: [
            stepperNavigatorElement,
            spinnerElement,
            cancelLink,
          ],
        } as UISchemaLayout,
        {
          type: UISchemaLayoutType.VERTICAL,
          elements: [
            descriptionElement,
            openOktaVerifyButton,
            cancelLink,
          ],
        } as UISchemaLayout],
    });

    return formBag;
  }
...
```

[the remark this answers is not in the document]
leemchale-okta: Yeah name may be confusing.  What about `AutoStepIndex`?

test/testcafe/spec/DeviceChallengePollView_spec.js
glenfannin-okta: Ahh I see, so it actually looks like there is a parity issue here, it looks like the link should not be `Back to sig in`, based on the implementation [here](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/device/DeviceChallengePollView.js#L61), it looks like it should actually be `Cancel and take me to sign in`.

src/v3/src/components/AutoStepper/AutoStepper.tsx
leemchale-okta: My thinking with making this component was that it may be more flexible since it could be placed in any of the `elements` indexes lets say if you didn't want it to do the auto step on the first page but when the user lands on a different page/index.  You could also place multiple in different indexes if you wanted to have multiple transitions.  But that is only for possible future use not guaranteed that it will be needed so your method if having the callback in the `AutoStepperElement` would work too.

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -12,60 +12,158 @@
...
+    ? 'authenticatorChallenge-cancel' : 'currentAuthenticator-cancel';
+  const cancelLink: LinkElement = {
     type: 'Link',
+    contentType: 'footer',
     options: {
-      label: loc('customUri.required.content.download.linkText', 'login'),
-      href: deviceChallengePayload.downloadHref,
+      label: loc('goback', 'login'),
```
glenfannin-okta, line 96: I think updating the key used here will fix the issue with the test, mimicking what is used in the v2 view [here](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/views/device/DeviceChallengePollView.js#L61)

src/v3/src/components/AutoStepper/AutoStepper.tsx
glenfannin-okta: I see, yea the above modifications would enable the same, the only real difference is that this approach delegates to the transformer to tell the component _how_ it should navigate between steps. 

For example, the current approach only allows it to navigate using a setTimeout, whereas, there could be other scenarios where it doesn't want to use a setTimeout, but instead maybe navigate based on a data change or something else.

But its up to you guys, if you feel like this is overkill for now then I totally understand.

shawyu-okta: I generally favor an approach of simpler encapsulated behavior as long as it doesn't prevent us from expanding to more generalized behavior in the future. That would favor the existing approach for now, although if we run into other use-cases we should implement as @glenfannin-okta suggested. I don't think it would be too much work to update the transformers that use this component if we do rework this component to take a callback in the future.

leemchale-okta: ahh I see the call back is in the `StepperNavigator` component so it will have the same flexibility with the added flexibility of being able to navigate on a custom condition.  I do like that and am cool with implementing it that way.

leemchale-okta: @shawyu-okta  true it wouldn't be too difficult to update it in the future so that sounds good to me.  What do you think of the name should it remain the same or change it to 'StepperNavigator'?

leemchale-okta: Yeah im gonna see if I can implement glens approach shouldnt be too much work

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -12,60 +12,158 @@
...
   TitleElement,
+  UISchemaLayout,
+  UISchemaLayoutType,
 } from '../../../types';
 import { loc } from '../../../util';
 
+const getTitleText = (challengeMethod = CHALLENGE_METHOD.CUSTOM_URI) => {
+  if (challengeMethod === CHALLENGE_METHOD.APP_LINK) {
+    return loc('appLink.title', 'login');
+  } if (challengeMethod === CHALLENGE_METHOD.UNIVERSAL_LINK) {
```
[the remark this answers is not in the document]
leemchale-okta: Yeah my mistake I think I had `else ifs` there and got rid of the `elses` but forgot to move to new line

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -12,60 +12,158 @@
...
+    return loc('universalLink.title', 'login');
+  }
+  return loc('customUri.title', 'login');
+};
+
+const getDescriptionText = (challengeMethod = CHALLENGE_METHOD.CUSTOM_URI) => {
```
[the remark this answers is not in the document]
leemchale-okta: Not needed my bad forgot to remove

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -12,60 +12,158 @@
...
     },
-  } as LinkElement);
+  };
+
+  uischema.elements.unshift(titleElement);
+
+  // if the previous step was identify and the challenge method is APP_LINK,
+  // we delay displaying the content because of a cold start issue with Okta Verify
+  if (challengeMethod === CHALLENGE_METHOD.APP_LINK
+      && prevTransaction?.nextStep?.name === IDX_STEP.IDENTIFY) {
```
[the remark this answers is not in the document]
leemchale-okta: is that because the previous step to CHALLENGE_POLL is always IDENTIFY?

src/v3/src/transformer/layout/idxTransformerMapping.ts
leemchale-okta: Yeah looks like a custom cancel link was needed for all methods for cancel polling

test/testcafe/spec/DeviceChallengePollView_spec.js
leemchale-okta: yeah for some reason it looks like in the APP_LINK method during the 'delay', the link is `Cancel and take me to sign in`, then after the delay is finished they update the link the 'Back to sign in'... lol very confusing

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -12,60 +12,158 @@
...
+          ],
+        } as UISchemaLayout,
+        {
+          type: UISchemaLayoutType.VERTICAL,
+          elements: [
+            descriptionElement,
+            openOktaVerifyButton,
+            cancelLink,
+          ],
+        } as UISchemaLayout],
```
glenfannin-okta, line 138: Can you add the `viewIndex` properties to these arrays, similar to https://github.com/okta/okta-signin-widget/blob/0.1/src/v3/src/transformer/layout/googleAuthenticator/transformGoogleAuthenticatorEnroll.ts#L105

leemchale-okta: Okay will do.  What is that used for?

test/testcafe/spec/DeviceChallengePollView_spec.js
leemchale-okta: at this part [here](https://github.com/okta/okta-signin-widget/blob/09fb3ad250f41534934be7f19ccafaac51d2ab0c/src/v2/view-builder/utils/ChallengeViewUtil.js#L121)

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -12,60 +12,164 @@
...
+            cancelLink,
+          ].map((ele: UISchemaElement) => ({ ...ele, viewIndex: 0 })),
+        } as UISchemaLayout,
+        {
+          type: UISchemaLayoutType.VERTICAL,
+          elements: [
+            descriptionElement,
+            openOktaVerifyButton,
+            cancelLink,
+          ].map((ele: UISchemaElement) => ({ ...ele, viewIndex: 0 })),
```
glenfannin-okta, line 143:
```suggestion
          ].map((ele: UISchemaElement) => ({ ...ele, viewIndex: 1 })),
```

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
glenfannin-okta: Its is used by one of the generic transformers to distinguish between the different views of a Stepper. When it traverses the layout, before this change, there is no way for it to know what elements belong to which view.

leemchale-okta: ah awesome got it thanks!

test/testcafe/spec/DeviceChallengePollView_spec.js
glenfannin-okta: Oh I see, we should probably create a ticket to implement this logic and either disable these tests or at least add that ticket number here so we can remove this conditional once we get it completely working.

leemchale-okta: okay I may be able to implement a work around

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyFPLoopbackPoll.ts
```diff
@@ -55,7 +55,7 @@ export const transformOktaVerifyFPLoopbackPoll: IdxStepTransformer = ({
     type: 'Link',
     contentType: 'footer',
     options: {
-      label: loc('goback', 'login'),
```
leemchale-okta, line 58: had to change this to match parity with v2.

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyFPLoopbackPoll.ts
```diff
@@ -55,7 +55,7 @@ export const transformOktaVerifyFPLoopbackPoll: IdxStepTransformer = ({
...
+      label: loc('loopback.polling.cancel.link', 'login'),
```
leemchale-okta, line 58: had to change this to match parity with v2.

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
leemchale-okta: I was able to resolve the test issue by updating the key in the FPLoopback transformer aswell as in the device challenge transformer

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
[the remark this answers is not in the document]
shawyu-okta: That's kind of what I'm wondering - it seems like the case but wasn't sure.

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -247,13 +247,19 @@ const universalLinkMock = RequestMock()
...
   .onRequestTo(/(2000|6511|6512|6513)\/probe/)
   .respond(null, 500, { 'access-control-allow-origin': '*' })
   .onRequestTo(/\/idp\/idx\/authenticators\/poll\/cancel/)
   .respond(identifyWithLoopbackFallbackAndroidWithoutLink)
   .onRequestTo(/\/idp\/idx\/authenticators\/okta-verify\/launch/)
   .respond(identifyWithLaunchAppLink)
-  .onRequestTo(/\/idp\/idx\/authenticators\/poll/)
+  .onRequestTo(request => {
+    return request.url.match(/\/idp\/idx\/authenticators\/poll/) && appLinkLoopBackFailed;
+  })
```
shawyu-okta, line 262: nit: are you able to place this up with the other /poll/ request? Also is the lack of the `$` at the end intentional?

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -12,60 +12,173 @@
...
     },
-  } as LinkElement);
+  };
+
+  uischema.elements.unshift(titleElement);
+
+  // if the current step is device-challenge-poll and the challenge method is APP_LINK,
+  // we delay displaying the content because of a cold start issue with Okta Verify
+  if (challengeMethod === CHALLENGE_METHOD.APP_LINK
+      && nextStep?.name === IDX_STEP.DEVICE_CHALLENGE_POLL) {
```
shawyu-okta, line 108: Can you add a unit test that covers the code in this block?

test/testcafe/framework/page-objects/DeviceChallengePollPageObject.js
```diff
@@ -75,27 +75,41 @@ export default class DeviceChallengePollViewPageObject extends BasePageObject {
...
   }
 
   async clickAppLink() {
-    await this.t.click(Selector('.al-button'));
+    if(userVariables.v3) {
+      await this.form.clickButton('Open Okta Verify');
+    } else {
+      await this.t.click(this.form.getLink('Open Okta Verify'));
+    }
   }
```
shawyu-okta, line 113: not required, and up to you, but we could combine these two into the `clickLaunchOktaVerifyButton` method below by updating line 119 with the `await this.t.click(this.form.getLink('Open Okta Verify'));` code you used here.

shawyu-okta (approved):

PR #3023, opened 2023-01-17
OKTA-557948 : Fastpass apple SSO custom biometrics error

leemchale-okta:
## Description:

This PR adds the custom biometrics error with bullet points for fastpass apple SSO extension.  This custom error matches parity with v2.

## PR Checklist

### Issue:

- [OKTA-557948](https://oktainc.atlassian.net/browse/OKTA-557948)

### Screenshot/Video:

image

### DIFF
--- a/src/v3/src/components/InfoBox/InfoBox.tsx
+++ b/src/v3/src/components/InfoBox/InfoBox.tsx
@@ -64,12 +64,7 @@ const InfoBox: UISchemaElementComponent<{
       >
         {title && <AlertTitle>{title}</AlertTitle>}
         {message}
-        {bullets?.length > 0
-        && (
-        <Box>
-          <List uischema={bulletList} />
-        </Box>
-        )}
+        {bullets?.length > 0 && <List uischema={bulletList} />}
       </Alert>
     </Box>
   );
--- a/src/v3/src/components/InfoBox/InfoBox.tsx
+++ b/src/v3/src/components/InfoBox/InfoBox.tsx
@@ -58,7 +58,7 @@ const InfoBox: UISchemaElementComponent<{
       >
         { title && <AlertTitle>{title}</AlertTitle> }
         { message }
-        { listOptions && <List uischema={ {type: 'List', options: listOptions } } /> }
+        { listOptions && <List uischema={{ type: 'List', options: listOptions }} /> }
       </Alert>
     </Box>
   );
--- a/src/v3/src/components/InfoBox/InfoBox.tsx
+++ b/src/v3/src/components/InfoBox/InfoBox.tsx
@@ -20,7 +20,6 @@ import { h } from 'preact';
 import { useWidgetContext } from '../../contexts';
 import {
   InfoboxElement,
-  ListElement,
   MessageType,
   MessageTypeVariant,
   UISchemaElementComponent,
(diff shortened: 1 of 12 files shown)


src/v3/src/components/InfoBox/InfoBox.tsx
```diff
@@ -51,7 +63,9 @@ const InfoBox: UISchemaElementComponent<{
         {...({ 'data-se': dataSe })}
         className={`infobox-${messageClass.toLowerCase()}`}
       >
-        {message}
+        { title && <AlertTitle>{title}</AlertTitle> }
```
glenfannin-okta, line 66: Since this is added, I think you can remove the `title` property from line 61 right?

src/v3/src/types/schema.ts
```diff
@@ -400,6 +400,7 @@ export interface InfoboxElement extends UISchemaElement {
     contentType: string;
     title?: string;
     dataSe?: string;
+    bullets?: string[];
```
glenfannin-okta, line 403: I think `contentType` was an incomplete way of trying to implement something like this but given its not used here, I think you can delete it. 

Additionally, I would rename `bullets` to something like `listOptions` and then make the type match the options from the ListElement like 

`listOptions?: ListElement['options']`

src/v3/src/components/InfoBox/InfoBox.tsx
```diff
@@ -51,7 +63,9 @@ const InfoBox: UISchemaElementComponent<{
...
+        { message }
+        { bullets?.length > 0 && <List uischema={bulletList} /> }
```
glenfannin-okta, line 68: If the changes are made to the type (in schema), you can then just check the `listOptions` value for truthiness to display the List element. 

i.e. 
```
{ listOptions && <List uischema={{type: 'List', options={listOptions}}} /> }
```

glenfannin-okta:
> image

cc: @alexdahl-okta Any thought on this design wise? I was thinking maybe some space is needed between the list and the text above it but I am totally sure.

alexdahl-okta:
@glenfannin-okta Hmm we probably don't have to stress it right now?

I agree that there should be more space between the paragraph and list. I'd expect that to be automatic, and I'm hesitant to hack our own fragile solution... I'll start a topic with Odyssey!

src/v3/src/components/InfoBox/InfoBox.tsx
leemchale-okta: Yeah I think this can be removed.  I just wasnt sure if it was still needed for accessibility reasons.

test/testcafe/spec/ChallengeOktaVerifySSOExtensionView_spec.js
```diff
@@ -103,11 +103,11 @@ test
     await t.expect(getStateHandleFromSessionStorage()).eql(null);
   });
 
-// TODO: OKTA-557948 Implement custom biometrics error and enable this test
-test.meta('v3', false)
+test.meta('v3', true)
```
glenfannin-okta, line 106: Since the entire spec is enabled for v3, you can just remove the disable meta flag at the test level and it will automatically be included for v3.
```suggestion
test
```

test/testcafe/spec/ChallengeOktaVerifySSOExtensionView_spec.js
```diff
@@ -119,11 +119,11 @@ test.meta('v3', false)
     await t.expect(errorText).notContains('Your device\'s biometric sensors are accessible');
   });
 
-// TODO: OKTA-557948 Implement custom biometrics error and enable this test
-test.meta('v3', false)
+test.meta('v3', true)
```
glenfannin-okta, line 122:
```suggestion
test
```

glenfannin-okta (approved):

PR #3024, opened 2023-01-18
OKTA-566238 speed up testcafe

lesterchoi-okta:
## Description:

- Increase testcafe concurrency
- Prevent fallback to Dyson mock server by explicitly setting MockRequest

## PR Checklist

### Issue:

- [OKTA-566238](https://oktainc.atlassian.net/browse/OKTA-566238)


package.json
```diff
@@ -77,7 +77,7 @@
     "test:e2e": "yarn workspace @okta/e2e test",
     "test:e2e:lang": "yarn workspace @okta/e2e test:lang",
     "test:testcafe-setup": "mkdir -p test-reports/testcafe && yarn start --mock mockDuo --open false",
-    "test:testcafe-run": "yarn test -t testcafe -c 2 chrome:headless --reporter spec,xunit:build2/reports/junit/testcafe-results.xml",
+    "test:testcafe-run": "yarn test -t testcafe -c 5 chrome:headless --reporter spec,xunit:build2/reports/junit/testcafe-results.xml",
```
glenfannin-okta, line 80: Just curious, any reason why a default concurrency wasn't added to the testcaferc.js config file?

glenfannin-okta (approved):

glenfannin-okta:
Also, with this change, does this eliminate the need to split the tests into groups?

package.json
lesterchoi-okta: TLDR: developer experience

When concurrency is added to the `testcaferc.js` file it'll naively spin up `n` containers regardless of what other flags/args `testcafe` was invoked with, e.g., `-t`, `-T` (we have a vscode launch configs to "Debug with TestCafe" which will typically run one fixture). If we set the default concurrency on the config, it causes a disruptive "flash" as the browsers are opened.

I kept the concurrency here in the script to avoid the need to override with `-c1` when you're just running one test.

lesterchoi-okta:
> Also, with this change, does this eliminate the need to split the tests into groups?

It should for now. We may eventually want to split "slow" tests (with long waits, e.g., resend sms) from the rest, but the suite should now run in about half the time it used to take.

PR #3031, opened 2023-01-23
OKTA-560815 : Add support for engFastpassMultipleAccounts

leemchale-okta:
## Description:

This PR implements the OVFP behavior for when the engFastpassMultipleAccounts widget feature is enabled.

## PR Checklist

### Issue:

- [OKTA-560815](https://oktainc.atlassian.net/browse/OKTA-560815)

### DIFF
--- a/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
+++ b/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
@@ -22,7 +22,7 @@ import {
   LaunchAuthenticatorButtonElement,
   UISchemaElementComponent,
 } from '../../types';
-import { setUrlQueryParams, getTranslation, isAndroid } from '../../util';
+import { getTranslation, isAndroid, setUrlQueryParams } from '../../util';
 import { OktaVerifyIcon } from '../Icon';
 
 const LaunchAuthenticatorButton: UISchemaElementComponent<{
@@ -53,7 +53,7 @@ const LaunchAuthenticatorButton: UISchemaElementComponent<{
       setUserIdentifier(data.identifier as string);
     }
     if (deviceChallengeUrl) {
-      const loginHintQueryParam = userIdentifier ? {login_hint: userIdentifier} : undefined;
+      const loginHintQueryParam = userIdentifier ? { login_hint: userIdentifier } : undefined;
       if (isAndroid() && challengeMethod !== CHALLENGE_METHOD.APP_LINK) {
         Util.redirectWithFormGet(setUrlQueryParams(deviceChallengeUrl, loginHintQueryParam));
       } else {
--- a/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
+++ b/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
@@ -22,7 +22,7 @@ import {
   LaunchAuthenticatorButtonElement,
   UISchemaElementComponent,
 } from '../../types';
-import { appendLoginHint, getTranslation, isAndroid } from '../../util';
+import { setUrlQueryParams, getTranslation, isAndroid } from '../../util';
 import { OktaVerifyIcon } from '../Icon';
 
 const LaunchAuthenticatorButton: UISchemaElementComponent<{
@@ -50,13 +50,14 @@ const LaunchAuthenticatorButton: UISchemaElementComponent<{
   const handleClick: ClickHandler = async () => {
     if (features?.engFastpassMultipleAccounts && data.identifier) {
       // set userIdentifier in widget context to the current Username input field data
-      setUserIdentifier(encodeURIComponent(data.identifier as string));
+      setUserIdentifier(data.identifier as string);
     }
     if (deviceChallengeUrl) {
+      const loginHintQueryParam = userIdentifier ? {login_hint: userIdentifier} : undefined;
       if (isAndroid() && challengeMethod !== CHALLENGE_METHOD.APP_LINK) {
-        Util.redirectWithFormGet(appendLoginHint(deviceChallengeUrl, userIdentifier));
+        Util.redirectWithFormGet(setUrlQueryParams(deviceChallengeUrl, loginHintQueryParam));
       } else {
-        window.location.assign(appendLoginHint(deviceChallengeUrl, userIdentifier));
+        window.location.assign(setUrlQueryParams(deviceChallengeUrl, loginHintQueryParam));
       }
     }
     onSubmitHandler({
--- a/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
+++ b/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
@@ -22,7 +22,9 @@ import {
   LaunchAuthenticatorButtonElement,
   UISchemaElementComponent,
 } from '../../types';
-import { getTranslation, isAndroid, setUrlQueryParams } from '../../util';
+import {
+  getBaseUrl, getTranslation, isAndroid, setUrlQueryParams,
+} from '../../util';
 import { OktaVerifyIcon } from '../Icon';
 
 const LaunchAuthenticatorButton: UISchemaElementComponent<{
@@ -44,6 +46,7 @@ const LaunchAuthenticatorButton: UISchemaElementComponent<{
     loginHint,
     setloginHint,
     data,
+    widgetProps,
   } = useWidgetContext();
 
   const handleClick: ClickHandler = async () => {
@@ -53,10 +56,11 @@ const LaunchAuthenticatorButton: UISchemaElementComponent<{
     }
     if (deviceChallengeUrl) {
       const loginHintQueryParam = loginHint ? { login_hint: loginHint } : undefined;
+      const urlObj = new URL(deviceChallengeUrl, getBaseUrl(widgetProps));
       if (isAndroid() && challengeMethod !== CHALLENGE_METHOD.APP_LINK) {
-        Util.redirectWithFormGet(setUrlQueryParams(deviceChallengeUrl, loginHintQueryParam));
+        Util.redirectWithFormGet(setUrlQueryParams(urlObj, loginHintQueryParam));
       } else {
-        window.location.assign(setUrlQueryParams(deviceChallengeUrl, loginHintQueryParam));
+        window.location.assign(setUrlQueryParams(urlObj, loginHintQueryParam));
       }
     }
     onSubmitHandler({
--- a/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
+++ b/src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
@@ -41,22 +41,22 @@ const LaunchAuthenticatorButton: UISchemaElementComponent<{
 
   const focusRef = useAutoFocus<HTMLButtonElement>(focus);
   const {
-    loginHint,
-    setLoginHint,
+    userIdentifier,
+    setUserIdentifier,
     data,
     widgetProps: { features },
   } = useWidgetContext();
 
   const handleClick: ClickHandler = async () => {
     if (features?.engFastpassMultipleAccounts && data.identifier) {
-      // set loginHint to the current Username input field data
-      setLoginHint(encodeURIComponent(data.identifier as string));
+      // set userIdentifier in widget context to the current Username input field data
+      setUserIdentifier(encodeURIComponent(data.identifier as string));
     }
     if (deviceChallengeUrl) {
       if (isAndroid() && challengeMethod !== CHALLENGE_METHOD.APP_LINK) {
-        Util.redirectWithFormGet(appendLoginHint(deviceChallengeUrl, loginHint));
+        Util.redirectWithFormGet(appendLoginHint(deviceChallengeUrl, userIdentifier));
       } else {
-        window.location.assign(appendLoginHint(deviceChallengeUrl, loginHint));
+        window.location.assign(appendLoginHint(deviceChallengeUrl, userIdentifier));
       }
     }
     onSubmitHandler({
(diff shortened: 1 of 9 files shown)


src/v3/src/types/context.ts
```diff
@@ -42,6 +42,8 @@ export type IWidgetContext = {
   loading: boolean;
   setLoading: StateUpdater<boolean>;
   setWidgetRendered: StateUpdater<boolean>;
+  loginHint?: string;
+  setLoginHint: StateUpdater<string | undefined>;
```
shawyu-okta, line 46: Is there a way to avoid adding these to the widget context? I'm not sure if we need state variables to hold the `loginHint`. I'm thinking it can potentially be pushed out to the transformers and we only use the `widgetContext` to get the feature flag info.

leemchale-okta: The problem was we needed a way for the loginHint/username to persist across views.  Since the `loginHint` is grabbed from the data in the identifier input field, we also need to be able to access it in the next flow with the Open Okta Verify button.

lesterchoi-okta (changes requested):

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -693,8 +690,9 @@ test
     await t.expect(deviceChallengePollPageObject.getFormTitle()).eql('Sign in with Okta FastPass');
 
     await t.expect(await deviceChallengePollPageObject.hasSpinner()).eql(true);
-    await t.expect(deviceChallengePollPageObject.getFooterCancelPollingLink().innerText).eql('Cancel and take me to sign in');
+    await t.expect(deviceChallengePollPageObject.getFooterCancelPollingLink().exists).eql(true);
 
+    await t.wait(5000); // wait for FASTPASS_FALLBACK_SPINNER_TIMEOUT
```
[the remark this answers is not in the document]
leemchale-okta: usually you can use the `waitFor` function to achieve that, but for some reason the `waitFor` function is not included in this testcafe context

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
```diff
@@ -39,13 +40,23 @@ const LaunchAuthenticatorButton: UISchemaElementComponent<{
...
+    userIdentifier,
+    setUserIdentifier,
+    data,
+    widgetProps: { features },
+  } = useWidgetContext();
 
   const handleClick: ClickHandler = async () => {
+    if (features?.engFastpassMultipleAccounts && data.identifier) {
+      // set userIdentifier in widget context to the current Username input field data
+      setUserIdentifier(encodeURIComponent(data.identifier as string));
```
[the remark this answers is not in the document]
shawyu-okta: Just double-checking - is our ES6 polyfill configured to include this? Assuming we're using `core-js` then it should be included since it's part of most of the usual entry points: https://github.com/zloirock/core-js#url-and-urlsearchparams

src/v3/src/util/appendLoginHint.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+export const appendLoginHint = (deviceChallengeUrl = '', loginHint = ''): string => {
+  if (deviceChallengeUrl && loginHint) {
+    return `${deviceChallengeUrl}&login_hint=${loginHint}`;
+  }
+  return deviceChallengeUrl;
+};
```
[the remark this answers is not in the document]
shawyu-okta: Agree on both counts, we can rename the fn and also improve on the original implementation in v2 by using `URL` and `URLSearchParams` while we're at it

`URL`: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL
`URL.searchParams` is a `URLSearchParams` object: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

src/v3/src/components/Widget/index.tsx
```diff
@@ -282,6 +283,8 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
       loading,
       setLoading,
       setWidgetRendered,
+      userIdentifier,
+      setUserIdentifier,
```
shawyu-okta, line 287: We need to make sure this value is cleared in the case that a user clears the username field (should go back to `null`/`undefined` OR we need to check for empty strings)

Also if a user backs out of a flow by clicking the back button in the widget and the widget shows a blank username field again, we need to ensure the value is cleared as well.

Finally, what happens if the username is prefilled via widgetConfig or rememberMe functionality?

src/v3/src/components/Widget/index.tsx
```diff
@@ -96,6 +96,7 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
   const [loading, setLoading] = useState<boolean>(false);
   const [widgetRendered, setWidgetRendered] = useState<boolean>(false);
   const brandedTheme = mapMuiThemeFromBrand(brandColors);
+  const [userIdentifier, setUserIdentifier] = useState<string | undefined>(undefined);
```
shawyu-okta, line 99: To go along with the comment below, if we still want to allow non-string values `null | undefined` I favor using `null` when the value is intentionally empty.

Another way to go is _only_ allow `string` and check for empty strings in the component code before appending the parameters to the URL.

src/v3/src/util/appendLoginHint.ts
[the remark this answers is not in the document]
leemchale-okta: I found a problem in that this breaks the spec test since the spec test uses `okta-verify.html` as the dummy url.  The `URL` object constructor does not accept that as a valid url and throws an error:
image

How do we want to handle this do we want to change the test mock file?

src/v3/src/components/Widget/index.tsx
leemchale-okta: if the username is prefilled via widgetConfig or rememberMe it will still work properly since it sets the `userIdentifier` value to whatever is in the username field only after the launch authenticator button is pressed.  The username field data is populated before the authenticator button is pressed.  

Since the `userIdentifier` value is set only after the launch auth button is pressed, we shouldnt have any issue of possibly having the wrong username set in it or the user clearing the field.  So it shouldnt be a problem if someone backs out of the flow, but technically the old username would still be leftover in `userIdentifier` until the launch auth button is pressed again which may be a security concern.  I think a simple solution to this would be to clear the `userIdentifier` as soon as the identify flow loads.

shawyu-okta: > if the username is prefilled via widgetConfig or rememberMe it will still work properly since it sets the userIdentifier value to whatever is in the username field only after the launch authenticator button is pressed. The username field data is populated before the authenticator button is pressed.

👍 

> So it shouldnt be a problem if someone backs out of the flow, but technically the old username would still be leftover in userIdentifier until the launch auth button is pressed again which may be a security concern. I think a simple solution to this would be to clear the userIdentifier as soon as the identify flow loads.

Cool, I think as a part of good hygiene we should clear out these widget context values properly in this case even if it's not affecting the widget behavior in the current use-case.

src/v3/src/types/context.ts
```diff
@@ -42,6 +42,8 @@ export type IWidgetContext = {
   loading: boolean;
   setLoading: StateUpdater<boolean>;
   setWidgetRendered: StateUpdater<boolean>;
+  userIdentifier?: string;
+  setUserIdentifier: StateUpdater<string | undefined>;
```
glenfannin-okta, line 46: Any thoughts on this name? In v2 it is set as `login_hint`, we don't have to mimic this exactly, but userIdentifier is a bit misleading because it may not actually be the real identifier, and it's only used in specific scenarios. I just don't want this to be confused later down the road with something it isn't.

leemchale-okta: Thats a good point since we are grabbing the username from the input field, it may not necessarily be a valid username.  I believe it was changed because we wanted to make the name more generic.  @shawyu-okta what do you think?

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
```diff
@@ -39,13 +40,24 @@ const LaunchAuthenticatorButton: UISchemaElementComponent<{
...
   const focusRef = useAutoFocus<HTMLButtonElement>(focus);
+  const {
+    userIdentifier,
+    setUserIdentifier,
+    data,
+    widgetProps: { features },
+  } = useWidgetContext();
 
   const handleClick: ClickHandler = async () => {
+    if (features?.engFastpassMultipleAccounts && data.identifier) {
```
lesterchoi-okta, line 51:
```suggestion
    if (data.identifier) {
```

lesterchoi-okta: Per latest comment in Jira, we are removing this flag check

lesterchoi-okta: https://github.com/okta/okta-signin-widget/pull/3034

src/v3/src/types/context.ts
lesterchoi-okta: Anything wrong with the obvious alternative?

```
  loginHint?: string;
  setLoginHint: StateUpdater<string | undefined>;
```

shawyu-okta: Yeah I had @leemchale-okta change it to `userIdentifier` because I wanted to avoid a very specific global context value if possible, but based on the comments above I don't have an issue reverting to `loginHint`

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
shawyu-okta: Let's remove the check here and update the tests in the same fashion as in #3034. It will cause some conflicts on rebase/merge to the trunk but they shouldn't be difficult to resolve.

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -168,6 +169,10 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
           i18n: { key: 'oform.errorbanner.title' },
         } as IdxMessage);
       }
+      // clear the userIdentifier value when returning to the identify flow
+      if (newTransaction.nextStep?.name === IDX_STEP.IDENTIFY) {
+        setUserIdentifier(null);
+      }
```
shawyu-okta, line 175: It might be better to put this setter somewhere where we react to the transaction changing, rather than where we are setting the changed transaction, if that makes sense. I view this hook as a place where we are updating the transaction value, whereas the bit about clearing the userIdentifier value is a reaction to when that changes.

We have some effect hooks that are in `Widget/index` that might be more appropriate since they are run whenever transaction changes.

src/v3/src/components/OpenOktaVerifyFPButton/OpenOktaVerifyFPButton.tsx
```diff
@@ -53,6 +54,9 @@ const OpenOktaVerifyFPButton: UISchemaElementComponent<{
   } = uischema;
   const [key, setKey] = useState<number>(0);
   const label = getTranslation(translations, 'label');
+  const { userIdentifier } = useWidgetContext();
+  const loginHintQueryParam = userIdentifier ? { login_hint: userIdentifier } : undefined;
```
shawyu-okta, line 58: nit: consider `null`
```suggestion
  const loginHintQueryParam = userIdentifier ? { login_hint: userIdentifier } : null;
```

src/v3/src/types/context.ts
glenfannin-okta: Ahh, that makes sense, my only concern was if a home team eng comes across this, they may think this is some quick way to access the user's identifier. Granted that could be dispelled rather quickly with a codebase search, but just wanted to get your opinion on this. I am good with either way 🙏

src/v3/src/util/setUrlQueryParams.ts
```diff
@@ -0,0 +1,16 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+export const setUrlQueryParams = (
+  url: URL | string,
+  params: Record<string, string> = {},
+) => url + '?' + new URLSearchParams(params).toString();
```
shawyu-okta, line 16: https://github.com/okta/okta-signin-widget/pull/3031#discussion_r1087255593
> I found a problem in that this breaks the spec test since the spec test uses okta-verify.html as the dummy url. The URL object constructor does not accept that as a valid url and throws an error:

There's a corner case here where if the `url` already has the `?` in it then this might return an invalid URL, but in practice I don't think we'll run into it.

Using the `URL` class should mitigate this and you might be able to use the `URL(url, base)` overload of the constructor and pass the widget `baseUrl` to it. We have a utility in `settingsUtils` called `getBaseUrl(widgetProps)`, although it looks like it can still return `undefined`.

shawyu-okta (approved):
Only blocking comments to address are to remove the FF / update tests, changing the context value for identifier to be named hint, and the location of where we clear out the identifier/hint value.

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
leemchale-okta: 👍

src/v3/src/hooks/useOnSubmit.ts
leemchale-okta: Okay yeah I see what youre saying.  The `useOnSubmit` seemed like a reliable place to put it but the `Widget/index` is probably more appropriate.

src/v3/src/types/context.ts
leemchale-okta: okay makes sense I will change it to `  loginHint?: string;
  setLoginHint: StateUpdater<string | undefined>;`

src/v3/src/components/OpenOktaVerifyFPButton/OpenOktaVerifyFPButton.tsx
```diff
@@ -53,6 +56,10 @@ const OpenOktaVerifyFPButton: UISchemaElementComponent<{
   } = uischema;
   const [key, setKey] = useState<number>(0);
   const label = getTranslation(translations, 'label');
+  const { loginHint, widgetProps } = useWidgetContext();
+  const urlObj = new URL(href as string, getBaseUrl(widgetProps));
```
lesterchoi-okta, line 60: 👍

lesterchoi-okta (approved):

PR #3035, opened 2023-01-24
OKTA-557574 : Implement autoFocus widget option

glenfannin-okta:
## Description:
The purpose of this PR is to implement the autoFocus widget option and set the default value as false. Product and design requested this to assist users with disabilities / using screen readers preventing the app from automatically reading out the auto focused field and allowing the user to scan through the page from top to bottom to gather content/instructions. 

## PR Checklist

### Issue:

- [OKTA-557574](https://oktainc.atlassian.net/browse/OKTA-557574)

### Screenshot/Video:
https://user-images.githubusercontent.com/97472729/214378165-154722c4-8ca7-4bd9-9aef-84333efbdbd2.mov


src/v3/src/types/schema.ts
```diff
@@ -411,14 +411,6 @@ export interface SuccessCallback extends UISchemaElement {
   }
 }
 
-export type StepperNavButtonConfigDirection = 'next' | 'prev';
-
-export type StepperNavButtonConfigAttrs = {
-  variant?: 'primary' | 'secondary';
-  label?: string;
-  id?: string;
-};
```
glenfannin-okta, line 420: These are no longer used.

src/v3/src/types/schema.ts
```diff
@@ -432,7 +424,7 @@ export interface AccordionLayout {
   elements: AccordionPanelElement[];
 }
 
-export interface StepperButtonElement {
+export interface StepperButtonElement extends UISchemaElement {
```
glenfannin-okta, line 427: These need to extend UISchemaElement because there are properties (i.e. focus) needed in dependent code.

shawyu-okta (commented):
Is there a way this can be implemented so that rather than having to use the FF as to guard adding `focus` in the transformer layer to every element, we instead ignore the autofocus behavior within each component if the FF is off?

glenfannin-okta:
> Is there a way this can be implemented so that rather than having to use the FF as to guard adding `focus` in the transformer layer to every element, we instead ignore the autofocus behavior within each component if the FF is off?

Oh yes actually, this can totally be done in the useAutoFocus hook instead and would only result in a single change in that singular location. 🙏 Will update! Ty!

src/v3/src/util/settingsUtils.ts
```diff
@@ -95,3 +95,10 @@ export const getBaseUrl = (widgetProps: WidgetProps): string | undefined => {
   const [parsedBaseUrl] = issuerPath?.split('/oauth2/') ?? [];
   return parsedBaseUrl;
 };
+
+export const getAutoFocus = (widgetProps: WidgetProps): boolean => {
+  // Product requests autoFocus default to be false
+  const { features: { autoFocus = false } = {} } = widgetProps;
+
+  return autoFocus;
+};
```
shawyu-okta, line 104: I think we could do without this hook now that it's only used in one place

src/v3/src/types/widget.ts
```diff
@@ -189,6 +189,7 @@ export type OktaWidgetFeatures = {
   securityImage?: boolean;
   rememberMe?: boolean;
   autoPush?: boolean;
+  autoFocus?: boolean;
```
shawyu-okta, line 192: Curious if there was already some discussion about the name of this feature. This seems in line with the other feature names but I also considered `allowAutoFocus`. Seeing the other ones I am more convinced this is the right name though.

src/v3/test/integration/authenticator-verification-email.test.tsx
```diff
@@ -49,13 +53,18 @@ describe('Email authenticator verification when email magic link = undefined', (
...
         mockResponse: authenticatorVerificationEmail,
+        widgetOptions: { features: { autoFocus: true } },
       });
 
       // renders the form
       await findByText(/Verify with your email/);
+      const codeEntryBtn = await findByRole(
+        'button', { name: 'Enter a code from the email instead' },
+      ) as HTMLButtonElement;
+      await waitFor(() => expect(codeEntryBtn).toHaveFocus());
```
shawyu-okta, line 67: Is the `waitFor` necessary because the focus behavior happens very minimally later than the completion of widget render (due to the effect hook)?

shawyu-okta (approved):

src/v3/src/types/widget.ts
glenfannin-okta: This is actually already an existing [feature](https://github.com/okta/okta-signin-widget/blob/master/src/models/Settings.ts#L109) that v2 uses, I assumed we will want to keep the same name so v3 is backwards compatible with existing features.

src/v3/test/integration/authenticator-verification-email.test.tsx
glenfannin-okta: Yea, without the `waitFor` it fails (sometimes), there is a slight delay between when the element renders and when the focus happens.

src/v3/src/types/widget.ts
shawyu-okta: Ah okay great, thanks for attaching the link

PR #3038, opened 2023-01-24
V3 parity spec for Challenge Phone Authenticator

vijetmahabaleshwar-okta:
## PR Checklist

### Issue:

- [OKTA-509024](https://oktainc.atlassian.net/browse/OKTA-509024)


test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
```diff
@@ -145,6 +158,7 @@ async function setupInteractionCodeFlow(t) {
 test
   .requestHooks(smsPrimaryMock)('SMS primary mode - has the right labels', async t => {
     const challengePhonePageObject = await setup(t);
+    await t.expect(challengePhonePageObject.formExists()).eql(true);
```
glenfannin-okta, line 161: I would move this into the `setup` function so it is available to all tests.

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
```diff
@@ -155,7 +169,7 @@ test
     const pageTitle = challengePhonePageObject.getFormTitle();
     const pageSubtitle = challengePhonePageObject.getFormSubtitle();
     const primaryButtonText = challengePhonePageObject.getSaveButtonLabel();
-    const secondaryButtonText = challengePhonePageObject.getSecondaryLinkText();
+    const secondaryButtonText = challengePhonePageObject.getSecondaryLinkText('Receive a voice call instead');
```
glenfannin-okta, line 172: Instead of getting the text using the text and then asserting on it, I think it would be simpler to just assert the button with this text exists. 

You can use this function from the BaseFormObject: `getButton` and just pass in the button label like: `getButton('Receive a voice call instead');` 

So ultimately, it would be like this: 

```
await t.expect(
   await challengePhonePageObject.form.getButton('Receive a voice call instead').exists
).eql(true);
```

Then you can get rid of the `getSecondaryLinkText` since this should work for both v2 and v3.

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
```diff
@@ -221,6 +235,7 @@ test
   .requestHooks(voicePrimaryMock)('Voice primary mode - has the right labels', async t => {
     const challengePhonePageObject = await setup(t);
 
+    await t.expect(challengePhonePageObject.formExists()).eql(true);
```
glenfannin-okta, line 238: Can remove if you put this into the `setup` function.

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
```diff
@@ -512,42 +527,42 @@ test
   .requestHooks(invalidCodeMock)('Entering invalid passcode results in an error', async t => {
     const challengePhonePageObject = await setup(t);
     await challengePhonePageObject.verifyFactor('credentials.passcode', 'abcd');
-    await challengePhonePageObject.clickNextButton();
+    await challengePhonePageObject.clickNextButton('Verify');
     await challengePhonePageObject.waitForErrorBox();
-    await t.expect(challengePhonePageObject.getInvalidOTPFieldError()).contains('Invalid code. Try again.');
+    // await t.expect(challengePhonePageObject.getInvalidOTPFieldError()).contains('Invalid code. Try again.');
```
glenfannin-okta, line 532: Is the field level error not showing up?

vijetmahabaleshwar-okta: Yeah, this error doesn't show up in v3 for some reason. I verified that it's not an actual bug in production though. Not sure if it's some issue with the mocks.

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
vijetmahabaleshwar-okta: Screen Shot 2023-01-25 at 2 21 04 PM
Not working in v2, since the secondary text is NOT a button in v2. Might have to stick to the old way

glenfannin-okta: Ahh, that stinks, ok, no worries.

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
```diff
@@ -512,42 +527,48 @@ test
...
     const challengePhonePageObject = await setup(t);
     await challengePhonePageObject.verifyFactor('credentials.passcode', 'abcd');
-    await challengePhonePageObject.clickNextButton();
+    await challengePhonePageObject.clickVerifyButton();
     await challengePhonePageObject.waitForErrorBox();
-    await t.expect(challengePhonePageObject.getInvalidOTPFieldError()).contains('Invalid code. Try again.');
+    // in v3, the "invalid code" field error is not shown. Test mock issue, not product issue
+    if (!userVariables.v3) {
+      await t.expect(challengePhonePageObject.getInvalidOTPFieldError()).contains('Invalid code. Try again.');
+    }
```
glenfannin-okta, line 535: If you could create a jira and comment here so we can come back and clean up the mock. We try to avoid adding this v3 conditional in the spec unless there is a true difference between v2 and v3.

glenfannin-okta (approved):

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
vijetmahabaleshwar-okta: This is fixed now with the new mock file

PR #3040, opened 2023-01-25
OKTA-564968 : Adds signed nonce error title

leemchale-okta:
## Description:

This PR adds a custom title to signed nonce errors to match parity with v2.

## PR Checklist

### Issue:

- [OKTA-564968](https://oktainc.atlassian.net/browse/OKTA-564968)

### Screenshot/Video:

image


src/v3/src/components/InfoBox/InfoBox.tsx
```diff
@@ -56,7 +56,14 @@ const InfoBox: UISchemaElementComponent<{
         {...({ 'data-se': dataSe })}
         className={`infobox-${messageClass.toLowerCase()}`}
       >
-        { title && <AlertTitle>{title}</AlertTitle> }
+        { title && (
+        <AlertTitle
+          role="heading"
+          aria-level="3"
```
leemchale-okta, line 62: Since v2 uses an `<h3>` header tag for its error title, this would match parity for accessibility and testing.  Correct me if im wrong

test/testcafe/spec/DeviceChallengePollViewFailure_spec.js
```diff
@@ -145,12 +145,12 @@ test.requestHooks(logger, mock)('probing and polling APIs are sent and responded
...
 test
-  .meta('v3', false) // Need to add title when this specific type of error is returned (OKTA-564968)
   .requestHooks(logger, deviceInvalidatedErrorMsg)('add title when device or account is invalidated', async t => {
     mockCalls = 0;
     const deviceChallengePollPageObject = await setup(t);
-    await t.expect(deviceChallengePollPageObject.form.getErrorBoxText()).eql(
-      'Couldn’t verify your identity\n\nYour device or account was invalidated. If this is unexpected, contact your administrator for help.');
+    await t.expect(deviceChallengePollPageObject.form.getErrorBoxTitle()).eql('Couldn’t verify your identity');
+    await t.expect(deviceChallengePollPageObject.hasText(
+      'Your device or account was invalidated. If this is unexpected, contact your administrator for help.')).eql(true);
```
leemchale-okta, line 153: made separate assertions for title and description text due to v2 error text containing more new line characters.

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -259,6 +259,12 @@ export default class BaseFormObject {
     return errors[index];
   }
 
+  getErrorBoxTitle() {
```
glenfannin-okta, line 262: I would either rename this to something more generic (i.e. do not include ErrorBox since the result of the function would not matter if its in an alert box) or update the function to first find by role using "alert" and then within the alert box search for the Title/heading element.

glenfannin-okta (commented):
I would also add a new test case to the transformTerminalMessages spec to cover this new case.

test/testcafe/framework/page-objects/components/BaseFormObject.js
leemchale-okta: oh yeah good call!

src/v3/src/components/InfoBox/InfoBox.tsx
shawyu-okta: In case anyone is curious where these allowed props are coming from like I was, `AlertTitle` is a styled version of the `Typography` component, so can be passed any properties that are allowed on `Typography`. The MUI documentation doesn't make that clear at all so I had to dig into the source code to find that 🙂

shawyu-okta: Does this work if we apply `component` and `variant='h3'` instead? If so it should come with free implied role/aria-level attributes.

leemchale-okta: > In case anyone is curious where these allowed props are coming from like I was, `AlertTitle` is a styled version of the `Typography` component, so can be passed any properties that are allowed on `Typography`. The MUI documentation doesn't make that clear at all so I had to dig into the source code to find that 🙂

awesome didnt know that thanks for that!

leemchale-okta: > 

`component` and `variant='h3'` dont seem to be valid for `<AlertTitle>` and I can't find anything in the docs that says otherwise

leemchale-okta: I think maybe because we are using the `<AlertTitle>` from `@okta/odyssey-react-mui`?

src/v3/src/transformer/terminal/transformTerminalMessages.ts
```diff
@@ -34,9 +34,10 @@ type ModifiedIdxMessage = Modify<IdxMessage, {
     key: string;
     params?: unknown[];
   };
+  title?: string;
 }>;
 
-const appendMessageElements = (uischema: UISchemaLayout, messages: IdxMessage[]): void => {
+const appendMessageElements = (uischema: UISchemaLayout, messages: ModifiedIdxMessage[]): void => {
```
lesterchoi-okta, line 40: NOTE: Adding this property forces the type change on `appendMessageElements()`. Otherwise the `Modify` is just making those properties optional.

Not much to do about this unless/until we overhaul this transformer. The if/else chain below smells and should be refactored.

src/v3/src/components/InfoBox/InfoBox.tsx
```diff
@@ -56,7 +56,13 @@ const InfoBox: UISchemaElementComponent<{
         {...({ 'data-se': dataSe })}
         className={`infobox-${messageClass.toLowerCase()}`}
       >
-        { title && <AlertTitle>{title}</AlertTitle> }
+        { title && (
+        <AlertTitle
+          role="heading"
```
lesterchoi-okta, line 61: Is there a way to use a semantic tag instead of applying an aria attr? They should generally be used as a last-resort. Do we need to talk to Odyssey?

lesterchoi-okta (approved):
Looks ok since the PR is targeting just the missing title for the spec. We'll come back to clean up this transformer.

src/v3/src/components/InfoBox/InfoBox.tsx
shawyu-okta: As far as I can tell the `AlertTitle` in `@okta/odyssey-react-mui` is just directly reexported from `@mui/material`

shawyu-okta: Oh, it doesn't look like the MUI material AlertTitle passes all the props through to the inner Typography component :/
https://github.com/mui/material-ui/blob/5d387c0e588c163a0c8c159ab53243b9d895a10b/packages/mui-material/src/AlertTitle/AlertTitle.js#L20-L52

src/v3/test/integration/__snapshots__/okta-verify-enrollment-version-upgrade.test.tsx.snap
```diff
@@ -138,7 +138,9 @@ exports[`okta-verify-enrollment-version-upgrade should render form 1`] = `
                       class="MuiAlert-message emotion-18"
                     >
                       <div
+                        aria-level="3"
```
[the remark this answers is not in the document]
leemchale-okta: Sorry just an old snapshot i forgot to update

src/v3/src/components/InfoBox/InfoBox.tsx
shawyu-okta: This is from MUI unfortunately, not Odyssey. It looks like MUI uses a `div` with styling for this rather than use a semantic heading element.

image

shawyu-okta: See also this hidden thread: https://github.com/okta/okta-signin-widget/pull/3040#discussion_r1086209062

leemchale-okta: `role="heading"` is no longer needed and has been removed.  `AlertTitle` should already have the necessary props built in.

leemchale-okta: ah yeah thats why the `role="heading"` was needed since MUI doesnt use a semantic title

leemchale-okta: I suppose an alternative would be to use the Typography component?

shawyu-okta: Does it look OK if we use Typography? I wanted to use AlertTitle since it seems to be tailor made for the Alert component but if it isn't suiting our needs we can look at the other components.

PR #3045, opened 2023-01-27
OKTA-561820 : MUI Theme options override implementation in SIW v3 

glenfannin-okta:
## Description:
The purpose of this PR is to provide customers/orgs the ability to override any MUI/ODS theme option via the widget config options.

## PR Checklist

### Issue:

- [OKTA-561820](https://oktainc.atlassian.net/browse/OKTA-561820)

### Screenshot/Video:
Screen Shot 2023-01-27 at 1 46 11 PM
Screen Shot 2023-01-27 at 1 46 40 PM


src/v3/src/types/widget.ts
```diff
@@ -79,6 +80,8 @@ export type WidgetOptions = {
 
   // theming
   theme?: ThemeOptions;
+  // Override MUI Theming
+  muiThemeOptions?: MuiThemeOptions;
```
glenfannin-okta, line 84: Not entirely sure what to name this since there is already a `theme` property. Also, not sure if I should have included it inside the existing `theme` property. Looking for feedback on these.

alexdahl-okta: 👏 I'll defer to developers!

Separate properties for different kinds of theming is probably helpful clarity?

On one hand, I'd prefer MUI to be an implementation detail that customers don't have to worry about. But on the other hand, if the API is defined by MUI and we'll be sending them to MUI docs for reference, having MUI in the name is probably helpful clarity?

glenfannin-okta: Exactly this, that is where I struggled because I wanted to keep MUI abstracted, however, since we don't control the properties/API that is needed to be passed in, I figured it would make more sense to have `mui` in the name so the enduser can make the correlation. 

Let's see what others think, I am torn with this as well 😅

src/v3/test/e2e/tests/theming.ts
```diff
@@ -29,19 +29,17 @@ test('Theme configuration applies correctly', async (t) => {
     .typeText("input[data-se='credentials.passcode']", 'password')
     .click(submitButton);
 
-  // TODO: Assertion is skipped until we update auth coin component with theming
```
glenfannin-okta, line 32: Verified this is now working as expected

src/v3/test/e2e/tests/theming.ts
```diff
@@ -13,7 +13,7 @@
 import { Selector } from 'testcafe';
 
 fixture('Theming')
-  .page('http://localhost:3000/?siw-use-mocks=true&siw-mock-response=/idp/idx/introspect/default');
```
glenfannin-okta, line 16: Changed to use scenario because this mock does not allow navigation to different screens (as the test is attempting to do).

src/v3/src/types/widget.ts
shawyu-okta: I think leaving `mui` makes sense, since in some place we will likely need to doc that it's MUI underlying anyways.

Any thoughts on using `overrides` rather than `options`?

glenfannin-okta: I actually prefer `overrides` lol (I used it as the param name to the function) not sure why I put options here 😅

src/v3/src/util/theme.ts
```diff
@@ -113,8 +117,9 @@ export const mapMuiThemeFromBrand = (brand: BrandColors | undefined): Theme => {
       odysseyThemeCopy.palette.text.primary = derivedTheme.textColor;
     }
   }
+
   // @ts-expect-error Error thrown from difference of component props between ODS and MUI
-  return odysseyThemeCopy;
+  return merge(odysseyThemeCopy, muiThemeOverrides || {});
```
shawyu-okta, line 122: I think you can omit the `|| {}`

In Node REPL:
```node
> lodash.merge({}, undefined)
{}
```

shawyu-okta (approved):

PR #3050, opened 2023-01-30
V3 parity spec for enroll phone authenticator data view

vijetmahabaleshwar-okta:
## PR Checklist

### Issue:

- [OKTA-509015](https://oktainc.atlassian.net/browse/OKTA-509015)

### DIFF
--- a/test/testcafe/framework/page-objects/EnrollPhonePageObject.js
+++ b/test/testcafe/framework/page-objects/EnrollPhonePageObject.js
@@ -4,7 +4,8 @@ import { userVariables } from 'testcafe';
 const PASSCODE_FIELD_NAME = 'credentials.passcode';
 const PHONE_NUMBER_SELECTOR = '.phone-authenticator-enroll__phone';
 const PHONE_NUMBER_EXTENSION_SELECTOR = '.phone-authenticator-enroll__phone-ext';
-const phoneFieldName = 'authenticator\\.phoneNumber';
+const PHONE_FIELD_NAME = 'authenticator\\.phoneNumber';
+const PHONE_CODE_FIELD_NAME = "phoneCode";
 const RESEND_VIEW_SELECTOR = '.phone-authenticator-enroll--warning';
 
 export default class EnrollAuthenticatorPhonePageObject extends BasePageObject {
@@ -17,18 +18,12 @@ export default class EnrollAuthenticatorPhonePageObject extends BasePageObject {
     return this.form.selectRadioButtonOptionByValue('authenticator\\.methodType', methodType);
   }
 
-  extensionIsHidden() {
+  async extensionIsHidden() {
     if (userVariables.v3) {
-      return !this.form.fieldByLabelExists('Extension');
+      const exists = await this.form.fieldByLabelExists('Extension');
+      return !exists;
     }
     return this.form.getElement(PHONE_NUMBER_EXTENSION_SELECTOR).hasClass('hide');
-  }
-
-  extensionText() {
-    if (userVariables.v3) {
-      return this.form.getElement('label[for="phoneExtension"]').innerText;
-    }
-    return this.form.getElement(PHONE_NUMBER_EXTENSION_SELECTOR).innerText;
   }
 
   countryCodeText() {
@@ -44,7 +39,10 @@ export default class EnrollAuthenticatorPhonePageObject extends BasePageObject {
   }
 
   hasPhoneNumberError() {
-    return this.form.hasTextBoxErrorMessage(phoneFieldName);
+    if (userVariables.v3) {
+      return this.form.hasTextBoxErrorMessage(PHONE_FIELD_NAME);    
+    }
+    return this.form.hasTextBoxErrorMessage(PHONE_CODE_FIELD_NAME);
   }
 
   clickSaveButton(name) {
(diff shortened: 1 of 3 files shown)


test/testcafe/spec/EnrollAuthenticatorDataPhoneView_spec.js
```diff
@@ -59,12 +64,12 @@ test.requestHooks(mock)('voice mode click and extension will get shown', async t
   await t.expect(enrollPhonePage.getSaveButtonLabel()).eql('Receive a code via voice call');
 
   // Extension field is shown
-  await t.expect(enrollPhonePage.extensionIsHidden()).eql(false);
-  const extensionText = await enrollPhonePage.getElement('.phone-authenticator-enroll__phone-ext').innerText;
+  await t.expect(await enrollPhonePage.extensionIsHidden()).eql(false);
+  const extensionText = await enrollPhonePage.extensionText();
   await t.expect(extensionText.trim()).eql('Extension');
```
glenfannin-okta, line 69: Instead of these where we get the text then assert that it is what we expect, I think we could make is simpler by just trying to get the field by the label/text and if it isnt there it will fail. 

We have [this ](https://github.com/okta/okta-signin-widget/blob/0.1/test/testcafe/framework/page-objects/components/BaseFormObject.js#L34)function in the baseformObject. Then you can use it here like:
```
await t.expect(await enrollPhonePage.formFieldExistsByLabel('Extension')).eql(true);
```

test/testcafe/spec/EnrollAuthenticatorDataPhoneView_spec.js
```diff
@@ -42,13 +48,12 @@ test.requestHooks(mock)('default sms mode', async t => {
   await t.expect(enrollPhonePage.getSaveButtonLabel()).eql('Receive a code via SMS');
 
   // Extension field is hidden
-  await t.expect(enrollPhonePage.extensionIsHidden()).eql(true);
+  await t.expect(await enrollPhonePage.extensionIsHidden()).eql(false);
```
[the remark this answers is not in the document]
vijetmahabaleshwar-okta: Wasn't meant to be changed. I've fixed it but I should have created this as a draft PR. 
Thanks for reviewing it early :D

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -32,7 +32,7 @@ export default class BaseFormObject {
   }
 
   fieldByLabelExists(label, options = undefined) {
-    return within(this.el).getByLabelText(new RegExp(label), options).exists;
```
vijetmahabaleshwar-okta, line 35: fieldByLabelExists() will fail for cases where label doesn't exist, if we use `getByLabelText()`
Using `queryByLabelText()` allows us to use this method for non existent label scenarios

glenfannin-okta: Ohh yes 🤕 

I'll leave it up to you if you feel its worth adding that function. 🙏

test/testcafe/spec/EnrollAuthenticatorDataPhoneView_spec.js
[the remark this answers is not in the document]
glenfannin-okta: Ahh, sorry about that hahaa

test/testcafe/framework/page-objects/EnrollPhonePageObject.js
```diff
@@ -38,10 +57,13 @@ export default class EnrollAuthenticatorPhonePageObject extends BasePageObject {
...
 
   fillPhoneNumber(value) {
-    return this.form.setTextBoxValue(phoneFieldName, value);
+    return this.form.setTextBoxValue(PHONE_FIELD_NAME, value);
   }
 
   phoneNumberFieldIsSmall() {
+    if (userVariables.v3) {
+      return this.form.elementExist('[inputmode="tel"]');
+    }
```
glenfannin-okta, line 66: nit: 
so it seems like this class `phone-authenticator-enroll__phone--small` is very specific to v2, the purpose of this is to make the input box shorter in length to allow the extension input field to fit on the same row. 

In v2, we don't really need to do this check because we are using flexbox for alignment, and that automatically happens with the css we are using.

What I would do to not cause confusion is to leave this function as it was, and then in the test, wrap this assertion with the `userVariables.v3` condition to only execute it when it isn't v3.

test/testcafe/spec/EnrollAuthenticatorDataPhoneView_spec.js
```diff
@@ -116,15 +120,15 @@ test.requestHooks(voiceOnlyOptionMock)('default is voice mode', async t => {
   await t.expect(enrollPhonePage.getSaveButtonLabel()).eql('Receive a code via voice call');
 
   // Extension field is not hidden
-  await t.expect(enrollPhonePage.extensionIsHidden()).eql(false);
+  await t.expect(await enrollPhonePage.extensionIsHidden()).eql(false);
   // Phone field is rendered small
   await t.expect(enrollPhonePage.phoneNumberFieldIsSmall()).eql(true);
```
glenfannin-okta, line 125: nit: 
```suggestion
  // Add comment here saying that this is only applicable to v2
  // since v3 uses flexbox to render phone and extension fields on same line.
  if (!userVariables.v3) {
    await t.expect(enrollPhonePage.phoneNumberFieldIsSmall()).eql(true);
  }
```

glenfannin-okta (approved):

PR #3065, opened 2023-02-01
Handles non-IDX error responses

shawyu-okta:
## Description:

This PR updates the gen3 widget to handle non-ion/non-IDX error responses from the server, particularly when the JSON response is `{}` with a non-success HTTP code.

## PR Checklist

### Issue:

- [OKTA-564970](https://oktainc.atlassian.net/browse/OKTA-564970)

### DIFF
--- a/src/v3/src/components/LoopbackProbe/LoopbackProbe.test.tsx
+++ b/src/v3/src/components/LoopbackProbe/LoopbackProbe.test.tsx
@@ -31,6 +31,7 @@ jest.mock('../../../src/contexts', () => ({
         stateHandle: 'fake-state-handle',
       },
     },
+    setIdxTransaction: jest.fn(),
   }),
 }));
 jest.mock('../../../../util/Logger');
@@ -84,10 +85,7 @@ describe('LoopbackProbe', () => {
     await waitFor(() => expect(proceedStub).toHaveBeenCalledTimes(1), { timeout: 100 });
 
     expect(proceedStub).toHaveBeenCalledWith({
-      actions: [{
-        name: step,
-        params: undefined,
-      }],
+      step,
       stateHandle: 'fake-state-handle',
     });
   });
@@ -204,10 +202,7 @@ describe('LoopbackProbe', () => {
     await waitFor(() => expect(proceedStub).toHaveBeenCalledTimes(1), { timeout: 100 });
 
     expect(proceedStub).toHaveBeenCalledWith({
-      actions: [{
-        name: 'device-challenge-poll',
-        params: undefined,
-      }],
+      step: 'device-challenge-poll',
       stateHandle: 'fake-state-handle',
     });
   });
--- a/src/v3/src/components/LoopbackProbe/LoopbackProbe.test.tsx
+++ b/src/v3/src/components/LoopbackProbe/LoopbackProbe.test.tsx
@@ -19,7 +19,7 @@ import { LoopbackProbeElement } from 'src/types';
 import LoopbackProbe from './LoopbackProbe';
 
 const proceedStub = jest.fn();
-jest.mock('../../../src/contexts', () => ({
+jest.mock('../../contexts', () => ({
   useWidgetContext: () => ({
     authClient: {
       idx: {
--- a/src/v3/src/components/LoopbackProbe/LoopbackProbe.test.tsx
+++ b/src/v3/src/components/LoopbackProbe/LoopbackProbe.test.tsx
@@ -18,9 +18,20 @@ import { LoopbackProbeElement } from 'src/types';
 
 import LoopbackProbe from './LoopbackProbe';
 
-const onSubmitHandler = jest.fn();
-jest.mock('../../hooks', () => ({
-  useOnSubmit: () => onSubmitHandler,
+const proceedStub = jest.fn();
+jest.mock('../../../src/contexts', () => ({
+  useWidgetContext: () => ({
+    authClient: {
+      idx: {
+        proceed: proceedStub,
+      },
+    },
+    idxTransaction: {
+      context: {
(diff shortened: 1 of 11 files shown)


src/v3/src/components/LoopbackProbe/LoopbackProbe.test.tsx
```diff
@@ -18,9 +18,21 @@ import { LoopbackProbeElement } from 'src/types';
 
 import LoopbackProbe from './LoopbackProbe';
 
-const onSubmitHandler = jest.fn();
-jest.mock('../../hooks', () => ({
-  useOnSubmit: () => onSubmitHandler,
+const proceedStub = jest.fn();
+jest.mock('../../../src/contexts', () => ({
```
glenfannin-okta, line 22: super nit 😅 (or maybe there was a reason for doing it this way?): 
```suggestion
jest.mock('../../contexts', () => ({
```

glenfannin-okta (approved):

src/v3/src/components/LoopbackProbe/LoopbackProbe.test.tsx
shawyu-okta: Oh oops I think just a mistake

src/v3/src/transformer/terminal/transformUnhandledErrors.ts
```diff
@@ -52,7 +52,11 @@ const getErrorMessage = (error?: AuthApiError, widgetProps?: WidgetProps) : stri
       message: () => loc('oie.feature.disabled', 'login'),
     },
     {
-      tester: (err?: AuthApiError) => err?.errorCode && !!err?.errorSummary,
+      tester: (err?: AuthApiError) => err?.errorCode && err?.errorSummary,
+      message: (err?: AuthApiError) => err?.errorSummary,
+    },
+    {
+      tester: (err?: AuthApiError) => err?.errorCode && !err?.errorSummary,
```
shawyu-okta, line 59: This diff is important @glenfannin-okta -- I removed a `!` here so instead of testing for the existence of `errorSummary`, we test for the non-existence of `errorSummary`. In the case where the _is_ an `errorSummary`, we use the tester above this one to directly render the `errorSummary` as the message in the infobox.

See the logic [here](https://github.com/okta/okta-signin-widget/blob/ff6816c5012387e4cfab4f25f16ba9d386b0570e/src/v2/controllers/FormController.ts#L343-L350) that I am copying. I saw we had unit tests asserting the existing behavior so I want to make sure we didn't intentionally change the behavior for gen3 before I merge this.

src/v3/src/types/schema.ts
```diff
@@ -327,6 +327,14 @@ export interface ReminderElement extends UISchemaElement {
...
   type: 'List';
   options: {
+    /**
+     * Items to render in the list.
+     *
+     * **NOTE**: Only string and UISchemaElement with type
+     * 'Button', 'Description', or 'TextWithHtml'
+     * are supported. Other UISchemaElement types will
+     * not render and print a warning to the console.
+     */
```
shawyu-okta, line 337: no idea where this came from 🤔

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
```diff
@@ -93,14 +95,29 @@ const LoopbackProbe: FunctionComponent<{ uischema: LoopbackProbeElement }> = ({
...
+        name: cancelStep,
         params,
-      });
+      }],
+    };
+    if (typeof idxTransaction?.context.stateHandle !== 'undefined') {
+      payload.stateHandle = idxTransaction.context.stateHandle;
     }
+    const newTransaction = await authClient?.idx.proceed(payload);
+    setIdxTransaction(newTransaction);
```
shawyu-okta, line 120: I considered refactoring these into one method but the payloads differ in a way that I think is a bit messy to try to reconcile, and the remaining duplication is minimal so I'm keeping them separate.

glenfannin-okta: Yea, I totally agree with this, the action payload vs the standard step + params payload is a mess to deal with, totally good with how you have it.

shawyu-okta:
@glenfannin-okta can you re-review some changes? I made some not-insignificant changes to the error message transformer

src/v3/src/transformer/terminal/transformUnhandledErrors.ts
```diff
@@ -44,26 +51,41 @@ const getErrorMessage = (error?: AuthApiError, widgetProps?: WidgetProps) : stri
...
+      break;
+    case 'OAuthError':
+      message = oauthErrorChecks
+        .find(({ tester }) => tester(error as OAuthError))?.message(error as OAuthError);
+      break;
+    default:
+      // intentionally fall through
+  }
+  // default fall back for unknown errors
+  return message || loc('oform.error.unexpected');
```
glenfannin-okta, line 88: nit: wondering if this is easier to follow?
```suggestion
    default:
      // default fall back for unknown errors
      message = loc('oform.error.unexpected');
  }
  return message;
```

glenfannin-okta (approved):
LGTM, I didn't know there were multiple types of errors auth-js can return. This is nice!

src/v3/src/transformer/terminal/transformUnhandledErrors.ts
shawyu-okta: I considered this but the issue with this is if we enter one of the switch cases but the search for a message fails, then this `default` case doesn't get hit and we'll return `undefined`.

glenfannin-okta: Ahh, that makes complete sense. Ty!

PR #3069, opened 2023-02-06
OKTA-541610 : Adding RTL support for BYOL

glenfannin-okta:
## Description:
The purpose of this PR is to ensure RTL direction is enabled when a language that requires RTL is used as a BYOL.

## PR Checklist

### Issue:

- [OKTA-541610](https://oktainc.atlassian.net/browse/OKTA-541610)

### Screenshot/Video:
image

https://user-images.githubusercontent.com/97472729/218834959-1e93b36d-2948-4698-b35f-ee9e50918b87.mov


src/v3/src/util/getLanguageDirection.ts
```diff
@@ -0,0 +1,35 @@
...
+  'ha', // Hausa
+  'he', // Hebrew
+  'khw', // Khowar
+  'ks', // Kashmiri
+  'ps', // Pashto
+  'sd', // Sindhi
+  'ur', // Urdu
+  'uz', // Uzbeki Afghanistan
+  'yi', // Yiddish
+]);
```
glenfannin-okta, line 29: I derived this list from the below sources: 
https://meta.wikimedia.org/wiki/Template:List_of_language_names_ordered_by_code
and 
https://www.loc.gov/standards/iso639-2/php/code_list.php

src/v3/package.json
```diff
@@ -39,7 +39,9 @@
     "codegen": "yarn workspace @okta/okta-signin-widget codegen",
     "dev": "yarn codegen && preact watch --no-prerender --no-sw --port 3000",
     "lint": "yarn codegen && eslint .",
+    "lint:css": "npx stylelint **/*.css",
     "lint:report": "scripts/buildtools/lint-report.sh",
+    "lint:tsx": "npx stylelint **/*.tsx",
```
glenfannin-okta, line 44: Should I remove this since it technically does not work?

src/v3/src/components/Form/Form.tsx
```diff
@@ -92,9 +95,8 @@ const Form: FunctionComponent<{
     <form
       noValidate
       onSubmit={handleSubmit}
-      className="o-form" // FIXME update page objects using .o-form selectors
+      className={classes} // FIXME update page objects using .o-form selectors
       data-se="o-form"
-      style={{ maxWidth: '100%', wordBreak: 'break-word' }}
```
glenfannin-okta, line 97: This should also resolve CSP violations.

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -175,10 +181,11 @@ const AuthenticatorButton: UISchemaElementComponent<{
...
           </Box>
-          <RightArrowIcon
-            name={`${iconName}-arrow-right`}
-            description={ctaLabel}
-          />
+          {
+            typeof languageDirection !== 'undefined'
+              ? <ArrowLeftIcon titleAccess={ctaLabel} />
+              : <ArrowRightIcon titleAccess={ctaLabel} />
+          }
```
glenfannin-okta, line 188: Must flip arrow for RTL languages

src/v3/src/components/Form/Form.tsx
shawyu-okta: Interestingly I didn't see any CSP violations for this when I was working on my ticket 🤔

glenfannin-okta: Oh weird 🤔 maybe its only an issue when you use javascript to manipulate the dom/style? Or maybe not lol I know in v2 this was an issue but maybe because we use Odyssey its a non issue?

src/v3/src/util/getLanguageDirection.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// ltr is default so mark as undefined when ltr
+type LanguageDirection = 'rtl' | undefined;
```
shawyu-okta, line 32: IMO we should either name this `LanguageIsRTL` or something similar or strictly make it a string union of `'rtl' | 'ltr'`

src/v3/src/util/getLanguageDirection.ts
shawyu-okta: Just for my understanding, is this an exhaustive list of RTL languages or is it just the RTL languages we want to support?

src/v3/src/components/Form/Form.tsx
```diff
@@ -92,9 +95,8 @@ const Form: FunctionComponent<{
     <form
       noValidate
       onSubmit={handleSubmit}
-      className="o-form" // FIXME update page objects using .o-form selectors
+      className={classes} // FIXME update page objects using .o-form selectors
```
shawyu-okta, line 98: I just made this JIRA for tech debt, can you attach the ticket number? https://oktainc.atlassian.net/browse/OKTA-578584

src/v3/src/util/getLanguageDirection.ts
glenfannin-okta: This is meant to be an exhaustive list of all RTL languages.

src/v3/src/util/getLanguageDirection.ts
glenfannin-okta: Yea I was trying to make it so that when the language is `ltr` it doesnt add the property to the dom, it doesnt hurt anything having it there I just wanted to avoid unnecessary noise in snapshots and in the html adding `ltr` when it is set that way by default. But I agree it is confusing only having it as `rtl`

src/v3/package.json
```diff
@@ -39,6 +39,7 @@
     "codegen": "yarn workspace @okta/okta-signin-widget codegen",
     "dev": "yarn codegen && preact watch --no-prerender --no-sw --port 3000",
     "lint": "yarn codegen && eslint .",
+    "lint:styles": "npx stylelint **/*.css",
     "lint:report": "scripts/buildtools/lint-report.sh",
```
shawyu-okta, line 43: We should update the script to also run the style lint

src/v3/src/components/AuthContainer/AuthContainer.tsx
```diff
@@ -15,14 +15,14 @@ import { Box } from '@okta/odyssey-react-mui';
...
 import { FunctionComponent, h } from 'preact';
 
-import { useWidgetContext } from '../../contexts';
-import { getLanguageCode } from '../../util';
 import style from './style.css';
 
-const AuthContainer: FunctionComponent = ({ children }) => {
+type AuthContainerProps = { languageCode: string; languageDirection?: 'rtl' };
+const AuthContainer: FunctionComponent<AuthContainerProps> = ({
+  languageDirection, languageCode, children,
```
shawyu-okta, line 22: do we want to consider allowing `...rest` props and pass them through directly to `Box`?

glenfannin-okta: There are two issues with this approach, if we extract children, `rest` will still contain props like `jsx` and `ref` which are incompatible with the `Box` component. Also, we disallow spreading props via lint so we would also need to disable it for this line if we were to do it. 

image

src/v3/src/components/PasswordRequirements/PasswordMatches.tsx
```diff
@@ -58,16 +57,16 @@ const PasswordMatches: UISchemaElementComponent<{
 
   return (
     <Box data-se="password-authenticator--matches">
-      <List
+      <Box
+        component="ul"
```
glenfannin-okta, line 61: Removing the usage of the `List` MUI element because in the latest version of Odyssey they are customizing the style that forces us to override them. This component only uses list for semantic reasoning not for the style so just using `Box`.

src/v3/src/components/PasswordRequirements/PasswordRequirementListItem.tsx
```diff
@@ -26,15 +23,20 @@ const PasswordRequirementListItem: FunctionComponent<PasswordRequirementProps> =
   const { status, label } = props;
 
   return (
-    <ListItem
-      disablePadding
-      disableGutters
+    <Box
+      component="li"
+      sx={{ marginBlockEnd: (theme) => theme.spacing(2) }}
```
glenfannin-okta, line 28: Removing the usage of the `List` MUI element because in the latest version of Odyssey they are customizing the style that forces us to override them. This component only uses list for semantic reasoning not for the style so just using `Box`.

src/v3/src/transformer/layout/email/emailChallengeConsentTransformer.ts
```diff
@@ -76,7 +76,7 @@ export const transformEmailChallengeConsent: IdxStepTransformer = ({ transaction
     label: loc('oie.consent.enduser.accept.label', 'login'),
     options: {
       type: ButtonType.BUTTON,
-      variant: 'secondary',
+      variant: 'primary',
```
glenfannin-okta, line 79: This should have been primary not secondary.

src/v3/.eslintrc.js
```diff
@@ -150,6 +150,48 @@ module.exports = {
...
+              { propName: 'borderBottomLeftRadius', message: 'Use "borderEndStartRadius" instead.' },
+              { propName: 'borderBottomRightRadius', message: 'Use "borderEndEndRadius" instead.' },
+
+              { propName: 'top', message: 'Use "insetBlockStart" instead.' },
+              { propName: 'bottom', message: 'Use "insetBlockEnd" instead.' },
+              { propName: 'left', message: 'Use "insetInlineStart" instead.' },
+              { propName: 'right', message: 'Use "insetInlineEnd" instead.' },
+            ],
+          },
+        ],
```
shawyu-okta, line 194: Nice! Fixit day project idea to extract this into a custom ESLint plugin or config that can be shared with rest of the UI projects

src/v3/src/components/AuthContainer/AuthContainer.tsx
```diff
@@ -33,7 +34,8 @@ const AuthContainer: FunctionComponent = ({ children }) => {
       className={classes}
       data-version={VERSION}
       data-commit={COMMITHASH}
-      lang={getLanguageCode(widgetProps)}
+      lang={lang}
+      dir={dir}
```
shawyu-okta, line 38: Actually, wondering if we should pull from context rather than passed as a prop. Thoughts?

shawyu-okta (approved):

PR #3087, opened 2023-02-15
Adds `cspNonce` to widget config

shawyu-okta:
## Description:

Allow a `cspNonce` to be passed to the widget config so it can be used by the MUI cache provider when CSP enforcement is desired.

## PR Checklist

### Issue:

- [OKTA-552075](https://oktainc.atlassian.net/browse/OKTA-552075)

### DIFF
--- a/src/v3/.eslintrc.js
+++ b/src/v3/.eslintrc.js
@@ -129,6 +129,20 @@ module.exports = {
           json: 'always',
         }],
 
+        // prevent inline styles which cause CSP violations
+        'react/forbid-component-props': ['error', {
+          forbid: [{
+            propName: 'style',
+            message: 'Inline styles cause CSP issues',
+          }],
+        }],
+        'react/forbid-dom-props': ['error', {
+          forbid: [{
+            propName: 'style',
+            message: 'Inline styles cause CSP issues',
+          }],
+        }],
+
         // use @typescript-eslint/no-unused-vars
         'no-unused-vars': 'off',
 
--- a/src/v3/preact.config.js
+++ b/src/v3/preact.config.js
@@ -100,6 +100,7 @@ export default {
           ...rule,
           use: rule.use.reduce((acc, loader) => {
             acc.push(
+              // force MiniCssExtractPlugin.loader for watch mode and static builds
               /style-loader/.test(loader) ? MiniCssExtractPlugin.loader : loader
             );
             return acc;
--- a/src/v3/preact.config.js
+++ b/src/v3/preact.config.js
@@ -190,7 +190,7 @@ export default {
       devServer: {
         headers: {
           'Content-Security-Policy': mergeContentSecurityPolicies(
-            "style-src 'self' 'nonce-playground'; object-src 'self'; script-src 'self' 'unsafe-eval'",
+            "style-src 'self'; object-src 'self'; script-src 'self' 'unsafe-eval' 'nonce-playground'",
             playgroundConfig.devServer.headers['Content-Security-Policy'] || '',
           ),
         },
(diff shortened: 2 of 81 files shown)


src/v3/src/components/Widget/index.tsx
```diff
@@ -304,9 +305,9 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     >
       {/* Note: we need to keep both themeproviders (MUI/ODS) until ODS exports all MUI components */}
       <MuiThemeProvider theme={brandedTheme}>
-        {/* the style is to allow the widget to inherit the parent's bg color */}
-        <ScopedCssBaseline sx={{ backgroundColor: 'inherit' }}>
-          <OdysseyCacheProvider>
+        <OdysseyCacheProvider nonce={cspNonce}>
+          {/* the style is to allow the widget to inherit the parent's bg color */}
+          <ScopedCssBaseline sx={{ backgroundColor: 'inherit' }}>
```
shawyu-okta, line 310: This is still being imported directly from MUI. Once we upgrade odyssey to > `0.18` then we can pull the component from there.

src/v3/src/components/Widget/index.tsx
```diff
@@ -304,9 +305,9 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     >
       {/* Note: we need to keep both themeproviders (MUI/ODS) until ODS exports all MUI components */}
       <MuiThemeProvider theme={brandedTheme}>
```
shawyu-okta, line 307: This is still causing some CSP issues. The theme provider pulls `cspNonce` only from `window.cspNonce` and does not directly accept it. This means in the playground we will still see the CSP failures but in the Okta-hosted environment it will pick up the nonce and work correctly.

Ultimately, we don't need to worry about this because we are planning to retire all usage of the older Odyssey theme provider so this will get removed soon.

src/v3/preact.config.js
```diff
@@ -105,7 +100,8 @@ export default {
           ...rule,
           use: rule.use.reduce((acc, loader) => {
             acc.push(
-              loader !== MiniCssExtractPlugin.loader ? loader : 'style-loader',
+              // force MiniCssExtractPlugin.loader for watch mode and static builds
+              /style-loader/.test(loader) ? MiniCssExtractPlugin.loader : loader,
```
lesterchoi-okta, line 104: doesn't look like this is matching anything right now:
```js
      if (rule.use) {
        return {
          ...rule,
          use: rule.use.reduce((acc, loader) => {
            console.log(/style-loader/.test(loader) && loader); // false for everything on `yarn build`
            acc.push(
              // force MiniCssExtractPlugin.loader for watch mode and static builds
              /style-loader/.test(loader) ? MiniCssExtractPlugin.loader : loader,
            );
            return acc;
          }, []),
        };
      }
```

src/v3/.eslintrc.js
```diff
@@ -129,6 +129,20 @@ module.exports = {
...
+            propName: 'style',
+            message: 'Inline styles cause CSP issues',
+          }],
+        }],
+        'react/forbid-dom-props': ['error', {
+          forbid: [{
+            propName: 'style',
+            message: 'Inline styles cause CSP issues',
+          }],
+        }],
```
lesterchoi-okta, line 144: 👍

lesterchoi-okta (approved):
see note in `preact.config.js`

src/v3/preact.config.js
shawyu-okta: This only matches when you run `dev` - the "watch mode" causes it to run `style-loader` rather than `MiniCssExtractPlugin.loader`

glenfannin-okta (approved):

PR #3088, opened 2023-02-15
OKTA-574017 : i18n key HTML content parser implementation

glenfannin-okta:
## Description:
The purpose of this PR is to add the `no-translate` class to the appropriate element when necessary to prevent selenium test suite from failing when executing psuedo-loc english leak tests. 

## PR Checklist

### Issue:

- [OKTA-574017](https://oktainc.atlassian.net/browse/OKTA-574017)

### Screenshot/Video:
image


src/v3/package.json
```diff
@@ -61,6 +61,7 @@
     "@okta/odyssey-react-mui": "^0.17.0",
     "@okta/odyssey-react-theme": "^0.15.3",
     "@okta/okta-auth-js": "^7.2.0",
+    "@types/react-html-parser": "^2.0.2",
```
glenfannin-okta, line 64: Is dependencies the correct place for this? Or should the types only be added to devDependencies?

shawyu-okta: should be devDeps since it's not used at runtime

src/v3/src/components/Form/style.module.css
```diff
@@ -0,0 +1,4 @@
+:global(span.strong) {
+  font-weight: bold;
+  word-break: break-all;
+}
```
shawyu-okta, line 4: Curious if this can be handled in the `ScopedCssBaseline` component: https://mui.com/material-ui/react-css-baseline/#typography

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -1277,6 +1277,8 @@ oie.user.profile.division = Division
 oie.user.profile.department = Department
 oie.user.profile.managerId = Manager ID
 oie.user.profile.manager = Manager
+oie.selfservice.unlock_user.landing.to.app.success.message = Account successfully unlocked! Verify your account with a security method to continue.
+oie.selfservice.unlock_user.landing.to.app.signing.in.message = Account successfully unlocked!
```
glenfannin-okta, line 1281: These already exist in master, we just have not backported yet, adding now to prevent content issue (server side sends `<br>` tag in text)

See: https://github.com/okta/okta-signin-widget/blob/b00690bacaef60dec1e1508c04993ad1dcea56ec/packages/@okta/i18n/src/properties/login.properties#L1304-L1305

packages/@okta/i18n/src/properties/login_ok_PL.properties
```diff
@@ -940,6 +940,8 @@ oie.user.profile.division = 》Ðîṽîšîöñ 䀕ヸ€홝한Ӝฐโ《
 oie.user.profile.department = 》Ðéþåŕţɱéñţ ⾼ئ䀕ヸ€홝한Ӝฐโ《
 oie.user.profile.managerId = 》Ṁåñåĝéŕ ÎÐ ⾼ئ䀕ヸ€홝한Ӝฐโ《
 oie.user.profile.manager = 》Ṁåñåĝéŕ ヸ€홝한Ӝฐโ《
+oie.selfservice.unlock_user.landing.to.app.success.message = 》Åççöûñţ šûççéššƒûļļý ûñļöçķéð¡ Ṽéŕîƒý ýöûŕ åççöûñţ ŵîţĥ å šéçûŕîţý ɱéţĥöð ţö çöñţîñûé· ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ《
+oie.selfservice.unlock_user.landing.to.app.signing.in.message = 》Åççöûñţ šûççéššƒûļļý ûñļöçķéð¡ 䀕ヸ€홝한Ӝฐโ⾼ئ䀕ヸ€홝한Ӝฐโ《
```
glenfannin-okta, line 944: Already exists in master see: https://github.com/okta/okta-signin-widget/blob/b00690bacaef60dec1e1508c04993ad1dcea56ec/packages/@okta/i18n/src/properties/login_ok_PL.properties#L962-L963

src/v3/src/components/Form/style.module.css
glenfannin-okta: Totally forgot about this 😅

src/v3/package.json
```diff
@@ -76,6 +77,7 @@
     "js-cookie": "^3.0.1",
     "lodash": "^4.17.21",
     "preact": "^10.7.3",
+    "react-html-parser": "^2.0.2",
```
shawyu-okta, line 80: Do we still need to do extra sanitization of the input strings before passing them to this library? https://github.com/peternewnham/react-html-parser#security

src/v3/src/types/schema.ts
```diff
@@ -134,6 +135,11 @@ export interface UISchemaElement {
    * view/step within the {@link StepperLayout} this element belongs to.
    */
   viewIndex?: number;
+  /**
+   * This function should be defined when your text contains <$1> tags
+   * which require custom interpolation
+   */
+  contentTransformer?: Options['transform'];
```
shawyu-okta, line 142: Is there any possibility this would be anything other than the function that `util/getHTMLTransformer` returns?

src/v3/package.json
glenfannin-okta: Good point, I questioned this as well, but initially I thought it wasn't necessary since the content is coming from our properties/translation file. But I can see if somehow that was compromised it could be an added layer of protection. What do you think, should I add the `dompurify` lib? It seems relatively small ~700KB

src/v3/src/types/schema.ts
shawyu-okta: Oh I didn't think about the fact that we need to pass parameters to this HOF to get the transformer we want. I think this makes more sense now with that in mind.

src/v3/src/types/schema.ts
```diff
@@ -387,6 +394,7 @@ export interface ImageWithTextElement extends UISchemaElement {
     SVGIcon: FunctionComponent;
     textContent?: string;
     alignment?: string;
+    textClassName?: string;
```
shawyu-okta, line 397: I'd like to avoid passing along className along if possible, since it will lead to us relying on className rather than other semantic attributes as a selector in the long term. Is it possible to use a more specific property to control the addition of `no-translate` as a class on each component? (e.g. `noTranslate?: boolean` with default `false`)

src/v3/src/types/schema.ts
glenfannin-okta: It shouldn't be, there are other options the ReactHTMLParser function accepts but I had no need to use those. If we want to expose all possible options to our transformers we could convert that to something like `parserOptions` and have our transformer fill in the object as needed. I went with just the transform function to limit the capabilities but I see no reason against exposing the entire `Options` object to our transformers. Curious what you think about that.

src/v3/src/types/schema.ts
glenfannin-okta: Great idea actually, I disliked the approach I added but was hoping to get some feedback :D ty! I will update 🙏

src/v3/src/components/TextWithHtml/TextWithHtml.tsx
```diff
@@ -66,8 +67,9 @@ const TextWithHtml: UISchemaElementComponent<{
           },
         })}
         onClick={handleClick}
-        dangerouslySetInnerHTML={{ __html: content }}
-      />
+      >
+        {ReactHtmlParser(content, { transform: uischema.contentTransformer })}
```
shawyu-okta, line 71: Could `ReactHtmlParser` also be encapsulated into the transformer? If so, are there major downsides?

glenfannin-okta: I considered this (I also considered adding it to our `loc` util function), the problem with that is it alters what is returned (so we will need to type it as `string | h.jsx.Element`. But additionally, it would require us to go through each flow/scenario and determine which contains spans, anchors, etc embedded in the translation keys we use, to determine if we need to use the parser (not just the ones that use the `$1` tags). So this approach is more of a catch all.

src/v3/package.json
glenfannin-okta: *Update* 

I actually gave this a try and when I use the `sanitize` function on the content from the properties file, it actually remove the `</$1>` token from the string and eventually breaks the content from displaying properly :( 

I don't think its a terrible thing to not sanitize since the content is sort of trusted (from .properties)

packages/@okta/i18n/src/properties/login.properties
nikhilvenkatraman-okta: @glenfannin-okta If the server side sends <br> tag should this also have the <br> tag then ?

nikhilvenkatraman-okta: As discussed please file a JIRA for the home team to remove the br tag from the enduser-backend bundle to keep this string in sync.

glenfannin-okta: 🙏 Created the [ticket](https://oktainc.atlassian.net/browse/OKTA-579103), I will follow up with that team as well to make sure it is handled.

src/v3/src/components/List/List.tsx
```diff
@@ -52,12 +50,6 @@ const renderElement = (item: UISchemaElement) => {
           <InformationalText uischema={item as DescriptionElement} />
         </Container>
       );
-    case 'TextWithHtml':
-      return (
-        <Container>
-          <TextWithHtml uischema={item as TextWithHtmlElement} />
-        </Container>
-      );
```
glenfannin-okta, line 60: For now this is no longer needed, the ODA screen that used this is now using the standard Description (informational text) component (which contains the parser logic) so the anchor link will get transformed correctly (updated screenshot attached to PR).

src/v3/src/components/TextWithActionLink/TextWithActionLink.tsx
```diff
@@ -31,45 +31,35 @@ const TextWithHtml: UISchemaElementComponent<{
...
   };
 
   return (
-    <Box>
-      <Box
-        sx={(theme) => ({
-          [`& .${contentClassname}`]: {
-            color: theme.palette.primary.main,
-          },
-        })}
```
glenfannin-okta, line 67: the parser automatically converts anchor tags to the ODS MUI link component, so no need to override the color.

src/v3/src/hooks/useHtmlContentParser.tsx
```diff
@@ -0,0 +1,46 @@
...
+    if (node instanceof Element && node.type === 'tag' && node.name === 'a') {
+      const props = attributesToProps(node.attribs);
+      // eslint-disable-next-line react/jsx-props-no-spreading
+      return <Link {...props}>{domToReact(node.children, parserOptions)}</Link>;
+    }
+    return undefined;
+  };
+  if (typeof parserOptions.replace === 'undefined') {
+    parserOptions.replace = replace;
+  }
```
glenfannin-okta, line 43: Setting this as a default to avoid having to redefine this at the transformer layer each time we need it. Also, it would require us to inspect keys to see which contains an anchor tag to determine if we need to add this function or not.

src/v3/src/types/schema.ts
```diff
@@ -300,12 +301,11 @@ export interface DescriptionElement extends UISchemaElement {
...
 }
 
-export interface TextWithHtmlElement extends UISchemaElement {
-  type: 'TextWithHtml';
+export interface TextWithActionLinkElement extends UISchemaElement {
+  type: 'TextWithActionLink';
   options: ActionOptions & {
     content: string;
     contentClassname: string;
-    submitOnClick: boolean;
```
glenfannin-okta, line 308: Removed this because now the TextWithHtml component is renamed to TextWithActionLink. This was done to limit the usage of this component to only be used for translation keys that contain a link which does not contain an `href` value (i.e. it operates as a button instead of a link).

src/v3/src/util/locUtil.ts
```diff
@@ -17,12 +17,13 @@ export const loc = (
   key: string,
   bundleName?: string,
   params?: Array<string | number | boolean | unknown>,
+  replacerFn?: (content: string) => string,
```
glenfannin-okta, line 20: This function is used to replace tokens like `<$1></$1>` with an actual HTML element. The reason for this is because the parser removes these unknown tags.

src/v3/package.json
glenfannin-okta: _Additional update_ 

Now that I switched libraries (react-html-parser to html-react-parser). It seems to be doing some sort of sanitization on its own as it also removes the `<$1>` tags. However, I have a workaround for that, but now that this is in place, I think I can sanitize the string before parsing it, and should not have an issue like before, should I go ahead and do that?

glenfannin-okta: _Final update_

I went ahead and added the dompurify package and utilized it in the parser hook :D

src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
```diff
@@ -50,10 +51,8 @@ const ReminderPrompt: UISchemaElementComponent<{
     sessionStorage.setItem(RESEND_TIMESTAMP_SESSION_STORAGE_KEY, token);
   };
 
-  // eslint-disable-next-line arrow-body-style
-  const getResendTimestamp = (): string | null => {
-    return sessionStorage.getItem(RESEND_TIMESTAMP_SESSION_STORAGE_KEY);
-  };
+  const getResendTimestamp = (): string => sessionStorage
+    .getItem(RESEND_TIMESTAMP_SESSION_STORAGE_KEY) as string;
```
shawyu-okta, line 55: This isn't specifically related to this change at all but I missed on the review for this PR that we cannot rely on `string` type after `getItem`. Later if we `parseInt` on the value here and it's `null` it will return `NaN`, which is probably not what we want to do math with. Let's set a sensible default in case it's not set and also put this return type back to `string | null`. @leemchale-okta

src/v3/src/util/locUtil.ts
```diff
@@ -17,12 +17,13 @@ export const loc = (
...
-  const localizedText = localize(key, bundleName, params);
+  const localizedText: string = localize(key, bundleName, params);
 
-  return localizedText
-    .replace('<$1>', '')
-    .replace('</$1>', '')
-    .replace('<span class="strong">', '')
-    .replace('</span>', '');
+  if (typeof replacerFn === 'function') {
+    return replacerFn(localizedText);
```
shawyu-okta, line 25: Rather than allow passing an unrestricted function, what do you think of passing a map of key/value replacements? Right now the function can modify the string in any way it wishes.

src/v3/src/hooks/useHtmlContentParser.tsx
```diff
@@ -0,0 +1,52 @@
...
+  options?: HTMLReactParserOptions,
+): string | h.JSX.Element | h.JSX.Element[] | undefined => {
+  if (typeof content === 'undefined') {
+    return undefined;
+  }
+
+  const parserOptions: HTMLReactParserOptions = {
+    ...options,
+    // eslint-disable-next-line global-require
+    library: require('preact'),
```
shawyu-okta, line 35: Does this still work if you `import` this at the top and then pass it as a variable?

shawyu-okta (approved):

src/v3/src/components/ImageWithText/ImageWithText.tsx
```diff
@@ -47,6 +48,7 @@ const ImageWithText: UISchemaElementComponent<{
         <Box>
           <Typography
             component="span"
+            className={noTranslate ? 'no-translate' : undefined}
```
queeniechen-okta, line 51: Why do we need a no-translate here? Is the content of the text supposed to represent an image so we don't want to count it as an "English Leak"? 

Although if that is the case, would we ever want "localized" images?

src/v3/src/transformer/i18n/transformSecondEmailInputExplain.ts
```diff
@@ -37,6 +37,7 @@ export const transformSecondEmailInputExplain: TransformStepFnWithOptions = ({
         element,
         name: 'bottomExplain',
         i18nKey: 'oie.profile.additional.secondemail.subtitle',
+        replacementTokens: { '<$1>': '<span class="strong">', '</$1>': '</span>' },
```
queeniechen-okta, line 40: So I started writing this when you had the replacementFn. Now that you updated this, I think most of my concerns are gone but I'm wondering if we can make it even more concise?

I.e.:
```
add Translation({
....,
arguments: {
  $1:  { 
element: span,
class: "no-translate"
}
  }
}
})

```

glenfannin-okta: I like this, we can definitely do this.

glenfannin-okta: I am afraid that using `.replace` on the `$1` alone could be dangerous. I was considering using regex to only find `$1` that is wrapped with brackets but then it would also require me to handle the open tag differently than the closing tag. 

Also, I think this logic will be short lived, we were discussing today how we can get away from using these tags/html embedded in translation strings because it is a sort of anti-pattern.

queeniechen-okta: I think one potential solution might be to match /(<`argument`>)(.*)(</`argument`>)/, and then do some magic with matching groups. 

Otherwise perhaps the logic here might help? : https://github.com/okta/okta-ui/blob/master/packages/courage/src/util/handlebars/helper-i18n.ts#L14

However, that might be a bit complex for now and I agree if this is not going to be the long term solution, I'm fine with revisiting it another day :)

glenfannin-okta: All set! Thank you so much for the recommendation!

src/v3/src/components/ImageWithText/ImageWithText.tsx
glenfannin-okta: This text is a number, it is the code that is pushed to their OV app. SIW v2 expects it to have the no-translate class on it so their english leaks test can ignore it. 

FYI, I did create the spike for our team dependent on the ticket you shared yesterday 🙏

glenfannin-okta:
cc: @zihanwang-okta for translation review.

mauriciocastillosilva-okta (approved):
Changes to properties file look good. I looked quickly at no-translate too, and looks good as well

src/v3/src/util/locUtil.ts
```diff
@@ -13,16 +13,41 @@
...
   key: string,
   bundleName?: string,
   params?: Array<string | number | boolean | unknown>,
+  tokenReplacement?: Record<TokenSearchValue, TokenReplacementValue>,
 ): string => {
-  const localizedText = localize(key, bundleName, params);
+  const localizedText: string = localize(key, bundleName, params);
+
+  if (typeof tokenReplacement !== 'undefined') {
+    let updatedText = localizedText;
```
queeniechen-okta, line 36: curious, does this work with nested tags? so like `text text <$1> text text <$1> text </$1> </$1>`

queeniechen-okta (approved):
Thanks for addressing my comments! the token replacement thing looks very epic, just have one more minor question and if that is fine, LGTM 💖 💖

src/v3/src/util/locUtil.ts
glenfannin-okta: It would not work with that :( , but I don't believe that is expected to work. With these tokens, if there are multiple open/closing tags in one string, the token value should change. 

i.e. 
```
hcaptcha.footer.label = This site is protected by hCaptcha and its <$1>Privacy Policy</$1> and <$2>Terms of Service</$2> apply.
```

So in this case, if we wanted to, we would just need to add `$2` to the TokenSearchValue type and it will work as expected.

queeniechen-okta: What if I wanted to format lets say:
`<a href="#" > <strong> here is a test string </strong> </a>`

do we support that?

glenfannin-okta: We _shouldn't_ support that 😅 adding HTML elements inside the translation string is sort of considered an anti-pattern and we actually are discussing how we can clean up these translations and limit how things can be styled

queeniechen-okta: ah I see, so is signin widget going for no HTML elements at all?

Because from what I'm aware of, we don't want HTML in the `properties` explicitly but it's ok to be using these `<$1>` notations. 

What I was going for is, if that was the theoretical string that we wanted to represent, would `<$1> <$2> here is a test string </$2> </$1>` work with the function you have built out if I fill in the proper arguments for `$1` and `$2`?

glenfannin-okta: Ah, then yes that is possible, I had to modify my type a bit so TS would not complain but here is the commit. Also added a unit test to show it working. 

https://github.com/okta/okta-signin-widget/pull/3088/commits/9e461921955d2bac6294891807f5cb78e79f30fa

queeniechen-okta: Epic, thank you 👍

PR #3110, opened 2023-03-02
OKTA-576951 : Registration callbacks implementation for g3

glenfannin-okta:
## Description:

The purpose of this PR is to implement the Registration callbacks for the G3 SIW. 

** Note **
The assumption made here is that even though the API does not accept new (unregistered) fields to be passed successfully (without triggering an error), to still allow this in G3 because it currently works this way in G2. Customers _could_ be adding fields onto the page with the parseSchema callback, then, before submission to the API, using the preSubmit callback to collect the data and remove it from the payload before it is sent on to the API (avoid the IDX error for trying to utilize unregistered fields).

## PR Checklist

### Issue:

- [OKTA-576951](https://oktainc.atlassian.net/browse/OKTA-576951)

### Screenshot/Video:
Registration callback demo:
https://drive.google.com/file/d/1bVrKfmrk4XheiqBpaewG-si-5fIFeqXK/view?usp=sharing

Added update to display monochrome style link within the ReminderPrompt component: 
image

image

https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&page=1&pageSize=6&sha=d083d152f54199df30c59872f52f7b1c32e80c74&tab=main

### DIFF
--- a/src/types/options.ts
+++ b/src/types/options.ts
@@ -148,7 +148,7 @@ interface Piv {
 }
 
 // Types for config
-type UserOperation =
+export type UserOperation =
   'PRIMARY_AUTH' |
   'FORGOT_PASSWORD' |
   'UNLOCK_ACCOUNT';
--- a/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
+++ b/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
@@ -106,6 +106,7 @@ const ReminderPrompt: UISchemaElementComponent<{
       <Link
         // eslint-disable-next-line no-script-url
         href="javascript:void(0);"
+        variant="monochrome"
         onClick={() => resendHandler()}
       >
         {buttonText}
--- a/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
+++ b/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
@@ -126,6 +126,7 @@ const ReminderPrompt: UISchemaElementComponent<{
               step,
               isActionStep,
               actionParams,
+              variant: 'monochrome',
             },
           }}
         />
--- a/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
+++ b/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
@@ -11,12 +11,14 @@
  */
 
 import { Alert, Box, Link } from '@okta/odyssey-react-mui';
+import { HTMLReactParserOptions } from 'html-react-parser';
 import { h } from 'preact';
 import { useEffect, useRef, useState } from 'preact/hooks';
 
 import { useWidgetContext } from '../../contexts';
 import { useHtmlContentParser, useOnSubmit } from '../../hooks';
 import { ReminderElement, UISchemaElementComponent } from '../../types';
+import { getLinkReplacerFn } from '../../util';
 import TextWithActionLink from '../TextWithActionLink';
 
 export const DEFAULT_TIMEOUT_MS = 30_000;
@@ -116,11 +118,13 @@ const ReminderPrompt: UISchemaElementComponent<{
 
   const renderAlertContent = () => {
     if (contentHasHtml && contentClassname) {
+      const parserOptions: HTMLReactParserOptions = {};
+      parserOptions.replace = getLinkReplacerFn(parserOptions, 'monochrome');
       return (
         <TextWithActionLink
           uischema={{
             type: 'TextWithActionLink',
-            parserOptions: uischema.parserOptions,
+            parserOptions,
             options: {
               contentClassname,
               content,
--- a/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
+++ b/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
@@ -38,7 +38,7 @@ const ReminderPrompt: UISchemaElementComponent<{
     contentHasHtml,
   } = uischema.options;
   const onSubmitHandler = useOnSubmit();
-  const parsedContent = useHtmlContentParser(content);
+  const parsedContent = useHtmlContentParser(content, uischema.parserOptions);
 
   const [show, setShow] = useState<boolean>(false);
   const timerRef = useRef<number | undefined>();
@@ -120,13 +120,13 @@ const ReminderPrompt: UISchemaElementComponent<{
         <TextWithActionLink
           uischema={{
             type: 'TextWithActionLink',
+            parserOptions: uischema.parserOptions,
             options: {
               contentClassname,
               content,
               step,
               isActionStep,
               actionParams,
-              variant: 'monochrome',
             },
           }}
         />
--- a/src/v3/src/hooks/useHtmlContentParser.tsx
+++ b/src/v3/src/hooks/useHtmlContentParser.tsx
@@ -10,20 +10,15 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { Link } from '@okta/odyssey-react-mui';
 import dompurify from 'dompurify';
-import HtmlReactParser, {
-  attributesToProps,
-  DOMNode,
-  domToReact,
-  Element,
-  HTMLReactParserOptions,
-} from 'html-react-parser';
+import HtmlReactParser, { HTMLReactParserOptions } from 'html-react-parser';
 import preact, { h } from 'preact';
+
+import { getLinkReplacerFn } from '../util';
 
 export const useHtmlContentParser = (
   content: string | undefined,
-  options?: HTMLReactParserOptions & { variant?: 'monochrome'; },
+  options?: HTMLReactParserOptions,
 ): string | h.JSX.Element | h.JSX.Element[] | undefined => {
   if (typeof content === 'undefined') {
     return undefined;
@@ -35,22 +30,8 @@ export const useHtmlContentParser = (
   };
 
   if (typeof parserOptions.replace === 'undefined') {
-    // default replacer function
-    parserOptions.replace = (node: DOMNode) => {
-      if (node instanceof Element && node.type === 'tag' && node.name === 'a') {
-        const props = attributesToProps(node.attribs);
-        return (
-          <Link
-            // eslint-disable-next-line react/jsx-props-no-spreading
-            {...props}
-            variant={options?.variant}
-          >
-            {domToReact(node.children, parserOptions)}
-          </Link>
-        );
-      }
-      return undefined;
-    };
+    // default link replacer function
+    parserOptions.replace = getLinkReplacerFn(parserOptions);
   }
 
   const sanitizedContent = dompurify.sanitize(content);
--- a/src/v3/src/hooks/useHtmlContentParser.tsx
+++ b/src/v3/src/hooks/useHtmlContentParser.tsx
@@ -23,7 +23,7 @@ import preact, { h } from 'preact';
 
 export const useHtmlContentParser = (
   content: string | undefined,
-  options?: HTMLReactParserOptions,
+  options?: HTMLReactParserOptions & { variant?: 'monochrome'; },
 ): string | h.JSX.Element | h.JSX.Element[] | undefined => {
   if (typeof content === 'undefined') {
     return undefined;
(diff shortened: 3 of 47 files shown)


src/types/options.ts
```diff
@@ -137,7 +137,7 @@ interface Piv {
 }
 
 // Types for config
-type UserOperation =
+export type UserOperation =
```
glenfannin-okta, line 140: Wanted to re-use this type without re-defining it in v3 folder.

src/v3/src/components/Form/renderers.tsx
```diff
@@ -201,7 +202,7 @@ export default [
     renderer: InputPassword,
   },
   {
-    tester: ({ options: { inputMeta: { type } = {} as Input } }: FieldElement) => type === 'boolean',
+    tester: ({ options: { inputMeta: { type } = {} as Input } }: FieldElement) => type === 'boolean' || type === 'checkbox',
```
glenfannin-okta, line 205: Customers can send in `checkbox` as the `Input` type and it is expected to be respected and render as such (tested against v2).

src/v3/src/components/Form/renderers.tsx
```diff
@@ -167,21 +167,22 @@ export default [
...
     tester: ({
       options: {
-        inputMeta: { options } = {} as Input,
+        inputMeta: { options, type } = {} as Input,
         format,
         customOptions,
       },
-    }: FieldElement) => (Array.isArray(customOptions) || Array.isArray(options)) && format === 'select',
+    }: FieldElement) => (Array.isArray(customOptions) || Array.isArray(options) || typeof options === 'object')
+      && ([format, type].includes('select')),
```
glenfannin-okta, line 185: Customers can send in `select` as the `Input` type and it is expected to be respected and render as such (tested against v2).

src/v3/src/components/Form/renderers.tsx
```diff
@@ -167,21 +167,22 @@ export default [
   {
     tester: ({
       options: {
-        inputMeta: { options } = {} as Input,
+        inputMeta: { options, type } = {} as Input,
         format,
         customOptions,
       },
-    }: FieldElement) => (Array.isArray(customOptions) || Array.isArray(options)) && format === 'radio',
+    }: FieldElement) => (Array.isArray(customOptions) || Array.isArray(options)) && ([format, type].includes('radio')),
```
glenfannin-okta, line 174: Customers can send in `radio` as the `Input` type and it is expected to be respected and render as such (tested against v2).

src/v3/src/components/Select/Select.tsx
```diff
@@ -47,6 +47,20 @@ const Select: UISchemaElementComponent<UISchemaElementComponentWithValidationPro
...
+  const getOptions = (): IdxOption[] | undefined => {
+    if (Array.isArray(customOptions) || Array.isArray(options)) {
+      return customOptions ?? options;
+    }
+
+    if (typeof options === 'object') {
+      return Object.entries(options)
+        .filter(([key, val]) => key !== '' && val !== '')
+        .map(([key, val]) => ({ label: val, value: key } as IdxOption));
+    }
```
glenfannin-okta, line 59: Supports registration callback parseSchema format for select fields.

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -140,36 +174,73 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
...
+          setMessage({
+            message: loc('oform.errorbanner.title', 'login'),
+            class: 'ERROR',
+            i18n: { key: 'oform.errorbanner.title' },
+          } as IdxMessage);
+        }
+        setStepToRender(stepToRender);
+        setLoading(false);
+        resolve?.(true);
+      };
```
glenfannin-okta, line 202: Needed a re-usable way to execute this group of functions (between the reg callback and in non registration flows/scenarios)

src/v3/src/transformer/field/type.ts
```diff
@@ -39,6 +39,7 @@ const map = new Map<string, JsonSchemaType>([
   ['pdatetime', 'string'],
   ['ptime', 'string'],
   ['string', 'string'],
+  ['text', 'string'],
```
glenfannin-okta, line 42: Registration parseSchema can send `text` as the `Input` type, so we need to make sure this is rendered properly.

src/v3/src/transformer/profile/transformEnrollProfile.ts
```diff
@@ -97,6 +97,13 @@ export const transformEnrollProfile: IdxStepTransformer = ({ transaction, formBa
       validate: (data: FormBag['data']) => {
         const newPw = data['credentials.passcode'] as string;
         const errorMessages: WidgetMessage[] = [];
+        if (!newPw && passwordElement.options.inputMeta.required) {
+          errorMessages.push({
+            class: 'ERROR',
+            message: loc('model.validation.field.blank', 'login'),
+            i18n: { key: 'model.validation.field.blank' },
+          });
+        }
```
glenfannin-okta, line 106: Since we are overriding the default validation function, we need to check if the field is required and trigger the empty field error message.

src/v3/src/transformer/transaction/transform.ts
```diff
@@ -44,6 +50,41 @@ const transformRemediationNameForIdp: TransformStepFnWithOptions = (options) =>
...
+    (error: APIError) => {
+      // Prioritize server side errors over custom error message
+      if (isClientTransaction) {
+        return;
+      }
+      triggerRegistrationErrorMessages(error, inputs, setMessage);
+    },
+  );
+  return formbag;
+};
```
glenfannin-okta, line 85: This is the meat of the parseSchema functionality.

TLDR;

I am taking the inputs sent to us from the API response and converting them into the Registration schema format (which is what v2 is currently expecting). Then, I am passing it to the callback (to allow the customer to add/update/etc) the schema, and in the body of the callback, I execute a function that converts the modified schema back into the IDX `Input` object array to allow it to pass through the SIW standard transformation functions (where we will get things like validation and element creation for free).

src/v3/src/types/schema.ts
```diff
@@ -146,6 +146,15 @@ export type TokenReplacementValue = {
...
 export type TokenReplacement = Partial<Record<TokenSearchValue, TokenReplacementValue>>;
 
+export type RegistrationElementSchema = Modify<Input, {
+  'label-top'?: boolean;
+  placeholder?: string;
+  'data-se'?: string;
+  options?: IdxOption[] | Record<string, string>[];
+  sublabel?: string;
+  wide?: boolean;
+}>;
```
glenfannin-okta, line 156: The v2 Registration schema is just a derivative of the `Input` object with some additional fields. 

I am going to file a new ticket to respect some of these new fields, i.e. `sublabel` should display a label under the main label, but it requires changes to many components and I wanted this PR to be as small as possible.

src/v3/src/types/widget.ts
```diff
@@ -67,6 +75,21 @@ export type WidgetResetArgs = {
...
+    schema: RegistrationElementSchema[],
+    onSuccess: RegistrationSchemaCallbackV3,
+    onFailure: RegistrationErrorCallback
+  ) => void;
+  preSubmit?: (
+    postData: IdxActionParams,
+    onSuccess: RegistrationDataCallbackV3,
+    onFailure: RegistrationErrorCallback
+  ) => void;
+}>;
```
glenfannin-okta, line 91: I had to modify this type because it looked like this was originally made for v1 (the current structure did not match v2 at all).

src/v3/src/util/idxUtils.ts
```diff
@@ -229,3 +238,90 @@ export const convertIdxMessageToWidgetMessage = (
...
+  inputs.reduce((acc: Input[], input: Input) => {
+    const flattenedInputs = flattenInputs(input);
+    return [...acc, ...flattenedInputs];
+  }, []).map((flattendInput: Input) => ({
+    ...flattendInput,
+    'label-top': true,
+    'data-se': flattendInput.name,
+    wide: true,
+  } as RegistrationElementSchema))
+);
```
glenfannin-okta, line 254: The schema array that is sent to the customer in the callback is a flattened version of the `Input` array provided to use from IDX. So this function just flattens the objects and adds default values for those extra properties (which is what v2 sets them as too).

src/v3/src/util/idxUtils.ts
```diff
@@ -229,3 +238,90 @@ export const convertIdxMessageToWidgetMessage = (
...
+        // Existing Field (modify)
+        Object.assign(idxInputs[existFieldIdx], schemaEle);
+      } else {
+        // New Field (create)
+        // @ts-expect-error Registration Schema Element has extra fields that are not in Input
+        idxInputs.push(schemaEle);
+      }
+    }
+  });
+};
```
glenfannin-okta, line 302: This function converts the flattened Registration schema back into an `Input` object array. There is an assumption made here that we wont nest objects more than 1 level (because testing in v2, those properties are ignore when I tried them).

src/v3/src/hooks/useHtmlContentParser.tsx
```diff
@@ -23,7 +23,7 @@ import preact, { h } from 'preact';
 
 export const useHtmlContentParser = (
   content: string | undefined,
-  options?: HTMLReactParserOptions,
+  options?: HTMLReactParserOptions & { variant?: 'monochrome'; },
```
shawyu-okta, line 26: This is a little bit smelly, we might want to think a little more (but not necessarily block this work) about how to do this without drilling the prop that doesn't really have anything superficially to do with this hook

glenfannin-okta: I agree, initially what I thought I could do was just set the attribute on the link before we pass it to `loc`. What happens though is that the Parser doesn't recognize the `variant` attribute and strips it out :(

glenfannin-okta: I found another way around this issue. I think this approach is a little cleaner and re-usable for other scenarios.
https://github.com/okta/okta-signin-widget/pull/3110/commits/d2c2ad7ba73fef0ee61392a819d90a54b614e4a2

src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
```diff
@@ -119,6 +120,7 @@ const ReminderPrompt: UISchemaElementComponent<{
         <TextWithActionLink
           uischema={{
             type: 'TextWithActionLink',
+            parserOptions: uischema.parserOptions,
```
shawyu-okta, line 123: Do we always want a monochrome link in `ReminderPrompt`? If so, we could hard-code that here instead of pulling from `uischema`

src/v3/src/transformer/transaction/transform.ts
```diff
@@ -10,11 +10,17 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { IdxRemediation } from '@okta/okta-auth-js';
-import { flow } from 'lodash';
```
shawyu-okta, line 14: I _think_ this actually is the preferred way to import from `lodash` now: https://lodash.com/per-method-packages

shawyu-okta (approved):
strong work!

src/v3/src/util/settingsUtils.ts
```diff
@@ -107,3 +120,88 @@ export const getDefaultCountryCode = (widgetProps: WidgetProps): string => {
...
+    (error: APIError) => {
+      const errorObj = error || {
+        errorSummary: loc('registration.default.callbackhook.error', 'login'),
+      };
+      onFailure(errorObj);
+    },
+  );
+};
+
+export const transformIdentifier = (
```
shawyu-okta, line 196: Clarifying from yesterday's meeting, does this satisfy the `transformUsername` hook requirement?

glenfannin-okta: Yep! It does! 🙏 I updated the doc to reference this.

src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
glenfannin-okta: Ah, so you mean call the function here to build the parser options (for the link replacer) instead of doing it at the transformer layer. I like it, it removes duplicated effort! 🙏

PR #3111, opened 2023-03-03
OKTA-549614 : Hide redundant Authenticator button icon text

leemchale-okta:
## Description:

This PR adds `aria-hidden` to the Authenticator button icon titles as they are redundant to the button label content.

## PR Checklist

### Issue:

- [OKTA-549614](https://oktainc.atlassian.net/browse/OKTA-549614)

### DIFF
--- a/src/v3/src/components/AuthCoin/AuthCoin.test.tsx
+++ b/src/v3/src/components/AuthCoin/AuthCoin.test.tsx
@@ -36,21 +36,21 @@ describe('AuthCoin tests', () => {
 
     const { container, findByRole } = render(<AuthCoin {...props} />);
 
-    expect(await findByRole('img')).toBeDefined();
+    expect(await findByRole('img', { hidden: true })).toBeDefined();
     expect(container).toMatchSnapshot();
   });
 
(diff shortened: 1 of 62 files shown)


glenfannin-okta (approved):
LGTM, wondering, for the AuthCoin in the Header, should we also use the aria-hidden attr? cc: @alexdahl-okta 

If so this implementation can change a bit by adding it directly to the AuthCoin SVGs

alexdahl-okta:
Ah, good point! I think that's true. https://oktainc.atlassian.net/browse/OKTA-555963

As long we're vigilant about always pairing auth icons with descriptive text nearby, the icons themselves probably don't need labels.

leemchale-okta:
> LGTM, wondering, for the AuthCoin in the Header, should we also use the aria-hidden attr? cc: @alexdahl-okta
> 
> If so this implementation can change a bit by adding it directly to the AuthCoin SVGs

@glenfannin-okta what do you think of adding the `aria-hidden` to the AuthCoin component?  Since the Auth button uses the AuthCoin aswell

PR #3112, opened 2023-03-03
updates loc string to match backend

jaredperreault-okta:
## PR Checklist

### Issue:
- [OKTA-551586](https://oktainc.atlassian.net/browse/OKTA-551586)


mauriciocastillosilva-okta (commented):
Changes look good. One thing, can you test with any other language? Reason I am asking is that now English string has an argument, but other languages do not. I believe this will work fine, as in, the argument pass to non-English will be ignored, but want to be sure.

glenfannin-okta (approved):

glenfannin-okta:
One question I had was how will this work with customers that may have overridden this key? I assume it should continue to work? 

Also, will this only be for english? What happens with the other languages? Or was this only affecting English strings originally?

jaredperreault-okta:
The L10N error will occur if the backend sends a param and the message (on the SIW side) does not have a placeholder character. This is true across languages and customized strings

Targeting master so all languages will get the same update before this fix is released

zihanwang-okta (approved):

mauriciocastillosilva-okta (approved):

PR #3113, opened 2023-03-03
OKTA-577972 : g3 : Applying fix for OV Biometrics error message format

glenfannin-okta:
## Description:

The purpose of this PR is to format the Biometrics error message as a list to meet parity with g2 SIW as well as enable the parity spec test.

## PR Checklist

### Issue:

- [OKTA-577972](https://oktainc.atlassian.net/browse/OKTA-577972)

### Screenshot/Video:
image

### DIFF
--- a/src/v3/src/components/InfoBox/InfoBox.tsx
+++ b/src/v3/src/components/InfoBox/InfoBox.tsx
@@ -56,7 +56,7 @@ const InfoBox: UISchemaElementComponent<{
       >
         { message.title && (
         <Typography
-          component="h2"
+          component="h3"
           variant="h6"
         >
           {message.title}
--- a/src/v3/src/components/InfoBox/InfoBox.tsx
+++ b/src/v3/src/components/InfoBox/InfoBox.tsx
@@ -56,7 +56,7 @@ const InfoBox: UISchemaElementComponent<{
       >
         { message.title && (
         <Typography
-          component="h3"
+          component="h2"
           variant="h6"
         >
           {message.title}
--- a/src/v3/src/transformer/messages/transform.ts
+++ b/src/v3/src/transformer/messages/transform.ts
@@ -10,9 +10,10 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+import flow from 'lodash/flow';
+
 import { IDX_STEP, OV_UV_ENABLE_BIOMETRIC_SERVER_KEY } from '../../constants';
 import {
-  FormBag,
   InfoboxElement,
   TransformStepFnWithOptions,
   UISchemaElement,
@@ -33,8 +34,7 @@ const fipsComplianceKeys = [
 ];
 
 export const CUSTOM_MESSAGE_KEYS = [
-  OV_OVERRIDE_MESSAGE_KEY.OV_FORCE_FIPS_COMPLIANCE_UPGRAGE_KEY_IOS,
-  OV_OVERRIDE_MESSAGE_KEY.OV_FORCE_FIPS_COMPLIANCE_UPGRAGE_KEY_NON_IOS,
+  ...fipsComplianceKeys,
   OV_OVERRIDE_MESSAGE_KEY.OV_QR_ENROLL_ENABLE_BIOMETRICS_KEY,
   OV_UV_ENABLE_BIOMETRIC_SERVER_KEY,
 ];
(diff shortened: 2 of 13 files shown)


src/v3/src/components/InfoBox/InfoBox.tsx
```diff
@@ -56,7 +56,7 @@ const InfoBox: UISchemaElementComponent<{
       >
         { message.title && (
         <Typography
-          component="h2"
+          component="h3"
```
glenfannin-okta, line 59: The title in InfoBox should at be an H3 at minimum because we anticipate each page to have an H2 title. Additionally this matches parity with v2.

src/v3/src/transformer/terminal/transformTerminalMessages.ts
```diff
@@ -75,10 +75,6 @@ const appendBiometricsErrorBox = (
     options: {
       class: 'ERROR',
       dataSe: 'callout',
-      listOptions: {
-        type: 'ul',
-        items: listMessages,
-      },
```
glenfannin-okta, line 81: Remnant from the WidgetMessage refactor. The infoBoxElement no longer respects the listOptions prop since it is now apart of the WidgetMessage.

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -45,6 +45,12 @@ export default class BaseFormObject {
     }).innerText;
   }
 
+  getNthTitle(index) {
+    return screen.findAllByRole('heading', {
+      level: 2,
```
shawyu-okta, line 50: Is it an issue that we only target h2 as a "title"?

src/v3/src/transformer/messages/transform.ts
```diff
@@ -55,9 +55,32 @@ const overrideMessagesWithTitle = (msgs: WidgetMessage[]): WidgetMessage[] => {
...
+    description: loc('oie.authenticator.app.method.push.verify.enable.biometrics.description', 'login'),
+    message: messageBullets,
+  };
+  msgs.splice(msgIndex, 1, message);
+};
+
 const transformCustomMessages = (formBag: FormBag, messages: WidgetMessage[]): FormBag => {
   const { uischema } = formBag;
   const formattedMessages = overrideMessagesWithTitle(messages);
+  overrideBiometricsErrorMessage(messages);
```
shawyu-okta, line 83: It would be safer if we didn't rely on the mutation of the passed param here. It looks like this is working because `messages` and `formattedMessages` are references to the same array but it could cause confusion

test/testcafe/framework/page-objects/components/BaseFormObject.js
glenfannin-okta: I tried making it an h3, but since our messages are the first item in the container displayed (below the h1), this causes an a11y failure because it expects H tags to be in ascending order. 

I considered mutating the elements array, finding the Title element and displaying the error box below the title then making the error box title an h3 but that would also require updating the Infobox element to accept an option prop to instruct which H level to use. 

The above may not be a bad thing, but wanted to get your input on this before I made this sort of change (I actually did it but converted it back because I thought some may question it lol)

src/v3/src/transformer/messages/transform.ts
glenfannin-okta: refactored this to operate like our other transformers.

shawyu-okta (approved):

src/v3/src/transformer/messages/transform.ts
```diff
@@ -33,8 +34,7 @@ const fipsComplianceKeys = [
 ];
 
 export const CUSTOM_MESSAGE_KEYS = [
-  OV_OVERRIDE_MESSAGE_KEY.OV_FORCE_FIPS_COMPLIANCE_UPGRAGE_KEY_IOS,
-  OV_OVERRIDE_MESSAGE_KEY.OV_FORCE_FIPS_COMPLIANCE_UPGRAGE_KEY_NON_IOS,
+  ...fipsComplianceKeys,
```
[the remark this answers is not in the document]
glenfannin-okta: Oh, no, this is referencing a local variable that combines multiple related keys: 
```
const fipsComplianceKeys = [
  OV_OVERRIDE_MESSAGE_KEY.OV_FORCE_FIPS_COMPLIANCE_UPGRAGE_KEY_IOS,
  OV_OVERRIDE_MESSAGE_KEY.OV_FORCE_FIPS_COMPLIANCE_UPGRAGE_KEY_NON_IOS,
];
```

PR #3134, opened 2023-03-09
Fixes mobile OV enrollment reading transaction context

shawyu-okta:
## PR Checklist

### Issue:

- [OKTA-579900](https://oktainc.atlassian.net/browse/OKTA-579900)

### DIFF
--- a/src/v3/src/transformer/oktaVerify/transformOktaVerifyChannelSelection.test.ts
+++ b/src/v3/src/transformer/oktaVerify/transformOktaVerifyChannelSelection.test.ts
@@ -52,7 +52,10 @@ describe('TransformOktaVerifyChannelSelection Tests', () => {
   });
 
   it('should only append (sms/email) channel options when on mobile and qrcode is the selectedChannel', () => {
+    // on mobile, the data we need is on current transaction context rather than previous transaction
     prevTransaction.context = {
+    } as unknown as IdxContext;
+    transaction.context = {
       // TODO: OKTA-503490 temporary sln access missing relatesTo obj
       currentAuthenticator: {
         value: {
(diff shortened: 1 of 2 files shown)


src/v3/src/transformer/oktaVerify/transformOktaVerifyChannelSelection.test.ts
```diff
@@ -52,7 +52,10 @@ describe('TransformOktaVerifyChannelSelection Tests', () => {
   });
 
   it('should only append (sms/email) channel options when on mobile and qrcode is the selectedChannel', () => {
+    // on mobile, the data we need is on current transaction context rather than previous transaction
```
shawyu-okta, line 55: @glenfannin-okta I was wondering if we have a better explanation of why this happens?

glenfannin-okta: Yea, so what happens is, when you are on desktop, the first view that get loaded is the QR code view (which allows the user to scan the QR code for enrollment). On Mobile, the idea is that they cant scan, so instead of showing them a view that they cant do anything with, we automatically load the channel selection view (which shows radio options for them to pick either email or SMS to get a link for enrollment). Well, because of that process of pushing them to the channel selection view, the previous transaction doesnt have the expected remediation that desktop does. 

So the flow on desktop goes like this: 

(Current transaction object) Load OV Enroll Remediation where qrcode is selected channel -> if the user clicks cant scan; loads the channel selection view, and update the previous transaction object with the first remediation (which contains the context we expect in the next transformer) -> channel selection view is loaded and is able to grab the correct data from the previous transaction's context. 

But the flow on Mobile goes like this: 

(Current transaction object) Load OV Enroll Remediation where qrcode is selected channel -> notices user is on mobile and synthetically pushes them to channel selection view (previous transaction object is not updated in this scenario) -> channel selection view is loaded but is unable to grab the correct data from the previous transaction's context because the previous transaction is either null or not what the transformer expects.

I am sorry this was so long winded but I wanted to make sure I provided all the context :(

glenfannin-okta (approved):

src/v3/src/transformer/oktaVerify/transformOktaVerifyChannelSelection.test.ts
shawyu-okta: This is an amazing comment

PR #3144, opened 2023-03-13
OKTA-587104 : g3 : Localize icon titles properly

glenfannin-okta:
## Description:

The purpose of this PR is to fix an issue that was found where icon titles were not being localized properly (i.e. Authcoin and authenticator list icons).

## PR Checklist

### Issue:

- [OKTA-587104](https://oktainc.atlassian.net/browse/OKTA-587104)

### Screenshot/Video:

Screen Shot 2023-03-13 at 6 16 12 PM
Screen Shot 2023-03-13 at 6 16 29 PM
Screen Shot 2023-03-13 at 6 16 36 PM

### DIFF
--- a/src/v3/src/components/AuthCoin/authCoinConfigUtil.ts
+++ b/src/v3/src/components/AuthCoin/authCoinConfigUtil.ts
@@ -42,7 +42,7 @@ type AuthCoinConfig = {
   iconClassName: string;
 };
 
-const AuthCoinByAuthenticatorKey: Record<string, AuthCoinConfig> = {
+export const getAuthCoinConfiguration = (): Record<string, AuthCoinConfig> => ({
   [AUTHENTICATOR_KEY.CUSTOM_OTP]: {
     icon: CustomOTPIcon,
     customizable: true,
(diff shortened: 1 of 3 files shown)


src/v3/src/components/AuthCoin/authCoinConfigUtil.ts
```diff
@@ -42,7 +42,7 @@ type AuthCoinConfig = {
   iconClassName: string;
 };
 
-const AuthCoinByAuthenticatorKey: Record<string, AuthCoinConfig> = {
+export const getAuthCoinConfiguration = (): Record<string, AuthCoinConfig> => ({
```
shawyu-okta, line 45: Why do we need to convert this to a function? Or is this basically the fix?

shawyu-okta (approved):

src/v3/src/components/AuthCoin/authCoinConfigUtil.ts
shawyu-okta: Ok, that makes sense, thanks. Any thoughts on storing description as the i18n key and calling `loc` at the point of usage?

glenfannin-okta: :D I actually did consider that, the problem with that was since the authcoin is also used by Authenticator buttons, those contain their own localized descriptions which are translated within the transformers :( 

I considered adding a `key` property for this exact scenario, but then I would have to add logic to check if the key exists and use that otherwise use description but thought that would complicate things more :(

PR #3148, opened 2023-03-14
Enables safe mode and safe mode polling screens

shawyu-okta:
## Description:

* implement safe mode view
* implement safe mode polling view
* re-enable tests related to safe mode and safe mode polling

## PR Checklist

### Issue:

- [OKTA-546860](https://oktainc.atlassian.net/browse/OKTA-546860)

### DIFF
--- a/src/v3/src/transformer/layout/safeMode/transformSafeModePoll.ts
+++ b/src/v3/src/transformer/layout/safeMode/transformSafeModePoll.ts
@@ -13,8 +13,13 @@
 import {
   DescriptionElement,
   IdxStepTransformer,
+  IStepperContext,
   SpinnerElement,
+  StepperNavigatorElement,
   TitleElement,
+  UISchemaElement,
+  UISchemaLayout,
+  UISchemaLayoutType,
 } from '../../../types';
 import { generateRandomString, loc } from '../../../util';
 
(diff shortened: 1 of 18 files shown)


src/v3/src/types/schema.ts
```diff
@@ -171,34 +171,58 @@ export interface UISchemaElement {
...
-export interface UISchemaLayout {
+/**
+ * Parent interface for all Layout types with common properties
+ */
+interface Layout {
+  type: string;
+  key?: string;
+  elements: unknown[];
+  options?: Record<string, unknown>;
+}
```
shawyu-okta, line 182: Not sure if this is really helpful so feel free to veto if you think this parent type is more harmful than helpful

src/v3/src/types/schema.ts
```diff
@@ -450,19 +474,6 @@ export interface SuccessCallback extends UISchemaElement {
...
-  options?: {
-    defaultStepIndex: () => number;
-  }
-}
-
-export interface AccordionLayout {
-  type: UISchemaLayoutType.ACCORDION;
-  elements: AccordionPanelElement[];
-}
-
```
shawyu-okta, line 465: just a reordering so the layout types are grouped together

src/v3/src/transformer/layout/safeMode/transformSafeModePoll.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  const descriptionElement: DescriptionElement = {
+    type: 'Description',
+    contentType: 'subtitle',
+    options: {
+      content: loc(
+        'poll.form.message',
+        'login',
+        [pollIntervalSec],
+        { $1: { element: 'span' } },
```
glenfannin-okta, line 55: 🥳 Love that this is getting usage 🙏

src/v3/src/transformer/uischema/updateElementKeys.ts
```diff
@@ -23,7 +23,18 @@ import { traverseLayout } from '../util';
...
+      if ([
+        UISchemaLayoutType.STEPPER,
+        UISchemaLayoutType.ACCORDION,
+        UISchemaLayoutType.HORIZONTAL,
+        UISchemaLayoutType.VERTICAL,
+      ].includes((element.type) as UISchemaLayoutType)) {
+        return false;
+      }
+      return !!element.type;
+    },
```
glenfannin-okta, line 37: nit:
```suggestion
    // never add a synthetic key to layout elements' uischema
    predicate: (element) => ![
        UISchemaLayoutType.STEPPER,
        UISchemaLayoutType.ACCORDION,
        UISchemaLayoutType.HORIZONTAL,
        UISchemaLayoutType.VERTICAL,
      ].includes((element.type) as UISchemaLayoutType),
```

glenfannin-okta (approved):

src/v3/src/transformer/layout/safeMode/transformSafeModePoll.ts
glenfannin-okta: Also minor nit, should this have the `strong` class added to bold the number?

shawyu-okta: I _believe_ we generally don't want to use bold text within strings any more to emphasize text. @alexdahl-okta @narayun-okta can you confirm?

alexdahl-okta: Good question!

If we've landed on a static string (the number doesn't count down), then we probably don't need bold (we don't need to bring attention to a changing value).

I don't think we have a blanket rule against bold, and some draft guidelines from content design have a least one case where bold is recommended. Bold is a helpful tool in the toolbox designers would like to use sometimes.

But I believe there is globalization pressure against bold – it's not native to all writing systems, and makes translation more difficult. So do agree that we can avoid bold unless we have a compelling case for emphasis.

PR #3150, opened 2023-03-15
OKTA-584936 : FastPass mobile show Download OV page when not installed

leemchale-okta:
## Description:

This PR fixes the issue where an error was being thrown when pressing the FastPass button without Okta Verify installed on mobile devices.

## PR Checklist

### Issue:

- [OKTA-584936](https://oktainc.atlassian.net/browse/OKTA-584936)

### DIFF
--- a/src/v3/src/components/Widget/index.tsx
+++ b/src/v3/src/components/Widget/index.tsx
@@ -133,9 +133,11 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     try {
       if (typeof proxyIdxResponse !== 'undefined') {
         setIdxTransaction({
-          // @ts-expect-error proxyIdxResponse has missing types
+          // @ts-expect-error OKTA-589168 - RawIdxResponse type does not contain the properties from ProxyIdxResponse type yet
+          // remove ts-expect-error ASAP to avoid risk of trying to access property that does not exist
           rawIdxState: proxyIdxResponse,
-          // @ts-expect-error proxyIdxResponse has missing types
+          // @ts-expect-error OKTA-589168 - IdxContext type does not contain the properties from ProxyIdxResponse type yet
+          // remove ts-expect-error ASAP to avoid risk of trying to access property that does not exist
           context: proxyIdxResponse,
         });
         return;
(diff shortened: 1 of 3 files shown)


src/v3/src/components/Widget/index.tsx
```diff
@@ -130,6 +131,15 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
   const bootstrap = useCallback(async () => {
     await initLanguage();
     try {
+      if(typeof proxyIdxResponse !== 'undefined') {
+        setIdxTransaction({
+          // @ts-expect-error proxyIdxResponse has missing types
```
glenfannin-okta, line 136: nit: 
I would say its less of missing types and more incompatible types with `rawIdxState` and `context`

glenfannin-okta (approved):

src/v3/src/components/Widget/index.tsx
shawyu-okta: is there a better way to pass this typecheck without `ts-expect-error` if it's not a missing types issue? Setting `idxTransaction` is a central part of the widget behavior so I am concerned about ignoring the types we pass to it.

glenfannin-okta: We could create a ticket for the SDK team to add these properties (which exist in the proxyIdxResponse) to the IdxTransaction interface.

shawyu-okta: I think I'm a bit confused. The `ProxyIdxReponse` type is vastly different than `RawIdxState` or `IdxContext` types.

glenfannin-okta: Yep to both, or more like `context` and `RawIdxState` doesn't have the properties that `proxyIdxResponse` has. So if we just added those properties to `context` and `RawIdxState` in auth-js then we can avoid disabling the type check here.

shawyu-okta: @leemchale-okta this is ok for now since we don't have a good way to solve this immediately before EA.

Please file a JIRA to fix these typings. If we keep it like this for too long we run the real risk of accidentally reading a property we think is safe to read from `IdxTransaction` but actually does not exist.

shawyu-okta: Add the comment in both cases (here and below)

shawyu-okta (approved):

src/v3/src/components/Widget/index.tsx
leemchale-okta: okay will do 👍

PR #3161, opened 2023-03-20
g3: Adds back link to ODA Android enrollment screens

shawyu-okta:
## PR Checklist

### Issue:

- [OKTA-588701](https://oktainc.atlassian.net/browse/OKTA-588701)

### DIFF
--- a/src/v3/src/components/Link/Link.tsx
+++ b/src/v3/src/components/Link/Link.tsx
@@ -63,12 +63,15 @@ const Link: UISchemaElementComponent<{
   return (
     typeof href === 'undefined' ? (
       <LinkMui
-        // eslint-disable-next-line no-script-url
-        href="javascript:void(0)"
         onClick={onClick}
         aria-describedby={ariaDescribedBy}
         ref={focusRef}
         data-se={dataSe}
+        sx={{
+          '&:hover': {
+            cursor: 'pointer',
+          },
+        }}
       >
         {label}
       </LinkMui>
--- a/src/v3/src/components/Link/Link.tsx
+++ b/src/v3/src/components/Link/Link.tsx
@@ -74,6 +74,7 @@ const Link: UISchemaElementComponent<{
           '&:hover': {
             cursor: 'pointer',
           },
+          verticalAlign: 'baseline',
         }}
       >
         {label}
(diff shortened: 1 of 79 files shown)


src/v3/src/components/StepperLink/StepperLink.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const focusRef = useAutoFocus<HTMLButtonElement>(focus);
+  const widgetContext = useWidgetContext();
+
+  const handleClick = () => {
+    if (typeof nextStepIndex === 'function') {
+      setStepIndex!(nextStepIndex(widgetContext));
+    } else {
+      setStepIndex!(nextStepIndex);
+    }
```
glenfannin-okta, line 42: super nit since I know this is also the same way in the StepperButton 😅 
```suggestion
    if (typeof nextStepIndex === 'function') {
      setStepIndex(nextStepIndex(widgetContext));
    } else {
      setStepIndex(nextStepIndex);
    }
```

glenfannin-okta: Feel free to leave as is and we can clean up later as part of tech debt.

src/v3/src/components/Link/Link.tsx
```diff
@@ -63,12 +63,19 @@ const Link: UISchemaElementComponent<{
...
+        role="link"
         onClick={onClick}
         aria-describedby={ariaDescribedBy}
         ref={focusRef}
         data-se={dataSe}
+        sx={{
+          '&:hover': {
+            cursor: 'pointer',
+          },
+          verticalAlign: 'baseline',
```
glenfannin-okta, line 77: I know sometime a while back we used verticalAlign on a link and it caused this view to look uneven, I assume that is not the case anymore? I think it should be fine, because last time we used `middle` and that caused some issues, so I think `baseline` is ok. 
image

glenfannin-okta (approved):

src/v3/src/components/Link/Link.tsx
shawyu-okta: Yeah, the "Sign up" text was appearing slightly below the rest of the text (~3px) without any extra styling. When I put `baseline` it appears correctly.

src/v3/src/components/StepperLink/StepperLink.tsx
shawyu-okta: Oh yeah I can update this, I copy/pasted but I agree the type guard ensures we can call the function

PR #3166, opened 2023-03-22
OKTA-591835 : Focus on Title when page loads

leemchale-okta:
## Description:

This PR adds focusing on the Title element when a new page is loaded in SIW Gen 3.  This is to prevent a screen reader from continuing to read content from a previous page.  It is also to let screen reader users know that a new page has loaded and provides context.

## PR Checklist

### Issue:

- [OKTA-591835](https://oktainc.atlassian.net/browse/OKTA-591835)

### Screenshot/Video:

https://user-images.githubusercontent.com/107433508/227124613-75e78476-40a9-48f3-b284-dcdaaac43cfc.mov

### DIFF
--- a/src/v3/src/components/Title/Title.tsx
+++ b/src/v3/src/components/Title/Title.tsx
@@ -12,6 +12,8 @@
 
 import { Box, Typography } from '@okta/odyssey-react-mui';
 import { h } from 'preact';
+import { useEffect } from 'preact/hooks';
+import { useWidgetContext } from '../../contexts';
 
 import { TitleElement, UISchemaElementComponent } from '../../types';
 
@@ -19,7 +21,13 @@ const Title: UISchemaElementComponent<{
   uischema: TitleElement
 }> = (
   { uischema: { id, options } },
-) => (
+) => {
+  const { setStatusText } = useWidgetContext();
+  useEffect(() => {
+    setStatusText(options?.content + " page has loaded")
+  }, [])
+
+  return (
   <Box
     display="flex"
     justifyContent="flex-start"
@@ -33,6 +41,6 @@ const Title: UISchemaElementComponent<{
       {options?.content}
     </Typography>
   </Box>
-);
+)};
 
 export default Title;
--- a/src/v3/src/components/Title/Title.tsx
+++ b/src/v3/src/components/Title/Title.tsx
@@ -15,6 +15,7 @@ import { h } from 'preact';
 import { useEffect, useRef } from 'preact/hooks';
 
 import { TitleElement, UISchemaElementComponent } from '../../types';
+import style from './style.css';
 
 const Title: UISchemaElementComponent<{
   uischema: TitleElement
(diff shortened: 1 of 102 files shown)


src/v3/src/components/Title/Title.tsx
```diff
@@ -12,27 +12,40 @@
...
-    <Typography
-      id={id}
-      component="h2"
-      variant="h4"
-      data-se="o-form-head"
+) => {
+  const titleRef = useRef<HTMLTitleElement>(null);
+  useEffect(() => {
+    titleRef.current?.focus();
+  }, []);
```
glenfannin-okta, line 27: I am curious how does this work when autoFocus config option is set to true? Will it essentially be ignored? If so, we should probably check the autoFocus value to ensure this does not override that functionality? 

Also, is it semantically correct to focus on the Title instead of the H1? (Maybe that doesn't matter but just wanted to call it out to make sure we arent doing something unexpected).

shawyu-okta: Agree we should probably check on the `autoFocus` option and perhaps ignore this effect if another element should be focused instead?

leemchale-okta: autoFocus config option does still work properly when enabled, since the title gets focused first, and then the interactive element will take focus from it.  But I agree I think its better to add a check for `autoFocus` there to make sure.

alexdahl-okta: Yeah, good question! I think Title is best semantically. It's the start of the content that's changed, and its announcement will be what users want to know. (The H1 doesn't change between views.)

src/v3/test/integration/authenticator-enroll-email-magic-link-false.test.tsx
```diff
@@ -35,6 +36,7 @@ describe('Email authenticator enroll when magic link = false Tests', () => {
     });
 
     const headerEle = await findByRole('heading', { level: 2 });
+    await waitFor(() => expect(headerEle).toHaveFocus());
```
leemchale-okta, line 39: Need to wait for Title to have focus first, or else it will take away focus from the input field, causing it to fail to type the mock input

glenfannin-okta (approved):

src/v3/src/components/Title/Title.tsx
```diff
@@ -12,27 +12,46 @@
...
+      display="flex"
+      justifyContent="flex-start"
     >
-      {options?.content}
-    </Typography>
-  </Box>
-);
+      <Typography
+        id={id}
+        className={style.noOutline}
```
shawyu-okta, line 44: Can you add this type in `sx` rather than using `className`?

src/v3/src/components/Title/Title.tsx
```diff
@@ -12,27 +12,46 @@
...
+        component="h2"
+        variant="h4"
+        data-se="o-form-head"
+        ref={titleRef}
+        tabIndex={-1}
```
shawyu-okta, line 49: Is using `tabIndex` required for this to work?

shawyu-okta (approved):
Two comments but nothing blocking

src/v3/src/components/Title/Title.tsx
leemchale-okta: yeah cant focus on it without it, but setting it to -1 will prevent it from being able to navigate focus to it using the keyboard tab key

src/v3/src/components/Title/Title.tsx
leemchale-okta: ah yes we can good call.  i forgot about that `sx` prop lol

PR #3167, opened 2023-03-22
Remove the strict-ssl false check

vijetmahabaleshwar-okta:
## PR Checklist

### Issue:

- [OKTA-592627](https://oktainc.atlassian.net/browse/OKTA-592627)


shuowu-okta (approved):

scripts/setup.sh
```diff
@@ -9,14 +9,12 @@ export AUTHJS_VERSION=""
 export REGISTRY_REPO="npm-topic"
 export REGISTRY="${ARTIFACTORY_URL}/api/npm/${REGISTRY_REPO}"
 setup_service node v14.18.2
-# Use the cacert bundled with centos as okta root CA is self-signed and cause issues downloading from yarn
-setup_service yarn 1.21.1 /etc/pki/tls/certs/ca-bundle.crt
+setup_service yarn 1.22.19
```
shawyu-okta, line 12: can we change this to
```
setup_service node-and-yarn "14.18.2" "1.22.19"
```

shawyu-okta: This will put us on our newest way to install `yarn` as well

vijetmahabaleshwar-okta: Build fails with this error when I use `node-and-yarn` 
```
2023-03-22 21:26:38 Expected to find .nvmrc at /root/okta/okta-signin-widget/.nvmrc
2023-03-22 21:26:38 Repo must have its own .nvmrc file for this entry point.
```
We can tackle this as a separate issue I guess. Will go ahead and merge without this change.

vijetmahabaleshwar-okta: Well... Turns out, I'll just need to add an `.nvmrc` file in the root directory with the version. Will try that and see if it works.

.yarnrc.yml
```diff
@@ -0,0 +1,11 @@
+caFilePath: "${HOME}/.strap/okta/cacerts/Okta-Root-CA.pem"
+
+enableStrictSsl: true
```
shawyu-okta, line 3: I think this is default `true` but I like keeping it here explicitly

shawyu-okta (approved):

PR #3169, opened 2023-03-23
OKTA-586845 : g3 : Removing duplicate custom validation function

glenfannin-okta:
## Description:
`confirmPassword` field is a custom field added by the client, however, it is bundled with the `credentials.passcode` field for validation. We do not need to define a validation function for the `confirmPassword` field since it is already defined in the original field from the server, doing so causes issues like the field relaying the incorrect message to screenreader upon validation.

## PR Checklist

### Issue:

- [OKTA-586845](https://oktainc.atlassian.net/browse/OKTA-586845)

### Screenshot/Video:

https://user-images.githubusercontent.com/97472729/227276896-1ec83d6b-f024-4703-b741-c757cd0ffaf8.mov

### DIFF
--- a/src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
+++ b/src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
@@ -154,22 +154,6 @@ export const transformEnrollPasswordAuthenticator: IdxStepTransformer = ({
   dataSchema.fieldsToValidate.push('confirmPassword');
   dataSchema.fieldsToValidate.push('passwordMatchesValidation');
 
-  // Controls live field change validation
-  dataSchema.confirmPassword = {
-    validate: (data: FormBag['data']) => {
-      const confirmPw = data.confirmPassword;
-      if (!confirmPw) {
-        return [{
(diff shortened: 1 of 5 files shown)


src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
```diff
@@ -154,22 +154,6 @@ export const transformEnrollPasswordAuthenticator: IdxStepTransformer = ({
...
-        return [{
-          name: 'confirmPassword',
-          class: 'ERROR',
-          message: loc('model.validation.field.blank', 'login'),
-          i18n: { key: 'model.validation.field.blank' },
-        }];
-      }
-      return undefined;
-    },
-  };
```
glenfannin-okta, line 171: This was hiding in plain sight and didnt notice it on my last fix. 😢

glenfannin-okta: re: the comment, we no longer have live field validation, it only validates on blur now, so this is why this is no longer needed.

shawyu-okta (approved):

PR #3174, opened 2023-03-27
OKTA-588958 : Email magic link mobile browser icon fix

leemchale-okta:
## Description:

This PR adds the mobile browser icon to the email magic link flow

## PR Checklist

### Issue:

- [OKTA-588958](https://oktainc.atlassian.net/browse/OKTA-588958)

### Screenshot/Video:
image

### DIFF
--- a/src/v3/src/img/16pxDevice.svg
+++ b/src/v3/src/img/16pxDevice.svg
@@ -1,4 +1,4 @@
-<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="desktop" role="img">
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="desktop" role="img" aria-hidden="true">
 <title id="desktop-icon">Desktop browser</title>
 <path d="M4.5 14.5H8M11.5 14.5H8M8 14.5V13" stroke="#1662DD" class="siwIconStrokePrimary"/>
 <path d="M3 1.5H13C13.8284 1.5 14.5 2.17157 14.5 3V10C14.5 10.8284 13.8284 11.5 13 11.5H3C2.17157 11.5 1.5 10.8284 1.5 10V3C1.5 2.17157 2.17157 1.5 3 1.5Z" stroke="#1662DD" class="siwIconStrokePrimary"/>
--- a/src/v3/src/img/16pxDevice.svg
+++ b/src/v3/src/img/16pxDevice.svg
@@ -1,4 +1,4 @@
-<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="desktop" role="img">
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="desktop-icon" role="img">
 <title id="desktop-icon">Desktop browser</title>
 <path d="M4.5 14.5H8M11.5 14.5H8M8 14.5V13" stroke="#1662DD" class="siwIconStrokePrimary"/>
 <path d="M3 1.5H13C13.8284 1.5 14.5 2.17157 14.5 3V10C14.5 10.8284 13.8284 11.5 13 11.5H3C2.17157 11.5 1.5 10.8284 1.5 10V3C1.5 2.17157 2.17157 1.5 3 1.5Z" stroke="#1662DD" class="siwIconStrokePrimary"/>
--- a/src/v3/src/img/16pxDevice.svg
+++ b/src/v3/src/img/16pxDevice.svg
@@ -1,4 +1,4 @@
-<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="desktop-icon" role="img">
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="desktop-icon" role="img" aria-hidden="true">
 <title id="desktop-icon">Desktop browser</title>
 <path d="M4.5 14.5H8M11.5 14.5H8M8 14.5V13" stroke="#1662DD" class="siwIconStrokePrimary"/>
 <path d="M3 1.5H13C13.8284 1.5 14.5 2.17157 14.5 3V10C14.5 10.8284 13.8284 11.5 13 11.5H3C2.17157 11.5 1.5 10.8284 1.5 10V3C1.5 2.17157 2.17157 1.5 3 1.5Z" stroke="#1662DD" class="siwIconStrokePrimary"/>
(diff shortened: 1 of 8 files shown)


src/v3/src/img/16pxDevice.svg
```diff
@@ -1,5 +1,5 @@
 <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="desktop" role="img">
-<title id="desktop-icon">Browser</title>
+<title id="desktop-icon">Desktop browser</title>
```
shawyu-okta, line 2: Are these considered english leaks?

src/v3/src/transformer/terminal/transformEmailMagicLinkOTPOnlyElements.ts
```diff
@@ -85,16 +86,19 @@ export const transformEmailMagicLinkOTPOnly: TerminalKeyTransformer = (transacti
     };
   }
 
-  if (client?.value?.browser && client?.value?.os) {
+  const clientOs = client?.value?.os;
+  const clientBrowser = client?.value?.browser;
+  if (clientBrowser && clientOs) {
+    const isMobileDevice = clientOs === 'Android' || clientOs === 'iOS';
```
shawyu-okta, line 92: Check out `src/v3/src/util/browserUtils.ts` -- there are existing utilities to detect both mobile clients so we should use those if possible

src/v3/src/img/16pxDevice.svg
glenfannin-okta: These may be a good candidate for using aria-hidden since they are really just for visual effect. 

cc: @alexdahl-okta

src/v3/src/img/16pxMobileDevice.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="mobile" role="img">
+<title id="mobile-icon">Mobile browser</title>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M11.5 1H5.5C4.94772 1 4.5 1.44772 4.5 2V14C4.5 14.5523 4.94772 15 5.5 15H11.5C12.0523 15 12.5 14.5523 12.5 14V2C12.5 1.44772 12.0523 1 11.5 1ZM5.5 0C4.39543 0 3.5 0.895431 3.5 2V14C3.5 15.1046 4.39543 16 5.5 16H11.5C12.6046 16 13.5 15.1046 13.5 14V2C13.5 0.895431 12.6046 0 11.5 0H5.5ZM10 13V14H7V13H10ZM8.5 3C8.77614 3 9 2.77614 9 2.5C9 2.22386 8.77614 2 8.5 2C8.22386 2 8 2.22386 8 2.5C8 2.77614 8.22386 3 8.5 3Z" fill="#1662DD" class="siwIconFillPrimary"/>
+</svg>
```
glenfannin-okta, line 4: This may be a good time to use the `siwIconFillPrimaryDark` instead of primary, or we could just leave it for the ticket dedicated tot this 🤔

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -462,6 +462,9 @@ export default class BaseFormObject {
     return within(this.getButton(buttonName)).getByRole('img');
   }
 
+  getImageByTitle(title) {
+    return within(this.el).findByRole('img', {name: title});
+  }
```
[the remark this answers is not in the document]
leemchale-okta: Yeah I thought we did and was looking for it but apparently not

src/v3/src/img/16pxMobileDevice.svg
shawyu-okta: We should keep it as-is and do all of these at once so there is no inconsistency in the SVG colors

leemchale-okta: although the ticket only specifies Auth coins to be themed with primary dark.  Would this be included aswell?

src/v3/src/transformer/terminal/transformEmailMagicLinkOTPOnlyElements.ts
leemchale-okta: yeah I think it would make sense if we make an optional transaction parameter where we can pass in a transaction object and check there first

glenfannin-okta: instead of the entire transaction, maybe just the context? Or even just the client property from the context. You never know what someone in the future may do with it if the transaction is given to that function 😅

leemchale-okta: haha true good point i think just the client property would be okay

shawyu-okta: Hmm, I'm looking closely at the utility and I'm realizing these _might_ not be compatible, since they are checking different types of inputs (userAgent string vs what comes from `client` object). If you see a good way to handle this gracefully go for it, but I don't see as clear of a need to refactor these into one any more if they are too different.

leemchale-okta: Yeah I think youre right after looking at it more it looks like there isnt really a graceful way to handle it, they check very different inputs.  It could create more confusion than helpfulness with the util function so its probably best not to refactor into one

src/v3/src/img/16pxDevice.svg
alexdahl-okta: I think that's fair! We're using these icons as visual labels (instead of text labels) to decrease cognitive load and increase speed to the content. For low-vision users, they probably also want to get right to the content, and the content should be self-evident and sufficient for users to evaluate if the request is legitimate.

src/v3/src/img/16pxMobileDevice.svg
alexdahl-okta: Ah, yeah single color icons like these should be primary **base**.

(Two-color illustrations, including auth coins, should be primary dark and light.)

src/v3/src/img/16pxDevice.svg
```diff
@@ -1,5 +1,5 @@
 <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="desktop" role="img">
```
shawyu-okta, line 1: The aria-labelled by values should match the `id` below in `title` -- I think these should fail a11y checks if we check a view that has this SVG

src/v3/src/img/16pxMobileDevice.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="mobile" role="img">
```
shawyu-okta, line 1: Same comment about `aria-labelledby` here

src/v3/src/img/16pxDevice.svg
leemchale-okta: yeah looks like we will use aria-hidden on these.  I added them at first to see if the spec test fail, which it did since we cant find by title when it is aria-hidden.  I removed it because I was thinking this should be added in the work for this ticket https://oktainc.atlassian.net/browse/OKTA-594789

leemchale-okta: if its okay to just add it here I can do that

shawyu-okta (approved):

test/testcafe/framework/page-objects/components/BaseFormObject.js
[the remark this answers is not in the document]
shawyu-okta: Is this method still needed? I don't see it used any longer in this PR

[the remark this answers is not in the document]
leemchale-okta: ah yeah no longer needed.  Removing now

PR #3176, opened 2023-03-28
OKTA-589318 : siw g3 : enabling english leak detection test resolving issues

glenfannin-okta:
## Description:

The purpose of this PR is to enable the english leak detection test and resolve any pending english leaks that were found for implemented features. 

## PR Checklist

### Issue:

- [OKTA-589318](https://oktainc.atlassian.net/browse/OKTA-589318)


src/v3/src/transformer/layout/securityQuestion/securityQuestionEnroll.ts
```diff
@@ -74,7 +74,10 @@ export const transformSecurityQuestionEnroll: IdxStepTransformer = ({ transactio
   const predefinedQuestions = contextualData?.questions?.map((question: Question) => ({
     value: question.questionKey,
     label: question.question,
-  } as IdxOption));
+  } as IdxOption)).map((opt: IdxOption) => ({
+    ...opt,
+    label: loc(`security.${opt.value}`, 'login'),
+  }));
```
glenfannin-okta, line 80: Fixes english leaks for security question view.

src/v3/src/transformer/layout/idxTransformerMapping.ts
```diff
@@ -99,6 +99,10 @@ const TransformerMap: {
   }
 } = {
   [IDX_STEP.AUTHENTICATOR_ENROLLMENT_DATA]: {
+    [AUTHENTICATOR_KEY.EMAIL]: {
+      transform: transformEmailVerification,
+      buttonConfig: { showDefaultSubmit: false },
+    },
```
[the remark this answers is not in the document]
shawyu-okta: Was there a test that detected this?

src/v3/src/transformer/layout/securityQuestion/securityQuestionEnroll.ts
shawyu-okta: A bit wary of the string interpolation in the `loc` function. Is there any other way to do this?

src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
```diff
@@ -41,7 +41,9 @@ export const transformSecurityQuestionVerify: IdxStepTransformer = ({ transactio
   answerElement.translations = [{
     name: 'label',
     i18nKey: '',
-    value: relatesTo?.value?.profile?.question as string,
+    value: relatesTo?.value?.profile?.questionKey === 'custom'
+      ? relatesTo?.value?.profile?.question as string
+      : loc(`security.${relatesTo?.value?.profile?.questionKey}`, 'login'),
```
shawyu-okta, line 46: Same q here re: string interpolation

src/v3/src/transformer/phone/__snapshots__/phoneChallengeTransformer.test.ts.snap
```diff
@@ -71,7 +71,7 @@ Object {
       Object {
         "contentType": "subtitle",
         "options": Object {
-          "content": "oie.phone.verify.sms.codeSentText <span class=\\"strong no-translate\\">+1 XXX-XXX-4601</span>. oie.phone.verify.enterCodeText",
+          "content": "oie.phone.verify.sms.codeSentText <span class=\\"strong no-translate\\">+1 XXX-XXX-4601.</span> oie.phone.verify.enterCodeText",
```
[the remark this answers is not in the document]
shawyu-okta: that's odd, is there any way to change that behavior?

test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
```diff
@@ -27,12 +27,44 @@ const ignoredMocks = [
...
+  'terminal-return-otp-only-full-location-mobile-icon-unlock.json',
+  'terminal-return-otp-only-full-location.json',
+  'terminal-return-otp-only-no-location.json',
+  'terminal-return-otp-only-partial-location.json',
+  /** END of "page" issue **/
+  /** The below are erroring due to the user identifier, but it has the no-translate class on the element and doesn't fail in other mocks, must investigate TODO: OKTA-594871 */
+  'success-with-app-user.json',
+  'success.json',
+  /** END user identifier issue */
+];
```
shawyu-okta, line 66: Nice, thanks for annotating these too

test/testcafe/spec/TerminalView_spec.js
```diff
@@ -70,7 +70,6 @@ const terminalCustomAccessDeniedErrorMessageMock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/introspect')
   .respond(customAccessDeniedErrorMessage);
 
-fixture('Terminal view');
 fixture('Terminal view').meta('v3', true);
```
shawyu-okta, line 73: 🤔 hopefully this test was still running the whole time

glenfannin-okta: I hope so too lol I will find out on this branch, hopefully it doesn't fail 😅

src/v3/src/transformer/layout/idxTransformerMapping.ts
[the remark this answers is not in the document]
shawyu-okta: I see. Do you think we have a separate testcafe test to cover the flow that would use this mock as well?

src/v3/src/transformer/layout/securityQuestion/securityQuestionEnroll.ts
glenfannin-okta: The translation keys are all postfixed with the question key, so like `security.disliked_food`, `security.favorite_plush_toy`, etc. 

We will have to iterate through the options to get those which are present in the response to determine which keys are used. Granted if a new key is added (without it being added to translations) it would cause a L10 error, but hopefully that would be caught with leaks detection in okta-core.

src/v3/src/transformer/phone/__snapshots__/phoneChallengeTransformer.test.ts.snap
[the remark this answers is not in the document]
glenfannin-okta: The assert English leaks function is what makes that determination, any character (that is non empty) found that doesn't have the ok-pl tokens, it fails. 

The only real way to fix it is if we included the period inside of the translation key's value (this way OK-pl will convert it properly).

src/v3/src/transformer/layout/idxTransformerMapping.ts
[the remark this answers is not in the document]
glenfannin-okta: I just found it and enabled it, which also found another small bug 😓 but fixed that as well, will annotate where the bug was found and the fix. 

https://github.com/okta/okta-signin-widget/blob/master/test/testcafe/spec/EnrollAuthenticatorEmailFirst_spec.js

src/v3/src/transformer/phone/__snapshots__/phoneChallengeTransformer.test.ts.snap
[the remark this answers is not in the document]
glenfannin-okta: Yea, however, we should probably include the period inside of the translation text, tbh this entire should should be revisited because they are concatenating different keys together which is an anti-pattern w.r.t. translation. I can file a bug for this. That would ultimately resolve all of this 🙏

src/v3/src/transformer/layout/email/transformEmailAuthenticatorEnroll.ts
```diff
@@ -70,12 +68,12 @@ export const transformEmailAuthenticatorEnroll: IdxStepTransformer = ({ transact
   if (typeof useEmailMagicLink === 'undefined') {
     subTitleElement.options.content = loc('oie.email.enroll.subtitle', 'login');
   } else {
-    const redactedEmailAddress = nextStep.relatesTo?.value?.profile?.email;
```
glenfannin-okta, line 73: In enrollment flow (which is what this transformer is for) the identifier is pulled from the context instead of the authenticator block of the remediation. My assumption is because for enrolling, we are just using the email that is already displayed at the top of the page (not totally sure on this but just my guess). However, I confirmed this is how v2 does it for the enroll email flow and the test that was enabled verified as such.

test/testcafe/spec/EnrollAuthenticatorEmailFirst_spec.js
```diff
@@ -7,18 +7,21 @@ import { checkConsoleMessages } from '../framework/shared';
...
+  await challengeEmailPageObject.formExists();
   return challengeEmailPageObject;
 }
 
 const sendEmailMock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/introspect')
   .respond(emailAuthenticatorPreEnrollData)
   .onRequestTo('http://localhost:3000/idp/idx/credential/enroll')
+  .respond(emailVerification)
+  .onRequestTo('http://localhost:3000/idp/idx/challenge/poll')
```
glenfannin-okta, line 24: I added this because the original remediation has a poll apart of it, if the test takes too long, that poll could occur and fail the test. That could result in flaky tests, adding this here to return the same response resolves this issue.

src/v3/src/transformer/phone/__snapshots__/phoneChallengeTransformer.test.ts.snap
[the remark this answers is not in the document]
shawyu-okta: Agreed re: don't concatenate keys. We can improve this in the JIRA you filed, thanks!

test/testcafe/spec/EnrollAuthenticatorEmailFirst_spec.js
shawyu-okta: Nice! Thanks for making this test less flaky <3

src/v3/src/transformer/layout/email/transformEmailAuthenticatorEnroll.ts
shawyu-okta: Interesting, so the identifier in the transaction context is already redacted as well? Or it's not redacted but meant to not be redacted for enrollment?

src/v3/test/integration/__snapshots__/authenticator-enroll-email.test.tsx.snap
```diff
@@ -140,9 +140,15 @@ exports[`Email authenticator enroll when email magic link = true Tests should re
                     <p
                       class="MuiTypography-root MuiTypography-body1 MuiTypography-paragraph emotion-20"
                       data-se="o-form-explain"
-                      id="enroll-authenticator_Description_enroll-authenticator_Description_We_sent_you_a_verification_email_Click_the_verification_link_in_your_email_to_continue_or_enter_the_code_below_okta_email_aut2m2nnGGzJy8uO80g4_okta_email_aut2m2nnGGzJy8uO80g4_8"
+                      id="enroll-authenticator_Description_enroll-authenticator_Description_We_sent_an_email_to_<span_class=strong_no-translate>testUser@oktacom</span>_Click_the_verification_link_in_your_email_to_continue_or_enter_the_code_below_okta_email_aut2m2nnGGzJy8uO80g4_okta_email_aut2m2nnGGzJy8uO80g4_8"
```
shawyu-okta, line 143: whoa, this id value is huge and has html in it 😅 . is this expected?

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
```diff
@@ -74,11 +74,6 @@ export default class EnrollAuthenticatorPhonePageObject extends BasePageObject {
   async verificationLinkTextExists(emailAddress) {
     const text = await this.form.getSubtitle();
 
-    if (userVariables.v3) {
-      return text.includes('We sent you a verification email.');
-    }
-    else {
-      return text.includes(`We sent an email to ${emailAddress}.`);
-    }
```
shawyu-okta, line 82: For some reason I have a memory that we intentionally changed the text for this but I can't recall. Do you have any recollection of if we made a product change here?

shawyu-okta (approved):

src/v3/test/integration/__snapshots__/authenticator-enroll-email.test.tsx.snap
glenfannin-okta: Yea, I didn't anticipate this 😅  I will update the id generation method to strip HTML

test/testcafe/framework/page-objects/EnrollEmailPageObject.js
glenfannin-okta: I think in the early days of the widget, we had a different message (when we were trying to introduce new keys before settling on going for parity first). But since then we made it match v2 (or tried), this seems like it slipped through the cracks, when I compared the v2 view to our transformer, we were trying to grab the email from the profile (within the authenticator) but it doesn't exist there for enroll flow. 

v2 pulls it from the context and the test asserts it this way too :(

src/v3/src/transformer/phone/__snapshots__/phoneChallengeTransformer.test.ts.snap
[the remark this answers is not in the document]
glenfannin-okta: 🙏 https://oktainc.atlassian.net/browse/OKTA-595114

src/v3/src/transformer/layout/email/transformEmailAuthenticatorEnroll.ts
glenfannin-okta: Based on v2, it looks like its not redacted for enrollment, but it is redacted for verification flow. The redacted email is not present in the idx response for enrollment flow.

src/v3/test/integration/__snapshots__/authenticator-enroll-email.test.tsx.snap
glenfannin-okta: There was one dependency on this auto generated id. I created a tech debt ticket to remove this and update the dependency so it no longer relies on that 🙏 

https://oktainc.atlassian.net/browse/OKTA-595198

.bacon.yml
```diff
@@ -109,7 +109,7 @@ test_suites:
       sort_order: '15'
       timeout: '20'
       script_name: detect-english-leaks
-      criteria: OPTIONAL
+      criteria: MERGE
```
nikhilvenkatraman-okta, line 112: 💪

nikhilvenkatraman-okta (approved):

PR #3179, opened 2023-03-29
Added view and logic to start Chrome DTC probing

hanyuanzhang-okta:
## Description:
- Added the logic to start Chrome DTC probing by hitting `GET /idp/devices/{deviceIntegrationConfigId}/challenge?transactionHandle=XXX` and follow redirect to `GET /idp/devices/{deviceIntegrationConfigId}/challenge-response?transactionHandle=XXX` with an `iFrame`
- Added a spinner view, title and content during this process
- The remediation that triggers this logic is in the response of `POST /idp/idx/introspect` (no silent OV probing) and `POST /idp/idx/authenticators/poll` (OV probing): same remediation, but the `authenticatorChallenge` have a different `challengeMethod` of "CHROME_DTC", and the `href` points to the 1st GET above

## PR Checklist

### Issue:

- [OKTA-577451](https://oktainc.atlassian.net/browse/OKTA-577451)

### Reviewers:
@shawyu-okta @jaredperreault-okta 

### Screenshot/Video:
The new view shows up at 00:06 in this video
https://okta.box.com/s/xluo0dccp2kpili76gtkkt6hm4rt2p8h

https://bacon-go.aue1e.saasure.net/tasks/DOWNSTREAM_ARTIFACT_CREATE_COMMIT?taskId=45bedc826ff94afd83db8fabcd4db9c5


src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
```diff
@@ -170,18 +170,13 @@ const Body = BaseFormWithPolling.extend({
 
   doCustomURI() {
     this.ulDom && this.ulDom.remove();
+    const IframeView = createInvisibleIFrame('custom-uri-container', this.customURI);
```
jaredperreault-okta, line 173: Is there a scenario where both a `customURI` and `chromeDTC` will be rendered on the same page? If not, think the iframe id should be consistent

test/testcafe/framework/page-objects/DeviceChallengePollPageObject.js
```diff
@@ -24,8 +24,8 @@ export default class DeviceChallengePollViewPageObject extends BasePageObject {
     return this.body.find('iframe');
   }
 
-  getIframeAttributes() {
-    return Selector('#custom-uri-container').attributes;
+  getIframeAttributes(id) {
+    return Selector(id).attributes;
```
jaredperreault-okta, line 28: DOM selectors should remain in PageObject

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -573,14 +573,14 @@ test
     const deviceChallengePollPageObject = await setup(t);
     let iframe = await deviceChallengePollPageObject.getIframe();
     await t.expect(iframe.exists).ok({ timeout: 100 });
-    let attributes = await deviceChallengePollPageObject.getIframeAttributes();
+    let attributes = await deviceChallengePollPageObject.getIframeAttributes('#custom-uri-container');
```
jaredperreault-okta, line 576: DOM selectors should remain in PageObject

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
hanyuanzhang-okta: No, only one will be rendered on the same page.
Do you mean I can use the same iFrame id?

test/testcafe/spec/DeviceProbingChromeDTC_spec.js
```diff
@@ -0,0 +1,118 @@
+import { RequestMock } from 'testcafe';
+import { checkA11y } from '../framework/a11y';
+import DeviceChallengePollPageObject from '../framework/page-objects/DeviceChallengePollPageObject';
+import IdentityPageObject from '../framework/page-objects/IdentityPageObject';
+import identify from '../../../playground/mocks/data/idp/idx/identify';
+import deviceProbingChromeDTC from '../../../playground/mocks/data/idp/idx/device-probing-chrome-dtc';
+import identifyWithDeviceProbingLoopback from '../../../playground/mocks/data/idp/idx/identify-with-device-probing-loopback';
+
+const mockChromeProbingView = RequestMock()
```
hanyuanzhang-okta, line 9: Tests here are not complete end 2 end because device signals are not used in current backend implementation. Will create a jira to update the tests once the policy part (evaluate device signals based on ASOP) is done.

test/testcafe/framework/page-objects/DeviceChallengePollPageObject.js
jaredperreault-okta: I think you can use the same id for both iframes, so in this case, this PageObject method can remain unchanged.

If you did have a reason to use unique ids, then you should add another method to this PageObject class. Anytime you need to hardcode an id in a test, it should be done in the PageObject rather than the test itself. This way the DOM can be changed overtime, but the test will still pass (ideally)

hanyuanzhang-okta: Thank you Jared. I will create a new method with a new hard-coded id then, it will be easier for us to add tests on it.

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
shawyu-okta: Agree we could use the same iframe id, which simplifies the page object a bit since there is only one selector needed to get the iframe

hanyuanzhang-okta: Current design (short-term solution) is we do one device probing at a time (always OV probing first if eligible, then Chrome DTC probing if eligible). But in the future we might do all device probings at the same step, that means we could have both a `customURI` and `chromeDTC`, see https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2748980124/Device+Identification+Step+Modeling
If you have a strong opinion I can change them to use the same iFrame id (probably rename the id to be like `iframe-container`), if not I think we can keep them separated.

shawyu-okta: I don't feel strongly. I was thinking there wasn't a lot of downside to keeping the ids different to allow for flexibility for things like you mentioned. @jaredperreault-okta any objections?

jaredperreault-okta (approved):

PR #3185, opened 2023-03-30
OKTA-594871 : removing success* mock files from v3 ignore list

glenfannin-okta:
## Description:
The purpose of this PR is to remove the success* mock files from the **ignore** list for english leak detection. 

Context, these were added because they were failing locally, however, after some additional testing, it looks like because of the redirect, it failed to detect the `no-translate` class on the element because the app navigated away from the view. These are passing for v2 so my conclusion is that it is occurring fast enough to detect the content before redirection.  

My theory was correct, not an issue on bacon because of the speed at which the test executes (prior to redirection).
Bacon Link: https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&page=1&pageSize=6&sha=9a8f5004c5cedf2d8e99510a7387360857887907&tab=main

## PR Checklist

### Issue:

- [OKTA-594871](https://oktainc.atlassian.net/browse/OKTA-594871)

### DIFF
--- a/test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
+++ b/test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
@@ -58,10 +58,6 @@ const v3IgnoredMocks = [
   'terminal-return-otp-only-no-location.json',
   'terminal-return-otp-only-partial-location.json',
   /** END of "page" issue **/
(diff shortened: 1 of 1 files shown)


nikhilvenkatraman-okta (commented):
IIUC we might be losing coverage for the success mocks if we remove them right ? 
Couple of follow up question 
- Can localization for this view be tested via a e2e/unit test ?
- Have we considered using a different mock for success that avoids the redirection to ensure we have coverage

nikhilvenkatraman-okta (approved):

PR #2734, opened 2022-08-22
Display terminal view on OAuth error

oleksandrpravosudko-okta:
## Description:
* wrap auth-js errors with specialized error types
* show error view for 'terminal' errors in v1 and v2
* reject top level promise on 'non-recoverable' errors
* introduce user-facing error message for out-of-sync system clock

## PR Checklist

- [x] Have you verified the basic functionality for this change?
- [x] Did you add tests, as appropriate, following our [Automated Test guidelines](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2676497890/Automated+Testing+in+the+Signin+Widget)?
- [x] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?
- [x] Did you verify the change by running [downstream monolith artifact](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/102897979/Sign-in+Widget+Development#Sign-inWidgetDevelopment-Instructionstocreateandrunthedownstreamartifact(d16t))? (Provide link to build below)
- [x] Does this PR include noticeable changes to the UI? (If yes, attach screenshots/video below)

### Issue:

- [OKTA-516591](https://oktainc.atlassian.net/browse/OKTA-516591)

### Screenshot/Video:

https://user-images.githubusercontent.com/71440851/187463449-ea9fa5d0-242c-4934-badf-bec7bc2ac39e.mov

https://user-images.githubusercontent.com/71440851/187463737-f14b9cc7-8458-47fd-af08-ecd38341c417.mov

https://user-images.githubusercontent.com/71440851/187465870-4cd4afbf-51f2-4432-8154-888c5de88356.mov

### Downstream Monolith Build:

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&branch=d16t-okta-signin-widget-87ecbc8-6319c176&page=1&pageSize=6&tab=main

PR #3192, opened 2023-03-31
OKTA-594851 : SIW g3 : interaction code flow implementation and spec update

glenfannin-okta:
## Description:

The purpose of this PR is to remove a mock file from the english leak spec ignore list to increase test coverage for english leak testing. 

This PR also meets parity with SIW g2 for the interaction code flow scenario. When this was originally implemented in G3 it was a year ago before the recent changes in SIW g2. The changes in this PR uplifts g3 to meet parity with G2.

Bacon build: https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&page=1&pageSize=6&sha=17f56fc62f8f3625043eaa9690a5b296237abae9&tab=main

## PR Checklist

### Issue:

- [OKTA-594851](https://oktainc.atlassian.net/browse/OKTA-594851)


src/util/Errors.ts
```diff
@@ -57,7 +57,7 @@ export class U2FError extends CustomError {
   xhr: ErrorXHR;
   name = Enums.U2F_ERROR;
   constructor(err: ErrorContextData) {
-    super(err.xhr.responseJSON.errorSummary);
+    super(err.xhr?.responseJSON?.errorSummary);
```
glenfannin-okta, line 60: TS error was being thrown because xhr is undefinable.

src/util/Errors.ts
```diff
@@ -12,7 +12,7 @@
 
 import { loc } from '@okta/courage';
 import Enums from './Enums';
-import { ErrorXHR, ErrorContextData } from 'types/errors';
+import { ErrorXHR, ErrorContextData } from '../types/errors';
```
glenfannin-okta, line 15: Using this in v3 we had to correct the path here.

src/util/OAuthErrors.ts
```diff
@@ -23,15 +23,15 @@ class TypedOAuthError<T extends ErrorType> extends OAuthError {
   orginalError: AuthSdkError | SdkOAuthError;
   errorDetails: ErrorDetails
 
-  constructor(originalError, errorTypeCtor: new () => T) {
+  constructor(originalError: AuthSdkError | SdkOAuthError, errorTypeCtor: new () => T) {
```
glenfannin-okta, line 26: TS complained (error) that this was an `any` type and disallowed it.

src/v3/src/OktaSignIn/index.ts
```diff
@@ -163,8 +163,8 @@ export default class OktaSignIn {
           render(h(Widget, {
             events: this.events,
             authClient: this.authClient,
-            onSuccess: onSuccessWrapper,
-            onError: onErrorWrapper,
+            globalSuccessFn: onSuccessWrapper,
+            globalErrorFn: onErrorWrapper,
```
glenfannin-okta, line 167: Updating to match callback names from documentation (README).

test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
```diff
@@ -48,7 +48,6 @@ const v3IgnoredMocks = [
   'error-new-signup-email-exists.json',
   'error-new-signup-email.json',
   /** END of asterisk issue */
-  'success-with-interaction-code.json', // Receiving error regarding codeVerifier from auth-js, must investigate TODO: OKTA-594851
```
glenfannin-okta, line 51: Removed because it no longer fails with the recent changes.

src/v3/src/hooks/useInteractionCodeFlow.ts
```diff
@@ -0,0 +1,139 @@
...
+                  class: 'ERROR',
+                  i18n: { key: '' },
+                  message: typedError.errorDetails.errorSummary
+                    || loc('oform.error.unexpected', 'login'),
+                },
+                class: 'ERROR',
+              },
+            } as InfoboxElement);
+            setFormBag(terminalErrorFb);
+          } else {
```
glenfannin-okta, line 127: For information as to _why_ we are doing this, you can see this: https://github.com/okta/okta-signin-widget/pull/2734

src/v3/src/hooks/useInteractionCodeFlow.ts
```diff
@@ -0,0 +1,139 @@
...
+import { RenderResult } from '../../../types';
+import { getTypedOAuthError, RecoverableError } from '../../../util/OAuthErrors';
+import { redirectTransformer } from '../transformer/redirect';
+import { createForm } from '../transformer/utils';
+import { FormBag, InfoboxElement, WidgetProps } from '../types';
+import {
+  getMode, isAuthClientSet, isOauth2Enabled, loc,
+} from '../util';
+
+export const useInteractionCodeFlow = (
```
glenfannin-okta, line 25: On why we had to make these changes, see the latest in [interactionCodeFlow.js](https://github.com/okta/okta-signin-widget/blob/master/src/v2/client/interactionCodeFlow.js)

src/util/OAuthErrors.ts
```diff
@@ -23,15 +23,15 @@ class TypedOAuthError<T extends ErrorType> extends OAuthError {
...
     super(originalError.message);
     this.errorType = new errorTypeCtor();
     this.orginalError = originalError;
 
     this.errorDetails = {
       errorSummary: this.getErrorSummary(),
       errorCode: originalError.errorCode,
-      errorCauses: originalError.errorCauses,
+      errorCauses: 'errorCauses' in originalError ? originalError.errorCauses : undefined,
```
shawyu-okta, line 34: Does the default type of `originalError.errorCauses` not work here?

glenfannin-okta: The reason TS complains about this is because the type of the original error can be either `AuthSdkError` or `SdkOAuthError`. `AuthSdkError` contains errorCauses whereas `SdkOAuthError` does not, so I had to perform the check to ensure `errorCauses` exists before trying to use it. I believe this complains because strict type checking is enabled in v3, so when we use this util from v2 it enforces the strict type checking.

shawyu-okta: Interesting, since we sometimes return `undefined` in the ternary and TS is ok with that, I'd have expected it to be OK with `originalError.errorCauses` on its own since that would return `errorCauses` or `undefined`

src/v3/src/hooks/useInteractionCodeFlow.test.ts
```diff
@@ -0,0 +1,298 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import {
+  AuthSdkError, IdxStatus, IdxTransaction, TokenResponse,
+} from '@okta/okta-auth-js';
+import { waitFor } from '@testing-library/preact';
+import { renderHook } from '@testing-library/preact-hooks';
+import { getStubTransaction, getStubTransactionWithNextStep } from 'src/mocks/utils/utils';
+import { DescriptionElement, RedirectElement, WidgetProps } from 'src/types';
```
leemchale-okta, line 19: does using absolute paths here not cause any issues?

src/v3/src/hooks/useInteractionCodeFlow.ts
```diff
@@ -0,0 +1,139 @@
...
+    const shouldRedirect = redirect === 'always';
+    if (shouldRedirect) {
+      if (!redirectUri) {
+        throw new Error('redirectUri is required');
+      }
+
+      const searchParams = new URLSearchParams(redirectParams);
+      const redirectFormBag = redirectTransformer(
+        transaction,
+        `${redirectUri}?${searchParams.toString()}`,
```
leemchale-okta, line 78: if `interaction_code` or `state` in redirectParams are empty strings, would it still show up in the search params of the redirect uri with empty values?

src/v3/src/hooks/useInteractionCodeFlow.test.ts
glenfannin-okta: In tests they do not seem to cause a problem, however, in actual implementation files they definitely do 😅 

But I can update these for consistency 🙏

src/util/OAuthErrors.ts
glenfannin-okta: Yea, I thought the exact same thing, I was confused why it was even complaining in the first place, I made this update on a whim and it seemed to appease the TS gods 😅 

Removing the logic: 
image

Re-adding the logic:
image

src/v3/src/hooks/useInteractionCodeFlow.ts
glenfannin-okta: Actually, in this case interaction_code will never be undefined so I am updating that to remove the `|| ''`. Regarding state, based on the types it _could_ be undefined but in the real world, it would never be undefined. However, I will update the logic to properly account for this. 

Thank you!

nikhilvenkatraman-okta (approved):

src/v3/src/hooks/useInteractionCodeFlow.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  useEffect(() => {
+    if (typeof transaction === 'undefined') {
+      setFormBag(undefined);
+      return;
+    }
+
+    if (!([IdxStatus.TERMINAL, IdxStatus.SUCCESS].includes(transaction.status)
+        || !transaction.nextStep)) {
```
leemchale-okta, line 44: nit:  this looks a little hard to read, hard to tell if its all one condition or not.  I think extracting it out to a variable would make it easier to tell?

leemchale-okta (approved):

PR #3200, opened 2023-04-07
OKTA-594847 : SIW g3 : Granular/Enduser/Admin consent view implementation

glenfannin-okta:
## Description:
The purpose of this ticket is to implement the Granular consent view for SIW G3. This was a parity gap that was identified when we enabled the english leak spec test. 

In additional to enabling the view, we are now removing the consent-granular.json mock from from the ignore list of the english leak spec test to increase coverage for this spec.

Bacon Link: https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&page=1&pageSize=6&sha=ce51ca0a1520f81c3d0f43ecd94b12bf274d4a45&tab=main

Figma Links: 
https://www.figma.com/file/nteaCw12Lob4CPIORnQREc/branch/QdrZXPAamTc14WlceFU7Oz/Sign-In-Widget-Flow-Library?type=design&node-id=1029-14032&t=CjjtP6h22rGQCRKf-0
https://www.figma.com/file/nteaCw12Lob4CPIORnQREc/branch/XqrQo9ewJJyFNnQEme7qWy/Sign-In-Widget-Flow-Library?type=design&node-id=852-14860&t=WWsX2zBzuloNKMpS-0
https://www.figma.com/file/nteaCw12Lob4CPIORnQREc/Sign-In-Widget-Flow-Library?type=design&t=CDm9HOsOEihlP08A-0

## PR Checklist

### Issue:

- [OKTA-594847](https://oktainc.atlassian.net/browse/OKTA-594847)

### Screenshot/Video:
**Granular Consent View:** 

consent-granular

**Enduser Consent View:**
image

**Admin Consent View:**
image


src/v3/src/transformer/field/transform.ts
```diff
@@ -81,7 +81,8 @@ export const transformStepInputs = (
       const flattenedInputs = flattenInputs(input);
       return [...acc, ...flattenedInputs];
     }, [])
-    .filter((input) => input.visible !== false && input.mutable !== false)
+    .filter((input) => input.visible === true
+        || (input.visible !== false && input.mutable !== false))
```
glenfannin-okta, line 85: When fields are visible but immutable, we still need to transform them for display. This is what this update handles.

test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
```diff
@@ -33,7 +33,6 @@ const ignoredMocks = [
 const v3IgnoredMocks = [
   'authenticator-expiry-warning-custom-password.json', // TODO: Not yet implemented OKTA-594840
   'authenticator-expired-custom-password.json', // TODO: Not yet implemented OKTA-594842
-  'consent-granular.json', // TODO: Not yet implemented OKTA-594847
```
glenfannin-okta, line 36: Removing this increases the coverage for english leak detection for SIW Gen3

src/v3/src/components/Widget/index.tsx
```diff
@@ -351,6 +352,7 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
                 authCoinProps={buildAuthCoinProps(idxTransaction)}
               />
               <AuthContent>
+                <ConsentHeader />
```
glenfannin-okta, line 355: Created a tech debt ticket to handler all header elements separately: https://oktainc.atlassian.net/browse/OKTA-598852

nikhilvenkatraman-okta (approved):

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
```diff
@@ -0,0 +1,151 @@
...
+          titleAccess={altText}
+          classes="client-logo default-logo"
+          sx={{ width: '32px !important', height: '32px !important' }}
+          aria-hidden
+        />
+      )
+  );
+
+  const getAppIcon = () => {
+    const href: string | undefined = (clientUri as Record<string, string | undefined>)?.href;
```
leemchale-okta, line 77: how come this type of `Record<string, string | undefined>` was needed?

glenfannin-okta: Ahh, that is due to the type of the `app` object values are `Record<string, unknown>` since the value of the record is unknown, we have to cast it to what it is expected to be, otherwise TS complains.

src/v3/src/components/Checkbox/Checkbox.tsx
```diff
@@ -75,7 +84,14 @@ const Checkbox: UISchemaElementComponent<UISchemaElementComponentWithValidationP
         )}
         label={(
           <Fragment>
-            {label}
+            {labelInfo?.noTranslate ? (
+              <Box
+                component="span"
+                className="no-translate"
```
shawyu-okta, line 90: Is there any undesirable behavior if this `Box` is always present and only add the `className` conditionally like the other places?

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
```diff
@@ -0,0 +1,169 @@
...
+          src={logoHref}
+          width="32px"
+          height="32px"
+          alt={altText}
+          className="client-logo custom-logo"
+          aria-hidden="true"
+        />
+      )
+      : (
+        // TODO: This is a temporary icon used until UX provides one
```
shawyu-okta, line 67: Is there a followup ticket for this work?

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
shawyu-okta: Note that `clientUri` and `logo` can also be `undefined`. I don't think that will cause any runtime issues since we're using optional chaining but just be careful when casting since I think we're telling TS it won't be `undefined` when casting to this type.

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
```diff
@@ -0,0 +1,169 @@
...
+            className="title-text"
+            textAlign="center"
+          >
+            <Typography
+              component="h1"
+              variant="h3"
+              className="no-translate"
+            >
+              {appName}
+              &nbsp;
```
shawyu-okta, line 124: Curious why the nbsp is needed here

src/v3/src/transformer/layout/consent/transformAdminConsent.ts
```diff
@@ -0,0 +1,145 @@
...
+  };
+
+  const cancelButton: ButtonElement = {
+    type: 'Button',
+    label: loc('oform.cancel', 'login'),
+    options: {
+      type: ButtonType.BUTTON,
+      includeData: false,
+      actionParams: { consent: false },
+      step: stepName!,
```
shawyu-okta, line 135: Could we avoid the non-null assertion here by doing a typecheck with early return at the top of the transform?

Something like
```ts
if (!stepName) {
  return formBag;
}
```

src/v3/src/transformer/layout/consent/transformEnduserConsent.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  const termsOfServiceHref = app?.value?.termsOfService
+    && (app.value.termsOfService as Record<string, Record<string, unknown>>).href;
+  if (termsOfServiceHref) {
+    const tosLink: LinkElement = {
+      type: 'Link',
+      options: {
+        label: loc('consent.required.termsOfService', 'login'),
+        href: termsOfServiceHref as string,
```
shawyu-okta, line 96: According to the casting above it seems like `termsOfServiceHref` should be of type `Record<string, unknown>` but we cast to `string` here. Is this correct?

src/v3/src/transformer/layout/consent/transformGranularConsent.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  const submitButton: ButtonElement = {
+    type: 'Button',
+    label: loc('consent.required.consentButton', 'login'),
+    options: {
+      type: ButtonType.BUTTON,
+      includeData: true,
+      actionParams: { consent: true },
+      step: stepName!,
```
shawyu-okta, line 95: Same question here as in the other transformer about avoiding the non-null assertions.

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -211,6 +211,10 @@ export default class BaseFormObject {
   }
 
   getCancelButtonLabel() {
+    // in v3 buttons dont have a value prop
+    if (userVariables.v3) {
+      return this.el.find(CANCEL_BUTTON_SELECTOR).textContent;
```
shawyu-okta, line 216: Does this also work on the gen2 test?

src/v3/src/components/Checkbox/Checkbox.tsx
glenfannin-okta: This scenario should never happen, but if for some reason someone configured this incorrectly that caused this to happen, it would not have a negative impact on the UI as it would just display an empty span with that classname, and the class doesnt do anything other than allow our selenium tests to avoid asserting on it for translated (ok-pl) text.

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
glenfannin-okta: I will create one and link it here 🙏

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
glenfannin-okta: We need a space between teh app name and the Title, otherwise the content runs together, I changed this to use `{' '}` instead since in the vite branch, there is a lint rule disallowing &nbsp;

src/v3/src/transformer/layout/consent/transformEnduserConsent.ts
glenfannin-okta: `app?.value?.termsOfService` is of type `Record<string, unknown>`, since we want to get the actual href, we have to cast it as `Record<string, Record<string, unknown>>` then point to the `href` property. 

This is what the data looks like and why we have to do this: 

```
"termsOfService": {
     "href": "https://example.com/tos"
 },
```


And this is the typecasting logic: 

```const termsOfServiceHref = app?.value?.termsOfService
    && (app.value.termsOfService as Record<string, Record<string, unknown>>).href;```
 
 So `termsOfServiceHref` is actually of type unknown.

File: test/testcafe/framework/page-objects/components/BaseFormObject.js (hunk 11 above)
Comment (glenfannin-okta) [reply test/testcafe/framework/page-objects/components/BaseFormObject.js:216]: It doesn't because in Gen 2 the button is actually an `input` so `textContent` does not work, whereas in gen 3 it is a button (where `.value` does not work like in Gen 2).

File: src/v3/src/components/Checkbox/Checkbox.tsx (hunk 5 above)
Comment (shawyu-okta) [reply src/v3/src/components/Checkbox/Checkbox.tsx:90]: Oh I meant if we could restructure this like this:
```tsx
<Fragment>
  <Box
    component="span"
    className={labelInfo?.noTranslate ? 'no-translate' : undefined}
  >
    {labelInfo?.value as string}
  </Box>
//...
```

shawyu-okta (approved):
Just some minor type issues we discussed, otherwise lgtm!

src/v3/src/transformer/layout/consent/transformEnduserConsent.ts
glenfannin-okta: Discussed this, applying fix to type cast.

src/v3/src/components/Checkbox/Checkbox.tsx
glenfannin-okta: Ohhh, ok actually, I did originally consider this, the reason I went back on it was because it caused alot of (unnecessary) noise in the snapshot files. So to limit the number of changes in snaopshots I did this. But it would not hurt. Did you want me to do it this way?

shawyu-okta: Oh I see, I don't feel strongly, so up to you whichever you prefer

PR #3205, opened 2023-04-12
OKTA-566071 : SIW G3 : Applying optional label to fields

glenfannin-okta:
## Description:
The purpose of this PR is to apply the `Optional` label for optional field elements throughout the SIW. Additionally, for required fields displayed on the profile enrollment page, the ask is to display an asterisk next to the label of required fields along with the sub title denoting that all fields are required unless otherwise stated.

PR for the added translation: https://github.com/okta/okta-signin-widget/pull/3207

Bacon link: https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&page=1&pageSize=6&sha=35b17a53dd2fff85ade3163b56c07d2701adc218&tab=main

## PR Checklist

### Issue:

- [OKTA-566071](https://oktainc.atlassian.net/browse/OKTA-566071)

### Screenshot/Video:
Screen Shot 2023-04-12 at 6 49 27 PM
Screen Shot 2023-04-12 at 6 49 36 PM
Screen Shot 2023-04-12 at 6 49 52 PM

### DIFF
--- a/src/v3/src/components/InputPassword/InputPassword.tsx
+++ b/src/v3/src/components/InputPassword/InputPassword.tsx
@@ -20,6 +20,7 @@ import {
   InputBase,
   InputLabel,
   Tooltip,
+  Typography,
 } from '@okta/odyssey-react-mui';
 import { h } from 'preact';
 import { useState } from 'preact/hooks';
@@ -50,15 +51,16 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
   const value = useValue(uischema);
   const { loading } = useWidgetContext();
   const {
-    translations = [], focus, required, parserOptions, noTranslate,
+    translations = [], focus, parserOptions, noTranslate, showAsterisk,
   } = uischema;
   const {
     attributes,
-    inputMeta: { name },
+    inputMeta: { name, required },
   } = uischema.options;
   const label = getTranslation(translations, 'label');
   const hint = getTranslation(translations, 'hint');
   const explain = getTranslation(translations, 'bottomExplain');
+  const optionalLabel = getTranslation(translations, 'optionalLabel');
   const focusRef = useAutoFocus<HTMLInputElement>(focus);
   const parsedExplainContent = useHtmlContentParser(explain, parserOptions);
   const hasErrors = typeof errors !== 'undefined';
@@ -81,9 +83,22 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
     <Box>
       <InputLabel
         htmlFor={name}
-        required={required}
+        // To prevent asterisk from shifting far right
+        sx={{ justifyContent: showAsterisk ? 'flex-start' : undefined }}
       >
         {label}
+        {showAsterisk && (
+          <Box
+            component="span"
+            className="no-translate"
+            aria-hidden
+          >
+            &nbsp;&#42;
+          </Box>
+        )}
+        {required === false && (
+          <Typography variant="subtitle1">{optionalLabel}</Typography>
+        )}
       </InputLabel>
       {hint && (
         <FormHelperText
--- a/src/v3/test/integration/__snapshots__/enroll-profile-new.test.tsx.snap
+++ b/src/v3/test/integration/__snapshots__/enroll-profile-new.test.tsx.snap
@@ -54,29 +54,43 @@ exports[`enroll-profile-new should render form 1`] = `
                 class="MuiBox-root emotion-7"
               >
                 <div
-                  class="MuiBox-root emotion-11"
+                  class="MuiBox-root emotion-8"
+                >
+                  <p
+                    class="MuiTypography-root MuiTypography-body1 MuiTypography-paragraph emotion-12"
+                    data-se="o-form-explain"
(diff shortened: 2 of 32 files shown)


test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
```diff
@@ -32,20 +32,6 @@ const ignoredMocks = [
...
-  'enroll-profile-new-with-recaptcha-v2.json',
-  'enroll-profile-new.json',
-  'enroll-profile-with-idps.json',
-  'enroll-profile-with-password-returns-error.json',
-  'enroll-profile-with-password-returns-multiple-errors.json',
-  'enroll-profile-with-password.json',
-  'enroll-profile.json',
-  'error-new-signup-email-exists.json',
-  'error-new-signup-email.json',
-  /** END of asterisk issue */
```
glenfannin-okta, line 48: These were ignored because of the asterisk next to the label (rendered by MUI). We now either do not display the asterisk or when we do we apply the `no-translate` class to it to prevent it from failing the english leak test.

nikhilvenkatraman-okta: 💪

nikhilvenkatraman-okta (approved):

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -81,9 +83,22 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
     <Box>
       <InputLabel
         htmlFor={name}
-        required={required}
+        // To prevent asterisk from shifting far right
+        sx={{ justifyContent: showAsterisk ? 'flex-start' : undefined }}
```
leemchale-okta, line 87: what happens if `justifyContent` is set to `undefined`?  Does it fall back to a default?

glenfannin-okta: Yep, it falls back to css styles defined higher up. So in this case it will fall back to what Odyssey has defined for this element.

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -81,9 +83,22 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
...
+        {showAsterisk && (
+          <Box
+            component="span"
+            className="no-translate"
+            aria-hidden
+          >
+            &nbsp;&#42;
+          </Box>
+        )}
+        {required === false && (
```
leemchale-okta, line 99: are both 'showAsterisk` and 'required' properties needed?  Do they both share the same purpose of saying if the input is required or not?

leemchale-okta: Or is it because maybe the case where we want an input to be not required, but still show an asterisk, or vice versa?

src/v3/test/integration/__snapshots__/enroll-profile-new.test.tsx.snap
```diff
@@ -54,29 +54,43 @@ exports[`enroll-profile-new should render form 1`] = `
...
+                      class="no-translate MuiBox-root emotion-14"
                     >
-                       
-                      *
+                       *
                     </span>
                   </label>
                   <div
-                    class="MuiInputBase-root MuiInputBase-colorPrimary MuiInputBase-fullWidth emotion-14"
+                    class="MuiInputBase-root MuiInputBase-colorPrimary MuiInputBase-fullWidth emotion-17"
```
leemchale-okta, line 87: what is this change in `emotion` from?

src/v3/src/components/InputPassword/InputPassword.tsx
glenfannin-okta: Ah yes, so the asterisk will only show up on the profile enrollment page (per UX's request) and since we cant override the required property (because we need to know for those fields that are not required they should show the optional label) I had to create a new property to depict when we should show the asterisk. 

Added a blurb here: https://github.com/okta/okta-signin-widget/pull/3205/files#diff-bc22796207c263f6ef1317c9b0844cb530619b0f5ca1512fdb0314df25bd85faR245-R248

At the component level, we don't have the context of whether its profile enrollment or some other remediation, this is why I had to create a new property specifically for this.

src/v3/test/integration/__snapshots__/enroll-profile-new.test.tsx.snap
glenfannin-okta: This happened because a new [subtitle](https://github.com/okta/okta-signin-widget/pull/3205/files#diff-4a73bc5dbaad590a7b9422758173375c1db48abbcd3cceed5370ec963f54f381R59-R66) was added to this view which forced an increase in the total number of elements which causes the emotion element count to increase.

src/v3/src/components/InputPassword/InputPassword.tsx
leemchale-okta: ahh okay cool I got it now, thanks for the explanation 🙏

leemchale-okta (approved):

PR #3233, opened 2023-05-09
Add rel attribute to anchors for EndUserRemediationTerminalMessage

vincentle-okta:
## Description:

Updated EndUserRemediationTerminalMessage to have rel attributes added to all anchors with the correct values.

## PR Checklist

### Issue:

- [OKTA-606968](https://oktainc.atlassian.net/browse/OKTA-606968)

### Reviewers:

- @okta/device-platform

### DIFF
--- a/test/testcafe/framework/page-objects/components/BaseFormObject.js
+++ b/test/testcafe/framework/page-objects/components/BaseFormObject.js
@@ -124,6 +124,10 @@ export default class BaseFormObject {
     return this.el.find(FORM_INFOBOX_ERROR).innerText;
   }
 
+  getErrorBox() {
+    return Selector(CALLOUT);
+  }
+
   getErrorBoxHtml() {
     return this.getCallout(CALLOUT).innerHTML;
   }
--- a/test/testcafe/spec/TerminalView_spec.js
+++ b/test/testcafe/spec/TerminalView_spec.js
@@ -195,44 +195,68 @@ test.requestHooks(terminalCustomAccessDeniedErrorMessageMock)('should render cus
   await t.expect(terminalViewPage.form.getErrorBoxHtml()).eql('<span data-se="icon" class="icon error-16"></span><div class="custom-access-denied-error-message"><p>You do not have permission to perform the requested action.</p><ul class="custom-links"><li><a href="https://www.okta.com/" target="_blank" rel="noopener noreferrer">Help link 1</a></li><li><a href="https://www.okta.com/help?page=1" target="_blank" rel="noopener noreferrer">Help link 2</a></li></ul></div>');
 });
 
+const assertAnchorWithBlankTargetHaveCorrectRelAttribute = async (t, selector) => {
+  await t.expect(
+    selector
+      .find('a[target="_blank"]')
+      .filter((node) => {
+        const relValues = (node.getAttribute('rel') || '').split(' ');
+        return relValues.indexOf('noopener') < 0 || relValues.indexOf('noreferrer') < 0;
+      })
+      .exists).eql(false);
+};
+
+const assertSIWEndUserRemediationHeaders = async (t, terminalViewPage, explanationExists) => {
+  await t.expect(terminalViewPage.form.getErrorBox().find('.end-user-remediation-title').withExactText('Your device doesn\'t meet the security requirements').exists).eql(true);
+  await t.expect(terminalViewPage.form.getErrorBox().find('.end-user-remediation-explanation').withExactText('To sign in, make the following updates. Then, access the app again.').exists).eql(explanationExists);
+};
+
+const endUserRemediationActionFound = async (t, terminalViewPage, url, text, optionIndex) => {
+  await t.expect(
+    terminalViewPage.form.getErrorBox()
+      .find('.end-user-remediation-options')
+      .child((node) => {
+        let currentOptionIndex = 0;
+        let prevNode = node.previousSibling;
+
+        // Count the number of "Option N" siblings before the current node, to figure out what position are in
+        while (prevNode) {
+          if (prevNode.classList.contains('end-user-remediation-option')) {
+            currentOptionIndex += 1;
+          }
+          prevNode = prevNode.previousSibling;
+        }
+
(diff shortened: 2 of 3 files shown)


test/testcafe/spec/TerminalView_spec.js
```diff
@@ -195,44 +195,84 @@ test.requestHooks(terminalCustomAccessDeniedErrorMessageMock)('should render cus
...
 
+const assertAnchorWithBlankTargetHaveCorrectRelAttribute = async (t, selector) => {
+  await t.expect(
+    selector
+      .find('a[target="_blank"]')
+      .filter((node) => {
+        const relValues = (node.getAttribute('rel') || '').split(' ');
+        return relValues.indexOf('noopener') < 0 || relValues.indexOf('noreferrer') < 0;
+      })
+      .exists).eql(false);
```
jane-kuang-okta, line 206: Trying to understand this logic.. let me know if this is correct:
1. You take the selector and find the anchor tag.
2. You filter through the nodes in the anchor tag until we find the rel attr
3. You check if either of the `noopener` or `noreferrer` tags index are -1. If -1, it implies that it's not present in the rel attr.
4. Assuming the html has the rel attr, this should always return an empty object, therefore exists should be false.

Ok I actually think I understand it, but just felt a bit convoluted since you're checking for the inverse. Maybe it would be clearer if you checked for `return relValues.indexOf('noopener') >= 0 || relValues.indexOf('noreferrer') >= 0;` and assert exists is true?

Not sure if there's any flakiness in checking for a hardcoded `0` and `1` index in the return statement, if no flakiness then that would be my suggestion.

vincentle-okta: The goal of the test is to check:
- for all anchors that have `target="_blank"`
- filter for any that don't rel attributes with either value `noopener` or `noreferrer`
- if it exists, fail the the test.

The goal of the check is to assert not a single anchor with the missing attributes exist, as [it's considered bad](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)

jane-kuang-okta (approved):

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -124,6 +124,10 @@ export default class BaseFormObject {
     return this.el.find(FORM_INFOBOX_ERROR).innerText;
   }
 
+  getErrorBox() {
+    return Selector(CALLOUT);
```
shawyu-okta, line 128: Does `this.getCallout(CALLOUT)` work here instead?

test/testcafe/spec/TerminalView_spec.js
shawyu-okta: @vincentle-okta can you please move the selector/filter logic into the page object for the terminal view?

test/testcafe/spec/TerminalView_spec.js
```diff
@@ -195,44 +195,84 @@ test.requestHooks(terminalCustomAccessDeniedErrorMessageMock)('should render cus
...
+};
+
+const assertSIWEndUserRemediationHeaders = async (t, terminalViewPage, explanationExists) => {
+  await t.expect(terminalViewPage.form.getErrorBox().find('.end-user-remediation-title').withExactText('Your device doesn\'t meet the security requirements').exists).eql(true);
+  await t.expect(terminalViewPage.form.getErrorBox().find('.end-user-remediation-explanation').withExactText('To sign in, make the following updates. Then, access the app again.').exists).eql(explanationExists);
```
shawyu-okta, line 211: We want to avoid selecting (and particularly sub-selecting) on exact classnames in tests. We're undergoing an effort to eliminate all selectors in spec files, and try to eliminate whenever possible CSS classname selectors in page objects / form objects as well.

Would asserting that `terminalViewPage.form.getErrorBoxText().contains('target string').eql(true)` in these be sufficient?

test/testcafe/spec/TerminalView_spec.js
vincentle-okta: Moved the selector to `getAnchorsWithBlankTargetsWithoutRelevantAttributes`.

test/testcafe/framework/page-objects/components/BaseFormObject.js
vincentle-okta: Yes, that works, switched to that instead.

test/testcafe/spec/TerminalView_spec.js
vincentle-okta: I have changed some of the tests to not be as specific with selecting elements. Let me know if there are any more changes I should make.

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -248,6 +252,18 @@ export default class BaseFormObject {
     await this.t.click(this.el.find(selector));
   }
 
+  getAnchorsWithBlankTargetsWithoutRelevantAttributes() {
+    return Selector('a[target="_blank"]')
+      .filter((node) => {
+        const relValues = (node.getAttribute('rel') || '').split(' ');
+        return relValues.indexOf('noopener') < 0 || relValues.indexOf('noreferrer') < 0;
```
shawyu-okta, line 259: Use `.includes` instead

test/testcafe/spec/TerminalView_spec.js
```diff
@@ -199,40 +199,31 @@ test.requestHooks(endUserRemediationOneOptionMock)('should render end user remed
   const terminalViewPage = await setup(t);
   await checkA11y(t);
 
-  await t.expect(terminalViewPage.form.getErrorBoxHtml()).eql([
-    '<span data-se="icon" class="icon error-16"></span>',
```
shawyu-okta, line 203: Thanks for updating this test, it's much better already now that we aren't asserting on exact HTML.

shawyu-okta (approved):
one small comment but lgtm, thank you @vincentle-okta for making test improvements!

PR #3235, opened 2023-05-10
OKTA- 603136 : Gen 3 Idp authenticator

leemchale-okta:
## Description:

This PR implements the Idp authenticator into SIW G3.  This includes displaying Idp buttons on the identify & Enroll profile page.
This PR also enables IdpDiscovery in SIW G3, as well as adding Idp buttons in the widget config.

## PR Checklist

### Issue:

- [OKTA-603136](https://oktainc.atlassian.net/browse/OKTA-603136)

### Screenshot/Video:
**Identify/Register flow:**
Screenshot 2023-05-10 at 7 54 51 PM
Screenshot 2023-05-10 at 12 27 49 PM
Screenshot 2023-05-10 at 7 55 03 PM

**Authenticator flow:**

Screenshot 2023-05-10 at 7 56 48 PM
Screenshot 2023-05-10 at 7 57 01 PM


src/v3/src/transformer/button/transform.test.ts
```diff
@@ -39,9 +39,6 @@ jest.mock('./transformRegisterButton', () => ({
 jest.mock('./transformCancelButton', () => ({
   transformCancelButton: () => () => ({}),
 }));
-jest.mock('./transformLaunchAuthenticatorButton', () => ({
```
leemchale-okta, line 42: removed from this file since transformLaunchAuthenticatorButton file is no longer used

src/v3/src/transformer/button/transform.ts
```diff
@@ -32,5 +31,4 @@ export const transformButtons: TransformStepFnWithOptions = (options) => (formba
   transformRegisterButton(options),
   transformCancelButton(options),
   transformIDPButtons(options),
-  transformLaunchAuthenticatorButton(options),
```
leemchale-okta, line 35: moved `transformLaunchAuthenticatorButton` logic into `transformIDPButtons`

src/v3/src/transformer/button/transformLaunchAuthenticatorButton.ts
leemchale-okta, line 1: This file was removed and logic for the Launch authenticator was moved into `transformIdpButtons`

src/v3/src/transformer/button/transformIDPButtons.ts
```diff
@@ -10,93 +10,58 @@
...
-          ...transaction,
-          messages: [],
-          neededToProceed: pivRemediations,
-          availableSteps: pivRemediations as NextStep[],
-          nextStep: pivRemediations.find(({ name }) => name === IDX_STEP.PIV_IDP) as NextStep,
-        });
-      },
-    },
-  };
+  const fastPassButtonElement = getFastPassButtonElement(transaction);
```
leemchale-okta, line 39: moved fastpass button here to make it easier to group, order, and insert idp/custom buttons, along with the divider element.

src/v3/src/transformer/transaction/transform.ts
```diff
@@ -50,6 +61,54 @@ const transformRemediationNameForIdp: TransformStepFnWithOptions = (options) =>
   return formbag;
 };
 
+/**
+ * To support `idps` configuration in OIE.  Adds IDP buttons from widget config.
+ * https://github.com/okta/okta-signin-widget#openid-connect
+ */
+const injectIdPConfigButtonToRemediation: TransformStepFnWithOptions = (options) => (formbag) => {
```
leemchale-okta, line 68: This is for the openid-connect feature where idp buttons can be added in the Widget Config file.  More info on this can be found here: 
https://github.com/okta/okta-signin-widget/blob/b911ee7b11ecd5938726e651c1ad937a597bbdb8/docs/classic.md?plain=1#LL1295C7-L1295C7
https://github.com/okta/okta-signin-widget/pull/1237

test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js
```diff
@@ -125,20 +126,22 @@ test.requestHooks(mockWithoutIdentify)('should only render idp buttons with iden
...
   await setup(t);
 
-  await checkConsoleMessages({
-    controller: null,
-    formName: 'success-redirect',
-  });
+  if(!userVariables.v3) {
+    await checkConsoleMessages({
+      controller: null,
+      formName: 'success-redirect',
```
leemchale-okta, line 142: did not need to redirect to 'success-redirect' in V3

src/v3/src/transformer/button/transformIDPButtons.ts
```diff
@@ -10,93 +10,58 @@
...
+
+  // must be on identify or enroll profile page to display Idp buttons
+  if (!(containsIdentifyStep || containsEnrollProfileStep)) {
     return formbag;
   }
 
-  const { piv, idpDisplay } = widgetProps;
-  const isBottomDivider = typeof idpDisplay === 'undefined'
-    ? true
-    : idpDisplay.toUpperCase() === 'PRIMARY';
```
glenfannin-okta, line 48: We may want to keep this since it is a supported feature in Gen 2: https://github.com/okta/okta-signin-widget#idpdisplay

src/v3/src/transformer/layout/idp/transformIdpAuthenticator.ts
```diff
@@ -0,0 +1,72 @@
...
+        Util.redirectWithFormGet(nextStep?.href);
+      },
+    },
+  };
+
+  uischema.elements = [
+    titleElement,
+    descriptionElement,
+    submitButton,
+  ];
```
glenfannin-okta, line 69: Is there a scenario where fields are sent from the backend with these or is it always just custom UI elements added to the view?

src/v3/src/transformer/transaction/transform.ts
```diff
@@ -13,26 +13,37 @@
...
+  const { transaction } = options;
+  const { neededToProceed: remediations, nextStep, context } = transaction;
   if (!remediations.length) {
     return formbag;
   }
 
+  if (nextStep?.name === IDX_STEP.REDIRECT_IDP
+    && getAuthenticatorKey(transaction) === AUTHENTICATOR_KEY.IDP) {
+    // idp authenticator
+    const isVerifyFlow = Object.prototype.hasOwnProperty.call(context, 'currentAuthenticatorEnrollment');
```
glenfannin-okta, line 38: I think you can use es6 for this 

```
// currentAuthenticator is from enrollment flows and currentAuthenticatorEnrollment is from verify flows
const { rawIdxState: { currentAuthenticator, currentAuthenticatorEnrollment } } = transaction;
const isVerifyFlow = typeof currentAuthenticatorEnrollment !== 'undefined';
```

To make this re-usable, you can create it as a util function.

src/v3/src/transformer/transaction/transform.ts
```diff
@@ -50,6 +61,54 @@ const transformRemediationNameForIdp: TransformStepFnWithOptions = (options) =>
...
+          idp.className = idpConfig.className;
+        }
+        return {
+          name: IDX_STEP.REDIRECT_IDP,
+          type: idpConfig.type,
+          idp,
+          href: redirectUri,
+        };
+      });
+    transaction.neededToProceed = widgetRemediations.concat(redirectIdpRemediations);
```
glenfannin-okta, line 106: Usually (specifically for Gen 3), any modifications done to the `neededToProceed` property in the transaction, should also be replicated to the `nextStep` and `availableSteps` property of the transaction object. The reason for this is because _most_ of the transformers in Gen 3, rely heavily on the `nextSTep` or `availableSteps` properties and only in rare cases do we ignore those and go directly into the neededToProceed property.

src/v3/test/integration/__snapshots__/authenticator-piv-cac-verification.test.tsx.snap
```diff
@@ -755,44 +785,14 @@ exports[`authenticator-piv-cac-verification renders form with idpDisplay as SECO
...
-                  type="button"
-                >
-                  <span
-                    class="MuiButton-startIcon MuiButton-iconSizeMedium emotion-37"
-                  >
-                    <mocksvgicon />
-                  </span>
-                  Sign in with PIV / CAC card
-                </button>
-              </div>
```
glenfannin-okta, line 795: Just a reminder here that this may need attention since the test is explicitly testing `idpDisplay` config option feature.

test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js
glenfannin-okta: I think we will need this in v3 (because customers hook on these form values). 

I think we may be missing some logic for this scenario, see: https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/responseTransformer.js#L146-L171

test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js
```diff
@@ -159,19 +162,20 @@ test.requestHooks(logger, mockOnlyOneIdp)('Direct auth: does not auto redirect t
...
-    controller: null,
-    formName: 'success-redirect',
-  });
-
+  if (!userVariables.v3) {
+    await checkConsoleMessages({
+      controller: null,
+      formName: 'success-redirect',
+    });
+  }
```
glenfannin-okta, line 178: same as above: https://github.com/okta/okta-signin-widget/blob/master/src/v2/ion/responseTransformer.js#L146-L171

src/v3/src/transformer/button/transformIDPButtons.ts
leemchale-okta: based on this comment from Nara I thought we were going to remove this feature: https://oktainc.atlassian.net/browse/OKTA-602160?focusedCommentId=3581371

leemchale-okta: ah I see we can still keep this feature but just set the IDP buttons to the top by default,  but still allow the user the set the idpDisplay to `SECONDARY` if they want to display them at the bottom.

src/v3/src/transformer/layout/idp/transformIdpAuthenticator.ts
```diff
@@ -0,0 +1,72 @@
...
+    label: isEnrollStep
+      ? loc('mfa.enroll', 'login')
+      : loc('mfa.challenge.verify', 'login'),
+    options: {
+      type: ButtonType.BUTTON,
+      step: nextStep!.name,
+      isActionStep: false,
+      onClick: () => {
+        Util.redirectWithFormGet(nextStep?.href);
+      },
```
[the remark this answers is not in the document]
leemchale-okta: ohh awesome okay thanks!

test/testcafe/spec/IdentifyWithThirdPartyIdps_spec.js
leemchale-okta: ah okay yeah good point I will add this logic in

src/v3/src/transformer/button/transformIDPButtons.ts
glenfannin-okta: Yep exactly, if I am not mistaken I believe that is how we did it for Gen 3 with PIV (displayed it at the top by default instead of bottom unless they configured it to be displayed at the bottom).

src/v3/src/transformer/button/transformIDPButtons.test.ts
```diff
@@ -88,7 +89,8 @@ describe('IDP Button transformer tests', () => {
       .toBe('socialauth.divider.text');
   });
 
-  it('should add PIV/CAC IDP button when exists in remediation with SECONDARY display', async () => {
+  /* IDP buttons in G3 currently only show on top (no 'SECONDARY' idpDisplay) */
+  it.skip('should add PIV/CAC IDP button when exists in remediation with SECONDARY display', async () => {
```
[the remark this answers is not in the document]
leemchale-okta: feature added back in

src/v3/src/transformer/transaction/transform.ts
leemchale-okta: yeah this whole function for `openid-connect` was pretty much ported over quickly completely from v2 and doesnt really fit as well in v3 context.  Im going to see if I can extract this feature out of here in a way that wont need to modify the transaction at all.

src/v3/src/transformer/layout/idp/transformIdpAuthenticator.ts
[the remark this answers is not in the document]
leemchale-okta: I tried using the `LinkElement` but it still looks like a link.  Is there an example of how its used to look like a button?

src/v3/src/util/formUtils.ts
```diff
@@ -37,3 +54,214 @@ export const handleFormFieldBlur = (
...
+      type: 'Button',
+      label: displayName,
+      options: {
+        type: ButtonType.BUTTON,
+        step: IDX_STEP.PIV_IDP,
+        dataSe: 'piv-card-button',
+        classes: classNames.join(' '),
+        variant: 'secondary',
+        Icon: getIdpButtonIcon(type),
+        iconAlt: loc('piv.card', 'login'),
```
glenfannin-okta, line 255: `iconAlt` here needs to match the icon that is actually being used. 

I think the `getIdpButtonIcon(type)` could possibly return an object that contains the Icon as well as the `alt` text. One issue I see though is that the button label cannot be the same as the icon alt text (there is a explicit a11y rule against this). Looking into login.properties I don't see any keys specifically for these social auth icons that can be used (that isnt the same as the label) so you may have to create some like 

```
socialauth.facebook.icon.alt=Facebook
socialauth.google.icon.alt=Google
...
```

src/v3/test/integration/__snapshots__/authenticator-piv-cac-verification.test.tsx.snap
```diff
@@ -62,7 +62,6 @@ exports[`authenticator-piv-cac-verification renders form with custom PIV/CAC But
                     class="MuiButton-startIcon MuiButton-iconSizeMedium emotion-12"
                   >
                     <img
-                      alt="PIV Card"
```
glenfannin-okta, line 65: These should get fixed once you fix the comment above with the iconAlt property.

glenfannin-okta (commented):
Overall lgtm just those few comments re: the alt text. 

Also, just another observation, does the custom IDP button not display a button icon like the others? If not, I assume that is intentional?

glenfannin-okta (commented):
Overall lgtm just those few comments re: the alt text. 

Also, just another observation, does the custom IDP button not display a button icon like the others? If not, I assume that is intentional?

src/v3/src/util/formUtils.ts
glenfannin-okta: An alternative to the above if you dont want to create alt text for each of these icons, is you could just leave it as an empty string since the icon on the button uses aria-hidden. The only downside to that is when you hover over the image it wont display alternative text but that may be acceptable since the button label will exist. 

I'll leave that up to you. 🙏

leemchale-okta: Oops my bad this was supposed to be up above for the piv button lol.

I think since the button uses aria-hidden its probably fine to leave it as an empty string

src/v3/src/transformer/transaction/transform.ts
```diff
@@ -13,26 +13,37 @@
...
 const transformRemediationNameForIdp: TransformStepFnWithOptions = (options) => (formbag) => {
-  const { transaction: { neededToProceed: remediations, nextStep } } = options;
+  const { transaction } = options;
+  const { neededToProceed: remediations, nextStep } = transaction;
   if (!remediations.length) {
     return formbag;
   }
 
+  if (nextStep?.name === IDX_STEP.REDIRECT_IDP
+    && getAuthenticatorKey(transaction) === AUTHENTICATOR_KEY.IDP) {
```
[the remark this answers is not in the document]
leemchale-okta: i dont think there should be any overlap since piv cant be used as a factor for MFA, only for SSO on identify.  So there shouldnt be a case where there is a remediation with type `PIV_TYPE` with an authenticator key of `IDP`

leemchale-okta:
> Overall lgtm just those few comments re: the alt text.
> 
> Also, just another observation, does the custom IDP button not display a button icon like the others? If not, I assume that is intentional?

yeah looks like we decided to not have a default icon here: https://oktainc.atlassian.net/browse/OKTA-609477?focusedCommentId=3604948

src/v3/src/util/formUtils.ts
```diff
@@ -37,3 +54,215 @@ export const handleFormFieldBlur = (
...
+  const idpButtonElements = redirectIdpRemediations.map((idpObject) => {
+    let type = idpObject.type?.toLowerCase() || '';
+    let displayName;
+
+    if (IDP.SUPPORTED_SOCIAL_IDPS.includes(type) === false) {
+      type = 'general-idp';
+      // OKTA-396684 - makes sure that custom idps always have a name
+      displayName = loc('customauth.sign.in.with.label', 'login', [idpObject.idp?.name]);
+    } else {
+      displayName = loc(`socialauth.${type}.label`, 'login');
```
glenfannin-okta, line 228: I believe we try to avoid using variables when referencing translation keys, the reason for this is because I think we eventually want to enforce that the key exists (possibly using typescript?) and using a variable like this will prevent that. 

What we have done in the past is put the string in a Record/map and based on the key (which can be accessed by passing it the variable) passing that value to the `loc` function . See [this](https://github.com/okta/okta-signin-widget/blob/0.5/src/v3/src/constants/idxConstants.ts#L213-L218) for reference

glenfannin-okta (approved):
One last comment, sorry I missed that on the first two passes 😢

glenfannin-okta (approved):
One last comment, sorry I missed that on the first two passes 😢

PR #3274, opened 2023-06-13
Add enhanced polling support to resolve 400 bad request error

SohamKolhatkar-okta:
## Description:
- Few customers have reported an intermittent issue during authentication, where a bad request error is shown on SIW
- This is occurring due to 2 poll calls, which are made at nearly the same time.
When first poll hits server, it completes the transaction and deletes it. When the second poll hits the server, it cannot find the transaction and returns a 400. This 400 response returns to SIW quicker, which is why the error shows up on the page. Then the response for the first poll call returns to SIW, which is a success redirect and the user gets access.

Wiki: https://oktawiki.atlassian.net/wiki/spaces/~988051222/pages/2779389341/SIW+Polling+400+errors+enhancement

working on tests...

## PR Checklist

### Issue:

- [OKTA-609078](https://oktainc.atlassian.net/browse/OKTA-609078)

### Screenshot/Video:
FF disabled:

https://github.com/okta/okta-signin-widget/assets/86258428/26bde582-cb8b-4f6a-bbb6-eb87a975ebf1

FF enabled:

https://github.com/okta/okta-signin-widget/assets/86258428/cbbac614-5748-4b44-8051-9409db4889c5

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&branch=d16t-okta-signin-widget-274a99e-64895b56&page=1&pageSize=6&tab=main


src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
```diff
@@ -63,6 +63,8 @@ const Body = BaseFormWithPolling.extend({
     let challengeRequest = deviceChallenge.challengeRequest !== undefined ? deviceChallenge.challengeRequest : '';
     let probeTimeoutMillis = deviceChallenge.probeTimeoutMillis !== undefined ?
       deviceChallenge.probeTimeoutMillis : 100;
+    let isEnhancedPollingEnabled = deviceChallenge.enhancedPollingEnabled !== undefined
+      && deviceChallenge.enhancedPollingEnabled === true;
```
glenfannin-okta, line 67: nit:
```suggestion
    const isEnhancedPollingEnabled = typeof deviceChallenge.enhancedPollingEnabled === 'boolean'
      && deviceChallenge.enhancedPollingEnabled;
```

It doesn't look like `isEnhancedPollingEnabled` is reassigned, but correct me if I am wrong, but if not, should use const instead of let.

yannongli-okta: @glenfannin-okta we can have a future enhancement to put all possible values as const. Right now we will just keep it as let for consistency and also because we have tested it end to end. @SohamKolhatkar-okta please create a Jira for this purpose.

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
```diff
@@ -109,6 +111,9 @@ const Body = BaseFormWithPolling.extend({
               // once the OV challenge succeeds,
               // triggers another polling right away without waiting for the next ongoing polling to be triggered
               // to make the authentication flow goes faster 
+              if (isEnhancedPollingEnabled === true) {
```
shawyu-okta, line 114: See our conversations offline -- we should change this:
```suggestion
              if (deviceChallenge.enhancedPollingEnabled !== false) {
```

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
shawyu-okta: See my comment below, if you use the exact code change I provided this line shouldn't be necessary. Additionally in general any `typeof` check and checks for `undefined` are unnecessary if we directly do a strict equals/not-equals check against a boolean.

SohamKolhatkar-okta: Sure, will create a JIRA tomorrow, when I have time.

shawyu-okta (approved):

glenfannin-okta (approved):

yannongli-okta (approved):

PR #3279, opened 2023-06-15
OKTA-609477 : Add missing social idp icons

leemchale-okta:
## Description:

This PR adds the remaining missing social idp button icons in SIW gen 3.

## PR Checklist

### Issue:

- [OKTA-609477](https://oktainc.atlassian.net/browse/OKTA-609477)


src/v3/src/util/getSocialIdpButtonIcon.ts
```diff
@@ -0,0 +1,59 @@
...
+import orcidIconSvg from '../img/socialButtonIcons/orcid.svg';
+import paypalIconSvg from '../img/socialButtonIcons/paypal.svg';
+import quickbooksIconSvg from '../img/socialButtonIcons/quickbooks.svg';
+import salesforceIconSvg from '../img/socialButtonIcons/salesforce.svg';
+import spotifyIconSvg from '../img/socialButtonIcons/spotify.svg';
+import xeroIconSvg from '../img/socialButtonIcons/xero.svg';
+import yahooIconSvg from '../img/socialButtonIcons/yahoo.svg';
+import yahooJapanIconSvg from '../img/socialButtonIcons/yahoo-japan.svg';
+
+export const getSocialIdpButtonIcon = (idpType: string) : string => {
```
glenfannin-okta, line 34: Instead of a function, would it make more since to just export the constant `idpIconMapping`, then use that constant in the formUtils?

leemchale-okta: yeah that does make more sense, changing that now

src/v3/src/util/idpButtonIconMap.ts
```diff
@@ -0,0 +1,56 @@
...
+import orcidIconSvg from '../img/socialButtonIcons/orcid.svg';
+import paypalIconSvg from '../img/socialButtonIcons/paypal.svg';
+import quickbooksIconSvg from '../img/socialButtonIcons/quickbooks.svg';
+import salesforceIconSvg from '../img/socialButtonIcons/salesforce.svg';
+import spotifyIconSvg from '../img/socialButtonIcons/spotify.svg';
+import xeroIconSvg from '../img/socialButtonIcons/xero.svg';
+import yahooIconSvg from '../img/socialButtonIcons/yahoo.svg';
+import yahooJapanIconSvg from '../img/socialButtonIcons/yahoo-japan.svg';
+
+export const idpButtonIconMap: Record<string, string> = {
```
shawyu-okta, line 34: Is there any way to type the Record return type so the key is an enum based on the supported IDPs in `src/util/IDP.js`?

lesterchoi-okta (approved):
nit: is it necessary to call them "social **button** icons"? Wouldn't social icons cover it?

src/v3/src/img/socialButtonIcons/github.svg
lesterchoi-okta, line 1: @alexdahl-okta I'm concerned about these icons against branded/themed buttons... thoughts?

alexdahl-okta: Definitely debatable!

Out of the box, I don't think I'm worried. Themed buttons have a neutral or extremely light background color. We (SIW) control the lightness of the themed button color. Contrast and brand-friendliness feels okay.
Frame 449

But could be more worried for custom theme overrides, such as dark button background colors. We could bake a background/frame into the icons. Or, as the customer is defining styles for the button, rely to them to define styles for the icons to match their design goal.

We could debate in a followup?

src/v3/src/util/idpButtonIconMap.ts
shawyu-okta: Seems okay to not do the work to enforce type safety here as long as we have decent default behavior in the case that the lookup in the map returns `undefined`. If we want a "default icon" (not sure if we'd actually want this) we could add an icon to the map with a key of `''`.

shawyu-okta: I want to add this definitely isn't blocking -- just was curious if we wanted to consider doing it while we're here.

shawyu-okta (approved):

leemchale-okta:
> nit: is it necessary to call them "social **button** icons"? Wouldn't social icons cover it?

My thinking was that the icons were specific to the social Idp buttons, but on second thought I think just social icons would cover it since it should be more generic

src/v3/src/util/idpButtonIconMap.ts
leemchale-okta: looks like default icon is not necessary: https://oktainc.atlassian.net/browse/OKTA-609477?focusedCommentId=3601369

leemchale-okta: and currently we do have default behavior in place to handle if the social icon map lookup return `undefined` (it wont try and render an svg)

alexdahl-okta:
> My thinking was that the icons were specific to the social Idp buttons, but on second thought I think just social icons would cover it since it should be more generic

FWIW I think these are IdP icons, not "social" icons. (For example: sign in with Salesforce or smart card aren't really social login, but they are IdPs.)

But it's internals and the set is fuzzy (FastPass isn't an IdP), so 🤷‍♂️

leemchale-okta:
> FWIW I think these are IdP icons, not "social" icons. (For example: sign in with Salesforce or smart card aren't really social login, but they are IdPs.)
> 
> But it's internals and the set is fuzzy (FastPass isn't an IdP), so 🤷‍♂️

Thats a good point, I agree I think Idp icons is more appropriate

src/v3/src/util/formUtils.ts
```diff
@@ -238,7 +225,7 @@ export const getIdpButtonElements = (
     let type = idpObject.type?.toLowerCase() || '';
     let displayName;
 
-    const buttonI18key = SOCIAL_IDP_TYPE_TO_I18KEY[type];
+    const buttonI18key = SOCIAL_IDP_TYPE_TO_I18KEY[type as IdpKey];
```
shawyu-okta, line 228: I think we should move this type casting as close to where it's needed as possible. In this case we would move it to the definition of `let type` since we already know here that this is not an accurate type on this line (`type` can also be `''`)

That might look like this (although not certain of the TS syntax here):
```ts
let type = idpObject.type?.toLowerCase() as (IdpKey | undefined) || '';
```

shawyu-okta: This will make it clear when the value is which type. i.e. the left side of the logical OR operator has a union type of `(IdpKey | undefined)` and then we may also look at the right side if the left side had value `undefined`, so the total union type of `type` (a bit confusing 😄 ) assigned to the variable is `IdpKey | ''`.

shawyu-okta (approved):

PR #3282, opened 2023-06-22
Update login_fr.properties

nicoduj:
## Description:

Add parenthesis in case of 1 last password configuration (no "s" in this case)


shawyu-okta:
@nicoduj Thanks for the suggestion. We will pass this info to our translation team and will get back to you here.

_Internal ref: OKTA-621256_

nicoduj:
Thanks for considering. I have updated with another suggestion, but feel free to check with your transaltion team

PR #3290, opened 2023-06-28
OKTA-623397 : SIW G3 : Remove the unnecessary afterRender event trigger

glenfannin-okta:
## Description:
SessionStorage_spec.js was flaky in SIW Gen 3 due to an extra (unnecessary) event being triggered. This PR prevents the unnecessary event from being triggered. 

## PR Checklist

### Issue:

- [OKTA-623397](https://oktainc.atlassian.net/browse/OKTA-623397)

### DIFF
--- a/src/v3/src/util/idxUtils.ts
+++ b/src/v3/src/util/idxUtils.ts
@@ -15,7 +15,6 @@ import {
   FieldError,
   IdxMessage,
   IdxRemediation,
-  IdxStatus,
   IdxTransaction,
   Input,
   NextStep,
@@ -355,19 +354,3 @@ export const isVerifyFlow = (transaction: IdxTransaction): boolean => {
 
 // @ts-expect-error OKTA-627610 captcha missing from context type
 export const isCaptchaEnabled = (transaction: IdxTransaction): boolean => typeof transaction.context?.captcha !== 'undefined';
-
-export const shouldSkipAfterRender = (idxTransaction: IdxTransaction): boolean => {
-  const {
-    nextStep: { name: stepName } = {},
-    neededToProceed,
-    messages,
-    context,
-  } = idxTransaction;
-  // When a transaction is in this scenario, it usually indicates a successful
-  // interaction code flow token response which means we should not trigger the afterRender event.
-  return idxTransaction.status === IdxStatus.SUCCESS
-    && stepName === 'cancel'
-    && neededToProceed.length === 0
-    && !messages?.length
-    && !context.messages;
-};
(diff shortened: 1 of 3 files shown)


test/testcafe/spec/SessionStorage_spec.js
```diff
@@ -263,6 +264,9 @@ test.requestHooks(identifyChallengeMockWithError)('shall clear when session.stat
 
   // Refresh
   await challengeEmailPageObject.refresh();
+  // wait for refresh to complete
+  await t.wait(1000);
+  await t.expect(challengeEmailPageObject.formExists()).eql(true);
```
shawyu-okta, line 269: Are both of these needed? I thought that the second assertion would be sufficient since it should wait a certain amount of time for the assertion to become true

shawyu-okta: You might also be able to extend the assertion timeout inline instead of using both

src/v3/src/util/idxUtils.ts
```diff
@@ -351,3 +352,19 @@ export const isVerifyFlow = (transaction: IdxTransaction): boolean => {
...
+    messages,
+    context,
+  } = idxTransaction;
+  // When a transaction is in this scenario, it usually indicates a successful
+  // interaction code flow token response which means we should not trigger the afterRender event.
+  return idxTransaction.status === IdxStatus.SUCCESS
+    && stepName === 'cancel'
+    && neededToProceed.length === 0
+    && !messages?.length
+    && !context.messages;
```
shawyu-okta, line 369: This all feels a bit smelly -- we're carving out a very specific set of cases to skip this hook. Is it necessarily bad to run the hook in this scenario?

glenfannin-okta: Oh yea, I agree with the smelly part (but wondering if there is another place that can handle this), the problem is that we are triggering an event that isn't really representative of anything on the UI. The event that is triggered is like this `{ controller: null, form: 'cancel' }`, so if a customer consumes this and relies on the afterRender event to do things or gain information, this could negatively impact them or at least confuse them since this is triggered right before an interaction code flow success event (which provides tokens). Also Gen 2 does not register this event.

glenfannin-okta: Also the flakiness is caused by the success event/tokens triggering before this erroneous event. So for example, sometimes the tokens success event will happen first then this afterRender cancel event, other times it is reversed. This is why avoiding triggering this event fixes the flakiness.

glenfannin-okta: Tests are passing with the above change, I was mostly concerned about interaction code flow where it could render view elements but I saw the code for that implementation and view elements will get added to `uischema` so we are safe. 🙏

shawyu-okta (approved):

PR #3318, opened 2023-07-26
OKTA-615821 : feat : removing default consent icon when logo not provided in response

glenfannin-okta:
## Description:

The purpose of this PR is to remove the default settings/gear icon when the logo is not provided by the org admin for the application. 

**Without Logo:**

https://github.com/okta/okta-signin-widget/assets/97472729/226e0cf5-bf24-4b4d-b71b-33c6feb547ee

**With Logo:** 

https://github.com/okta/okta-signin-widget/assets/97472729/fd9da5f6-166a-4f54-a5ad-79cb5a925b2f

## PR Checklist

### Issue:

- [OKTA-615821](https://oktainc.atlassian.net/browse/OKTA-615821)


src/v3/src/util/settingsUtils.ts
```diff
@@ -134,7 +134,7 @@ export const getCustomHelpLinks = (widgetProps: WidgetProps): CustomLink[] => {
 
 export const getFactorPageCustomLink = (widgetProps: WidgetProps): Omit<CustomLink, 'target'> | undefined => {
   const { factorPage } = widgetProps.helpLinks || {};
-  return factorPage;
+  return typeof factorPage !== 'undefined' && 'href' in factorPage ? factorPage : undefined;
```
glenfannin-okta, line 137: This was an issue found while testing, prior to the fix, it added an empty div to the UI because we were not properly checking if the `factorPage` property was provided.

src/v3/src/util/idxUtils.ts
```diff
@@ -56,7 +56,8 @@ export const getUserInfo = (transaction: IdxTransaction): UserInfo => {
 };
 
 export const getAppInfo = (transaction: IdxTransaction): AppInfo => {
-  const { context: { app } } = transaction;
+  // @ts-expect-error OKTA-598868 app is missing from rawIdxState type
+  const { rawIdxState: { app } } = transaction;
```
glenfannin-okta, line 60: Had to use app value from raw Idx state due to missing data from context.

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
```diff
@@ -119,12 +105,11 @@ const ConsentHeader: FunctionComponent = () => {
             textAlign="center"
           >
             <Typography
-              component="h1"
```
glenfannin-okta, line 122: a11y issue; since admin consent view has a group heading, (which is h3), all heading elements should ascend in order without skipping.

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
```diff
@@ -119,12 +105,11 @@ const ConsentHeader: FunctionComponent = () => {
...
+              component="h2"
               variant="h3"
               className="no-translate"
             >
               {appName}
-              {' '}
```
glenfannin-okta, line 127: unnecessary space since we are using flexbox to align elements.

src/v3/src/transformer/layout/consent/transformAdminConsent.ts
```diff
@@ -90,6 +90,7 @@ export const transformAdminConsent: IdxStepTransformer = ({ transaction, formBag
         data[fieldName] = true;
         return {
           type: 'Field',
+          key: fieldName,
```
glenfannin-okta, line 93: key is needed to avoid JS errors for duplicate keys.

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
```diff
@@ -85,7 +71,7 @@ const ConsentHeader: FunctionComponent = () => {
         display="flex"
         justifyContent="center"
       >
-        {typeof href !== 'undefined'
+        {typeof href !== 'undefined' && typeof logoHref !== 'undefined'
```
leemchale-okta, line 74: is the logo always a link?  Can there be a logo with no href?

glenfannin-okta: Yep absolutely! When there is no clientUri (href for the logo), then it will fall to the `else` condition which will just try to generate the logo without an anchor, see: https://github.com/okta/okta-signin-widget/pull/3318/files#diff-ccfe8154ab15451a346edff46636e742d6c8fcfe8089731a928c6421704fc104R87

src/v3/src/util/settingsUtils.ts
leemchale-okta: just curious what was it returning before if `factorPage` prop was not provided?  Was there sometimes a valid `factorPage` object but with no `href` prop?

glenfannin-okta: Prior to this change, it was actually returning an empty object, then, [this](https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/transformer/button/transformFactorPageCustomLink.ts#L35-L50) code which relied on that property only checked if it was set or not (because it expected it to be undefined if the factorPage property wasn't set). I only added the additional check for the href to ensure we have a valid target to redirect them to. Based on the docs, the href is required so there shouldn't be a scenario where it isn't provided but added this check to be safe (because the config options are exposed to external customers for use in embedded flows too).

src/v3/test/integration/admin-consent.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    const { container, findByRole } = await setup({
+      mockResponse: adminConsentResponseWithoutLogo,
+    });
+    const appNameHeading = await findByRole('heading', { level: 2 });
+    const groupHeading = await findByRole('heading', { level: 3 });
+
+    expect(appNameHeading.textContent).toBe('Native client');
+    expect(groupHeading.textContent).toBe('Resource and policies');
+
+    expect(container).toMatchSnapshot();
```
leemchale-okta, line 46: is there an assertion to check if the logo is present or not?

src/v3/test/integration/admin-consent.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+      ...createAuthJsPayloadArgs('POST', 'idp/idx/consent', {
+        consent: true,
+      }, 'application/ion+json; okta-version=1.0.0'),
+    );
+  });
+
+  it('should send correct payload when cancel is clicked', async () => {
+    const { authClient, user, findByText } = await setup({ mockResponse: adminConsentResponse });
+
+    const allowConsentBtn = await findByText('Cancel');
```
leemchale-okta, line 64: nit: is this variable name still appropriate in this cancel context?

src/v3/test/integration/admin-consent.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    await user.click(allowConsentBtn);
+    expect(authClient.options.httpRequestClient).toHaveBeenCalledWith(
+      ...createAuthJsPayloadArgs('POST', 'idp/idx/consent', {
```
leemchale-okta, line 67: should this be` idp/idx/cancel`?  Or does cancel in this scenario have the same path?

src/v3/test/integration/admin-consent.test.tsx
glenfannin-okta: No assertion, I was hoping we could rely on the snapshot to ensure it does not regress. I can add an assertion though to be extra sure.

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
```diff
@@ -106,7 +92,7 @@ const ConsentHeader: FunctionComponent = () => {
   const getHeaderContent = () => {
     if ([IDX_STEP.CONSENT_ADMIN, IDX_STEP.CONSENT_ENDUSER].includes(stepName)) {
       // @ts-expect-error OKTA-598777 authentication missing from IdxContext interface
-      const { context: { authentication: { value: { issuer } } = {} } } = idxTransaction;
+      const { rawIdxState: { authentication: { value: { issuer } } = {} } } = idxTransaction;
```
leemchale-okta, line 95: was `context` also missing data here?

src/v3/test/integration/admin-consent.test.tsx
leemchale-okta: okay yeah I think relying on the snapshot is fine, but if you could add an assertion to be extra sure that would be good

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
leemchale-okta: ahh okay got it I missed that part, thank you!

src/v3/src/util/settingsUtils.ts
leemchale-okta: okay got it, awesome nice catch

src/v3/test/integration/admin-consent.test.tsx
glenfannin-okta: No, the label says `Cancel` but the request still sends it to the `consent` endpoint (which is defined by the idx response)

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
glenfannin-okta: Yes it was, not sure why, but it was missing client URI data.

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
alexdahl-okta: While we're here.... could we make the visual presentation of the app name Heading 6?

(But not violate the semantic meaning for a11y.)

leemchale-okta (approved):

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
glenfannin-okta: Done! 
image

PR #3319, opened 2023-07-27
OKTA-631012 Re-trigger consistent polling if the ad hoc pollin doesn't return a different remediation

yannongli-okta:
## Description:
- Previously we did https://github.com/okta/okta-signin-widget/pull/3274 to reduce the 400 errors
- It worked well, but there is a bug that this ad-hoc polling will only be called once.
- If it returns a different remediation (login success or other factor challenge), view will be refreshed and we can continue normally.
- If it returns same device-challenge-poll form, there will not be following poll call which will cause the user to be stuck on the same page.
- It was discovered on Android, mainly due to Android uses HTTPS loopback and is taking more time for server to process the transaction.
- Other platforms have this issue as well, but loopback works pretty fast so it was not discovered.
- The fix is, once the ad hoc polling is finished, we restart a polling which will call /poll after a few seconds
- This is low risk since it's controlled by server side flag which is behind a CCS kill swtich.
- It was enabled on trex, demo and preview, and not on production. So we can continue to monitor its impact.
- Tests can be added later, but not mandatory if everything works fine.

## PR Checklist

### Issue:

- [OKTA-631012](https://oktainc.atlassian.net/browse/OKTA-631012)

### Reviewers:
@YangChen-okta @SohamKolhatkar-okta 

### Screenshot/Video:
Before fix: https://okta.box.com/s/cpkd671comp7e2h1e29mu1ct2iueeuhe
After fix: https://okta.box.com/s/3uvsjvpp6vcfgn7o6dultsyzaepgt04z

### DIFF
--- a/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
+++ b/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
@@ -108,6 +108,10 @@ const Body = BaseFormWithPolling.extend({
               foundPort = true;
               if (deviceChallenge.enhancedPollingEnabled !== false) {
                 this.stopPolling();
+                this.trigger('save', this.model);
+                // if this poll doesn't return a different remediation, start polling again
+                this.startPolling();
+                return;
               }
               // once the OV challenge succeeds,
               // triggers another polling right away without waiting for the next ongoing polling to be triggered


src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
```diff
@@ -108,6 +108,10 @@ const Body = BaseFormWithPolling.extend({
               foundPort = true;
               if (deviceChallenge.enhancedPollingEnabled !== false) {
                 this.stopPolling();
+                this.trigger('save', this.model);
+                // if this poll doesn't return a different remediation, start polling again
+                this.startPolling();
```
YangChen-okta, line 113: is the purpose of stopPolling->startPolling to avoid two poll at about the same time?

shawyu-okta: @yannongli-okta can you confirm?

shawyu-okta: Also, do we know if `this.trigger('save', this.model)` is synchronous? I'm guessing not.

I'm asking because we cannot assume that just because `this.trigger('save', this.model)` has executed that the response has already been returned and processed, so perhaps restarting polling is not appropriate yet. A simple example would be if the network speed is slow and the response just hadn't been returned yet, we would restart polling prematurely.

shawyu-okta: @yannongli-okta could this cause an increase in the number of "improper poll" requests if the response is very slow, which may cause users to see _more_ errors in the widget? Before this change, we would simply wait for a slow response to complete, but now we start the polling timer again.

yannongli-okta: @shawyu-okta the issue is, before this change it would only do one polling which is triggered by this.trigger('save', this.model);. if it's returning the same remediation as before (device-challenge-poll), UI will not refresh and there will be no further polling, hence user is stuck there. this is a worse case in general

yannongli-okta:
Future action items:
1. Create a Jira with all tech details
2. Start a thread in widget channel and propose the Jira
3. Join the widge office hour to discuss with widget team for an ideal fix

PR #3325, opened 2023-08-01
[G3] Migrates to webpack5, exports compatibility bundles

shawyu-okta:
## PR Checklist

### Issue:

- [OKTA-621535](https://oktainc.atlassian.net/browse/OKTA-621535)


src/v3/src/playground.ts
lesterchoi-okta, line 1: Consolidated into `playground/main.ts`, right? 👍

src/v3/package.json
```diff
@@ -115,21 +115,31 @@
...
     "msw": "^1.1.1",
+    "nodemon": "^3.0.1",
     "postcss": "^8.4.13",
+    "postcss-loader": "^7.3.3",
+    "postcss-preset-env": "^8.4.1",
     "postcss-syntax": "^0.36.2",
     "preact-render-to-string": "^5.1.19",
     "preact-svg-loader": "^0.2.1",
     "properties": "^1.2.1",
+    "sass-loader": "^13.2.2",
```
lesterchoi-okta, line 126: why do we need `sass-loader`?

polyfill/modern.js
```diff
@@ -0,0 +1,11 @@
...
+
+// include all core-js/stable
+// webpack @babel/preset-env with useBuiltIns: 'entry' | 'usage'
+// will replace this with only the polyfills required at build-time.
+require('core-js/stable');
+
+// crypto is needed for PKCE
+require('fast-text-encoding'); // TextEncoder
+require('webcrypto-shim'); // crypto.subtle
+require('create-html-document-polyfill');
```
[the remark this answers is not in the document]
shawyu-okta: We use a package called `html-react-parser` in `v3/src/hooks/useHtmlContentParser.tsx` which has sub-dependency that uses `document.implementation.createHtmlDocument` in a non-IE11 compatible way.

We can avoid this polyfill by using an old version of `html-react-parser` that uses an older sub-dependency that has IE11 compatibility code in it.

Both of these options are suboptimal:
1. The polyfill appears to be unmaintained, although this might be OK since it's simple and perhaps does not need any changing
2. An old version of `html-react-parser` would also be unmaintained and we would not be able to upgrade it for any security reasons stemming from it or subdependencies of it

One last thought I had was to have npm/yarn apply a patch on the package on postinstall since the issue can be solved with a 1 to 2 line change but I don't think that's a great solution either.

src/v3/package.json
shawyu-okta: Oh, we don't, I'll remove. I had it in there for a while for the playground entry point because I thought we needed to build a *.scss file for the styles on that page but then later realized any styles needed are baked into the SIW styles.

src/v3/package.json
```diff
@@ -115,21 +115,31 @@
...
     "sirv-cli": "^2.0.2",
+    "style-loader": "^3.3.3",
     "stylelint": "^14.16.1",
     "stylelint-use-logical-spec": "^5.0.0",
     "svgo": "^2.3.0",
+    "svg-inline-loader": "^0.8.2",
```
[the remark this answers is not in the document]
shawyu-okta: Not sure, I copied this in so I can investigate `raw-loader`. Is there something specific about `svg-inline-loader` that you are concerned about?

src/v3/src/playground.ts
shawyu-okta: Yeah exactly. I removed (or intended to remove) all of the playground code that was added for Vite.

src/v3/src/util/theme.ts
```diff
@@ -100,9 +100,11 @@ export const mapMuiThemeFromBrand = (
   // TODO: OKTA-517723 temporary override until odyssey-react-mui theme borderRadius value is fixed
   odysseyTheme.shape.borderRadius = 4;
 
-  const odysseyThemeCopy = cloneDeep(odysseyTheme);
   odysseyTheme.direction = languageDirection;
 
+  // Do not modify `odysseyTheme` after this line, it will have no effect on the resulting theme
```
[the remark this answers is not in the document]
shawyu-okta: There was no "bug" per-se but it would have resulted in a bug if we were using a version of Odyssey with components that read the `theme.direction` value.

src/v3/package.json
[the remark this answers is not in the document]
shuowu-okta: Seems like raw-loader has been deprecated for webpack v5

> DEPREACTED for v5: please consider migrating to [asset modules](https://v4.webpack.js.org/guides/asset-modules/).

src/v3/webpack.dev.config.js
```diff
@@ -0,0 +1,131 @@
...
+        filename: 'css/okta-sign-in.css',
+      }),
+      new webpack.DefinePlugin({
+        DEBUG: true,
+      }),
+    ],
+    devServer: {
+      host: HOST,
+      watchFiles: [...STATIC_DIRS],
+      static: [...STATIC_DIRS],
```
lesterchoi-okta: nit: spreading seems redundant. any reason?

src/v3/package.json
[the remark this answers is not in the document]
shuowu-okta: also seems like we can cleanup some remaining loaders from vite

src/v3/webpack.release.config.ts
```diff
@@ -0,0 +1,95 @@
...
+    mode: 'production',
+    devtool: 'source-map',
+    entry: {
+      default: {
+        import: [
+          // polyfill must appear first in entry point array
+          resolve(__dirname, '../..', './polyfill/modern.js'),
+          resolve(__dirname, 'src/index.ts'),
+        ],
+        filename: 'js/okta-sign-in.min.js',
```
shuowu-okta, line 39: should `next` be included in the filename? or any other script handles that logic?

src/v3/webpack.release.config.ts
```diff
@@ -0,0 +1,95 @@
...
+      }),
+      // OKTA-429162: webpack-bundle-analyzer does not report bundled modules stats after
+      // upgrade to webpack@5
+      new BundleAnalyzerPlugin({
+        openAnalyzer: false,
+        reportFilename: 'okta-sign-in.analyzer.next.html',
+        analyzerMode: 'static',
+        defaultSizes: 'stat',
+      }),
+      new webpack.DefinePlugin({
```
shuowu-okta, line 75: nit: move to common config then conditionally assign value based on webpack `mode`

src/v3/webpack.release.config.ts
```diff
@@ -0,0 +1,95 @@
...
+          resolve(__dirname, 'src/index.ts'),
+        ],
+        filename: 'js/okta-sign-in.min.js',
+        library: {
+          name: 'OktaSignIn',
+          type: 'umd',
+          export: 'default',
+        },
+      },
+      noPolyfill: {
```
shuowu-okta, line 46: maybe also add a polyfill entry? just to provide some flexibility for devs

src/v3/webpack.release.config.ts
```diff
@@ -0,0 +1,95 @@
...
+      },
+    },
+    resolve: {
+      alias: {
+        duo_web_sdk: 'duo_web_sdk',
+      }
+    },
+    plugins: [
+      new MiniCssExtractPlugin({
+        filename: 'css/okta-sign-in.min.css',
```
shuowu-okta, line 65: nit: add `next`?

src/v3/webpack.dev.config.ts
```diff
@@ -0,0 +1,136 @@
...
+    },
+    resolve: {
+      alias: {
+        '@okta/duo': `${PLAYGROUND}/mocks/spec-duo/duo-mock.js`,
+        duo_web_sdk: resolve(__dirname, 'src/__mocks__/duo_web_sdk'),
+      }
+    },
+    plugins: [
+      new MiniCssExtractPlugin({
+        filename: 'css/okta-sign-in.css',
```
shuowu-okta, line 93: nit: add `next` in filename?

src/v3/webpack.dev.config.ts
```diff
@@ -0,0 +1,136 @@
...
+      path: `${PLAYGROUND}/target`,
+    },
+    resolve: {
+      alias: {
+        '@okta/duo': `${PLAYGROUND}/mocks/spec-duo/duo-mock.js`,
+        duo_web_sdk: resolve(__dirname, 'src/__mocks__/duo_web_sdk'),
+      }
+    },
+    plugins: [
+      new MiniCssExtractPlugin({
```
shuowu-okta, line 92: move to common?

src/v3/webpack.dev.config.ts
```diff
@@ -0,0 +1,136 @@
...
+
+const devConfig: Configuration = merge(
+  baseConfig,
+  {
+    mode: 'development',
+    devtool: 'source-map',
+    entry: {
+      playground: {
+        import: [
+          `${PLAYGROUND}/main.ts`,
```
shuowu-okta, line 64: maybe consider add a minimal playground for IE11, so we don't need to bother with playground IE11 compat issue (not important)

src/v3/package.json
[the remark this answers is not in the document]
shawyu-okta: I moved to using asset modules in webpack5 and it works well. The only thing that the svg-inline-loader was doing for us (at least as configured) was stripping the `width` and `height` properties in the `svg` tag out for us, but I don't think it makes a difference in the rendered output in browsers.

src/v3/src/util/theme.ts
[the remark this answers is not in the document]
shawyu-okta: I was originally using this branch to test some Odyssey changes too but pivoted it to only handle the JS changes. Thought it would be good to let this in still since it needs to be fixed as it will result in real bugs in the future if forgotten

src/v3/webpack.dev.config.js
shawyu-okta: I copied this from the gen2 webpack configs. I don't know if there's a real reason but I agree it shouldn't matter.

src/v3/webpack.dev.config.ts
shawyu-okta: For this one we actually need to keep it this way for playground usage since the playground always loads the files as `okta-sign-in.css` and `okta-sign-in.js`

src/v3/webpack.release.config.ts
shawyu-okta: Good point, I'll update

src/v3/webpack.dev.config.ts
shawyu-okta: Since the options passed to the plugin constructor are different (different filename) it's much easier to keep them individually in separate files and it makes it clearer which instance of the plugin will be in the result. This is why we never actually use the `webpack.common.config.ts` file as an input to `webpack-cli` and also why I define the export type as `Partial<Configuration>` rather than `Configuration`.

src/v3/webpack.release.config.ts
shawyu-okta: Yeah good call, I'll update this along with the CSS filename for the release config

src/v3/webpack.dev.config.ts
shawyu-okta: The playground actually works in IE11 now since we are using this same webpack config with the playground entrypoint, so the same babel config that makes the widget bundle IE11 compatible are also applied to the playground now. This was a big change that webpack5 made to support multiple entry points much better so we can use this consolidated configuration to build the playground and the widget at the same time.

src/v3/webpack.release.config.ts
shawyu-okta: oh yeah, I have a TODO item that may need to go in another ticket to generate the polyfill for the g3 widget. Right now we load both entrypoints for the "full bundle" (with polyfill AND widget code) which is great, but since it's using `@babel/preset-env`, the resulting polyfill included in the bundle changes depending on the code within the bundle.

I'm not sure yet how to get webpack to also code split that polyfill into a separate file during the build process but I'm assuming it is possible with some careful code splitting plugin configuration.

src/v3/webpack.dev.config.js
```diff
@@ -0,0 +1,131 @@
...
+    resolve: {
+      alias: {
+        '@okta/duo': `${PLAYGROUND}/mocks/spec-duo/duo-mock.js`,
+        duo_web_sdk: resolve(__dirname, 'src/__mocks__/duo_web_sdk'),
+      }
+    },
+    plugins: [
+      new MiniCssExtractPlugin({
+        filename: 'css/okta-sign-in.css',
+      }),
```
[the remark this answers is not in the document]
shawyu-okta: We don't _need_ to extract the CSS for dev mode but since the playground already has an import for the CSS it's nice to avoid the 404. We also may soon be able to do away with any CSS output altogether since the only CSS output currently comes from some of our CSS modules definitions which can probably be moved into `Box sx={}` or Odyssey theme options for components. At that point, all the CSS will be injected by `emotion-js` and webpack won't need to emit any CSS

src/v3/webpack.release.config.ts
shawyu-okta: Partially related to this comment: https://github.com/okta/okta-signin-widget/pull/3325/files#r1283712828

I found it a lot easier to understand the configs if there aren't multiple places where the config can vary based on some variable, so duplicating this for `dev` and `release` simplifies reading it since you only need to mentally merge the two config objects rather than also needing to think about the `mode` setting to resolve what the resulting config will be.

src/v3/package.json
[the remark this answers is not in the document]
shawyu-okta: Ok, we needed to move to `asset/inline` as the asset module setting. I just learned this too, but FYI `asset/inline` makes webpack generate a `data` URI with base64 encoding so it can be inlined as a string to `src=""`. In this case, the SVGs are included as `img src={Icon}` so simply inlining the raw SVG tag fails. In gen2, it looks like we emit all the asset files and link them as relative URIs but since we have all SVG images in gen3 it probably makes sense to inline everything. This does add to the bundle size a bit so we should pay attention here and consider changing this in the future.

src/v3/webpack.common.config.js
```diff
@@ -0,0 +1,147 @@
...
+      v1: resolve(__dirname, '../v1'),
+      v2: resolve(__dirname, '../v2'),
+
+      // react -> preact alias
+      react: 'preact/compat',
+      'react-dom/test-utils': 'preact/test-utils',
+      'react-dom': 'preact/compat',
+      'react/jsx-runtime': 'preact/jsx-runtime',
+
+      // @mui -> @mui/legacy
```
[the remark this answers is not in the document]
shawyu-okta: Will do

src/v3/webpack.common.config.js
```diff
@@ -0,0 +1,147 @@
...
+            '/node_modules/parse-ms',
+            '/node_modules/@sindresorhus/to-milliseconds',
+            '/node_modules/@okta/odyssey-react-mui',
+            '/node_modules/@mui',
+            '/node_modules/@okta/okta-auth-js',
+            '/node_modules/p-cancelable',
+          ].some(filePathContains);
+          const shallBeExcluded = [
+            '/node_modules/',
+            'packages/@okta/qtip2',
```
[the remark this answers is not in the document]
shawyu-okta: I need to investigate these. These were copied from another webpack config but I will check it out as a part of the cleanups and nits I find now that tests run well and the widget bundle looks right.

src/v3/webpack.common.config.js
```diff
@@ -0,0 +1,147 @@
...
+        test: /\.[jt]sx?$/,
+        exclude(filePath) {
+          const filePathContains = (f) => filePath.indexOf(f) > 0;
+          const npmRequiresTransform = [
+            '/node_modules/parse-ms',
+            '/node_modules/@sindresorhus/to-milliseconds',
+            '/node_modules/@okta/odyssey-react-mui',
+            '/node_modules/@mui',
+            '/node_modules/@okta/okta-auth-js',
+            '/node_modules/p-cancelable',
```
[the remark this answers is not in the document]
shawyu-okta: Same as further up about the `qtip2` dep.

> I need to investigate these. These were copied from another webpack config but I will check it out as a part of the cleanups and nits I find now that tests run well and the widget bundle looks right.

scripts/buildtools/commands/prepack.js
```diff
@@ -20,6 +20,11 @@ exports.handler = async () => {
     'src'
   ], `${BUILD_DIR}`);
 
+  // remove node_modules/ from package in case it gets copied in previous step
+  shell.rm('-Rf', [
+    `${BUILD_DIR}/src/v3/node_modules/`,
+  ]);
```
shawyu-okta, line 26: I couldn't find a way to prevent this directory from being copied in the previous `shell.cp` step but this is the next best thing. It's just a bit of a waste of time to copy the `node_modules/` directory just to delete it (which also takes a bit of time).

scripts/buildtools/commands/verify-package.js
```diff
@@ -3,8 +3,8 @@ const { readFileSync } = require('fs');
 
 const KB = 1024;
 const MB = 1024 * 1024;
-const EXPECTED_PACKAGE_SIZE = 49 * MB;
-const EXPECTED_PACKAGE_FILES = 20000;
+const EXPECTED_PACKAGE_SIZE = 23.5 * MB;
+const EXPECTED_PACKAGE_FILES = 3760;
```
shawyu-okta, line 7: Deleting `node_modules/` in the "prepack" step (above) brought our size and file count way back down to Earth. It used to be closer to this range before the v3 code was brought into the output bundle: https://github.com/okta/okta-signin-widget/blob/eed162d9a1d060e9864bd06b132dfab92f3273fc/scripts/buildtools/commands/verify-package.js#L6-L7

playground/main.ts
```diff
@@ -159,4 +165,12 @@ if (typeof URL !== 'undefined') {
     render = false;
   }
 }
-render && renderPlaygroundWidget(window.additionalOptions ?? {});
+
+// set up msw
+getWorker()
```
glenfannin-okta, line 170: Quick question on this, I recall (a very long time ago 😅 ) there was an odd issue when we used both MSW with the Dyson mock server, where the same request would be triggered multiple times. Is this no longer an issue?

src/v3/public/mocks/labels/json/country_foo.json
lesterchoi-okta, line 1: @glenfannin-okta wasn't this used for a test? Will it be a problem not to have the "foo" l10ns?

glenfannin-okta: This was required with vite configuration, but I believe switching back to webpack resolved the issue which required us to add it to the v3 public folder. I see the tests passing (BYOL_spec.js) so I am pretty confident that this was only needed with vite.

lesterchoi-okta (approved):

playground/main.ts
shawyu-okta: I guess not, I don't see that happening when running the testcafe suits

PR #3327, opened 2023-08-10
g3: Support `hide` and `show` methods

denysoblohin-okta:
## Description:

Support `widget.hide()` and `widget.show()` in Gen3

Tip:
- Using `EventEmitter` pattern
- Using custom `useOnce` [hook](https://github.com/okta/okta-signin-widget/pull/3327/files#diff-d9fe8ef09b73767f05c2fc2b3edc767f2e9db1e1232ae515581b457f66196b0aR373-R379) (using `useRef` [under the hood](https://github.com/okta/okta-signin-widget/pull/3327/files#diff-d41bd1a554514d5099edfcbd0046ebaf5f6617e90b839490b874de77880393bbR15-R22), instead of `useEffect`) to add 'hide' event listener before first render and allow developer to call `widget.hide()` right after `widget.renderEl()` 

## PR Checklist

### Issue:

- [OKTA-594739](https://oktainc.atlassian.net/browse/OKTA-594739)


src/v3/src/components/Widget/index.tsx
```diff
@@ -361,6 +366,17 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [responseError]);
 
+  // listen to 'hide' event
+  const onHide = useCallback((hideValue: boolean) => {
+    setHide(hideValue);
+  }, []);
```
glenfannin-okta, line 372: nit: 
since this controls whether it is shown or hidden, I think it would make more sense to refer to the callback as toggling visibility: 
```suggestion
  const toggleVisibility = useCallback((hideValue: boolean) => {
    setHide(hideValue);
  }, []);
```

src/v3/src/components/Widget/style.module.css
```diff
@@ -10,7 +10,7 @@
   visibility: hidden;
 }
 :global(.hide) {
-  display: none;
+  display: none !important;
```
glenfannin-okta, line 13: instead of adding `!important` here, could you define a module level `.hide` css class in `src/v3/src/components/AuthContainer/style.module.css`?

Unless there is some sort of benefit to doing it this way vs the module level way? I only suggested this because I am not sure of the ramifications of changing this css since the classes here are mostly used by Okta hosted JSP.

test/testcafe/spec/OktaSignIn_spec.js
```diff
@@ -0,0 +1,58 @@
...
+const renderHiddenWidget = ClientFunction(({v3}) => {
+  // functions `renderPlaygroundWidget` and `getWidgetInstance` are defined in playground/main.js
+  window.renderPlaygroundWidget();
+  if (v3) {
+    window.getWidgetInstance().hide();
+  } else {
+    window.getWidgetInstance().on('ready', () => {
+      window.getWidgetInstance().hide();
+    });
+  }
```
glenfannin-okta, line 29: Can you add a comment with this ticket number: https://oktainc.atlassian.net/browse/OKTA-637702 here to update this test to remove the conditional statement. Currently G3 does not trigger the ready event but it should, so this conditional statement should be removed once we add that implementation.

src/v3/src/components/AuthContainer/AuthContainer.tsx
```diff
@@ -18,9 +18,9 @@ import { FunctionComponent, h } from 'preact';
 import { useWidgetContext } from '../../contexts';
 import style from './style.module.css';
 
-const AuthContainer: FunctionComponent = ({ children }) => {
+const AuthContainer: FunctionComponent<{ hide?: boolean }> = ({ children, hide }) => {
```
glenfannin-okta, line 21: nit: since `hide` is initialized with a default value and always passed to the component, you can make this non-optional
```suggestion
const AuthContainer: FunctionComponent<{ hide: boolean }> = ({ children, hide }) => {
```

glenfannin-okta (approved):

PR #3328, opened 2023-08-11
OKTA-637641 Fix FastPass polling issue behind CCS kill switch

yannongli-okta:
## Description:

- We made a temp fix for the polling issue but still sees bad request error https://github.com/okta/okta-signin-widget/pull/3319
- This time we come with the permanent fix: `startPolling(1)` will set the poll interval as 1ms which means a polling will be triggered almost immediately, and there is only one polling queue. Following polling calls will use the refresh interval that is returned from server, like 2s or 4s.
- There will still be a very small chance that, when we call `stopPolling`, there is already a triggered polling that is pending response. but this frequency will be much smaller than the original issue and hence this is a best effort fix.

## PR Checklist

### Issue:

- [OKTA-637641](https://oktainc.atlassian.net/browse/OKTA-637641)

### Reviewers:
@lesterchoi-okta @shawyu-okta

### Screenshot/Video:
- Last time, the end to end test didn't catch the bad request so i refactored the test.
- **before fix**: old code will fail on the refactored test https://okta.box.com/s/4pfzrb99dghavexwyt6sbn50z7f41f9f
- **after fix**: new code will succeed on the refactored test https://okta.box.com/s/tvjzs3l0yontmayooeuqxknmryl6pdye
- **end to end test flow in local env**: https://okta.box.com/s/z5icn8m9p3xjkwmydvsmprnvhp32549g
- All these details were discussed with the widget team and dedicated ui engineers

TBA

### DIFF
--- a/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
+++ b/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
@@ -107,10 +107,16 @@ const Body = BaseFormWithPolling.extend({
             .done(() => {
               foundPort = true;
               if (deviceChallenge.enhancedPollingEnabled !== false) {
+                // this way we can gurantee that
+                // 1. the polling is triggered right away (1ms interval)
+                // 2. Only one polling queue
+                // 3. follwoing polling will continue with refresh interval from previous polling response
+                // NOTE: technically, there could still be concurrency issue where when we called stopPolling,
+                // there is already a polling triggered and hasn't completed yet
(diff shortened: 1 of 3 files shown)


lesterchoi-okta (approved):
Confirmed that `test/testcafe/spec/ChallengeOktaVerifyFastPassView_spec.js` "in loopback server, no redundant polling if server returns enhancedPollingEnabled as true" fails consistently on previous implementation and passes consistently on this fix version

PR #3339, opened 2023-08-22
OKTA-615788 : Default LTR Input direction fix

leemchale-okta:
## Description:

This PR fixes the issue where inputs that expect Latin/English(LTR) characters are inheriting the RTL language direction when a RTL language is set.  This is fixed by setting inputs that only expect numeric/Latin characters to LTR only.

## PR Checklist

### Issue:

- [OKTA-615788](https://oktainc.atlassian.net/browse/OKTA-615788)

### Screenshot/Video:
Screenshot 2023-08-22 at 9 49 17 PM
Screenshot 2023-08-22 at 9 50 05 PM

### DIFF
--- a/src/v3/src/components/InputPassword/InputPassword.tsx
+++ b/src/v3/src/components/InputPassword/InputPassword.tsx
@@ -51,7 +51,12 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
   const value = useValue(uischema);
   const { loading } = useWidgetContext();
   const {
-    translations = [], focus, parserOptions, noTranslate, showAsterisk,
+    translations = [],
+    focus,
+    parserOptions,
+    noTranslate,
+    showAsterisk,
+    dir,
   } = uischema;
   const {
     attributes,
@@ -141,6 +146,7 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
           ...attributes,
         }}
         className={noTranslate ? 'no-translate' : undefined}
+        dir={dir}
         endAdornment={(
           <InputAdornment position="end">
             <Tooltip title={showPassword ? getTranslation(translations, 'hide') : getTranslation(translations, 'show')}>
--- a/src/v3/src/components/PhoneAuthenticator/PhoneAuthenticator.tsx
+++ b/src/v3/src/components/PhoneAuthenticator/PhoneAuthenticator.tsx
@@ -48,7 +48,11 @@ const PhoneAuthenticator: UISchemaElementComponent<UISchemaElementComponentWithV
   describedByIds,
 }) => {
   const {
-    data, dataSchemaRef, loading, widgetProps,
+    data,
+    dataSchemaRef,
+    loading,
+    widgetProps,
+    languageDirection,
   } = useWidgetContext();
   const {
     translations = [],
@@ -124,12 +128,19 @@ const PhoneAuthenticator: UISchemaElementComponent<UISchemaElementComponentWithV
   const renderExtension = () => (
     showExtension && (
       <Box width={0.25}>
-        <InputLabel htmlFor="phoneExtension">{extensionLabel}</InputLabel>
+        <InputLabel
+          htmlFor="phoneExtension"
+          // label should remain in rtl format if rtl language is set
+          dir={languageDirection}
+        >
+          {extensionLabel}
+        </InputLabel>
         <InputBase
           value={extension}
           type="text"
           name="extension"
           id="phoneExtension"
+          dir="ltr"
           disabled={loading}
           onChange={(e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
             setExtension(e.currentTarget.value);
@@ -215,6 +226,7 @@ const PhoneAuthenticator: UISchemaElementComponent<UISchemaElementComponentWithV
       <Box
         display="flex"
         flexWrap="wrap"
+        dir="ltr"
       >
         <Box
           width={showExtension ? 0.7 : 1}
@@ -224,6 +236,8 @@ const PhoneAuthenticator: UISchemaElementComponent<UISchemaElementComponentWithV
             htmlFor={fieldName}
             // To prevent asterisk from shifting far right
             sx={{ justifyContent: showAsterisk ? 'flex-start' : undefined }}
+            // label should remain in rtl format if rtl language is set
+            dir={languageDirection}
           >
             {mainLabel}
             {showAsterisk && (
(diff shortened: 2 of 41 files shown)


src/v3/src/components/PhoneAuthenticator/PhoneAuthenticator.tsx
```diff
@@ -215,6 +226,7 @@ const PhoneAuthenticator: UISchemaElementComponent<UISchemaElementComponentWithV
       <Box
         display="flex"
         flexWrap="wrap"
+        dir="ltr"
```
shawyu-okta, line 229: If this is set, do we still need to set the `dir` in the nested elements?

src/v3/src/transformer/uischema/setLtrFields.ts
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import {
+  FieldElement,
+  TransformStepFn,
+} from '../../types';
+import { traverseLayout } from '../util';
+
+export const setLtrFields: TransformStepFn = (formbag) => {
```
shawyu-okta, line 19: Curious - is there a benefit of using this transformer instead of explicitly setting `field.dir` like in `transformRSAAuthenticator`? It seems like we already have one exception that requires us to set it in the transformer.

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -141,6 +146,7 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
           ...attributes,
         }}
         className={noTranslate ? 'no-translate' : undefined}
+        dir={dir}
```
shawyu-okta, line 149: No strong opinion here but it seems like we know password is always LTR so we could consider hardcoding it in here instead of allowing it to be configurable

src/v3/src/transformer/uischema/setLtrFields.ts
```diff
@@ -0,0 +1,39 @@
...
+  traverseLayout({
+    layout: formbag.uischema,
+    predicate: (el) => {
+      if (el.type === 'Field') {
+        // fields with these names should always be LTR
+        const ltrOnlyFieldNames = ['credentials.passcode', 'identifier', 'credentials.newPassword', 'confirmPassword'];
+        const fieldElement = (el as FieldElement);
+        if (ltrOnlyFieldNames.includes(fieldElement.options.inputMeta.name)) {
+          return true;
+        }
```
shawyu-okta, line 29: nit: a bit safer if we
```ts
return ltrOnlyFieldNames.includes(fieldElement.options.inputMeta.name)
```

Could also restructure the logic to initially check is `el.type !== 'Field'` and short-circuit `return false;` early

src/v3/src/transformer/uischema/setLtrFields.ts
```diff
@@ -0,0 +1,39 @@
...
+import { traverseLayout } from '../util';
+
+export const setLtrFields: TransformStepFn = (formbag) => {
+  traverseLayout({
+    layout: formbag.uischema,
+    predicate: (el) => {
+      if (el.type === 'Field') {
+        // fields with these names should always be LTR
+        const ltrOnlyFieldNames = ['credentials.passcode', 'identifier', 'credentials.newPassword', 'confirmPassword'];
+        const fieldElement = (el as FieldElement);
```
shawyu-okta, line 26: I'm surprised that discriminated unions didn't automatically narrow this type once we've checked `el.type`

src/v3/src/transformer/uischema/setLtrFields.ts
```diff
@@ -0,0 +1,39 @@
...
+        if (ltrOnlyFieldNames.includes(fieldElement.options.inputMeta.name)) {
+          return true;
+        }
+      }
+      return false;
+    },
+    callback: (el) => {
+      const fieldElement = (el as FieldElement);
```
shawyu-okta, line 34: I know this is 'safe' because we are counting on the predicate to only run this callback if it was `type === 'Field'` but probably best to still run a check inside this function in case we ever change the logic above. Ideally we could rework this transformer function to have some kind generic with type-narrowing to guarantee the incoming type but I'm not clever enough to think of how to do that right now :P

src/v3/src/transformer/uischema/setLtrFields.ts
leemchale-okta: I think the benefit of this is that we can easily add and remove common fields that we want to be ltr only without having to go to each individual transformer and add/remove `field.dir` setting.  (eg. if later we start accepting RTL identifiers, we can simply remove the 'identifier' name from this transformer).  Also when making new transformers, it can be easy to miss adding the `field.dir` for common fields like passcode entry.  But yeah there will be some exceptions for specific non-common use cases where we would need to set the field direction directly in the transformer.  I do think its benefical to have both options

src/v3/src/components/InputPassword/InputPassword.tsx
leemchale-okta: actually this InputPassword is also used for the security question answer, which needs to inherit RTL

src/v3/src/transformer/uischema/setLtrFields.ts
leemchale-okta: haha yeah that does make sense to check in the callback aswell, will add that in 👍

src/v3/src/transformer/uischema/setLtrFields.ts
shawyu-okta: Ok, that's a good point. My only concern (not blocking) is that these transforms for uischema run later than the layout transforms (where custom transforms occur) so it's not possible to override the direction behavior for a particular view if the field name matches any of these below. That's probably fine and even desired, but just noting that.

src/v3/test/integration/authenticator-enroll-data-phone.test.tsx
```diff
@@ -79,6 +79,21 @@ describe('authenticator-enroll-data-phone', () => {
...
   });
 
+  it('phone number and extension fields should be ltr even when a rtl language is set', async () => {
+    const {
+      findByTestId, user, findByLabelText,
+    } = await setup({ mockResponse, widgetOptions: { language: 'ar' } });
+
+    const phoneNumberEle = await findByTestId('authenticator.phoneNumber') as HTMLInputElement;
+
+    const methodType = await findByLabelText(/Voice call/);
```
shawyu-okta, line 89: I see this label is still in English -- do we not actually load other languages in the tests?

shawyu-okta (approved):
I think it would be nice to fix the issue with discriminated unions in the schema type file. I also think we might be able to remove `jest.unmock('@okta/courage');` in `test/integration/util/setup.tsx` since it's taken care of in `jest.setup.js`

Let's discuss before you merge these changes, but otherwise lgtm

PR #3342, opened 2023-08-25
OKTA-639061 Factor nickname changes for SIW (Phone Enrollments)

diptidesai-okta:
## Description:
- End users can give a nickname to their enrollments of OV, Webauthn and Phone.
- This PR, include nickname (from `POST /identify` response) along with display name for PHONE enrollments that have been nicknamed
- OV and WebAuthn currently have indirect references to the authenticator object (instead of enrollment) in remediation response. So, we will need additional backend changes for OV and WebAuthn to render nickname.
- Changes behind FF `FACTOR_NICKNAME_SIW`

## PR Checklist

### Issue:

- [OKTA-639061](https://oktainc.atlassian.net/browse/OKTA-639061)

### Reviewers:
@AnirudhMani-okta 
@glenfannin-okta 
@jaredperreault-okta 

### Screenshot/Video:

https://github.com/okta/okta-signin-widget/assets/88336366/9390b118-d7b3-4485-ace4-f38dc11aefc3


jaredperreault-okta:
Can you add some testcafe tests for this change as well? The unit test you added is good, but we need to test the views as well

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -39,9 +39,17 @@ const Body = BaseForm.extend(
       const extraCssClasses =
         this.model.get('phoneNumber') !== loc('oie.phone.alternate.title', 'login') ?
           'strong no-translate' : '';
+      let nicknameText = '';
+      if (this.model.get('nickname')) {
+        nicknameText = '(' + this.model.escape('nickname') + ')';
+      }
+      if (nicknameText.length > 20) {
+        nicknameText = nicknameText.substring(0, 20) + '...';
+      }
```
glenfannin-okta, line 48: It is not really recommended to append to text string this way, specifically because of localization, in other languages, the way this is ordered may not make sense, but, we usually try to keep one translation key, that can be interpolated with parameter values. 

In this specific scenario, I would actually recommend creating new translation keys to handle the expected output. 

```
oie.phone.verify.sms.sendText.nickname = Send a code via SMS to <$1>{0}</$1> ({1})
oie.phone.verify.call.sendText.nickname = Send a code via voice call to <$1>{0}</$1> ({1})
```

Then to use it, you will have to just check for the existence of the `nickname` value, and if set and if phoneNumber is set, use one of the above keys and pass the values as parameters to the `loc` function. 

We can assist with this if you need help. However, if you go with this approach, you will have to create the translation keys and get them translated. 

cc: @jaredperreault-okta @shawyu-okta

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
```diff
@@ -57,12 +57,19 @@ const Body = BaseForm.extend(Object.assign(
         : loc('mfa.calling', 'login');
       const enterCodeText = loc('oie.phone.verify.enterCodeText', 'login');
       const carrierChargesText = loc('oie.phone.carrier.charges', 'login');
+      let nicknameText = '';
+      if (this.model.get('nickname')) {
+        nicknameText = '(' + this.model.escape('nickname') + ')';
+      }
+      if (nicknameText.length > 20) {
+        nicknameText = nicknameText.substring(0, 20) + '...';
+      }
```
glenfannin-okta, line 66: Similar to [this](https://github.com/okta/okta-signin-widget/pull/3342/files#diff-613a9a51a022846714b15e19706db7688f92eab855376f7e72de20bd5cd8c431R42-R48) comment,will want to create separate translation keys for this as well.

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
```diff
@@ -57,12 +57,19 @@ const Body = BaseForm.extend(Object.assign(
...
 
       const strongClass = this.model.get('phoneNumber') !== loc('oie.phone.alternate.title', 'login') ?
         'strong no-translate' : '';
       // Courage doesn't support HTML, hence creating a subtitle here.
       this.add(`<div class="okta-form-subtitle" data-se="o-form-explain">
-        ${sendText}&nbsp;<span class='${strongClass}'>${this.model.escape('phoneNumber')}.</span>
+        ${sendText}<br/><span class='${strongClass}'>${this.model.escape('phoneNumber')}<br/>.${nicknameText}</span>
```
glenfannin-okta, line 72: Any reason why we are adding a line break here but not in the other view? Is this necessary? Has anyone from Design reviewed this change?

cc: @alexdahl-okta

src/v2/view-builder/utils/AuthenticatorUtil.js
```diff
@@ -52,6 +56,7 @@ const getAuthenticatorData = function(authenticator, isVerifyAuthenticator) {
 
   case AUTHENTICATOR_KEY.PHONE:
     Object.assign(authenticatorData, {
+      nickname: nicknameText
```
glenfannin-okta, line 59:
```suggestion
      nickname: nicknameText,
```

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
diptidesai-okta: Line breaks were added based on informal feedback from Alex and Joel. However, I don't see them show up at all even after re-deploying my server. Is there any override that could be preventing it? Thanks

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
diptidesai-okta: The nicknames should not be translated because those would not be Okta constants and will instead of be user defined strings.

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
```diff
@@ -87,6 +94,7 @@ export default BaseAuthenticatorView.extend({
   createModelClass() {
     const relatesToObject = this.options.currentViewState.relatesTo;
     const { methods, profile } = relatesToObject?.value || {};
+    const nickname = relatesToObject?.value || '';
```
glenfannin-okta, line 97:
```suggestion
```

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
```diff
@@ -87,6 +94,7 @@ export default BaseAuthenticatorView.extend({
   createModelClass() {
     const relatesToObject = this.options.currentViewState.relatesTo;
     const { methods, profile } = relatesToObject?.value || {};
```
glenfannin-okta, line 96:
```suggestion
    const { methods, profile, nickname } = relatesToObject?.value || {};
```

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -77,6 +85,7 @@ export default BaseAuthenticatorView.extend({
     const { options: methods } = _.find(uiSchema, schema => schema.name === 'authenticator.methodType');
     const relatesToObject = this.options.currentViewState.relatesTo;
     const { profile } = relatesToObject?.value || {};
+    const nickname = relatesToObject?.value || '';
```
glenfannin-okta, line 88:
```suggestion
```

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -77,6 +85,7 @@ export default BaseAuthenticatorView.extend({
     const { options: methods } = _.find(uiSchema, schema => schema.name === 'authenticator.methodType');
     const relatesToObject = this.options.currentViewState.relatesTo;
     const { profile } = relatesToObject?.value || {};
```
glenfannin-okta, line 87:
```suggestion
    const { profile, nickname } = relatesToObject?.value || {};
```

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -91,6 +100,10 @@ export default BaseAuthenticatorView.extend({
         'value': profile?.phoneNumber ? profile.phoneNumber : loc('oie.phone.alternate.title', 'login'),
         'type': 'string',
       },
+      nickname: {
+        'value': nickname?.nickname,
+        'type': 'string',
+      },
```
glenfannin-okta, line 106:
```suggestion
      nickname: {
        'value': nickname,
        'type': 'string',
      },
```

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
```diff
@@ -96,6 +104,10 @@ export default BaseAuthenticatorView.extend({
       phoneNumber: {
         'value': profile?.phoneNumber ? profile.phoneNumber : loc('oie.phone.alternate.title', 'login'),
         'type': 'string',
+      },
+      nickname: {
+        'value': nickname ? nickname : '',
```
glenfannin-okta, line 109:
```suggestion
        'value': nickname,
```

glenfannin-okta (commented):
Additionally, you will want to add a spec test (including adding the mock file where the feature flag is enabled - so it returns the nickname value)

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
```diff
@@ -185,6 +202,32 @@ test
...
+test
+  .requestHooks(smsPrimaryMockWithNickname)('SMS primary mode - shows nickname when present in API response', async t => {
+    const challengePhonePageObject = await setup(t);
+    await checkA11y(t);
+    await checkConsoleMessages({
+      controller: null,
+      formName: 'authenticator-verification-data',
+      authenticatorKey: 'phone_number',
+      methodType: 'sms',
+    });
```
glenfannin-okta, line 214:
```suggestion
    await checkConsoleMessages({
      controller: 'mfa-verify-passcode',
      formName: 'challenge-authenticator',
      authenticatorKey: 'phone_number',
      methodType: 'sms',
    });
```

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
diptidesai-okta: Reviewed feedback from Joel and Alex with Glen. Resolving this thread as removing the line breaks seems to address it.
cc: @glenfannin-okta

playground/mocks/data/idp/idx/authenticator-verification-phone-sms-nickname.json
```diff
@@ -0,0 +1,312 @@
...
+        "id": "autwa6eD9o02iBbtv0g3",
+        "methods": [
+          {
+            "methodType": "email"
+          }
+        ]
+      },
+      {
+        "displayName": "Okta Phone",
+        "nickname": "phone-nn",
```
glenfannin-okta, line 264: This doesn't seem to be in the correct location, if you look at [this](https://github.com/okta/okta-signin-widget/pull/3342/files#diff-9396975b565d396e86d9d556a15dc53d090538a44eec282d8b7c86c0df673560R14-R16) it is pointing to the currentAuthenticatorEnrollment, which does not have a nickname set. So the test will fail because the nickname wont be available in the view.

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -39,9 +39,17 @@ const Body = BaseForm.extend(
       const extraCssClasses =
         this.model.get('phoneNumber') !== loc('oie.phone.alternate.title', 'login') ?
           'strong no-translate' : '';
+      let nicknameText = this.model.get('nickname') || '';
+      if (nicknameText.length > 20) {
+        nicknameText = nicknameText.substring(0, 20) + '...';
+      }
+      if (nicknameText !== '') {
+        nicknameText = ' (' + nicknameText + ')';
```
glenfannin-okta, line 47: Should add a space before the `(` here, otherwise the phonenumber and nickname will run together.

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -39,9 +39,17 @@ const Body = BaseForm.extend(
...
+      }
       // Courage doesn't support HTML, hence creating a subtitle here.
       this.add(`<div class="okta-form-subtitle" data-se="o-form-explain">${sendText}
-        <span ${ extraCssClasses ? 'class="' + extraCssClasses + '"' : ''}>${this.model.escape('phoneNumber')}</span>
+        <span ${ extraCssClasses ? 'class="' + extraCssClasses + '"' : ''}>
+        ${this.model.escape('phoneNumber')}${nicknameText}</span>
```
glenfannin-okta, line 52: `nicknameText` should only be added if `this.model.get('phoneNumber') !== loc('oie.phone.alternate.title', 'login')`, meaning the phoneNumber is like `XXX-XXX-5455` instead of the text `phone`.

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
```diff
@@ -57,12 +57,19 @@ const Body = BaseForm.extend(Object.assign(
         : loc('mfa.calling', 'login');
       const enterCodeText = loc('oie.phone.verify.enterCodeText', 'login');
       const carrierChargesText = loc('oie.phone.carrier.charges', 'login');
+      let nicknameText = this.model.get('nickname') || '';
+      if (nicknameText.length > 20) {
+        nicknameText = nicknameText.substring(0, 20) + '...';
+      }
+      if (nicknameText !== '') {
+        nicknameText = '(' + nicknameText + ')';
```
glenfannin-okta, line 65: Should add a space before the ( here, otherwise the phonenumber and nickname will run together.

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
```diff
@@ -57,12 +57,19 @@ const Body = BaseForm.extend(Object.assign(
...
+      }
 
       const strongClass = this.model.get('phoneNumber') !== loc('oie.phone.alternate.title', 'login') ?
         'strong no-translate' : '';
       // Courage doesn't support HTML, hence creating a subtitle here.
       this.add(`<div class="okta-form-subtitle" data-se="o-form-explain">
-        ${sendText}&nbsp;<span class='${strongClass}'>${this.model.escape('phoneNumber')}.</span>
+        ${sendText}&nbsp;<span class='${strongClass}'>${this.model.escape('phoneNumber')}.${nicknameText}</span>
```
glenfannin-okta, line 72: nicknameText should only be added if this.model.get('phoneNumber') !== loc('oie.phone.alternate.title', 'login'), meaning the phoneNumber is like XXX-XXX-5455 instead of the text phone.

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
```diff
@@ -96,6 +101,18 @@ const smsPrimaryMockEmptyProfile = RequestMock()
...
+  .onRequestTo('http://localhost:3000/idp/idx/introspect')
+  .respond(smsVerificationWithNickname)
+  .onRequestTo('http://localhost:3000/idp/idx/challenge')
+  .respond(smsVerificationWithNickname)
+  .onRequestTo('http://localhost:3000/idp/idx/challenge/resend')
+  .respond(smsVerificationWithNickname)
+  .onRequestTo('http://localhost:3000/idp/idx/challenge/answer')
+  .respond(success)
+  .onRequestTo(/^http:\/\/localhost:3000\/app\/UserHome.*/)
+  .respond(oktaDashboardContent);
```
glenfannin-okta, line 114: The way this mock is setup should not be necessary for your test case since you are only asserting on the subtitle text in the view and not actually performing any interactions with the input field or buttons.

```suggestion
const smsPrimaryMockWithNickname = RequestMock()
  .onRequestTo('http://localhost:3000/idp/idx/introspect')
  .respond(smsVerificationNickname);
```

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
```diff
@@ -29,6 +30,10 @@ const smsVerificationEmptyProfile = JSON.parse(JSON.stringify(smsVerificationNoP
 // add empty profile to test
 smsVerificationEmptyProfile.remediation.value[0].profile = {};
 
+const smsVerificationWithNickname = JSON.parse(JSON.stringify(smsVerificationNickname));
+// add empty profile to test
+smsVerificationWithNickname.remediation.value[0].profile = {};
+
```
glenfannin-okta, line 36: This is not needed, can delete.
```suggestion
```

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
```diff
@@ -185,6 +202,32 @@ test
...
+test
+  .requestHooks(smsPrimaryMockWithNickname)('SMS primary mode - shows nickname when present in API response', async t => {
+    const challengePhonePageObject = await setup(t);
+    await checkA11y(t);
+    await checkConsoleMessages({
+      controller: 'mfa-verify-passcode',
+      formName: 'challenge-authenticator',
+      authenticatorKey: 'phone_number',
+      methodType: 'sms',
+    });
```
glenfannin-okta, line 214: This is not necessary. 
```suggestion
```

playground/mocks/data/idp/idx/authenticator-verification-phone-sms-nickname.json
diptidesai-okta: yes good catch. Thanks

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
diptidesai-okta: Addressed. Thanks

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
diptidesai-okta: Addressed.

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
diptidesai-okta: Addressed. Thanks

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
diptidesai-okta: Addressed. Thanks

playground/mocks/data/idp/idx/authenticator-verification-phone-sms-nickname.json
diptidesai-okta: Moved to `authenticatorEnrollments`. Thanks

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
alexdahl-okta: @diptidesai-okta Can we see screenshots with the line break working? I want to make sure the nickname line wraps normally.

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -39,9 +39,17 @@ const Body = BaseForm.extend(
       const extraCssClasses =
         this.model.get('phoneNumber') !== loc('oie.phone.alternate.title', 'login') ?
           'strong no-translate' : '';
+      let nicknameText = '';
+      if (this.model.get('nickname')) {
+        nicknameText = '(' + this.model.escape('nickname') + ')';
+      }
+      if (nicknameText.length > 20) {
+        nicknameText = nicknameText.substring(0, 20) + '...';
```
alexdahl-okta, line 43: Why would we truncate the string itself based on a character count? (If we do need to, why is there also a function for that in `AuthenticatorUtil.js`?)

I'd expect handling the string in CSS would be more resilient.

For example, in the authenticator list, it looks like `VERYLONGNICKNAMEVERY...` is interfering with layout inappropriately:
Screenshot 2023-08-30 at 2 40 35 PM 2

An approach with `text-overflow: ellipsis;` should be able to preserve layout, and let layout determine when truncation is necessary.

And on the verify view, the closing `)` appears inadvertently truncated:
Screenshot 2023-08-30 at 2 41 20 PM 2

Perhaps the verify view doesn't need any truncation; its style could make that determination.

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
```diff
@@ -185,6 +198,26 @@ test
...
+    const pageTitle = challengePhonePageObject.getFormTitle();
+    const pageSubtitle = challengePhonePageObject.getFormSubtitle();
+    const primaryButtonText = challengePhonePageObject.getSaveButtonLabel();
+    const secondaryButtonText = challengePhonePageObject.getSecondaryLinkText('Receive a voice call instead');
+    await t.expect(pageTitle).contains('Verify with your phone');
+    await t.expect(pageSubtitle).contains('Send a code via SMS to');
+    await t.expect(pageSubtitle).contains('+1 XXX-XXX-2342');
+    await t.expect(pageSubtitle).contains(' (phone-nn)');
+    await t.expect(primaryButtonText).contains('Receive a code via SMS');
+    await t.expect(secondaryButtonText).contains('Receive a voice call instead');
```
glenfannin-okta, line 215: These assertions are not needed as they are for a different view.
```suggestion
```

playground/mocks/data/idp/idx/authenticator-verification-select-authenticator.json
```diff
@@ -464,6 +464,7 @@
         "key": "phone_number",
         "id":"pae5ykx4eIvfslkO90g4",
         "displayName":"Phone Number",
+        "nickname":"ph-nn",
```
glenfannin-okta, line 467: Preferably, we should leave the existing mock files as is, and create a new one with this nickname value. This is to showcase support for when the nickname feature is not enabled but the view still works.

diptidesai-okta: Got it. Added new mock and test.

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
diptidesai-okta: Addressed.

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
diptidesai-okta: Thanks. Removed the manual ellipsis/truncation logic. This is how it looks
image
image

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -251,6 +258,137 @@ test.requestHooks(mockChallengePassword)('should load select authenticator list'
   await t.expect(await selectFactorPage.signoutLinkExists()).eql(true);
 });
 
+test.requestHooks(mockChallengeWithNickname)('should load select authenticator list with nicknames if available', async t => {
```
glenfannin-okta, line 261:
```suggestion
test.meta('v3', false).requestHooks(mockChallengeWithNickname)('should load select authenticator list with nicknames if available', async t => {
```

playground/mocks/data/idp/idx/authenticator-verification-phone-sms-nickname.json
```diff
@@ -0,0 +1,315 @@
...
...VH1jNVdV7Ou3nc9hzE9zcV2ueRZoSYq6h7WZ0KjLjhQTqRKe3g9XGR_DL05a26czfGJL-R1omHxUuFDlYSCd_jtKZE4TseUXQ5qfUzk0X9syqqVA36HEiJ5_CWsQ2SpxIgqir4DztivsQE1ubJfVrvPZw_-6GU391NkXE8e_B64X6ZSqgUgm5H4LuXj9Eu2VKG5z4KLKWieuHDK0Hjc1RRGuxFqcYhzySnTtK41Za4abTqVM-YYxokb9ah2yxauzpveIO0rDqjGo3kqZdVDV81kdvIzqgclooUb5X1o_fdmPjB_IUkVDkIHN9sVu1F0HkqQWp7xDmhbokAB7kVa6WIiax08UtTUfPvVK4aOLP_R61VfzImlTZ6otI6yC1yZAETauuADIz1YkB9bcBOOpJeqrom3keGeVwCmU-i0dJYGi1NLEPDm0px0OhVW77Cpi1C0aBX295Dju4-luOM3EuaBad-U5FMwDAO00qi0c3XDiaY0Bgkrmcqeo1yIl5p4cktsBs2ArzpzpEzJAmpLaIbuaihov8uy1K6e9y0Ko2VcOR930K-Cfnr-0QxX_WLtf8pB9LYENR2Gx65BxVMi9E4M1_rLG5UwBq4ATEH72a1kkTeXvDgMpwMmhIEz_nPuCcDJVZ8UDQoZU8C_T3yzOrlUHc8x2qN-FEWG3mLBGw7SAdxs3pH4eZnnQQVN4Ch4Tuc_wqMrophaxMLtWAt7Sa0L0zBpJUJsCKg_pTKH6gkjL_Tco6eIY7o78JoX7M_GSFIdF9qR3FLj9rXJrvP0qeWMrbhwgy6DhJg_bBWeKNQdclh1uRLLmbKpdCZH--lqLubZGJtd8JlkEKWW_PnDYFdDGzMwMz83owDHvw8GQ9KpqoPME-Ty0SEoMwKKdeRHCwUf90ycp7D5Jc6yHY17tr8axpjy_yCNhyI7ChtK0cnyJVsQPwkCH5q5fiiQrwX-nEny95PCpwnJzA85qGyYJusFlBO-HH8Ojcp_jTJNr3GeiBwfPQtOwQZ7AkUL7tLsls7q6MiLzCju_MGQlsFSzvV2N7wyFkCoLf9zyk7vKpU3QxfZFA6jqR-f9FOEXrWHr_0Orsd1s2k-RZ8_-SicRI5VieZj7YApi6v0LHqklqkzluwOfH-wSI4tf0En-A-rrR-1yQei7JCa0s8YXpp0mVOE.EC3UkB3_VQpE75Jet9kM4g",
```
jaredperreault-okta, line 2: This can be truncated, it doesn't need to be a real value

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
glenfannin-okta: Want to disable this test for v3 since there is a ticket to add this logic into V3.

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
```diff
@@ -57,12 +57,21 @@ const Body = BaseForm.extend(Object.assign(
...
+      }
 
       const strongClass = this.model.get('phoneNumber') !== loc('oie.phone.alternate.title', 'login') ?
         'strong no-translate' : '';
       // Courage doesn't support HTML, hence creating a subtitle here.
       this.add(`<div class="okta-form-subtitle" data-se="o-form-explain">
-        ${sendText}&nbsp;<span class='${strongClass}'>${this.model.escape('phoneNumber')}.</span>
+        ${sendText}&nbsp;<span class='${strongClass}'>${this.model.escape('phoneNumber')}</span>
+        <span ${ extraNicknameCssClasses ? 'class="' + extraNicknameCssClasses + '"' : ''}>
+        ${nicknameText}.</span>
```
glenfannin-okta, line 74: You should probably only add the `nickName` span element if the nickname value exists, (it also depends on if the phone number is provided). I dont think it will cause a problem visually to have an empty span here but that may be an a11y issue. 

Can you create a variable above `this.add(...` logic, and set the span with the nickName only if nickName is set. If it is not set, you can leave it as an empty string.

Then, in this area, you can just append the variable, so regardless on if it exists or not, it will render properly.

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -36,12 +36,20 @@ const Body = BaseForm.extend(
...
+        nicknameText = ' (' + nicknameText + ')';
+        extraNicknameCssClasses = 'strong no-translate authenticator-enrollment-nickname';
+      }
       // Courage doesn't support HTML, hence creating a subtitle here.
       this.add(`<div class="okta-form-subtitle" data-se="o-form-explain">${sendText}
-        <span ${ extraCssClasses ? 'class="' + extraCssClasses + '"' : ''}>${this.model.escape('phoneNumber')}</span>
+        <span ${ extraCssClasses ? 'class="' + extraCssClasses + '"' : ''}>
+        ${this.model.escape('phoneNumber')}</span>
+        <span ${ extraNicknameCssClasses ? 'class="' + extraNicknameCssClasses + '"' : ''}>
+        ${nicknameText}.</span>
```
glenfannin-okta, line 52: Same as the comment left in ChallengeAuthenticatorPhoneView. 

You should probably only add the nickName span element if the nickname value exists, (it also depends on if the phone number is provided). I dont think it will cause a problem visually to have an empty span here but that may be an a11y issue.

Can you create a variable above this.add(... logic, and set the span with the nickName only if nickName is set. If it is not set, you can leave it as an empty string.

Then, in this area, you can just append the variable, so regardless on if it exists or not, it will render properly.

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
diptidesai-okta: Addressed.

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
diptidesai-okta: Addressed.

src/v2/view-builder/utils/AuthenticatorUtil.js
```diff
@@ -24,11 +24,15 @@ const getButtonDataSeAttr = function(authenticator) {
...
-/* eslint complexity: [0, 0], max-statements: [2, 24] */
+/* eslint complexity: [0, 0], max-statements: [2, 25] */
 const getAuthenticatorData = function(authenticator, isVerifyAuthenticator) {
   const authenticatorKey = authenticator.authenticatorKey;
   const key = _.isString(authenticatorKey) ? authenticatorKey.toLowerCase() : '';
   let authenticatorData = {};
+  let nicknameText = isVerifyAuthenticator ? authenticator.relatesTo?.nickname : undefined;
+  if (nicknameText && nicknameText.length > 20) {
+    nicknameText = nicknameText.substring(0, 20) + '...';
+  }
```
glenfannin-okta, line 35: This should no longer be necessary right? Since you created the css in `assets/sass/v2/_select-authenticator.scss`

src/v2/view-builder/views/phone/ChallengeAuthenticatorPhoneView.js
```diff
@@ -57,12 +57,23 @@ const Body = BaseForm.extend(Object.assign(
...
+      let nicknameText = isPhoneNumberAvailable ? this.model.get('nickname') : '';
+      let extraNicknameCssClasses = '';
+      if (nicknameText !== '') {
+        nicknameText = ' (' + nicknameText + ')';
+        extraNicknameCssClasses = 'strong no-translate authenticator-enrollment-nickname';
+      }
 
+      const nicknameTemplate = nicknameText ? `<span ${ extraNicknameCssClasses ? 'class="' + 
+      extraNicknameCssClasses + '"' : ''}>
+      ${nicknameText}.</span>` : '';
```
glenfannin-okta, line 70:
```suggestion
      ${nicknameText}.</span>` : '<span class="no-translate">.</span>';
```

test/testcafe/spec/SelectAuthenticatorForVerification_spec.js
```diff
@@ -251,6 +258,137 @@ test.requestHooks(mockChallengePassword)('should load select authenticator list'
   await t.expect(await selectFactorPage.signoutLinkExists()).eql(true);
 });
 
+test.meta('v3', false).requestHooks(mockChallengeWithNickname)('should load select authenticator list with nicknames if available', async t => {
```
glenfannin-okta, line 261:
```suggestion
// OKTA-642786 - awaiting Gen3 implementation
test.meta('v3', false).requestHooks(mockChallengeWithNickname)('should load select authenticator list with nicknames if available', async t => {
```

test/testcafe/spec/ChallengeAuthenticatorPhone_spec.js
```diff
@@ -185,6 +198,21 @@ test
     await t.expect(challengePhonePageObject.getSignoutLinkText()).eql('Back to sign in');
   });
 
+test.meta('v3', false)
```
glenfannin-okta, line 201:
```suggestion
// OKTA-642786 - awaiting Gen3 implementation
test.meta('v3', false)
```

src/v2/view-builder/views/phone/ChallengeAuthenticatorDataPhoneView.js
```diff
@@ -36,12 +36,21 @@ const Body = BaseForm.extend(
...
+      const extraCssClasses = isPhoneNumberAvailable ? 'strong no-translate' : '';
+      let nicknameText = isPhoneNumberAvailable ? this.model.get('nickname') : '';
+      let extraNicknameCssClasses = '';
+      if (nicknameText !== '') {
+        nicknameText = ' (' + nicknameText + ')';
+        extraNicknameCssClasses = 'strong no-translate authenticator-enrollment-nickname';
+      }
+      const nicknameTemplate = nicknameText ? `<span ${ extraNicknameCssClasses ? 
+        'class="' + extraNicknameCssClasses + '"' : ''}>
+      ${nicknameText}.</span>` : '';
```
glenfannin-okta, line 49:
```suggestion
      ${nicknameText}.</span>` : '<span class="no-translate">.</span>';
```

src/v2/view-builder/utils/AuthenticatorUtil.js
diptidesai-okta: correct - will remove

diptidesai-okta: Removed.

jaredperreault-okta (approved):

PR #3356, opened 2023-09-05
Use @prefresh/webpack for HMR

denysoblohin-okta:
## Description:

Fix HMR when running `yarn workspace v3 dev`

- Added [`optimization: { runtimeChunk: 'single' }`](https://github.com/okta/okta-signin-widget/pull/3356/files#diff-bde2c72a896868e214e0d91cc48c9fcbc19dac3aae65607a15c6692fe37ef48dR162) to fix error `Cannot set properties of undefined (setting './src/components/Button/Button.tsx')`, see [issue](https://github.com/webpack/webpack-dev-server/issues/2792)
- Added `@prefresh/webpack` for hot module reload without page refresh
- Added `@prefresh/babel-plugin` only for webpack dev config.  And  [only for files in `/src/`](https://github.com/okta/okta-signin-widget/pull/3356/files#diff-bde2c72a896868e214e0d91cc48c9fcbc19dac3aae65607a15c6692fe37ef48dR101), otherwise there will be error `Maximum call stack size exceeded` (probably caused by the fact that `@okta/odyssey-react-mui` uses `react` not `preact`, so need to exclude react libs and include only widget src)
- [These changes](https://github.com/okta/okta-signin-widget/pull/3356/files#diff-0c99dcd54f4308c8b962b658ada5f2690761c9c4fd3575e0a684eb6186ad5a5dR159) are needed for correct [component recognition](https://github.com/preactjs/prefresh#recognition) by prefresh. Otherwise you can see strange bugs when modifying files - like "Keep me signed in" will be rendered as Select instead of Checkbox
- [This](https://github.com/okta/okta-signin-widget/pull/3356/files#diff-d9fe8ef09b73767f05c2fc2b3edc767f2e9db1e1232ae515581b457f66196b0aR330-R349) is a hack to prevent agressive widget state reset because of `bootstrap` call -  eg. if you edit file `idxUtil.ts`, any transformer etc. then `Widget/index.tsx` will be reloaded, its effects will be fired including the one that calls `bootstrap`

## PR Checklist

### Issue:

- [OKTA-639419](https://oktainc.atlassian.net/browse/OKTA-639419)


src/v3/src/components/hocs/withFormValidationState.tsx
```diff
@@ -82,6 +82,6 @@ export const withFormValidationState: WrappedFunctionComponent<
     // eslint-disable-next-line react/jsx-props-no-spreading
     return <Component {...combinedProps} />;
   };
-  ParentComponent.displayName = `WithFormValidationState(${getDisplayName(ParentComponent)})`;
+  ParentComponent.displayName = `WithFormValidationState(${getDisplayName(Component)})`;
```
denysoblohin-okta, line 85: Otherwise `displayName` will always equal `WithFormValidationState(ParentComponent)`

lesterchoi-okta (approved):

src/v3/webpack.common.config.ts
```diff
@@ -48,37 +73,24 @@ const baseConfig: Partial<Configuration> = {
...
-                runtime: 'automatic',
-                importSource: 'preact',
-              },
-            ],
-          ],
-        },
+        options: babelOptions,
+      },
+      {
+        test: /\.[jt]sx?$/,
```
shawyu-okta, line 88: Can you add a comment mentioning we need to separate these because of the dev config? Just so future developers don't see this and think they can simplify by combining these.

src/v3/src/components/Widget/index.tsx
```diff
@@ -327,12 +329,33 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
...
+      };
+      const prevDeps = __PREFRESH__.widgetBootstrapDeps as typeof deps;
+      const depsChanged = !prevDeps || Object.keys(prevDeps).some((k) => deps[k] !== prevDeps[k]);
+      __PREFRESH__.widgetBootstrapDeps = deps;
+      if (!depsChanged) {
+        // eslint-disable-next-line no-console
+        console.info('[HMR] Skip Widget bootstrap');
+        return;
+      }
+    }
```
shawyu-okta, line 351: Can you extract this out into a separate file? I'd like to avoid more code in the Widget index file, especially code that is working around a development-time issue.

May even possibly even guard it with a compile-time env variable like NODE_ENV so webpack might totally strip it out in the production bundle.

shawyu-okta (approved):
minor comments

PR #3358, opened 2023-09-05
OKTA-643896 : SIW Gen3 : Fixing Duo authenticator

glenfannin-okta:
## Description:

SIW Gen3 was failing to load the DUO iframe due to a build tooling change we made. This PR fixes this issue. 

## PR Checklist

### Issue:

- [OKTA-643896](https://oktainc.atlassian.net/browse/OKTA-643896)

### Screenshot/Video:

https://github.com/okta/okta-signin-widget/assets/97472729/ddcaf519-ac07-4bd4-bb4e-8d43487c34f7

### DIFF
--- a/src/v3/webpack.common.config.ts
+++ b/src/v3/webpack.common.config.ts
@@ -132,8 +132,6 @@ const baseConfig: Partial<Configuration> = {
       'react-dom': 'preact/compat',
       'react/jsx-runtime': 'preact/jsx-runtime',
 
-      'duo_web_sdk': 'duo_web_sdk',
-
       // @mui -> @mui/legacy
       // use the legacy @mui/* bundles for ie11 support
       // https://mui.com/material-ui/guides/minimizing-bundle-size/#legacy-bundle
--- a/src/v3/webpack.common.config.ts
+++ b/src/v3/webpack.common.config.ts
@@ -27,10 +27,6 @@ const getShortGitHash = () => {
   return hash.slice(0, 7);
 };
 
-const {
-  MOCK_DUO
-} = process.env;
-
 const baseConfig: Partial<Configuration> = {
   mode: 'production',
   devtool: 'source-map',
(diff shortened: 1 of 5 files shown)


src/v3/webpack.dev.config.ts
```diff
@@ -82,13 +82,6 @@ const devConfig: Configuration = merge<Partial<Configuration>>(
     output: {
       path: `${PLAYGROUND}/target`,
     },
-    resolve: {
-      alias: {
-        // mock duo
-        '@okta/duo': resolve(PLAYGROUND, '/mocks/spec-duo/duo-mock.js'),
```
glenfannin-okta, line 88: This is not necessary for Gen3, this is only for Gen2.

src/v3/webpack.dev.config.ts
```diff
@@ -82,13 +82,6 @@ const devConfig: Configuration = merge<Partial<Configuration>>(
...
-        duo_web_sdk: resolve(__dirname, 'src/__mocks__/duo_web_sdk'),
```
glenfannin-okta, line 89: Removed this here because I am now passing an env flag to yarn command when executed for testcafe, this is to allow us to run the real version of DUO when testing locally. 

Another way to do this is to create a dedicated webpack config file specifically for testcafe, with a new yarn command/script, this way, no flag is required, and we can run the standard dev command to run the server locally and have the real DUO integration work. 

Looking for feedback on this approach.

src/v3/webpack.testcafe.config.ts
```diff
@@ -0,0 +1,136 @@
+/*
```
glenfannin-okta, line 1: This is the same as the dev config, the only difference being the duo package is using mocks instead of the real module.

src/v3/webpack.common.config.ts
```diff
@@ -132,6 +132,8 @@ const baseConfig: Partial<Configuration> = {
       'react-dom': 'preact/compat',
       'react/jsx-runtime': 'preact/jsx-runtime',
 
+      'duo_web_sdk': 'duo_web_sdk',
```
glenfannin-okta, line 135: Use the real module by default unless otherwise specified not to.

glenfannin-okta:
This is no longer necessary

PR #3363, opened 2023-09-11
OKTA-641233: SIW Gen 3 : Implement User Code (device code) authentication

trevoring-okta:
## Description:
This PR adds support for the user code / device code authentication flow within SIW Gen 3. 

Some areas of note:
- The v2 version of this flow hyphenates the device code after the 4th character purely for UX/aesthetic purposes (verified with the outbound fed team that the backend can handle hyphenated and unhyphenated codes). Discussed with @alexdahl-okta and decided to reimplement hyphenation using the `addHyphen.ts` utility for parity with v2 despite the manual hyphenation logic being somewhat finicky.
- The v2 version of this flow displays the success and error terminal states using success and error icons (see below). Discussed with Alex again and decided to use `InfoBox` to replicate these success and error states, which required some one-off changes to `transformTerminalMessages.ts` since v3 was defaulting to rendering the terminal states only using titles and descriptions.

#### v2 terminal state w/ icon:
v2-device-code-terminal-states

#### v3 terminal state before infobox change:
v3-device-code-terminal-states-without-infobox

#### v3 terminal state after infobox change:
v3-terminal-state-with-infobox

## PR Checklist

### Issue:

- [OKTA-641233](https://oktainc.atlassian.net/browse/OKTA-641233)
- [Bacon Link](https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&branch=ti-OKTA-641233-user-code-authenticator&page=1&pageSize=6&sha=d700b98344661643fc32450b6125d0a909eb0855&tab=main)

#### Success Flow:
https://github.com/okta/okta-signin-widget/assets/106110543/53481a23-8e70-4cca-b14c-3a47227c9eb4

#### Invalid Code Flow:
https://github.com/okta/okta-signin-widget/assets/106110543/aeea9e3c-62f8-48bc-9374-0584ff6395d6

#### Error Device Not Activated Flow:
https://github.com/okta/okta-signin-widget/assets/106110543/e72ed153-7a84-44fb-8fe8-654295db6efd


src/v3/src/components/InputText/InputText.tsx
```diff
@@ -123,6 +124,11 @@ const InputText: UISchemaElementComponent<UISchemaElementComponentWithValidation
         onBlur={(e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
           handleBlur?.(e?.currentTarget?.value);
         }}
+        onKeyUp={(e: KeyboardEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+          if (hyphenate) {
+            addHyphen(e);
+          }
+        }}
```
shawyu-okta, line 131:
```suggestion
        onKeyUp={hyphenate ? addHyphen : undefined}
```

src/v3/src/types/schema.ts
```diff
@@ -271,6 +271,11 @@ export interface FieldElement extends UISchemaElement {
    * This only applies for profile enrollment view.
    */
   showAsterisk?: boolean;
+  /**
+   * @description Determines if the app should add a hyphen after the field's 4th character
+   * This only applies for device code view.
+  */
+  hyphenate?: boolean;
```
shawyu-okta, line 278: I'm concerned about the number of view-specific properties we're adding to elements. Can we explore making this more generic and then allow the transformer to pass in the very specific function to execute after the key input is pressed.

shawyu-okta: Curious if in this case we can leverage form field validation in `handleChange` and `handleBlue` to do this. If so it would also solve the issue above mentioned in this code comment:

> // NOTE: Hyphenation is triggered by a KeyboardEvent and only occurs upon typing, not pre-filling

@glenfannin-okta are we able to modify the field data from the validator functions right now in a reliable way? If not, is it just a matter of adding another property to the validator function callback? If so, I'm in favor of extending the validator's functionality to also rewrite the current value of the field so long as it isn't terribly complex or causes weird looping issues.

src/v3/src/transformer/terminal/transformTerminalMessages.ts
```diff
@@ -157,6 +157,23 @@ export const transformTerminalMessages: TerminalKeyTransformer = (transaction, f
...
+    displayedMessages.unshift(deviceActivatedAsSuccessMessage);
+  } else if (containsOneOfMessageKeys([TERMINAL_KEY.DEVICE_NOT_ACTIVATED_CONSENT_DENIED,
+    TERMINAL_KEY.DEVICE_NOT_ACTIVATED_INTERNAL_ERROR], displayedMessages)) {
+    // Displays device not activated terminal states titles as an error InfoBox instead of a title
+    const deviceNotActivatedAsErrorMessage: WidgetMessage = {
+      class: 'ERROR',
+      message: loc('device.code.activated.error.title', 'login'),
+      i18n: { key: 'device.code.activated.error.title' },
+    };
+    displayedMessages.unshift(deviceNotActivatedAsErrorMessage);
```
shawyu-okta, line 176: Not in scope but: Oh man, we really need to refactor the terminal message transformations 😅. This is getting to be a real mess.

src/v3/src/types/schema.ts
trevoring-okta: Would making this change to use form field validation provide value to any other flows? 

If not, I feel like the changes required to add this form field validation might not be worth it, especially because other OTP inputs don't use hyphenation, so it might actually be more consistent to have device code match this behavior since it's also essentially an OTP? 

It already seemed a bit odd to me that the device code is hyphenated purely for UX reasons, but other OTPs are not.

glenfannin-okta: @shawyu-okta It may be worth a try, I think a new hook would need to be created, then use that hook within `src/v3/src/components/hocs/withFormValidationState.tsx` and either add it to the `handleChange` function or create a new property (for keyup) and add it to the keyup attribute of the input. 

@trevoring-okta Wanna give this a try, I can assist with it.

test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
```diff
@@ -9,10 +10,16 @@ export default class DeviceCodeActivatePageObject extends BasePageObject {
...
+    }
     return this.form.getElement('.okta-form-subtitle').textContent;
   }
 
-  clickNextButton() {
+  clickNextButton(name) {
+    if (userVariables.v3) {
+      return this.form.clickSaveButton(name);
+    }
     return this.form.clickSaveButton();
```
glenfannin-okta, line 23: I think this would work.
```suggestion
    return this.form.clickSaveButton(name);
```

test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
```diff
@@ -9,10 +10,16 @@ export default class DeviceCodeActivatePageObject extends BasePageObject {
   }
 
   getPageSubtitle() {
+    if (userVariables.v3) {
+      return this.form.getElement('[data-se="o-form-explain"]').textContent;
```
glenfannin-okta, line 14:
```suggestion
      return this.form.getSubtitle();
```

test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
```diff
@@ -52,15 +59,31 @@ export default class DeviceCodeActivatePageObject extends BasePageObject {
...
+    if (userVariables.v3) {
+      return this.getPageSubtitle();
+    }
     return this.form.getTerminalContent();
   }
 
   isTerminalSuccessIconPresent() {
+    if (userVariables.v3) {
+      return this.form.getElement('.infobox-success').exists && this.form.getElement('.MuiAlert-icon').exists;
+    }
```
glenfannin-okta, line 79: To determine if it is the success info box for v3, I would replicate what is being done in this file `test/testcafe/framework/page-objects/components/CalloutObject.js` in this function `isError` but instead of `infobox-error` as the class, it would be `infobox-success`. We want to stay away from asserting on MUI classnames in the case we ever move away from MUI.

test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
```diff
@@ -52,15 +59,31 @@ export default class DeviceCodeActivatePageObject extends BasePageObject {
...
     return this.form.getElement('.device-code-terminal--icon.success-24-green').exists;
   }
 
   isTerminalErrorIconPresent() {
+    if (userVariables.v3) {
+      return this.form.getElement('.infobox-error').exists && this.form.getElement('.MuiAlert-icon').exists;
+    }
```
glenfannin-okta, line 86: Would recommend same as the comment above.

src/v3/src/transformer/terminal/transformTerminalTransaction.test.ts
```diff
@@ -181,13 +181,10 @@ describe('Terminal Transaction Transformer Tests', () => {
...
 
     expect(formBag).toMatchSnapshot();
-    expect(formBag.uischema.elements.length).toBe(2);
-    expect(formBag.uischema.elements[0].type).toBe('Title');
-    expect((formBag.uischema.elements[0] as TitleElement).options?.content)
-      .toBe(TERMINAL_TITLE_KEY[TERMINAL_KEY.DEVICE_NOT_ACTIVATED_CONSENT_DENIED]);
-    expect(formBag.uischema.elements[1].type).toBe('Link');
-    expect((formBag.uischema.elements[1] as LinkElement).options?.label).toBe('oie.try.again');
-    expect((formBag.uischema.elements[1] as LinkElement).options?.href).toBe(mockHref);
+    expect(formBag.uischema.elements.length).toBe(1);
```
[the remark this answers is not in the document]
glenfannin-okta: Oh, maybe that happens in another section of the code?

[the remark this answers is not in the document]
trevoring-okta: Yeah I believe it doesn't reach the logic for InfoBox because `transformTerminalMessages` is mocked at the top of this test

test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
trevoring-okta: Makes sense in terms of not wanting to assert on MUI classnames, I only did that because it doesn't seem like there are any non-MUI selectors that I can target for the icon itself and this function is supposed to look specifically for a success icon. 

Screenshot 2023-09-12 at 4 21 47 PM

That being said, I feel like looking for an `infobox-success` and `infobox-error` classname is sufficient for the v3 version of this function because those components should always come with their respective icons, otherwise the error is bigger than just this spec, do you agree? If so, something along the lines of either `this.form.elementExist('.infobox-success')` or `this.form.getAlertBox().hasClass('infobox-success')` should be sufficient.

alexdahl-okta:
Looking great, thanks!

test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
glenfannin-okta: ^ Yep! Totally agree with the last statement 🙏

src/v3/src/components/InputText/InputText.tsx
```diff
@@ -17,7 +17,6 @@ import {
   InputLabel,
   Typography,
 } from '@okta/odyssey-react-mui';
-import { h } from 'preact';
```
glenfannin-okta, line 20: Was this removed on purpose? I think this is added to any file that contains jsx

trevoring-okta: Oh oops, will add back I think I thought at one point it was giving me a lint error but maybe I was mistaken

PR #3366, opened 2023-09-13
OKTA-642777 : SIW G3 Failure redirect view

leemchale-okta:
## Description:

This PR implements support for the failure redirect terminal view in SIW Gen 3.

## PR Checklist

### Issue:

- [OKTA-642777](https://oktainc.atlassian.net/browse/OKTA-642777)

### Screenshot/Video:
**View in Gen 2:**
Screenshot 2023-09-12 at 10 13 39 PM
**View in Gen 3:**
Screenshot 2023-09-12 at 10 15 19 PM

### DIFF
--- a/src/v3/src/transformer/terminal/transformTerminalTransaction.ts
+++ b/src/v3/src/transformer/terminal/transformTerminalTransaction.ts
@@ -171,6 +171,20 @@ export const transformTerminalTransaction = (
     );
   }
 
+  if (transaction.context?.failure?.href) {
+    // Direct auth clients display the error instead of redirecting
+    // when redirect option is set to 'always' it will override the default behavior
+    const shouldRedirect = isOauth2Enabled(widgetProps) === false || widgetProps.redirect === 'always';
+    if (shouldRedirect) {
+      SessionStorage.removeStateHandle();
+      return redirectTransformer(
+        transaction,
+        transaction.context.failure.href,
+        widgetProps,
+      );
+    }
+  }
+
   const { messages } = transaction;
 
   if (containsMessageKey(TERMINAL_KEY.SESSION_EXPIRED, messages)) {
--- a/test/testcafe/framework/page-objects/PlaygroundErrorPageObject.js
+++ b/test/testcafe/framework/page-objects/PlaygroundErrorPageObject.js
@@ -9,6 +9,6 @@ export default class PlaygroundErrorPageObject extends BasePageObject {
   }
 
   hasTitle() {
-    return !!this.title;
+    return this.title.exists;
   }
 }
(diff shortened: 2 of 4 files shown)


test/testcafe/framework/page-objects/PlaygroundErrorPageObject.js
```diff
@@ -9,6 +9,6 @@ export default class PlaygroundErrorPageObject extends BasePageObject {
   }
 
   hasTitle() {
-    return !!this.title;
```
leemchale-okta, line 12: this was causing the function to always return true, making the spec test always pass.  The spec test has not been working properly until now due to this.  Thats probably why we didn't catch this missing view earlier

src/v3/src/transformer/terminal/transformTerminalTransaction.ts
```diff
@@ -171,6 +171,20 @@ export const transformTerminalTransaction = (
     );
   }
 
+  if (transaction.context?.failure?.href) {
```
shawyu-okta, line 174: Can we add tests to cover this scenario?

test/testcafe/framework/page-objects/PlaygroundErrorPageObject.js
shawyu-okta: Thanks for fixing! I was going to write a comment about the return type of `this.title.exists` being `Promise<boolean>` and that you'd need to `await` it but TIL that testcafe auto `await`s certain Promises: https://testcafe.io/documentation/402702/reference/test-api/testcontroller/expect#:~:text=TestCafe%20awaits%20Promises%20from%20compatible%20asynchronous%20functions

leemchale-okta: ohh yeah good point.   Interesting okay cool good to know!

src/v3/src/transformer/terminal/transformTerminalTransaction.ts
leemchale-okta: would an integration test be necessary?  It would essentially test the same thing as the testcafe test.  Or do you mean in the unit test for `transformTerminalTransaction`?

shawyu-okta: I was thinking of the unit tests for `transformTerminalTransaction`

leemchale-okta: ah yes okay for sure I will add that 👍

test/testcafe/spec/FailureRedirect_spec.js
```diff
@@ -47,7 +48,8 @@ test.requestHooks(userNotAssignedMock)('oauth: will redirect if `redirect === "a
     authParams: {
       pkce: true // required for interaction code flow
     },
-    redirect: 'always'
+    authScheme: 'oauth2',
```
shawyu-okta, line 51: Is this something the gen2 widget didn't need but we do for gen3? Put another way, is this something we should have had set on the gen2 spec anyways?

leemchale-okta: yeah gen2 didn't need this because in gen2 `authScheme` is set to `oauth2` by default.  Where as in gen3 it isn't so it needs to be explicitly set here.  So this will work in both cases

shawyu-okta (approved):

PR #3367, opened 2023-09-13
g3: Support on/off

denysoblohin-okta:
## Description:

Adds support of `on` and `off` methods of `OktaSignIn` for Gen3

Fixes bug for v1/v2: `.off(event, callback)` was not working because of wrapping the callback with error handing to new callback instance.
Fixed with using `WeakMap` (original callback -> wrapped callback)

WeakMap polyfill works in IE11

## PR Checklist

### Issue:

- [OKTA-594740](https://oktainc.atlassian.net/browse/OKTA-594740)

### DIFF
--- a/src/v3/src/OktaSignIn/index.ts
+++ b/src/v3/src/OktaSignIn/index.ts
@@ -56,12 +56,8 @@ export default class OktaSignIn {
   readonly authClient: OktaSignInAPI['authClient'];
 
   /**
-   * Registered event handlers
+   * Event emitter
    */
-  private events: {
-    [key in OktaWidgetEventType]: OktaWidgetEventHandler
-  };
-
   private eventEmitter: EventEmitter;
 
   el: string | null;
@@ -71,12 +67,6 @@ export default class OktaSignIn {
     this.options = options;
     this.el = null;
     this.eventEmitter = new EventEmitter();
-
-    this.events = {
-      ready: () => { },
-      afterRender: () => { },
-      afterError: () => { },
-    };
 
     // if authClient is set, authParams are disregarded
     if (options.authClient) {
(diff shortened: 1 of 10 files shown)


src/v3/src/OktaSignIn/index.ts
```diff
@@ -236,10 +233,34 @@ export default class OktaSignIn {
   getUser(): void { }
 
   on(eventName: OktaWidgetEventType, eventHandler: OktaWidgetEventHandler): void {
-    this.events[eventName] = eventHandler;
+    let registeredEventHandler = eventHandler;
+    if (EVENTS_LIST.includes(eventName)) {
```
leemchale-okta, line 237: should we trap third party errors for all events and not just ones that are included in `EVENTS_LIST`?

denysoblohin-okta: We can. But other events are not public. I tried to keep parity with Gen2:
https://github.com/okta/okta-signin-widget/blob/master/src/widget/OktaSignIn.ts#L192

For handlers of public events (there are only 3 of those: `ready`, `afterRender`, `afterError`) we need to catch errors to prevent  throwing unhandled errors in SIW components (because event callbacks are executed synchronously in React lifecycle method `useEffect`).

There is currently one private event (undocumented, developers should not use, used in SIW code) - `hide`: https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/components/Widget/index.tsx#L402
There are also private events in Gen1/2 (like `route`) that developers could theoretically subscribe on, and if their handlers throw error, it would be uncaught.

leemchale-okta: I see okay yeah that does make sense that with events open to the public this is definitely needed.  Where as with private events we are/should be in control.  I think if we are in parity with gen2 then it should be okay.

@shawyu-okta any opinion on this?

leemchale-okta (approved):

PR #3375, opened 2023-09-19
Supports showPasswordToggleOnSignInPage FF in v3

trevoring-okta:
## Description:
- Supports the `showPasswordToggleOnSignInPage` FF for v3, though FF will eventually get deprecated post-GA for accessibility reasons (see [OKTA-623544](https://oktainc.atlassian.net/browse/OKTA-623544))

## PR Checklist

### Issue:

- [OKTA-594734](https://oktainc.atlassian.net/browse/OKTA-594734)
- [Bacon Link](url)

#### Demo of toggle visibility for different widget configs:
https://github.com/okta/okta-signin-widget/assets/106110543/59c53a58-dec0-41c0-ad83-0e4576ef5f1f


test/testcafe/spec/IdentifyWithPassword_spec.js
```diff
@@ -139,7 +138,6 @@ test.meta('v3', false).requestHooks(identifyRequestLogger, identifyWithPasswordM
   await t.expect(await identityPage.hasShowTogglePasswordIcon()).notOk();
 });
 
-// OKTA-555502 - showPasswordToggleOnSignInPage feature not supported yet
 test.meta('v3', false).requestHooks(identifyRequestLogger, identifyWithPasswordMock)('should not have password toggle if "features.showPasswordToggleOnSignInPage" is false', async t => {
```
leemchale-okta, line 141: this looks like a dupe of the previous test.  Could you change this to test the default case where `features.showPasswordToggleOnSignInPage` is undefined?

leemchale-okta:
`showPasswordToggleOnSignInPage` is a misleading FF name when it affects not just the sign in page but all password inputs lol.  But its getting deprecated anyway so dont think we need to be concerned about that

leemchale-okta (approved):

test/testcafe/spec/IdentifyWithPassword_spec.js
shawyu-okta: This is the test that tests the "dot notation" way of passing the FF which is not supported in gen3. We need to assess if we want to support this (it's not documented) but for the moment we can just skip the test in gen3.

@trevoring-okta can you add a comment with a link to the new Jira and also a comment explaining why we skip it right now?

shawyu-okta (approved):
just the request to add a comment otherwise lgtm

shawyu-okta:
One more thing -- should we move the FF to the 'deprecated' section? cc: @lesterchoi-okta

glenfannin-okta (commented):
cc: @shawyu-okta @lesterchoi-okta 

In reviewing this, I found this in the readme: 

```
Shows eye icon to toggle visibility of the user entered password on the Okta Sign-In page. Password is hidden by default, even when this flag is enabled. Passwords are visible for 30 seconds and then hidden automatically.
```

Specifically on the `Passwords are visible for 30 seconds and then hidden automatically.` part, this feature does not exist in Gen3 right now, but I figured we will want to support this (assuming it was implemented for security reasons). 
Should we create a ticket to get this implemented?

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -148,26 +153,28 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
...
-                    outlineStyle: 'solid',
-                    outlineWidth: '1px',
-                  },
-                }}
-              >
-                {showPassword ? <EyeOffIcon /> : <EyeIcon />}
-              </IconButton>
-            </Tooltip>
-          </InputAdornment>
+          showPasswordToggleOnSignInPage && (
```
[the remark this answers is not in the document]
glenfannin-okta: I only ask this because I see in Gen2 this is only applied to the identify form, but the other flows that use this component in Gen3 will be impacted (Security question, Password reset, etc). But maybe you all discussed this already and deemed this acceptable since we want to deprecate this eventually?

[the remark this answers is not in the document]
shawyu-okta: Thanks for the callout and checking to both of you. If the existing gen2 behavior is the same we can keep this the same in gen3.

test/testcafe/spec/IdentifyWithPassword_spec.js
leemchale-okta: Ohh I see okay got it my mistake.  Adding the comment will help distinguish each test

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -148,26 +160,27 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
...
         dir={dir}
         endAdornment={(
-          <InputAdornment position="end">
-            <Tooltip title={showPassword ? getTranslation(translations, 'hide') : getTranslation(translations, 'show')}>
-              <IconButton
-                aria-label={getTranslation(translations, 'visibilityToggleLabel')}
-                aria-pressed={showPassword}
-                aria-controls={name}
-                onClick={handleClickShowPassword}
-                onMouseDown={handleMouseDownPassword}
```
trevoring-okta, line 158: @shawyu-okta  @glenfannin-okta This `mouseDown` handler was the culprit for the cursor jumping to the front of the input. I spoke to @leemchale-okta because he originally added this handler since it used to be present on the Odyssey password component, but Lee was able to verify that the new Odyssey component doesn't have this handler anymore

test/testcafe/spec/IdentifyWithPassword_spec.js
```diff
@@ -149,6 +149,23 @@ test.meta('v3', false).requestHooks(identifyRequestLogger, identifyWithPasswordM
   await t.expect(await identityPage.hasShowTogglePasswordIcon()).notOk();
 });
 
+test.requestHooks(identifyRequestLogger, identifyWithPasswordMock)('should auto-hide password after 30 seconds', async t => {
```
trevoring-okta, line 152: Would appreciate some eyes on this test since it's my first that I've added with `wait()`s before. I saw other tests use a straight up `t.wait(30500)` for a 30 second timeout, but I thought adding the extra buffer time to the short `timeout` option for the `expect()` call made more sense because it overrides the longer default wait time for `expect()` calls. Otherwise, the default wait time for the assertion counts toward the total wait time and this test could pass even with something like `t.wait(20000)`

shawyu-okta: We might want to revisit this soon and see if doing this in a unit test is the right place. Nothing urgent we need to handle but waiting 30s of wall clock time is costly. Ideally we can use the fake timers functionality in jest to ensure this works but we have to move it to a unit test that renders a view with the password field and does the same thing.

trevoring-okta: Yeah the 30s wall clock was what I wasn't sure about. Probably should've waited to merge until I get a definitive response on this test, apologies. I can create a task ticket to revisit this test

shawyu-okta: no worries, I was just going to add a comment saying I should have reviewed sooner. Create the ticket, add to our triage and let's revisit soon.

PR #3376, opened 2023-09-20
G3: Support before/after hooks API

denysoblohin-okta:
## Description:

Adds support of [hooks](https://github.com/okta/okta-signin-widget#hooks) for Gen 3.
Hooks can be added with config and added in runtime with `.before` and `.after` methods.

Notes:
1. Emitting `ready` and `afterRender` events [moved](https://github.com/okta/okta-signin-widget/pull/3376/files#diff-d9fe8ef09b73767f05c2fc2b3edc767f2e9db1e1232ae515581b457f66196b0aR386) to useEffect on `widgetRendered` change. This is needed to keep the order of `ready`, `afterRender` events and `after` hooks same as in Gen2
2. `OktaSignIn` class owns instance of `WidgetHooks` class and passes it in widget props
3. There is only 1 discrepancy between in Gen2 and Gen3 in terms of events/hooks:
https://github.com/okta/okta-signin-widget/blob/3fff3b30de02ce50b4f9fd8a1f7ff9f7ecf19d45/test/testcafe/spec/OktaSignIn_spec.js#L604-L610
When user submits form which results in error, context of `afterError` event contains `terminal` form for Gen3, not current form as in Gen2. This is also noticeable in UI: Gen3 hides the form and renders only alert, while Gen2 still presents the form after error.

## PR Checklist

- [x] Did you add tests, as appropriate, following our [Automated Test guidelines](hxtps://oktawiki.atlassian.net/wiki/spaces/eng/pages/2676497890/Automated+Testing+in+the+Signin+Widget)?

### Issue:

- [OKTA-594745](https://oktainc.atlassian.net/browse/OKTA-594745)
- [OKTA-508861](https://oktainc.atlassian.net/browse/OKTA-508861)

### DIFF
--- a/src/util/Hooks.js
+++ b/src/util/Hooks.js
@@ -33,9 +33,9 @@ export async function executeHooksAfter(hook) {
 export function mergeHook(hooks, formName, hookToMerge) {
   const existingHook = hooks[formName] = hooks[formName] || { before: [], after: [] };
   if (hookToMerge.before) {
-    existingHook.before = existingHook.before.concat(hookToMerge.before);
+    existingHook.before = (existingHook.before || []).concat(hookToMerge.before);
   }
   if (hookToMerge.after) {
-    existingHook.after = existingHook.after.concat(hookToMerge.after);
+    existingHook.after = (existingHook.after || []).concat(hookToMerge.after);
   }
 }
--- a/src/v3/src/OktaSignIn/index.ts
+++ b/src/v3/src/OktaSignIn/index.ts
@@ -16,6 +16,8 @@ import { h, render } from 'preact';
 import { TinyEmitter as EventEmitter } from 'tiny-emitter';
 
 import {
+  EventContext, EventErrorContext,
+  HookFunction,
   OktaSignInAPI, RenderErrorCallback, RenderResult, RenderSuccessCallback,
 } from '../../../types';
 import { Widget } from '../components/Widget';
@@ -24,14 +26,16 @@ import {
   OktaWidgetEventHandler,
   OktaWidgetEventType,
   RenderOptions,
+  WidgetOptions,
   WidgetProps,
 } from '../types/widget';
+import { WidgetHooks } from '../util/widgetHooks';
 
 const EVENTS_LIST = ['ready', 'afterError', 'afterRender'];
 
 console.debug(`${OKTA_SIW_VERSION}-g${OKTA_SIW_COMMIT_HASH.substring(0, 7)}`);
 
-export default class OktaSignIn {
+export default class OktaSignIn implements OktaSignInAPI {
   /**
    * Version string
    */
@@ -63,6 +67,11 @@ export default class OktaSignIn {
   private eventEmitter: EventEmitter;
 
   /**
+   * Hooks
+   */
+  private widgetHooks: WidgetHooks;
+
+  /**
    * Map original event handler to wrapped one
    */
   private eventCallbackMap: WeakMap<OktaWidgetEventHandler, OktaWidgetEventHandler>;
@@ -75,6 +84,7 @@ export default class OktaSignIn {
     this.el = null;
     this.eventEmitter = new EventEmitter();
     this.eventCallbackMap = new WeakMap();
+    this.widgetHooks = new WidgetHooks(this.options.hooks);
 
     // if authClient is set, authParams are disregarded
     if (options.authClient) {
@@ -139,7 +149,7 @@ export default class OktaSignIn {
     onError?: RenderErrorCallback,
   ): Promise<RenderResult> {
     const { el } = options;
-    this.el = el;
+    this.el = el!;
 
     return new Promise<RenderResult>((resolve, reject) => {
       const onSuccessWrapper = (res: RenderResult): void => {
@@ -158,11 +168,12 @@ export default class OktaSignIn {
         if (target) {
           // @ts-ignore OKTA-508744
           render(h(Widget, {
+            ...this.options,
             authClient: this.authClient,
             globalSuccessFn: onSuccessWrapper,
             globalErrorFn: onErrorWrapper,
             eventEmitter: this.eventEmitter,
-            ...this.options,
+            widgetHooks: this.widgetHooks,
           }), target);
         } else {
           throw new Error(`could not find element ${el}`);
@@ -205,9 +216,13 @@ export default class OktaSignIn {
     );
   }
 
-  before(): void { }
+  before(formName: string, hook: HookFunction): void {
+    this.widgetHooks.addHook('before', formName, hook);
+  }
 
-  after(): void { }
+  after(formName: string, hook: HookFunction): void {
+    this.widgetHooks.addHook('after', formName, hook);
+  }
 
   hide(): void {
     this.eventEmitter.emit('hide', true);
@@ -236,13 +251,13 @@ export default class OktaSignIn {
     if (EVENTS_LIST.includes(eventName)) {
       // trap third-party callback errors
       const origHandler = eventHandler;
-      registeredEventHandler = (...callbackArgs) => {
+      registeredEventHandler = ((ctx: EventContext, error: EventErrorContext) => {
         try {
-          origHandler.call(this, ...callbackArgs);
+          origHandler.call(this, ctx, error);
         } catch (err) {
           console.error(`[okta-signin-widget] "${eventName}" event handler error:`, err);
         }
-      };
+      }) as OktaWidgetEventHandler;
       this.eventCallbackMap.set(origHandler, registeredEventHandler);
     }
     this.eventEmitter.on(eventName, registeredEventHandler);
@@ -262,7 +277,7 @@ export default class OktaSignIn {
   }
 
   private buildRenderOptions(
-    options: WidgetProps & Record<string, string> = {},
+    options: WidgetOptions & Record<string, string> = {},
   ): RenderOptions {
     const widgetOptions = this.options;
     // @ts-expect-error OKTA-508744
--- a/src/v3/src/OktaSignIn/index.ts
+++ b/src/v3/src/OktaSignIn/index.ts
@@ -16,19 +16,23 @@ import { h, render } from 'preact';
 import { TinyEmitter as EventEmitter } from 'tiny-emitter';
 
 import {
-  EventContext, EventErrorContext,
+  EventContext,
+  EventErrorContext,
   HookFunction,
-  OktaSignInAPI, RenderErrorCallback, RenderResult, RenderSuccessCallback,
+  RenderErrorCallback,
+  RenderResult,
+  RenderSuccessCallback,
 } from '../../../types';
-import { Widget } from '../components/Widget';
-import { JsonObject } from '../types';
 import {
+  JsonObject,
   OktaWidgetEventHandler,
   OktaWidgetEventType,
+  OktaSignInAPI,
   RenderOptions,
   WidgetOptions,
   WidgetProps,
-} from '../types/widget';
+} from '../types';
+import { Widget } from '../components/Widget';
 import { WidgetHooks } from '../util/widgetHooks';
 
 const EVENTS_LIST = ['ready', 'afterError', 'afterRender'];
@@ -148,8 +152,7 @@ export default class OktaSignIn implements OktaSignInAPI {
     onSuccess?: RenderSuccessCallback,
     onError?: RenderErrorCallback,
   ): Promise<RenderResult> {
-    const { el } = options;
-    this.el = el!;
+    this.el = options.el ?? null;
 
     return new Promise<RenderResult>((resolve, reject) => {
       const onSuccessWrapper = (res: RenderResult): void => {
@@ -176,7 +179,7 @@ export default class OktaSignIn implements OktaSignInAPI {
             widgetHooks: this.widgetHooks,
           }), target);
         } else {
-          throw new Error(`could not find element ${el}`);
+          throw new Error(`could not find element ${this.el}`);
         }
       } catch (error) {
         if (typeof onError === 'function') {
--- a/src/v3/src/OktaSignIn/index.ts
+++ b/src/v3/src/OktaSignIn/index.ts
@@ -260,7 +260,7 @@ export default class OktaSignIn implements OktaSignInAPI {
         } catch (err) {
           console.error(`[okta-signin-widget] "${eventName}" event handler error:`, err);
         }
-      }) as OktaWidgetEventHandler;
+      });
       this.eventCallbackMap.set(origHandler, registeredEventHandler);
     }
     this.eventEmitter.on(eventName, registeredEventHandler);
--- a/src/v3/src/OktaSignIn/index.ts
+++ b/src/v3/src/OktaSignIn/index.ts
@@ -23,16 +23,16 @@ import {
   RenderResult,
   RenderSuccessCallback,
 } from '../../../types';
+import { Widget } from '../components/Widget';
 import {
   JsonObject,
+  OktaSignInAPI,
   OktaWidgetEventHandler,
   OktaWidgetEventType,
-  OktaSignInAPI,
   RenderOptions,
   WidgetOptions,
   WidgetProps,
 } from '../types';
-import { Widget } from '../components/Widget';
 import { WidgetHooks } from '../util/widgetHooks';
 
 const EVENTS_LIST = ['ready', 'afterError', 'afterRender'];
--- a/src/v3/src/components/Widget/index.tsx
+++ b/src/v3/src/components/Widget/index.tsx
@@ -101,6 +101,7 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     eventEmitter,
     otp,
     flow,
+    widgetHooks,
   } = widgetProps;
 
   const [hide, setHide] = useState<boolean>(false);
@@ -125,6 +126,7 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
   const dataSchemaRef = useRef<FormBag['dataSchema']>();
   const [loading, setLoading] = useState<boolean>(false);
   const [widgetRendered, setWidgetRendered] = useState<boolean>(false);
+  const widgetRenderedOnce = useRef<boolean>(false);
   const [loginHint, setloginHint] = useState<string | null>(null);
   const languageCode = getLanguageCode(widgetProps);
   const languageDirection = getLanguageDirection(languageCode);
@@ -222,7 +224,8 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  const handleError = (error: unknown) => {
+  const handleError = async (error: unknown) => {
+    await widgetHooks.callHooks('before', undefined);
     // TODO: handle error based on types
     // AuthApiError is one of the potential error that can be thrown here
     // We will want to expose development stage errors from auth-js and file jiras against it
@@ -288,21 +291,17 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
         });
       }
 
+      await widgetHooks.callHooks('before', transaction);
+
       setResponseError(null);
       setIdxTransaction(transaction);
-      // ready event
-      eventEmitter?.emit?.('ready', {
-        stepName: transaction.nextStep?.name,
-      });
     } catch (error) {
       if (usingStateHandleFromSession) {
         // Saved stateHandle is invalid. Remove it from session
         // Bootstrap will be restarted with stateToken from widgetProps
         unsetStateHandle();
       } else {
-        eventEmitter?.emit?.('ready');
-
-        handleError(error);
+        await handleError(error);
       }
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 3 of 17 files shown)


src/v3/src/util/widgetHooks.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+}
+
+export const mergeWidgetHooks = (
+  hooks: HooksOptions, hookType: HookType, formName: string, hook: HookFunction,
+) => {
+  const newHooks = { ...hooks };
+  newHooks[formName] = newHooks[formName] || {};
+  newHooks[formName][hookType] = [
+    ...(newHooks[formName][hookType] || []),
```
leemchale-okta, line 49: nice I like this clean way of checking if there are existing hooks and populating them in one line

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -262,9 +252,14 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
         // No need to continue since onSuccess is called in registration hook callback
         return;
       }
+
+      if (!isClientTransaction) {
+        await executeWidgetHooks('before', hooksRef.current, newTransaction);
```
leemchale-okta, line 257: does gen 2 have any situations where they would execute the `before` hook before a client transaction?

src/v3/src/components/Widget/index.tsx
```diff
@@ -386,23 +392,33 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     if (isClientTransaction) {
       return;
     }
-    if (widgetRendered && typeof idxTransaction !== 'undefined'
-      && uischema.elements.length > 0) {
-      eventEmitter?.emit?.('afterRender', getEventContext(idxTransaction));
+    if (widgetRendered) {
+      (async () => {
+        const isFirstRender = !widgetRenderedOnce.current;
```
leemchale-okta, line 397:
```suggestion
        const isFirstRender = typeof widgetRenderedOnce.current === 'undefined';
```

leemchale-okta: nit just to be more specific since its a ref

leemchale-okta: or I believe you can just give the ref a default of `false` above so its set to a boolean

src/util/Hooks.js
```diff
@@ -33,9 +33,9 @@ export async function executeHooksAfter(hook) {
 export function mergeHook(hooks, formName, hookToMerge) {
   const existingHook = hooks[formName] = hooks[formName] || { before: [], after: [] };
```
shawyu-okta, line 34: Is this all still necessary if we have the default value setting (i.e. `|| []`) below?

src/v3/src/components/Widget/index.tsx
```diff
@@ -306,6 +315,7 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
 
   const resume = useCallback(async () => {
     await initLanguage();
+    let transaction: IdxTransaction | undefined;
```
shawyu-okta, line 318: I see you want to pass `transaction` to `handleError` in the `catch` block. Can you use `idxTransaction` instead and eliminate this pattern of having an uninitialized variable?

shawyu-okta: Same thing above in `bootstrap`

src/v3/src/components/Widget/index.tsx
```diff
@@ -188,33 +201,29 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
...
         return;
       }
-      const transaction: IdxTransaction = await authClient.idx.start({
+      transaction = await authClient.idx.start({
         stateHandle,
         // Required to prevent auth-js from clearing sessionStorage and breaking interaction code flow
         exchangeCodeForTokens: false,
       });
-      const hasError = !transaction.requestDidSucceed || transaction.messages?.some(
+      const hasError = !transaction!.requestDidSucceed || transaction!.messages?.some(
```
shawyu-okta, line 209: Are the `!` non-null assertions needed? I expect the return type of `await authClient.idx.start` would be `idxTransaction` with no possibility of `null | undefined`

src/v3/src/util/widgetHooks.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { IdxStatus, IdxTransaction } from '@okta/okta-auth-js';
+
+import { HookFunction, HooksOptions, HookType } from '../../../types';
+import { getFormNameForTransaction } from './getEventContext';
+
+/* eslint-disable no-await-in-loop, no-restricted-syntax */
+export async function executeWidgetHooks(
+  hookType: HookType,
+  hooks?: HooksOptions,
```
shawyu-okta, line 21: Curious if `hooks` can be kept as a singleton in this file instead of using a `ref` in the `Widget/index.tsx` file and needing to pass the ref around, awkwardly needing to access `ref.current`, etc.

Since it's a `ref` we already are implying it does not affect component rendering so we could potentially simplify by using a singleton. I might also suggest using `Map`s here instead of arrays if we can make that move.

src/v3/src/hooks/useOnSubmit.ts
leemchale-okta: cc: @glenfannin-okta

src/util/Hooks.js
leemchale-okta: I think `hooks[formName]` could contain other hooks other than `before` or `after`?  Which would make this not initialize with default `{ before: [], after: [] };`

leemchale-okta: ignore my last comment, it does look like this is not necessary with the default value setting below

src/v3/src/hooks/useOnSubmit.ts
denysoblohin-okta: I use `isClientTransaction` as an indicator that request did not succeed. 
In this case "before" and "after" hooks should not be executed for "terminal" form. As it's done for Gen2.

See test `should send "afterError" event (and not execute hooks for "terminal" form) if submit results with error`: https://github.com/okta/okta-signin-widget/pull/3376/files#diff-189986d3f484c4c778c28206906c6fb9e91e4d71404d94ad174eddba3e0acfdcR565

src/v3/src/components/Widget/index.tsx
denysoblohin-okta: Actually there is no need to pass transaction to `handleError`. Removed unnecessary code changes.

src/util/Hooks.js
denysoblohin-okta: It's necessary.
I've found a bug in Gen 2 when added new tests for hooks.
If we have only "after" hook in options, and then trying to add new "before" hook with `.before()` API, we'll get `TypeError: Cannot read property 'concat' of undefined`

https://github.com/okta/okta-signin-widget/blob/bc61d536e2aece3e6e1afe2741d25e47295e301f/.widgetrc.sample.js#L20-L25

https://github.com/okta/okta-signin-widget/blob/36fc36650f9a229f0e2ff556c0fea63cead245fc/test/testcafe/spec/OktaSignIn_spec.js#L144-L146

src/v3/src/types/widget.ts
```diff
@@ -69,7 +71,8 @@ export type RenderOptions = {
 export type AuthenticationMode = 'remediation' | 'relying-party';
 
 export type OktaWidgetEventHandler = {
-  (...args: unknown[]): void;
+  (context: EventContext, error?: EventErrorContext): void;
```
denysoblohin-okta, line 74: to allow `class OktaSignIn implements OktaSignInAPI`

src/v3/src/util/widgetHooks.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  /* eslint-disable no-restricted-syntax */
+  constructor(hooksOptions?: HooksOptions) {
+    this.hooks = new Map();
+    if (hooksOptions) {
+      for (const [formName, formHooks] of Object.entries(hooksOptions)) {
+        const formHooksMap = new Map<string, HookFunction[]>();
+        for (const [hookType, hooksByType] of Object.entries(formHooks)) {
+          if (hooksByType) {
+            formHooksMap.set(hookType, hooksByType);
```
shawyu-okta, line 29: Can we use `this.addHook` instead here? Would be better to use the same public method to add hooks.

I also see the returned type of `hookType` and `hooksByType` is `any` right now so we should address that as well

src/v3/src/components/Widget/index.tsx
```diff
@@ -210,7 +212,10 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  const handleError = (error: unknown) => {
+  const handleError = async (error: unknown) => {
+    if (widgetHooks) {
+      await widgetHooks.callHooks('before', undefined, error as Error);
+    }
```
shawyu-okta, line 218: Anywhere we are doing a `if(widgetHooks)` guard we can take advantage of optional chaining syntax:
```suggestion
    await widgetHooks?.callHooks('before', undefined, error as Error);
```

shawyu-okta: We could also make `widgetHooks` a required prop

src/v3/src/components/Widget/index.tsx
```diff
@@ -438,23 +439,33 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     if (isClientTransaction) {
       return;
     }
-    if (widgetRendered && typeof idxTransaction !== 'undefined'
-      && uischema.elements.length > 0) {
-      eventEmitter?.emit?.('afterRender', getEventContext(idxTransaction));
+    if (widgetRendered) {
+      (async () => {
+        const isFirstRender = typeof widgetRenderedOnce.current === 'undefined';
```
shawyu-okta, line 444: We can set `widgetRenderedOnce` to be `false` by default when creating and avoid needing do this check:
```ts
const widgetRenderedOnce = useRef<boolean>(false);
```
Then we can directly check the value in the conditional below with `if (widgetRenderedOnce.current)`

src/v3/src/components/Widget/index.tsx
```diff
@@ -438,23 +439,33 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
...
+        const executeAfterHooks = () => widgetHooks && widgetHooks.callHooks('after', idxTransaction, responseError);
```
shawyu-okta, line 445: Optional chaining simplifies this statement:
```suggestion
        const executeAfterHooks = () => widgetHooks?.callHooks('after', idxTransaction, responseError);
```

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -78,27 +76,17 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
...
-      const error = formatError(resp);
-      return {
-        xhr: error as unknown as ErrorXHR,
-        errorSummary: error.responseJSON && error.responseJSON.errorSummary,
-      };
-    };
-
     // TODO: Revisit and refactor this function as it is a dupe of handleError fn in Widget/index.tsx
-    const handleError = (transaction: IdxTransaction | undefined, error: unknown) => {
+    const handleError = async (transaction: IdxTransaction | undefined, error: unknown) => {
```
shawyu-okta, line 80: `async` no longer needed

shawyu-okta (commented):
Some comments. Thank for you adding testcafe tests! I think you may be to move some of the testcafe spec tests to unit tests for `Hooks.js` to make tests run a bit faster. Either way I would like to see some unit test coverage on `Hooks.js`

src/v3/src/util/widgetHooks.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  /* eslint-disable no-await-in-loop */
+  public async callHooks(
+    hookType: HookType,
+    idxTransaction?: IdxTransaction,
+    error?: Error | null,
+  ): Promise<void> {
+    if (!(idxTransaction || error)) {
+      return;
+    }
```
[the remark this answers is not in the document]
denysoblohin-okta: I think it's better to not pass `error` at all because its value is not used in method body.
```
callHooks(
    hookType: HookType,
    idxTransaction?: IdxTransaction | null
  )
```
If `idxTransaction` is null, then it's error case and need to execute hooks for "terminal"

src/v3/src/types/widget.ts
```diff
@@ -59,7 +61,7 @@ export interface ErrorXHR {
 }
 
 export type RenderOptions = {
-  el: string;
+  el?: string;
```
denysoblohin-okta, line 64: to allow `class OktaSignIn implements OktaSignInAPI`

src/v3/src/components/Widget/index.tsx
denysoblohin-okta: I made `widgetHooks` a required prop and pass it to `<Widget>` in [`setup()`](https://github.com/okta/okta-signin-widget/blob/master/src/v3/test/integration/util/setup.tsx#L43) for tests
Otherwise there will be error `Cannot read properties of undefined (reading 'callHooks')`

src/v3/src/util/widgetHooks.ts
```diff
@@ -0,0 +1,58 @@
...
+  /* eslint-disable no-restricted-syntax */
+  constructor(hooksOptions?: HooksOptions) {
+    this.hooks = new Map();
+    if (hooksOptions) {
+      for (const [formName, formHooks] of Object.entries(hooksOptions)) {
+        for (const [hookType, hooksByType] of Object.entries(formHooks)) {
+          for (const hook of (hooksByType || []) as HookFunction[]) {
+            this.addHook(hookType as HookType, formName, hook);
+          }
+        }
```
shawyu-okta, line 30: TS is making you cast `hookType as HookType` because it can't infer the types of anything that isn't `before` or `after` so it gives you `hooksByType: any`. We both know there will be only `before` or `after` but TS isn't smart enough to know that. Since there's only two cases we could avoid the type cast by directly handling both cases explicitly and writing the inner `for` loop write, once for each case.

Another option is to force TS to understand by using a string array of the literals we care about, which effectively does the same thing but explicitly:
```ts
['before', 'after'].forEach(hookType => {
  for (const hook of (formHooks[hookType] || []) as HookFunction[]) {
    this.addHook(hookType, formName, hook);
  }
});
```

src/v3/src/util/widgetHooks.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  /* eslint-disable no-await-in-loop */
+  public async callHooks(
+    hookType: HookType,
+    idxTransaction?: IdxTransaction,
+  ): Promise<void> {
+    const formName = getFormNameForTransaction(idxTransaction);
+    if (formName) {
+      const hooksToExecute = this.hooks.get(formName)?.get(hookType) || [];
+      if (hooksToExecute.length) {
```
shawyu-okta, line 51: this length check isn't necessary. if the array is empty the for loop below just won't run

src/v3/src/OktaSignIn/index.ts
```diff
@@ -237,13 +255,13 @@ export default class OktaSignIn {
...
-      registeredEventHandler = (...callbackArgs) => {
+      registeredEventHandler = ((ctx: EventContext, error: EventErrorContext) => {
         try {
-          origHandler.call(this, ...callbackArgs);
+          origHandler.call(this, ctx, error);
         } catch (err) {
           console.error(`[okta-signin-widget] "${eventName}" event handler error:`, err);
         }
-      };
+      }) as OktaWidgetEventHandler;
```
shawyu-okta, line 264: Can you file a ticket to fix this casting? I think it's possible to avoid this but I see why you've done it. Will just take some time with typescript docs and a little bit of creativity but I don't want this one to block the PR. Once you make the jira, please link it in a comment above this type cast.

shawyu-okta (approved):
Some suggestions I think will be helpful, plus some questions that just need to be answered before merging. Overall lgtm from my side if the notes are addressed.

I'd like to see @lesterchoi-okta take a look over this tomorrow before you merge, especially to ensure that our external `OktaSignIn` API is acceptable since we cannot change that later.

src/v3/src/OktaSignIn/index.ts
```diff
@@ -159,14 +172,15 @@ export default class OktaSignIn {
         if (target) {
           // @ts-ignore OKTA-508744
           render(h(Widget, {
+            ...this.options,
             authClient: this.authClient,
             globalSuccessFn: onSuccessWrapper,
             globalErrorFn: onErrorWrapper,
             eventEmitter: this.eventEmitter,
-            ...this.options,
+            widgetHooks: this.widgetHooks,
```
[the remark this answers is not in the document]
denysoblohin-okta: Putting `eventEmitter` and `widgetHooks` after `...this.options` is needed to prevent lint error `This spread always overwrites this property.`
I looks fine to put spread at the top as we don't expect overwriting any of listed properties (`authClient`, `globalSuccessFn`, `globalErrorFn `, `eventEmitter` and `widgetHooks`) from options.

src/v3/src/components/Widget/index.tsx
```diff
@@ -448,34 +443,42 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     if (isClientTransaction) {
       return;
     }
-    if (widgetRendered && typeof idxTransaction !== 'undefined'
-      && uischema.elements.length > 0) {
-      eventEmitter?.emit?.('afterRender', getEventContext(idxTransaction));
+    if (widgetRendered) {
+      (async () => {
```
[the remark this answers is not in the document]
denysoblohin-okta: Using `async` with `useEffect` will emit type error `'() => Promise<void>' is not assignable to parameter of type 'EffectCallback'`
https://github.com/preactjs/preact/issues/3239
https://github.com/facebook/react/blob/f2c381131fb58c107e6153255bc8d1d6340db506/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L103

src/v3/src/components/Widget/index.tsx
```diff
@@ -448,34 +443,42 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
     if (isClientTransaction) {
       return;
     }
-    if (widgetRendered && typeof idxTransaction !== 'undefined'
-      && uischema.elements.length > 0) {
```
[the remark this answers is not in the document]
denysoblohin-okta: Good point. 
I saw your conversation at https://github.com/okta/okta-signin-widget/pull/3290#discussion_r1245414155
Condition `idxTransaction?.status !== IdxStatus.SUCCESS` can be changed to `uischema.elements.length > 0`

src/v3/src/OktaSignIn/index.ts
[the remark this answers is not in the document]
shawyu-okta: Ok, yeah I'm convinced now that in fact we shouldn't spread at the bottom of the object anyways since we don't want to overwrite the `authClient` (or any of the other ones, actually) accidentally since it's always set in the constructor function.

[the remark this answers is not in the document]
shawyu-okta: Actually we're kind of spreading in extra properties to the `Widget` component we don't need, like `onSuccess`, since it's still in the `options` object after we add the wrappers around them. Not a big deal and no change needed for your PR but realizing we could clean this up a bit.

src/v3/src/OktaSignIn/index.ts
denysoblohin-okta: Changed `OktaWidgetEventHandler` declaration to
```
export type OktaWidgetEventHandler = EventCallback | EventCallbackWithError;
```

src/v3/src/types/widget.ts
```diff
@@ -107,7 +110,10 @@ export type PageTitleCallback = (context: EventContext, param: PageTitleCallback
 export type OktaWidgetEventType = 'ready' | 'afterError' | 'afterRender';
 export type IDPDisplayType = 'PRIMARY' | 'SECONDARY';
 
-export type WidgetProps = Partial<WidgetOptions>;
+export type WidgetProps = Partial<WidgetOptions> & {
+  eventEmitter: EventEmitter;
+  widgetHooks: WidgetHooks; // instance of class WidgetHooks
```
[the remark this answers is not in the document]
denysoblohin-okta: As I understand, `WidgetOptions` are "public" options that can be passed to `OktaSignIn` constructor. 
And `WidgetProps` are props passed to `<Widget />`. 
`eventEmitter` and `widgetHooks` are not public, so I moved them to `WidgetProps` (and marked as required as [you suggested](https://github.com/okta/okta-signin-widget/pull/3376#discussion_r1338846208)).

But maybe we can just use
```ts
export type WidgetProps = WidgetOptions;
```
since `Partial` is not needed
and move `eventEmitter` and `widgetHooks`  to `WidgetOptions` as optional. For simplicity.
@lesterchoi-okta ?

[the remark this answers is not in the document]
shawyu-okta: Oh you're right, it might be right to keep `WidgetOptions` and `WidgetProps` separate the way you already have them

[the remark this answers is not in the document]
shawyu-okta: I wrote a utility type yesterday that might be helpful:
```ts
type PartialExcept<T, K extends keyof T> = Partial<Omit<T, K>> & Pick<T, K>;
```

lesterchoi-okta (approved):

PR #3381, opened 2023-09-25
SIW Gen 3: Min & Max Length Validation on Sign Up Page

trevoring-okta:
## Description:
- [OKTA-648346](https://oktainc.atlassian.net/browse/OKTA-648346) found that a backend error message was being propagated to the frontend because we had not yet implemented client-side validation and error handling for string fields whose value exceeded their `maxLength` property present on the `auth-js` `Input` type
- This PR adds support for throwing a validation error and blocking form submission if a field does not meet `maxLength` or `minLength` constraints , which is parity with gen 2
- This PR also refactors the field validation error testing in `src/v3/src/transformer/field/transform.ts` to follow the `ErrorTransformer` and `ErrorTester` pattern used in `src/v3/src/transformer/terminal/transformUnhandledErrors.ts` for readability and extensibility purposes

## PR Checklist

### Issue:

- [OKTA-648346](https://oktainc.atlassian.net/browse/OKTA-648346)
- [Bacon Link](https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&branch=ti-OKTA-648346-ssr-sign-up-errors&page=1&pageSize=6&tab=main)

#### Before implementing max length validation:
https://github.com/okta/okta-signin-widget/assets/106110543/0b051142-3b6a-474c-b558-130e73bdc282

#### After implementing max length validation:
https://github.com/okta/okta-signin-widget/assets/106110543/74ba0e7c-552a-439e-99d1-608dc2d5be61

#### After implementing min length validation:
https://github.com/okta/okta-signin-widget/assets/106110543/5ae6fc19-8457-4d04-8dc2-149fa5eb1115


src/v3/src/transformer/field/transform.ts
```diff
@@ -41,28 +49,72 @@ const mapUiElement = (input: Input): FieldElement => {
...
+          message: loc('model.validation.field.blank', 'login'),
+          i18n: { key: 'model.validation.field.blank' },
+        }
+      ),
+    },
+  ];
+
+  let messages: WidgetMessage[] | undefined;
+  // Validation errors can be divided into populated and blank field checks
+  if (fieldValue) {
```
trevoring-okta, line 110: Would like some thoughts on whether populated error checks vs. blank error checks is a good (and exhaustive) split for this error testing pattern? Per description, see `transformUnhandledErrors.ts` for the pattern that this logic was modeled on

src/v3/src/transformer/field/transform.ts
```diff
@@ -103,10 +155,10 @@ export const transformStepInputs = (
           validate(data) {
             // in the case of a required boolean input, just return true
             // if the backend requires the checbox val to be true, error will be displayed
-            if (typeof data[name] === 'boolean' || !!data[name]) {
```
trevoring-okta, line 106: `!!data[name]` was originally here to check if the field was populated, and if so, short circuit and return `undefined` since we didn't have any validation for populated fields before adding max length validation

src/v3/src/transformer/field/transform.ts
```diff
@@ -41,28 +49,72 @@ const mapUiElement = (input: Input): FieldElement => {
...
+
+  let messages: WidgetMessage[] | undefined;
+  // Validation errors can be divided into populated and blank field checks
+  if (fieldValue) {
+    messages = populatedFieldErrorChecks
+      .filter(({ tester }) => (tester(fieldValue)))?.map((error) => error.message());
+  } else {
+    messages = blankFieldErrorChecks
+      .filter(({ tester }) => (tester(fieldValue)))?.map((error) => error.message());
   }
```
trevoring-okta, line 116: Would like some thoughts on whether populated field vs. blank field error checks is a good (and exhaustive) split for classifying the different validation checks we are doing? Per the PR description, this testing pattern was based on the logic seen in `getWidgetMessage()` within `transformUnhandledErrors.ts`

src/v3/src/transformer/field/transform.ts
```diff
@@ -41,28 +49,72 @@ const mapUiElement = (input: Input): FieldElement => {
...
-    && isCustomizedI18nKey(customizedErrorConfig.key, widgetProps)) {
-    errorMessage.message = loc(customizedErrorConfig.key, 'login');
-    errorMessage.i18n.key = customizedErrorConfig.key;
+  const useCustomizedBlankErrorMessage = !!customizedErrorConfig && name === IDX_STEP.IDENTIFY
+                                    && isCustomizedI18nKey(customizedErrorConfig.key, widgetProps);
+
+  const populatedFieldErrorChecks: ValidationErrorTester[] = [
+    {
+      // Validate maxLength for string inputs that have the property set
+      tester: (value?: unknown) => !!maxLength && type === 'string' && !!value && (value as string).length > maxLength,
```
glenfannin-okta, line 73: While we are here, we can probably also define a validator for min length (i.e in Enroll profile allows org admins to set the string length minimums and maximums). 

This is the key:
`model.validation.field.string.minLength`

src/v3/src/transformer/field/transform.ts
```diff
@@ -41,28 +49,72 @@ const mapUiElement = (input: Input): FieldElement => {
...
+  ];
+
+  let messages: WidgetMessage[] | undefined;
+  // Validation errors can be divided into populated and blank field checks
+  if (fieldValue) {
+    messages = populatedFieldErrorChecks
+      .filter(({ tester }) => (tester(fieldValue)))?.map((error) => error.message());
+  } else {
+    messages = blankFieldErrorChecks
+      .filter(({ tester }) => (tester(fieldValue)))?.map((error) => error.message());
```
glenfannin-okta, line 115: Can the `blankFieldErrorChecks` and `populatedFieldErrorChecks` be combined into one array? Then, if the `find` function is used on the array, it will just step through the list until the tester function returns true. So as long as the items are in order (that we want them displayed) in the array, I _think_ these could be combined into one.

test/testcafe/spec/IdentifyRegistration_spec.js
```diff
@@ -136,6 +134,82 @@ test.meta('v3', false).requestHooks(mock)('should show errors after empty requir
   await t.expect(registrationPage.hasEmailErrorMessage()).eql(true);
 });
 
+test.requestHooks(mock)('should show max length field validation errors', async t => {
```
lesterchoi-okta, line 137: @trevoring-okta will this behave for non-Latin character, e.g., Chinese, Japanese, Korean (CJK)? Maybe worth testing

https://help.okta.com/en-us/content/topics/reference/ref-supported-languages.htm

playground/mocks/data/idp/idx/enroll-profile-new.json
```diff
@@ -20,18 +20,27 @@
               "value": [
                 {
                   "name": "lastName",
+                  "type": "string",
```
lesterchoi-okta, line 23: Just curious, did something force you to add `"type": "string"`? IIRC It's implied according to the Ion Spec.

lesterchoi-okta (approved):

yarn.lock
```diff
@@ -13546,12 +13546,12 @@ json-stringify-safe@^5.0.1, json-stringify-safe@~5.0.1:
...
   integrity sha512-ZClg6AaYvamvYEE82d3Iyd3vSSIjQ+odgjaTzRuO3s7toCdFKczob2i0zCh7JE8kWn17yvAWhUVxvqGwUalsRA==
 
-json5@2.x, json5@^0.5.1, json5@^1.0.1, json5@^2.1.0, json5@^2.1.2, json5@^2.1.3, json5@^2.2.1, json5@^2.2.2:
+json5@2.x, json5@^0.5.1, json5@^2.1.0, json5@^2.1.3, json5@^2.2.1, json5@^2.2.2:
   version "2.2.3"
   resolved "https://registry.yarnpkg.com/json5/-/json5-2.2.3.tgz#78cd6f1a19bdc12b73db5ad0c61efd66c1e29283"
   integrity sha512-XmOWe7eyHYH14cLdVPoyg+GOH3rYX++KpzrylJwSW98t3Nk+U8XOl8FWKOgwtzdb8lXGf6zYwDUzeHMWfxasyg==
 
-json5@^1.0.2:
+json5@^1.0.1, json5@^1.0.2, json5@^2.1.2:
```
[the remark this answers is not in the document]
trevoring-okta: Not intentional, will revert

playground/mocks/data/idp/idx/enroll-profile-new.json
trevoring-okta: I have `type ==== 'string'` as part of my test logic before I can cast the field value to a `string` and compare its `length` to `maxLength` and `minLength`, so I had to add this to the mock otherwise the error isn't thrown

test/testcafe/spec/IdentifyRegistration_spec.js
```diff
@@ -136,6 +134,82 @@ test.meta('v3', false).requestHooks(mock)('should show errors after empty requir
...
+  await t.expect(registrationPage.hasLastNameError()).eql(true);
+  await t.expect(registrationPage.hasLastNameErrorMessage()).eql(true);
+  await t.expect(registrationPage.getLastNameErrorMessage()).contains('This field cannot exceed the maximum allowed characters');
+  await t.expect(registrationPage.hasFirstNameError()).eql(true);
+  await t.expect(registrationPage.getFirstNameErrorMessage()).contains('This field cannot exceed the maximum allowed characters');
+  await t.expect(registrationPage.hasEmailError()).eql(true);
+  await t.expect(registrationPage.hasEmailErrorMessage()).eql(true);
+  await t.expect(registrationPage.getEmailErrorMessage()).contains('This field cannot exceed the maximum allowed characters');
+
+  // Populate fields (minLenght = 2) with 2 characters
```
trevoring-okta, line 160: Realizing I mixed up test logic for min and max test, resolving in next push

test/testcafe/spec/IdentifyRegistration_spec.js
```diff
@@ -136,6 +134,82 @@ test.meta('v3', false).requestHooks(mock)('should show errors after empty requir
...
+  await registrationPage.fillFirstNameField('ab');
+  await registrationPage.fillLastNameField('ab');
+  await registrationPage.fillEmailField('ab');
+  await registrationPage.focusRegisterButton();
+
+  // All three enroll fields should not show min length validation error
+  await t.expect(registrationPage.hasLastNameError()).eql(false);
+  await t.expect(registrationPage.hasLastNameErrorMessage()).eql(false);
+  await t.expect(registrationPage.hasFirstNameError()).eql(false);
```
trevoring-okta, line 169: Realizing I mixed up test logic for min and max test, resolving in next push

test/testcafe/spec/IdentifyRegistration_spec.js
trevoring-okta: Good callout, I'm assuming you asked this because some non-Latin characters can be more than one "code unit", which I just learned is what JS `String.length` actually uses to calculate length. I tested gen2 with the 😄 emoji since it takes up 2 code units and it seems that gen2 must use code units to enforce length rather than actual characters because the name fields generated by the mock with `"minLength": 2` allowed a single emoji as a first name.

@lesterchoi-okta Do we want to go with parity for gen2, which would mean leaving the usage of `String.length` as the validation logic, or would you consider this a bug and we should switch to logic that always returns the exact # of characters?

This link has an example for getting the true # of characters: [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length)

playground/mocks/data/idp/idx/enroll-profile-new.json
shawyu-okta: @trevoring-okta does the real API response send this too?

trevoring-okta: @shawyu-okta I believe so, here's an example response for the email field that my preview org sends me:
`label
: 
"Email"
maxLength
: 
100
name
: 
"userProfile.email"
required
: 
true
type
: 
"string"`

trevoring-okta: Is it possible that not all orgs send the `type`?

shawyu-okta: I think all orgs and endpoints should send `type` then because I imagine they should use the same serializer for Ion Spec. I guess we could be even safer by assuming `type === 'string'` iff `type` is `undefined`.

PR #3389, opened 2023-09-27
SIW Gen 3: Enable Chrome DTC view and spec

trevoring-okta:
## Description:
- Implements the Chrome DTC polling view for gen 3
- Enables and updates the corresponding spec tests (one sidenote is that none of the tests in this spec were actually running since this spec got created because we were missing `await` before calling `assertChromeDTCView()`)

## PR Checklist

### Issue:

- [OKTA-616189](https://oktainc.atlassian.net/browse/OKTA-616189)
- [Bacon Link](https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&branch=ti-OKTA-616189-dtc-view&page=1&pageSize=6&sha=ab44c49962e1f782a035b223f185f0b6ff7c6120&tab=main)

### Screenshot/Video:
Screenshot 2023-09-27 at 5 35 22 PM

### DIFF
--- a/src/v3/src/components/ChromeDtcContainer/ChromeDtcContainer.tsx
+++ b/src/v3/src/components/ChromeDtcContainer/ChromeDtcContainer.tsx
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2022-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { h } from 'preact';
+
+import { useWidgetContext } from '../../contexts';
+import {
+  ChromeDtcContainerElement, UISchemaElementComponent,
+} from '../../types';
+import { getBaseUrl } from '../../util';
+import HiddenIFrame from '../HiddenIFrame';
+
+const ChromeDtcContainer: UISchemaElementComponent<{
+  uischema: ChromeDtcContainerElement
+}> = ({ uischema }) => {
+  const {
+    options: {
+      href,
+    },
+  } = uischema;
+  const { widgetProps } = useWidgetContext();
+  const deviceChallengeUrl = new URL(href as string, getBaseUrl(widgetProps)).toString();
+
+  return (
+    <HiddenIFrame
+      id="chrome-dtc-container"
+      src={deviceChallengeUrl}
+    />
+  );
+};
+
+export default ChromeDtcContainer;
--- a/src/v3/src/components/ChromeDtcContainer/ChromeDtcContainer.tsx
+++ b/src/v3/src/components/ChromeDtcContainer/ChromeDtcContainer.tsx
@@ -28,7 +28,7 @@ const ChromeDtcContainer: UISchemaElementComponent<{
     },
   } = uischema;
   const { widgetProps } = useWidgetContext();
-  const deviceChallengeUrl = new URL(href as string, getBaseUrl(widgetProps)).toString();
+  const deviceChallengeUrl = new URL(href, getBaseUrl(widgetProps)).toString();
 
   return (
     <HiddenIFrame
--- a/src/v3/src/components/HiddenIFrame/HiddenIFrame.tsx
+++ b/src/v3/src/components/HiddenIFrame/HiddenIFrame.tsx
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2022-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { Box } from '@mui/material';
+import { FunctionComponent, h } from 'preact';
+
+type HiddenIFrameProps = {
+  id?: string;
+  src: string;
+};
+const HiddenIFrame: FunctionComponent<HiddenIFrameProps> = ({ id, src }) => (
+  <Box
+    id={id}
+    component="iframe"
+    src={src}
+    sx={{
+      display: 'none',
+    }}
+  />
+);
+
+export default HiddenIFrame;
--- a/src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
+++ b/src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
@@ -14,6 +14,7 @@ import { NextStep } from '@okta/okta-auth-js';
 
 import { CHALLENGE_METHOD, IDX_STEP } from '../../../constants';
 import {
+  ChromeDtcContainerElement,
   DescriptionElement,
   IdxStepTransformer,
   IStepperContext,
@@ -37,11 +38,14 @@ const getTitleText = (challengeMethod: string) => {
     case CHALLENGE_METHOD.CHROME_DTC:
       return loc('chrome_dtc.title', 'login');
 
+    case CHALLENGE_METHOD.CUSTOM_URI:
+      return loc('customUri.title', 'login');
+
     case CHALLENGE_METHOD.UNIVERSAL_LINK:
       return loc('universalLink.title', 'login');
 
     default:
-      return loc('customUri.title', 'login');
+      return '';
   }
 };
 
@@ -50,14 +54,14 @@ const getDescriptionText = (challengeMethod: string) => {
     case CHALLENGE_METHOD.APP_LINK:
       return loc('appLink.content', 'login');
 
+    case CHALLENGE_METHOD.CUSTOM_URI:
+      return loc('customUri.required.content.prompt', 'login');
+
     case CHALLENGE_METHOD.UNIVERSAL_LINK:
       return loc('universalLink.content', 'login');
 
-    case CHALLENGE_METHOD.CHROME_DTC:
-      return '';
-
     default:
-      return loc('customUri.required.content.prompt', 'login');
+      return '';
   }
 };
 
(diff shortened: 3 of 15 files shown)


src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -30,23 +30,35 @@ import {
...
-    return loc('universalLink.content', 'login');
+  switch (challengeMethod) {
+    case CHALLENGE_METHOD.APP_LINK:
+      return loc('appLink.content', 'login');
+
+    case CHALLENGE_METHOD.UNIVERSAL_LINK:
+      return loc('universalLink.content', 'login');
+
+    case CHALLENGE_METHOD.CHROME_DTC:
+      return '';
```
trevoring-okta, line 57: Is it worth adding this empty string as an insurance policy so we don't accidentally render the custom URI description text for DTC view in case the `DescriptionElement` accidentally gets added to the `uischema` despite the if check added below?
```   
if (challengeMethod !== CHALLENGE_METHOD.CHROME_DTC) {
    uischema.elements.push(descriptionElement);
    uischema.elements.push(openOktaVerifyButton);
}
```

leemchale-okta: imo I think its fine, I think it also makes it more clear and consistent that the chrome dtc method should not have a description

leemchale-okta (approved):

test/testcafe/spec/DeviceProbingChromeDTC_spec.js
```diff
@@ -24,48 +24,47 @@ const mockChromeProbingViewAfterOVProbing = RequestMock()
...
-  .respond((req, res) => {
+  .respond((_, res) => {
     res.statusCode = '200';
     res.headers['content-type'] = 'application/json';
     res.setBody(identify);
   });
 
-// TODO: OKTA-616189 - implement this view in Gen3
-fixture('Device Challenge Polling View for Chrome DTC').meta('v3', false);
+fixture('Device Challenge Polling View for Chrome DTC').meta('v3', true);
```
shawyu-okta, line 67: Sorry to cause you a merge conflict because of this but I just merged the change that means you can remove the `.meta('v3',` part and it will run the test by default.

test/testcafe/spec/DeviceProbingChromeDTC_spec.js
```diff
@@ -8,13 +8,13 @@ import identifyWithDeviceProbingLoopback from '../../../playground/mocks/data/id
 
 const mockChromeProbingView = RequestMock()
   .onRequestTo(/\/idp\/idx\/introspect/)
-  .respond((req, res) => {
+  .respond((_, res) => {
```
shawyu-okta, line 11: Was this causing a linting error for you?

trevoring-okta: I think I thought it was but I just re-added and it passes. Do we have a preference to leave `req` in there for readability even if they're always unused variables?

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
shawyu-okta: This isn't a bad idea, but you could also see if moving the `default` case to a case for `CHALLEGE_METHOD.CUSTOM_URI` and then change the fallthrough `default` to always return `''`

trevoring-okta: That makes more sense, thanks for the suggestion

shawyu-okta: I'm just sure if for some reason we depend on the `default` case to be the custom uri text. You could try it and see what happens in tests

shawyu-okta: If you do this may also want to consider updating the other switch case above as well

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -180,12 +192,14 @@ export const transformOktaVerifyDeviceChallengePoll: IdxStepTransformer = ({
     return formBag;
   }
 
-  if (challengeMethod === CHALLENGE_METHOD.APP_LINK
-    || challengeMethod === CHALLENGE_METHOD.UNIVERSAL_LINK) {
+  if (challengeMethod !== CHALLENGE_METHOD.CUSTOM_URI) {
```
shawyu-okta, line 195: This is OK, but consider using something like this to be clear which ones we're including for adding the spinner:
```ts
if ([
  CHALLENGE_METHOD.APP_LINK,
  CHALLENGE_METHOD.UNIVERSAL_LINK,
  CHALLENGE_METHOD.CHROME_DTC,
].contains(challengeMethod)) {
```

test/testcafe/spec/DeviceProbingChromeDTC_spec.js
shawyu-okta: I think either way is fine, and we do have a linting exception to allow the `_` as a placeholder when you need to skip a parameter like this

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
leemchale-okta: IIRC I dont think custom uri actually needs to be the default case so it should be fine

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -30,23 +31,38 @@ import {
...
+      return loc('chrome_dtc.title', 'login');
+
+    case CHALLENGE_METHOD.CUSTOM_URI:
+      return loc('customUri.title', 'login');
+
+    case CHALLENGE_METHOD.UNIVERSAL_LINK:
+      return loc('universalLink.title', 'login');
+
+    default:
+      return '';
```
glenfannin-okta, line 48: It seems like (based on the logic before the update) uses `return loc('customUri.title', 'login');` as the default?

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -30,23 +31,38 @@ import {
...
+      return loc('appLink.content', 'login');
+
+    case CHALLENGE_METHOD.CUSTOM_URI:
+      return loc('customUri.required.content.prompt', 'login');
+
+    case CHALLENGE_METHOD.UNIVERSAL_LINK:
+      return loc('universalLink.content', 'login');
+
+    default:
+      return '';
```
glenfannin-okta, line 64: Similar, should this be used as the default? `loc('customUri.required.content.prompt', 'login');`

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
leemchale-okta: yeah I remember I wrote that and made it look like custom uri is the default fallback case but thats not actually needed.  I looked at how gen 2 handles it and confirmed they use a switch with a specific custom uri case: https://github.com/okta/okta-signin-widget/blob/9d0a36dbe47c28a4df989da89b669e51d26a6996/src/v2/view-builder/utils/ChallengeViewUtil.js#L46

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
leemchale-okta: same as above

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
trevoring-okta: Yeah we had a discussion about this above already @glenfannin-okta, but thanks for the callout and thanks for following up on this @leemchale-okta!

src/v3/src/components/HiddenIFrame/HiddenIFrame.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Box } from '@mui/material';
+import { FunctionComponent, h } from 'preact';
+
+type HiddenIFrameProps = {
+  id?: string;
+  src: string;
+};
+const HiddenIFrame: FunctionComponent<HiddenIFrameProps> = ({ id, src }) => (
+  <Box
+    id={id}
```
shawyu-okta, line 22: Can you just check to make sure this won't render `id="undefined"` when `id` isn't passed?

src/v3/src/components/ChromeDtcContainer/ChromeDtcContainer.tsx
```diff
@@ -0,0 +1,41 @@
...
+const ChromeDtcContainer: UISchemaElementComponent<{
+  uischema: ChromeDtcContainerElement
+}> = ({ uischema }) => {
+  const {
+    options: {
+      href,
+    },
+  } = uischema;
+  const { widgetProps } = useWidgetContext();
+  const deviceChallengeUrl = new URL(href as string, getBaseUrl(widgetProps)).toString();
```
shawyu-okta, line 31: is the `as string` cast needed? `href` should already be `string`

shawyu-okta: also just wondering if the `new URL` part is strictly necessary to get the full URL for the iframe. I just don't see this being done in gen2 but it also uses the courage `View` which might do stuff like this for us. If we do need the fully qualified URL then using this is the right way to do it (as opposed to string concatentation)

shawyu-okta (approved):
last few clarifying comments that hopefully don't block. add some unit testing to cover the transformer changes and then you're good to go.

src/v3/src/components/ChromeDtcContainer/ChromeDtcContainer.tsx
trevoring-okta: This URL construction logic was copied over from how `OpenOktaVerifyFPButton.tsx` constructs its URL, which was added via Lee's PR here: https://github.com/okta/okta-signin-widget/pull/3031

You're right that the `as string` cast is not necessary here because I made `href` required for the `ChromeDtcContainerElement` `options`, good callout. In terms of the constructing the URL via a URL object, I think for consistency's sake it's fine to leave it since that's how the button component constructs the URL, unless you have strong opinions

shawyu-okta: > In terms of the constructing the URL via a URL object, I think for consistency's sake it's fine to leave it since that's how the button component constructs the URL, unless you have strong opinions

agree. let's keep it then, thanks for the explanation

src/v3/src/components/HiddenIFrame/HiddenIFrame.tsx
trevoring-okta: In regards to this, do you mean do a sanity check that the string `"undefined"` doesn't get set as the `id` when the `id` is omitted?

shawyu-okta: @trevoring-okta yeah exactly. Just want to make sure it omits the property instead of the value getting coerced to the string "undefined".

trevoring-okta: Got it, just checked and it omits the `id` 👍

PR #3083, opened 2023-02-14
fix: Adds error handler for Authenticator picker on Unlock Account page

LazarVuckovic-okta:
## Description:

Adds an error handler for when the user is trying to unlock account with only the identifier entered, but no authenticator selected.

## PR Checklist

- [x] Have you verified the basic functionality for this change?
- [x] Did you add tests, as appropriate, following our [Automated Test guidelines](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2676497890/Automated+Testing+in+the+Signin+Widget)?
- [x] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?
- [x] Did you verify the change by running [downstream monolith artifact](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/102897979/Sign-in+Widget+Development#Sign-inWidgetDevelopment-Instructionstocreateandrunthedownstreamartifact(d16t))? (Provide link to build below)
- [x] Does this PR include noticeable changes to the UI? (If yes, attach screenshots/video below)

### Issue:

- [OKTA-520182](https://oktainc.atlassian.net/browse/OKTA-520182)

### Reviewers:
@glenfannin-okta 

### Screenshot/Video:
https://okta.box.com/s/b05z6crke0v8j2dlcrq9tdgs10rivpvd

### Downstream Monolith Build:

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&branch=d16t-okta-signin-widget-032ae2e-63ebf3e2&page=1&pageSize=6&sha=223730e3fa478c2c2d7df10f40ef9e5332166aa7&tab=main

PR #3396, opened 2023-10-02
SIW Gen 3: Unlock Account Back to Sign In Error

trevoring-okta:
## Description:
- In gen3, we changed our `Link` component to be an HTML `button` under the hood but omitted an explicit `type="button"`, so the default value for `type` ended up being `type="submit"`. This was causing an error where the user was being sent back to the sign in screen when pressing Enter in the input of the unlock account screen because it was targeting the "Back to sign in" link as the only valid submit button in the form
- This small fix is pre-GA and doesn't fully replicate the gen2 UX, which also raises an error on pressing Enter (see [#3083](https://github.com/okta/okta-signin-widget/pull/3083)), but we will do a post-GA fast follow to meet parity here 

## PR Checklist

### Issue:

- [OKTA-653788](https://oktainc.atlassian.net/browse/OKTA-653788)

#### Back to sign in error
https://github.com/okta/okta-signin-widget/assets/106110543/761cf629-51fc-4cff-a4bb-2564418e1179

### DIFF
--- a/src/v3/src/components/Link/Link.tsx
+++ b/src/v3/src/components/Link/Link.tsx
@@ -65,6 +65,8 @@ const Link: UISchemaElementComponent<{
     typeof href === 'undefined' ? (
       <LinkMui
         component="button"
+        // Fixes OKTA-653788 (see comments) - Currently we treat all links as buttons
+        type="button"
         variant="body1"
         role="link"
         onClick={onClick}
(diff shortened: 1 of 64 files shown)


shawyu-okta (approved):
based on the video it looks like it still jumps back to the main widget flow after pressing "enter", is that right? I was expecting it to stay on the same view.

trevoring-okta:
@shawyu-okta sorry I put a small header above the video but that is the video of the error for more context on it. I didn't include a video after this fix because I figured it's easier to visualize that nothing happens now after pressing Enter

shawyu-okta:
@trevoring-okta oh i see, thanks for clarifying

PR #3430, opened 2023-10-24
Reenable IdentifyWithRememberUsername_spec

trevoring-okta:
## Description:
- Gen3 version of [OKTA-552341](https://oktainc.atlassian.net/browse/OKTA-552341), which was fixed in #3063
- Reenables IdentifyWithRememberUsername_spec by storing identifier in `ln` username cookie on form submission

## PR Checklist

### Issue:

- [OKTA-585939](https://oktainc.atlassian.net/browse/OKTA-585939)
- [Bacon Link](https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&branch=ti-OKTA-585939-reenable-identify-cookie-spec&page=1&pageSize=6&sha=ead268860e18edc00e9b9e3eb060e085d792e42c&tab=main)

### DIFF
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -141,15 +141,13 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     // Allow username transformation if applicable
     if ('identifier' in payload) {
       payload.identifier = transformIdentifier(widgetProps, step, payload.identifier as string);
-    }
 
-    // Widget rememberMe feature stores the entered identifier in a cookie, to pre-fill the form on subsequent visits to page
-    if (features?.rememberMe) {
-      if (payload.identifier) {
+      // Widget rememberMe feature stores the entered identifier in a cookie, to pre-fill the form on subsequent visits to page
+      if (features?.rememberMe) {
         setUsernameCookie(payload.identifier as string);
+      } else {
+        removeUsernameCookie();
       }
-    } else {
-      removeUsernameCookie();
     }
 
     // For Granular Consent remediation, scopes within the `optedScopes`
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -310,6 +310,7 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     widgetProps,
     eventEmitter,
     widgetHooks,
+    features,
     setResponseError,
     setIdxTransaction,
     setIsClientTransaction,
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -34,7 +34,9 @@ import {
   loc,
   postRegistrationSubmit,
   preRegistrationSubmit,
+  removeUsernameCookie,
   SessionStorage,
+  setUsernameCookie,
   toNestedObject,
   transformIdentifier,
   triggerRegistrationErrorMessages,
@@ -64,7 +66,7 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     setStepToRender,
     widgetProps,
   } = useWidgetContext();
-  const { eventEmitter, widgetHooks } = widgetProps;
+  const { eventEmitter, widgetHooks, features } = widgetProps;
 
   return useCallback(async (options: OnSubmitHandlerOptions) => {
     setLoading(true);
@@ -139,6 +141,15 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     // Allow username transformation if applicable
     if ('identifier' in payload) {
       payload.identifier = transformIdentifier(widgetProps, step, payload.identifier as string);
+    }
+
+    // Widget rememberMe feature stores the entered identifier in a cookie, to pre-fill the form on subsequent visits to page
+    if (features?.rememberMe) {
+      if (payload.identifier) {
+        setUsernameCookie(payload.identifier as string);
+      }
+    } else {
+      removeUsernameCookie();
     }
 
     // For Granular Consent remediation, scopes within the `optedScopes`
--- a/src/v3/test/integration/identify-with-password-error-flow.test.tsx
+++ b/src/v3/test/integration/identify-with-password-error-flow.test.tsx
@@ -55,6 +55,7 @@ describe('identify-with-password-error-flow', () => {
   });
 
   it('should display warning message when invalid identifier is entered and should allow user to resubmit same information without showing client-side errors', async () => {
+    jest.spyOn(cookieUtils, 'getUsernameCookie').mockReturnValue('');
     const {
       authClient,
       user,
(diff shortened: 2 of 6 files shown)


src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -141,6 +143,15 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
...
     }
 
+    // Widget rememberMe feature stores the entered identifier in a cookie, to pre-fill the form on subsequent visits to page
+    if (features?.rememberMe) {
+      if (payload.identifier) {
+        setUsernameCookie(payload.identifier as string);
+      }
+    } else {
+      removeUsernameCookie();
+    }
```
shawyu-okta, line 153: Should this go inside the block above on L142 guarded by `if ('identifier' in payload)`?

test/testcafe/spec/IdentifyWithRememberUsername_spec.js
```diff
@@ -225,9 +225,7 @@ test.requestHooks(identifyRequestLogger, identifyMock)('should pre-fill identifi
   await t.expect(identifier).eql('testUsername@okta.com');
 });
 
-
-// OKTA-585939 Cookie is not updated in gen 3 widget
-test.meta('gen3', false).requestHooks(identifyRequestLogger, identifyMock)('should store identifier in ln cookie when updated', async t => {
+test.requestHooks(identifyRequestLogger, identifyMock)('should store identifier in ln cookie when updated', async t => {
```
shawyu-okta, line 228: I think I've seen the username field pre-filled in some of our test deployments even before the changes in this PR. Is there something else that was causing the field to be pre-filled?

src/v3/src/hooks/useOnSubmit.ts
trevoring-okta: Ah yeah good catch

test/testcafe/spec/IdentifyWithRememberUsername_spec.js
trevoring-okta: Yeah we also have logic for pre-populating the username field [in the identify transformer](https://github.com/okta/okta-signin-widget/blob/dc0726438fcd15f8d716463f46307259eaf96846/src/v3/src/transformer/identify/transformIdentify.ts#L49C1-L59C1) but I've checked in a preview org and the original bug that Jared fixed for gen2 is present in gen3. I'm not 100% sure how adding this logic to set the cookie upon saving the form fixes the bug, maybe @jaredperreault-okta could give a bit more info?

trevoring-okta: @shawyu-okta chatted with Jared on Slack about the gen2 ticket and got some more info. Was also able to verify that this change should fix the bug from reappearing in gen3

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -139,6 +141,13 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     // Allow username transformation if applicable
     if ('identifier' in payload) {
       payload.identifier = transformIdentifier(widgetProps, step, payload.identifier as string);
+
+      // Widget rememberMe feature stores the entered identifier in a cookie, to pre-fill the form on subsequent visits to page
+      if (features?.rememberMe) {
+        setUsernameCookie(payload.identifier as string);
+      } else {
+        removeUsernameCookie();
+      }
```
shawyu-okta, line 150: This code looks sound, but wondering if the placement is what we want. We already have very similar code in https://github.com/okta/okta-signin-widget/blob/5e12b8f1430d6dbfa16391199a71e57fe80ebdc7/src/v3/src/transformer/terminal/transformTerminalTransaction.ts#L139 but it only runs on a success terminal transaction. I'm wondering if we should not set this cookie until we reach a success state because otherwise we will overwrite the cookie before the user's authentication is successful.

trevoring-okta: Posting what Jared said about the motivation for the original fix he made in gen2 from our DM: 

_"The rememberMyUsername feature was implemented differently in g1 and g2. g1 it stored the username in a cookie after the user typed, where g2 wait until a terminal success state was reached. I believe a customer reported an issue with MFA, where the username was being extracted from the IDX response and when MFA was enabled, the IDX pipeline lost the context of the username and feature broke. The fix was to store the username when the user entered their username and proceeded, mimicking the g1 implementation and fixed the MFA issue."_

So basically the current state of gen3 matches the state of gen2 BEFORE Jared made the fix because it only store the username cookie upon reaching a success terminal state. I was able to repro the original bug in the current state of gen3 and verify that this change does fix it. Since this cookie is only used to pre-populate the username field on the identify screen, I believe the thought is that it's safe to overwrite it whenever the user proceeds from the identify screen even if they don't reach a success terminal state because we presume that the user entered their username correctly and even if their transaction failed they would want to try again with the same username pre-populated.

Let me know if any of that isn't clear

glenfannin-okta: Quick question on this, does that mean we no longer need the logic in the terminal transaction transformer?

src/v3/test/integration/identify-with-password-error-flow.test.tsx
```diff
@@ -71,6 +71,7 @@ describe('identify-with-password-error-flow', () => {
           status: 200,
         },
       },
+      widgetOptions: { features: { rememberMe: false } },
```
glenfannin-okta, line 74: Just curious, why was this needed? 🤔

trevoring-okta: If you see this code in `transformIdentify` https://github.com/okta/okta-signin-widget/blob/8beb7299d230864090837133fd30ffc7b600f7bd/src/v3/src/transformer/identify/transformIdentify.ts#L53-L58
we actually default `rememberMe` to true if it's undefined or falsey

src/v3/src/hooks/useOnSubmit.ts
trevoring-okta: That I'm not sure about, I can try removing it and see how it affects our tests?

glenfannin-okta: Does this logic still exist in Gen 2? Maybe Jared may know, I would assume he would have removed it when fixing this for that version.

src/v3/test/integration/identify-with-password-error-flow.test.tsx
glenfannin-okta: Ohh, that makes sense, I forgot that feature is defaulted to true in g2 and we wanted to preserve it 🙏

src/v3/src/hooks/useOnSubmit.ts
trevoring-okta: Oh wait yeah you're right, if you look at this part of his gen2 PR he removed the analogous terminal transaction code:
https://github.com/okta/okta-signin-widget/pull/3063/files#diff-9d2ede01662be7d3291032cd313089a2f2a7399c19f8f078a7b2910c2a38f141
I have a new Bacon run going right now without the gen3 cookie terminal transaction code so hopefully that passes 🙏

shawyu-okta: Thanks @trevoring-okta for the context -- that's clear to me.

shawyu-okta (approved):

PR #3433, opened 2023-10-24
OKTA-658263 : Odyssey 1.x migration testcafe VRT

leemchale-okta:
## Description:

This PR adds support for visual regression testing in the testcafe UI components demo spec test.  It also adds VRT for RTL regressions in UI components.  This is to assist in catching ui regressions as odyssey is being migrated to 1.x in gen 3.

## PR Checklist

### Issue:

- [OKTA-658263](https://oktainc.atlassian.net/browse/OKTA-658263)
- [OKTA-612762](https://oktainc.atlassian.net/browse/OKTA-612762)


package.json
```diff
@@ -100,7 +100,7 @@
     "copy:types": "grunt copy:types",
     "build:types": "yarn clean:types && yarn generate:types && yarn copy:types",
     "test:parity-setup": "mkdir -p test-reports/testcafe && yarn generate-config && yarn workspace v3 dev",
-    "test:parity-run": "OKTA_SIW_GEN3=true yarn testcafe -c 2 chrome:headless --reporter spec,xunit:build2/reports/junit/testcafe-results.xml",
+    "test:parity-run": "OKTA_SIW_GEN3=true yarn testcafe -c 2 \"chrome:headless --force-device-scale-factor=1\" --reporter spec,xunit:build2/reports/junit/testcafe-results.xml",
```
leemchale-okta, line 103: must set `--force-device-scale-factor=1` in chrome browser to keep screenshots quality consistent between different computers

test/testcafe/spec/uiDemo_spec.js
```diff
@@ -0,0 +1,39 @@
...
+
+    // freeze the spinner element so screenshots are consistent
+    await pageObject.stopSpinnerAnimation();
+
+    await t.customActions.compareScreenshot('one');
+    await pageObject.scrollToNextPage();
+
+    // possible bug in testcafe where the animated progress bar at the bottom is displaying after scrolling
+    // causing inconsistent screenshots
+    await pageObject.hideProgressBar();
```
leemchale-okta, line 30: this was a bug in testcafe that was found previously and fixed: https://github.com/DevExpress/testcafe/issues/2076

but the bug persists after scrolling the page down for some reason

src/v3/package.json
```diff
@@ -125,7 +125,7 @@
     "stylelint-use-logical-spec": "^5.0.0",
     "svgo": "^2.3.0",
     "terser-webpack-plugin": "^5.2.4",
-    "testcafe": "^1.20.1",
+    "testcafe": "^3.3.0",
```
shawyu-okta, line 128: 🎉 

Wonder if this can be done for the root version of `testcafe` as well?

test/testcafe/spec/uiDemo_spec.js
```diff
@@ -0,0 +1,39 @@
...
+  const pageObject = new UiDemoPageObject(t);
+  await pageObject.navigateToPage();
+  await t.expect(pageObject.formExists()).eql(true);
+  return pageObject;
+}
+
+fixture('UI demo');
+
+test
+  .requestHooks(uiDemoMock)('UI demo VRT', async t => {
```
shawyu-okta, line 19: Should this be marked only to run in Gen3?

test/testcafe/spec/uiDemo_spec.js
```diff
@@ -0,0 +1,39 @@
...
+    const pageObject = await setup(t);
+
+    // freeze the spinner element so screenshots are consistent
+    await pageObject.stopSpinnerAnimation();
+
+    await t.customActions.compareScreenshot('one');
```
shawyu-okta, line 25: nit: use numbers instead of the string for better sorting of the output filenames

test/testcafe/framework/page-objects/UiDemoPageObject.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  async stopSpinnerAnimation() {
+    await ClientFunction(() => {
+      const spinnerRoot = document.querySelector('.MuiCircularProgress-root');
+      const spinnerCircle = document.querySelector('.MuiCircularProgress-circle');
+      spinnerRoot.setAttribute('style', 'animation: none;');
+      spinnerCircle.setAttribute('style', 'animation: none;');
+    })();
+  }
```
shawyu-okta, line 32: This is a good idea to halt the spinner or put it into a predictable state, but I don't actually see the spinner at all in the screenshots right now.

vrtUtil/vrtUtil.js
```diff
@@ -0,0 +1,77 @@
...
+const getDiffImageName = (imagePath) => {
+  return path.join(
+    path.dirname(imagePath),
+    `${path.basename(imagePath, path.extname(imagePath))}-diff.png`,
+  );
+};
+
+const doVisualRegression = async (testObject, name, updateScreenshot) => {
+  // set window size so screenshots are consistent
+  await testObject.resizeWindow(800, 800);
```
shawyu-okta, line 20: We may need to consider putting the window back to the original size at the end of this method

vrtUtil/vrtUtil.js
```diff
@@ -0,0 +1,77 @@
...
+  if(typeof testObject === 'undefined') {
+    return;
+  }
```
shawyu-okta, line 23: I think this should be the first statement in this method

vrtUtil/ImageDiff.js
```diff
@@ -0,0 +1,77 @@
...
+    }));
+  }
+
+  getImageOutput() {
+    return this.options.imageOutputPath;
+  }
+
+  getDifference() {
+    // eslint-disable-next-line no-mixed-operators
+    return Math.round(100 * 100 * this.differences / (this.width * this.height)) / 100;
```
shawyu-okta, line 68: What does this math do? Can you add a comment explaining what / why we calculate here?

vrtUtil/vrtUtil.js
```diff
@@ -0,0 +1,77 @@
+const path = require('path');
+const fs = require('fs');
+const ImageDiff = require('./ImageDiff');
+
+const VISUAL_REGRESSION_THRESHOLD = 0.01;
```
shawyu-okta, line 5: nit: if this a percentage? if so, we can name it a bit more clearly

vrtUtil/ImageDiff.js
```diff
@@ -0,0 +1,77 @@
...
+
+class ImageDiff {
+  constructor(opts) {
+    this.options = { ...opts };
+    this.width = 0;
+    this.height = 0;
+    this.differences = 0;
+  }
+
+  async run(callback) {
```
shawyu-okta, line 38: Consider directly returning the results instead of using a callback pattern

vrtUtil/ImageDiff.js
```diff
@@ -0,0 +1,77 @@
...
+  return new Promise((resolve, reject) => {
+    fs.createReadStream(image).pipe(new PNG())
+      .on('parsed', function onParse() {
+        resolve(this);
+      })
+      .on('error', reject);
+  });
+}
+
+function fixPngImage(image, width, height) {
```
shawyu-okta, line 15: Can we rename this function to clarify what it's doing or add comments to describe what it's doing?

test/testcafe/framework/page-objects/UiDemoPageObject.js
```diff
@@ -0,0 +1,33 @@
...
+
+  async hideProgressBar() {
+    await ClientFunction(() => {
+      // the testcafe progress bar element is in the shadow dom so we cannot query it directly
+      const style = document.createElement('style');
+      style.textContent = '.progress-bar-hammerhead-shadow-ui { visibility: hidden !important; }';
+      document.head.appendChild(style);
+    })();
+  }
```
shawyu-okta, line 23: Some of these _might_ actually be good for the `BasePageObject` or a utility function

src/v3/package.json
leemchale-okta: @shawyu-okta im having issues in bacon with this upgrade.  I might have to revert this back to the original

test/testcafe/framework/page-objects/UiDemoPageObject.js
```diff
@@ -0,0 +1,33 @@
...
+
+  async stopSpinnerAnimation() {
+    await ClientFunction(() => {
+      const spinnerRoot = document.querySelector('.MuiCircularProgress-root');
+      const spinnerCircle = document.querySelector('.MuiCircularProgress-circle');
+      spinnerRoot.setAttribute('style', 'animation: none;');
+      spinnerCircle.setAttribute('style', 'animation: none;');
```
trevoring-okta, line 30: I know you mentioned you were having issues with getting the Spinner to freeze yesterday. I just tried it locally and was able to get it to freeze by using `animation: none!important`. I know `!important` isn't great style, but I think it's fine for this use case, have you tried using it here?

src/v3/package.json
leemchale-okta: testcafe version was reverted back to original due to bugs and performance issues

test/testcafe/spec/uiDemo_spec.js
```diff
@@ -0,0 +1,42 @@
...
+  .respond(uiDemoResponse);
+
+async function setup(t) {
+  const pageObject = new UiDemoPageObject(t);
+  await pageObject.navigateToPage();
+  await t.expect(pageObject.formExists()).eql(true);
+  return pageObject;
+}
+
+fixture('UI demo').meta('gen3', false).meta('gen2', false);
```
leemchale-okta, line 18: this spec runs in its own bacon suite and is excluded from running in the gen 2 `testcafe` and `parity-v3` run

test/testcafe/spec/uiDemo_spec.js
leemchale-okta: this fixture is now disabled from `testcafe` and `parity-v3` above

vrtUtil/ImageDiff.js
leemchale-okta: this method calculates the amount of difference between the two images as a percentage (not a decimal)

vrtUtil/vrtUtil.js
leemchale-okta: I think since the vrt is only used in the uiDemo spec now outside of the parity-v3 suite, this shouldn't be a problem

vrtUtil/vrtUtil.js
leemchale-okta: will add a comment

test/testcafe/framework/page-objects/UiDemoPageObject.js
leemchale-okta: I think since these are now only used by the uiDemo spec outside of the `parity-v3`/`testcafe` suites, it can probably stay here for now

vrtUtil/ImageDiff.js
```diff
@@ -0,0 +1,77 @@
+const fs = require('fs');
+const { PNG } = require('pngjs');
+const pixelmatch = require('pixelmatch');
```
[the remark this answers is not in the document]
leemchale-okta: yeah this file was being referenced to by the testcafe config file before when we were using the `customActions` feature and it didn't support that syntax.  But it has since been refactored and no longer referenced there so this can be changed back to use `import` 👍

test/testcafe/framework/page-objects/UiDemoPageObject.js
leemchale-okta: Hmm I wasn't able to get it to work using that.  The spinner appears in your screenshot?

leemchale-okta: was able to get it to work by appending the style to the `head` instead

test/testcafe/framework/page-objects/UiDemoPageObject.js
leemchale-okta: fixed by appending the style to the `head` instead

test/testcafe/spec/uiDemo_spec.js
```diff
@@ -0,0 +1,28 @@
...
+
+test
+  .requestHooks(uiDemoMock)('UI demo VRT', async t => {
+    const pageObject = await setup(t);
+
+    // freeze the spinner element so screenshots are consistent
+    await pageObject.stopSpinnerAnimation();
+
+    await compareScreenshot(t, { fullPage: true, strictMode: true });
```
leemchale-okta, line 27: instead of having multiple screenshots we take one of the fullpage.  This way it is easier to manage and all the ui components are shown together as a whole, without being cut off

src/v3/test/e2e/.testcaferc.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   },
   appCommand: 'test/e2e/start-app-e2e.sh',
   browsers: 'chrome:headless',
-  concurrency: 3,
+  concurrency: 1,
```
leemchale-okta, line 13: new version of testcafe does not like when concurrency is higher than the amount of tests

.bacon.yml
```diff
@@ -146,3 +146,17 @@ test_suites:
...
+    script_path: /root/okta/okta-signin-widget/scripts
+    sort_order: '32'
+    timeout: '40'
+    script_name: vrt-v3
+    criteria: MERGE
+    queue_name: small
+  - name: vrt-v3-get-screenshots
+    script_path: /root/okta/okta-signin-widget/scripts
+    sort_order: '33'
+    timeout: '40'
```
shawyu-okta, line 159: Do these suites actually take this long? If not, we should set them to something closer to the usual run time.

playground/fonts.css
```diff
@@ -0,0 +1,171 @@
+@font-face {
+    font-family: Proxima Nova;
+    font-style: normal;
+    font-weight: 400;
+    src: url(https://op3static.oktacdn.com/assets/loginpage/font/assets/proximanova-reg-webfont.9d5837512674046fa816.eot);
```
shawyu-okta, line 5: Can you file a ticket and add a comment in this file to address this hard-coded preview static URL?

vrtUtil/ImageDiff.js
```diff
@@ -0,0 +1,87 @@
...
+  // this method calculates the amount of difference as a percentage (not a decimal)
+  getDifferencePercent() {
+    // eslint-disable-next-line no-mixed-operators
+    return Math.round(100 * 100 * this.differences / (this.width * this.height)) / 100;
+  }
+
+  hasPassed() {
+    // strict mode does not allow a single difference in pixels.  Exact 1:1 match
+    if (this.options.strictMode) {
+      return this.differences === 0;
```
shawyu-okta, line 80: I'm not sure I understand how this works with `strictMode`, can you explain?

vrtUtil/vrtUtil.js
```diff
@@ -0,0 +1,110 @@
+const path = require('path');
+const fs = require('fs');
+const ImageDiff = require('./ImageDiff');
+
+// Matching threshold as a decimal percentage of the maximum acceptable square distance between two colors;
+// Ranges from 0 to 1. Smaller values make the comparison more sensitive.
+const DEFAULT_VRT_THRESHOLD = 0.1;
```
shawyu-okta, line 7: This value is a decimal fraction but the comparison in `ImageDiff` is denoted as a percentage

vrtUtil/vrtUtil.js
```diff
@@ -0,0 +1,110 @@
...
+const BACON_CI_BASE_PATH = path.join('build2', 'reports', 'vrt', 'artifacts', 'screenshots');
```
shawyu-okta, line 8:
```suggestion
const BASE_PATH = path.join('build2', 'reports', 'vrt', 'artifacts', 'screenshots');
```

shawyu-okta (commented):
I know we already created the topic branch but I see the button overflow bug present in the initial VRTs so we might consider rebasing the topic branch so we have a clean baseline image

.bacon.yml
leemchale-okta: yeah no this should be changed to about 15 mins

scripts/vrt-v3-get-screenshots.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+export CHROME_HEADLESS=true
+setup_service google-chrome-stable 119.0.6045.105-1
```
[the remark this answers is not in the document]
leemchale-okta: i dont believe they do and can be moved down

vrtUtil/ImageDiff.js
leemchale-okta: since we are now taking our base screenshots in the same environment every time (in bacon), the benefit of this is that there will be no outside variability in the screenshot quality/rendering.  This way can do exact 1:1 pixel matching comparison.  So in this `hasPassed` method we can check if there are 0 different pixels, if not it will fail the VRT.

shawyu-okta:
Should we use an actual RTL language for the RTL screenshot?

vrtUtil/vrtUtil.js
leemchale-okta: so this value is actually the threshold that is passed into the `pixelmatch` comparison algorithm, and isnt a threshold directly for the passing percentage, which could be confusing.  In `pixelmatch` the threshold is not exactly a minimum percentage of difference, its the percentage of the maximum acceptable square distance between two colors which i commented above. 
 ref: https://github.com/mapbox/pixelmatch/blob/ced69a33f0a3c8e9377a221e274693690f69be22/index.js#L45
 
 I think what I can do is rename this var to be more accurate, as well as create another var for a maximum difference percentage threshold

leemchale-okta:
> Should we use an actual RTL language for the RTL screenshot?

I think not totally necessary but would be beneficial.  Ill see if I can get an rtl language into the playground

src/v3/package.json
```diff
@@ -30,7 +30,10 @@
     "test:dev": "NODE_ENV=development jest",
     "test:e2e": "testcafe test/e2e/tests/**/*.js --config-file test/e2e/.testcaferc.js",
     "test:e2e:screenshots": "yarn test:e2e -s path=test/e2e/screenshots,thumbnails=false",
-    "test:report": "PREACT_APP_USE_PKCE=false NODE_ENV=test jest --coverage"
+    "test:report": "PREACT_APP_USE_PKCE=false NODE_ENV=test jest --coverage",
+    "test:vrt": "testcafe \"chrome:headless --force-device-scale-factor=1\" --config-file test/vrt/.testcaferc.js",
+    "test:vrt-ci": "VRT_CI=true yarn test:vrt",
```
shawyu-okta, line 35: Oh I was thinking the env variable change would be set in the shell script like `export VRT_CI=true` so this script wouldn't be needed

leemchale-okta: ohh okay ill try that

leemchale-okta: yep looks like that works fine 👍

vrtUtil/ImageDiff.js
leemchale-okta: the `differences` variable represents the actual number of pixels that are different in the image

test/testcafe/framework/page-objects/UiDemoPageObject.js
```diff
@@ -0,0 +1,39 @@
...
+
+  // freezes mui spinner components
+  async stopSpinnerAnimation() {
+    await ClientFunction(() => {
+      const rootSpinnerStyle = document.createElement('style');
+      rootSpinnerStyle.textContent = '.MuiCircularProgress-root { animation: none !important; }';
+      document.head.appendChild(rootSpinnerStyle);
+
+      const circleSpinnerStyle = document.createElement('style');
+      circleSpinnerStyle.textContent = '.MuiCircularProgress-circle { animation: none !important; }';
```
shawyu-okta, line 35: Curious if this can be expanded to use a single rule to stop **all** animations on the page:
```css
disableAnimationsStyle.textContent = '* { animation: none !important; }';
```

src/v3/test/vrt/start-app-vrt.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+export NODE_OPTIONS="--max-old-space-size=5120"
+yarn dev
```
shawyu-okta, line 4: We should be able to use the other script: `test/e2e/start-app-e2e.sh` and remove this one since they are identical

scripts/buildtools/commands/verify-package.js
```diff
@@ -17,10 +17,10 @@ const EXPECTED_BUNDLE_SIZES = {
...
   'okta-sign-in.polyfill.js': 504 * KB,
   'okta-sign-in.polyfill.min.js': 108 * KB,
-  //  Revert this to 1.9 when odyssey 1.x migration is complete and legacy package is removed
-  'okta-sign-in.next.js': 2.1 * MB,
-  //  Revert this to 1.7 when odyssey 1.x migration is complete and legacy package is removed
-  'okta-sign-in.next.no-polyfill.js': 1.9 * MB,
+  //  Revert this when odyssey 1.x migration is complete and legacy package is removed
+  'okta-sign-in.next.js': 1.9 * MB,
+  //  Revert this when odyssey 1.x migration is complete and legacy package is removed
+  'okta-sign-in.next.no-polyfill.js': 1.7 * MB,
```
shawyu-okta, line 23: Are the bundles smaller now?

shawyu-okta (approved):
Curious if the update to `src/v3/test/integration/__snapshots__/authenticator-enroll-security-question-error.test.tsx.snap` was intended

scripts/buildtools/commands/verify-package.js
leemchale-okta: yeah after rebasing with master that has the bundle reducing logic, the bundles got smaller

leemchale-okta:
> Curious if the update to `src/v3/test/integration/__snapshots__/authenticator-enroll-security-question-error.test.tsx.snap` was intended

yeah it was intended, after rebasing and running the unit tests, somehow this one didnt get updated

test/testcafe/framework/page-objects/UiDemoPageObject.js
leemchale-okta: yep this should work fine

PR #3440, opened 2023-10-26
Fixing lockfile drift

trevoring-okta:
## Description:
- Adds `yarn_sync()` in `setup.sh` to check for changes to `yarn.lock` in CI and force users to check in an updated lockfile
- Removes `@okta/okta-signin-widget: "link:../../dist"` in test app `package.json` because it was leading to an empty resolution in the lockfile on every `yarn install`, which would raise a false positive for the `yarn_sync()` logic
- Ensures that `yarn install` in `setup` uses public yarn registry and then reverts to internal artifactory for `@okta:registry` for all following scripts
- Cleans up naming of different registries used for `setup`, `publish`, and `verify-registry-install` scripts

## PR Checklist

### Issue:

- [OKTA-653621](https://oktainc.atlassian.net/browse/OKTA-653621)
- [Bacon Link](https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&branch=ti-OKTA-653621-fix-lockfile-drift&page=1&pageSize=12&sha=8faa60bdc0f49018cb90c18f75af49d3a3e516bc&tab=main)

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -5,10 +5,18 @@ set -eo pipefail
 # This is available from the "downstream artifact" menu on any okta-auth-js build in Bacon.
 # DO NOT MERGE ANY CHANGES TO THIS LINE!!
 export AUTHJS_VERSION=""
+export INTERNAL_REGISTRY="${ARTIFACTORY_URL}/api/npm/npm-okta-release"
+export PUBLIC_REGISTRY="https://registry.yarnpkg.com"
 
-# Install required node version
-export REGISTRY_REPO="npm-topic"
-export REGISTRY="${ARTIFACTORY_URL}/api/npm/${REGISTRY_REPO}"
+function yarn_sync() {
+  echo "Checking to see if the yarn.lock file is in sync with given dependencies"
+  if test "$( git status --short -- '*yarn.lock' | wc -l )" -ne 0
+  then
+    return 1
+  fi
+  echo "yarn sync OK"
+  return 0
+}
 
 if ! setup_service node v16.19.1 &> /dev/null; then
   echo "Failed to install node"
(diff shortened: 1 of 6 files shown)


test/app/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "dependencies": {
-    "@okta/okta-signin-widget": "link:../../dist"
```
shawyu-okta, line 12: @denysoblohin-okta you had originally added this for reasons related to the ESM build for the test app. @trevoring-okta and I couldn't find any problems running the ESM build now with this change, so maybe it's not longer a problem now. Can you take a look and explain if there's any cases that you think will fail?

scripts/verify-registry-install.sh
```diff
@@ -2,7 +2,7 @@
 
 # NOTE: MUST BE RAN *AFTER* THE PUBLISH SUITE
 
-export REGISTRY="${ARTIFACTORY_URL}/npm-topic"
+export PUBLISHED_REGISTRY="${ARTIFACTORY_URL}/npm-topic"
```
shawyu-okta, line 5: Should this be `PUBLISH_REGISTRY` to match the one in `publish.sh` with the same value?

test/app/package.json
trevoring-okta: @shawyu-okta Sorry forgot to update you directly via Slack DM but I already checked with @denysoblohin-okta and he verified that the `link:../../dist` is no longer necessary

scripts/setup.sh
```diff
@@ -20,13 +28,23 @@ if ! setup_service yarn 1.22.19 &> /dev/null; then
   exit ${FAILED_SETUP}
 fi
 
-cd ${OKTA_HOME}/${REPO}
+cd "${OKTA_HOME}"/"${REPO}" || exit
```
denysoblohin-okta, line 31: Just curious, why  this change is needed?

trevoring-okta: Ah that was leftover from copying some logic from a different script, it's not necessary I can change it back to minimize changes

test/app/package.json
shawyu-okta: Ok nice, no worries

shawyu-okta (approved):
lgtm. Did you run a test to make sure the `yarn_sync` is catching things as you'd expect in CI?

trevoring-okta:
@shawyu-okta Yeah before removing the `link:../../dist`, the `yarn_sync()` was causing failures on Bacon for `setup.sh` after `yarn install` because the lockfile was getting changed to have the empty resolution

PR #3446, opened 2023-10-27
Changes text in Gen3 for optional authenticator enrollment view

shawyu-okta:
## PR Checklist

### Issue:

- [OKTA-549620](https://oktainc.atlassian.net/browse/OKTA-549620)

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&branch=d16t-okta-signin-widget-17a1bd8-65491acb&page=1&pageSize=6&sha=8ff907569d26412a0ce5dd369e3e6cfe23e6caf3&tab=main


src/v3/src/components/Heading/Heading.tsx
```diff
@@ -37,8 +37,8 @@ const Heading: UISchemaElementComponent<{
     >
       <Typography
         id={uischema.id}
-        variant={`h${level ?? 2}`}
-        component={`h${visualLevel ?? 3}`}
+        variant={`h${visualLevel ?? 2}`}
+        component={`h${level ?? 3}`}
```
shawyu-okta, line 41: Turns out these were reversed so I've fixed that in this PR and you'll see the updates to transformers (and snapshots) that were rendering this component to reflect that.

src/v3/src/components/AuthCoin/AuthCoin.test.tsx
```diff
@@ -25,10 +25,6 @@ describe('AuthCoin tests', () => {
     };
   });
 
-  afterEach(() => {
-    cleanup();
-  });
```
shawyu-okta, line 30: Apparently this happens automatically and has been for a long time. I was the one that added these lines in a PR to reduce test flakiness but perhaps it wasn't the issue and didn't help (nor did it harm) our tests.

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-cleanup

src/v3/.eslintrc.js
```diff
@@ -307,7 +313,18 @@ module.exports = {
...
+        'testing-library/no-container': 'warn',
+        'testing-library/prefer-screen-queries': 'off',
+        'testing-library/render-result-naming-convention': 'off',
+        'testing-library/prefer-explicit-assert': [
+          'warn',
+          {
+            assertion: 'toBeInTheDocument',
+            includeFindQueries: true,
+          },
+        ],
```
shawyu-okta, line 327: I don't know how I ended up here, got carried away in a yak shave. Result is I refactored and updated some eslint configs. I can back these out if we think it's too much or is distracting.

src/v3/src/transformer/layout/googleAuthenticator/transformGoogleAuthenticatorEnroll.test.ts
```diff
@@ -17,22 +17,28 @@ import {
...
 import { transformGoogleAuthenticatorEnroll } from '.';
 
 describe('Google Authenticator Enroll Transformer Tests', () => {
   const transaction = getStubTransactionWithNextStep();
-  const widgetProps: WidgetProps = {};
-  const formBag = getStubFormBag();
+  const widgetProps: WidgetProps = {
+    eventEmitter: null as unknown as TinyEmitter,
+    widgetHooks: null as unknown as WidgetHooks,
+  };
```
shawyu-okta, line 38: Not sure what the best thing to do here is... It's throwing a TS error because these two properties are required on the type but this isn't a whole lot better. I also don't really think we _need_ to instantiate real versions of each since they aren't used in the tests. I guess I could write up some mock versions of both to be used in tests that would have spied methods to track calls if we wanted to do that.

src/v3/src/transformer/layout/googleAuthenticator/transformGoogleAuthenticatorEnroll.test.ts
```diff
@@ -98,6 +104,8 @@ describe('Google Authenticator Enroll Transformer Tests', () => {
 
   it('should add Stepper layout to UI Schema elements '
     + 'when GA Enroll params exists in Idx response', () => {
+    expect.hasAssertions();
```
shawyu-okta, line 107: eslint isn't happy that the assertions are all within the `validateForm` helper function so this helps ensure we do actually make assertions and also makes the linter happy.

glenfannin-okta (approved):
LGTM, only question I had was changing the i18n key value, will it prompt changes in selenium tests? 

Maybe not if they use selector values to interact with the element instead of the actual text values, so this may be just fine 🙏

PR #3449, opened 2023-11-01
Reenable Device Fingerprinting Tests in Identify_spec

trevoring-okta:
## Description:
- Reenables device fingerprinting tests within `Identify_spec`
- Adds `deviceFingerprintingUtils` temporarily to gen3 utils folder until `auth-js` implements device fingerprinting
- Refer to [gen2 version of utility](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/utils/DeviceFingerprinting.js) for comparison with `deviceFingerprintingUtils`. Core logic stayed the same, changes were mainly modernization of syntax and removing jQuery code.

## PR Checklist

### Issue:

- [OKTA-654456](https://oktainc.atlassian.net/browse/OKTA-654456)

### DIFF
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -27,6 +27,7 @@ import { MessageType } from '../types';
 import {
   areTransactionsEqual,
   containsMessageKey,
+  getBaseUrl,
   getErrorEventContext,
   getImmutableData,
   isConfigRecoverFlow,
@@ -40,6 +41,7 @@ import {
   triggerRegistrationErrorMessages,
 } from '../util';
 import { getEventContext } from '../util/getEventContext';
+import DeviceFingerprintingUtils from 'src/util/deviceFingerprintingUtils';
 
 type OnSubmitHandlerOptions = {
   includeData?: boolean;
@@ -64,7 +66,7 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     setStepToRender,
     widgetProps,
   } = useWidgetContext();
-  const { eventEmitter, widgetHooks } = widgetProps;
+  const { eventEmitter, widgetHooks, features } = widgetProps;
 
   return useCallback(async (options: OnSubmitHandlerOptions) => {
     setLoading(true);
@@ -192,10 +194,20 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
       }
     }
     // Login flows that mimic step up (moving forward in login pipeline) via internal api calls,
-    // need to clear stored stateHandles.
+    // need to clear stored stateHandles. 
     // This way the flow can maintain the latest state handle. For eg. Device probe calls
     if (step === IDX_STEP.CANCEL_TRANSACTION) {
       SessionStorage.removeStateHandle();
+    }
+    if (step === IDX_STEP.IDENTIFY && features?.deviceFingerprinting) {
+      const baseUrl = getBaseUrl(widgetProps);
+      if (baseUrl) {
+        // function should catch any errors and return undefined to avoid crashing here
+        const fingerprint = await DeviceFingerprintingUtils.generateDeviceFingerprint(baseUrl);
+        if (fingerprint) {
+          authClient.http.setRequestHeader('X-Device-Fingerprint', fingerprint);
+        }
+      }
     }
     setMessage(undefined);
     try {
@@ -302,6 +314,7 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     widgetProps,
     eventEmitter,
     widgetHooks,
+    features,
     setResponseError,
     setIdxTransaction,
     setIsClientTransaction,
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -203,7 +203,7 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
       const baseUrl = getBaseUrl(widgetProps);
       if (baseUrl) {
         // Proceeds with form submission even if device fingerprinting fails
-        const fingerprint = await generateDeviceFingerprint(baseUrl);
+        const fingerprint = await generateDeviceFingerprint(baseUrl).catch(() => undefined);
         if (fingerprint) {
           authClient.http.setRequestHeader('X-Device-Fingerprint', fingerprint);
         }
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -202,7 +202,7 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     if (step === IDX_STEP.IDENTIFY && features?.deviceFingerprinting) {
       const baseUrl = getBaseUrl(widgetProps);
       if (baseUrl) {
-        // function should catch any errors and return undefined to avoid crashing here
+        // Proceeds with form submission even if device fingerprinting fails
         const fingerprint = await generateDeviceFingerprint(baseUrl);
         if (fingerprint) {
           authClient.http.setRequestHeader('X-Device-Fingerprint', fingerprint);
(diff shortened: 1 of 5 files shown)


src/v3/src/util/deviceFingerprintingUtils.ts
```diff
@@ -0,0 +1,102 @@
...
+    iframe.id = 'device-fingerprint-container';
+    iframe.style.display = 'none';
+    // Create and load devicefingerprint page inside the iframe
+    iframe.src = `${oktaDomainUrl}/auth/services/devicefingerprint`;
+
+    const formElement = document.querySelector('form[data-se="o-form"]');
+    if (formElement === null) {
+      handleError('Form does not exist');
+    }
+    formElement!.appendChild(iframe);
```
trevoring-okta, line 95: Gen2 utility uses jQuery to create the `iframe`, switched to dynamically creating `iframe` and adding it to the DOM. Also, gen2 utility was passed the `form` element so `iframe` can be appended, but we do not have access to that `form` element in `useOnSubmit` so we search for the `form` using `data-se="o-form"` instead

src/v3/src/util/deviceFingerprintingUtils.ts
```diff
@@ -0,0 +1,102 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+const getUserAgent = (): string => navigator.userAgent;
+
+const isWindowsPhone = (userAgent: string): RegExpMatchArray | null => userAgent.match(/windows phone|iemobile|wpdesktop/i);
```
glenfannin-okta, line 15: Can you move this into `src/v3/src/util/browserUtils.ts`

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -197,6 +199,16 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     if (step === IDX_STEP.CANCEL_TRANSACTION) {
       SessionStorage.removeStateHandle();
     }
+    if (step === IDX_STEP.IDENTIFY && features?.deviceFingerprinting) {
+      const baseUrl = getBaseUrl(widgetProps);
+      if (baseUrl) {
+        // Proceeds with form submission even if device fingerprinting fails
+        const fingerprint = await generateDeviceFingerprint(baseUrl);
```
glenfannin-okta, line 206: Do you need to catch errors here and return an empty string (or undefined) from the catch block? I am thinking if the function rejects and throws an error, this would halt the remaining progress of this hook which I don't think should happen.

glenfannin-okta: something like: 
```suggestion
        const fingerprint = await generateDeviceFingerprint(baseUrl).catch(() => undefined);
```

src/v3/src/util/deviceFingerprintingUtils.ts
```diff
@@ -0,0 +1,102 @@
...
+export const isMessageFromCorrectSource = (iframe: HTMLIFrameElement, event: MessageEvent)
+: boolean => (
+  event.source === iframe.contentWindow
+);
+
+// NOTE: This utility is similar to the DeviceFingerprinting.js file used for V2 authentication flows.
+export const generateDeviceFingerprint = (oktaDomainUrl: string): Promise<string> => (
+  // eslint cannot detect that setTimeout() at end of function will eventually return with Error rejection
+  // eslint-disable-next-line consistent-return
+  new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
glenfannin-okta: Actually, if you just remove the `return` from all places that uses `resolve(...)` or `reject(...)`. This is what is causing the linting issue.

src/v3/src/util/deviceFingerprintingUtils.ts
```diff
@@ -0,0 +1,102 @@
...
+            type: 'GetFingerprint',
+          });
+        } else if (message && message.type === 'FingerprintAvailable') {
+          removeIframe();
+          resolve(message.fingerprint);
+        } else {
+          handleError('No data');
+        }
+      } catch (error) {
+        // Ignore any errors since we could get other messages too
```
glenfannin-okta, line 79: I wonder why they didn't use the resolve or handle error here? 🤔 Maybe we should?

src/v3/src/util/deviceFingerprintingUtils.ts
trevoring-okta: Do you just mean these two functions, or the entire file?

src/v3/src/hooks/useOnSubmit.ts
trevoring-okta: Oh yeah, didn't port this over from the gen2 code, thanks

src/v3/src/util/deviceFingerprintingUtils.ts
trevoring-okta: The comment seems to suggest that they might've not returned (essentially "ignored) any errors here because we could get an error from trying to parse a message that isn't a fingerprint specific error and we don't want to short-circuit from something irrelevant since the `setTimeout` at the bottom of the utility will send a default fingerprint-specific error after 2 seconds anyway?

src/v3/src/util/deviceFingerprintingUtils.ts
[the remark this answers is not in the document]
trevoring-okta: Yeah so we can do that to bypass the linting issue, but I believe the gen2 code purposefully added `return` statements specifically in the `userAgent` checks at the very beginning of the function because without the explicit `return`, the control flow will proceed past these `reject` calls and still instantiate the `iframe` but never clean it up, which we don't want to happen. Here's the gen2 code:
https://github.com/okta/okta-signin-widget/blob/acd4c0dc62b24bc9ae83f48cdda36f39c57f14df/src/v2/view-builder/utils/DeviceFingerprinting.js#L32-L36

The gen2 code would fail this eslint rule because it also isn't consistent with returning, but I thought it would be safer to keep this behavior and immediately return control and not create the `iframe` if we know right away that the `userAgent` check failed, so I ended up adding `return` to all locations where we `resolve` and `reject`, but `eslint` doesn't know about the `setTimeout` fallback logic, which will always return after 2 seconds if no other `return` is hit.

Does this make sense?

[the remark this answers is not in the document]
trevoring-okta: One alternative that I _think_ could work would be to move the `userAgent` and `windowsPhone` checks below the instantiation of the `iframe` and then use `handleError` (which depends on the `iframe` being instantiated since it calls `removeIFrame()`) instead of `return reject ...`, but then we're instantiating the `iframe` in cases where we could have detected that it should have never been instantiated in the first place

src/v3/src/util/deviceFingerprintingUtils.ts
glenfannin-okta: Oh, just the functions, sorry.

src/v3/src/util/deviceFingerprintingUtils.ts
glenfannin-okta: Ohh I see, so assuming that the message event is triggered multiple times during that timeout, we wouldnt want to return early and miss the chance of actually obtaining the fingerprint. That makes sense! Thanks!

src/v3/src/util/deviceFingerprintingUtils.ts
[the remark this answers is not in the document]
glenfannin-okta: Ah I see, could you place a single return statement after the reject/resolve callbacks? Not sure of that would appease the lint error.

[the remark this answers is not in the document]
trevoring-okta: I was able to refactor by moving the `userAgent` and `isWindowsPhone` checks outside of the overall `return new Promise...` and now the lint error isn't throwing, so this should be resolved!

glenfannin-okta (approved):

PR #3460, opened 2023-11-06
Reenable Interact_spec Clear Recovery Token

trevoring-okta:
## Description:
- Reenables test in `Interact_spec` that expected gen3 to meet parity with gen2 and clear the `recoveryToken` config option used by embedded widgets for password reset
- `recoveryToken` has to be directly deleted from `authClient.options`, currently `auth-js` does not support an alternative way to clear this option 

## PR Checklist

### Issue:

- [OKTA-654463](https://oktainc.atlassian.net/browse/OKTA-654463)
- [Bacon Link](https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&branch=ti-OKTA-654463-reenable-interact-clear-token&page=1&pageSize=6&sha=5ac690487b3b33bfb61fbd303b1d1af22ae09dba&tab=main)

### DIFF
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -194,9 +194,10 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     // Required to prevent auth-js from clearing sessionStorage and breaking interaction code flow
     payload.exchangeCodeForTokens = false;
     if (step === 'cancel') {
-      authClient?.transactionManager.clear();
+      authClient?.transactionManager.clear({ clearIdxResponse: false });
       SessionStorage.removeStateHandle();
       if (isOauth2Enabled(widgetProps)) {
+        authClient?.transactionManager.clear();
         // We have to directly delete the recoveryToken since it is set once upon authClient instantiation
         delete authClient.options.recoveryToken;
         // In this case we need to restart login flow and recreate transaction meta
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -67,6 +67,7 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     setMessage,
     setStepToRender,
     widgetProps,
+    bootstrap,
   } = useWidgetContext();
   const { eventEmitter, widgetHooks, features } = widgetProps;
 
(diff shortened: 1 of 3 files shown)


src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -197,6 +197,8 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
       authClient?.transactionManager.clear({ clearIdxResponse: false });
       SessionStorage.removeStateHandle();
       if (isOauth2Enabled(widgetProps)) {
+        // We have to directly delete the recoveryToken since it is set once upon authClient instantiation
+        delete authClient.options.recoveryToken;
```
glenfannin-okta, line 201: Can you also add this to the transform terminal transaction transformer? Above [this](https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/transformer/terminal/transformTerminalTransaction.ts#L122) line.

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -194,9 +194,11 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     // Required to prevent auth-js from clearing sessionStorage and breaking interaction code flow
     payload.exchangeCodeForTokens = false;
     if (step === 'cancel') {
-      authClient?.transactionManager.clear({ clearIdxResponse: false });
```
glenfannin-okta, line 197: I think we need to keep this as is to not cause regression, see: https://github.com/okta/okta-signin-widget/blob/master/src/v2/controllers/FormController.ts#L165

However, I think you can add another call to `clear` (without the param) within the `isOauth2Enabled` condition (this is what v2 does as well).

src/v3/src/transformer/terminal/transformTerminalTransaction.ts
```diff
@@ -119,6 +119,8 @@ const appendViewLinks = (
       cancelLink.options.href = backToSigninUri;
     } else if (isOauth2Enabled(widgetProps)) {
       cancelLink.options.onClick = async () => {
+        // We have to directly delete the recoveryToken since it is set once upon authClient instantiation
+        delete authClient?.options.recoveryToken;
```
glenfannin-okta, line 123: If you add the clear in the `useOnSubmit` hook (within the condition), can you also add that here?

src/v3/src/hooks/useOnSubmit.ts
trevoring-okta: I see, thanks. Yeah I changed this because the re-enabled test is failing in CI but not locally so I was trying to change some things around to potentially diagnose what the issue is because the error is now not the same as the original reason this test was failing (which was the `recoveryToken` not being cleared)

test/testcafe/spec/Interact_spec.js
```diff
@@ -114,6 +114,7 @@ async function setup(t, options = {}) {
     recoveryToken: options.recoveryToken,
     clientId: 'fake',
     redirectUri: 'http://doesnot-matter',
+    authScheme: 'oauth2',
     authParams: {
```
trevoring-okta, line 118: Test was failing on Bacon because Bacon's widget config didn't have explicit `authScheme: 'oauth2'` so in `useOnSubmit` the `isOAuth2Enabled` check was not passing and the `recoveryToken` clearing logic was not getting hit

glenfannin-okta (approved):

PR #3465, opened 2023-11-09
Gen3: Fix ARIA description for authenticator buttons

denysoblohin-okta:
## Problem:

The ARIA description of authenticator button is constructed with `aria-describedby` attribute from text content of 3 elements: `description`, `usageDescription` (optional) and `ctaLabel` .

### Example for "Okta Verify" button:
```
description: "Okta Verify is an authenticator app, installed on your phone, used to prove your identity"
usageDescription: "Used for access"
ctaLabel: "Set up"
```
740b23e7-09b8-4311-9e91-5b6a7545934d
Value of `aria-describedby` :
```
okta_verify_0-description okta_verify_0-usageDescription okta_verify_0-ctaLabel
````

Which results in ARIA description:
```
Okta Verify is an authenticator app, installed on your phone, used to prove your identity Used for access Set up
```
There are no punctuation separators between 3 parts which results in continous sentence speech with Voice Over or other screen reader app

## Solution:

Potential solutions:
1. Add period to all translations `oie.*.authenticator.description`, `oie.enroll.authenticator.usage.text.*`
Also need to do this for [dynamic values](https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/transformer/selectAuthenticator/utils.ts#L170) of user's email, phone, custom app names. It will also affect UI and adding period for phone/email is not desirable, see screenshot:
Screenshot 2023-11-09 at 18 03 42

2. Use [`aria-details`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-details) as suggested by @shawyu-okta . Cons: it has limited support for screen reader apps: https://a11ysupport.io/tech/aria/aria-details_attribute
3. Use [`aria-description`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description). Cons: it's in W3C draft
4. Use [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) for elements used in  `aria-describedby` to override text used by screen readers (adding period at the end if there is no punctuation already)

#4 is used in this PR

### Screenshots:

(See "Description")

Before:

Screenshot 2023-11-09 at 15 55 41

After:

Screenshot 2023-11-09 at 15 51 13

## PR Checklist

### Issue:

- [OKTA-660588](https://oktainc.atlassian.net/browse/OKTA-660588)

### CODE REVIEW

File: src/v3/src/util/makeAriaLabel.ts [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+};
+
+export const makeAriaLabel = (str: string) => {
+  // If string has no punctuation at the end, add period. Otherwise return undefined.
+  // This util is useful to add `aria-label` for element with a text content
+  //  that is not a valid sentence.
+  // When ids of such elements are used in `aria-describedby`, punctuation
+  //  is necessary to separate texts.
+  if (!endsWithPunctuation(str)) {
+    return `${str.trim()}.`;
```
lesterchoi-okta, line 32: We may have to adjust this for RTL languages

src/v3/src/util/makeAriaLabel.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+  return undefined;
```
lesterchoi-okta, line 34: You should return the original string

src/v3/src/util/makeAriaLabel.ts
```diff
@@ -0,0 +1,35 @@
...
+  // Need to support Unicode punctuation characters.
+
+  // This regular expression checks that last char is any kind of punctuation
+  //  except opening/closing quotes, connector (underscore), opening bracket
+  const matchRes = str.trim().match(/(\p{Po}|\p{Pd}|\p{Pe})$/mu);
+  // Exclude some punctuation chars that are not used to separate parts of sentence
+  return matchRes && !['"', '\''].includes(matchRes[0]);
+};
+
+export const makeAriaLabel = (str: string) => {
```
lesterchoi-okta, line 25:
```suggestion
export const punctuate = (str: string) => {
```

src/v3/src/util/makeAriaLabel.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    [
+      'keeps string as-is if it has trailing whitespaces and trimmed string ends with punctuation',
+      'We sent you a verification email. ',
+      undefined,
+    ],
+    [
+      'supports Unicode punctuation chars',
+      '请求无法完成。',
+      undefined,
+    ],
```
lesterchoi-okta, line 86: This case is pretty interesting. What if the original string didn't have punctuation?

```suggestion
    [
      'supports Unicode punctuation chars',
      '请求无法完成。',
      undefined,
    ],
    [
      'adds Unicode punctuation chars',
      '请求无法完成',
      '请求无法完成。',
    ],
```

lesterchoi-okta (approved):

test/testcafe/framework/page-objects/BasePageObject.js
```diff
@@ -285,4 +285,15 @@ export default class BasePageObject {
   isVisible() {
     return this.form.el.visible;
   }
+
+  async getAriaDescription(el) {
```
shawyu-okta, line 289: Might be a worthwhile place to put a comment describing why this method formats the returned value the way that it does (i.e. this is how screenreaders handle multiple elements in the aria-description by concatenating the values with a space.

src/v3/test/integration/__snapshots__/authenticator-verification-select-authenticator.test.tsx.snap
```diff
@@ -619,6 +620,7 @@ exports[`authenticator-verification-select-authenticator renders form 1`] = `
                               Phone
                             </h3>
                             <p
+                              aria-label="+1 XXX-XXX-5310."
```
shawyu-okta, line 623: Is it OK if the phone number on the label here gets the period at the end, too? Want to make sure that screenreaders read this correctly and doesn't end up turning it into a decimal or something.

src/v3/src/util/makeAriaLabel.test.ts
denysoblohin-okta: Then period (full stop ".", but not Chinese full stop "。") will be added to the end

src/v3/src/util/makeAriaLabel.ts
denysoblohin-okta: There is no issue with adding period to the end of RTL text.
Screenshot 2023-11-10 at 18 05 19

denysoblohin-okta:
@lesterchoi-okta @shawyu-okta 
As for RTL text, I've noticed an issue with a phone number in authenticator button description. Not related to this PR.

Screenshot 2023-11-10 at 18 15 37

Description is rendered here: https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx#L169
As description is just [a string](https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/transformer/selectAuthenticator/utils.ts#L170), I think it can be fixed by adding [`\u200E`  LTR mark](https://en.wikipedia.org/wiki/Left-to-right_mark) (similar to `&lrm;`) to the beginning of string.
(Similar to https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/transformer/phone/phoneEnrollmentCodeTransformer.ts#L61)
What do you think?

lesterchoi-okta (approved):
Thanks for taking care of RTL and full-stop punctuation

src/v3/src/transformer/selectAuthenticator/utils.ts
```diff
@@ -251,13 +251,17 @@ const formatAuthenticatorOptions = (
...
+        AUTHENTICATOR_KEY.PHONE,
+      ];
       const authenticator = option.relatesTo;
 
       return {
         type: 'AuthenticatorButton',
         label: option.label,
         id: `auth_btn_${authenticatorKey}_${enrollmentId || id}`,
         noTranslate: !isEnroll && AUTHENTICATORS_WITH_NO_TRANSLATE_CLASS.includes(authenticatorKey),
+        dir: !isEnroll && AUTHENTICATORS_WITH_LTR_DESCRIPTION.includes(authenticatorKey) ? 'ltr' : undefined,
```
shawyu-okta, line 264: FYI there is a `uischema/setLtrFields` transformer but it may not make sense. Just wanted to make sure you were aware of it and considered it.

shawyu-okta (approved):

glenfannin-okta (commented):
In AuthenticatorButton.tsx, I think you may also want to add the punctuate function to the nickname element.

PR #3468, opened 2023-11-14
Fix issue with CustomEvent in IE11

denysoblohin-okta:
## Description:

PR https://github.com/okta/okta-signin-widget/pull/3405 have introduced regression for IE11 for Gen1 if `features.passwordlessAuth` is enabled

### Bug description

Issue  with function [`createCustomEvent`](https://github.com/okta/okta-signin-widget/blob/16e7a4273858255e425c212ef2424fdf74a23b54/src/util/loc.ts#L132C1-L132C1) called from [`emitL10nError`](https://github.com/okta/okta-signin-widget/blob/16e7a4273858255e425c212ef2424fdf74a23b54/src/util/loc.ts#L120) (line 133):
https://github.com/okta/okta-signin-widget/blob/16e7a4273858255e425c212ef2424fdf74a23b54/src/util/loc.ts#L133-L137
For IE11 `CustomEvent` is an object, not a function. 
So trying to call `new CustomEvent(..)` triggers error `Object doesn't support this action`:
Screenshot 2023-11-14 at 15 04 14

Fix - check that `CustomEvent` is a function before calling a constructor, as in original code from courage:
https://github.com/okta/okta-signin-widget/blob/16e7a4273858255e425c212ef2424fdf74a23b54/packages/%40okta/courage-dist/esm/src/courage/util/StringUtil.js#L90

### Why `emitL10nError` is called

 For `password` factor `FactorUtil.getFactorDescription()` is called from [`Factor`](https://github.com/okta/okta-signin-widget/blob/16e7a4273858255e425c212ef2424fdf74a23b54/src/v1/models/Factor.js#L110) model to compute `factorDescription` property.
But `password` factor has no description: 
https://github.com/okta/okta-signin-widget/blob/16e7a4273858255e425c212ef2424fdf74a23b54/src/util/FactorUtil.js#L140
So it can't be localized: https://github.com/okta/okta-signin-widget/blob/16e7a4273858255e425c212ef2424fdf74a23b54/src/util/FactorUtil.js#L363
Last line will [emit `okta-i18n-error`](https://github.com/okta/okta-signin-widget/blob/16e7a4273858255e425c212ef2424fdf74a23b54/src/util/loc.ts#L65) with empty `key` 

Proposing to not try to localize empty description: https://github.com/okta/okta-signin-widget/pull/3468/files#diff-99e80e48c63c1ecdbd265022ebb8c35d77e3033842f1e014cda08b25fa507aecR363

## PR Checklist

### Issue:

- [OKTA-667106](https://oktainc.atlassian.net/browse/OKTA-667106)

### Screenshot/Video:

Before:

https://github.com/okta/okta-signin-widget/assets/72614880/82c592d0-65cf-4fa5-aaa5-52907d581216

After:

https://github.com/okta/okta-signin-widget/assets/72614880/51cac152-635f-452c-8837-d9555f6287cd

### DIFF
--- a/src/util/loc.ts
+++ b/src/util/loc.ts
@@ -25,6 +25,13 @@ type Bundles = {
   [key in BundleName]: Bundle;
 };
 
+interface L10nErrorDetail {
+  type: 'l10n-error';
+  key: string;
+  bundleName: string;
+  reason: string;
+}
+
 declare global {
   interface Window {
     okta?: {
@@ -126,13 +133,15 @@ function emitL10nError(key: string, bundleName: string, reason: string) {
       reason: reason
     }
   });
-  document.dispatchEvent(event);
+  if (event) {
+    document.dispatchEvent(event);
+  }
 }
 
-const createCustomEvent = function (event: string, params: CustomEventInit) {
+const createCustomEvent = function (event: string, params: CustomEventInit<L10nErrorDetail>) {
   if (typeof window.CustomEvent === 'function') {
     return new CustomEvent(event, params);
-  } else {
+  } else if (!window.CustomEvent) {
     /**
      * CustomEvent polyfill for IE
      * https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#polyfill
--- a/src/util/loc.ts
+++ b/src/util/loc.ts
@@ -130,7 +130,7 @@ function emitL10nError(key: string, bundleName: string, reason: string) {
 }
 
 const createCustomEvent = function (event: string, params: CustomEventInit) {
-  if (window.CustomEvent) {
+  if (typeof window.CustomEvent === 'function') {
     return new CustomEvent(event, params);
   } else {
     /**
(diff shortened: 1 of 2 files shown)


glenfannin-okta:
Can you rebase and point this at 7.12 branch?

src/util/loc.ts
```diff
@@ -130,7 +130,7 @@ function emitL10nError(key: string, bundleName: string, reason: string) {
 }
 
 const createCustomEvent = function (event: string, params: CustomEventInit) {
-  if (window.CustomEvent) {
+  if (typeof window.CustomEvent === 'function') {
```
shawyu-okta, line 133: This behavior in IE11 is still slightly different it seems. In the older version, it checks `!window.CustomEvent` and then sets it to `IECustomEvent`, _then_ checks if `window.CustomEvent` is a function before calling it. However if in IE11 `window.CustomEvent` is an Object,, it won't polyfill since `window.CustomEvent` is defined, just is an Object.

In the new code, we check first if the type is `'function'` and polyfill when it's not. Seems like the correct behavior but it is a bit different.

lesterchoi-okta (approved):
Let's keep old behavior and get this merged

PR #3476, opened 2023-11-28
Use Odyssey 1.x icons

denysoblohin-okta:
## PR Checklist

### Issue:

- [OKTA-658927](https://oktainc.atlassian.net/browse/OKTA-658927)

### Screenshot/Video:

UI_demo_VRT


src/v3/src/components/PhoneAuthenticator/PhoneAuthenticator.tsx
```diff
@@ -111,7 +111,9 @@ const PhoneAuthenticator: UISchemaElementComponent<UISchemaElementComponentWithV
   }, [phone, fieldName]);
 
   useEffect(() => {
-    dataSchemaRef.current![fieldName].validate = validate;
+    if (dataSchemaRef.current) {
```
denysoblohin-okta, line 114: Without check it causes issues with HMR
<details>
  <summary>Error</summary>
Screenshot 2023-11-28 at 12 26 50
</details>

src/v3/src/components/LaunchAuthenticatorButton/LaunchAuthenticatorButton.tsx
```diff
@@ -82,6 +82,8 @@ const LaunchAuthenticatorButton: UISchemaElementComponent<{
       fullWidth
       onClick={handleClick}
       ref={focusRef}
+      // Fixes text overflow
+      sx={{ display: 'flex', whiteSpace: 'normal' }}
```
denysoblohin-okta, line 86: Before:
Screenshot 2023-11-28 at 11 31 07
After:
Screenshot 2023-11-28 at 11 30 58

src/v3/webpack.dev.config.ts
```diff
@@ -107,7 +107,9 @@ const devConfig: Configuration = mergeWithRules({
           loader: 'babel-loader',
           options: {
             plugins: [
-              ...process.env.IE11_COMPAT_MODE === 'true' ? [] : ['@prefresh/babel-plugin'],
+              // Disabled #3 in https://github.com/okta/okta-signin-widget/pull/3356
+              //  because of issue "Maximum call stack size exceeded"
+              // ...process.env.IE11_COMPAT_MODE === 'true' ? [] : ['@prefresh/babel-plugin'],
```
denysoblohin-okta, line 112: Fixes HMR

src/v3/test/integration/__snapshots__/email-challenge-consent.test.tsx.snap
```diff
@@ -151,42 +151,29 @@ exports[`email-challenge-consent should render form 1`] = `
                           data-se="icon-browser"
                         >
                           <svg
-                            aria-hidden="true"
```
glenfannin-okta, line 154: Is there any concern that these will no longer use the aria-hidden attr?

src/v3/src/components/Images/MobileDeviceIcon.tsx
```diff
@@ -10,29 +10,19 @@
...
-      class="siwIconFillPrimary"
-    />
-  </svg>
+  <DevicesIcon
+    sx={(theme: Theme) => ({
+      color: theme.palette.primary.main,
+      width: '16px',
+      height: '16px',
+    })}
+    titleAccess={loc('icon.title.mobileBrowser', 'login')}
```
glenfannin-okta, line 26: I see Odyssey now combines both desktop and mobile into one image, possibly a question for @alexdahl-okta; Should we create a new title to appropriately match the new icon? Or should we continue with the separate text depending if they are on Mobile vs desktop? 

When on mobile the text would read "Mobile browser" whereas on desktop it will read "Browser" 
This will be the new icon: 
image

glenfannin-okta (commented):
Overall lgtm.

src/v3/src/components/Icon/OdysseyIcons.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const ArrowLeftIcon = _ArrowLeftIcon as OdysseyIcon;
+export const ArrowRightIcon = _ArrowRightIcon as OdysseyIcon;
+export const CheckCircleFilledIcon = _CheckCircleFilledIcon as OdysseyIcon;
+export const CloseIcon = _CloseIcon as OdysseyIcon;
+export const DevicesIcon = _DevicesIcon as OdysseyIcon;
+export const GlobeIcon = _GlobeIcon as OdysseyIcon;
+export const HideIcon = _HideIcon as OdysseyIcon;
+export const InformationCircleIcon = _InformationCircleIcon as OdysseyIcon;
+export const ShowIcon = _ShowIcon as OdysseyIcon;
+export const UserIcon = _UserIcon as OdysseyIcon;
```
trevoring-okta, line 49: Can you explain how this type conversion works? I was also trying to figure out why TS was letting us directly render icons imported from `@okta/odyssey-react-mui/icons`

src/v3/src/components/Images/DeviceIcon.tsx
```diff
@@ -10,40 +10,19 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+import { Theme } from '@mui/material';
 import { FunctionComponent, h } from 'preact';
 
 import { loc } from '../../util';
+import { DevicesIcon } from '../Icon/OdysseyIcons';
 
 export const DeviceIcon: FunctionComponent = () => (
```
trevoring-okta, line 19: nit: Since we're defaulting to Odyssey whenever possible, can we have the name match the naming convention that Odyssey uses, so in this case renaming this component and file to also be `DevicesIcon` rather than `DeviceIcon`

src/v3/src/components/PasswordRequirements/Icon.tsx
```diff
@@ -25,12 +25,13 @@ const Icon: FunctionComponent<PasswordRequirementIconProps> = (
...
 ) => {
   const { status } = props;
-  const statusToIconProps = {
-    incomplete: { name: 'close', color: 'grey.500' },
-    complete: { name: 'check-circle-filled', color: 'success' },
-    info: { name: 'information-circle', color: 'info' },
+  const statusToIcon = {
+    incomplete: { component: CloseIcon, color: 'grey.500' },
+    complete: { component: CheckCircleFilledIcon, color: 'success' },
+    info: { component: InformationCircleIcon, color: 'info' },
```
trevoring-okta, line 31: Can we change the `color: .....` values to map to Odyssey 1.x tokens that we get from the `useOdysseyDesignTokens` hook? Also, right now I believe the `color: 'info'` does not exist in 1.x so the screenshots show this icon rendering as black/grey when it should actually map to `Tokens.HueBlue500` (I checked and 0.x used to map 'info' to blue 500 hue)

trevoring-okta: Actually, I'd like @glenfannin-okta and maybe @alexdahl-okta 's opinions on this, but should we map the password requirement colors to all be palette-based rather than direct hues. I know we're aiming to give customers the ability to customize as much as possible and mapping to palette tokens rather than their underlying hues allows them to customize without breaking what they define "blue", "green", etc. to be.

If you agree that we should use palette values whenever possible, then I think these status colors should map to:
```
incomplete: { component: CloseIcon, color: 'Tokens.PaletteNeutralMain' }
complete: { component: CheckCircleFilledIcon, color: Tokens.PalettePrimarySuccess },
info: { component: InformationCircleIcon, color: Tokens.PalettePrimaryMain }
```

src/v3/src/components/Images/DeviceIcon.tsx
```diff
@@ -10,40 +10,19 @@
...
-      height="1"
-      fill="#1662DD"
-      class="siwIconFillPrimary"
-    />
-  </svg>
+  <DevicesIcon
+    sx={(theme: Theme) => ({
+      color: theme.palette.primary.main,
+      width: '16px',
+      height: '16px',
```
trevoring-okta, line 24: It's probably fine to leave this style for `width` and `height` right now since that's parity with the original custom SVG icon we were building, but I think we're going to want to reassess if we should be setting `width` and `height` or not because Odyssey does not explicitly set a `width` and `height` and instead uses MUI default values of `1em`

trevoring-okta: I'll create a follow-up ticket in the epic to look at any icon sizing differences between icons that are unique to the SIW and icons that we get from Odyssey

trevoring-okta:
LGTM besides outstanding questions!

src/v3/test/integration/__snapshots__/email-challenge-consent.test.tsx.snap
denysoblohin-okta: Good catch!

Yes, there is a concern. Hiding icon title from accessibility seems desirable behaviour.

However Odyssey uses MUI's  [SvgIcon](https://mui.com/material-ui/api/svg-icon/#SvgIcon-prop-titleAccess) which sets `aria-hidden="true"` only if `titleAccess` is not passed:
https://github.com/mui/material-ui/blob/a656b90a3b616a268c7a8c6ddc88e7d536263b25/packages/mui-material/src/SvgIcon/SvgIcon.js#L107

If we [pass `titleAccess`](https://github.com/okta/okta-signin-widget/pull/3476/files#diff-794b3e864714ca56eafbdcb9160fc3c3f734f51e74f2bf1e264eec34f11aae29R26), MUI adds `<title>` with icon description inside `<svg>` (in UI it will present tooltip when we hover on icon, also description will be read by speech synthesizers like VoiceOver)
If we don't pass `titleAccess`, there will be no tooltip. 

@alexdahl-okta @lesterchoi-okta

denysoblohin-okta: Fixed

src/v3/src/components/PasswordRequirements/Icon.tsx
denysoblohin-okta: Changed to proposed tokens

src/v3/src/components/Icon/OdysseyIcons.tsx
```diff
@@ -0,0 +1,25 @@
...
+  ArrowRightIcon,
+  CheckCircleFilledIcon,
+  CloseIcon,
+  DevicesIcon,
+  GlobeIcon,
+  HideIcon,
+  InformationCircleIcon,
+  ShowIcon,
+  UserIcon,
+} from '@okta/odyssey-react-mui/icons';
```
shawyu-okta, line 25: Is there any reason we don't directly import from `@okta/odyssey-react-mui/icons` in each place we need an icon?

src/v3/webpack.dev.config.ts
shawyu-okta: What is the effect here on IE11 testing in playground?

src/v3/src/components/Images/MobileDeviceIcon.tsx
alexdahl-okta: Great question.

My hunch is to combine, and just say "browser" for both. The distinction between mobile and desktop was already subtle, and I don't think that context adds much value.

Do we know what the values for the browser context are like?

I recall that they're something like "Chrome on MAC_OS_X"... If they're likely to include the operating system, then that'd make it even safer to remove the context from the icon label.

src/v3/src/components/PasswordRequirements/Icon.tsx
alexdahl-okta: Yeah, those more semantic tokens would be great! And those sound right – the main tokens for neutral/success/primary.

(Though are we still using the info status in password requirements? 🤔 May not matter here.)

src/v3/src/components/Images/MobileDeviceIcon.tsx
glenfannin-okta: Yep! For mobile, it would say something like: `FIREFOX on Android` or `Safari on iOS`. Whereas on desktop, it would say `FIREFOX on Mac OS X`. 

That makes sense to consolidate it to browser since more context is provided in the label. 

@denysoblohin-okta Would it be possible to get rid of the `MobileDeviceIcon.tsx` and then in [here](https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/transformer/terminal/transformEmailMagicLinkOTPOnlyElements.ts#L97) you can just reference the combined DevicesIcon (which has the browser title).

src/v3/src/components/PasswordRequirements/Icon.tsx
denysoblohin-okta: @alexdahl-okta We don't use info in password requirements.
However in case if server response contain new requirement that is not supported in SIW, it will render it with info icon:
Screenshot 2023-12-01 at 11 43 57

src/v3/src/components/Icon/OdysseyIcons.tsx
denysoblohin-okta: No reason. Changed to direct imports

src/v3/webpack.dev.config.ts
denysoblohin-okta: Live HMR was not working in IE11 (despite [this note](https://github.com/preactjs/prefresh#usage-in-ie11), I couldn't make it work with IE11 in [this PR](https://github.com/okta/okta-signin-widget/pull/3356))
With `IE11_COMPAT_MODE=true`  plugins `@prefresh/babel-plugin` and `@prefresh/webpack` are disabled (lines 112 , 126).
So this change should have no effect on IE11.

Checked how playground works on IE11 and added a fix: https://github.com/okta/okta-signin-widget/pull/3476/commits/5b74801fab0a044b9afb0bd912170d40f18d48e2#diff-11ee664012ddfbbe9392744c520696a96dd8ebb553f4b64eaaeaf422a6b0ade7R97

PR #3479, opened 2023-11-29
fix: Addresses an XSS vuln on the identifier element

LazarVuckovic-okta:
## Description:

The injected HTML was not being escaped properly for the title attribute, as it was missing quotes. This allowed a malicious user to inject a script. Since the CSP headers were rolled out, this was no longer an issue as far as JS execution goes, but I addressed the HTML injection anyways.

I made sure that the test was failing without the fix in place.

## PR Checklist

### Issue:

- [OKTA-633280](https://oktainc.atlassian.net/browse/OKTA-633280)

### Screenshot/Video:

https://drive.google.com/file/d/1m_LeCaE_vBh58ZZFf998FgBsha5t9uvc/view?usp=sharing

image

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&page=1&pageSize=6&sha=705f1c263623b2cec4984c2c85062b7d1d78375d&tab=main

### DIFF
--- a/src/v2/view-builder/internals/BaseForm.ts
+++ b/src/v2/view-builder/internals/BaseForm.ts
@@ -104,7 +104,7 @@ export default Form.extend({
 
     const header = this.$el.find('[data-se="o-form-head"]');
     const identifierHTMLString = hbs('<div class="identifier-container">\
-        <span class="identifier no-translate" data-se="identifier" title={{identifier}}>{{identifier}}</span>\
+        <span class="identifier no-translate" data-se="identifier" title="{{identifier}}">{{identifier}}</span>\
       </div>')({identifier});
 
     if (header.length) {
--- a/test/testcafe/framework/page-objects/BasePageObject.js
+++ b/test/testcafe/framework/page-objects/BasePageObject.js
@@ -80,6 +80,14 @@ export default class BasePageObject {
     return this.getTextContent('[data-se="identifier"]');
   }
 
+  getIdentifierTitle() {
+    return Selector('[data-se="identifier"]').getAttribute('title');
+  }
+
+  identifierHasContenteditable() {
+    return Selector('[data-se="identifier"]').hasAttribute('contenteditable');
+  }
+
   getFormFieldLabel(field) {
     return this.form.getFormFieldLabel(field);
   }
--- a/test/testcafe/spec/ChallengeAuthenticatorPassword_spec.js
+++ b/test/testcafe/spec/ChallengeAuthenticatorPassword_spec.js
@@ -43,6 +43,14 @@ const resetPasswordSuccess = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/challenge/answer')
   .respond(xhrSSPRSuccess, 200);
 
+const xhrAuthenticatorRequiredPasswordXSS = JSON.parse(JSON.stringify(xhrAuthenticatorRequiredPassword));
+const identifierXSS = '<img contenteditable onbeforeinput=alert(document.domain) id>test';
+xhrAuthenticatorRequiredPasswordXSS.user.value.identifier = identifierXSS;
+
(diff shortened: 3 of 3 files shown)


src/v2/view-builder/internals/BaseForm.ts
```diff
@@ -104,7 +104,7 @@ export default Form.extend({
 
     const header = this.$el.find('[data-se="o-form-head"]');
     const identifierHTMLString = hbs('<div class="identifier-container">\
-        <span class="identifier no-translate" data-se="identifier" title={{identifier}}>{{identifier}}</span>\
+        <span class="identifier no-translate" data-se="identifier" title="{{identifier}}">{{identifier}}</span>\
```
lesterchoi-okta, line 107: LGTM. Handlebars doesn't auto-escape (like React)? I thought there was something with `{{evil}}` vs `{{{evil}}}`

test/testcafe/spec/ChallengeAuthenticatorPassword_spec.js
```diff
@@ -235,3 +243,11 @@ test.requestHooks(mockChallengeAuthenticatorPassword)('should show custom factor
   await t.expect(challengePasswordPage.getFactorPageHelpLinksLabel()).eql('custom factor page link');
   await t.expect(challengePasswordPage.getFactorPageHelpLink()).eql('https://acme.com/what-is-okta-autheticators');
 });
+
+test.requestHooks(mockChallengeAuthenticatorPasswordXSS)('should properly escape the identifier and avoid XSS', async t => {
+  const challengePasswordPage = await setup(t);
+  await checkA11y(t);
+  await t.expect(challengePasswordPage.getIdentifier()).eql(identifierXSS);
+  await t.expect(challengePasswordPage.identifierHasContenteditable()).eql(false);
```
lesterchoi-okta, line 251: Clever. I assume you turned it from red to green to confirm the vuln existed in the first place?

test/testcafe/framework/page-objects/BasePageObject.js
```diff
@@ -80,6 +80,14 @@ export default class BasePageObject {
...
   }
 
+  getIdentifierTitle() {
+    return Selector('[data-se="identifier"]').getAttribute('title');
+  }
+
+  identifierHasContenteditable() {
+    return Selector('[data-se="identifier"]').hasAttribute('contenteditable');
+  }
+
```
lesterchoi-okta, line 90: Could you move these to a more specific PageObject? `BasePageObject` is pretty broad.

lesterchoi-okta (changes requested):
See comment re; moving methods from `BasePageObject`. Otherwise LGTM

src/v2/view-builder/internals/BaseForm.ts
LazarVuckovic-okta: It does escape with the double curly braces, which we were doing and it worked fine with the identifier being passed into the span content, but it wasn't working well with the attribute, without it being in quotes.
The triple curly braces syntax is the evil one, that does not escape :)

test/testcafe/framework/page-objects/BasePageObject.js
LazarVuckovic-okta: Sure, I can move it into `ChallengePasswordPageObject`. The only reason why I put it here was because the `getIdentifier` method was here, so I wanted it to be grouped together.

test/testcafe/spec/ChallengeAuthenticatorPassword_spec.js
LazarVuckovic-okta: Thank you. Yes, I made sure the test was failing first and then applied the fix to verify it was good.

lesterchoi-okta (approved):
Can fast-follow to fix page object change

jakebacher2-okta (approved):

PR #3482, opened 2023-12-07
Use OdysseyProvider with cspNonce and stylisPlugins instead of CustomPluginsOdysseyCacheProvider

denysoblohin-okta:
## Description:

- Use `OdysseyProvider` with `nonce` and `stylisPlugins`
- Removed `CustomPluginsOdysseyCacheProvider`

Some tweaks for flacky jest tests:
- Added `asyncUtilTimeout: 5 * 1000,` ([timeout for `waitFor`](https://testing-library.com/docs/dom-testing-library/api-configuration/#asyncutiltimeout), default is 1000) in jest config
- added `await waitFor(...)` after setup and after user clicks

## PR Checklist

### Issue:

- [OKTA-670666](https://oktainc.atlassian.net/browse/OKTA-670666)


src/v3/test/integration/authenticator-verification-email.test.tsx
```diff
@@ -167,7 +167,7 @@ describe('Email authenticator verification when email magic link = undefined', (
         },
         widgetOptions: { features: { autoFocus: true } },
       });
-      await findByText(/Verify with your email/);
+      await waitFor(() => findByText(/Verify with your email/));
```
lesterchoi-okta, line 170: Does changing the global `asyncUtilTimeout` config not affect `findBy...` queries?

> findBy methods are a combination of getBy* queries and waitFor. They accept the waitFor options as the last argument (i.e. await screen.findByText('text', queryOptions, waitForOptions))
https://testing-library.com/docs/queries/about/#types-of-queries

denysoblohin-okta: You are right. Reverted changes like this.

trevoring-okta:
Generally LGTM, do you have a guess why the switch to OdysseyProvider caused these longer rendering times?

src/v3/jest.setup.js
```diff
@@ -31,6 +31,7 @@ const { configure } = require('@testing-library/preact');
 
 configure({
   testIdAttribute: 'data-se',
+  asyncUtilTimeout: 5 * 1000,
```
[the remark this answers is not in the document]
denysoblohin-okta: Reverted to `5000` because of lint error
```
 34:22  error  Parsing error: Identifier directly after number
```

denysoblohin-okta:
@trevoring-okta 
> do you have a guess why the switch to OdysseyProvider caused these longer rendering times?

I think I know the reason. 
With `CustomPluginsOdysseyCacheProvider` (before this PR) `stylisPlugins` actually were not working.
You can check it by inspecting `<style>` tags in DOM - they do not contain styles like `#okta-sign-in:not([dir="rtl"]) ...` and `[dir="rtl"]  ...`
It happens because [OdysseyProvider](https://github.com/okta/okta-signin-widget/blob/odyssey-1.x/src/v3/src/components/Widget/index.tsx#L507) creates its own `OdysseyCacheProvider` but without `stylisPlugins` (before this PR) and overrides previously created one.

Without stylis plugins tests run just faster.

BTW Odyssey 1.9.1 has a bug regarding `stylisPlugins` prop, need to update to 1.9.2
https://github.com/okta/odyssey/pull/2066/files

trevoring-okta:
> Without stylis plugins tests run just faster (there are ~2x less styles needs to be parsed by jsdom).
> 
> BTW Odyssey 1.9.1 has a bug regarding `stylisPlugins` prop, need to update to 1.9.2 https://github.com/okta/odyssey/pull/2066/files

Ahhh okay that makes a lot of sense thanks for looking into that. Np on the bump to 1.9.2, I figure we're going to continue bumping the package for a while since they're releasing at a very fast cadence

trevoring-okta (approved):

PR #3484, opened 2023-12-08
Use Callout instead of Alert

denysoblohin-okta:
## Description:

- Use `Callout ` from Odyssey 1.x instead of legacy `Alert`
- Fixed tests, because new `Callout ` does not support classes (`infobox-*`) and has different DOM structure: it [adds visually hidden span](https://github.com/okta/odyssey/blob/main/packages/odyssey-react-mui/src/Callout.tsx#L54) with text like "error:" (severity) which is not visible but is included in parent's `innerText`

## PR Checklist

### Issue:

- [OKTA-642557](https://oktainc.atlassian.net/browse/OKTA-642557)

### DIFF
--- a/test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
+++ b/test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
@@ -1,6 +1,5 @@
 import BasePageObject from './BasePageObject';
-import {Selector} from 'testcafe';
-import { userVariables } from 'testcafe';
+import { Selector, userVariables } from 'testcafe';
 
 const USER_CODE_FIELD = 'userCode';
 
@@ -72,14 +71,14 @@ export default class DeviceCodeActivatePageObject extends BasePageObject {
 
   isTerminalSuccessIconPresent() {
     if (userVariables.gen3) {
-      return this.form.elementExist('.infobox-success');
+      return this.form.getAlertBox().hasClass('MuiAlert-calloutSuccess');
     }
     return this.form.getElement('.device-code-terminal--icon.success-24-green').exists;
   }
 
   isTerminalErrorIconPresent() {
     if (userVariables.gen3) {
-      return this.form.elementExist('.infobox-error');
+      return this.form.getAlertBox().hasClass('MuiAlert-calloutError');
     }
     return this.form.getElement('.device-code-terminal--icon.error-24-red').exists;
   }
--- a/test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
+++ b/test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
@@ -72,14 +72,14 @@ export default class DeviceCodeActivatePageObject extends BasePageObject {
 
   isTerminalSuccessIconPresent() {
     if (userVariables.gen3) {
-      return this.form.elementExist('.infobox-success');
+      return this.form.elementExist('[data-se="infobox-success"]');
     }
     return this.form.getElement('.device-code-terminal--icon.success-24-green').exists;
   }
 
   isTerminalErrorIconPresent() {
     if (userVariables.gen3) {
-      return this.form.elementExist('.infobox-error');
+      return this.form.elementExist('[data-se="infobox-error"]');
     }
     return this.form.getElement('.device-code-terminal--icon.error-24-red').exists;
   }
--- a/test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
+++ b/test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
@@ -1,5 +1,6 @@
 import BasePageObject from './BasePageObject';
-import { Selector, userVariables } from 'testcafe';
+import {Selector} from 'testcafe';
+import { userVariables } from 'testcafe';
 
 const USER_CODE_FIELD = 'userCode';
 
@@ -71,14 +72,14 @@ export default class DeviceCodeActivatePageObject extends BasePageObject {
 
   isTerminalSuccessIconPresent() {
     if (userVariables.gen3) {
-      return this.form.getAlertBox().hasClass('MuiAlert-calloutSuccess');
+      return this.form.elementExist('.infobox-success');
     }
     return this.form.getElement('.device-code-terminal--icon.success-24-green').exists;
   }
 
   isTerminalErrorIconPresent() {
     if (userVariables.gen3) {
-      return this.form.getAlertBox().hasClass('MuiAlert-calloutError');
+      return this.form.elementExist('.infobox-error');
     }
     return this.form.getElement('.device-code-terminal--icon.error-24-red').exists;
   }
--- a/test/testcafe/framework/page-objects/EnrollOktaVerifyPageObject.js
+++ b/test/testcafe/framework/page-objects/EnrollOktaVerifyPageObject.js
@@ -3,7 +3,7 @@ import { userVariables, Selector } from 'testcafe';
 import { within } from '@testing-library/testcafe';
 
 const FORM_INFOBOX_ERROR_TITLE = '[data-se="o-form-error-container"] [data-se="callout"] > h3';
-const FORM_INFOBOX_ERROR_TITLE_V3 = '[data-se="callout"] div > h2';
+const FORM_INFOBOX_ERROR_TITLE_V3 = '[data-se="callout"] h2';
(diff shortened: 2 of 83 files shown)


test/testcafe/framework/page-objects/components/CalloutObject.js
```diff
@@ -13,22 +15,24 @@ export default class CalloutObject {
...
     }
-    return hasInfoBoxErrorClass &&
+    return this.el.hasClass('infobox-error') &&
       this.el.find('[data-se="icon"]').hasClass('error-16');
   }
 
   getTextContent() {
     if (userVariables.gen3) {
-      return this.el.textContent;
+      return this.el.find(CALLOUT_CONTENT_V3).textContent;
```
glenfannin-okta, line 35: Is the `find` necessary here? The logic in the constructor [here](https://github.com/okta/okta-signin-widget/pull/3484/files#diff-4bcc86ed383d602f2dbea0af4ee097d300d074ec0c3b90f1319dedd8266197a9R10-R14) assigns the `this.el` to the current alert/callout, so I would think it should already have reference to the correct element without the need to find?

denysoblohin-okta: Yes. The idea here is to filter out non-visible `span` inside callout (see span with text "warning" in the screenshot). So instead of taking `textContent ` for whole alert, we need to get it for content div only. 

Screenshot 2023-12-12 at 17 37 23

```javascript
document.querySelector('[role="alert"]').textContent
// outputs: 'warningReminder contentResend'
```

test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
```diff
@@ -72,14 +71,14 @@ export default class DeviceCodeActivatePageObject extends BasePageObject {
 
   isTerminalSuccessIconPresent() {
     if (userVariables.gen3) {
-      return this.form.elementExist('.infobox-success');
+      return this.form.getAlertBox().hasClass('MuiAlert-calloutSuccess');
```
trevoring-okta, line 74: I believe the `data-se="message"` on the top level `Box` that wraps the `Callout` component is not actually used in our page objects (I did a CMD+F for [data-se="message"]). If that's so, we could replace that `data-se` with the ``infobox-${messageClass.toLowerCase()}`` and then use that for all selectors where we need the callout variant

test/testcafe/framework/page-objects/components/CalloutObject.js
trevoring-okta: Just spoke to Lester about this. First off, if you trace the references it looks like this specific `getTextContent()` is barely used and doesn't specifically need `textContent`. We could use `innerText`, which we use much more frequently in our tests. 

`innerText` will get us a string of all text in the alert delimited by a `\n` character and we can update all text assertions that use `.getTextContent().eql(...)` to instead use something like `.contains(....)`. 

This will get us past the issue of having to use MUI class selectors as a way of targeting the exact parent element of the text content we want

test/testcafe/framework/page-objects/components/CalloutObject.js
```diff
@@ -13,22 +15,24 @@ export default class CalloutObject {
...
+    }
     return this.el.hasClass('infobox-warning') &&
       this.el.child('[data-se="icon"]').hasClass('warning-16');
   }
 
   isError() {
-    const hasInfoBoxErrorClass = this.el.hasClass('infobox-error');
     if (userVariables.gen3) {
-      return hasInfoBoxErrorClass;
+      return this.el.hasClass('MuiAlert-calloutError');
```
trevoring-okta, line 27: Assuming my comment about `[data-se="message"]` not being used in our page objects is true, we can switch these selectors to target `[data-se="infobox-error"]`, `[data-se="infobox-warning"]`, etc.

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -262,7 +263,11 @@ export default class BaseFormObject {
   }
 
   getErrorBoxText() {
-    return this.getErrorBox().innerText;
+    if (userVariables.gen3) {
+      return this.getErrorBox().find(CALLOUT_CONTENT_V3).innerText;
```
trevoring-okta, line 267: Same idea as comment made in `CalloutObject.js`. I think we can remove the `CALLOUT_CONTENT_V3` selector and have this function and other relevant functions return all of the `innerText` in the alert component and then switch our spec assertions to use `.contains()` instead of `.eql()`

E.g. this function would become
```
if (userVariables.gen3) {
  return this.getErrorBox().innerText;
}
```

test/testcafe/framework/page-objects/components/CalloutObject.js
```diff
@@ -13,22 +15,24 @@ export default class CalloutObject {
...
     }
-    return hasInfoBoxErrorClass &&
+    return this.el.hasClass('infobox-error') &&
       this.el.find('[data-se="icon"]').hasClass('error-16');
```
trevoring-okta, line 30: This is sort of random but I'm a bit skeptical of `error-16` and `warning-16`. Are these `emotion` classes? I guess this has been working for gen2 tests but it seems like the presence of `'infobox-error` should be enough to tell us that the callout is the error variant

test/testcafe/framework/page-objects/components/CalloutObject.js
denysoblohin-okta: Ok.
It was my initial idea to fix tests by replacing `.eql()` with `.contains()`, but I thought thus tests would be less accurate.

test/testcafe/framework/page-objects/components/CalloutObject.js
denysoblohin-okta: These are not emotion classes. They are used in Gen2. `16` means width and height of icon. For Gen3 we should use `infobox-*` data-se attribute

test/testcafe/framework/page-objects/components/CalloutObject.js
denysoblohin-okta: Done. Replaced `.eql` with `.contains`
https://github.com/okta/okta-signin-widget/pull/3484/commits/9e4cb2b505c43b91257e17263bb9392dbecb3a5e

test/testcafe/framework/page-objects/DeviceCodeActivatePageObject.js
denysoblohin-okta: Done
https://github.com/okta/okta-signin-widget/pull/3484/commits/11c32491a5bdbde59e9f375e6bda4c5fd31587f5#diff-3c0545bd20e65728cb0870b895809baafc3b5da0137f9d83d738ee775f687c8dR46

test/testcafe/framework/page-objects/components/CalloutObject.js
denysoblohin-okta: Done
https://github.com/okta/okta-signin-widget/pull/3484/commits/11c32491a5bdbde59e9f375e6bda4c5fd31587f5#diff-4bcc86ed383d602f2dbea0af4ee097d300d074ec0c3b90f1319dedd8266197a9R25

test/testcafe/framework/page-objects/components/BaseFormObject.js
denysoblohin-okta: Reverted

test/testcafe/framework/page-objects/components/CalloutObject.js
trevoring-okta: Ah okay I see, thanks for the clarification

test/testcafe/framework/page-objects/EnrollOktaVerifyPageObject.js
```diff
@@ -3,7 +3,7 @@ import { userVariables, Selector } from 'testcafe';
 import { within } from '@testing-library/testcafe';
 
 const FORM_INFOBOX_ERROR_TITLE = '[data-se="o-form-error-container"] [data-se="callout"] > h3';
-const FORM_INFOBOX_ERROR_TITLE_V3 = '[data-se="callout"] > div > h2';
+const FORM_INFOBOX_ERROR_TITLE_V3 = '[data-se="callout"] div > h2';
```
trevoring-okta, line 6: nit: I think we can assume that Odyssey Callout will only ever render one `h2` element so we can probably remove the `div >`

trevoring-okta (approved):
LGTM besides one extra nit. All good to merge once we get green Bacon!

PR #3487, opened 2023-12-15
Use Tag in IdentifierContainer

denysoblohin-okta:
## Description:

- Updated Odyssey to 1.9.16
- Added `translate="no"` for `Callout` and `Tag`. For now it's being rendered as `translate="yes"` ([example in snapshots](https://github.com/okta/okta-signin-widget/pull/3487/files#diff-55b815d63b27f3d8b8ee531db22a4eca16ac367376f011d84340191275916e2aR34)) due to bug in preact https://github.com/preactjs/preact/issues/3799 until https://github.com/preactjs/preact/pull/3800 is merged
- Modified English Leaks test to support [both `.no-translate, [translate="no"]`](https://github.com/okta/okta-signin-widget/pull/3487/files#diff-20ecefc4d164e1f22564fa444b02083080727de4a7c205be486bfbacbbbd050dR11)   (until all `no-translate` classes are removed and issue with `translate="yes"` is fixed). 
No-translation content should be get with [using `innerText`](https://github.com/okta/okta-signin-widget/pull/3487/files#diff-20ecefc4d164e1f22564fa444b02083080727de4a7c205be486bfbacbbbd050dR174) instead of `textContent` - same as for [getting all visible text on page](https://github.com/okta/okta-signin-widget/pull/3487/files#diff-20ecefc4d164e1f22564fa444b02083080727de4a7c205be486bfbacbbbd050dR165), to to avoid including hidden elements

## PR Checklist

### Issue:

- [OKTA-674223](https://oktainc.atlassian.net/browse/OKTA-674223)

### Screenshot/Video:

Before:
Screenshot 2023-12-18 at 17 33 51

After:
Screenshot 2023-12-18 at 17 29 42

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&branch=d16t-okta-signin-widget-3790c27-658356cc&page=1&pageSize=6&sha=79b5070bbf78117c42ab596fe872692da49173da&tab=main


src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
```diff
@@ -35,45 +34,27 @@ const IdentifierContainer: UISchemaElementComponent<{
...
-        >
-          {identifier}
-        </Box>
-      </Box>
+      <Tag
+        icon={(
+          <Box
+            sx={{
+              width: '16px',
+              height: '16px',
```
lesterchoi-okta, line 42: Shouldn't these use design tokens?

src/v3/test/integration/__snapshots__/authenticator-email-verification-data.test.tsx.snap
```diff
@@ -111,12 +112,9 @@ exports[`authenticator-email-verification-data should render form 1`] = `
                                 User
                               </title>
                             </svg>
-                          </span>
+                          </div>
                           <span
-                            aria-label="tester@okta1.com"
-                            class="identifier no-translate MuiBox-root emotion-17"
-                            data-se="identifier"
-                            title="tester@okta1.com"
```
lesterchoi-okta, line 119: Are these attributes ok to remove? `data-se`, `aria-label`, `title` seem important

lesterchoi-okta (commented):
Question about missing attributes in snapshots after change

src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
denysoblohin-okta: There is the dedicated ticket to review icon sizing: [OKTA-671459](https://oktainc.atlassian.net/browse/OKTA-671459)

src/v3/test/integration/__snapshots__/authenticator-email-verification-data.test.tsx.snap
glenfannin-okta: I believe there are selenium tests within the monolith that uses this data-se attribute (we should probably be periodically running the downstreams against this topic branch to avoid a delay once this is completed. 

Also, re: the title, this was used because if the user ID is extremely long, we have css that truncates it and adds an ellipsis, the title was for visual users that allows them to hover the container to see the full user ID. For that reason I think we may need to keep this unless design says otherwise (it was originally a directive from UX to add this).

src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
```diff
@@ -35,45 +34,27 @@ const IdentifierContainer: UISchemaElementComponent<{
...
-        flexGrow="0"
-        paddingX={4}
-        paddingY={2}
-        sx={{
-          borderRadius: '36px',
-          backgroundColor: tokens.HueNeutral50,
-          textOverflow: 'ellipsis',
-          whiteSpace: 'nowrap',
-          overflow: 'hidden',
-        }}
```
glenfannin-okta, line 50: Was there a reason this css was removed (specifically the textOverflow, whitespace and overflow properties)? This is to truncate long user IDs to avoid the container from taking up multiple rows (this was a UX decision).

src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
```diff
@@ -23,7 +23,6 @@ const IdentifierContainer: UISchemaElementComponent<{
   const { options: { identifier } } = uischema;
 
   const mainContainerClasses = classNames('identifier-container');
-  const identifierSpanClasses = classNames('identifier', 'no-translate');
```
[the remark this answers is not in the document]
trevoring-okta: We've been having ongoing discussions with Odyssey about how we can move `no-translate` to instead use the `translate="no"` global HTML attribute (this is what I talked to you about for the checkbox component, @glenfannin-okta). I have an open PR to get this merged into Odyssey right now, which would unblock this for the identifier container. Maybe we should wait until that gets merged and then we can bump to latest version of Odyssey

src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
trevoring-okta: I think this is because Odyssey basically does not expose `sx` on any non-Box component. I've been anticipating that many of our components have built-in one-offs from product/UX like this one that won't play well with Odyssey. I'm not sure if there's another way to achieve this fix so if it's deemed critical we might have to flag this component as one that cannot be converted to use Odyssey's `Tag`.

src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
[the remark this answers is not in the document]
glenfannin-okta: Ahh, that makes sense, this actually reminds me, if you guys change how things are tagged as "no to be translated" do you also plan to update the corresponding Selenium tests in the monolith that looks for that class on the SIW elements? I just want to make sure we have accounted for those tests as well.

src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
glenfannin-okta: Ah I see, could we just wrap the text in a box component (span) and apply this styling? Would that cause any issues?

src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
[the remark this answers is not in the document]
trevoring-okta: Yeah I believe we had this discussion sometime when you were on PTO but @lesterchoi-okta was doing an investigation into the selectors that are used in monolith and whether it will be feasible to convert all of them over to semantic selectors or `data-se`. I'll start a wiki right now to keep track of selectors that we know we've broken or removed as part of the migration

src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
trevoring-okta: Yeah I just tested this and I think we can do this. @denysoblohin-okta it seems to me like we can move this text overflow styling to the root `Box` element and also remove some of the styles on that root element because they're not needed now that we have `Tag` styling. Note: we'll have to continue using the normal MUI `Box` that we import from `@okta/odyssey-react-mui-legacy` because OD 1.x `Box` doesn't support `title`

I'm thinking something like this:
```
<Box
  title={identifier}
  sx={{
    textOverflow: 'ellipsis',
    whiteSpace: 'nowrap',
    overflow: 'hidden',
  }}
  data-se="identifier-container"
>
  <Tag
    icon={(
      <Box
        sx={{
          width: '16px',
          height: '16px',
        }}
      >
        <UserIcon
          titleAccess="User"
          sx={{
            color: tokens.PalettePrimaryMain,
            width: '16px',
            height: '16px',
          }}
        />
      </Box>
    )}
    label={identifier}
    testId="identifier"
  />
</Box>
```

src/v3/test/integration/__snapshots__/authenticator-email-verification-data.test.tsx.snap
denysoblohin-okta: As for `title` attribute, I can add it to the outer `Box` instead of `Tag` (which doesn't support it). Then it would show tooltip as before.

As for `data-se="identifier"`, it's applied to outer `div` (line 92)
I hope monolith selenium tests don't use `.identifier` class.
And I hope it apply `innerText` instead of `textContent` against element with `data-se="identifier"`. Otherwise we can remove `titleAccess="User"` from `<UserIcon>` to delete `<title>User</title>` inside icon

src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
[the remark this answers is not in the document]
denysoblohin-okta: As a quick fix for enleaks tests in SIW, I've modified spec by adding `.identifier-container span` in addition to `.no-translate` as selector for text without translation
https://github.com/okta/okta-signin-widget/pull/3487/commits/60e0ccb46256f738c3fc1acbfc4fda9f093b1149#diff-20ecefc4d164e1f22564fa444b02083080727de4a7c205be486bfbacbbbd050d
But it's bad workaround, should remove it.

A better way would be to add class `no-translate` to the outer `Box` (the one with `identifier-container` class). But the problem is that its `.textContent` would include `User` from `<title>User</title>` inside icon.
Any reason why `.textContent` is used instead of `.innerText` ?
https://github.com/okta/okta-signin-widget/blob/b67d788ed46fad0d11c981895969243103fed752/test/testcafe/spec-en-leaks/EnglishLeaks_spec.js#L173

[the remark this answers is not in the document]
trevoring-okta: I'm not 100% sure how our English leak tests work but I'm assuming we use `.textContent` to also exclude non-visible text that appears in the DOM but not on the page because I believe otherwise our English leak tests will flag non-visible text like `<title>User</title>` as a leak?

TLDR: I think you should be safe to move the `.no-translate` to the outer `Box` for now and then we can remove it once https://github.com/okta/odyssey/pull/2074 is merged in to Odyssey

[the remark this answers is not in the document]
denysoblohin-okta: Actually `.textContent` **includes** non-visible text (like content of `title` tag) while `.innerText` does not include.

So if we mix `.textContent`  and `.inerText`
https://github.com/okta/okta-signin-widget/blob/b67d788ed46fad0d11c981895969243103fed752/test/testcafe/spec-en-leaks/EnglishLeaks_spec.js#L164
https://github.com/okta/okta-signin-widget/blob/b67d788ed46fad0d11c981895969243103fed752/test/testcafe/spec-en-leaks/EnglishLeaks_spec.js#L173
it cause problems with hidden text

test/testcafe/spec-en-leaks/EnglishLeaks_spec.js
```diff
@@ -8,6 +8,11 @@ const path = require('path');
 const PLAYGROUND = path.resolve(__dirname, '../../../playground');
 const mocksFolder = `${PLAYGROUND}/mocks/data/idp/idx`;
 const mocksOauth2Folder = `${PLAYGROUND}/mocks/data/oauth2`;
+const noTranslateSelector = [
+  '.no-translate',
+  // hidden text for screen readers
```
denysoblohin-okta, line 13: Should be removed after merging of https://github.com/okta/odyssey/pull/2074/files#diff-ba66395dd166ceaa1fb9be3e394dc3b399a088fdf32cbe51311b64f83f99ab8dR61

denysoblohin-okta:
Checked monolith downstream tests. 
Looks like this PR hasn't introduced new failures.
Downstream for this PR: https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&branch=d16t-okta-signin-widget-3790c27-658356cc&page=1&pageSize=6&sha=79b5070bbf78117c42ab596fe872692da49173da&tab=main
Downstream for base commit from `odyssey-1.x` branch: https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&branch=d16t-okta-signin-widget-56e48ef-6583ff9f&page=1&pageSize=6&sha=3ecf5a9647cb66857cc850b45c3f0cf7ff43eff2&tab=main

src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
```diff
@@ -33,47 +31,32 @@ const IdentifierContainer: UISchemaElementComponent<{
       alignItems="center"
       marginBlockEnd={4}
       maxWidth={1}
-      className={mainContainerClasses}
+      className="no-translate"
```
trevoring-okta, line 34: Can we switch this to `translate="no"` since that should be possible with a MUI `Box` and then remove the `className`

trevoring-okta (approved):
LGTM besides one question about removing `.no-translate` from MUI `Box` in `IdentifierContainer` component

src/v3/src/components/IdentifierContainer/IdentifierContainer.tsx
trevoring-okta: That's fair, so you already tested and removing this `.no-translate` will break english leak CI? I might have a potential fix for the Preact issue coming

denysoblohin-okta: Actually for now english leaks tests are being run in CI only for v2, not for v3 
I see test failures when running manually `yarn test:testcafe-run-en-leaks` (with `yarn workspace v3 dev`)
So I think it's fine to remove classes `no-translate` already and probably enable english leaks tests for v3 in the future

PR #3492, opened 2023-12-20
Support BYOL and sync languageCode with OD Provider

trevoring-okta:
## Description:
- Passes in SIW-defined `LanguageCode` to `OdysseyProvider`'s `languageCode` prop so strings in Odyssey components can be translated to the same language as the SIW
- Constructs translation overrides object to pass into `OdysseyProvider`'s `translationOverrides` prop. This allows users to override Odyssey strings from the widget config in the same way that they can for SIW strings
- Creates translation overrides object with a generic `string` type to support BYOL for languages that aren't natively supported by Odyssey (e.g. RTL languages like 'ar')

## PR Checklist

### Issue:

- [OKTA-666640](https://oktainc.atlassian.net/browse/OKTA-666640)

### Demo video
https://github.com/okta/okta-signin-widget/assets/106110543/a9379473-bf52-4a5e-a329-3d545d52932e

First shows Odyssey `TextField` "Optional" label overridden by locally-hosted Arabic translation file to simulate BYOL, then shows that string being overridden by user-defined string from widget config. This mirrors the order of precedence in SIW string overrides.


src/v3/src/components/Widget/index.tsx
```diff
@@ -134,8 +135,13 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
   const languageCode = getLanguageCode(widgetProps);
   const languageDirection = getLanguageDirection(languageCode);
   const { stateHandle, unsetStateHandle } = useStateHandle(widgetProps);
+  // Odyssey language codes use '_' instead of '-' (e.g. zh-CN -> zh_CN)
+  const odyLanguageCode: string = languageCode.replace('-', '_');
+  // Only pick Odyssey translation keys from translation bundle
+  const odyTranslationOverrides: TranslationOverrides<string> = {
+    [odyLanguageCode]: pick(Bundles.login, odysseyI18nResourceKeysList)
+  } as TranslationOverrides<string>;
```
trevoring-okta, line 143: Was debating whether these should be moved to a utility file, @glenfannin-okta do you have any ideas on whether these would fit better in a different file?

Also, what's our stance on `lodash` and `underscore`? I could implement pick natively if needed, see: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore?tab=readme-ov-file#_pick

src/v3/screenshots/base/UI_demo/UI_demo_RTL_VRT.png
trevoring-okta, line 1: Diff was just very minor spacing changes that I believe must be coming from us pulling in some styling changes in the Odyssey bump to `1.9.9`

src/v3/src/components/Widget/index.tsx
```diff
@@ -134,8 +135,13 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
   const languageCode = getLanguageCode(widgetProps);
   const languageDirection = getLanguageDirection(languageCode);
   const { stateHandle, unsetStateHandle } = useStateHandle(widgetProps);
+  // Odyssey language codes use '_' instead of '-' (e.g. zh-CN -> zh_CN)
+  const odyLanguageCode: string = languageCode.replace('-', '_');
+  // Only pick Odyssey translation keys from translation bundle
+  const odyTranslationOverrides: TranslationOverrides<string> = {
+    [odyLanguageCode]: pick(Bundles.login, odysseyI18nResourceKeysList),
+  } as TranslationOverrides<string>;
```
glenfannin-okta, line 143: I see this is working as expected, however, I'm just curious, does this work when load a language that is not cached in the browser session properties? 

The reason I ask is because of [this](https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/components/Widget/index.tsx#L151-L156) logic that was added to load the current selected language, I remember an issue that was found before this was added that caused english to display instead of their preferred language because english is the default in the Bundle.login property. 

Wondering if this property can be set as state instead and then passed into the provider, then it can be updated in the initLang callback using the pick function to ensuring it has the correct overrides (since the Bundles.login gets updated after `loadLanguage()` is called)?

src/v3/src/components/Widget/index.tsx
glenfannin-okta: Left a comment on the `odyTranslationOverrides`, but re: the util, I usually only lean into creating a util if it is going to be re-used or if there is enough logic to warrant a separate dedicated file. This seems pretty straightforward and may not warrant a separate util file.

Re: lodash, initially we were using it alot, then we began to only use when absolutely necessary due to the size of lodash and us not having tree shaking. I think in this case, if there is simple logic that can be created to avoid using lodash functions we should go for that instead.

src/v3/src/components/Widget/index.tsx
trevoring-okta: Ah good callout, I'll look into this fix. Thanks!

src/v3/src/components/Widget/index.tsx
```diff
@@ -158,6 +166,18 @@ export const Widget: FunctionComponent<WidgetProps> = (widgetProps) => {
...
+      setOdyTranslationOverrides({
+        [odyLanguageCode]: odysseyI18nResourceKeysList
+          .reduce((overrides: Partial<OdysseyI18nResourceKeys>,
+            key: typeof odysseyI18nResourceKeysList[number]) => {
+            const updatedOverrides = { ...overrides };
+            if (Bundles.login && Object.prototype.hasOwnProperty.call(Bundles.login, key)) {
+              updatedOverrides[key] = (Bundles.login as Partial<OdysseyI18nResourceKeys>)[key];
+            }
+            return updatedOverrides;
+          }, {}),
```
glenfannin-okta, line 179: nit: 

I know I said I usually would only create a util for something that is re-used or complex, but this, at first glance, seems like it would be beneficial to have a util function instead of having the logic here 😅 (I'm sorry). Not required though, I leave it totally up to you 🙏

glenfannin-okta (approved):
LGTM! Nice job with this!

src/v3/src/components/Widget/index.tsx
trevoring-okta: Yeah agreed that it seems like a bit too much business logic for this file. Could this fit in `languageUtils.ts`?

PR #3392, opened 2023-09-28
Show email as a description for email authenticator button

denysoblohin-okta:
## Description:

If masked email is present in IDX response in `authenticatorEnrollments` (`profile.email` for email authenticator, same as `profile.phoneNumber` for phone authenticator), show it as description

## PR Checklist

- [x] Have you verified the basic functionality for this change?
- [x] Did you add tests, as appropriate, following our [Automated Test guidelines](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2676497890/Automated+Testing+in+the+Signin+Widget)?
- [x] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?
- [ ] Did you verify the change by running [downstream monolith artifact](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/102897979/Sign-in+Widget+Development#Sign-inWidgetDevelopment-Instructionstocreateandrunthedownstreamartifact(d16t))? (Provide link to build below)
- [ ] Does this PR include noticeable changes to the UI? (If yes, attach screenshots/video below)

### Issue:

- [OKTA-620984](https://oktainc.atlassian.net/browse/OKTA-620984)

### Screenshot/Video:

Before:
before

After:
after

Gen 2:
Screenshot 2023-10-09 at 17 42 38

PR #3499, opened 2024-01-05
Gen3: Update Authenticator List styles

denysoblohin-okta:
## PR Checklist

### Issue:

- [OKTA-506882](https://oktainc.atlassian.net/browse/OKTA-506882)

### Screenshot/Video:

Before / After:

Screenshot 2024-01-09 at 19 02 31
Screenshot 2024-01-09 at 19 01 21

<hr>

Screenshot 2024-01-09 at 19 06 09
Screenshot 2024-01-09 at 19 07 03

<hr>

Screenshot 2024-01-09 at 19 09 00
Screenshot 2024-01-09 at 19 08 04

<hr>

Screenshot 2024-01-09 at 19 26 25
Screenshot 2024-01-09 at 19 25 34


src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -87,34 +89,42 @@ const AuthenticatorButton: UISchemaElementComponent<{
     });
   };
 
+  // eslint-disable-next-line no-nested-ternary
+  const ctaIcon = isAdditionalEnroll
+    ? <AddIcon titleAccess={ctaLabel} />
+    : languageDirection === 'rtl'
+      ? <ArrowLeftIcon titleAccess={ctaLabel} />
+      : <ArrowRightIcon titleAccess={ctaLabel} />;
```
trevoring-okta, line 97: Any strong reason why you decided to disable `no-nested-ternary` instead of using `if-else`?

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -170,9 +181,10 @@ const AuthenticatorButton: UISchemaElementComponent<{
             paragraph
             id={`${iconName}-description`}
             sx={{
-              fontSize: '.875rem',
+              fontSize: tokens.TypographySizeSubordinate,
+              color: tokens.TypographyColorSubordinate,
```
trevoring-okta, line 185: Per Alex's Figma, do we need to explicitly set `fontWeight: tokens.TypographyWeightBody` or is that already the default here for the description?

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -170,9 +181,10 @@ const AuthenticatorButton: UISchemaElementComponent<{
...
               margin: 0,
-              marginBlockEnd: '6px',
+              marginBlockStart: '6px',
```
trevoring-okta, line 187: Can we update all spacing styling (margins, padding, etc.) to use the closest `tokens.SpacingX` value? I believe we are planning on updating the Okta-hosted page to match Odyssey's expected base font size of `14px` so that would mean `tokens.Spacing1 = 4px` and `tokens.Spacing2 = 8px`. I'd say try those values out and pick the one that looks the best

See: https://github.com/okta/odyssey/blob/main/packages/odyssey-design-tokens/src/spacing.json

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -207,49 +220,69 @@ const AuthenticatorButton: UISchemaElementComponent<{
             id={`${iconName}-usageDescription`}
             textAlign="start"
             sx={{
-              fontSize: '.875rem',
+              fontSize: tokens.TypographySizeSubordinate,
+              color: tokens.TypographyColorSubordinate,
               margin: 0,
```
trevoring-okta, line 225: Following up on comment about using spacing tokens, we should ideally even update styling like this to use `tokens.Spacing0`

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -207,49 +220,69 @@ const AuthenticatorButton: UISchemaElementComponent<{
...
             }}
             data-se="authenticator-button-usage-text"
             className="authenticator-usage-text"
           >
             {usageDescription}
           </Typography>
         )}
+        {isEnroll && (
+          <Box
+            className="cta-button authenticator-button"
```
trevoring-okta, line 236: Even though we aren't updating Authenticator List components to Odyssey components, we should still try to remove all uses of `className`. It seems like this might be a big ask and this PR already has a lot of file changes, so I'm open to making a separate ticket for going through these components and removing `className` as well as updating relevant spec tests. Lmk what you think

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
trevoring-okta: Actually, I'd like @lesterchoi-okta 's thoughts on whether we want to update spacing values to use tokens as part of the migration. I could see an argument that we don't want this spacing to change based on tokens? I just looked at a Slack thread we had with Odyssey and they were thinking of keeping non-type-related spacing values as hard-coded `px`, see: https://okta.slack.com/archives/C7T2H3KNJ/p1701230964122719?thread_ts=1701193300.526959&cid=C7T2H3KNJ

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -207,49 +220,69 @@ const AuthenticatorButton: UISchemaElementComponent<{
...
+              data-se="cta-button-label"
+              className="button select-factor link-button"
+            >
+              {ctaLabel}
+            </Box>
+            {ctaIcon}
+          </Box>
+        )}
+      </Box>
+      {!isEnroll && (
```
trevoring-okta, line 267: nit: combine with the `{isEnroll && ... }` code above and make it a ternary

src/v3/src/components/Icon/OktaVerifyIcon.tsx
```diff
@@ -27,13 +27,6 @@ export const OktaVerifyIcon: FunctionComponent<IconProps> = ({
     viewBox="0 0 48 48"
   >
     <title id={name}>{description}</title>
-    <circle
-      cx="24"
-      cy="24"
-      r="24"
-      fill="#F5F5F6"
-      class="siwFillBg"
-    />
```
trevoring-okta, line 36: @alexdahl-okta In your ticket you said "Remove background from Auth Coins throughout the widget, making them just Auth Icons." I want to make sure that is still true and we don't want ANY instances of our icons to have a background anymore? E.g. this `OktaVerifyIcon` is also rendered in the `LaunchAuthenticatorButton` and would lose its background as a result of this change
Screenshot 2024-01-11 at 1 06 55 PM

src/v3/src/transformer/selectAuthenticator/utils.ts
```diff
@@ -238,33 +273,35 @@ const formatAuthenticatorOptions = (
...
+        && isAuthenticatorAlreadyEnrolled(authenticator, authenticatorEnrollments);
       const AUTHENTICATORS_WITH_METHOD_TYPE = [
         AUTHENTICATOR_KEY.ON_PREM,
         AUTHENTICATOR_KEY.OV,
         AUTHENTICATOR_KEY.RSA,
       ];
       const AUTHENTICATORS_WITH_NO_TRANSLATE_CLASS = [
         AUTHENTICATOR_KEY.PHONE,
-        AUTHENTICATOR_KEY.CUSTOM_APP,
+        AUTHENTICATOR_KEY.EMAIL,
```
trevoring-okta, line 290: Was `AUTHENTICATOR_KEY.EMAIL` just originally missing from this list or did you make a change that caused this?

src/v3/src/transformer/selectAuthenticator/utils.ts
trevoring-okta, line 1: Really good job writing clean code for all of the updates you had to make here for switching the label and description as well as adjusting the CTA label and icon depending on whether it's first enroll or additional enroll!

src/v3/src/transformer/selectAuthenticator/utils.ts
denysoblohin-okta: `AUTHENTICATOR_KEY.EMAIL` should be in the list after my changes in https://github.com/okta/okta-signin-widget/pull/3392
Otherwise english leaks tests for gen 3 will fail on mock `authenticator-verification-select-authenticator-with-email`
(For now english leaks tests  are not enabled for gen 3 in Bacon so I've tested manually)

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
denysoblohin-okta: No. Changed to if-else

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
denysoblohin-okta: It's default. But it's better to set explicitly

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
denysoblohin-okta: I'm fine with removing classnames in this PR.

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
denysoblohin-okta: It can't be easily converted to ternary, cause the Box for enroll has [this parent](https://github.com/okta/okta-signin-widget/pull/3499/files#diff-af107efbe8a423311293438bc3e6b500a1955db3708b05954bbac1abb936e026R151) (content flex box with column direction) and the Box for verify has [another parent](https://github.com/okta/okta-signin-widget/pull/3499/files#diff-af107efbe8a423311293438bc3e6b500a1955db3708b05954bbac1abb936e026R102) (root button box)

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
denysoblohin-okta: Changed

src/v3/src/components/Icon/OktaVerifyIcon.tsx
alexdahl-okta: Yes, thanks! Good for this context to also lose the circle background. 👍

alexdahl-okta:
🎉 Looking great!

In the "Verify it's you" view, it looks like we need some space between "Select from the following options" and the authenticators list.

Perhaps `SpaceScale5` (24px)?

Or, Odyssey paragraphs should have automatic end (bottom) margin of `SpaceScale4`. Maybe the problem is that's not showing up for some reason?

Screenshot 2024-01-12 at 9 41 30 AM

test/testcafe/framework/page-objects/SelectAuthenticatorPageObject.js
```diff
@@ -2,15 +2,32 @@ import { Selector, userVariables } from 'testcafe';
...
+  ? `${factorListRowSelector} [data-se="authenticator-button-nickname"]`
+  : `${factorListRowSelector} .authenticator-enrollment-nickname`;
+const factorIconSelector = userVariables.gen3
+  ? `${factorListRowSelector} [data-se="authenticator-icon"] [data-se="factor-beacon"]`
+  : `${factorListRowSelector} .authenticator-icon-container .authenticator-icon`;
+const factorCustomLogoSelector = userVariables.gen3
+  ? `${factorListRowSelector} [data-se="authenticator-icon"]`
+  : `${factorListRowSelector} .authenticator-icon-container`;
+const factorSelectForVerificationButtonDiv = userVariables.gen3
+  ? `${factorListRowSelector} > div:nth-child(3)`
```
trevoring-okta, line 20: I assume this is the best we could get for this selector, right? I'm a bit hesitant on having the selector depend on the exact position of this button div within the component, but if this is what we have to do to get rid of `className` that's fine.

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorVerify.ts
```diff
@@ -75,7 +75,6 @@ export const transformSelectAuthenticatorVerify: IdxStepTransformer = ({
   const informationalTextElement: DescriptionElement = {
     type: 'Description',
     contentType: 'subtitle',
-    noMargin: true,
```
trevoring-okta, line 78: This was done to fix the lack of bottom margin that Alex was talking about, right?

test/testcafe/framework/page-objects/SelectAuthenticatorPageObject.js
denysoblohin-okta: Instead we can get `.parentElement` of div with `[data-se="cta-button-icon"]`

trevoring-okta: If this selector is only for the verification code path which, correct me if I'm wrong, is trying to select [this div](https://github.com/okta/okta-signin-widget/blob/93e758241579aa7864b4989cae96ce82c05708b0/src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx#L275-L289) then could we switch this to `div:last-child`? That feels a bit safer than specifically `div:nth-child(3)`

denysoblohin-okta: Yes, it's correct div. `div:last-child` is better, thanks!

trevoring-okta (approved):
LGTM, great work! Please make the changes to spacing values to use Odyssey tokens before merging

PR #3501, opened 2024-01-10
Enable Gen3 Device Fingerprinting

trevoring-okta:
## Description:
- Enables device fingerprinting in gen3 to meet parity with gen2 fingerprinting behavior
- Rewrites fingerprinting utility to remove jQuery, enforce type safety, and clean up logic
    - New logic more closely mirrors auth-js implementation of device fingerprinting: https://github.com/okta/okta-auth-js/blob/master/lib/browser/fingerprint.ts
    - Old PR with IE11 regression, for comparison: https://github.com/okta/okta-signin-widget/pull/3449/files#diff-e1279c643ed9af1d0e83c4a1140a7fbb107b1dab0ba73391dfb037528618bf47
- See video below for manual IE11 testing to ensure that updated logic does not use unsupported methods (since this reenablement originally caused IE11 regression)
    - Manual testing was also done in Chrome, Firefox, Safari, and Edge

## PR Checklist

### Issue:

- [OKTA-674340](https://oktainc.atlassian.net/browse/OKTA-674340)
- [Bacon Link](https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&branch=ti-OKTA-674340-gen3-device-fingerprinting&page=1&pageSize=6&sha=fe271c0c3d2fcf41a3ed3a55fd44e06a69850449&tab=main)
- [Downstream](https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&page=1&pageSize=6&sha=4fea7e0a3beaefe2311eb00e98bc489799a88fe4&tab=main)

#### SauceLabs IE11 demo
https://github.com/okta/okta-signin-widget/assets/106110543/bfbdb6e5-1803-4aa0-9950-1b4ff818724d

### DIFF
--- a/src/v3/src/hooks/useOnSubmit.ts
+++ b/src/v3/src/hooks/useOnSubmit.ts
@@ -20,6 +20,7 @@ import {
 } from '@okta/okta-auth-js';
 import { cloneDeep, merge, omit } from 'lodash';
 import { useCallback } from 'preact/hooks';
+import { generateDeviceFingerprint } from 'src/util/deviceFingerprintingUtils';
 
 import { IDX_STEP, ON_PREM_TOKEN_CHANGE_ERROR_KEY } from '../constants';
 import { useWidgetContext } from '../contexts';
@@ -27,6 +28,7 @@ import { MessageType } from '../types';
 import {
   areTransactionsEqual,
   containsMessageKey,
+  getBaseUrl,
   getErrorEventContext,
   getImmutableData,
   isConfigRecoverFlow,
@@ -208,6 +210,16 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     // This way the flow can maintain the latest state handle. For eg. Device probe calls
     if (step === IDX_STEP.CANCEL_TRANSACTION) {
       SessionStorage.removeStateHandle();
+    }
+    if (step === IDX_STEP.IDENTIFY && features?.deviceFingerprinting) {
+      const baseUrl = getBaseUrl(widgetProps);
+      if (baseUrl) {
+        // Proceeds with form submission even if device fingerprinting fails
+        const fingerprint = await generateDeviceFingerprint(baseUrl).catch(() => undefined);
+        if (fingerprint) {
+          authClient.http.setRequestHeader('X-Device-Fingerprint', fingerprint);
+        }
+      }
     }
     setMessage(undefined);
     try {
--- a/src/v3/src/util/browserUtils.ts
+++ b/src/v3/src/util/browserUtils.ts
@@ -24,3 +24,7 @@ export const isIOS = (): boolean => (
 );
 
 export const isAndroidOrIOS = (): boolean => isAndroid() || isIOS();
+
+export const getUserAgent = (): string => navigator.userAgent;
+
+export const isWindowsPhone = (userAgent: string): RegExpMatchArray | null => userAgent.match(/windows phone|iemobile|wpdesktop/i);
--- a/src/v3/src/util/browserUtils.ts
+++ b/src/v3/src/util/browserUtils.ts
@@ -27,4 +27,4 @@ export const isAndroidOrIOS = (): boolean => isAndroid() || isIOS();
 
 export const getUserAgent = (): string => navigator.userAgent;
 
-export const isWindowsPhone = (userAgent: string): RegExpMatchArray | null => userAgent.match(/windows phone|iemobile|wpdesktop/i);
+export const isWindowsPhone = (userAgent: string): boolean => /windows phone|iemobile|wpdesktop/i.test(userAgent);
--- a/src/v3/src/util/deviceFingerprintingUtils.test.ts
+++ b/src/v3/src/util/deviceFingerprintingUtils.test.ts
@@ -1,0 +1,130 @@
+/*
+ * Copyright (c) 2023-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import * as DeviceFingerprintingUtils from './deviceFingerprintingUtils';
+
+describe('DeviceFingerprintingUtils', () => {
+  const oktaDomainUrl = '';
+
+  beforeAll(() => {
+    const mockForm = document.createElement('form');
+    mockForm.setAttribute('data-se', 'o-form');
+    document.body.append(mockForm);
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  const mockIFrameMessages = (success: boolean, errorMessage?: { type: string }) => {
+    const message = success
+      ? { type: 'FingerprintAvailable', fingerprint: 'thisIsTheFingerprint' }
+      : errorMessage;
+
+    // TODO (jest): event is missing `origin` property
+    window.postMessage(JSON.stringify(message), '*');
+  };
+
+  const mockUserAgent = (userAgent: string) => {
+    jest.spyOn(navigator, 'userAgent', 'get').mockReturnValue(userAgent);
+  };
+
+  const bypassMessageSourceCheck = () => {
+    jest.spyOn(DeviceFingerprintingUtils, 'isMessageFromCorrectSource').mockReturnValue(true);
+  };
+
+  it('creates hidden iframe during fingerprint generation', async () => {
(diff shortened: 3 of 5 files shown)


src/v3/src/util/deviceFingerprintingUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+
+// NOTE: This utility is similar to the DeviceFingerprinting.js file used for V2 authentication flows.
+export const generateDeviceFingerprint = (oktaDomainUrl: string): Promise<string> => {
+  const userAgent = getUserAgent();
+  if (!userAgent) {
+    return Promise.reject(new Error('User agent is not defined'));
+  }
+  if (isWindowsPhone(userAgent)) {
+    return Promise.reject(new Error('Device fingerprint is not supported on Windows phones'));
+  }
```
trevoring-okta, line 26: Short circuits fingerprint generation if either of these checks fail

src/v3/src/util/deviceFingerprintingUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (formElement === null) {
+      reject(new Error('Form does not exist'));
+    }
+    formElement!.appendChild(iframe);
+
+    timeout = setTimeout(() => {
+      // If the iframe does not load, receive the right message type, or there is a slow connection, throw an error
+      reject(new Error('Service not available'));
+    }, 2000);
+  });
```
trevoring-okta, line 79: All business logic for generating iframe, attaching it as a child element of `o-form`, and setting up listeners for relevant device fingerprint messages is contained in here and results in either a resolved `Promise` with the fingerprint as a string or a rejected `Promise` with an error message

src/v3/src/util/deviceFingerprintingUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  return promise.finally(() => {
+    clearTimeout(timeout);
+    window.removeEventListener('message', listener);
+    if (formElement?.contains(iframe)) {
+      iframe.parentElement?.removeChild(iframe);
+    }
+  }) as Promise<string>;
+};
```
trevoring-okta, line 88: All clean up logic for clearing the timeout that is initialized while waiting for a fingerprint message, removing event listeners, and removing the iframe is contained within this `promise.finally` block

src/v3/src/util/deviceFingerprintingUtils.test.ts
trevoring-okta, line 1: These tests used https://github.com/okta/okta-signin-widget/blob/master/test/unit/spec/v2/view-builder/utils/DeviceFingerprinting_spec.js as a starting point

src/v3/src/util/browserUtils.ts
```diff
@@ -24,3 +24,7 @@ export const isIOS = (): boolean => (
 );
 
 export const isAndroidOrIOS = (): boolean => isAndroid() || isIOS();
+
+export const getUserAgent = (): string => navigator.userAgent;
+
+export const isWindowsPhone = (userAgent: string): RegExpMatchArray | null => userAgent.match(/windows phone|iemobile|wpdesktop/i);
```
lesterchoi-okta, line 30:
```suggestion
export const isWindowsPhone = (userAgent: string): boolean => /windows phone|iemobile|wpdesktop/i.test(userAgent);
```

src/v3/src/hooks/useOnSubmit.ts
```diff
@@ -209,6 +211,16 @@ export const useOnSubmit = (): (options: OnSubmitHandlerOptions) => Promise<void
     if (step === IDX_STEP.CANCEL_TRANSACTION) {
       SessionStorage.removeStateHandle();
     }
+    if (step === IDX_STEP.IDENTIFY && features?.deviceFingerprinting) {
+      const baseUrl = getBaseUrl(widgetProps);
+      if (baseUrl) {
+        // Proceeds with form submission even if device fingerprinting fails
+        const fingerprint = await generateDeviceFingerprint(baseUrl).catch(() => undefined);
```
lesterchoi-okta, line 218: Won't this bury the exception/rejection? Should it log before returning `undefined`?

src/v3/src/util/deviceFingerprintingUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+  let msg;
+  const formElement = document.querySelector('form[data-se="o-form"]');
+  const promise = new Promise((resolve, reject) => {
+    iframe = document.createElement('iframe');
+    iframe.style.display = 'none';
+    iframe.id = 'device-fingerprint-container';
+
+    listener = (event: MessageEvent) => {
+      if (!isMessageFromCorrectSource(iframe, event)) {
+        return undefined;
```
lesterchoi-okta, line 40: Should these conditions reject as well? The invocation catches and returns `undefined` anyway

src/v3/src/util/deviceFingerprintingUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+    iframe.src = `${oktaDomainUrl}/auth/services/devicefingerprint`;
+    if (formElement === null) {
+      reject(new Error('Form does not exist'));
+    }
+    formElement!.appendChild(iframe);
+
+    timeout = setTimeout(() => {
+      // If the iframe does not load, receive the right message type, or there is a slow connection, throw an error
+      reject(new Error('Service not available'));
+    }, 2000);
```
lesterchoi-okta, line 78: Shouldn't there be a way to control the timeout (for testing)? `auth-js` [implementation](https://github.com/okta/okta-auth-js/blob/ff14212bc8e692e081621d579740eed9f50e96c5/lib/browser/fingerprint.ts#L69) defaults to `15_000` but it can be overridden.

src/v3/src/util/deviceFingerprintingUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (isWindowsPhone(userAgent)) {
+    return Promise.reject(new Error('Device fingerprint is not supported on Windows phones'));
+  }
+
+  let timeout: NodeJS.Timeout;
+  let iframe: HTMLIFrameElement;
+  let listener: (this: Window, ev: MessageEvent) => void;
+  let msg;
+  const formElement = document.querySelector('form[data-se="o-form"]');
+  const promise = new Promise((resolve, reject) => {
```
lesterchoi-okta, line 33: I'd recommend setting the Promise return type to string. The `return undefined` lines below should then show type errors -- IMO it's true that returning undefined is type-unsafe since the method return type is set to `Promise<string>`.

```suggestion
  const promise = new Promise<string>((resolve, reject) => {
```

The type cast at the bottom `as Promise<string>` basically means to ignore the returned type in the method body.

src/v3/src/util/deviceFingerprintingUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+    }, 2000);
+  });
+
+  return promise.finally(() => {
+    clearTimeout(timeout);
+    window.removeEventListener('message', listener);
+    if (formElement?.contains(iframe)) {
+      iframe.parentElement?.removeChild(iframe);
+    }
+  }) as Promise<string>;
```
lesterchoi-okta, line 87: Remove the type cast to see that you're returning `undefined` when the `Promise` expects to resolve to `string`.

```suggestion
 promise.finally(() => {
    clearTimeout(timeout);
    window.removeEventListener('message', listener);
    if (formElement?.contains(iframe)) {
      iframe.parentElement?.removeChild(iframe);
    }
  });
  return promise;
```

Alternatively, return from L33 instead of naming it

```ts
return new Promise<string>((resolve, reject) => {
  // ...
}).finally(() => {
  // ...
});
```

lesterchoi-okta (changes requested):
type concerns

src/v3/src/util/deviceFingerprintingUtils.ts
trevoring-okta: I wrote these conditions to not reject because of the case where the window can get messages from other sources in the same origin such as browser extensions (this was called out in a comment within the auth-js implementation [here](https://github.com/okta/okta-auth-js/blob/ff14212bc8e692e081621d579740eed9f50e96c5/lib/browser/fingerprint.ts#L46-L48)). This still seemed applicable here because, otherwise fingerprinting might throw an error before it can finish because of something like receiving a message from an extension

trevoring-okta: I think you're right that the ` if (!event || !event.data || event.origin !== oktaDomainUrl)` check should reject with an error though

src/v3/src/util/deviceFingerprintingUtils.ts
lesterchoi-okta: Re-reading the logic here, ignore my previous about `return undefined` and type errors. Still think setting the `Promise<string>` generic is a good idea if it's the expected return type for the method.

src/v3/src/hooks/useOnSubmit.ts
lesterchoi-okta: Keep it the same as Gen2 for now. TestCafe specs and downstreams may fail if there are unexpected logs/errors in console.

lesterchoi-okta (dismissed):
LGTM

lesterchoi-okta (approved):

PR #3504, opened 2024-01-12
Patch Preact translate bug with patch-package

trevoring-okta:
## Description:
- Uses [patch-package](https://yarnpkg.com/package?name=patch-package) to patch Preact [bug](https://github.com/preactjs/preact/issues/3799) with the same fix found [here](https://github.com/preactjs/preact/pull/3800)
- Adds support for passing a `SKIP_POSTINSTALL` flag to a `yarn install` call which prevents the `postinstall` `patch-package` call from running. This is needed for the `verify-package.sh` script to work because the `yarn install` calls from that script were coming from the installation of the SIW as a dependency for a test app, which was breaking since the test app itself did not have `patch-package` installed. The english leak test suite doesn't need `patch-package` anywhere except for `yarn install` calls from the package root
- Other small lint fixes

## PR Checklist

### Issue:

- [OKTA-684531](https://oktainc.atlassian.net/browse/OKTA-684531)

#### Highlighted `div` with "Label" is imported from Odyssey and was passed `translate="no"` as a prop
Screenshot 2024-01-12 at 6 43 17 PM


package.json
```diff
@@ -102,7 +102,8 @@
     "test:parity-setup": "mkdir -p test-reports/testcafe && yarn generate-config && yarn workspace v3 dev",
     "test:parity-run": "OKTA_SIW_GEN3=true yarn testcafe -c 4 chrome:headless --reporter spec,xunit:build2/reports/junit/testcafe-results.xml",
     "test:parity-ci": "OKTA_SIW_SKIP_FLAKY=true yarn codegen && run-p -r test:parity-setup test:parity-run",
-    "test:parity-ci-flaky": "OKTA_SIW_GEN3=true OKTA_SIW_ONLY_FLAKY=true yarn test:parity-ci"
+    "test:parity-ci-flaky": "OKTA_SIW_GEN3=true OKTA_SIW_ONLY_FLAKY=true yarn test:parity-ci",
+    "postinstall": "if [[ -z \"$SKIP_POSTINSTALL\" ]]; then patch-package ; fi"
```
denysoblohin-okta, line 106: This command would not work on Windows. 
Tried with GitBash and received error "-z was unexpected at this time."

Maybe we can just strip out `postinstall` command from npm package [here](https://github.com/okta/okta-signin-widget/blob/master/scripts/buildtools/commands/prepack.js#L41) (since we don't export Gen3 anyway in npm package, and I hope this fix is temporary)?

trevoring-okta: Ooh, didn't know about this `prepack` routine. Good idea, will look into it

denysoblohin-okta (approved):
Please rebase to the latest `odyssey-1.x` commit. 
Should resolve `translate="yes"` issues in snapshots

PR #3511, opened 2024-01-17
Migrate Spinner to Odyssey CircularProgress

trevoring-okta:
## Description:
- Migrate SIW `Spinner` to Odyssey `CircularProgress`
- Add theme override for rendering a white spinner when it is rendered as a child of a primary button because Odyssey `CircularProgress` does not support changing the color since they don't expect that a spinner would be used as the icon for a button
    - This theme override is guaranteed to be safe from an accessibility POV because the primary palette token (`PalettePrimaryMain`) is generated by the Leonardo package which always ensures that it adheres to WCAG contrast ratios with white text

## PR Checklist

### Issue:

- [OKTA-678806](https://oktainc.atlassian.net/browse/OKTA-678806)

#### Spinner only renders as white on a primary button, otherwise it renders as `PrimaryPaletteMain`
https://github.com/okta/okta-signin-widget/assets/106110543/58b9aaff-eb67-488c-91c1-3f9928d29163

### DIFF
--- a/src/v3/src/util/theme.ts
+++ b/src/v3/src/util/theme.ts
@@ -123,16 +123,6 @@ export const createThemeAndTokens = (
   // Merge default Odyssey 1.x theme with component overrides
   const themeOverride = mergeThemes(baseOdysseyTheme, {
     components: {
-      MuiButton: {
-        styleOverrides: {
-          root: {
-            display: 'flex',
-            justifyContent: 'center',
-            alignItems: 'center',
-            whiteSpace: 'normal',
-          },
-        },
-      },
       MuiAlert: {
         styleOverrides: {
           root: {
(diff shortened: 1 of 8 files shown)


src/v3/src/util/theme.ts
```diff
@@ -134,6 +134,19 @@ export const createThemeAndTokens = (
...
         },
       },
+      MuiButton: {
+        styleOverrides: {
+          root: ({ ownerState }) => ({
+            // Odyssey CircularProgress does not allow color change but SIW needs a white spinner
+            // when rendering it as the startIcon for primary buttons
+            ...(ownerState.variant === 'primary' && {
+              '& .MuiCircularProgress-root': {
+                color: 'white',
```
denysoblohin-okta, line 144: Can we use some property from `mergedTokens ` , like `HueNeutralWhite`?

denysoblohin-okta (approved):
LGTM. Just one question about using token instead of `'white'`

PR #3527, opened 2024-01-30
Migrate `InputText` to `TextField`

trevoring-okta:
## Description:
- Migrate `InputText` to use Odyssey `TextField`
- Updates React Testing Library integration tests to use better selectors where possible (prioritized `findByLabelText`)

NOTES: 
- All migration PRs have the same changes to update `buttonFocusRef` and `linkFocusRef`, which I was asked to change upstream in Odyssey. Once one of these PRs gets merged in those changes will no longer be marked as file changes
- All migration PRs switch to use a `buildFieldLevelErrorMessages.ts` utility that combines logic that was present in the `FieldLevelMessageContainer` and `WidgetMessageContainer` components. This utility takes in the list of error messages and returns an object containing the appropriate values to pass to `errorMessage` and `errorMessageList` for Odyssey `Field` components.
- Tested each component migration manually in IE11 and fixed any bugs I saw
- Ran english leak detection suite locally for all migrations since english leak script is not on gen 3 in Bacon

## PR Checklist

### Issue:

- [OKTA-678814](https://oktainc.atlassian.net/browse/OKTA-678814)

### DIFF
--- a/src/v3/src/components/InputText/InputText.tsx
+++ b/src/v3/src/components/InputText/InputText.tsx
@@ -10,15 +10,10 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { useOdysseyDesignTokens } from '@okta/odyssey-react-mui';
-import {
-  Box,
-  FormHelperText,
-  InputBase,
-  InputLabel,
-  Typography,
-} from '@okta/odyssey-react-mui-legacy';
+import { Box } from '@mui/material';
+import { TextField } from '@okta/odyssey-react-mui';
 import { h } from 'preact';
+import { buildFieldLevelErrorMessages } from 'src/util/buildFieldLevelErrorMessages';
 
 import { useWidgetContext } from '../../contexts';
 import {
@@ -31,16 +26,13 @@ import {
   UISchemaElementComponent, UISchemaElementComponentWithValidationProps,
 } from '../../types';
 import { getTranslation } from '../../util';
-import FieldLevelMessageContainer from '../FieldLevelMessageContainer';
 import { withFormValidationState } from '../hocs';
 
 const InputText: UISchemaElementComponent<UISchemaElementComponentWithValidationProps> = ({
-  type,
   uischema,
   errors,
   handleChange,
   handleBlur,
-  describedByIds,
 }) => {
   const value = useValue(uischema);
   const { loading } = useWidgetContext();
@@ -48,112 +40,46 @@ const InputText: UISchemaElementComponent<UISchemaElementComponentWithValidation
     translations = [],
     focus,
     parserOptions,
-    showAsterisk,
     dir,
   } = uischema;
   const label = getTranslation(translations, 'label');
   const hint = getTranslation(translations, 'hint');
   const explain = getTranslation(translations, 'bottomExplain');
-  const optionalLabel = getTranslation(translations, 'optionalLabel');
   const {
     attributes,
     inputMeta: { name, required },
     dataSe,
   } = uischema.options;
+  const { autocomplete, inputmode } = attributes || {};
   const focusRef = useAutoFocus<HTMLInputElement>(focus);
   const parsedExplainContent = useHtmlContentParser(explain, parserOptions);
-  const hasErrors = typeof errors !== 'undefined';
-  // TODO: OKTA-569647 - refactor logic
-  const hintId = hint && `${name}-hint`;
-  const explainId = explain && `${name}-explain`;
-  const ariaDescribedByIds = [describedByIds, hintId, explainId].filter(Boolean).join(' ')
-    || undefined;
-  const tokens = useOdysseyDesignTokens();
+  const { errorMessage, errorMessageList } = buildFieldLevelErrorMessages(errors);
 
   return (
-    <Box>
-      <InputLabel
-        htmlFor={name}
-        // To prevent asterisk from shifting far right
-        sx={{ justifyContent: showAsterisk ? 'flex-start' : undefined }}
-      >
-        {label}
-        {showAsterisk && (
-          <Box
-            component="span"
-            sx={{
-              marginInlineStart: tokens.Spacing2,
-              marginInlineEnd: tokens.Spacing2,
-            }}
-            className="no-translate"
-            aria-hidden
-          >
-            *
-          </Box>
-        )}
-        {required === false && (
-          <Typography
-            variant="subtitle1"
-            sx={{ whiteSpace: 'nowrap' }}
-          >
-            {optionalLabel}
-          </Typography>
-        )}
-      </InputLabel>
-      {
-        hint && (
-          <FormHelperText
-            id={hintId}
-            className="o-form-explain"
-            data-se={hintId}
-            // TODO: OKTA-577905 - Temporary fix until we can upgrade to the latest version of Odyssey
-            sx={{ textAlign: 'start' }}
-          >
-            {hint}
-          </FormHelperText>
-        )
-      }
-      <InputBase
-        value={value}
-        type={type || 'text'}
+    <Box dir={dir}>
+      <TextField
+        autoCompleteType={autocomplete}
+        errorMessage={errorMessage}
+        errorMessageList={errorMessageList}
+        hint={hint ?? parsedExplainContent as string}
         id={name}
+        inputFocusRef={focusRef}
+        inputMode={inputmode}
+        isDisabled={loading}
+        isFullWidth
+        isOptional={!required}
         name={name}
-        error={hasErrors}
+        label={label ?? ''}
         onChange={(e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
           handleChange?.(e.currentTarget.value);
         }}
         onBlur={(e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
           handleBlur?.(e?.currentTarget?.value);
         }}
-        disabled={loading}
-        fullWidth
-        inputProps={{
-          'data-se': dataSe,
-          'aria-describedby': ariaDescribedByIds,
-          ...attributes,
-        }}
-        inputRef={focusRef}
-        dir={dir}
+        testId={dataSe}
+        type="text"
+        value={value as string | undefined}
       />
-      {hasErrors && (
-        <FieldLevelMessageContainer
-          messages={errors}
-          fieldName={name}
-        />
-      )}
-      {
-        explain && (
-          <FormHelperText
-            id={explainId}
-            className="o-form-explain"
-            data-se={explainId}
-            // TODO: OKTA-577905 - Temporary fix until we can upgrade to the latest version of Odyssey
-            sx={{ textAlign: 'start' }}
-          >
-            {parsedExplainContent}
-          </FormHelperText>
-        )
-      }
     </Box>
   );
 };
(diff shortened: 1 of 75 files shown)


src/v3/src/components/InputText/InputText.tsx
```diff
@@ -31,130 +25,58 @@ import {
...
     translations = [],
     focus,
     parserOptions,
-    showAsterisk,
-    dir,
   } = uischema;
   const label = getTranslation(translations, 'label');
   const hint = getTranslation(translations, 'hint');
   const explain = getTranslation(translations, 'bottomExplain');
-  const optionalLabel = getTranslation(translations, 'optionalLabel');
```
trevoring-okta, line 57: Optional label is automatically added by Odyssey via `isOptional` prop

src/v3/src/components/InputText/InputText.tsx
```diff
@@ -31,130 +25,58 @@ import {
...
-            {parsedExplainContent}
-          </FormHelperText>
-        )
-      }
-    </Box>
+    <TextField
+      autoCompleteType={autocomplete}
+      errorMessage={errorMessage}
+      errorMessageList={errorMessageList}
+      hint={hint ?? parsedExplainContent as string}
```
trevoring-okta, line 61: `parsedExplainContent` is only used for a single type of input that shows a description for a secondary email address during profile enrollment. `hint` is only used on the primary identify screen if users pass in a custom string for `primaryauth.password.placeholder`. As a result, these two strings are mutually exclusive and can both use the Odyssey `hint` prop

src/v3/src/components/InputText/InputText.tsx
```diff
@@ -31,130 +25,58 @@ import {
...
   const label = getTranslation(translations, 'label');
   const hint = getTranslation(translations, 'hint');
   const explain = getTranslation(translations, 'bottomExplain');
-  const optionalLabel = getTranslation(translations, 'optionalLabel');
   const {
     attributes,
     inputMeta: { name, required },
     dataSe,
   } = uischema.options;
+  const { autocomplete, inputmode } = attributes || {};
```
trevoring-okta, line 51: The only values in `attributes` are `autocomplete` and `inputmode` so we are not losing any attributes by not spreading the object as props passed to `InputBase` like the old code was doing

src/v3/src/util/buildFieldLevelErrorMessages.ts
```diff
@@ -0,0 +1,47 @@
...
+      const error = errors[0];
+      // Covers nested messages, e.g. password requirements not met
+      if (Array.isArray(error.message)) {
+        return {
+          errorMessage: error.description,
+          errorMessageList: error.message
+            .filter((msg) => typeof msg.message === 'string')
+            .map((msg) => msg.message as string),
+        };
+      }
```
trevoring-okta, line 36: Confirmed with @glenfannin-okta  that we only ever see one level of "nested messages" in the SIW, so this utility function does not have to be recursive. The **only** case where we even see nesting within field level error messages is the password requirements not met error with a list of requirements that haven't been met (see screenshot). In this case, we use the `errorMessage` as the place to pass in the `error.description` and the `errorMessageList` to pass in the list of errors

Screenshot 2024-01-30 at 5 36 53 PM

src/v3/src/util/isLtrField.ts
```diff
@@ -0,0 +1,22 @@
...
+// fields with these names should stay LTR by default
+const ltrOnlyFieldNames = [
+  'credentials.passcode',
+  'identifier',
+  'credentials.newPassword',
+  'confirmPassword',
+  'extension',
+];
+
+export const isLtrField = (fieldName: string) => ltrOnlyFieldNames.includes(fieldName) || fieldName.endsWith('phoneNumber');
```
trevoring-okta, line 22: Extends the logic used in https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/transformer/uischema/setLtrFields.ts to also include phone number fields since the `PhoneAuthenticator` is also always LTR. This utility is then used in `theme.ts` to add MUI theme overrides for LTR exceptions

src/v3/src/util/theme.ts
```diff
@@ -151,11 +152,21 @@ export const createThemeAndTokens = (
...
+            display: 'block',
           },
+        },
+      },
+      MuiInputBase: {
+        styleOverrides: {
+          root: ({ ownerState }) => ({
+            width: '100%',
+            ...(ownerState.name && isLtrField(ownerState.name) && {
+              direction: 'ltr',
```
trevoring-okta, line 167: Makes all components that use `InputBase` under the hood (which happens to align nicely with all of the components that the SIW forces to be LTR) to be LTR at the theming layer since Odyssey does not support passing in direction props anywhere

src/v3/src/util/theme.ts
```diff
@@ -151,11 +152,21 @@ export const createThemeAndTokens = (
           }),
         },
       },
-      MuiInputBase: {
+      MuiFormHelperText: {
         styleOverrides: {
           root: {
-            width: '100%',
+            display: 'block',
```
trevoring-okta, line 158: IE11 fix

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -337,7 +337,7 @@ export default class BaseFormObject {
 
   getNthErrorMessage(fieldName, value) {
     if (userVariables.gen3) {
-      return this.el.find(`#${fieldName}-error-${value}`).innerText;
+      return this.el.find(`#${fieldName}-error`).child('div').child('ul').child('li').nth(value).innerText;
```
trevoring-okta, line 340: Odyssey field level error component is a single parent element with each error as a child, while old SIW behavior was each error is its own element

test/testcafe/spec/IdentifyWithPassword_spec.js
```diff
@@ -83,7 +83,7 @@ test.requestHooks(identifyWithPasswordMock)('should show customized error if req
   await identityPage.waitForErrorBox();
 
   await t.expect(identityPage.hasIdentifierErrorMessage()).eql(true);
-  await t.expect(identityPage.getIdentifierErrorMessage()).eql('Username is required!');
+  await t.expect(identityPage.getIdentifierErrorMessage()).match(/Username is required!/);
```
trevoring-okta, line 86: We have to use `.match(/..../)` because any optional Odyssey fields will bring in the "Optional" string as text content when trying to use a selector

test/testcafe/spec/Identify_spec.js
```diff
@@ -173,7 +173,7 @@ test.requestHooks(identifyMock)('should show errors if required fields are empty
   await identityPage.clickNextButton();
   await identityPage.waitForErrorBox();
 
-  await t.expect(identityPage.getIdentifierErrorMessage()).eql('This field cannot be left blank');
+  await t.expect(identityPage.getIdentifierErrorMessage()).match(/This field cannot be left blank/);
```
trevoring-okta, line 176: We have to use .match(/..../) because any error messages in Odyssey fields will bring in an "Error:" string that is from Odyssey's `ScreenReaderText` component as text content when trying to use a selector

src/v3/test/integration/identify-with-password.test.tsx
```diff
@@ -37,10 +37,10 @@ describe('identify-with-password', () => {
     });
 
     await findByText('Sign in', { selector: 'button' });
-    const usernameEl = await findByTestId('identifier-hint') as HTMLLabelElement;
+    const usernameEl = await findByLabelText(/Username/) as HTMLInputElement;
```
trevoring-okta, line 40: We have to use `/Username/` because optional Odyssey fields will bring in the "Optional" string as text content when trying to select a field by its label

src/v3/src/components/InputText/InputText.tsx
denysoblohin-okta: It is desired for explanation text to be rendered above the text input field (instead of below the field as was before migration), right?

Screenshot 2024-02-02 at 12 23 45

src/v3/test/integration/identify-with-password.test.tsx
```diff
@@ -69,13 +69,13 @@ describe('identify-with-password', () => {
...
     const {
-      findByTestId,
+      findByLabelText,
     } = await setup({
       mockResponse,
       widgetOptions: { username: mockUsername },
     });
 
-    const usernameEl = await findByTestId('identifier') as HTMLInputElement;
+    const usernameEl = await findByLabelText(/Username/) as HTMLInputElement;
```
denysoblohin-okta, line 78: Why can't we use `findByTestId` here?

test/testcafe/spec/EnrollAuthenticatorPhoneView_spec.js
```diff
@@ -167,9 +167,9 @@ test
     const enrollPhonePageObject = await setup(t);
     await checkA11y(t);
     await enrollPhonePageObject.clickNextButton();
-    await t.expect(await enrollPhonePageObject.resendCodeExists(2)).eql(false);
+    await t.expect(await enrollPhonePageObject.resendCodeExists(1)).eql(false);
```
denysoblohin-okta, line 170: Just curious, why the nth index was 2 and now it should equal 1?

test/testcafe/framework/page-objects/IdentityPageObject.js
```diff
@@ -272,6 +272,9 @@ export default class IdentityPageObject extends BasePageObject {
   }
 
   getIdentifierSubLabelValue() {
+    if (userVariables.gen3) {
+      return Selector('#identifier-hint').textContent;
```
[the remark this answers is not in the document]
denysoblohin-okta: Sorry, haven't noticed your last commit 
https://github.com/okta/okta-signin-widget/pull/3527/commits/788ab9f20c956129ce5aa988d1a25dfb8c72d2ce

denysoblohin-okta (approved):
LGTM

src/v3/src/components/InputText/InputText.tsx
denysoblohin-okta: Casting `parsedExplainContent` to string seems a workaround. In the screenshot above `hint` value is an array of React elements, but Ody expcts string

trevoring-okta: For the explanation text above the field, yes we decided to move it to reuse the `hint` prop location because there's no support for a bottom explain location and we only need it for this one case. For the casting of `parsedExplainContent`, yes I think we are forced to cast as a workaround because Odyssey doesn't allow arrays of React elements. I'm not sure we have another option besides asking Odyssey to support this unless you have other ideas?

test/testcafe/spec/EnrollAuthenticatorPhoneView_spec.js
trevoring-okta: These tests started failing maybe due to some flakiness with timers and the resend code alert not appearing right away. I've also tested locally and seems like the form-level "We found some errors" pop up does not always appear when testcafe tries to simulate clicking the "Verify" button. Regardless, I think the resend code should always at least be the 0th or 1st index anyway if you see the screenshot. I think since the original test was only asserting the presence of an element with `role="alert"` at the 2nd index and not the content of that element, that was actually asserting the presence of the field-level "This field cannot be left blank"
Screenshot 2024-02-02 at 5 45 37 PM

PR #3533, opened 2024-01-30
Migrate `Select`

trevoring-okta:
## Description:
- Migrate `Select` to use Odyssey `NativeSelect`

## PR Checklist

### Issue:

- [OKTA-642558](https://oktainc.atlassian.net/browse/OKTA-642558)

### DIFF
--- a/src/v3/src/components/Select/Select.tsx
+++ b/src/v3/src/components/Select/Select.tsx
@@ -39,6 +39,7 @@ const Select: UISchemaElementComponent<UISchemaElementComponentWithValidationPro
   const { errorMessage, errorMessageList } = buildFieldLevelErrorMessages(errors);
 
   const {
+    attributes,
     inputMeta: {
       name,
       options,
@@ -46,6 +47,7 @@ const Select: UISchemaElementComponent<UISchemaElementComponentWithValidationPro
     },
     customOptions,
   } = uischema.options;
+  const { autocomplete } = attributes || {};
   const focusRef = useAutoFocus<HTMLSelectElement>(focus);
 
   const getOptions = (): IdxOption[] | undefined => {
@@ -63,11 +65,12 @@ const Select: UISchemaElementComponent<UISchemaElementComponentWithValidationPro
 
   return (
     <NativeSelect
+      autoCompleteType={autocomplete}
       errorMessage={errorMessage}
       errorMessageList={errorMessageList}
       hint={!required ? optionalLabel : undefined}
       id={name}
-      inputFocusRef={focusRef}
+      inputRef={focusRef}
       isDisabled={loading}
       label={label}
       onChange={(e: SelectChangeEvent<string>) => {
--- a/src/v3/src/components/Select/Select.tsx
+++ b/src/v3/src/components/Select/Select.tsx
@@ -11,10 +11,7 @@
  */
 
 import { SelectChangeEvent } from '@mui/material';
-import { useOdysseyDesignTokens } from '@okta/odyssey-react-mui';
-import {
-  Box, FormControl, InputLabel, Select as MuiSelect, Typography,
-} from '@okta/odyssey-react-mui-legacy';
+import { NativeSelect } from '@okta/odyssey-react-mui';
 import { IdxOption } from '@okta/okta-auth-js/types/lib/idx/types/idx-js';
 import { h } from 'preact';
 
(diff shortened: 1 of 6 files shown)


test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -348,7 +348,7 @@ export default class BaseFormObject {
 
   getValueFromDropdown(fieldName, index = 0) {
     if (userVariables.gen3) {
-      const selectEle = this.el.find(`[data-se="${fieldName}"]`);
+      const selectEle = this.el.find(`[id="${fieldName}"]`);
```
trevoring-okta, line 351: As part of these migrations I've been trying to fix failing tests by using alternatives to `data-se`

src/v3/src/components/Select/Select.tsx
```diff
@@ -24,24 +21,22 @@ import {
...
 const Select: UISchemaElementComponent<UISchemaElementComponentWithValidationProps> = ({
   uischema,
   errors,
   handleChange,
   handleBlur,
-  describedByIds,
 }) => {
   const value = useValue(uischema);
   const { loading } = useWidgetContext();
-  const { focus, translations = [], showAsterisk } = uischema;
```
denysoblohin-okta, line 40: It's intentional to remove asterisks in Odyssey 1 migration, right?

src/v3/src/components/Select/Select.tsx
```diff
@@ -69,93 +64,50 @@ const Select: UISchemaElementComponent<UISchemaElementComponentWithValidationPro
...
 
   return (
-    <FormControl
-      disabled={loading}
-      error={hasErrors}
+    <NativeSelect
+      autoCompleteType={autocomplete}
+      errorMessage={errorMessage}
+      errorMessageList={errorMessageList}
+      hint={!required ? optionalLabel : undefined}
```
denysoblohin-okta, line 71: Why not use `isOptional` prop instead of `hint`?

test/testcafe/framework/page-objects/components/BaseFormObject.js
denysoblohin-okta: Ody 1.11.1 renders `data-se` attribute for `select` (1.9.x versions were rendering `data-se` for parent `div` instead). Does it still causes test failures?

src/v3/src/components/Select/Select.tsx
```diff
@@ -69,93 +64,50 @@ const Select: UISchemaElementComponent<UISchemaElementComponentWithValidationPro
...
+        handleChange?.(selectTarget.value);
+      }}
+      onBlur={(e: SelectChangeEvent<string>) => {
+        const selectTarget = (
+          e?.target as SelectChangeEvent['target'] & { value: string; name: string; }
+        );
+        handleBlur?.(selectTarget.value);
+      }}
+      testId={name}
+      value={value as string}
```
denysoblohin-okta, line 89: This should be changed to 
```js
value={value ?? ''}
```

Otherwise Odyssey would set [`controlState`](https://github.com/okta/odyssey/blob/00daa33877428125405ce716101dd4c089081922/packages/odyssey-react-mui/src/NativeSelect.tsx#L162) to `UNCONTROLLED`
https://github.com/okta/odyssey/blob/00daa33877428125405ce716101dd4c089081922/packages/odyssey-react-mui/src/inputUtils.ts#L36-L37

This leads to a bug with rendering default value. 
Eg. for mock `authenticator-enroll-security-question` select should render [first option](https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/transformer/layout/securityQuestion/securityQuestionEnroll.ts#L97) as default value:
https://github.com/okta/okta-signin-widget/blob/master/src/v3/test/e2e/screenshots/authenticator-enroll-security-question.png

On first render of `Select` component `value` prop is `undefined` (which leads to uncontrolled state detected by Odyssey, and it would not be changed to controlled), on second render `value` prop is `'disliked_food'`, but it will be rendered as "Select option"

src/v3/src/components/Select/Select.tsx
trevoring-okta: Not sure how this happened. I think one of the other components I migrated didn't support `isOptional` so I was forced to do this and then I just assumed that `NativeSelect` doesn't support it as well. I'll change that here and in the `PhoneAuthenticator` migration, nice catch

src/v3/src/components/Select/Select.tsx
trevoring-okta: Ah, yeah, another good catch. I handled this uncontrolled state issue in a few of the other migrations but missed it here.

denysoblohin-okta (approved):

PR #3537, opened 2024-01-30
Migrate `Checkbox`

trevoring-okta:
## Description:
- Migrate `Checkbox` to Odyssey `CheckboxGroup` and `Checkbox`

## PR Checklist

### Issue:

- [OKTA-678787](https://oktainc.atlassian.net/browse/OKTA-678787)


src/v3/src/components/Checkbox/Checkbox.tsx
```diff
@@ -10,121 +10,70 @@
...
-    <FormControl
-      component="fieldset"
-      error={hasErrors}
-      aria-describedby={describedByIds}
-      className={noTranslate ? 'no-translate' : undefined}
+    <CheckboxGroup
+      errorMessage={errorMessage}
+      errorMessageList={errorMessageList}
+      id={name}
+      label=""
```
trevoring-okta, line 58: Because of the granular consent views, we have situations where individual checkboxes can get errors rendered under them that are specific to that checkbox. Thus, we have to wrap each checkbox in an Odyssey `CheckboxGroup` so we can pass in error messages because the individual `Checkbox` component in Odyssey does not support rendering error messages below it

src/v3/src/components/Checkbox/Checkbox.tsx
```diff
@@ -10,121 +10,70 @@
...
 
   const {
     options: { inputMeta: { name, mutable } },
     focus,
-    noTranslate,
     translations = [],
-    showAsterisk,
   } = uischema;
   const isReadOnly = mutable === false;
+  const checkboxId = `${name}-checkbox`;
```
trevoring-okta, line 41: We set the field name as the ID for the `CheckboxGroup` since that's the field that gets associated with aria values and then create this checkbox ID to pass to the underlying `Checkbox` component

src/v3/src/components/Checkbox/Checkbox.tsx
```diff
@@ -10,121 +10,70 @@
...
   const labelInfo = getTranslationInfo(translations, 'label');
+  const label = useHtmlContentParser(labelInfo?.noTranslate
+    ? wrapInTranslateNo(labelInfo.value)
+    : labelInfo?.value) as string | undefined;
   const descriptionInfo = getTranslationInfo(translations, 'description');
+  const description = useHtmlContentParser(descriptionInfo?.noTranslate
+    ? wrapInTranslateNo(descriptionInfo.value)
+    : descriptionInfo?.value) as string | undefined;
```
trevoring-okta, line 49: The granular consent views are some of the only ones where we have to be able to apply `translate="no"` to part of a component but not another (e.g. don't translate the checkbox label but do translate the description/hint below it). To get around this, I wrote this utility to wrap a string to not be translated and then use the HTML content parser hook in the `Checkbox` component to parse this HTML string to a React element. Fortunately, Odyssey does not actually strictly type check the type once it is passed in and will render a React element at runtime so casting the wrapped `label` and `description` is a (not so great) workaround for this issue

test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
```diff
@@ -106,15 +106,11 @@ export default class ChallengeOktaVerifyPushPageObject extends ChallengeFactorPa
     return label;
   }
 
-  getAutoChallengeCheckboxLabel() {
-    return this.form.getElement(AUTO_CHALLENGE_CHECKBOX_LABEL_SELECTOR);
-  }
-
```
trevoring-okta, line 112: Was not needed after updating corresponding spec

src/v3/src/components/Checkbox/Checkbox.tsx
denysoblohin-okta: Can we create a ticket to review `translate` prop later? 
Possibly changing shape of it from `yes | no` to map like `{ hint: 'yes' | 'no', label: 'yes' | 'no', root: 'yes' | 'no' }` and add support of it to Odyssey and SIW.

trevoring-okta: Yeah, we can talk to the Odyssey team about this. This issue was something we knew was going to come up when we implemented this in Odyssey, but we didn't want to make the API too complex since we only have to do this workaround for the granular consent view in SIW

denysoblohin-okta (approved):

PR #3544, opened 2024-01-31
Migrate `ConsentHeader`

trevoring-okta:
## Description:
- Migrate `ConsentHeader` to use Odyssey components

## PR Checklist

### Issue:

- [OKTA-678813](https://oktainc.atlassian.net/browse/OKTA-678813)


src/v3/test/integration/__snapshots__/admin-consent.test.tsx.snap
```diff
@@ -42,7 +42,6 @@ exports[`admin-consent should render form with logo 1`] = `
                     class="MuiBox-root emotion-1"
                   >
                     <a
-                      aria-label="Logo for the app"
```
glenfannin-okta, line 45: aria-label is probably required here since this is an anchor tag with no text.

src/v3/src/components/ConsentHeader/ConsentHeader.tsx
```diff
@@ -69,12 +68,11 @@ const ConsentHeader: FunctionComponent = () => {
       >
         {typeof href !== 'undefined' && typeof logoHref !== 'undefined'
           ? (
-            <Box component="div">
+            <Box>
               <Link
                 href={href}
                 target="_blank"
                 rel="noopener noreferrer"
-                aria-label={altText}
```
glenfannin-okta, line 77: aria-label is probably required here since this is an anchor tag with no text.

trevoring-okta: Got it. I think I saw this issue raised as an `a11y` error at some point but this PR was also giving me weird test failures so I reverted to get it back to this state. I think the `a11y` runs via e2e tests that I didn't run locally though so I'm guessing Bacon will tell me you're right. This means we'll have to import this `Link` component from `@mui/material` since Odyssey `Link` doesn't support aria-label, but that's not the end of the world

glenfannin-okta: Oh interesting, did Odyssey say they will eventually support that prop? That seems like a standard thing to allow especially given how anchors can wrap anything (and not necessarily rely on the inner content to represent the text for screen readers).

trevoring-okta: I think they expect that `Link` components will always render some sort of text, though I can ask

glenfannin-okta (approved):

PR #3549, opened 2024-02-12
OKTA-689219 : Updating authenticator button aria labels

glenfannin-okta:
## Description:
The purpose of this PR is to correct the spelling on some authenticator button i18n keys and add new dedicated keys/labels for OV method types to provide more context to users utilizing a screenreader. 

## PR Checklist

### Issue:

- [OKTA-689219](https://oktainc.atlassian.net/browse/OKTA-689219)


shuowu-okta (dismissed):

trevoring-okta:
One thing I noticed that should probably be added at some point with all of these updates to `AuthenticatorButton` aria labels is that lots of the transformer tests in `src/v3/src/transformer/selectAuthenticator` now have TS errors because `ariaLabel` was added as a required prop but many of the mock auth button objects that were not updated to have an `ariaLabel`

trevoring-okta:
Another somewhat related question that I've been curious about since I've been encountering a lot of merge conflicts with `AuthenticatorButton` is why we don't have `aria-labelledby` set to `{`${iconName}-label`}` on the authenticator button component?

glenfannin-okta:
> One thing I noticed that should probably be added at some point with all of these updates to `AuthenticatorButton` aria labels is that lots of the transformer tests in `src/v3/src/transformer/selectAuthenticator` now have TS errors because `ariaLabel` was added as a required prop but many of the mock auth button objects that were not updated to have an `ariaLabel`. If that could be fixed in this PR it would probably prevent confusion in the future and create a cleaner Git history. Otherwise, LGTM!

Done, I didn't even notice those tests had TS errors, I forget I have to open the v3 workspace directly to see reported TS errors.

glenfannin-okta:
> Another somewhat related question that I've been curious about since I've been encountering a lot of merge conflicts with `AuthenticatorButton` is why we don't have `aria-labelledby` set to `{`${iconName}-label`}` on the authenticator button component?

In Gen 3, we were doing that, however, the label didn't provide enough context to screen readers (eg. hovering a button, the reader would just say "Email" or "Enter a code" which doesn't make sense, but "Select Email." or "Select phone" made much more sense), and since we also had to do this in Gen 2, we felt adding labels that provide clear instructions and match the labels between both generations were the proper way to go.

trevoring-okta (approved):

PR #3561, opened 2024-03-05
Gen1: Add translation for errors.E0000068

denysoblohin-okta:
## Description:

Adds translation to localize `E0000068` error.

Currently (for Gen1 only) a custom 'okta-i18n-error' event is triggered for i18n key `errros.E0000068`
Error is triggered from https://github.com/okta/okta-signin-widget/blob/e63951ffcff0fd6843199d546c52ee332db93c0a/src/util/Util.js#L102

Example:
https://github.com/okta/okta-signin-widget/blob/e63951ffcff0fd6843199d546c52ee332db93c0a/test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_errorActivate.js#L5-L11
https://github.com/okta/okta-signin-widget/blob/e63951ffcff0fd6843199d546c52ee332db93c0a/test/unit/spec/v1/EnrollCall_spec.js#L519
In this test an unlocalized `errorSummary` was used to return an error message, because of the fallback in this line:
https://github.com/okta/okta-signin-widget/blob/e63951ffcff0fd6843199d546c52ee332db93c0a/src/util/Util.js#L98
This PR fixes this issue by localizing error message with code `E0000068` . Error message in the test above would equal `Invalid Passcode\/Answer`

Error description in Okta API: https://developer.okta.com/docs/reference/error-codes/#E0000068

## PR Checklist

### Issue:

- [OKTA-700843](https://oktainc.atlassian.net/browse/OKTA-700843)

### Screenshot/Video:

For https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/xhr/MFA_VERIFY_invalid_password.js

Before:
Screenshot 2024-03-06 at 14 54 51

After:
Screenshot 2024-03-06 at 15 06 13


test/unit/spec/v1/EnrollSms_spec.js
```diff
@@ -724,7 +724,13 @@ Expect.describe('EnrollSms', function() {
         });
     });
     itp('shows error if error response on verification', function() {
-      return sendValidCodeFn()
+      return sendValidCodeFn({
+        i18n: {
+          en: {
+            'errors.E0000068': 'Invalid Passcode or Answer'
```
lesterchoi-okta, line 730: Looks like the translation string and this response mismatch?

"errors.E0000068 = Invalid Passcode/Answer"
vs
'errors.E0000068': 'Invalid Passcode or Answer'

Does the test pass?

test/unit/spec/v1/MfaVerify_spec.js
```diff
@@ -2232,22 +2238,17 @@ Expect.describe('MFA Verify', function() {
         })
         .then(function(test) {
           expect(test.form.hasErrors()).toBe(true);
-          expect(test.form.errorMessage()).toBe('Password is incorrect');
```
lesterchoi-okta, line 2235: Where did this message come from previously?

test/unit/spec/v1/MfaVerify_spec.js
```diff
@@ -1126,7 +1132,7 @@ Expect.describe('MFA Verify', function() {
         })
         .then(function(test) {
           expect(test.form.hasErrors()).toBe(true);
-          expect(test.form.errorMessage()).toBe('Your answer doesn\'t match our records. Please try again.');
```
lesterchoi-okta, line 1129: Where did this message come from previously?

test/unit/spec/v1/EnrollSms_spec.js
```diff
@@ -31,8 +31,8 @@ Expect.describe('EnrollSms', function() {
       el: $sandbox,
       baseUrl: baseUrl,
       authClient: authClient,
-      defaultCountryCode: routerOptions.defaultCountryCode,
```
lesterchoi-okta, line 34: Why did this get removed?

lesterchoi-okta (changes requested):

test/unit/spec/v1/EnrollSms_spec.js
denysoblohin-okta: Yes, they mismatch. 
`Invalid Passcode/Answer` is the original error message that comes from server response ([this mock](https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_errorActivate.js)).
`Invalid Passcode or Answer` is the translation string, I made it different in purpose to demonstrate that the final error message should be translated.
The test passes.

test/unit/spec/v1/MfaVerify_spec.js
denysoblohin-okta: Previously this message came from `errorCauses[0].errorSummary` in mock response
https://github.com/okta/okta-signin-widget/blob/9d0b5302faa8eac667197cc20598be96c6711336/test/unit/helpers/xhr/MFA_VERIFY_invalid_password.js#L11
because of this line
https://github.com/okta/okta-signin-widget/blob/e63951ffcff0fd6843199d546c52ee332db93c0a/src/util/Util.js#L98

test/unit/spec/v1/MfaVerify_spec.js
denysoblohin-okta: Previously this message came from `errorCauses[0].errorSummary` in mock response
https://github.com/okta/okta-signin-widget/blob/9d0b5302faa8eac667197cc20598be96c6711336/test/unit/helpers/xhr/MFA_VERIFY_invalid_answer.js#L11

test/unit/spec/v1/EnrollSms_spec.js
denysoblohin-okta: In line 35 I added 
https://github.com/okta/okta-signin-widget/blob/62ef3c5118b26826655fb35f70bf3aa2a1d84391/test/unit/spec/v1/EnrollSms_spec.js#L35
to destructure all properties from `routerOptions` including `defaultCountryCode ` and also `i18n`

lesterchoi-okta (commented):
LGTM @shuowu-okta

shuowu-okta (dismissed):

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -151,6 +151,7 @@ errors.E0000064 = Password is expired and must be changed.
 errors.E0000065 = Internal error processing app metadata.
 errors.E0000066 = APNS is not configured, contact your admin
 errors.E0000067 = Factors Service Error.
+errors.E0000068 = Invalid Passcode/Answer
```
mauriciocastillosilva-okta, line 154: I see in the screenshots that the field label is "Password", and now we have this here with "Passcode/Answer"... can we choose one label and try to be consistent with it?

test/unit/spec/v1/EnrollSms_spec.js
mauriciocastillosilva-okta: Good point Lester, it is definitely better to have a conjunction like "or" instead of a slash. 

Regarding if it is just for tests, our mock can have something like "Invalid Passscode (API error)", so we know where each comes from, as opposed to a small difference like the slash vs "or"

mauriciocastillosilva-okta (commented):
Couple of comments

packages/@okta/i18n/src/properties/login.properties
denysoblohin-okta: You are right.
Error code `E0000068` itself is a generic error code used for different factors (security question, TOTP, password etc.) for Gen1 (authn API). This can cause issues with translation.

The API response contains more detailed description (specific to current factor) in `errorCauses[0].errorSummary`. But it's not translated.
Exampels in docs:
https://d28m3l9ryqsunl.cloudfront.net/docs/reference/api/authn/#response-parameters-for-verify-security-question-factor
https://d28m3l9ryqsunl.cloudfront.net/docs/reference/api/authn/#response-parameters-for-verify-totp-factor

Here is the video demonstrating current presentation of error messages:


https://github.com/okta/okta-signin-widget/assets/72614880/cd6814b8-c1a7-4272-9903-88eb2a5c3c5f



Errors messages are specific, but not translated.
If we add translation for `errors.E0000068`, it should be common for all factors.

@lesterchoi-okta

lesterchoi-okta: @mauriciocastillosilva-okta I think @denysoblohin-okta's explanation says it all. The problem is that the translation takes no parameters, and the error code is overloaded. Modifying the translation in code with find/replace is error-prone, so I think this should be solved by coming up with generic wording for the translation. Is that something you or a content-writer could help with?

PR #3562, opened 2024-03-06
Gen1: Don't dispatch custom 'okta-i18n-error' event for server errors with required argument (like E0000001)

denysoblohin-okta:
## Description:

For i18n keys `errors.E*` with a required argument (like [`errors.E0000001`](https://github.com/okta/okta-signin-widget/blob/master/packages/%40okta/i18n/src/properties/login.properties#L85) which should be translated to `Api validation failed: {0}`, or E0000007 etc.) the localisation in this line:
https://github.com/okta/okta-signin-widget/blob/e63951ffcff0fd6843199d546c52ee332db93c0a/src/util/Util.js#L102
would fail because we don't pass parameters to the `loc()` util. And a custom 'okta-i18n-error' event will be triggered.

(Gen1/2 are affected. But API errors like `E*` are typically produced by authn API used in Gen1)

Error message will be constructed from `errorSummary` in server response as a fallback:
https://github.com/okta/okta-signin-widget/blob/e63951ffcff0fd6843199d546c52ee332db93c0a/src/util/Util.js#L98
See example: 
https://github.com/okta/okta-signin-widget/blob/9f4ff4a605057639aebe5445d69e0d9dfae195fe/test/unit/spec/v1/EnrollSms_spec.js#L402

This PR changes `transformErrorXHR` util to not dispatch 'okta-i18n-error' event in cases when translation for `errors.E*` requires an argument.

Another approach would be to modify `errors.E*` translations with removing `{0}`

Error description in Okta API: https://developer.okta.com/docs/reference/error-codes/#E0000001

## PR Checklist

### Issue:

- [OKTA-700848](https://oktainc.atlassian.net/browse/OKTA-700848)

### Screenshot/Video:

For https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_error.js

Screenshot 2024-03-06 at 14 49 11

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -100,7 +100,10 @@ Util.transformErrorXHR = function(xhr) {
   // Replace error messages
   if (!_.isEmpty(xhr.responseJSON)) {
     const errorMsg = xhr.responseJSON.errorCode
-      ? loc('errors.' + xhr.responseJSON.errorCode, 'login')
+      // We don't pass parameters to the `loc()` util
+      // However some i18n keys like `errors.E0000001` require one parameter
+      // Don't dispatch custom 'okta-i18n-error' event in this case
+      ? loc('errors.' + xhr.responseJSON.errorCode, 'login', [], true)
       : undefined;
 
     if (errorMsg?.indexOf('L10N_ERROR[') === -1) {
--- a/src/util/loc.ts
+++ b/src/util/loc.ts
@@ -45,15 +45,17 @@ declare global {
  * @param  {String} key The key
  * @param  {String} [bundle="login"] The name of the i18n bundle. Defaults to "login".
  * @param  {Array} [params] A list of parameters to apply as tokens to the i18n value
+ * @param  {Boolean} [ignoreIncorrectParams] If true, a custom 'okta-i18n-error' event would not be dispatched
  * @return {String} The localized value
  */
 export const loc = function (
   key: string,
   bundleName: BundleName = 'login',
-  params: Array<string | number | boolean | unknown> = []
+  params: Array<string | number | boolean | unknown> = [],
+  ignoreIncorrectParams = false
 ) {
   const bundle = getBundle(bundleName);
-  /* eslint complexity: [2, 6] */
+  /* eslint complexity: [2, 7] */
 
   if (!bundle) {
     emitL10nError(key, bundleName, 'bundle');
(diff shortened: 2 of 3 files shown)


shuowu-okta (dismissed):

src/util/Util.js
```diff
@@ -100,7 +100,10 @@ Util.transformErrorXHR = function(xhr) {
   // Replace error messages
   if (!_.isEmpty(xhr.responseJSON)) {
     const errorMsg = xhr.responseJSON.errorCode
-      ? loc('errors.' + xhr.responseJSON.errorCode, 'login')
+      // We don't pass parameters to the `loc()` util
+      // However some i18n keys like `errors.E0000001` require one parameter
+      // Don't dispatch custom 'okta-i18n-error' event in this case
```
lesterchoi-okta, line 105: I'm unclear on whether the missing required param is expected or not. Should this be resolved on the server-side? I don't want to flood Sentry logs, but I don't want to bury valid exceptions either

denysoblohin-okta: We could use `xhr.responseJSON.errorCauses[0].errorSummary`  as a parameter to translation for keys `errors.E*`

Then eg. for this response
https://github.com/okta/okta-signin-widget/blob/master/test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_error.js
error message would be like 
```
Api validation failed: Invalid Phone Number.
```
where `Api validation failed:` comes from translation but `Invalid Phone Number.` comes from server and would not be translated. 

There is no additional error keys that comes from server and we could rely on to build fully translated message

lesterchoi-okta: So it **does** sound like the server is not returning an expected parameter. I don't think we'll go back to fix i18n in G1 error responses so I think your solution is reasonable.

https://github.com/okta/okta-signin-widget/blob/9d0b5302faa8eac667197cc20598be96c6711336/packages/%40okta/i18n/src/properties/login.properties#L85

https://github.com/okta/okta-signin-widget/blob/9d0b5302faa8eac667197cc20598be96c6711336/test/unit/helpers/xhr/MFA_ENROLL_ACTIVATE_error.js#L11

lesterchoi-okta (commented):
approval pending change to use `errorSummary` as `{0}` parameter in l10n of `errors.E0000001`

shuowu-okta (approved):

src/util/loc.ts
```diff
@@ -73,7 +75,9 @@ export const loc = function (
       return 'L10N_ERROR[' + key + ']';
     }
   } catch (e) {
-    emitL10nError(key, bundleName, 'parameters');
+    if (!ignoreIncorrectParams) {
```
mauriciocastillosilva-okta, line 78: do we have some info about what Error is thrown here? As in, we might want to emit the l10n error for all others errors we can catch here, with the exception of the Error about arguments missing

mauriciocastillosilva-okta (commented):
one comment

PR #3372, opened 2023-09-15
Add `Open Okta Verify` button during OV enrollment

SarthakDhingra-okta:
## Description:

When users choose to enroll Okta Verify using a sign-in URL after already being logged into chrome on android, they are redirected to `https://login.okta.com/` after succesful authentication and cannot complete the enrollment.

Okta-verify has a [deep link](https://github.com/atko-devices/okta-android-verify/blob/1854471c1ceb9b542e7663f1ba933b13504eb916/okta-auth-app/src/main/AndroidManifest.xml#L355) on `login.okta.com`, but deep links [do not work without a user gesture](https://developer.chrome.com/docs/multidevice/android/intents/) on android. This is why some authentication methods like a 2FA push notification would cause the issue. As such the solution is to add a user gesture prior to redirection to `login.okta.com` so the deep link can open the app.

UI mocks provided by @stevennguyen-okta 

## PR Checklist

- [x] Have you verified the basic functionality for this change?
- [x] Did you add tests, as appropriate, following our [Automated Test guidelines](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2676497890/Automated+Testing+in+the+Signin+Widget)?
- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?
- [ ] Did you verify the change by running [downstream monolith artifact](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/102897979/Sign-in+Widget+Development#Sign-inWidgetDevelopment-Instructionstocreateandrunthedownstreamartifact(d16t))? (Provide link to build below)
- [x] Does this PR include noticeable changes to the UI? (If yes, attach screenshots/video below)

### Issue:

- [OKTA-635926](https://oktainc.atlassian.net/browse/OKTA-635926)

### Reviewers:
@lesterchoi-okta @jaredperreault-okta 

### Screenshot/Video:
Demo of bug: https://okta.box.com/s/9ik685wwbwpf2mb21bbs1vasjih82oz9
Demo with code in PR: https://okta.box.com/s/msywbks2q85nx0res57zypa1mea77bsz
New View:
final_view

PR #3566, opened 2024-03-07
OKTA-698916 : fix : Fixing OV enrollment view on Android devices when…

glenfannin-okta:
## Description:
We received reports that orgs utilizing a custom domain (with no template customizations) we unable to enroll into OV from android devices. After researching this with the device platform team, we found that when the interstitial value is set to anything other than `DEFAULT` it would not render the OV launch button as expected and would not advance the user as expected. 

This value within the AutoRedirect view is only used to control whether or not the spinner displays and the sub title text that is displayed. In the case of OV enrollment on Android, this value does not matter, so we re-arranged the condition under which we display the button. More information can be found in the ticket comments. 

For additional reference, see original implementation: https://github.com/okta/okta-signin-widget/pull/3372

## PR Checklist

### Issue:

- [OKTA-698916](https://oktainc.atlassian.net/browse/OKTA-698916)

### DIFF
--- a/src/v2/view-builder/views/AutoRedirectView.js
+++ b/src/v2/view-builder/views/AutoRedirectView.js
@@ -15,6 +15,12 @@ const Body = BaseForm.extend({
     const app = this.options.appState.get('app');
     const user = this.options.appState.get('user');
 
+    // OKTA-635926: add user gesture for ov enrollment on android
+    if (Util.isAndroidOVEnrollment()) {
+      titleString = loc('oie.success.text.signingIn.with.appName.android.ov.enrollment', 'login');
+      return titleString;
+    }
+
     // If the option is not set, we treat that as being the case where we don't render the spinner.
     // This is to account for the customer hosted scenario, because by default okta-core will pass in
     // the correct value as set by the option in the Admin UI (which by default is "DEFAULT").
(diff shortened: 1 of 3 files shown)


src/v2/view-builder/views/AutoRedirectView.js
```diff
@@ -73,22 +76,18 @@ const Body = BaseForm.extend({
...
-          title: loc('oktaVerify.open.button', 'login'),
-          id: 'launch-enrollment-ov',
-          click: () => {
-            Util.redirectWithFormGet(currentViewState.href);
-          }
-        }));
-      } else {
-        this.add('<div class="okta-waiting-spinner"></div>');
-      }
+    if (Util.isAndroidOVEnrollment()) {
```
glenfannin-okta, line 79: Displaying this button should be independent of the interstitial value.

src/v2/view-builder/views/AutoRedirectView.js
```diff
@@ -15,6 +15,12 @@ const Body = BaseForm.extend({
     const app = this.options.appState.get('app');
     const user = this.options.appState.get('user');
 
+    // OKTA-635926: add user gesture for ov enrollment on android
+    if (Util.isAndroidOVEnrollment()) {
+      titleString = loc('oie.success.text.signingIn.with.appName.android.ov.enrollment', 'login');
+      return titleString;
+    }
```
glenfannin-okta, line 22: Displaying the OV enrollment subtitle should also be independent of the interstitial value.

trevoring-okta (approved):
LGTM!

shuowu-okta (approved):

PR #3579, opened 2024-03-22
Gen1: Don't translate API error E0000068 message

denysoblohin-okta:
## Description:

This PR replaces https://github.com/okta/okta-signin-widget/pull/3561

Error [E0000068](https://developer.okta.com/docs/reference/error-codes/#E0000068) should not be translated on client-side.
Because it can appear in different contexts:  https://github.com/okta/okta-signin-widget/pull/3561#discussion_r1520488474
So a translation would be too generic like "Invalid Passcode/Answer".
Instead we can just use error message from API as before. It's factor specific and translated.

## PR Checklist

### Issue:

- [OKTA-700843](https://oktainc.atlassian.net/browse/OKTA-700843)

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -99,11 +99,16 @@ Util.transformErrorXHR = function(xhr) {
   }
   // Replace error messages
   if (!_.isEmpty(xhr.responseJSON)) {
-    const errorMsg = xhr.responseJSON.errorCode
+    const { errorCode } = xhr.responseJSON;
+    const untranslatedErrorCodes = [
+      // API already provides localized and factor specific error message in errorCauses
+      'E0000068',
+    ];
+    const errorMsg = errorCode && !untranslatedErrorCodes.includes(errorCode)
       // We don't pass parameters to the `loc()` util
       // However some i18n keys like `errors.E0000001` require one parameter
       // Don't dispatch custom 'okta-i18n-error' event in this case
-      ? loc('errors.' + xhr.responseJSON.errorCode, 'login', [], true)
+      ? loc('errors.' + errorCode, 'login', [], true)
       : undefined;
 
     if (errorMsg?.indexOf('L10N_ERROR[') === -1) {
--- a/test/unit/spec/v1/EnrollCall_spec.js
+++ b/test/unit/spec/v1/EnrollCall_spec.js
@@ -506,6 +506,8 @@ Expect.describe('EnrollCall', function() {
         });
     });
     itp('shows error if error response on verification', function() {
+      // spy on emitting of CustomEvent with type 'okta-i18n-error' in `loc()` util
+      const dispatchEventSpy = jest.spyOn(document, 'dispatchEvent');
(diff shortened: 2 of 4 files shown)


src/util/Util.js
```diff
@@ -99,11 +99,16 @@ Util.transformErrorXHR = function(xhr) {
   }
   // Replace error messages
   if (!_.isEmpty(xhr.responseJSON)) {
-    const errorMsg = xhr.responseJSON.errorCode
+    const { errorCode } = xhr.responseJSON;
+    const untranslatedErrorCodes = [
```
mauriciocastillosilva-okta, line 103: nit: name it something like `noTranslationNeededErrorCodes`

test/unit/spec/v1/EnrollCall_spec.js
```diff
@@ -545,6 +547,7 @@ Expect.describe('EnrollCall', function() {
               },
             },
           ]);
+          expect(dispatchEventSpy).not.toHaveBeenCalled();
```
mauriciocastillosilva-okta, line 550: Do we need a test where we check that the event is dispatched?

mauriciocastillosilva-okta (commented):
Just a couple of minor comments. Thanks!

test/unit/spec/v1/EnrollCall_spec.js
lesterchoi-okta: don't think so? the expectation is that it's not called, i.e., that it doesn't try to report to Sentry

src/util/Util.js
lesterchoi-okta: It's a pretty locally scoped tmp var, I think it's ok as-is?

lesterchoi-okta (approved):
LGTM. See @mauriciocastillosilva-okta's comments

test/unit/spec/v1/EnrollCall_spec.js
mauriciocastillosilva-okta: I meant to say with an invalid code. As in, test that the event is dispatched when the conditions for that happen

src/util/Util.js
mauriciocastillosilva-okta: yeah, its alright. Just a nit comment

test/unit/spec/v1/EnrollCall_spec.js
mauriciocastillosilva-okta: I meant to say with an invalid code. As in, test that the event is dispatched when the conditions for that happen

mauriciocastillosilva-okta (commented):
looks good. 

Just pending if you want another test to confirm the event is dispatched when there is an invalid code. Up to you

test/unit/spec/v1/EnrollCall_spec.js
denysoblohin-okta: I think such test should be added to `courage`

PR #3585, opened 2024-03-27
Gen 2/3: Perform redirect only on visible page

denysoblohin-okta:
## Description:

This PR fixes the issue with a redirect to app on mobile devices (Chrome for Android) when page is hidden (in background).
PR adds general rule for redirects - if page is hidden, wait for [visibility change](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event) to `visible` before changing URL.

## PR Checklist

### Issue:

- [OKTA-702402](https://oktainc.atlassian.net/browse/OKTA-702402)

### DIFF
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -1117,6 +1117,28 @@ const smartCardEnrollOrVerify = {
   ],
 };
 
+const redirectOnPoll = {  
+  '/idp/idx/introspect': [
+    'authenticator-verification-email',
+    // 'authenticator-verification-data-okta-verify-push-autoChallenge-off',
+  ],
+  '/idp/idx/challenge': [
+    'authenticator-verification-okta-verify-push-autoChallenge-on',
+  ],
+  '/idp/idx/challenge/poll': [
+    'success-redirect-remediation',
+    // 'error-with-failure-redirect',
+    // 'failure-redirect-remediation',
+    // 'success',
+  ],
+  '/idp/idx/authenticators/poll': [
+    'success-redirect-remediation',
+    // 'error-with-failure-redirect',
+    // 'failure-redirect-remediation',
+    // 'success',
+  ],
+};
+
 module.exports = {
   mocks: idx
 };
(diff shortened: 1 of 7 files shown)


playground/mocks/config/responseConfig.js
```diff
@@ -1117,6 +1117,28 @@ const smartCardEnrollOrVerify = {
   ],
 };
 
+const redirectOnPoll = {
```
denysoblohin-okta, line 1120: TODO: remove

test/testcafe/spec/Redirect_spec.js
```diff
@@ -0,0 +1,101 @@
...
+  // Get internal Chrome DevTools client
+  const openedBrowsers = t.testRun.browserConnection.provider.plugin.openedBrowsers;
+  const openedBrowser = openedBrowsers[Object.keys(openedBrowsers)[0]];
+  const clients = openedBrowser.browserClient._clients;
+  const client = clients[Object.keys(clients)[0]].client;
+  return client;
+}
+
+async function openNewTab(t, url = '') {
+  // https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-createTarget
```
denysoblohin-okta, line 32: There is no way to open new tab with TestCafe API, only new window. But we can use DevTools

src/v2/controllers/FormController.ts
```diff
@@ -247,7 +247,10 @@ export default Controller.extend({
       // if Util.isAndroidOVEnrollment() returns true we use a user gesture to complete the redirect in AutoRedirectView
       if (!Util.isAndroidOVEnrollment()) {
         const currentViewState = this.options.appState.getCurrentViewState();
-        Util.redirectWithFormGet(currentViewState.href);
+        // OKTA-702402: redirect only if/when the page is visible
+        Util.executeOnVisiblePage(() => {
+          Util.redirectWithFormGet(currentViewState.href);
+        });
```
glenfannin-okta, line 253: Do we want to do this at all times? Or only when we are in a chrome browser?

glenfannin-okta: cc: @lesterchoi-okta

glenfannin-okta (commented):
Overall LGTM, other than the one question re: if we should limit this to chrome browsers.

src/v2/controllers/FormController.ts
lesterchoi-okta: AFAICT the normal active tab behavior will not be affected, and this should handle the inactive tab scenario for other supported browsers that have an "inactive tab" feature -- Edge, Firefox, Safari. Is that correct?

denysoblohin-okta: Correct

glenfannin-okta (approved):

PR #3590, opened 2024-04-01
OKTA-703334 : fix : include user info for password validation during profile enrollment

glenfannin-okta:
## Description:

The purpose of this PR is to include user information (identifier, first name, last name) in password validation client-side function when enrolling in a new profile. During profile enrollment, this user info data is provided at the time of enrollment instead of provided by the server, so conditions was added to ensure we check the appropriate data before validation.

## PR Checklist

### Issue:

- [OKTA-703334](https://oktainc.atlassian.net/browse/OKTA-703334)

### Screenshot/Video:

https://github.com/okta/okta-signin-widget/assets/97472729/c4e10894-7ca1-4d7e-b544-68e3926f86b9

Parts of username, firstname and lastname:
image

### DIFF
--- a/src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
+++ b/src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
@@ -26,8 +26,9 @@ import {
   PasswordRequirementStatus,
   PasswordValidation,
   UISchemaElementComponent,
+  UserInfo,
 } from '../../types';
-import { validatePassword } from '../../util';
+import { getUserProvidedUserInfo, validatePassword } from '../../util';
 import PasswordRequirementListItem from './PasswordRequirementListItem';
 
 const PasswordRequirements: UISchemaElementComponent<{
@@ -47,6 +48,8 @@ const PasswordRequirements: UISchemaElementComponent<{
   const password = 'credentials.newPassword' in data
     ? data['credentials.newPassword']
     : data['credentials.passcode'];
+  // Create a reference to fields that may be used in password validation (Enroll Profile)
+  const userProvidedUserInfo = getUserProvidedUserInfo(data);
 
   const [passwordValidations, setPasswordValidations] = useState<PasswordValidation>({});
   const tokens = useOdysseyDesignTokens();
@@ -65,13 +68,13 @@ const PasswordRequirements: UISchemaElementComponent<{
     return 'info';
   };
 
-  const onValidatePassword = (pw: string): void => {
+  const onValidatePassword = (pw: string, userInfoObj: UserInfo): void => {
     if (!settings) {
       setPasswordValidations({});
       return;
     }
 
-    const validations = validatePassword(pw, userInfo, settings);
+    const validations = validatePassword(pw, userInfoObj, settings);
     if (!Object.keys(validations).length) {
       setPasswordValidations({});
       return;
@@ -85,9 +88,15 @@ const PasswordRequirements: UISchemaElementComponent<{
   );
 
   useEffect(() => {
-    passwordValidationHandler((password ?? '') as string);
+    const userInfoObj = Object.keys(userInfo).length ? userInfo : userProvidedUserInfo;
+    passwordValidationHandler((password ?? '') as string, userInfoObj);
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [password]);
+  }, [
+    password,
+    userProvidedUserInfo.identifier,
+    userProvidedUserInfo.profile?.firstName,
+    userProvidedUserInfo.profile?.lastName,
+  ]);
 
   return requirements?.length > 0 ? (
     <Box
--- a/src/v3/src/transformer/profile/transformEnrollProfile.test.ts
+++ b/src/v3/src/transformer/profile/transformEnrollProfile.test.ts
@@ -40,7 +40,7 @@ describe('Enroll Profile Transformer Tests', () => {
     transaction.nextStep = {
       name: '',
     };
-    widgetProps = {};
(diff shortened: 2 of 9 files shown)


src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
```diff
@@ -85,9 +88,15 @@ const PasswordRequirements: UISchemaElementComponent<{
   );
 
   useEffect(() => {
-    passwordValidationHandler((password ?? '') as string);
+    const userInfoObj = Object.keys(userInfo).length ? userInfo : userProvidedUserInfo;
```
trevoring-okta, line 91: So `userInfo` will be defined in all remediations besides enroll profile?

src/v3/test/integration/enroll-profile-with-password.test.tsx
```diff
@@ -89,6 +89,40 @@ describe('enroll-profile-with-password', () => {
     expect(container).toMatchSnapshot();
   });
 
+  it('should display field level error when password includes username', async () => {
```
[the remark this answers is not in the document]
glenfannin-okta: Oh yes! I will add a test for first name and last name as well 🙏

src/v3/src/transformer/profile/transformEnrollProfile.test.ts
```diff
@@ -152,23 +136,23 @@ describe('Enroll Profile Transformer Tests', () => {
...
       .toBe('password.complexity.requirements.header');
-    expect((updatedFormBag.uischema.elements[2] as PasswordRequirementsElement).options?.userInfo)
-      .toEqual(mockUserInfo);
-    expect((updatedFormBag.uischema.elements[2] as PasswordRequirementsElement).options?.settings)
+    expect((updatedFormBag.uischema.elements[5] as PasswordRequirementsElement).options?.userInfo)
+      .toEqual({});
+    expect((updatedFormBag.uischema.elements[5] as PasswordRequirementsElement).options?.settings)
       .toEqual({ complexity: { minNumber: 1, minSymbol: 1 } });
-    expect((updatedFormBag.uischema.elements[2] as PasswordRequirementsElement)
+    expect((updatedFormBag.uischema.elements[5] as PasswordRequirementsElement)
```
lesterchoi-okta, line 154: Hard to read. Consider changing to [toEqual](https://jestjs.io/docs/expect#toequalvalue) or [toMatchObject](https://jestjs.io/docs/expect#tomatchobjectobject)?

src/v3/src/util/idxUtils.ts
```diff
@@ -62,6 +63,26 @@ export const getUserInfo = (transaction: IdxTransaction): UserInfo => {
...
+
+  return {
+    identifier,
+    profile: {
+      firstName,
+      lastName,
+    },
+  };
+};
+
```
lesterchoi-okta, line 85: Looks awkward. Does it make sense to expand the dot-separated key/value pairs of `userProfile` into an object? Why is it stored this way in the `FormBag`?

src/v3/src/util/idxUtils.ts
```diff
@@ -62,6 +63,26 @@ export const getUserInfo = (transaction: IdxTransaction): UserInfo => {
   return user.value as UserInfo;
 };
 
+export const getUserProvidedUserInfo = (data: FormBag['data']): UserInfo => {
```
lesterchoi-okta, line 66: Is it possible to re-use the `getUserInfo` function? Or is it very different to get data from the Transaction vs the FormBag?

glenfannin-okta: Yes it is inherently different, the transaction stores the userInfo exactly as the interface dispays (`{ identifier: ..., profile: { firstName: ..., lastName: ... }}`. Whereas the user related data the is provided when the user is in enroll-profile flow is not available in the transaction but instead set in the formBag by form entry and the keys are `userProfile.email`, `userProfile.firstName`, `userProfile.lastName`, etc. I only created this convenience function `getUserProvidedUserInfo` to avoid having to modify the the functions the rely on userInfo.

src/v3/src/util/idxUtils.ts
glenfannin-okta: The reason its stored this way is because of how the widget interfaces with the IDX payload. The object is [flattened](https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/transformer/field/transform.ts#L141) (assuming it make it easier to work with - but also mimics how Gen 2 handles form fields) into simple key/value pairs, with the key being the flattened reference to the data field. 

Keep in mind, every keystroke will update this field value entry in the form bag, so expanding it here doesn't provide much value.

src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
glenfannin-okta: But also, yes, enroll-profile does not provide userInfo because they are essentially creating the user object here, so IDX doesn't have any information to provide since this is the step that creates the user.

src/v3/src/transformer/profile/transformEnrollProfile.test.ts
glenfannin-okta: I can update this to make it read easier, however, are you sure you prefer `toMatchObject` over `toEqual`? It seems like toMatchObject would loosen the restrictions that toEqual currently provides, in which case would not fail when significant changes are made. Granted this is a unit test but if someone adds some new properties to this settings object, I feel like it would be better to fail (to notify them that they need to update tests) than to have it pass since they didn't remove any expected keys. 

What do you think?

trevoring-okta (approved):

PR #3632, opened 2024-05-23
OKTA-722502 : Same device enrollment flow updates

leemchale-okta:
## Description:

This PR adds the following updates to the Okta Verify enrollment flow in SIW Gen2 & Gen 3:

-Updates to use new translation keys
-Adds the `Or set up Okta Verify on this device` link on the select enroll channel view when Same device enrollment is available. (this feature relies on the backend change below)

Backend PR to fix `samedevice` channel option not being included in response:
https://github.com/atko-eng/okta-core/pull/95705

Figma reference: https://www.figma.com/file/ZP8O9yyMXIsnOhyt8yOKcw/OV-local-enrollment-enhancements?type=design&node-id=2728-15464&mode=design&t=mZ4a71AGYREvZH2Q-0

Design approval: https://okta.slack.com/archives/C06NW6QAHLZ/p1721323879609869?thread_ts=1721087617.224389&cid=C06NW6QAHLZ

Downstream test on monolith: https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&author=lee.mchale%40okta.com&branch=LM-OKTA-722502-downstream-test&page=1&pageSize=6&tab=main

### Gen 2 `Or set up Okta Verify on this device` link flow:

https://github.com/user-attachments/assets/cd8c8f89-791a-4698-bf51-e706dde46767

### Gen 3 `Or set up Okta Verify on this device` link flow:

https://github.com/user-attachments/assets/e41c579b-2dd2-4eb0-a24e-f06be19f2953

## PR Checklist

### Issue:

- [OKTA-722502](https://oktainc.atlassian.net/browse/OKTA-722502)


src/v2/view-builder/views/ov/SameDeviceEnrollLink.js
```diff
@@ -0,0 +1,51 @@
...
+      const sameDeviceChannel = 'samedevice';
+      const remediations = appState.get('remediations');
+      const selectEnrollmentChannelRemediation = remediations.find((remediation) => {
+        return remediation.name === RemediationForms.SELECT_ENROLLMENT_CHANNEL;
+      });
+
+      if (!selectEnrollmentChannelRemediation) {
+        return;
+      }
+      const idField = _.find(selectEnrollmentChannelRemediation.uiSchema, (schema) =>
```
trevoring-okta, line 29: Do we need to use `_.find()` and `_.filter()` throughout this file as opposed to `selectEnrollmentChannelRemediation.uiSchema.find()`?

src/v2/view-builder/views/ov/SelectEnrollmentChannelOktaVerifyView.js
```diff
@@ -28,4 +32,18 @@ const Body = BaseForm.extend({
...
+      ?.find(val => val.name === 'channel').options?.find(channel => channel.value === 'samedevice');
+    // only add this link if the samedevice channel is available in the remediation
+    if (sameDeviceChannelAvailable) {
+      this.add(new SameDeviceEnrollLink({
+        model: this.model,
+        settings: this.settings,
+        appState: this.options.appState
+      }), 'form');
+    }
+  },
```
trevoring-okta, line 48: Does the adding of the link need to be gated by the corresponding SIW FF at all? I guess the presence of the `samedevice` channel indicates that the backend FF is on?

trevoring-okta: Ah, I can see that you're removing `features.sameDeviceOVEnrollmentEnabled` in some places. Is this SIW-side FF necessary anymore?

src/v3/src/components/List/ListItemText.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Box } from '@okta/odyssey-react-mui';
+import { FunctionComponent, h } from 'preact';
+
+import { useHtmlContentParser } from '../../hooks';
+
+type ListItemTextProps = {
+  text: string;
+};
+
+const ListItemText: FunctionComponent<ListItemTextProps> = ({ text }) => {
```
trevoring-okta, line 22: What is this new component for?

playground/mocks/data/idp/idx/authenticator-enroll-ov-same-device-ios-any-security.json
```diff
@@ -124,7 +132,18 @@
...
                         "options": [
-                            { "label": "SAMEDEVICE", "value": "samedevice" }
+                          {
+                              "label": "EMAIL",
+                              "value": "email"
+                          },
+                          {
+                              "label": "SMS",
+                              "value": "sms"
+                          },
```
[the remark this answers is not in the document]
leemchale-okta: yeah so actually before in this scenario the server was _only_ returning the same device option, which is only supposed to happen when security is set to HIGH.  So technically looking at this response change you could say the default channels are now being included with `samedevice`.  So I might have to change up my wording a bit so its less confusing

src/v2/view-builder/views/ov/SameDeviceEnrollLink.js
leemchale-okta: true I think we can just use `selectEnrollmentChannelRemediation.uiSchema.find()`, ill change it

src/v2/view-builder/views/ov/SelectEnrollmentChannelOktaVerifyView.js
leemchale-okta: yeah the presence of the `samedevice` channel means the backend FF is on.  I thought it would be better for the control of whether or not to include it to be on the server side, to avoid an extra layer of logic on the client side.

the FF is still needed mostly for text variations in descriptions when the FF is on vs when its off

src/v3/src/components/List/ListItemText.tsx
leemchale-okta: This is so we can use html in list item text.  Im using it [here](https://github.com/okta/okta-signin-widget/pull/3632/files#diff-b34c14683e246f719bcf66a192426e681520d5b5678062427aee0be52535339cR400) to have bold text in a list.

src/v2/view-builder/views/ov/EnrollChannelPollDescriptionView.js
```diff
@@ -9,23 +9,32 @@ export default View.extend({
...
           <li class="switch-channel-content"></li>
         </ul>
       {{/if}}
       {{#if phoneNumber}}
         <ul class="sms-info ov-info">
-          <li>{{{i18n code="oie.enroll.okta_verify.sms.info" bundle="login" arguments="phoneNumber"}}}</li>
+          <li>
+            {{{i18n code="oie.enroll.okta_verify.sms.info.updated"
+            bundle="login" arguments="phoneNumber" $1="<span class='strong'>$1</span>"}}}
+          </li>
```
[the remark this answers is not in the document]
leemchale-okta: yeah looks like most of them are unused now.  Whether or not they can be deleted now im not sure, I was thinking to leave them in just incase we might need them in the future.

playground/mocks/data/idp/idx/authenticator-enroll-ov-same-device-ios-any-security.json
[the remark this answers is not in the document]
trevoring-okta: Ah, got it 👍

src/v2/view-builder/views/ov/SelectEnrollmentChannelOktaVerifyView.js
trevoring-okta: Makes sense, thanks for clarifying

src/v3/src/components/List/ListItemText.tsx
trevoring-okta: One last somewhat nit/comment: any idea why the gen3 spacing is so tight in the "Set up Okta Verify" view? (The view about 5 seconds into your demo vid).

This feels much tighter than normal but maybe that's how all of our lists are?

leemchale-okta: actually that part isn't a list it just uses regular description elements.  But it does look tight not sure why, it looked like that even before my change.  Ill see if I can find why the description elements dont have any bottom margin

src/v2/view-builder/views/ov/EnrollChannelPollDescriptionView.js
```diff
@@ -79,7 +99,7 @@ export default View.extend({
             {{#if sameDeviceOVEnrollmentEnabled}}
               <li>
                 <span>{{i18n code="customUri.required.content.download.title" bundle="login"}}</span>&nbsp;
-                <a href="{{deviceMap.downloadHref}}" target="_blank" id="download-ov" class="orOnMobileLink">
```
leemchale-okta, line 82: this was causing the download link to navigate to the enroll on another device flow

src/v3/src/components/List/ListItemText.tsx
leemchale-okta: I added an optional bottomMargin property to the description element to create some space

src/v3/src/components/InformationalText/InformationalText.tsx
```diff
@@ -35,13 +35,15 @@ const InformationalText: UISchemaElementComponent<{
     content,
     dataSe,
     variant,
+    bottomMargin,
```
shawyu-okta, line 38: Is allowing the UI Schema should control the margin size like this a good idea? I think that the components should have a standardized spacing so they always appear the same way. If the design has more spacing we should evaluate if the prescribed spacing in the mocks is critical or not.

shawyu-okta: I just saw the other comment below about why this was added. If the spacing should be increased it would ideally be across the board for all `InformationalText` components rather than configurable via the UI Schema.

src/v3/src/components/OpenOktaVerifyFPButton/OpenOktaVerifyFPButton.tsx
```diff
@@ -40,18 +40,19 @@ const OpenOktaVerifyFPButton: UISchemaElementComponent<{
       step,
       href,
       challengeMethod,
+      label,
     },
   } = uischema;
   const [key, setKey] = useState<number>(0);
-  const label = getTranslation(translations, 'label');
+  const btnLabel = label || getTranslation(translations, 'label');
```
shawyu-okta, line 47: Why can't this using the `translations` transformer to set the label text?

src/v3/src/components/InformationalText/InformationalText.tsx
trevoring-okta: Oh waittt, the tight spacing was because these description elements have noMargin: true on them. I don't think the bottomMargin is necessary, we can remove that and then also remove the noMargin 👍

src/v3/src/transformer/oktaVerify/transformOktaVerifyChannelSelection.ts
```diff
@@ -62,11 +62,19 @@ export const transformOktaVerifyChannelSelection: IdxStepTransformer = ({
...
+  channelSelectionElement.label = loc('oie.enroll.okta_verify.select.channel.subtitle', 'login');
   channelSelectionElement.options.format = 'radio';
   const { options: { inputMeta: { options = [] } } } = channelSelectionElement;
   channelSelectionElement.options.customOptions = options
-    .filter((opt) => opt.value !== lastSelectedChannel)
+    .filter((opt) => {
+      if (opt.value === 'samedevice') {
+        // filter out the samedevice channel so it is not displayed as a radio
+        // and set flag to true so we know to add it as a link instead
+        sameDeviceChannelAvailable = true;
```
shawyu-okta, line 73: Is there a way to avoid using a stateful flag for this? This makes it more difficult to understand the code paths that will be taken, especially when the conditional reading this value is farther down. Even if that means iterating through the `options` a second time it is probably worth it since N is going to be small.

src/v3/src/transformer/oktaVerify/transformOktaVerifyChannelSelection.ts
```diff
@@ -62,11 +62,19 @@ export const transformOktaVerifyChannelSelection: IdxStepTransformer = ({
     'authenticator.channel',
     formBag.uischema.elements as UISchemaElement[],
   ) as FieldElement;
-  channelSelectionElement.label = loc('oie.enroll.okta_verify.select.channel.description', 'login');
+  channelSelectionElement.label = loc('oie.enroll.okta_verify.select.channel.subtitle', 'login');
   channelSelectionElement.options.format = 'radio';
   const { options: { inputMeta: { options = [] } } } = channelSelectionElement;
   channelSelectionElement.options.customOptions = options
-    .filter((opt) => opt.value !== lastSelectedChannel)
+    .filter((opt) => {
```
shawyu-okta, line 69: Is it possible for `samedevice` to be the only enrollment option? If so, `channelSelectionElement.options.customOptions[0].value` below will fail

assets/sass/v2/_okta-verify.scss
```diff
@@ -58,6 +58,11 @@
     text-decoration: none;
   }
 
+  .sameDevice-setup .download-ov-link {
+    color: #007cc0;
```
trevoring-okta, line 62: Why is this link not also `color: $link-text-color`?

src/v3/src/components/InformationalText/InformationalText.tsx
leemchale-okta: oh wow it was right there dont know how I missed that haha 🤦

src/v2/view-builder/views/ov/SameDeviceEnrollLink.js
```diff
@@ -0,0 +1,51 @@
...
+        return;
+      }
+      // filter selected channel
+      const sameDeviceChannelField = selectEnrollmentChannelRemediation.uiSchema?.find((schema) =>
+        schema.name === 'authenticator.channel');
+      if (!sameDeviceChannelField) {
+        return;
+      }
+      sameDeviceChannelField.options = sameDeviceChannelField.options?.filter((option) =>
+        option.value === sameDeviceChannel);
```
shawyu-okta, line 41: Same concern here about the filtered result in the event that `samedevice` is the only channel. Will that cause issues?

src/v2/view-builder/views/ov/EnrollementChannelDataOktaVerifyView.js
```diff
@@ -122,8 +122,6 @@ export default BaseAuthenticatorView.extend({
     });
   },
   postRender() {
-    if (!this.settings.get('features.sameDeviceOVEnrollmentEnabled')) {
-      this.add(SwitchEnrollChannelLinkView, 'form');
-    }
```
shawyu-okta, line 127: Since the conditional is gone, what is the impact of this change?

src/v2/view-builder/views/ov/EnrollChannelPollDescriptionView.js
shawyu-okta: Can you point out where this unwanted behavior was coming from?

src/v2/view-builder/views/ov/EnrollChannelPollDescriptionView.js
```diff
@@ -237,4 +252,25 @@ export default View.extend({
...
+});
+
+function getDeviceMap(appState) {
+  if (!appState) {
+    return null;
+  }
+
+  const contextualData = appState.get('currentAuthenticator').contextualData;
+  let deviceMap = {};
+  if (contextualData.samedevice && contextualData.samedevice?.setupOVUrl) {
```
shawyu-okta, line 264: The `contextualData.samedevice &&` check is redundant because you already are using `.?` in the next statement. Same below in the `else if`.

src/v2/view-builder/views/ov/EnrollChannelPollDescriptionView.js
```diff
@@ -151,24 +171,19 @@ export default View.extend({
...
-        deviceMap = contextualData.samedevice;
-      } else if (contextualData.devicebootstrap && contextualData.devicebootstrap?.setupOVUrl) {
-        deviceMap = contextualData.devicebootstrap;
-      }
     }
 
     if (deviceMap.platform) {
       deviceMap.platformLC = deviceMap.platform.toLowerCase();
-      deviceMap.isDesktop = !(deviceMap.platformLC === 'ios' || deviceMap.platformLC === 'android');
+      deviceMap.isDesktop = getDeviceIsDesktop(deviceMap);
```
shawyu-okta, line 186: Can all of the `deviceMap` mutation/setting logic happen inside `getDeviceMap`? I'd expect the returned value of that function to be a "complete" deviceMap object

src/v2/view-builder/views/ov/SelectEnrollmentChannelOktaVerifyView.js
```diff
@@ -2,24 +2,28 @@ import { loc, _ } from '@okta/courage';
...
     const schemas = BaseForm.prototype.getUISchema.apply(this, arguments);
     // filter selected channel
     const channelField = _.find(schemas, (schema) => schema.name === 'authenticator.channel');
-    channelField.options = _.filter(channelField?.options, (option) =>
-      option.value !== this.options.appState.get('currentAuthenticator')?.contextualData?.selectedChannel);
+    // do not show the 'samedevice' radio option, that option is displayed as a link instead
+    channelField.options = _.filter(channelField?.options, (option) => (
+      option.value !== this.options.appState.get('currentAuthenticator')?.contextualData?.selectedChannel
+      && option.value !== 'samedevice'
+    ));
```
shawyu-okta, line 21: I think my comment in `SameDeviceEnrollLink` should actually have been here
https://github.com/okta/okta-signin-widget/pull/3632/files#r1683443342

src/v3/src/components/List/ListItemText.tsx
shawyu-okta, line 22: Not a blocker at all, but is there much value in this component compared to inserting the `useHtmlContentParser` hook in `List`? This is essentially a thin wrapper around that

src/v2/view-builder/views/ov/EnrollChannelPollDescriptionView.js
leemchale-okta: yeah [here](https://github.com/okta/okta-signin-widget/blob/d314a99c3b0344f94e0144185404d004d2b60a57/src/v2/view-builder/views/ov/EnrollChannelPollDescriptionView.js#L196) is where the click event is attached to the Set up on another device link and looks for this classname.  Since this download link was using the same classname, that event was being triggered when this link was pressed.

src/v3/src/transformer/oktaVerify/transformOktaVerifyChannelSelection.ts
leemchale-okta: yeah I should be able to refactor this to not use a state flag

src/v3/src/transformer/oktaVerify/transformOktaVerifyChannelSelection.ts
leemchale-okta: it is possible for `samedevice` to be the only channel, but that is only when OV security setting is set to HIGH.  In that case, the user would not have to option to navigate to this channel selection screen anyway, so It probably wouldnt be a problem.  But I have added a safeguard if condition just to be safe.

assets/sass/v2/_okta-verify.scss
leemchale-okta: good question, not really sure, I copied the style from orOnMobileLink.  I think I should change both of them to use `link-text-color`

src/v2/view-builder/views/ov/EnrollementChannelDataOktaVerifyView.js
leemchale-okta: now with this change the switch channel link will display on the enrollment channel page (sms, email etc. ) when the 
sameDeviceOVEnrollmentEnabled feature is enabled.
Screenshot 2024-07-18 at 3 54 11 PM

which matches the figma here: https://www.figma.com/file/ZP8O9yyMXIsnOhyt8yOKcw/OV-local-enrollment-enhancements?type=design&node-id=2728-15464&mode=design&t=mZ4a71AGYREvZH2Q-0

src/v3/src/components/List/ListItemText.tsx
leemchale-okta: was not able to use the useHtmlContentParser hook in List without getting the 
`React Hook "useHtmlContentParser" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function  react-hooks/rules-of-hooks` error

src/v3/src/components/OpenOktaVerifyFPButton/OpenOktaVerifyFPButton.tsx
leemchale-okta: refactored it to use the translation transformer instead

src/v2/view-builder/views/ov/SameDeviceEnrollLink.js
leemchale-okta: I dont think it should be an issue because the user should not have access to the channel selection screen; when `samedevice` is the only channel, that means OV security is set to HIGH.  The `Or set up Okta Verify on a mobile device` link to enroll in other channels will not display when security is set to HIGH

src/v2/view-builder/views/ov/SelectEnrollmentChannelOktaVerifyView.js
leemchale-okta: I dont think it should be an issue because the user should not have access to the channel selection screen; when samedevice is the only channel, that means OV security is set to HIGH. The Or set up Okta Verify on a mobile device link to enroll in other channels will not display when security is set to HIGH

src/v3/src/components/List/ListItemText.tsx
shawyu-okta: In `List.tsx` there already is usage of a hook `useOdysseyDesignTokens()` - why is this hook not allowed?

leemchale-okta: looks like because we use `useHtmlContentParser` within a callback for the array map on line 100.  Which in that case the hooks call order cannot be guaranteed: https://www.dhiwise.com/post/fixing-the-react-hook-cannot-be-called-inside-a-callback#:~:text=This%20is%20because%20React%20relies%20on%20the%20order%20of%20hook%20calls%20to%20track%20the%20state%20and%20effects%20associated%20with%20each%20hook%20properly.%20When%20you%20place%20a%20hook%20inside%20a%20callback%20function%2C%20the%20hook%27s%20call%20order%20can%20no%20longer%20be%20guaranteed%2C%20as%20the%20callback%20function%20may%20not%20be%20called%20on%20every%20render.

leemchale-okta: im questioning whether `useHtmlContentParser` even needs to be a hook, seems to just act like a normal function that parses some text on the spot.  Maybe we could  just treat it as a regular function?

shawyu-okta: @leemchale-okta you're right - let's move this to another directory (maybe there's something for utils) and rename it so it doesn't get picked up by the linter for hooks rules.

leemchale-okta: yeah agreed sounds good will get that refactored.

slack discussion ref: https://okta.slack.com/archives/C04NYNLLRDL/p1721768477843139

src/v3/test/integration/__snapshots__/okta-verify-enrollment-version-upgrade.test.tsx.snap
```diff
@@ -199,7 +199,7 @@ exports[`okta-verify-enrollment-version-upgrade should render form 1`] = `
                           <li
                             class="MuiListItem-root emotion-33"
                           >
-                            On your mobile device, download the Okta Verify app from the App Store (iPhone and iPad) or Google Play (Android devices).
+                            On your other device, download the Okta Verify app from the App Store® (iPhone® and iPad®) or on Google Play (Android? devices).
```
shawyu-okta, line 202: Might want to check the translation string here - "Android" has a `?` showing here

leemchale-okta: looks like this key was originally added here: https://github.com/okta/okta-signin-widget/commit/df224e3be2611c88f9816301b02c049161be84ef#:~:text=oie.enroll.okta_verify.qrcode.step1.updated%20%3D%20On%20your%20other%20device%2C%20download%20the%20Okta%20Verify%20app%20from%20the%20App%20Store%C2%AE%20(iPhone%C2%AE%20and%20iPad%C2%AE)%20or%20on%20Google%20Play%20(Android%3F%20devices).

with a "?" by mistake.  Asking globalization if this can be fixed manually

shawyu-okta (approved):
Make sure the downstream monolith is passing before merging since there's some classname changes to the Gen2 widget in this PR

PR #3650, opened 2024-06-11
Okta 719439 https loopback

Luxu2-Okta:
## Description:

Add https loopback support to SIW v2 and v3

## PR Checklist

### Issue:

- [OKTA-719439](https://oktainc.atlassian.net/browse/OKTA-719439)

### DIFF
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -417,7 +417,7 @@ const windowAuthnLoopbackFailfast = {
   ],
 };
 
-const windowAuthnHttpsLoopback = {
+const macOSAuthnHttpsLoopback = {
   '/idp/idx/introspect': [
     'identify-with-device-probing-https-loopback', // 1 (response order)
   ],
--- a/playground/mocks/config/responseConfig.js
+++ b/playground/mocks/config/responseConfig.js
@@ -417,6 +417,20 @@ const windowAuthnLoopbackFailfast = {
   ],
 };
 
+const windowAuthnHttpsLoopback = {
+  '/idp/idx/introspect': [
+    'identify-with-device-probing-https-loopback', // 1 (response order)
+  ],
+  '/idp/idx/authenticators/poll': [
+    'identify-with-device-probing-https-loopback', // 2
+    'identify-with-device-probing-https-loopback', // 3
+    'identify', // 4: as a signal of success
+  ],
+  '/idp/idx/authenticators/poll/cancel': [
+    'authenticator-verification-select-authenticator',
+  ]
+};
+
 // device probe: silent probe failed, but probing is still required
 const desktopSmartProbe = {
   '/idp/idx/introspect': [
--- a/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
+++ b/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
@@ -1,3 +1,4 @@
+/* eslint max-statements: [2, 21] */
 import { $, View } from '@okta/courage';
 import { BaseFormWithPolling } from '../internals';
 import Logger from 'util/Logger';
@@ -163,6 +164,7 @@ const Body = BaseFormWithPolling.extend({
     // This only applies to MacOS for now
     if (authenticatorHttpsDomainUrl) {
       // if https domain are included, max number of ports to be probed should be doubled
+      Logger.info('httpsDomain enabled, will probe and challenge https first');
       maxNumberOfPorts += maxNumberOfPorts;
       ports.forEach(port => {
         probeChain = probeChain
--- a/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
+++ b/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
@@ -59,7 +59,9 @@ const Body = BaseFormWithPolling.extend({
 
   doLoopback(deviceChallenge) {
     let authenticatorDomainUrl = deviceChallenge.domain !== undefined ? deviceChallenge.domain : '';
+    let authenticatorHttpsDomainUrl = deviceChallenge.httpsDomain !== undefined ? deviceChallenge.httpsDomain : '';
     let ports = deviceChallenge.ports !== undefined ? deviceChallenge.ports : [];
+    let maxNumberOfPorts = ports.length;
     let challengeRequest = deviceChallenge.challengeRequest !== undefined ? deviceChallenge.challengeRequest : '';
     let probeTimeoutMillis = deviceChallenge.probeTimeoutMillis !== undefined ?
       deviceChallenge.probeTimeoutMillis : 100;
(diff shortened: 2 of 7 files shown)


test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -350,6 +415,82 @@ test.skip
     await t.expect(identityPage.getIdentifierValue()).eql('Test Identifier');
   });
 
+test
+  .requestHooks(loopbackSuccessLogger, loopbackSuccessWithHttpsMock)('in loopback server approach, https loopback succeeds', async t => {
+    failureCount = 0;
+    // after OKTA-715718 is fixed, should use ".eql(0)" for ".lte(otherProbeCount)" assertions
+    const otherProbeCount = process.env.OKTA_SIW_GEN3 === 'true' ? 0 : 1;
```
Luxu2-Okta, line 422: bacon test [flakiness](https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&branch=OKTA-719439-https-loopback&page=1&pageSize=6&sha=30742a3e24fa0e04fc2764c5edcc2593764245fc&tab=tests&testSuite=okta-signin-widget-testcafe)

@denysoblohin-okta has a fix in this [PR](https://github.com/okta/okta-signin-widget/pull/3654)

According to my local test, it is indeed flaky, especially when I add break points

SIW V2, should be only two `/probe` call like V3
but actually, four `/probe` calls
Screenshot 2024-06-14 at 11 54 49 AM

three `/probe` calls
Screenshot 2024-06-14 at 11 50 10 AM

SIW V3, it is using simple async/await, so only two `/probe` calls, which is expected
Screenshot 2024-06-13 at 1 55 53 PM

playground/mocks/config/responseConfig.js
```diff
@@ -417,6 +417,20 @@ const windowAuthnLoopbackFailfast = {
   ],
 };
```
yannongli-okta, line 419: maybe should name it as MacOS instead of windows, since we are technically supporting this feature for MacOS

test/testcafe/spec/DeviceChallengePollView_spec.js
yannongli-okta: any specific reasons we see different number of probe calls in V2 and V3? I'm fine with asserting different number of calls but would like to understand the root cause

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
```diff
@@ -124,11 +125,19 @@ const LoopbackProbe: FunctionComponent<{ uischema: LoopbackProbeElement }> = ({
   useEffect(() => {
     const doLoopback = async () => {
       let foundPort = false;
-      // loop over each port
+
+      let urls = ports.map((port) => `${domain}:${port}`);
+      if (httpsDomain) {
+        Logger.info('httpsDomain enabled, will probe and challenge https first');
```
yannongli-okta, line 131: do we want to add same log for V2?

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
```diff
@@ -180,7 +189,7 @@ const LoopbackProbe: FunctionComponent<{ uischema: LoopbackProbeElement }> = ({
           break;
         } catch (e) {
           // only for unexpected error conditions (e.g. fetch throws an error)
-          Logger.error(`Something unexpected happened while we were checking port ${port}`);
```
yannongli-okta, line 183: do we need to make similar changes like V2 to handle max ports reached use case?

yannongli-okta (commented):
please get some review from SIW team as well, especially on V3 code and tests

test/testcafe/spec/DeviceChallengePollView_spec.js
Luxu2-Okta: I believe the reason is in this [PR](https://github.com/okta/okta-signin-widget/pull/3654)'s description. Mainly because Promisified $.ajax in BaseOktaVerifyChallengeView were not chained correctly

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
Luxu2-Okta: good point, will add

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
Luxu2-Okta: V2 already handling the max ports reached use case [here](https://github.com/okta/okta-signin-widget/pull/3650/files#diff-5383ac39ca9bdcbf8c11884ac254718b02024b36a4b438c1917e16e444e10ac8R194)
I suppose you are referring to do the same checking in [the https here](https://github.com/okta/okta-signin-widget/pull/3650/files#diff-5383ac39ca9bdcbf8c11884ac254718b02024b36a4b438c1917e16e444e10ac8R176), but I think there is no need, because we always do https first and doubled `maxNumberOfPorts`, so the `countFailedPorts` should not reach to `maxNumberOfPorts`

yannongli-okta: sorry i meant V3 here

Luxu2-Okta: I think there is no need, the V3 code is much cleaner, it just iterate with each domain+port with async/await, and it didn't check max ports before. But I will ask SIW team to double check

yannongli-okta (approved):
lgtm!

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
```diff
@@ -157,18 +160,40 @@ const Body = BaseFormWithPolling.extend({
     };
 
     let probeChain = Promise.resolve();
+    // If https domain exists, do https domain probe first
+    // This only applies to MacOS for now
+    if (authenticatorHttpsDomainUrl) {
+      // if https domain are included, max number of ports to be probed should be doubled
+      Logger.info('httpsDomain enabled, will probe and challenge https first');
+      maxNumberOfPorts += maxNumberOfPorts;
+      ports.forEach(port => {
```
shuowu-okta, line 169: move this probing logic into a function? looks similar with the one used for regular domain.

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
```diff
@@ -124,11 +125,19 @@ const LoopbackProbe: FunctionComponent<{ uischema: LoopbackProbeElement }> = ({
   useEffect(() => {
     const doLoopback = async () => {
       let foundPort = false;
-      // loop over each port
+
+      let urls = ports.map((port) => `${domain}:${port}`);
```
shuowu-okta, line 129: nit: urls -> baseUrls

shuowu-okta (approved):

src/v3/src/components/LoopbackProbe/LoopbackProbe.tsx
Luxu2-Okta: Hi @shuowu-okta , would you also take a look at this question? I think we don't need to check the `maxNumberOfPorts` like in V2, but would be better if you can double check

lesterchoi-okta (approved):

PR #3654, opened 2024-06-14
Fix flaky TestCafe OV polling test in `DeviceChallengePollView_spec`

denysoblohin-okta:
## Description:

The fix consists of 2 parts:
1. Need to comment skipped tests. For some reason (bug in TestCafe 1.x?) mocks used in `test.skip.requestHooks(xxx)` interferes with mocks in other tests - other tests in spec can use mocks from skipped test
2. Promisified `$.ajax` in [BaseOktaVerifyChallengeView](https://github.com/okta/okta-signin-widget/blob/master/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js#L17) - migrated from jQuery's Promise-like `done/fail` to native `then/catch`. With old approach, `probe` and `challenge` requests are not chained correctly (et. probing port 6512 can be started in parallel with challenging port 6511). With native Promise it's always a chain `probe 1 -> challenge 1 -> probe 2 -> challenge 2 -> ...`

Notes:
- Tested the fix 10 times on Bacon and got constant green `testcafe` task
- Tried to upgrade TestCafe to 3.x to eliminate need to comment skipped tests, but it requires upgrading Chrome version and causes other issues, so postponed..

## PR Checklist

### Issue:

- [OKTA-715718](https://oktainc.atlassian.net/browse/OKTA-715718)

### DIFF
--- a/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
+++ b/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
@@ -105,7 +105,6 @@ const Body = BaseFormWithPolling.extend({
 
     const onFailure = () => {
       Logger.error(`Something unexpected happened while we were checking port ${currentPort}.`);
-      return $.Deferred().reject();
     };
 
     const doProbing = (domainUrl) => {
@@ -164,7 +163,7 @@ const Body = BaseFormWithPolling.extend({
         .catch(onFailure);
     };
 
-    let probeChain = $.Deferred().resolve();
+    let probeChain = Promise.resolve();
 
     const handlePortProbing = (port, baseUrl, checkPortMaxFailure) => {
       probeChain = probeChain
--- a/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
+++ b/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
@@ -105,13 +105,14 @@ const Body = BaseFormWithPolling.extend({
 
     const onFailure = () => {
       Logger.error(`Something unexpected happened while we were checking port ${currentPort}.`);
+      return $.Deferred().reject();
     };
 
     const doProbing = (domainUrl) => {
       return checkPort(getAuthenticatorUrl('probe', domainUrl))
-        .done(() => {
+        .then(() => {
           return onPortFound(getAuthenticatorUrl('challenge', domainUrl))
-            .done(() => {
+            .then(() => {
               foundPort = true;
               if (deviceChallenge.enhancedPollingEnabled !== false) {
                 // this way we can gurantee that
@@ -131,7 +132,7 @@ const Body = BaseFormWithPolling.extend({
               // to make the authentication flow goes faster 
               return this.trigger('save', this.model);
             })
-            .fail((xhr) => {
+            .catch((xhr) => {
               countFailedPorts++;
               // Windows and MacOS return status code 503 when 
               // there are multiple profiles on the device and
@@ -160,10 +161,10 @@ const Body = BaseFormWithPolling.extend({
               }
             });
         })
-        .fail(onFailure);
+        .catch(onFailure);
     };
 
-    let probeChain = Promise.resolve();
+    let probeChain = $.Deferred().resolve();
 
     const handlePortProbing = (port, baseUrl, checkPortMaxFailure) => {
       probeChain = probeChain
--- a/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
+++ b/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
@@ -105,6 +105,7 @@ const Body = BaseFormWithPolling.extend({
 
     const onFailure = () => {
       Logger.error(`Something unexpected happened while we were checking port ${currentPort}.`);
+      return $.Deferred().reject();
     };
 
     const doProbing = (domainUrl) => {
--- a/test/testcafe/spec/DeviceChallengePollView_spec.js
+++ b/test/testcafe/spec/DeviceChallengePollView_spec.js
@@ -381,9 +381,14 @@ async function setupLoopbackFallback(t, widgetOptions) {
   return deviceChallengeFalllbackPage;
 }
 
+const skipTest = () => {
+  return null;
+  // return test.skip;
+};
+
 // TODO: fix quarantined test - OKTA-645716
-test.skip
-  .requestHooks(loopbackSuccessLogger, loopbackSuccessMock)('in loopback server approach, probing and polling requests are sent and responded', async t => {
+skipTest()
+  ?.requestHooks(loopbackSuccessLogger, loopbackSuccessMock)?.('in loopback server approach, probing and polling requests are sent and responded', async t => {
     const deviceChallengePollPageObject = await setup(t);
     await checkA11y(t);
     await t.expect(deviceChallengePollPageObject.getBeaconSelector()).contains(BEACON_CLASS);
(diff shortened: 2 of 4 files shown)


test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -316,9 +316,14 @@ async function setupLoopbackFallback(t, widgetOptions) {
   return deviceChallengeFalllbackPage;
 }
 
+const skipTest = () => {
+  return null;
+  // return test.skip;
+};
```
lesterchoi-okta, line 322: Explain reasoning

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
```diff
@@ -14,7 +14,13 @@ const request = (opts) => {
     method: 'GET',
     contentType: 'application/json',
   }, opts);
-  return $.ajax(ajaxOptions);
+  return new Promise((resolve, reject) => {
```
lesterchoi-okta, line 17: Explain why/how this helps

lesterchoi-okta: @shuowu-okta did you have a specific behavior-change related concern w.r.t. Promisifying this?

test/testcafe/spec/DeviceChallengePollView_spec.js
denysoblohin-okta: I used DevTools to inspect network requests and saw that test `in loopback server approach, will cancel polling when challenge errors out` mistakenly uses request hooks of previous (skipped) test (`loopbackPollTimeoutMock`)
Screenshot 2024-06-18 at 19 01 37
Request to `:2000/probe` has status 200, but should have 500.
This issue can be solved by just removing / commenting out skipped tests

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
denysoblohin-okta: In current code Promises are mixed with jQuery's Deferred  (line 159 `let probeChain = Promise.resolve();` uses Promise, ` $.ajax` returns Deferred and `done/fail` APIs are used)

Tests expect that requests are performed in chain: `probe port 2000 -> challenge 2000 (if probe OK) -> probe 6511 -> challenge 6511 -> probe 6512 -> challenge 6512 -> probe 6513 -> challenge 6513`
With current code it's not always true, eg. `probe 6511` can run in parallel with `challenge 2000`
So eg. this check
https://github.com/okta/okta-signin-widget/blob/054cb76b313d5333dd4028833a956028ac4289de/test/testcafe/spec/DeviceChallengePollView_spec.js#L435
can fail

shuowu-okta: Changing the base xhr approach may cause unexpected issue for legacy browsers. As the goal is to fix flaky bacon test, I propose to find another way that fix against test directly, for example, migrate test from e2e to unit level (testcafe -> jest).

denysoblohin-okta: See latest commit that uses `$.Deferred` instead of `Promise` and `then+catch` API instead of `done+fail` for proper chaining.
The recommended API method to use chaining is `then`: https://api.jquery.com/deferred.then/

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
```diff
@@ -153,10 +154,10 @@ const Body = BaseFormWithPolling.extend({
               }
             });
         })
-        .fail(onFailure);
+        .catch(onFailure);
     };
 
-    let probeChain = Promise.resolve();
+    let probeChain = $.Deferred().resolve();
```
yannongli-okta, line 160: will this deferred cause any degradation of UX like slower authentication speed?

denysoblohin-okta: With this change the behaviour will be same as in Gen3 - probe/challenge calls will be running not in parallel, but in chain (probing/challenging for port 2 should wait for probing/challenging of port 1 to complete). 
This can be slower.
But such behaviour in expected judging from code and tests.

If it's desired to have parallel requests, then tests should be rewritten to reflect this.
For example, this line should be removed:
https://github.com/okta/okta-signin-widget/blob/c07d759f6e0e57ea6eecb690e1ae696c4bf3aa96/test/testcafe/spec/DeviceChallengePollView_spec.js#L430

lesterchoi-okta: cc @shuowu-okta @Luxu2-Okta

Luxu2-Okta: After talking with the team, we would like to fix it, to make V2 behave similar like V3, there shouldn't be a performance concern.
We can be more cautious by using a new flag from server, but this might be an overkill

Luxu2-Okta (dismissed):
LGTM, if we want to be very cautious, we can put this fix behind a FF or KS, but this maybe an overkill

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
lesterchoi-okta: We typically isolate higher-risk changes in a patch bump on the tip so it can be easily rolled back. I think this can target `7.22.1`

shawyu-okta: How does using `$.Deferred().resolve()` differ from `Promise.resolve()` and how does it cause the probe chain to become serial?

Luxu2-Okta: the `$.Deferred().resolve()` is the jquery's promise, not sure about the very low level, but jquery promise has the `.done` method (similar as `.then`), while es6 promise doesn't.
since all the probing will use jquery $.ajax method, so I guess using `$.Deferred().resolve()` and return the rejected promise in `onFailure` will make it serial, but I will leave it to @denysoblohin-okta  to confirm

denysoblohin-okta: > How does using $.Deferred().resolve() differ from Promise.resolve()

First one returns `$.Deferred` object instead of `Promise`. This will change `probeChain` from a chain of Promises to a chain of Deferred. `Deferred` is a "thenable" object, but not a real promise.


> how does it cause the probe chain to become serial?

In current code native Promises are mixed with jQuery's Deferred:
https://github.com/okta/okta-signin-widget/blob/f37fdb316d818e46f9f93349b3b534d84ebf6831/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js#L166
https://github.com/okta/okta-signin-widget/blob/f37fdb316d818e46f9f93349b3b534d84ebf6831/src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js#L22
(`$.ajax` returns `$.Deferred`)

`probeChain` uses native Promises with `then/catch` API. But `doProbing()` returns `$.Deferred` and uses `done/fail` API.
This is a problem. To achieve proper chaining we need to change `done/fail` API to `then/catch`.


Actually we can leave `Promise.resolve()` instead of `$.Deferred().resolve()`, then we still have mixing of Promise and Deferred, but they can play well together with `then/catch` API.

However there was a concern from @shuowu-okta to use Promises in Gen2 (when my previous fix contained promisification of `$.ajax`, to use Promises instead of Deferred)

denysoblohin-okta: Rebased PR onto `7.21`

Luxu2-Okta: just curious, this will target `7.21.2` in the 2024.09.0?

Luxu2-Okta (approved):
LGTM!

lesterchoi-okta (approved):

lesterchoi-okta (changes requested):
Hold for 7.22.1 release

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
shawyu-okta: > Actually we can leave Promise.resolve() instead of $.Deferred().resolve()

Yeah, I think now that the `Promise` API is standardized and canonically polyfilled we should try to use `Promise` when we can since it's better understood and well-documented (MDN).

Luxu2-Okta (approved):

lesterchoi-okta (approved):

PR #3693, opened 2024-08-16
[G2] Clear timeout and interval before setting new ones

shawyu-okta:
## Description:

Only one `timeoutId` in `FormUtil` should be set at a time, so anytime we set a timeout, we should clear any existing ones before setting a new one. Previously, if `timeoutId` was already set (and thus a timer was already running), we would overwrite the `timeoutId`, so that timer would never get cleared by the `clearTimeout` calls in this utility. This resulted in the `setInterval` timer being set _after_ we expected to already have cleared all of the timers, resulting in the interval timer running indefinitely (which blocked the usage of the submit button).

## PR Checklist

### Issue:

- [OKTA-792679](https://oktainc.atlassian.net/browse/OKTA-792679)

### DIFF
--- a/packages/@okta/courage-for-signin-widget/package.json
+++ b/packages/@okta/courage-for-signin-widget/package.json
@@ -38,7 +38,7 @@
     "@babel/plugin-transform-shorthand-properties": "^7.16.7",
     "@babel/preset-typescript": "^7.16.7",
     "@okta/babel-plugin-handlebars-inline-precompile": "3.1.0-beta.4200.gf3d0a27",
-    "@okta/courage": "4.5.0-9093-g1a39d30",
+    "@okta/courage": "4.5.0-9096-gf1f49cb",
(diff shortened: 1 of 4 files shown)


packages/@okta/courage-for-signin-widget/package.json
```diff
@@ -38,7 +38,7 @@
     "@babel/plugin-transform-shorthand-properties": "^7.16.7",
     "@babel/preset-typescript": "^7.16.7",
     "@okta/babel-plugin-handlebars-inline-precompile": "3.1.0-beta.4200.gf3d0a27",
-    "@okta/courage": "4.5.0-8728-gd0a9dd1",
```
shawyu-okta, line 41: Candidate version, will update with firm published version before merging

lesterchoi-okta (approved):

PR #3689, opened 2024-08-09
poc use i18next

denysoblohin-okta:
## PR Checklist

- [ ] Have you verified the basic functionality for this change?
- [ ] Did you add tests, as appropriate, following our [Automated Test guidelines](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2676497890/Automated+Testing+in+the+Signin+Widget)?
- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?
- [ ] Did you verify the change by running [downstream monolith artifact](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/102897979/Sign-in+Widget+Development#Sign-inWidgetDevelopment-Instructionstocreateandrunthedownstreamartifact(d16t))? (Provide link to build below)
- [ ] Does this PR include noticeable changes to the UI? (If yes, attach screenshots/video below)

### Issue:

- [OKTA-XXXXXX](https://oktainc.atlassian.net/browse/OKTA-XXXXXX)

PR #3703, opened 2024-08-29
Localize expiration timestamps for device assurance grace period

trevoring-okta:
## Description:
- Uses `Date.toLocaleString()` (has extensive browser coverage, including IE11) to render an ISO8601 formatted string in a format that is essentially Okta's `short-with-timezone` format (e.g. 11/29/2021, 4:15 PM EDT)
- Uses the `languageCode` widget prop/setting to pass the client's locale to the function

## PR Checklist

### Issue:

- [OKTA-798277](https://oktainc.atlassian.net/browse/OKTA-798277)

### Screenshot/Video:
Screenshot 2024-09-03 at 5 52 15 PM

Screenshot 2024-09-03 at 5 53 16 PM


src/v3/src/util/formUtils.ts
```diff
@@ -310,18 +311,39 @@ export const buildEndUserRemediationMessages = (
...
       // `messages` will already be localized at this point by transactionMessageTransformer, so we can directly set
       // widgetMsg.title equal to `message`
       widgetMsg.title = message;
+    } else if (key.startsWith(GRACE_PERIOD_TITLE_KEY)) {
+      if (params.length > 0) {
+        const expiry = params[0];
+        let localizedExpiry;
+        switch (key.split(GRACE_PERIOD_TITLE_KEY)[1]) {
+          case GRACE_PERIOD_DUE_BY_DATE_SUFFIX:
+            localizedExpiry = TimeUtil.formatUnixEpochToDeviceAssuranceGracePeriodDueDate(
```
denysoblohin-okta, line 333: Proposal for Gen3: after merging [PR](https://github.com/okta/okta-signin-widget/pull/3689) to use `i18next` we can utilize [DateTime localization](https://www.i18next.com/translation-function/formatting#datetime)

playground/mocks/data/idp/idx/device-assurance-grace-period-due-by-date.json
```diff
@@ -31,11 +31,11 @@
       "type": "array",
       "value": [
           {
-            "message": "Your device doesn't meet the security requirements. Fix the issue by 08/01/2024 to prevent lockout.",
+            "message": "Your device doesn't meet the security requirements. Fix the issue by 1725667200000 to prevent lockout.",
```
denysoblohin-okta, line 34: Does server actually send such `message`?

src/v3/src/util/formUtils.ts
trevoring-okta: Oh nice, I didn't know this was inflight. Looks like it would work well, I'll make a follow up ticket and include it as a comment

trevoring-okta: Internal Ref: [OKTA-798446](https://oktainc.atlassian.net/browse/OKTA-798446)

playground/mocks/data/idp/idx/device-assurance-grace-period-due-by-date.json
trevoring-okta: Yes. We don't actually use the `message` field at all on the client side since we have the matching i18n key in our SIW bundle, but we needed to send the strings so that we can pass along the parameters. In hindsight, we could have maybe explored passing these expiry values within the `remediation` section of the IDX response, but I don't think I've seen too much precedent for doing that when the value isn't also directly tied to a form control.

I just checked and the response actually renders the timestamp parameter with commas like: 
_"Your device doesn't meet the security requirements. Fix the issue by 1,726,790,400,000 to prevent lockout."_

trevoring-okta: Overall, it's an awkward pattern but the server-side `message` field should never actually get rendered anywhere.

src/v3/src/transformer/oktaVerify/transformOktaVerifyEnrollPoll.ts
```diff
@@ -437,7 +437,7 @@ export const transformOktaVerifyEnrollPoll: IdxStepTransformer = ({
 
   const stepper: StepperLayout = {
     type: UISchemaLayoutType.STEPPER,
-    key: 'stepper_' + channelType,
+    key: `stepper_${channelType}`,
```
trevoring-okta, line 440: This was pulled in as an automatic lint fix. I'll leave it here but it's unrelated to my actual code changes

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -1656,9 +1656,7 @@ idx.error.code.access_denied.device_assurance.remediation.android.zero.trust.and
 
 # Device Assurance Grace Period
 idx.device_assurance.grace_period.title = Device assurance reminder
-# {0} is the number of days until the grace period expires
-idx.device_assurance.grace_period.warning.title.due_by_days = Your device doesn't meet the security requirements. Fix the issue within {0} days to prevent lockout.
```
trevoring-okta, line 1660: We are consolidating to a single string for the expiration message. This can get deleted in the next translation run

playground/mocks/data/idp/idx/device-assurance-grace-period-one-option.json
```diff
@@ -31,11 +31,11 @@
       "type": "array",
       "value": [
           {
-            "message": "Your device doesn't meet the security requirements. Fix the issue within 7 days to prevent lockout.",
+            "message": "Your device doesn't meet the security requirements. Fix the issue by 09/05/2024, 12:00 AM UTC to prevent lockout.",
```
trevoring-okta, line 34: Update: we will send a localized string that specifies UTC as the timezone for this `message` field to still provide a semantically correct expiry string from the server. This will cover cases where we might have a L10N error and client-side `loc()` fails.

E.g. For this specific mock, the server will send an ISO8601 format string of "2024-09-05T00:00:00.000Z" as the `param` and the `message` field will be that same string localized to UTC rather than the client's timezone -> "09/05/2024, 12:00 AM UTC"

test/testcafe/spec/DeviceAssuranceGracePeriod_spec.js
```diff
@@ -95,13 +61,14 @@ test.requestHooks(dueByDaysMock)('should render correct messaging and navigate t
 
 test.requestHooks(oneOptionMock)('should render correct messaging for grace period with one option', async t => {
   const deviceAssuranceGracePeriodPage = await setup(t);
+  const expiryLocaleString = TimeUtil.formatDateToDeviceAssuranceGracePeriodExpiryLocaleString(new Date('2024-09-05T00:00:00.000Z'));
```
shawyu-okta, line 64: Is there any way to avoid using the `TimeUtil` here in the test as well? I assume you are doing this so the string is always formatted for the TZ of the machine running the test rather than using a static string. Not ideal, but could be acceptable if we're unit testing `formatDateToDeviceAssuranceGracePeriodExpiryLocaleString` well

shawyu-okta: There might be ways to override the timezone in testcafe, although forcing a timezone in a unit test is a lot easier.

playground/mocks/data/idp/idx/device-assurance-grace-period-one-option.json
shawyu-okta: This is a good thought to have the fallback, although does the code currently handle if a call to `loc` returns `L10N_ERROR` well?

src/util/TimeUtil.js
```diff
@@ -93,4 +93,22 @@ export default {
...
+  /**
+   * @method formatDateToDeviceAssuranceGracePeriodExpiryLocaleString
+   * Conversion from a Date object to a locale string that mimics Okta's `short-with-timezone` format
+   * e.g. new Date(2024-09-05T00:00:00.000Z) -> 09/05/2024, 8:00 PM EDT
+   *
+   * @param {Date} date The Date object for the grace period expiry
+   * @param {LanguageCode} languageCode The user's language code / locale
+   * @return {String} The formatted `short-with-timezone` local string
+   */
+  formatDateToDeviceAssuranceGracePeriodExpiryLocaleString: (date, languageCode) => date.toLocaleString(languageCode, {
```
shawyu-okta, line 106: Do we need to do any input validation or checks for null/undefined?

shawyu-okta, line 106: Related but separate: We might want to add a few simple unit tests here to show our expected output

playground/mocks/data/idp/idx/device-assurance-grace-period-one-option.json
trevoring-okta: You're right, I just checked and it will still render `L10N_ERROR` as the string. The main case I'm looking to add coverage for is something going wrong in the `TimeUtil` function when creating the localized expiry param. I'll add a check to use the fallback message if the param comes back falsy because of an invalid `Date` object.

Digging into this more, I don't think this fallback message with the server-side localized string will ever actually get rendered to the SIW because the generic transformer that localizes all IDX transaction `messages` will always run before my grace period code is hit, so the actual fallback string users might see is "Your device doesn't meet the security requirements. Fix the issue by 2024-09-05T00:00:00.000Z to prevent lockout." with the ISO8601 string as the parameter before it's processed into the ideal format via `TimeUtil`. Regardless, I think having this be the server-side `message` in the IDX payload is better than what I originally had, which was a Unix timestamp

test/testcafe/spec/DeviceAssuranceGracePeriod_spec.js
trevoring-okta: Yeah looks like TestCafe doesn't have a built-in way to enforce a specific timezone. During my first pass at this PR I was struggling to set a hard-coded timezone for Jest, but I just found a solution that works. Will add some unit tests

src/util/TimeUtil.js
trevoring-okta: Will move the check for an invalid `Date` object in here 👍 I just checked and `toLocaleString()` handles `null`/`undefined` values for the locale param so we shouldn't need to validate `languageCode`

trevoring-okta: Actually looks like `toLocaleString()` only handles `undefined`. Will add some validation for other cases

src/util/TimeUtil.js
```diff
@@ -93,4 +93,31 @@ export default {
...
+   */
+  formatDateToDeviceAssuranceGracePeriodExpiryLocaleString: (date, languageCode) => {
+    try {
+    // Invalid Date objects will return NaN for valueOf()
+      return date && !isNaN(date.valueOf()) && languageCode !== null ? date.toLocaleString(languageCode, {
+        year: 'numeric',
+        month: '2-digit',
+        day: '2-digit',
+        hour: 'numeric',
+        minute: 'numeric',
```
lesterchoi-okta, line 114: I thought you were truncating the `minute` -- or maybe you meant you'd round down? I thought `minute: 'numeric'` would show a single digit for values <10m after the hour, but it seems to be overridden by one of the other values

lesterchoi-okta (dismissed):

src/util/TimeUtil.js
trevoring-okta: Ah, yeah we're rounding down, though I actually forgot to enforce that. I looked up why `numeric` doesn't show a single digit value and apparently that happens for specific combinations of these options, but to be safe I'll use '2-digit' for `hour` and `minute`

lesterchoi-okta (approved):

shawyu-okta (approved):

PR #3718, opened 2024-09-13
OKTA-798570 : Enable samedevice enrollment page polling test

leemchale-okta:
## Description:

This PR updates the `EnrollAuthenticatorOktaVerify_spec.js` testcafe test to expect polling from the OV samedevice enrollment page.

Risk level: Low (only test change)

## PR Checklist

### Issue:

- [OKTA-798570](https://oktainc.atlassian.net/browse/OKTA-798570)

### DIFF
--- a/test/testcafe/spec/EnrollAuthenticatorOktaVerify_spec.js
+++ b/test/testcafe/spec/EnrollAuthenticatorOktaVerify_spec.js
@@ -35,6 +35,9 @@ import xhrAuthenticatorEnrollEnableBiometricsSMS from '../../../playground/mocks
 
 import xhrSuccess from '../../../playground/mocks/data/idp/idx/success.json';
 
+const POLLING_INTERVAL = 4000;
+const RESEND_REMINDER_PROMPT_INTERVAL = 30000
+
 const logger = RequestLogger(/introspect|poll|send|enroll/, {
   logRequestBody: true,
   stringifyRequestBody: true,
@@ -397,7 +400,7 @@ test.requestHooks(logger, enrollViaQRcodeMocks1)('should be able to enroll via q
     await t.removeRequestHooks(enrollViaQRcodeMocks1);
     await t.addRequestHooks(enrollViaQRcodeMocks2);
   }
-  await t.wait(4000);
+  await t.wait(POLLING_INTERVAL);
   // V3 - higher poll requests ~ 7
   if (!userVariables.gen3) {
     await t.expect(logger.count(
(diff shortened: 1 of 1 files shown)


test/testcafe/spec/EnrollAuthenticatorOktaVerify_spec.js
```diff
@@ -989,12 +985,13 @@ test
     await t.expect(await enrollOktaVerifyPage.returnToAuthenticatorListLinkExists()).ok();
     await t.expect(await enrollOktaVerifyPage.signoutLinkExists()).ok();
 
-    // expect no polling for same device page
+    // expect polling for same device page
     if (!userVariables.gen3) {
+      await t.wait(4000);
```
trevoring-okta, line 990: So we don't expect polling for gen3? It looks like we were originally gating `enrollSameDeviceMocks` w/ a request to `/challenge/poll` behind `gen3` variable above, though?

trevoring-okta: Also, is the 4 sec `wait()` an arbitrary number or some standardized interval? Regardless, I'd suggest making this a constant at the top of the spec since I just took a peek and it seems like it's used at least 1 other time

trevoring-okta: I also see `wait(30000)` used twice, if that is some sort of standardized interval it might make sense to make that a constant as part of this PR as well

leemchale-okta: yeah gen 3 was actually always expecting polling while only gen 2 wasn't, thats why it was gated.  But if I recall correctly the polling behavior between gen 2 and gen 3 is different, with gen 3 resulting in different poll counts after 4 seconds thats why only gen 2 polling was being tested here.

leemchale-okta: 4 seconds is the standard interval between polls, so yeah good call ill make that a constant.  Yeah the wait(30000) is for the 'send again' reminder prompt ill make that a constant as well.

trevoring-okta (approved):

test/testcafe/spec/EnrollAuthenticatorOktaVerify_spec.js
```diff
@@ -35,6 +35,9 @@ import xhrAuthenticatorEnrollEnableBiometricsSMS from '../../../playground/mocks
 
 import xhrSuccess from '../../../playground/mocks/data/idp/idx/success.json';
 
+const POLLING_INTERVAL = 4000;
+const RESEND_REMINDER_PROMPT_INTERVAL = 30000
```
shawyu-okta, line 39: Is there any way we can shorten the polling interval in the mock file without causing issues in this test?

trevoring-okta:
@leemchale-okta Any reason why this hasn't been merged yet?

leemchale-okta:
> @leemchale-okta Any reason why this hasn't been merged yet?

Yeah this was blocked because I believe the channel behavior was possibly going to change, but now the behavior is going to remain how it is, so I will merge this soon.

test/testcafe/spec/EnrollAuthenticatorOktaVerify_spec.js
leemchale-okta: Looks like it will cause issues in the tests; seems that there are tests that rely on the polling interval being 4 seconds.  Shortening the polling interval makes some tests poll too early, which will make them continue to another flow prematurely and causing assertions to fail.  I think the api mock variables would need to be refactored to do this.

trevoring-okta (approved):

PR #3720, opened 2024-09-23
fix: update subtitle and footer with legal terms

praveendurai-okta:
## Description:

Update subtitle and add terms of use in the footer for persona.

## PR Checklist

### Issue:

- [OKTA-808095](https://oktainc.atlassian.net/browse/OKTA-808095)

#### GEN 3 
Screenshot 2024-09-24 at 10 04 07 PM

Screenshot 2024-09-24 at 8 56 23 PM

#### GEN 2
Screenshot 2024-09-24 at 9 32 12 PM

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&author=praveen.durai%40okta.com&branch=d16t-okta-signin-widget-c8b1f89-66f315ed&page=1&pageSize=6&tab=main


src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.ts
```diff
@@ -44,6 +49,37 @@ export const transformIdvIdpAuthenticator: IdxStepTransformer = ({
...
+        $1: {
+          element: 'a',
+          attributes: {
+            class: 'terms-of-use', href: 'https://withpersona.com/legal/terms-of-use', target: '_blank', rel: 'noopener noreferrer',
+          },
+        },
+        $2: {
+          element: 'a',
+          attributes: {
+            class: 'privacy-policy', href: 'https://withpersona.com/legal/privacy-policy', target: '_blank', rel: 'noopener noreferrer',
```
praveendurai-okta, line 67: new tab icon will be done as a followup.

src/v2/view-builder/views/idp/RedirectIdvView.js
```diff
@@ -1,6 +1,22 @@
-import { loc } from '@okta/courage';
+import { loc, View } from '@okta/courage';
 import { BaseIdPAuthenticatorBody, BaseIdpAuthenticatorView } from './BaseIdpAuthenticator';
 import { FORMS } from 'v2/ion/RemediationConstants';
+import hbs from '@okta/handlebars-inline-precompile';
+import AuthenticatorFooter from 'v2/view-builder/components/AuthenticatorFooter';
+
+const PrivacyTermsFooterView = View.extend({
+  tagName: 'p',
+  className: 'margin-top-30',
```
trevoring-okta, line 9: I'd discourage you from using a `className` just to add `margin` here. I took a look and we don't follow that pattern anywhere in gen2 code and we also prefer more semantic class names if we _have_ to use them.

Could you recreate the spacing you're looking for by adding the CSS you need to either a new .scss file in `assets/sass/v2` or maybe the existing `_footer.scss` file?

Maybe something like replacing the `class="margin-btm-5"` with `class="redirect-idv-footer"` and then defining something like `.redirect-idv-footer {
  margin: 20px 0 5px;
}`
This would allow you to be more exact with your margins because it seems like `30px` is a bit larger than the mockups you shared

src/v2/view-builder/views/idp/RedirectIdvView.js
```diff
@@ -1,6 +1,22 @@
...
+  template: hbs`
+      <p class="margin-btm-5">{{i18n code="oie.idv.idp.description.termsOfUse" bundle="login" 
+        $1="<a href='https://withpersona.com/legal/terms-of-use' class='terms-of-use inline-link' 
+          target='_blank' rel='noopener noreferrer'>$1</a>"
+        $2="<a href='https://withpersona.com/legal/privacy-policy' class='privacy-policy inline-link'
```
trevoring-okta, line 14: If `.terms-of-use` and `.privacy-policy` are only used for tests, let's change them to `data-se`

src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.ts
```diff
@@ -44,6 +49,37 @@ export const transformIdvIdpAuthenticator: IdxStepTransformer = ({
...
+  const termsOfUseDescription: DescriptionElement = {
+    type: 'Description',
+    contentType: 'subtitle',
+    options: {
+      variant: 'subtitle1',
+      content: loc('oie.idv.idp.description.termsOfUse', 'login', undefined, {
+        $1: {
+          element: 'a',
+          attributes: {
+            class: 'terms-of-use', href: 'https://withpersona.com/legal/terms-of-use', target: '_blank', rel: 'noopener noreferrer',
```
trevoring-okta, line 61: Same comment as earlier about changing `.terms-of-use` and `.privacy-policy` to be `data-se` if they're for testing only

src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.ts
```diff
@@ -58,7 +94,36 @@ export const transformIdvIdpAuthenticator: IdxStepTransformer = ({
     },
   };
 
-  uischema.elements = [titleElement, descriptionElement, submitButton];
+  const divider: DividerElement = {
+    type: 'Divider',
+  };
```
trevoring-okta, line 99: Just curious why only gen3 mockups have a divider?

src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.ts
```diff
@@ -58,7 +94,36 @@ export const transformIdvIdpAuthenticator: IdxStepTransformer = ({
...
+
+  uischema.elements = [
+    titleElement,
+    descriptionElement,
+    submitButton,
+    divider,
+  ];
+
+  if (shouldShowCancelLink(widgetProps?.features)) {
```
trevoring-okta, line 108: Nice catch here to enforce the `shouldShowCancelLink()` logic. If you look at `transformCancelButton()` we also assert that the transaction has `cancel` [as an available step](https://github.com/okta/okta-signin-widget/blob/8c660a4277d5113e81efe9c9a70cdeaedde32955/src/v3/src/transformer/button/transformCancelButton.ts#L33-L34). 

Could you add that same assertion here before hardcoding in `step: 'cancel'`

src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.ts
```diff
@@ -58,7 +94,36 @@ export const transformIdvIdpAuthenticator: IdxStepTransformer = ({
...
+    const cancelLink: LinkElement = {
+      type: 'Link',
+      contentType: 'footer',
+      options: {
+        label: loc('goback', 'login'),
+        isActionStep: true,
+        step: 'cancel',
+      },
```
trevoring-okta, line 116: Let's also add in `dataSe: 'cancel'` here to be consistent with [transformCancelButton](https://github.com/okta/okta-signin-widget/blob/8c660a4277d5113e81efe9c9a70cdeaedde32955/src/v3/src/transformer/button/transformCancelButton.ts#L45)

src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.ts
praveendurai-okta: In persona it will always be `cancel`. So no need of additional check. Link will be shown/hidden only based on `shouldShowCancelLink`. 
where as `transformCancelButton` is a common code used in multiple places so they had to do additional check.

src/v2/view-builder/views/idp/RedirectIdvView.js
praveendurai-okta: `margin-btm-*` `margin-top-*` are commonly used in okta-ui so i have used the same :)

updated the code to use custom class.

src/v2/view-builder/views/idp/RedirectIdvView.js
```diff
@@ -9,16 +25,19 @@ const Body = BaseIdPAuthenticatorBody.extend({
...
 });
 
 export default BaseIdpAuthenticatorView.extend({
-  Body
+  Body,
+  Footer: AuthenticatorFooter.extend({
+    postRender() {
+      this.add(PrivacyTermsFooterView);
+    }
+  }),
```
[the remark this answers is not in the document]
praveendurai-okta: you are right. i will change to `BaseFooter`

src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.ts
praveendurai-okta: have asked designer. stay tuned :) will do it as follow up if needed.

trevoring-okta (dismissed):
LGTM, as long as you're fine w/ adding a divider and the new tab icons in a follow up as you've said 👍

trevoring-okta (approved):

PR #3747, opened 2024-11-07
fix: ignore form errors for removed ov challenge view

shuowu-okta:
## PR Checklist

### Issue:

- [OKTA-820509](https://oktainc.atlassian.net/browse/OKTA-820509)

### Screenshot/Video:

Setup:

Add 3s waiting time before calling `/challenge` request in BaseOktaVerifyChallengeView to simulate failed `/challenge` request after cancelling

Before:

https://github.com/user-attachments/assets/07459b38-e81f-4d1e-bad3-c9f6f4e3c843

After:

https://github.com/user-attachments/assets/fbeb4fd5-94b6-40a7-a456-73f86b1fa7d7


src/v2/controllers/FormController.ts
```diff
@@ -209,7 +210,10 @@ export default Controller.extend({
     } else {
       error = new ConfigError(`Invalid action selected: ${actionPath}`);
       this.options.settings.callGlobalError(error);
-      await this.showFormErrors(this.formView.model, error, this.formView.form);
+      // only show form errors when not intentionally ignored from actionParams
+      if (!actionParams[ACTION_PARAMS_IGNORE_FORM_ERRORS]) {
+        await this.showFormErrors(this.formView.model, error, this.formView.form);
+      }
```
Luxu2-Okta, line 216: How about we put a logger in a `else` here

src/v2/view-builder/internals/BaseOktaVerifyChallengeView.js
```diff
@@ -54,6 +57,7 @@ const Body = BaseFormWithPolling.extend({
 
   remove() {
     BaseFormWithPolling.prototype.remove.apply(this, arguments);
+    this.removed = true;
     this.stopProbing();
     this.stopPolling();
   },
```
Luxu2-Okta, line 63: I guess this `remove` method will be called when this courage component is removed from the DOM.
correct me if I am wrong, when the `remove` is called, it only happen in two cases
1. loopback is successful, SIW transit to next view
2. user clicked the cancel (or some other user actions), that SIW load other views

So when 2 happens, we stopped polling, and as consequence, also no need to show the error. I am still not SIW expert, but this apporach makes sense to me

shuowu-okta: Yes, based on the investigation, probe/challenge requests happen in parallel, it can be pretty hard to tell when to stop/abort certain requests, it's ok to let the request fail, in the end what we want is to prevent error message display in another view (after cancel).

src/v2/controllers/FormController.ts
shuowu-okta: Looks like global error/log should be tracked [here](https://github.com/okta/okta-signin-widget/pull/3747/files#diff-72a0e868d27f4dfd26d36bcba1d860aa25999401fa00788f3ac694304043dd4dR212) already

jaredperreault-okta (approved):

src/v2/controllers/FormController.ts
shawyu-okta: Can we change the logic here to `return` early if `ignoreFormErrors=true` instead of the negative logic?

src/v2/controllers/FormController.ts
```diff
@@ -152,6 +153,9 @@ export default Controller.extend({
   // eslint-disable-next-line max-statements
   async handleInvokeAction(actionPath = '', actionParams = {}) {
     const { appState, settings } = this.options;
+    const ignoreFormErrors = (actionParams as any)[ACTION_PARAMS_IGNORE_FORM_ERRORS];
+    // prevent passing ACTION_PARAMS_IGNORE_FORM_ERRORS in network payload
+    delete actionParams[ACTION_PARAMS_IGNORE_FORM_ERRORS];
```
shawyu-okta, line 158: Instead of doing this explicitly just for the ignore form errors params, can we do one of two things to make this more maintainable/scalable?

1. Create a list of actionParams constants that are allowed to be passed to `invokeAction` and filter the params with the list before calling `invokeAction`
2. Nest the current `actionParams` values inside a property that indicates the params we want to send versus other params that shouldn't be sent. i.e.:
```ts
type IdxRequestParamKeys = 'reason' | 'statusCode';

type ActionParams = {
  // this type could be made even stricter by setting allowed values for each possible key (e.g. `reason` can only be one of 3 values, `statusCode` is a number
  idxRequestParams: Record<IdxRequestParamKeys,string>;
  ignoreFormErrors: boolean;
}
```

src/v2/controllers/FormController.ts
shuowu-okta: @shawyu-okta Updated

src/v2/controllers/FormController.ts
shuowu-okta: @shawyu-okta  Updated by passing ignoreFormErrors as a separate arguments during `appState.trigger('invokeAction')` to avoid specific param handling for `actionParams`

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -361,6 +363,41 @@ const LoginHintAppLinkMock = RequestMock()
...
+    'access-control-allow-origin': '*',
+    'access-control-allow-headers': 'X-Okta-Xsrftoken, Content-Type'
+  })
+  .onRequestTo({ url: /challenge/, method: 'POST'})
+  .respond((req, res) => {
+    return new Promise((resolve) => setTimeout(function() {
+      res.statusCode = '400';
+      res.headers['access-control-allow-origin'] = '*';
+      res.headers['access-control-allow-headers'] = 'X-Okta-Xsrftoken, Content-Type';
+      res.statusCode = '400';
```
shawyu-okta, line 395: nit: this is set already

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -16,6 +17,7 @@ import identifyWithLaunchAppLink from '../../../playground/mocks/data/idp/idx/id
 import userIsNotAssignedError from '../../../playground/mocks/data/idp/idx/error-user-is-not-assigned.json';
 
 const BEACON_CLASS = 'mfa-okta-verify';
+const EARLY_CANCEL_CHALLENGE_REQUEST_WAIT_TIME = 4000;
```
shawyu-okta, line 20: Do we need to necessarily wait this long?

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -848,3 +885,17 @@ test
...
     await t.expect(getPageUrl()).contains('login_hint='+encodeURIComponent(username));
   });
+
+test
+  .requestHooks(loopbackEarlyCancelChallengeErrorLogger, loopbackEarlyCancelChallengeErrorMock)('expect no error message in view', async t => {
+    const deviceChallengePollPageObject = await setup(t);
+    await t.expect(deviceChallengePollPageObject.getFormTitle()).eql('Verifying your identity');
+    
+    await deviceChallengePollPageObject.clickCancelAndGoBackLink();
+    await t.wait(EARLY_CANCEL_CHALLENGE_REQUEST_WAIT_TIME + 1000); // wait for delayed challenge request
```
shawyu-okta, line 895: Add an assertion to ensure that the `POST /challenge/` request was actually made . You should be able to get that information from `loopbackEarlyCancelChallengeErrorLogger`.

shawyu-okta (commented):
Overall lgtm, can you document your testing steps and attach a before/after screenshot or video in the PR description?

test/testcafe/spec/DeviceChallengePollView_spec.js
shuowu-okta: added at the end of test

test/testcafe/spec/DeviceChallengePollView_spec.js
shuowu-okta: not sure what's the issue with testcafe, but if I set a shorter wait time, the title from first form cannot be asserted correctly.

shawyu-okta (approved):

PR #3761, opened 2024-12-09
Feat: Add support for new idv type CLEAR and Incode

praveendurai-okta:
## Description:

This is incremental development. Changes are behind the FF in backend.

Add new IDV Type (Clear and Incode) in SIW both V2 and V3 versions.

Note: `CLEAR` is always CAPS for branding purpose.

## PR Checklist

### Issue:

- [OKTA-815582](https://oktainc.atlassian.net/browse/OKTA-815582)

#### Incode
Screenshot 2024-12-10 at 11 38 37 AM
#### Clear
Screenshot 2024-12-10 at 11 36 01 AM

#### Incode
Screenshot 2024-12-10 at 11 47 09 AM

#### Clear

Screenshot 2024-12-10 at 11 48 46 AM


src/v3/src/util/idxUtils.ts
```diff
@@ -462,6 +468,33 @@ export const isValidPhoneMethodType = (
...
+      break;
+    case 'INCODE':
+      termsOfUse = 'https://incode.id/terms';
+      privacyPolicy = 'https://incode.id/privacy';
+      break;
+    default:
+      termsOfUse = '';
+      privacyPolicy = '';
+      break;
+  }
```
trevoring-okta, line 494: nit: From an accessibility POV, I believe having a `default` case with `termsOfUse = undefined` and `privacyPolicy = undefined` is better than empty strings. Empty strings will make the links focusable for screenreaders, but navigate the user to a nonsensical destination, while `undefined` will prevent the links from being focusable and reduce confusion.

trevoring-okta: Technically, `termsOfUse` and `privacyPolicy` are already `undefined` since they start out uninitialized, but it would be nice to explicitly set them to `undefined` here. A complete solution to this should probably also remove the `inline-link` class from the `<a/>` elements, but I'll leave that up to you on whether it's worth the LOE for something that really should never happen

src/v2/view-builder/utils/IdpUtil.js
```diff
@@ -0,0 +1,31 @@
...
+  case 'CLEAR':
+    termsOfUse = 'https://www.clearme.com/member-terms';
+    privacyPolicy = 'https://www.clearme.com/privacy-policy';
+    break;
+  case 'INCODE':
+    termsOfUse = 'https://incode.id/terms';
+    privacyPolicy = 'https://incode.id/privacy';
+    break;
+  }
+  return {
```
trevoring-okta, line 27: See my comment in the equivalent gen3 code about standardizing on a `default` case that explicitly returns `undefined` for both `termsOfUse` and `privacyPolicy`

trevoring-okta (commented):
Generally LGTM! One small suggestion 👍

PR #3764, opened 2024-12-13
[Gen3] Remove empty optional fields from form data on submission

trevoring-okta:
## Description:
- #2412 introduced explicit logic to remove optional fields with empty strings from the payload for profile enrollment views, but this was missed in gen3
- Adds logic at the generic transformer level to trim and remove empty/whitespace-only strings for optional fields since there is no downside that I am aware of to making this logic apply to all flows (even if profile enrollment is the only flow that this bug currently exists in)
- Updates the spec from #2412 that intended to test the behavior of entering text into an optional field — which adds it to the data schema in the payload — and then deleting that text to get an empty string. The spec actually only directly entered an empty string via TestCafe, which wasn't properly replicating this scenario and giving us a false positive in gen3. Instead, we should explicitly enter a value and then delete it/replace it with an empty or whitespace-only string
- #2219 introduced a special case for the update profile view because the backend requires that the `userProfile.secondEmail` field always have data, so new gen3 logic respects that

## PR Checklist

### Issue:
- [OKTA-834905](https://oktainc.atlassian.net/browse/OKTA-834905)

#### Before
https://github.com/user-attachments/assets/a9cd65c6-d1ac-405a-b104-fe0be34fe0ba

#### After
https://github.com/user-attachments/assets/e04137c8-f87d-4d31-8b97-534d1030ddea

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&author=trevor.ing%40okta.com&branch=ti-OKTA-834905-d16t&page=1&pageSize=6&sha=06f5dbc6df9aed0b6bb6c785e3c30c5dc23a6216&tab=main

### DIFF
--- a/src/v3/src/transformer/field/transform.ts
+++ b/src/v3/src/transformer/field/transform.ts
@@ -136,11 +136,10 @@ export const transformStepInputs = (
 
   const { inputs = [] } = step;
 
-  return inputs
-    .reduce((acc: Input[], input: Input) => {
-      const flattenedInputs = flattenInputs(input);
-      return [...acc, ...flattenedInputs];
-    }, [])
+  const flattenedInputs = inputs
+    .reduce((acc: Input[], input: Input) => [...acc, ...flattenInputs(input)], []);
+
+  const formBag = flattenedInputs
     .filter((input) => input.visible === true
         || (input.visible !== false && input.mutable !== false))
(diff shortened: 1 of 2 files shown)


glenfannin-okta (approved):

jaredperreault-okta (approved):

priyakathirvel-okta (commented):
LGTM

PR #3769, opened 2024-12-30
feat: add grace period support for authenticators

TiffanyNg-okta:
## Description:

- Figma: https://www.figma.com/design/lOOqLOMOii5l6lIbYCvUKz/Authenticator-enrollment-campaigns?node-id=44-2739&t=e0Tod2vF005BrOtc-4
- https://oktawiki.atlassian.net/wiki/spaces/eng/pages/3147727639/SIW+for+Authenticator+Enrollment+Grace+Period
- Adds grace period support for v2 & v3
- Sorts authenticators object into different lists depending on combination of skip remediation + valid grace period field
- If required authenticators are being returned, we sort them into `Required Now` or `Required Soon` list
- `Required Soon` authenticators will display grace period elements that show the exact grace period date + time (in user locale and time zone) along with a relative time statement i.e. `Required in {x} days` or `Required today`
- If all the required authenticators have a valid grace period + skip remediation, we show a `Remind me later` button (backend will handle how this button redirects to optional authenticators)
- `Optional` authenticators screen should remain the same
- Re-use device assurance's `formatDateToDeviceAssuranceGracePeriodExpiryLocaleString` to display exact grace period date + time
- For v2, we add a new container (AuthenticatorEnrollOptionsContainer) to house the different lists as we can now have more than one list shown on the view
- V3 was able to make use of existing functionality so no new files we needed outside of some refactoring. 
- We are opting to NOT round down to nearest hour like Device Assurance Grace Period

## PR Checklist

### Issue:

- [OKTA-826614](https://oktainc.atlassian.net/browse/OKTA-826614)
- [OKTA-829172](https://oktainc.atlassian.net/browse/OKTA-829172)
- [OKTA-829167](https://oktainc.atlassian.net/browse/OKTA-829167)
- [OKTA-829174](https://oktainc.atlassian.net/browse/OKTA-829174)

### Screenshot/Video:

V2 screenshots:
Screenshot 2025-01-02 at 10 25 52 AM

Screenshot 2025-01-07 at 10 34 00 AM

Screenshot 2025-01-07 at 10 47 42 AM

Screenshot 2025-01-02 at 10 30 23 AM

V3 screenshots:
Screenshot 2025-01-07 at 1 48 30 PM

Screenshot 2025-01-07 at 1 47 06 PM

Screenshot 2025-01-24 at 6 56 35 PM

Screenshot 2025-01-24 at 6 57 16 PM


src/v3/src/mocks/response/idp/idx/introspect/authenticator-enroll-grace-period-with-skip.json
```diff
@@ -0,0 +1,183 @@
+{
```
TiffanyNg-okta, line 1: Can probably remove the mock file here and everything under v3/src/mocks if preferred. I was previously using it with msw for local testing

test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
```diff
@@ -435,3 +456,82 @@ test.requestHooks(mockEnrollAuthenticatorWithCustomApp)('should load select auth
   await t.expect(selectFactorPage.getFactorSelectButtonDataSeByIndex(4, true)).eql('webauthn');
   await t.expect(await selectFactorPage.factorUsageTextExistsByIndex(4)).eql(false);
 });
+
+// re-enable with gen 3 changes
+test.meta('gen3', false).requestHooks(mockEnrollRequiredNowAndRequiredSoonAuthenticators)('should load select required now and required soon authenticator lists', async t => {
```
TiffanyNg-okta, line 461: @jakebacher-okta you'll likely want to enable gen3 here and update the selectors in the page object to run these test for v3 parity in your siw v3 changes

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.ts
```diff
@@ -64,23 +94,65 @@ export const transformSelectAuthenticatorEnroll: IdxStepTransformer = ({
...
+  if (authenticatorButtonsDueNow.length) {
+    authenticatorListElementDueNow.push({
+      type: 'AuthenticatorButtonList',
+      options: { buttons: authenticatorButtonsDueNow, dataSe: 'authenticator-enroll-list' },
+    });
+  }
+
   const skipButton: ButtonElement = {
     type: 'Button',
     label: loc('oie.optional.authenticator.button.title', 'login'),
```
TiffanyNg-okta, line 158: @jakebacher-okta were you waiting to change this to "Remind me later" in the case when authenticators have grace period + skip ?

src/util/TimeUtil.js
```diff
@@ -125,5 +129,31 @@ export default {
...
+  /**
+   * @method calculateDaysBetweenEpochTimestamps
+   * Calculate the number of days between two epoch milliseconds timestamps
+   * rounded down to the nearest day
+   *
+   * @param {Integer} epoch1 epoch timestamp in milliseconds as an integer
+   * @param {Integer} epoch2 epoch timestamp in milliseconds as an integer
+   * @return {Integer} The number of days between the two epoch timestamps
+   */
+  calculateDaysBetweenEpochTimestamps(epoch1, epoch2) {
```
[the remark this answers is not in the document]
TiffanyNg-okta: I have unit tests in TimeUtil_spec.js starting on line 155 here 🙂 : https://github.com/okta/okta-signin-widget/pull/3769/files/10a139eace839ce76ab7614508e4b65d47e746a9#diff-6f8c4193d90fd1b20ea606fe118b204009c54689dfe5dd519ffdbc5bbad68859R155

src/util/TimeUtil.js
```diff
@@ -125,5 +129,31 @@ export default {
...
+    try {
+      if (Number.isInteger(epoch1) && Number.isInteger(epoch2)) {
+        const diffInMs = Math.abs(epoch2 - epoch1);
+      
+        const msPerDay = 60 * 60 * 24 * 1000;
```
shawyu-okta, line 148: Use a constant (all caps) at the top of the file

src/util/TimeUtil.js
```diff
@@ -125,5 +129,31 @@ export default {
...
+        const days = Math.floor(diffInMs / msPerDay);
+    
+        return days;
+      } else {
+        return null;
+      }
+    } catch (e) {
+      return null;
```
shawyu-okta, line 156: nit: The `else` and the `return null` within the `catch` can be moved to the bottom of the method as the default return value.

The `catch` can just be empty if you wish to still swallow any errors

src/util/TimeUtil.js
```diff
@@ -102,14 +102,18 @@ export default {
...
     try {
     // Invalid Date objects will return NaN for valueOf()
       if (date && !isNaN(date.valueOf()) && languageCode !== null) {
-        // Round down the date to the nearest hour
-        date.setMinutes(0, 0, 0);
+        // eslint-disable-next-line max-depth
+        if (roundDownToNearestHour) {
+          // Round down the date to the nearest hour
+          date.setMinutes(0, 0, 0);
+        }
```
shawyu-okta, line 116: Would it be simpler to move the rounding down to the calling code that wants it rather than introducing a flag onto this function?

src/util/TimeUtil.js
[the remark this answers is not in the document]
shawyu-okta: Naming nit, this can be named `calculateDaysBetween(epoch1, epoch2)` since the JSDoc and parameter names are pretty clear. I assume not throwing within this function is intended, since otherwise I would suggest using errors to enforce parameter types

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -179,7 +183,54 @@ const AuthenticatorButton: UISchemaElementComponent<{
...
+                  margin: tokens.Spacing0,
+                  textAlign: 'start',
+                }}
+              >
+                {gracePeriodExpiry}
+              </Typography>
+            </Box>
+          </Box>
+        )}
+        {!gracePeriodRequiredDescription && description && (
```
shawyu-okta, line 233: Can this logic be refactored out so they are implicitly mutually exclusive rather than explicitly? (e.g. using `if/else` or moving to a function named something like `renderDescription()` that returns a component)

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.ts
```diff
@@ -33,6 +34,12 @@ const getContentDescrAndParams = (brandName?: string): TitleElement['options'] =
   return { content: loc('oie.select.authenticators.enroll.subtitle', 'login') };
 };
 
+const isGracePeriodStillActive = (expiry: string): boolean => {
+  const currentTimestampMs = new Date().getTime();
```
shawyu-okta, line 38: nit: `Date.now()`

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.ts
```diff
@@ -33,6 +34,12 @@ const getContentDescrAndParams = (brandName?: string): TitleElement['options'] =
...
+  const gracePeriod = new Date(expiry);
```
shawyu-okta, line 39: nit:
```suggestion
  const gracePeriodTimestampMs = new Date(expiry).getTime();
```

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.ts
```diff
@@ -64,39 +94,107 @@ export const transformSelectAuthenticatorEnroll: IdxStepTransformer = ({
...
 
-  uischema.elements = [
-    title,
-    informationalText,
-    description,
-    authenticatorListElement,
-    ...(skipStep ? [skipButton] : []),
-  ];
+  // 3 situations - required soon + required now, all required soon, all required now
+  // when grace periods are past required and should be treated as normal required
```
shawyu-okta, line 168: Could this be handled by sequentially building the array and assigning it to `uischema.elements` at the end? Like this:
```js
const elements = [title, description];
if (hasNow || hasSoon) {
  if (hasNow) {
    // add now heading, elements
  }
  if (hasSoon) {
    // add soon heading, elements
  }
} else {
  // add default heading, elements
}
if (skipStep) {
  // add skipButton
}
ui.elements = elements;
```

src/v3/src/transformer/selectAuthenticator/utils.ts
```diff
@@ -360,6 +376,30 @@ const formatAuthenticatorOptions = (
         AUTHENTICATOR_KEY.PHONE,
       ];
 
+      // @ts-ignore TODO: Add grace period fields to auth-js SDK https://oktainc.atlassian.net/browse/OKTA-848910
+      const date = new Date(authenticator?.gracePeriod?.expiry);
+      const gracePeriodEpochTimestampMs = !Number.isNaN(date.getTime()) ? date.getTime() : 0;
+      const currentTimestampMs = new Date().getTime();
```
shawyu-okta, line 382: nit: `Date.getTime()`

src/v2/view-builder/components/AuthenticatorEnrollOptionsContainer.js
```diff
@@ -0,0 +1,65 @@
...
+
+  className: 'authenticator-enroll-list-container',
+
+  initialize: function() {
+    const authenticatorsWithGracePeriod = [];
+    const authenticators = [];
+
+    this.options.optionItems.forEach((authenticator) => {
+      if (authenticator.relatesTo?.gracePeriod?.expiry
+        && new Date(authenticator.relatesTo?.gracePeriod?.expiry).getTime() > (new Date(Date.now()).getTime())) {
```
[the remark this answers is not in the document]
TiffanyNg-okta: No particular reason 😅 . It just started out as `new Date(`), then added `Date.now()` explicitly for tests. Updated to just `Date.now()`.

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
TiffanyNg-okta: Yes, moved it out to a `renderDescription()` with `if/else if`

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.ts
TiffanyNg-okta: Yup, updated!

src/v3/src/transformer/selectAuthenticator/utils.ts
TiffanyNg-okta: I think you meant `Date.now()` here? Updated to `Date.now()`

src/util/TimeUtil.js
TiffanyNg-okta: We would have to then do `date.setMinutes(0, 0, 0)` twice - once in v2 and once in v3 before calling this function, so it made sense to us to centralize it here especially since date logic/calculations can get tricky.

src/v3/src/transformer/selectAuthenticator/utils.ts
shawyu-okta: > I think you meant Date.now() here? Updated to Date.now()

Oh yes, my mistake

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -100,6 +104,79 @@ const AuthenticatorButton: UISchemaElementComponent<{
...
+          }}
+          data-se="authenticator-button-description"
+          aria-label={punctuate(description)}
+          dir={dir}
+          translate={noTranslate ? 'no' : undefined}
+        >
+          {description}
+        </Typography>
+      );
+  }
```
shawyu-okta, line 178: Not sure if TS complained but you should `return null` at the bottom as the default if nothing should render

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -100,6 +104,79 @@ const AuthenticatorButton: UISchemaElementComponent<{
     return <ArrowRightIcon titleAccess={ctaLabel} />;
   }
 
+  function renderDescription() {
```
shawyu-okta, line 107: Prefer arrow functions here

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.test.ts
```diff
@@ -190,4 +212,288 @@ describe('Enroll Authenticator Selector Transformer Tests', () => {
...
+            expiry: '2045-09-27 18:00:00.000.',
+          },
+        },
+      } as unknown as IdxOption,
+      {
+        label: 'Email',
+        value: 'okta_email',
+        relatesTo: {
+          gracePeriod: {
+            expiry: '2030-09-27 18:00:00.000.',
```
shawyu-okta, line 232: I assume you are using far in the future expiration dates to try to guarantee that this test will work as you expect. I would suggest we try to use mocking of `new Date` and `Date.now` whenever possible to create realistic unit test scenarios that are also stable over the long run.

P.S. Are you not expecting this test to need to pass after Sept 2030? 😛

shawyu-okta: The same comment also applies to other mocked date objects in this PR, including ones that are in the past. We should still mock the value of the present date so that snapshots and assertions are unchanging.

src/v3/src/transformer/selectAuthenticator/utils.ts
```diff
@@ -360,6 +376,30 @@ const formatAuthenticatorOptions = (
         AUTHENTICATOR_KEY.PHONE,
       ];
 
+      // @ts-ignore TODO: Add grace period fields to auth-js SDK https://oktainc.atlassian.net/browse/OKTA-848910
+      const date = new Date(authenticator?.gracePeriod?.expiry);
+      const gracePeriodEpochTimestampMs = !Number.isNaN(date.getTime()) ? date.getTime() : 0;
```
shawyu-okta, line 381: You may be able to make this more succinct by changing the const to
```js
const dateMs = new Date(authenticator?.gracePeriod?.expiry).getTime();
```
And then you only need to call `getTime()` once

shawyu-okta: Separately, prefer positive logic for readability. You can invert the ternary to avoid the `!` negation.

src/v3/src/transformer/selectAuthenticator/utils.ts
```diff
@@ -500,10 +543,12 @@ export const getAuthenticatorVerifyButtonElements = (
 export const getAuthenticatorEnrollButtonElements = (
   authenticatorOptions: IdxOption[],
   step: string,
+  locale?: string,
```
shawyu-okta, line 546: I'm not sure if this transformer util is the right place for `locale`. It smells just a little bit to have this passed down through all of these functions when the rendering layer (`AuthenticatorButton` component) has the locale context already.

I'm thinking of two possibilities:
* We could do the locale formatting of a timestamp in the component
* We can do locale formatting in another transformer layer like we handle translations, possibly even combined into a locale transformer layer

The latter feels like it might be unnecessary complexity, but it does fit the theme of localization for that transform.

Curious if you have thoughts. We can discuss live during work hours.

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
TiffanyNg-okta: ah yeah was just following the style in the above method, updated both to use arrow functions

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
TiffanyNg-okta: I was seeing some lint errors that I resolved, but updated to make it return an explicit null

src/util/TimeUtil.js
shawyu-okta: Makes sense, I think there's some refactoring that could be done here to improve the code so we don't have to do this with a flag, but that's just a result of these two features sharing a similar function that creates the awkward code and it's out of scope for this change. I think the reuse is still worthwhile even with the added flag so let's go ahead with this and revisit in the future.

src/v3/src/transformer/selectAuthenticator/utils.ts
```diff
@@ -360,6 +376,30 @@ const formatAuthenticatorOptions = (
...
+      }
+
+      const gracePeriodExpiry = (hasGracePeriods && locale
+        && TimeUtil.formatDateToDeviceAssuranceGracePeriodExpiryLocaleString(
+          new Date(gracePeriodEpochTimestampMs),
+          locale,
+          false,
+        )) || undefined;
+      const gracePeriodRequiredDescription = (gracePeriodExpiry
+        && getGracePeriodRequiredDescription(remainingGracePeriodDays)) || undefined;
```
shawyu-okta, line 401: Generally it's best to use `null` to represent intentionally missing values as opposed to `undefined` to represent unintentionally missing values, unless there's a reason we need to specifically expect `undefined`

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.test.ts
TiffanyNg-okta: 😄 I think this file is the only one we missed that depends on the present date? Let me know if we missed any other test.

We had already mocked `Date.now` in testcafe here: https://github.com/okta/okta-signin-widget/pull/3769/files#diff-2900b8cd411a393bd492a86b3fa5c19025d50ea6928545d641ea0341fbfacaf2R86

and v2 unit tests here: https://github.com/okta/okta-signin-widget/pull/3769/files#diff-2af053911968b4d4959d3566bd261769a8cc5324853397eb90b51f5b9052aed8R29

src/v3/src/transformer/selectAuthenticator/utils.ts
TiffanyNg-okta: hmm @jakebacher-okta was there a specific reason to have `undefined` here? I'm guessing no, but just wanted to check before updating to `null`

justinabrokwah-okta (dismissed):
LGTM! Great, thorough work on this

test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
```diff
@@ -435,3 +483,105 @@ test.requestHooks(mockEnrollAuthenticatorWithCustomApp)('should load select auth
   await t.expect(selectFactorPage.getFactorSelectButtonDataSeByIndex(4, true)).eql('webauthn');
   await t.expect(await selectFactorPage.factorUsageTextExistsByIndex(4)).eql(false);
 });
+
+// re-enable with gen 3 changes
```
shawyu-okta, line 487: Is this comment needed?

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.test.ts
```diff
@@ -190,4 +213,288 @@ describe('Enroll Authenticator Selector Transformer Tests', () => {
...
+        relatesTo: {
+          gracePeriod: {
+            expiry: '2020-09-27 18:00:00.000.',
+          },
+        },
+      } as unknown as IdxOption,
+      {
+        relatesTo: {
+          gracePeriod: {
+            expiry: '2020-9-27 18:00:00.000.',
```
shawyu-okta, line 401: Not sure if the missing `0` was intentional or not

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.test.ts
```diff
@@ -190,4 +213,288 @@ describe('Enroll Authenticator Selector Transformer Tests', () => {
...
+    // @ts-ignore TODO: Add grace period fields to auth-js SDK https://oktainc.atlassian.net/browse/OKTA-848910
+    expect(updatedFormBag.uischema.elements[2].options.content).toBe('oie.setup.optional.short');
+    expect(((updatedFormBag.uischema.elements[3] as AuthenticatorButtonListElement)
+      .options.dataSe)).toBe('authenticator-enroll-list');
+    expect(((updatedFormBag.uischema.elements[3] as AuthenticatorButtonListElement)
+      .options.buttons[0] as AuthenticatorButtonElement).options.type).toBe(ButtonType.BUTTON);
+    expect(((updatedFormBag.uischema.elements[3] as AuthenticatorButtonListElement)
+      .options.buttons[0] as AuthenticatorButtonElement).label).toBe('Email');
+    expect((updatedFormBag.uischema.elements[4] as ButtonElement).options.step)
+      .toBe('skip');
```
shawyu-okta, line 443: If all grace periods expired, I wouldn't expect there to be a `skip` element. Am I reading this test incorrectly?

shawyu-okta: It seems like the first new test should have the `skip` step included, but this one wouldn't get it

TiffanyNg-okta: Discussed with @jakebacher-okta , this is intended so that we test that we're reactive to the backend response i.e. skip in response should skip button in the UI. But no, it's not likely that this is a possible/valid idx response to have the skip + expired grace periods.

test/testcafe/spec/SelectAuthenticatorForEnroll_spec.js
TiffanyNg-okta: Nope, was just a reminder for us when developing. Removed!

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.test.ts
TiffanyNg-okta: Probably not intentional, updating the timestamps to match the rest of the mocks

src/v3/src/transformer/selectAuthenticator/utils.ts
TiffanyNg-okta: Updated to `null`

TiffanyNg-okta:
Also did some quick testing with mocks in ie11:
Gen 2 screenshots:
Screenshot 2025-01-28 at 1 39 27 PM
Screenshot 2025-01-28 at 1 38 59 PM
Screenshot 2025-01-28 at 1 37 46 PM
Screenshot 2025-01-28 at 1 37 13 PM
Gen 3 screenshots:
Screenshot 2025-01-28 at 12 36 26 PM
Screenshot 2025-01-28 at 12 35 50 PM
Screenshot 2025-01-28 at 12 35 02 PM
Screenshot 2025-01-28 at 12 31 54 PM

src/v2/view-builder/components/AuthenticatorEnrollOptions.js
```diff
@@ -80,6 +92,36 @@ const AuthenticatorRow = View.extend({
     }
 
     const data = View.prototype.getTemplateData.apply(this, arguments);
+
+    const currentTimestampMs = Date.now();
+    const gracePeriodEpochTimestampMs = new Date(this.model.get('relatesTo')?.gracePeriod?.expiry).getTime();
+    if (!isNaN(gracePeriodEpochTimestampMs) && currentTimestampMs < gracePeriodEpochTimestampMs) {
```
shawyu-okta, line 98: Note that `isNaN` coerces the value to a number first, but I think in this particular usage we aren't at risk of non-numeric types being input, since the possible values of `gracePeriodEpochTimestampMs` are either `Number` or `NaN` and no other possibility.

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.test.ts
shawyu-okta: Understood. I think I'm also confused by the assertion that `oie.setup.optional.short` string is shown even in the case where no authenticators have grace period. I understand that if `skip` is available, then the description being "optional" is consistent, but in terms of realistic scenarios, it is a bit misleading to any reader of the test. If there's a way to move some of the scenarios around to be more realistic that would be best

shawyu-okta (approved):
No blocking comments. Nice work -- thanks for the changes and testing in IE11.

src/v3/src/transformer/selectAuthenticator/transformSelectAuthenticatorEnroll.ts
```diff
@@ -38,7 +38,7 @@ const getContentDescrAndParams = (brandName?: string): TitleElement['options'] =
 const isGracePeriodStillActive = (expiry: string): boolean => {
   const currentTimestampMs = Date.now();
   const gracePeriodTimestampMs = new Date(expiry).getTime();
-  return !Number.isNaN(gracePeriodTimestampMs) && currentTimestampMs < gracePeriodTimestampMs;
+  return !isNaN(gracePeriodTimestampMs) && currentTimestampMs < gracePeriodTimestampMs;
```
shawyu-okta, line 43: This is fine and no change is needed, but the Gen3 webpack configuration should automatically add new `core-js` polyfills for the browsers we target, which includes IE11. If this works correctly, any code that is _only_ in Gen3 theoretically could use `Number.isNaN` without any issues, since it will get polyfilled at compile-time.

TiffanyNg-okta: Ah, no wonder I was only seeing issues with Gen 2 😅. I had applied it here also to be safe

shawyu-okta (approved):

PR #3781, opened 2025-01-28
feat(oie): supports device assurance custom errors

trevoring-okta:
## Description:
- Adds IDX mocks and extends terminal TestCafe spec for Device Assurance Custom Error Remediation (an extension of the existing end user remediation feature)
    - Tech design for feature [here](https://oktawiki.atlassian.net/wiki/x/UIfEv) 
- Minimal source code changes needed for this feature since it just allows admins to customize the text and link for the end user remediation options
    - Updated gen3 code specifically since it was not properly rendering remediation options w/o a link (which is now possible w/ this feature) within the bullet list

## PR Checklist

### Issue:

- [OKTA-839975](https://oktainc.atlassian.net/browse/OKTA-839975)

#### Gen2 Custom Message w/ Custom Link and Custom Message w/ No Link Demo
https://github.com/user-attachments/assets/a1104875-3986-4e6d-9a9a-f38286a3355e

#### Gen3 Custom Message w/ Custom Link and Custom Message w/ No Link Demo
https://github.com/user-attachments/assets/5231072d-86d3-47dc-ba0c-5b53cb36615b

### DIFF
--- a/playground/mocks/data/idp/idx/end-user-remediation-custom-message-no-url.json
+++ b/playground/mocks/data/idp/idx/end-user-remediation-custom-message-no-url.json
@@ -1,0 +1,53 @@
+{
+  "stateHandle": "01OCl7uyAUC4CUqHsObI9bvFiq01cRFgbnpJQ1bz82",
+  "version": "1.0.0",
+  "expiresAt": "2020-05-18T20:42:23.000Z",
+
+  "intent": "LOGIN",
+  "messages": {
+    "type": "array",
+    "value": [
+      {
+        "message": "Your device doesn't meet the security requirements",
+        "i18n": {
+          "key": "idx.error.code.access_denied.device_assurance.remediation.title"
+        },
+        "class": "ERROR"
+      },
+      {
+        "message": "To sign in, make the following updates. Then, access the app again.",
+        "i18n": {
+          "key": "idx.error.code.access_denied.device_assurance.remediation.explanation_one_rule"
+        },
+        "class": "ERROR"
+      },
+      {
+        "message": "It's our company policy that your Android device cannot be rooted",
+        "i18n": {
+          "key": "device.assurance.custom.remediation.ANDROID.Okta:DeviceAssurance.Jailbreak"
+        },
+        "class": "ERROR"
+      },
+      {
+        "links": [
+          {"url": "https://okta.com/android-biometric-lock"}
+        ],
+        "message": "Enable lock screen and biometrics",
+        "i18n": {
+          "key": "idx.error.code.access_denied.device_assurance.remediation.android.use_biometric_lock_screen"
+        },
+        "class": "ERROR"
+      },
+      {
+        "links": [
+          {"url": "https://okta.com/help"}
+        ],
+        "message": "For more information, follow the instructions on the help page or contact your administrator for help",
+        "i18n": {
+          "key": "idx.error.code.access_denied.device_assurance.remediation.additional_help_default"
+        },
+        "class": "ERROR"
+      }
+    ]
+  }
+}
--- a/src/v2/ion/i18nUtils.js
+++ b/src/v2/ion/i18nUtils.js
@@ -21,6 +21,7 @@ import { FORMS, AUTHENTICATOR_KEY } from './RemediationConstants';
 import { I18N_BASE_ATTRIBUTE_ENROLL_PROFILE_MAPPINGS } from '../view-builder/views/enroll-profile/i18nBaseAttributeMappings';
 
 const WEBAUTHN_API_GENERIC_ERROR_KEY = 'authfactor.webauthn.error';
+const DEVICE_ASSURANCE_CUSTOM_REMEDIATION_I18N_KEY_PREFIX = 'device.assurance.custom.remediation.';
 
 const SECURITY_QUESTION_PREFIXES = [
   'enroll-authenticator.security_question.credentials.questionKey.',
@@ -338,6 +339,11 @@ const getMessage = (message) => {
       // The WebAuthn api error doesn't make much sense to a typical end user, but useful for developer.
       // So keep the api message in response, but show a generic error message on UI.
       return loc(WEBAUTHN_API_GENERIC_ERROR_KEY, 'login');
+    }
+
+    // Return unlocalized IDX `message.message` for admin-defined custom device assurance remediation keys
+    if (i18nKey.startsWith(DEVICE_ASSURANCE_CUSTOM_REMEDIATION_I18N_KEY_PREFIX)) {
+      return message.message;
     }
   }
 
--- a/src/v3/src/components/WidgetMessageContainer/WidgetMessageContainer.tsx
+++ b/src/v3/src/components/WidgetMessageContainer/WidgetMessageContainer.tsx
@@ -25,7 +25,7 @@ import {
   ClickEvent,
   ListStyleType,
   WidgetMessage,
-  WidgetMessageLink,
+  WidgetMessageOption,
 } from '../../types';
 import { parseHtmlContent } from '../../util';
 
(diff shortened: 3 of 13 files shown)


src/v2/ion/i18nUtils.js
```diff
@@ -339,6 +340,11 @@ const getMessage = (message) => {
       // So keep the api message in response, but show a generic error message on UI.
       return loc(WEBAUTHN_API_GENERIC_ERROR_KEY, 'login');
     }
+
+    // Return unlocalized IDX `message.message` for admin-defined custom device assurance remediation keys
+    if (i18nKey.startsWith(DEVICE_ASSURANCE_CUSTOM_REMEDIATION_I18N_KEY_PREFIX)) {
+      return message.message;
+    }
```
trevoring-okta, line 347: All customized messages will come w/ this prefix and obviously cannot be localized

shawyu-okta: How will older versions of the widget behave if they receive a DA custom remediation message?

trevoring-okta: Functionally, they're fine. They'll fall through to the `Logger.warn()` and then still return `message.message`. This is just to be more explicit about returning the unlocalized fallback message

src/v3/src/components/WidgetMessageContainer/WidgetMessageContainer.tsx
```diff
@@ -46,6 +46,46 @@ const WidgetMessageContainer: FunctionComponent<{
...
+      return (
+        // @ts-expect-error error due to variant type applied, can be ignored
+        <MuiLink
+          component="button"
+          role="link"
+          onClick={(event: ClickEvent) => {
+            event.preventDefault();
+            link.onClick();
+          }}
+          // @ts-expect-error MUI variant type does not include monochrome but functions appropriately when set
```
shawyu-okta, line 60: Can you add type augmentations to fix one or either of these TS errors?

shawyu-okta: Oh, I see this was the way it was 😞 Not blocking, but something to consider if you're able to

src/v3/src/components/WidgetMessageContainer/WidgetMessageContainer.tsx
```diff
@@ -46,6 +46,46 @@ const WidgetMessageContainer: FunctionComponent<{
...
+          {link.label}
+        </Link>
+      );
+    }
+
+    // Custom error remediation allows admins to define a message without a URL
+    return link.label;
+  };
+
   const renderLinks = (links?: WidgetMessageLink[], listStyleType?: ListStyleType) => (links && (
```
shawyu-okta, line 89: It seems like the type `WidgetMessageLink` and the associated property names and methods named `link` are no longer accurate. It's definitely unexpected for an array of `WidgetMessageLink` to possibly represent elements that are not links or link-like at all. Can we refactor the types to broaden the definition so it encompasses links but also non-link elements? There's already actually two distinct types of behavior that come from a single shared type (with url and `isLinkButton=true`), so now the need is greater still.

trevoring-okta: Yeah, this makes sense. Will look into that right now

src/v3/src/transformer/field/transform.ts
```diff
@@ -191,7 +191,9 @@ export const transformStepInputs = (
     // string (which is set in transformEnrollProfileUpdate), so we should not remove it from the payload here
     if (fieldName === 'userProfile.secondEmail') {
       return acc;
-    } else if (typeof value === 'string' && value.trim().length === 0 && optionalInputNames.has(fieldName)) {
+    }
+
+    if (typeof value === 'string' && value.trim().length === 0 && optionalInputNames.has(fieldName)) {
```
trevoring-okta, line 196: automatic misc. lint fix that was pulled in

src/v3/src/types/schema.ts
```diff
@@ -58,33 +58,33 @@ export type WidgetMessage = Modify<IdxMessage, {
   title?: string;
   name?: string;
   description?: string;
-  links?: WidgetMessageLink[];
+  links?: MessageLink[];
+  options?: WidgetMessageOption[];
```
trevoring-okta, line 62: @shawyu-okta 
`links` are the raw type returned by IDX (maintains parity w/ existing functionality)
`options` are the new broadened/augmented type exclusively used by the Device Assurance end user remediation logic, which supports a union of link button, link, and plain text

src/v3/src/components/WidgetMessageContainer/WidgetMessageContainer.tsx
```diff
@@ -151,7 +167,8 @@ const WidgetMessageContainer: FunctionComponent<{
       <React.Fragment>
         {renderTitle(message.title)}
         {Array.isArray(message.message) ? createListMessages(message) : parsedContent}
-        {renderLinks(message.links, message.listStyleType)}
+        {/* `message.options` is an augmented version of `message.links`, so they are mutually exclusive */}
+        {renderOptions(message.options ?? message.links?.map((link) => ({ type: 'link', ...link })), message.listStyleType)}
```
trevoring-okta, line 171: We have one type of response ([error-identify-access-denied-custom-message.json](https://github.com/okta/okta-signin-widget/blob/2340053cf18b629ce684e865caba3018d00fc0e5/playground/mocks/data/idp/idx/error-identify-access-denied-custom-message.json#L4)), which also uses the `links` field in the IDX `messages` object, but does not need additional augmentation into the `options` field.

We can view `options` and `links` as mutually exclusive and add this mapping of links to augment it with the `type` [discriminator](https://basarat.gitbook.io/typescript/type-system/discriminated-unions) to allow it to reuse the new `renderOptions()` function

src/v3/src/types/schema.ts
shawyu-okta: So in the long run, perhaps we can replace `links` with `options` in all cases within widget code by processing the incoming raw response to move any data in `links` to `options`

src/v3/src/types/schema.ts
```diff
@@ -58,33 +58,33 @@ export type WidgetMessage = Modify<IdxMessage, {
...
-    label: MessageLink['label'];
-    onClick: () => void;
-    dataSe: string;
-  }
-  : {
-    isLinkButton: TIsLinkButton;
-    label: MessageLink['label'];
-    url: MessageLink['url'];
-  };
+export type WidgetMessageLinkButton = {
```
shawyu-okta, line 75: Can we consider a non-exported base type that lists the common properties and have the others extend it?

src/v3/src/util/formUtils.ts
```diff
@@ -360,6 +369,9 @@ export const buildEndUserRemediationMessages = (
     } = msg;
 
     const widgetMsg = { listStyleType: 'disc' } as WidgetMessage;
+    const isLinklessMessage = key.startsWith(DEVICE_ASSURANCE_CUSTOM_REMEDIATION_I18N_KEY_PREFIX)
```
shawyu-okta, line 372: Is it possible to refactor this logic into the `buildEnduserRemediationWidgetMessageOption` method?

playground/mocks/data/idp/idx/end-user-remediation-custom-message-no-url.json
```diff
@@ -0,0 +1,53 @@
...
+        ],
+        "message": "Enable lock screen and biometrics",
+        "i18n": {
+          "key": "idx.error.code.access_denied.device_assurance.remediation.android.use_biometric_lock_screen"
+        },
+        "class": "ERROR"
+      },
+      {
+        "links": [
+          {"url": "https://okta.com/help"}
```
shawyu-okta, line 43: Is this mock correct? It's named `-no-url`

test/testcafe/spec/TerminalView_spec.js
```diff
@@ -302,3 +322,62 @@ test.requestHooks(endUserRemediationNoOptionsMock)('should render end user remed
...
+});
+
+test.requestHooks(endUserRemediationCustomMessageCustomUrlMock)('should render end user remediation when there is an option with a custom message and custom URL', async t => {
+  const terminalViewPage = await setup(t);
+  await checkA11y(t);
+
+  await t.expect(terminalViewPage.form.getErrorBoxCallout().withText('Your device doesn\'t meet the security requirements').exists).eql(true);
+  await t.expect(terminalViewPage.form.getErrorBoxCallout().withText('To sign in, make the following updates. Then, access the app again.').exists).eql(true);
+
+  await t.expect(terminalViewPage.form.getErrorBoxAnchor('https://myremediationurl.com/docs').withExactText('It\'s our company policy that you upgrade your Android OS').exists).eql(true);
```
shawyu-okta, line 350: Is this assertion correct? I would have thought this would be asserting `false`

test/testcafe/spec/TerminalView_spec.js
```diff
@@ -302,3 +322,62 @@ test.requestHooks(endUserRemediationNoOptionsMock)('should render end user remed
...
+  await t.expect(terminalViewPage.form.getErrorBoxAnchor('https://okta.com/android-biometric-lock').withExactText('Enable lock screen and biometrics').exists).eql(true);
+
+  await t.expect(terminalViewPage.form.getErrorBoxCallout().withText('follow the instructions on the help page').find('a[href="https://okta.com/help"]').exists).eql(true);
+  await t.expect(terminalViewPage.form.getAnchorsWithBlankTargetsWithoutRelevantAttributes().exists).eql(false);
+});
+
+test.requestHooks(endUserRemediationCustomMessageNoUrlMock)('should render end user remediation when there is an option with a custom message and no URL', async t => {
```
shawyu-okta, line 357: Consider adding an assertion that there is no link. I realize a negative assertion is tricky, so I am looking for an assertion not on a specific link href missing, but that there is no anchor element or similar.

src/v3/src/components/WidgetMessageContainer/WidgetMessageContainer.tsx
shawyu-okta: This seems workable for now but along with my other comment, we should pre-process this earlier in the data flow so by the time it reaches the render component it's already in the data shape we want (i.e. no `links` field, all of the `links` data copied into `options`)

src/v3/src/types/schema.ts
trevoring-okta: Yeah, was thinking about that. That'd be ideal

src/v3/src/util/formUtils.ts
trevoring-okta: We need to know whether the message is linkless by line 399, which is the condition that determines whether we call `buildEndUserRemediationWidgetMessageOption()` at all, so not sure if refactoring is possible here unless we want to abstract this out into its own `isLinklessMessage()` function (similar to `isLoopbackDeviceRemediation()` and then call it on line 399 and again on line 332

shawyu-okta (approved):

PR #3783, opened 2025-02-04
Fix: Gen3 - Hide back to signin link in IDV based on IDX response

praveendurai-okta:
## Description:

Hide cancel/back to signin button if `cancel` object is missing in IDX response. similar to V2. Only for IDV View.

## PR Checklist

### Issue:

- [OKTA-863251](https://oktainc.atlassian.net/browse/OKTA-863251)

### Screenshot/Video:

Screenshot 2025-02-04 at 2 16 37 PM

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&page=1&pageSize=6&sha=bd4f09c66076c2bae466e21cc1154ac970441c24&tab=main

### DIFF
--- a/src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.test.ts
+++ b/src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.test.ts
@@ -57,6 +57,4 @@ describe('IDV IDP Authenticator transformer Tests', () => {
     expect(updatedFormBag.uischema.elements.length).toBe(7);
     expect(updatedFormBag).toMatchSnapshot();
   });
-
-
 });
--- a/src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.test.ts
+++ b/src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.test.ts
@@ -23,7 +23,6 @@ import {
   LinkElement,
   TitleElement,
 } from '../../../types';
-
 import { transformIdvIdpAuthenticator } from './transformIdvIdpAuthenticator';
 
 describe('IDV IDP Authenticator transformer Tests', () => {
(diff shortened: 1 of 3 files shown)


src/v3/src/transformer/layout/idp/transformIdvIdpAuthenticator.test.ts
```diff
@@ -36,6 +36,19 @@ describe('IDV IDP Authenticator transformer Tests', () => {
   });
 
   it('should add correct title, descriptions, and button', () => {
+    const updatedFormBag = transformIdvIdpAuthenticator({
+      formBag,
+      transaction,
+      widgetProps: {},
+    });
+    expect(updatedFormBag.uischema.elements.length).toBe(6);
+    expect(updatedFormBag).toMatchSnapshot();
```
shawyu-okta, line 45: It's good to have these snapshot assertions, but snapshot assertions alone aren't sufficient to verify expected behavior since they don't give the reader any information about the expectations for the code being tested.

There used to be some more explicit assertions. I'm not sure why they were removed in [this PR](https://github.com/okta/okta-signin-widget/pull/3720/files#diff-2882224631b9c352c93f7e33ba8337a72d1648a6c90430dfc7d30b479607c9a2) previously, but I'd like to see them put back and also improved (see `transformOktaVerifyCustomAppChallengePoll.test.ts`).

As a part of this change, there should be assertions added for the cancel element being present or not. In the test mentioned, it's the assertion that looks like this https://github.com/okta/okta-signin-widget/blob/2340053cf18b629ce684e865caba3018d00fc0e5/src/v3/src/transformer/oktaVerify/transformOktaVerifyCustomAppChallengePoll.test.ts#L99-L100

shawyu-okta (changes requested):
The code change itself looks good. This is a similar to some existing code that checks for the presence of the `cancel` step in other views. The only concern is around the unit tests --both existing and new -- for this transformer so we can be sure we're verifying the things we care about there. In this case we don't need any new browser-based test if we can cover it with unit tests properly.

shawyu-okta (dismissed):

shawyu-okta (approved):

PR #3786, opened 2025-02-12
fix: password field validation issue while clicking forgot password

shuowu-okta:
## Description:

https://github.com/user-attachments/assets/b5fa4820-2e83-4325-a0cc-0351e186ff4c

## PR Checklist

### Issue:

- [OKTA-856487](https://oktainc.atlassian.net/browse/OKTA-856487)


src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -78,6 +78,9 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
         handleChange?.(e.currentTarget.value);
       }}
       onBlur={(e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+        if (e.relatedTarget?.tagName === 'BUTTON') {
+          return;
+        }
```
lesterchoi-okta: Use of `relatedTarget` feels like the right approach, but could you

1. Update the conditional to handle `tagName === "A"` as well? Maybe there are other tags to consider
2. Handle the click on primary button (`type="submit"`) and other elements that **should** trigger form validation. When the "Next" button is clicked, this logic will suppress/prevent the form validation from triggering?
3. The logic should be generalized and applied to all input fields that can "push down" the "clicked and missed" element when validation errors appear.
4. Consider reversing the logic. Maybe design has opinions on progressive validation being performed `onBlur` vs `onSubmit`

Screenshot 2025-02-12 at 12 07 02

lesterchoi-okta (changes requested):

src/v3/src/components/InputPassword/InputPassword.tsx
```diff
@@ -78,6 +78,11 @@ const InputPassword: UISchemaElementComponent<UISchemaElementComponentWithValida
         handleChange?.(e.currentTarget.value);
       }}
       onBlur={(e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+        // NOTE: only skip onBlur handler for link button to avoid breaking existing a11y functionalities
+        // @ts-expect-error - preact does not have FocusEvent
+        if (e.relatedTarget?.tagName === 'BUTTON' && e.relatedTarget?.getAttribute('role') === 'link') {
```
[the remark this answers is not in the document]
lesterchoi-okta: Which elements do we expect to have`role="link"`? It feels brittle since the `role` attribute won't be set by default. This would have to be explicitly applied to links/elements on which we want to prevent validation

src/v3/src/components/Checkbox/Checkbox.tsx
```diff
@@ -70,8 +70,8 @@ const Checkbox: UISchemaElementComponent<UISchemaElementComponentWithValidationP
         isDisabled={loading || isReadOnly}
         label={label}
         name={name}
-        onBlur={(e: ChangeEvent<HTMLInputElement>) => {
-          handleBlur?.(e?.currentTarget?.checked);
+        onBlur={(e: FocusEvent) => {
+          handleBlur?.((e?.currentTarget as HTMLInputElement)?.checked, e as FocusEvent);
```
shawyu-okta, line 74: See `src/v3/src/types/handlers.ts`, which contains our definitions of the events. You shouldn't need casts with `as`.

src/v3/src/types/component.ts
```diff
@@ -24,6 +24,8 @@ export type TesterFunction = (
   uiElement: UISchemaElement | FieldElement | ButtonElement
 ) => boolean;
 
+export type HandleBlurFunction = (value: string | number | boolean, event: FocusEvent) => void;
```
shawyu-okta, line 27: Focus change handler should be moved to the `handlers.ts` file as well. See the `ClickHandler` definition that already exists. The `ChangeHandler` is actually unused and also incorrectly typed, so we could remove it while you're in there.

src/v3/src/components/Select/Select.tsx
```diff
@@ -78,11 +79,11 @@ const Select: UISchemaElementComponent<UISchemaElementComponentWithValidationPro
         );
         handleChange?.(selectTarget.value);
       }}
-      onBlur={(e: SelectChangeEvent<string>) => {
+      onBlur={(e: FocusEvent) => {
```
shawyu-okta, line 82:
```suggestion
      onBlur={(e: FocusEvent<HTMLSelectElement>) => {
```

src/v3/src/components/Select/Select.tsx
```diff
@@ -78,11 +79,11 @@ const Select: UISchemaElementComponent<UISchemaElementComponentWithValidationPro
...
         const selectTarget = (
           e?.target as SelectChangeEvent['target'] & { value: string; name: string; }
         );
-        handleBlur?.(selectTarget.value);
+        handleBlur?.(selectTarget.value, e);
```
shawyu-okta, line 86: I think this should have been `e.currentTarget` (which is guaranteed to be on `e` instead of `e.target`, which is nullable), but perhaps @trevoring-okta can comment on if using `e.target` was intentional.

src/v3/src/types/handlers.ts
```diff
@@ -14,6 +14,7 @@ import { h } from 'preact';
 
 export type ClickHandler<T extends EventTarget = HTMLButtonElement> = h.JSX.MouseEventHandler<T>;
 export type ChangeHandler<T extends EventTarget = HTMLInputElement> = h.JSX.TargetedEvent<T>;
+export type BlurHandler = (value: string | number | boolean, event: FocusEvent) => void;
```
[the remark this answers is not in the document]
shawyu-okta:
```suggestion
export type FocusHandler = h.JSX.FocusEventHandler<T>;
```
Although `BlurHandler` is also fine for naming since we use it for that much more and it's the same exact type in reality

shawyu-okta (approved):

PR #3787, opened 2025-02-12
fix(gen3): Updates loading state heading styles

shawyu-okta:
## PR Checklist

### Issue:

- [OKTA-862404](https://oktainc.atlassian.net/browse/OKTA-862404)

### Screenshot/Video:
<details>
<summary>Success redirect view (interstitial before redirect = NONE)</summary>

success-redirect-no-interstitial

</details>

<details>
<summary>Success redirect view with app name and identifier (interstitial before redirect = DEFAULT)</summary>

success-redirect-app-and-identifier

</details>

<details>
<summary>OV Fastpass loopback probing poll view</summary>

ov-fp-loopback-poll

</details>

<details>
<summary>OV device challenge poll view for Chrome DTC)</summary>

**Other OV device challenge methods (App Link, Custom URI, Universal Link) remain the same**

ov-device-challenge-poll-chrome-dtc

</details>

<details>
<summary>Apple SSO extension view</summary>

apple-ssoe

</details>


src/v3/src/components/ActionPending/ActionPending.tsx
```diff
@@ -0,0 +1,62 @@
...
+  uischema: ActionPendingElement
+}> = (
+  { uischema: { options } },
+) => {
+  const titleRef = useRef<HTMLTitleElement>(null);
+  const { widgetProps } = useWidgetContext();
+  const { features: { autoFocus = false } = {} } = widgetProps;
+
+  useEffect(() => {
+    if (!autoFocus) {
```
lesterchoi-okta, line 35: Is this condition backwards?

src/v3/src/transformer/redirect/redirectTransformer.test.ts
```diff
@@ -84,8 +85,8 @@ describe('Success Redirect Transform Tests', () => {
     );
 
     expect(formBag.uischema.elements.length).toBe(3);
-    expect(formBag.uischema.elements[0].type).toBe('Description');
-    expect((formBag.uischema.elements[0] as DescriptionElement).options?.content)
+    expect(formBag.uischema.elements[0].type).toBe('ActionPending');
```
lesterchoi-okta, line 88: nit: I'm seeing mixed use of the custom matcher vs built-in matcher: `expect(x).toBeElementType('value')` vs `expect(x.type).toBe('Value')`, IMHO it implies the custom matcher isn't doing enough work to be worthwhile.

src/v3/src/components/ActionPending/ActionPending.tsx
```diff
@@ -0,0 +1,62 @@
...
+      display="flex"
+      flexDirection="column"
+      justifyContent="center"
+      alignItems="center"
+      textAlign="center"
+      testId={options.dataSe}
+    >
+      <Typography
+        component="h2"
+        variant="h5"
```
lesterchoi-okta, line 52: Should these be props/options with sensible default values? And is H2 chosen because it's the correct level after the H1 logo heading where this is most likely to appear?

src/v3/src/components/ActionPending/ActionPending.tsx
```diff
@@ -0,0 +1,62 @@
...
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  return (
+    <Box
+      display="flex"
+      flexDirection="column"
+      justifyContent="center"
+      alignItems="center"
+      textAlign="center"
```
lesterchoi-okta, line 47: nit: Are all of these necessary if the component only handles one `options.content`? If it's always a single string value, I think `textAlign="center"` is sufficient?

src/v3/jest.setup.js
```diff
@@ -18,10 +18,29 @@ import {
 } from 'jest-extended';
 import mockBundles from '../util/Bundles.ts';
 
+const toBeElementType = (receivedElement, expectedType) => {
```
lesterchoi-okta, line 21: Not sure if we need this custom matcher

lesterchoi-okta (commented):
AFAICT the VRT is the only place that validates that the text is centered. Consider adding a more programmatic check using something like this? https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle 
or that the appropriate `textAlign="center"` prop is passed to the Box component by mocking/spying?

src/v3/src/components/ActionPending/ActionPending.tsx
lesterchoi-okta: https://www.a11yproject.com/posts/how-to-accessible-heading-structure/

> Do not choose a heading by its size, but by its level in the context of the content.
> Do not skip a heading level from the top down.

I'm guessing H2 and H5 are chosen for the case(s) you focused on, but I think need flexibility for accessibility

src/v3/src/components/ActionPending/ActionPending.tsx
glenfannin-okta: My guess is, if they configure autoFocus to be disabled, we want this to always be focused? If so, how will that play with [this](https://github.com/okta/okta-signin-widget/blob/master/src/v3/src/components/Title/Title.tsx#L30-L35)? Maybe this gets loaded onto the page after the title so this takes precedence?

src/v3/src/components/ActionPending/ActionPending.tsx
shawyu-okta: I created a special component for this to emphasize that this type of heading and centering is used for a specific purpose so I am using the same heading level as the `Title` component, which is also `h2` (`h1` is the logo and logo alt text)

src/v3/src/transformer/redirect/redirectTransformer.test.ts
shawyu-okta: Yeah, I was going through and didn't migrate all of them, but I realized that we shouldn't using the `as` cast  (it never fails, even if the element properties are invalid for the type) but instead use a different kind of type check. I agree that the `toBeElementType` isn't providing that rigor anyways so I can remove it.

src/v3/src/components/ActionPending/ActionPending.tsx
shawyu-okta: `Title` and this component don't appear at the same time in any of the scenarios, and this text was previously treated the same way as a title heading in the Gen2 widget, based on the testcafe test objects, so I am using the same autofocus behavior given that it will face the same type of a11y issue as `Title` as the first text element we would want a screenreader to focus on in the view.

shawyu-okta: And yes, to @glenfannin-okta 's point, this is the intended logic (I copied it from `Title.tsx`) -- so if no element is autofocused with the widget config, we want to move focus to the view title so the screenreader will jump to it when a new view loads.

shawyu-okta: Out of scope for this bug fix, but we want to consider moving all of this logic into the `setFocusOnFirstElement` transformer.

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -80,12 +102,7 @@ export const transformOktaVerifyDeviceChallengePoll: IdxStepTransformer = ({
     : transaction.nextStep?.relatesTo?.value?.contextualData?.challenge?.value;
   const { challengeMethod, href, downloadHref } = deviceChallengePayload;
 
-  const titleElement: TitleElement = {
-    type: 'Title',
-    options: {
-      content: getTitleText(challengeMethod),
-    },
-  };
+  const titleElement = getTitleElement(challengeMethod);
```
alexchen-okta, line 105: is there a reason why you changed the pattern here to use getTitleElement instead of getTitleText? the old pattern seems more consistent with the description element.

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
```diff
@@ -30,23 +31,44 @@ import {
...
-      return loc('appLink.title', 'login');
-
+      return {
+        type: 'Title',
+        options: {
+          content: loc('appLink.title', 'login'),
+        },
+      };
     case CHALLENGE_METHOD.CHROME_DTC:
       // reusing the existing message for Chrome DTC
```
alexchen-okta, line 44: Is this the only scope for this PR? Are there other views that the ActionPendingElement can be applied?

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
shawyu-okta: Yeah, it's because we return two different possible elements for the title element: `Title` and `ActionPending`

src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyDeviceChallengePoll.ts
shawyu-okta: I only updated this challenge method in this view because it presents to the user exactly the same as the OVFP probing view. The other device challenge screens have other text elements and some have a button, so we're leaving them as-is. We can consider them for future changes if warranted.

lesterchoi-okta (approved):

alexchen-okta (approved):

PR #3798, opened 2025-02-27
fix(gen2): transformUsername should only be used for background data request

shuowu-okta:
## Description:

transformUsername should only be applied to network request payload 

## PR Checklist

### Issue:

- [OKTA-858517](https://oktainc.atlassian.net/browse/OKTA-858517)

### Screenshot/Video:

before fix:

https://github.com/user-attachments/assets/8195d73f-20b3-4a5e-b983-ce1697f61735

after fix:

https://github.com/user-attachments/assets/aef71bea-2289-4554-a7d0-eb13e2cf0ed4

### DIFF
--- a/src/v2/controllers/FormController.ts
+++ b/src/v2/controllers/FormController.ts
@@ -273,8 +273,10 @@ export default Controller.extend({
     }
 
     const payload = transformPayload(formName, model);
+    // NOTE: this line should be called before triggering transformIdentifier
+    const originalIdentifier = payload.identifier;
     // Run hook: transform the user name (a.k.a identifier)
-    const [values, originalIdentifier] = this.transformIdentifier(formName, payload);
+    const values = this.transformIdentifier(formName, payload);
 
     // widget rememberMe feature stores the entered identifier in a cookie, to pre-fill the form on subsequent visits to page
     if (this.options.settings.get('features.rememberMe')) {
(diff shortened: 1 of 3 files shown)


test/testcafe/spec/IdentifyWithRememberUsername_spec.js
```diff
@@ -277,5 +277,51 @@ test.requestHooks(identifyRequestLogger, identifyMock)('should store transformed
...
+  });
```
lesterchoi-okta, line 317: IMO this is the only part under test?

```suggestion
  await t.expect(reqBody.identifier).eql('test.identifier@acme.com');
```

src/v2/controllers/FormController.ts
```diff
@@ -356,6 +356,7 @@ export default Controller.extend({
   },
 
   transformIdentifier(formName, modelJSON) {
+    const originalIdentifier = modelJSON.identifier;
```
shawyu-okta, line 359: Couldn't we do this in the code above instead?

Side note: this function is confusing because it returns an object but it actually modifies the passed object as well. It's worth update it to either return `void` to make it clear that the passed object is being mutated or it should return a single `string` with the value of transformed identifier.

shawyu-okta (approved):
Target branch `7.29`

jaredperreault-okta (approved):

PR #3807, opened 2025-03-13
Gen3: Auto check "Keep me signed in" checkbox based on `rememberMe` value in IDX payload

denysoblohin-okta:
## Description:

If FF `ENG_SET_REMEMBER_ME_BY_DEFAULT` is on, server sends `value: true` for `rememberMe` field in IDX payload.
Gen2 already handles this situation (auto check checkbox "Keep me signed in").
This PR fixes behavior for Gen3

## PR Checklist

### Issue:

- [OKTA-859336](https://oktainc.atlassian.net/browse/OKTA-859336)


jaredperreault-okta:
If gen2 already supports this feature, why does this require a new mock?
If the new mock is definitely required, we probably don't have test coverage for gen2, can you add it?

denysoblohin-okta:
> If gen2 already supports this feature, why does this require a new mock?

It's not a feature specific to FF "ENG_SET_REMEMBER_ME_BY_DEFAULT". 
It's not specific to Identify page either. 
There is a different logic to process `value` of fields from IDX payload.

Eg. if there were default values in [enroll profile payload](https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/data/idp/idx/enroll-profile.json) like
```
...
        "value": [
          {
            "name": "userProfile",
            "form": {
              "value": [
                {
                  "name": "lastName",
                  "label": "Last name",
                  "required": true,
                  "value": "Perreault"
                },
                {
                  "name": "firstName",
                  "label": "First name",
                  "required": true,
                  "value": "Jared"
                },
...
```
Gen2 would render "Jared" and "Perreault" in text fields, Gen3 would not

denysoblohin-okta:
I've added a testcafe test that checks both Gen2 and Gen3

src/v3/src/hooks/useValue.ts
```diff
@@ -14,9 +14,9 @@ import { useWidgetContext } from '../contexts';
 import { FieldElement } from '../types';
 
 export const useValue = (uischema: FieldElement): unknown => {
-  const { name } = uischema.options.inputMeta;
+  const { name, value } = uischema.options.inputMeta;
 
   const { data } = useWidgetContext();
 
-  return data[name];
+  return data[name] ?? value;
```
shawyu-okta, line 21: Should we limit this to only the identifier field or is the behavior in Gen2 to take values for any field from the server already as well?

denysoblohin-okta: > is the behavior in Gen2 to take values for any field from the server already as well?

Yes, this behavior is implemented in this line: 
https://github.com/okta/okta-signin-widget/blob/2d7c17ab7044cb7bf94d5d0176adb80f3c5ba9c5/src/v2/view-builder/internals/BaseModel.js#L18
It picks value from IDX response to model. If I remove `'value'` from the list, Gen2 will not take values from the server by default

shawyu-okta (approved):
Get approval from @jaredperreault-okta and target `master` for the monthly release

jaredperreault-okta (dismissed):

jaredperreault-okta (approved):

PR #3823, opened 2025-04-21
fix(gen3): Prevent translation plugins from modifying DOM

shawyu-okta:
## Description:

Google Translate (and other translation plugins) manipulate the DOM in a way that Preact/React cannot handle due to the assumption that React/Preact make about the ownership of the nodes within it's container.

Setting `translate="no"` prevents translation plugins from affecting the container node and child nodes. Preventing translation of the widget isn't limiting in any way from the current behavior since the widget does not work at all if a translation plugin is applied to it.

There _may_ be workarounds but they require us to write our JSX in a particular way to avoid the issue. There is a [lint plugin](https://github.com/getcouped/eslint-plugin-react-google-translate) that might also be able to help. Any pursuit of a workaround would be a greater scope of work so we are not currently pursuing that.

## PR Checklist

### Issue:

- [OKTA-901691](https://oktainc.atlassian.net/browse/OKTA-901691)

### DIFF
--- a/src/v3/src/OktaSignIn/index.ts
+++ b/src/v3/src/OktaSignIn/index.ts
@@ -171,6 +171,13 @@ export default class OktaSignIn implements OktaSignInAPI {
           : this.el;
 
         if (target) {
+          // OKTA-901691 Translation plugins break React/Preact since they directly maniuplate the DOM
+          // and React/Preact does not know about it so the next conditional render will fail.
+          // Setting this attribute will prevent translation plugins from affecting the Gen3 widget.
+          // This is a workaround until we can find a better solution.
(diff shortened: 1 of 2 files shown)


alexchen-okta (approved):

src/v3/src/OktaSignIn/index.ts
```diff
@@ -171,6 +171,13 @@ export default class OktaSignIn implements OktaSignInAPI {
           : this.el;
 
         if (target) {
+          // OKTA-901691 Translation plugins break React/Preact since they directly maniuplate the DOM
```
shawyu-okta, line 174:
```suggestion
          // OKTA-901691 Translation plugins break React/Preact since they directly manipulate the DOM
```

PR #3828, opened 2025-04-29
fix(gen3): Clear transaction and state handle in terminal view

shawyu-okta:
## PR Checklist

### Issue:

- [OKTA-826892](https://oktainc.atlassian.net/browse/OKTA-826892)

### Screenshot/Video:

https://github.com/user-attachments/assets/5ad329f0-cb7f-41c2-8d67-fb131a0c4be0


jaredperreault-okta (approved):

src/v3/src/constants/idxConstants.ts
```diff
@@ -139,6 +139,7 @@ export const TERMINAL_KEY: Record<string, string> = {
   RETURN_TO_ORIGINAL_TAB_KEY: 'idx.return.to.original.tab',
   SAFE_MODE_KEY_PREFIX: 'idx.error.server.safe.mode',
   SESSION_EXPIRED: 'idx.session.expired',
+  VERIFICATION_TIMED_OUT: 'core.system.transaction_resource_not_found',
```
trevoring-okta, line 142: [suggestion] When I made my original fix that we backed out, I noticed that we are missing localization for this specific IDX error. Can you please add it to the bundle (see my original PR [here](https://github.com/okta/okta-signin-widget/pull/3765/files)). Optionally/nice to have would also be to add a mock for this terminal error and add it as a test case in `TerminalView_spec.js` as I did in my PR

src/v3/src/transformer/terminal/transformTerminalTransaction.test.ts
```diff
@@ -236,21 +245,109 @@ describe('Terminal Transaction Transformer Tests', () => {
...
+    } as WidgetProps;
     const mockErrorMessage = 'Session expired';
     transaction.messages?.push(getMockMessage(
       mockErrorMessage,
       'ERROR',
       TERMINAL_KEY.SESSION_EXPIRED,
     ));
     const formBag = transformTerminalTransaction(transaction, widgetProps, mockBootstrapFn);
+    expect(SessionStorage.removeStateHandle).toHaveBeenCalledTimes(1);
+    expect(mockAuthClient.transactionManager.clear).toHaveBeenCalledTimes(1);
```
trevoring-okta, line 268: [question/clarification] I was originally confused why session storage state handle and transaction manager clear were being called here, but took a closer look and seems like this is a test case that covers a code path unrelated to your change in this PR and this behavior already existed specifically for `TERMINAL_KEY.SESSION_EXPIRED`?

src/v3/src/transformer/terminal/transformTerminalTransaction.ts
```diff
@@ -116,7 +116,23 @@ const appendViewLinks = (
       && isCancelAvailable) {
     const backToSigninUri = getBackToSignInUri(widgetProps);
     if (backToSigninUri) {
+      // If a backtoSigninUri is set, always use it
       cancelLink.options.href = backToSigninUri;
+    } else if (containsMessageKey(TERMINAL_KEY.VERIFICATION_TIMED_OUT, transaction.messages)) {
```
trevoring-okta, line 121: [question] You mentioned on Slack that:

> I actually think the behavior for this scenario probably applies to more views (or even should be the default behavior) but scoped it to be specific so we limit the number of views affected

Do we have any way of determining what other views might benefit from this "hard reset"? I agree that it is very likely that there are more cases, but it seems like we may have to rely on additional customer-reported issues to determine those scenarios 🤔

src/v3/src/transformer/terminal/transformTerminalTransaction.ts
```diff
@@ -116,7 +116,23 @@ const appendViewLinks = (
...
+      // This covers terminal views that should have a page reload to start a new transaction in the case
+      // that the URI is specifing a particular application to login to, so we need to stay on the same URI
```
trevoring-okta, line 123: [super nit (only if you're planning on addressing comments)] 
```suggestion
      // that the URI is specifying a particular application to login to, so we need to stay on the same URI
```

trevoring-okta (commented):
Just a few questions and a suggestion for pulling in some changes related to the `Verification timed out` error that I made in my original PR for this fix

src/v3/src/transformer/terminal/transformTerminalTransaction.ts
shawyu-okta: We will need to audit and think more carefully about what other terminal views this would apply to. For now we only will limit the scope to resolve this issue but i've specifically structured the code and left the comment so we know to use this branch in the future.

src/v3/src/transformer/terminal/transformTerminalTransaction.test.ts
shawyu-okta: Yeah, so in `SESSION_EXPIRED` the transformer immediately clears session storage and transaction manager as opposed to doing it only on click. I'm not sure if we want to use that as the generalized behavior or not, though. I would be willing to make a case for clearing the storage/manager as soon as the view is hit. The main side-effect of this would be that a refresh of the page would not keep the terminal state, but that might not be a bad thing, and it would allow us to greatly simplify the logic for clearing state when a terminal view is reached.

src/v3/src/constants/idxConstants.ts
shawyu-okta: Copied over your mock/test and added translation key

src/v3/src/transformer/terminal/transformTerminalTransaction.test.ts
trevoring-okta: Hm, yeah I think it could make sense to do that once we hit the code path that you're changing (within the `else if (!transaction.actions?.cancel
      && !containsOneOfMessageKeys(TERMINAL_KEYS_WITHOUT_CANCEL, transaction.messages)
      && isCancelAvailable` case), but let's reduce the potential blast radius and take that as a follow up alongside any potential audit we do.

trevoring-okta (approved):
Thanks for addressing questions + suggestions!

shuowu-okta:
@shawyu-okta This change adds a new i18n string, let's wait for monthly for translation team pick up.

lesterchoi-okta (approved):

shuowu-okta (approved):

PR #2794, opened 2022-09-19
OKTA-521313 : Trim whitespace from numeric type fields

glenfannin-okta:
## Description:

The purpose of this PR is to trim any whitespace from numeric fields (specifically, OTP code fields).

## PR Checklist

- [x] Have you verified the basic functionality for this change?
- [x] Did you add tests, as appropriate, following our [Automated Test guidelines](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2676497890/Automated+Testing+in+the+Signin+Widget)?
- [x] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?
- [ ] Did you verify the change by running [downstream monolith artifact](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/102897979/Sign-in+Widget+Development#Sign-inWidgetDevelopment-Instructionstocreateandrunthedownstreamartifact(d16t))? (Provide link to build below)
- [ ] Does this PR include noticeable changes to the UI? (If yes, attach screenshots/video below)

### Issue:

- [OKTA-521313](https://oktainc.atlassian.net/browse/OKTA-521313)

PR #3845, opened 2025-06-17
add identifier to fieldsToTrim

Luxu2-Okta:
Resolves: OKTA-948846

## Description:
Unlike gen2, gen3 username doesn't trim, so add the `identifier` to `fieldsToTrim`
passcode input is NOT trimmed

One nuance is
In Gen2, the trim happens in the courage model level, the UI still preserve the space
Screenshot 2025-06-17 at 1 47 29 PM

In Gen3, after we add the `identifier` to `fieldsToTrim`, it will also reflect in the UI, video below:

https://github.com/user-attachments/assets/8a4cd496-57a1-4f0d-a7f9-3d3e4f0045f4

## PR Checklist

### Issue:

- [OKTA-948846](https://oktainc.atlassian.net/browse/OKTA-948846)

### DIFF
--- a/src/v3/src/transformer/field/transform.ts
+++ b/src/v3/src/transformer/field/transform.ts
@@ -170,7 +170,8 @@ export const transformStepInputs = (
         };
         acc.dataSchema.fieldsToValidate.push(name);
         // Of the required fields, trim appropriately
-        if (uischema.options.attributes?.inputmode === 'numeric') {
+        if (uischema.options.attributes?.inputmode === 'numeric'
+          || uischema.key === 'identifier' ) {
           acc.dataSchema.fieldsToTrim.push(name);
         }
       }


src/v3/src/transformer/field/transform.ts
```diff
@@ -170,7 +170,8 @@ export const transformStepInputs = (
         };
         acc.dataSchema.fieldsToValidate.push(name);
         // Of the required fields, trim appropriately
-        if (uischema.options.attributes?.inputmode === 'numeric') {
+        if (uischema.options.attributes?.inputmode === 'numeric'
+          || uischema.key === 'identifier' ) {
```
Luxu2-Okta: Another place to trim can be in this [useOnSubmit](https://github.com/okta/okta-signin-widget/blob/20beb331d8e235d2ef8e5aa401e8b9398cb210ef/src/v3/src/hooks/useOnSubmit.ts#L144) before `transformIdentifier`, but I feel `transform.ts` is better
The other question is should I check more uischema properties, like `uischema.options.inputMeta`
Screenshot 2025-06-17 at 4 48 43 PM

Luxu2-Okta: synced with @shuowu , we can add `identifier` in the [dataSchema.fieldsToTrim](https://github.com/okta/okta-signin-widget/blob/4107b7a0d1b52182f0f43a90b7f1122cb8390e32/src/v3/src/transformer/main.ts#L66) array, but check it in the field `transform.ts` it a bit safer since we can also check the `step.name`

src/v3/src/transformer/field/transform.ts
```diff
@@ -170,7 +170,10 @@ export const transformStepInputs = (
         };
         acc.dataSchema.fieldsToValidate.push(name);
         // Of the required fields, trim appropriately
-        if (uischema.options.attributes?.inputmode === 'numeric') {
+        // also trim the identifier input
+        if (uischema.options.attributes?.inputmode === 'numeric'
+          || (step.name === 'identify' && name === 'identifier')
+        ) {
```
Luxu2-Okta: synced with @shuowu, we can add `identifier` in the [dataSchema.fieldsToTrim](https://github.com/okta/okta-signin-widget/blob/4107b7a0d1b52182f0f43a90b7f1122cb8390e32/src/v3/src/transformer/main.ts#L66) array, but checking in the field `transform.ts` is a bit safer since we can also check the `step.name`

And there is another edge case that may be a concern, we trim the identifier first, then call [transformIdentifier](https://github.com/okta/okta-signin-widget/blob/4107b7a0d1b52182f0f43a90b7f1122cb8390e32/src/v3/src/hooks/useOnSubmit.ts#L145), theoretically, a user can set up a custom `transformIdentifier` like `(username) => username.replace(' ', '_')`

* Before this change, when user name input value is `Test User `, it would be `Test_User_`
* After this change, it would be `Test_User`
 
but I think it is edge case, since no user would want to treat the space as some real input and final output would be based on leading or trailing spaces

shuowu-okta (approved):

alexchen-okta (approved):

test/testcafe/spec/Identify_spec.js
```diff
@@ -332,6 +332,44 @@ test.requestHooks(identifyThenSelectAuthenticatorMock)('navigate to other screen
...
...zMiLCJlbmMiOiJBMjU2R0NNIiwiYWxnIjoiZGlyIn0..4z7WWUdd0LzIJLxz.GmOyeMJ5XtS7kZipFanQxaMd2rblpEeaWv8U5IfaJv5F2V1otwft4q1tVo3yqbedhBjN-nO5qk6qR-0Op34lmecwwmHeRyzYbhrFiLZaR88nhCFMblP8Bo5d3Opl5gkX02e0FQL-Osorxvml0XYbuO7GdTH5EkIv0Q7h0Dq7L__h_uFLies8AkJZWIDQh25RlqcEjToHCvVW31A_NJJ1Vf5c8GFuyb9LsJ9kUpZikpK6C72GPrU_LGrIW8VBT4l24dDEre4J8XvTNO7fdVDiq-H7BEeIjaY06q1zqlVLWwqOHoKpGNQ0NMhBIXB0ZZC57Me9pFI5GLIUwRDUIm1vw3t_mHJDVIcCJe9kmt29tTccZ8Zo0N3q5bSiwMoNHPLxZSOrbx-bf4fniorNH5ypJnke7pc2Q3DFmqfPrB7CE1REjAyCKBHYDAfVexYCkMfCl0E8oMFJinnLbynb7Bqvbp_DqL8h0pNIoXUF4KTTsuKQg8yCCqBhkajxlvh9G7L3Sf76o4B2itB7ldeqXzAE9H60yqhIKEZPNOHUgRC2SkWkWlH6NIaNWQ2Bi2CjnL9YvUuQmO-dpf08KeCgwfVmT4GBTGfTkXwy3pBitacCqEREen2j2iUH9mhi2LOOFaGLh0TXslcBgkGuht6P7gyH2JN6yFInQyIp33xQsqYg7nqOZG1LCrQSqoviTfI72-AC6b7tju8YEn1P0nXGbSzlCztSXl2pa95tr4L5pyX8fNydKYMTLeHEnmNtXlRB6wQYP1ljf4Tzgus7O0etyJs75znsXHZ42znxlEKGhTo3ucFe3CI-vsHF1FDDj2DVeWl21zVOTehTbBaemoD1ekD5F8OHS7SrK9Bw7PTa-lpyls1OxvE_Wsco4_eGbax_DoPm6DbCwj8hWzb5wLEs6TClZKoUJeV1MSVB3OgGBZ3AGzhwfeG0sGi5DnUpAeKqgP6IN8kziNRDmW3YE0qIY2mLs7nI438RTu__6bg1E6SH1QHMNucNbmoDR6VDIUmlYc0xEpygH6PBVqiPD64MnD73_D9IinVNzqW7KQzAvuFFQW_LGDfjuh1D-oTs1gi1wWDylibjxdJabveoJ10NHgeb6SaYHg.kf5iTnjNKsKqhz0iE5K_Yw',
```
shawyu-okta, line 347: Any objection to extracting this out to a constant at the top of the test file instead? I'm not sure we can change the value in the mock file but we can DRY this a little bit so we don't have to have this long state handle string in every test

src/v3/src/transformer/field/transform.ts
shawyu-okta: If we aren't using `fieldsToTrim` for this, what is the purpose of that `dataSchema` property?

Luxu2-Okta: I think it is for more checks, currently the [dataSchema.fieldsToTrim](https://github.com/okta/okta-signin-widget/blob/4107b7a0d1b52182f0f43a90b7f1122cb8390e32/src/v3/src/transformer/main.ts#L66) can only find the fields by the key, the functionality is added in [this change](https://github.com/okta/okta-signin-widget/pull/2794/files#diff-5909dabad92ba072e785e92903b1509f10099132412da3cd25f3c2537c38fc41R105), but when we are actually deciding fields to trim, it is more complex, so we can not directly add `credentials.passcode` in [dataSchema.fieldsToTrim](https://github.com/okta/okta-signin-widget/blob/4107b7a0d1b52182f0f43a90b7f1122cb8390e32/src/v3/src/transformer/main.ts#L66) during initialization, but need to do some extra checking [here](https://github.com/okta/okta-signin-widget/blob/40c5762ad73459101715726aaf88a48679cbf045/src/v3/src/transformer/field/attributes.ts#L58), then push the `credentials.passcode` in `dataSchema.fieldsToTrim`

I added `'identifier'` directly in `dataSchema.fieldsToTrim`, then we will trim all the `identifier` in all cases.
Searching through the playground API mocks, there are four types of `identifier`
* [identify](https://github.com/okta/okta-signin-widget/blob/40c5762ad73459101715726aaf88a48679cbf045/playground/mocks/data/idp/idx/identify.json#L14-L22), this is exactly the customer case
* [select-authenticator-unlock-account](https://github.com/okta/okta-signin-widget/blob/40c5762ad73459101715726aaf88a48679cbf045/playground/mocks/data/idp/idx/user-account-unlock-choose-auth-after-unlock.json#L13-L21)
* [select-identify](https://github.com/okta/okta-signin-widget/blob/40c5762ad73459101715726aaf88a48679cbf045/playground/mocks/data/idp/idx/enroll-profile.json#L57-L65)
* [unlock-account](https://github.com/okta/okta-signin-widget/blob/40c5762ad73459101715726aaf88a48679cbf045/playground/mocks/data/idp/idx/user-unlock-account-identifier-first.json#L13-L21)

Here are videos that whitespaces are trimmed in all these cases in playground, but since they are all identifier, so I think we can trim for all these cases

#### identify

https://github.com/user-attachments/assets/6a3c0a44-73dc-406e-b31e-ee8d73c1d881

#### select-authenticator-unlock-account

https://github.com/user-attachments/assets/2a22fc8c-bf1e-498a-b35c-fc045aaf08a8

#### select-identify

https://github.com/user-attachments/assets/7f88fb87-5040-4151-a262-954bef9162a5

#### unlock-account

https://github.com/user-attachments/assets/73b45eb6-f26c-4cfd-9745-4ecbc3c97ec9

PR #3850, opened 2025-07-01
fix(oie): Consider browser region preference for date formats

shawyu-okta:
## PR Checklist

### Issue:

- [OKTA-939377](https://oktainc.atlassian.net/browse/OKTA-939377)

### DIFF
--- a/src/util/LanguageUtil.ts
+++ b/src/util/LanguageUtil.ts
@@ -87,7 +87,7 @@ export const getLanguageTags = (language: string | LanguageCallback | undefined,
     // Get the simple language code since that's how the languagesMap is keyed
     const [lang] = languageFromProps.split('-');
     // Pull the language tags for the specified language from the map first
-    languageTags.push(...languagesMap.get(lang) ?? [])
+    languageTags.push(...(languagesMap.get(lang) ?? []))
     // Then remove the language from the map so it doesn't get added again
     languagesMap.delete(lang);
   }


src/util/LanguageUtil.js
shawyu-okta, line 1: Moved to be a TS file

.testcaferc.js
```diff
@@ -180,6 +180,9 @@ const config = {
     return true;
   },
   screenshots: {
+    path: 'build2/reports/junit',
+    takeOnFails: true,
+    fullPage: true,
```
shawyu-okta, line 185: Gen3 testcafe config is set to take screenshots on failures already. This will help with debugging failed tests in CI.

playground/main.ts
```diff
@@ -33,7 +33,7 @@ declare global {
     // added in this file
     getWidgetInstance: () => OktaSignInAPI;
     createWidgetInstance: (options: WidgetOptions) => OktaSignInAPI;
-    renderPlaygroundWidget: (options: WidgetOptions) => void;
+    renderPlaygroundWidget: (options: WidgetOptions & { assertNoEnglishLeaks?: boolean, customize?: boolean }) => Promise<void>;
```
shawyu-okta: Returning the `Promise` is an addition, the union type in the parameter is just fixing the previous type to match reality.

src/util/TimeUtil.js
```diff
@@ -101,32 +101,39 @@ export default {
...
+        hour: '2-digit',
+        minute: '2-digit',
+        timeZoneName: 'short',
+      };
+      if (languageTags === undefined) {
+        return date.toLocaleString(undefined, formatOptions);
+      }
+      const languageTagsPrefs = Array.isArray(languageTags) ? languageTags : [languageTags];
+
+      return date.toLocaleString(languageTagsPrefs, formatOptions);
```
shawyu-okta: Might be able to consolidate this since `Date.prototype.toLocaleString` can accept `string | string[]` as the first parameter, so if the normalization of `languageTags` to an array isn't needed then this can be simplified to:
```suggestion
      return date.toLocaleString(languageTags, formatOptions);
```

test/unit/spec/v1/LoginRouter_spec.js
```diff
@@ -2044,45 +2044,46 @@ Expect.describe('v1/LoginRouter', function() {
...
+              'ru',
+              'sv',
+              'th',
+              'tr',
+              'uk',
+              'vi',
+              'zh-CN',
+              'zh-TW',
+            ]);
+          });
```
shawyu-okta, line 2086: Now that there are two calls (one in `getLanguageCode` and one in `getLanguageTags`), loop over both with the same expectations

playground/main.ts
```diff
@@ -167,6 +175,8 @@ const renderPlaygroundWidget = (options: WidgetOptions & { assertNoEnglishLeaks?
   if (customize) {
     addAfterTransformHooks(signIn as OktaSignInAPIV3);
   }
+  // promise resolves when widget ready event is received
+  return readyPromise;
```
shawyu-okta: Goal here is to only proceed after widget is ready to avoid trying to run assertions on the widget before it completely finishes bootstrapping.

test/testcafe/spec/AuthenticatorIdPView_spec.js
```diff
@@ -75,13 +75,9 @@ const verifyErrorCustomLogoMock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/introspect')
   .respond(xhrVerifyIdpAuthentiatorErrorCustomLogo);
 
-async function setup(t, {isVerify, expectAutoRedirect} = {}, widgetOptions = undefined) {
-  const options = widgetOptions ? { render: false } : {};
+async function setup(t, {isVerify, expectAutoRedirect} = {}) {
   const pageObject = new IdPAuthenticatorPageObject(t);
-  await pageObject.navigateToPage(options);
-  if (widgetOptions) {
-    await renderWidget(widgetOptions);
```
shawyu-okta, line 83: There is sometimes flakiness associated with waiting for the widget re-render, so instead we use `clientScripts` to override the widget options in the template before the widget bootstrap

src/util/LanguageUtil.ts
```diff
@@ -0,0 +1,125 @@
...
+  return Bundles.loadLanguage(languageCode, i18n, {
+    baseUrl: assetBaseUrl,
+    rewrite: assetRewrite,
+  }, supportedLanguages).then(function() {
+    clearTimeout(timeout);
+    appState.trigger('loading', false);
+    appState.trigger('removeLoading');
+  });
+  // TODO: what if load language error?
+}
```
shawyu-okta, line 125: This is existing code, just showing as added because I renamed `LanguageUtil.js` to `LanguageUtil.ts`

test/testcafe/spec/AuthenticatorYubikeyView_spec.js
```diff
@@ -29,13 +29,9 @@ const verifyMock = RequestMock()
   .onRequestTo('http://localhost:3000/idp/idx/challenge/answer')
   .respond(xhrSuccess);
 
-async function setup(t, widgetOptions) {
-  const options = widgetOptions ? { render: false } : {};
+async function setup(t) {
   const pageObject = new YubiKeyAuthenticatorPageObject(t);
-  await pageObject.navigateToPage(options);
-  if (widgetOptions) {
-    await renderWidget(widgetOptions);
```
shawyu-okta, line 37: Old usage of `renderWidget` to update the widget options was unreliable and caused a lot of flakiness and failed tests when I was making the changes in this PR. I updated things to override the options _before_ the widget loads/bootstraps for the first time instead by using the `clientScripts` option in some tests.

src/util/LanguageUtil.ts
```diff
@@ -0,0 +1,125 @@
...
+    languagesMap.delete(lang);
+  }
+
+  // Add any remaining language tags from the map to the returned array
+  languagesMap.forEach((values: Set<string>) => {
+    languageTags.push(...values);
+  });
+
+  return languageTags;
+};
```
Luxu2-Okta, line 101: For my curiosity, we use `Map` in this file, and v1 also uses `loadLanguage` method, do we need to worry about backward compatibility issues like very old IE?

src/util/LanguageUtil.ts
```diff
@@ -0,0 +1,125 @@
...
+  Util.expandLanguages([config.defaultLanguage.toLowerCase()]).forEach(addLanguageTag);
+
+  const languageTags: string[] = [];
+
+  // Put languages related to the language from widget properties first if it was specified
+  if (languageFromProps) {
+    // Get the simple language code since that's how the languagesMap is keyed
+    const [lang] = languageFromProps.split('-');
+    // Pull the language tags for the specified language from the map first
+    languageTags.push(...languagesMap.get(lang) ?? [])
```
Luxu2-Okta: should we add parenthesis to make sure the precedence? `languageTags.push(...(languagesMap.get(lang) ?? []))`

Luxu2-Okta (commented):
has two comments, but LGTM in general

src/v3/.eslintrc.js
```diff
@@ -181,7 +181,12 @@ module.exports = {
         'import/named': 'error',
         'import/prefer-default-export': 'off',
         'import/no-default-export': 'warn',
-        'import/no-unresolved': 'error',
+        'import/no-unresolved': ['error', {
+          ignore: [
+            'util\\/LanguageUtil',
+            'util\\/TimeUtil',
+          ],
```
Luxu2-Okta, line 188: any reason we have to ignore this rule in these two files? Or can we fix it?

src/util/LanguageUtil.ts
shawyu-okta: `babel` _should_ handle this by adding the polyfill but it's something I can check to make sure that the widget still loads in IE11

src/util/LanguageUtil.ts
shawyu-okta: I agree that the precedence is not clear enough to the reader, will update

src/v3/.eslintrc.js
shawyu-okta: The `ignore` array is the array of paths that eslint will not check for module resolution and not an array of paths that this rule is ignored for. I had trouble getting eslint to understand my alias settings to allow this to resolve correctly, probably due to the monorepo structure of this project. The webpack and jest settings files are able to handle these resolutions correctly, so this is only so eslint doesn't throw errors and so we can avoid adding `eslint-ignore-next-line` comments in every file that imports either one of these modules.

src/util/LanguageUtil.ts
```diff
@@ -0,0 +1,125 @@
...
+      values.add(languageTag)
+      languagesMap.set(lang, values);
+    } else {
+      languagesMap.set(lang, new Set<string>([languageTag]));
+    }
+  };
+
+  const supportedUserLanguages = userLanguages.filter((userLanguage: string) => {
+    // remove any user languages tags that do not match a supported language
+    return supportedLanguages.some((supportedLang: string) => userLanguage.startsWith(supportedLang.toLowerCase()));
```
leemchale-okta, line 56: does `userLanguage` also need to be lower cased before comparing?

shawyu-okta: `userLanguages` is always lowercase -- we do `.toLowerCase()` in the code above from L24-L33

leemchale-okta: ah got it thank you!

src/util/LanguageUtil.ts
```diff
@@ -0,0 +1,125 @@
...
+ * @see https://en.wikipedia.org/wiki/IETF_language_tag
+ * 
+ * @param {string | LanguageCallback | undefined} language 
+ * @param {string[]} supportedLanguages
+ * @returns {string[]}
+ */
+export const getLanguageTags = (language: string | LanguageCallback | undefined, supportedLanguages: string[]): string[] => {
+  // Get the user's preferred languages from the browser API
+  const userLanguages = BrowserFeatures.getUserLanguages().map((userLanguage: string) => {
+    // Map "simple" language codes to their full locale equivalents to match the
```
alexchen-okta, line 25: just confirming: what if the language code does not "match" browser location preferences? is there a fallback?

shawyu-okta: @alexchen-okta I think [this test](https://github.com/okta/okta-signin-widget/pull/3850/files/623b011f3912c6ee3aaa752291bfed935d271d5f#diff-9db856217b662c6a6c3f8601eca18da9dc44568549739d7556fb7e2c742da574R60) covers the behavior you are wondering about.

alexchen-okta (approved):

PR #3855, opened 2025-07-17
Lm set title h1 fix okta 974301

leemchale-okta:
## PR Checklist

### Issue:

### DIFF
--- a/src/v3/src/components/ActionPending/ActionPending.tsx
+++ b/src/v3/src/components/ActionPending/ActionPending.tsx
@@ -55,7 +55,7 @@ const ActionPending: UISchemaElementComponent<{
         }}
       >
         <Typography
-          component="h2"
+          component="h1"
           variant="h5"
           testId="o-form-head"
           typographyRef={titleRef}
--- a/test/testcafe/framework/page-objects/BasePageObject.js
+++ b/test/testcafe/framework/page-objects/BasePageObject.js
@@ -210,6 +210,10 @@ export default class BasePageObject {
     return this.form.getErrorBoxText();
   }
 
+  getErrorTitle() {
+    return this.form.getErrorTitle();
+  }
+
   getReturnToAuthenticatorListLink() {
     return this.form.getLink('Return to authenticator list');
   }
--- a/test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
+++ b/test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
@@ -63,13 +63,6 @@ export default class ChallengeOktaVerifyPushPageObject extends ChallengeFactorPa
     return this.form.getErrorBox();
   }
 
-  getErrorTitle() {
-    if (userVariables.gen3) {
-      return within(this.getWarningBox()).findByRole('heading', { level: 2 }).innerText;
-    }
-    return this.form.getElement(FORM_INFOBOX_ERROR_TITLE).innerText;
-  }
-
   getWarningBox() {
     if (userVariables.gen3) {
       return this.form.getAlertBox();
--- a/test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
+++ b/test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
@@ -65,14 +65,9 @@ export default class ChallengeOktaVerifyPushPageObject extends ChallengeFactorPa
 
   getErrorTitle() {
     if (userVariables.gen3) {
-      return this.form.getNthTitle(0);
+      return within(this.getWarningBox()).findByRole('heading', { level: 2 }).innerText;
     }
     return this.form.getElement(FORM_INFOBOX_ERROR_TITLE).innerText;
-  }
-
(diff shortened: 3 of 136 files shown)


test/testcafe/spec/ChallengeOktaVerifyResendPushView_spec.js
```diff
@@ -106,7 +106,7 @@ test
     const challengeOktaVerifyPushPageObject = await setup(t);
     await checkA11y(t);
     await challengeOktaVerifyPushPageObject.waitForErrorBox();
-    const pageTitle = challengeOktaVerifyPushPageObject.getFormTitleWithError();
+    const pageTitle = challengeOktaVerifyPushPageObject.getFormTitle();
```
leemchale-okta, line 109: don't need to use `getFormTitleWithError()` since the form title does not share an h2 tag with the error title anymore

test/testcafe/spec/ChallengeOktaVerifyResendPushView_spec.js
```diff
@@ -91,7 +91,7 @@ test
...
```
leemchale-okta, line 94: don't need to use `getFormTitleWithError()` since the form title does not share an h2 tag with the error title anymore

test/testcafe/framework/page-objects/BasePageObject.js
```diff
@@ -210,6 +210,10 @@ export default class BasePageObject {
     return this.form.getErrorBoxText();
   }
 
+  getErrorTitle() {
+    return this.form.getErrorTitle();
+  }
+
```
leemchale-okta, line 216: moved to base page object so it can be widely used

test/testcafe/framework/page-objects/ChallengeOktaVerifyPushPageObject.js
```diff
@@ -63,18 +62,6 @@ export default class ChallengeOktaVerifyPushPageObject extends ChallengeFactorPa
...
-    if (userVariables.gen3) {
-      return this.form.getNthTitle(0);
-    }
-    return this.form.getElement(FORM_INFOBOX_ERROR_TITLE).innerText;
-  }
-
-  getFormTitleWithError() {
-    const titlePosition = userVariables.gen3 ? 1 : 0;
-    return this.form.getNthTitle(titlePosition);
-  }
```
leemchale-okta, line 76: getFormTitleWithError() no longer needed since both form title and error title do not share an h2 tag anymore

test/testcafe/framework/page-objects/components/BaseFormObject.js
```diff
@@ -59,6 +60,19 @@ export default class BaseFormObject {
...
+      return within(this.getErrorBox()).findByRole('heading', { level: 2 }).innerText;
+    }
+    return this.getElement(FORM_INFOBOX_ERROR_TITLE).innerText;
+  }
+
+  getNthErrorTitle(index) {
+    return within(this.getErrorBox()).findAllByRole('heading', {
+      level: 2,
+    }).nth(index).innerText;
+  }
```
leemchale-okta, line 74: moved to base form object here to avoid duplication

src/v3/src/components/ActionPending/ActionPending.tsx
```diff
@@ -55,7 +55,7 @@ const ActionPending: UISchemaElementComponent<{
         }}
       >
         <Typography
-          component="h2"
+          component="h1"
```
shuowu-okta, line 58: Is this mandatory for this fix?

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
```diff
@@ -243,6 +243,7 @@ const AuthenticatorButton: UISchemaElementComponent<{
         }}
       >
         <Typography
+          component="h2"
```
Luxu2-Okta, line 246: similar question as @shuowu-okta , the customer's original request is for a specific view `"Verify its you" page title` from h2 to h1, and here we change the component from h3 to h2, I guess this is what we want to improve the a11y in general, but just would like to double check

src/v3/src/components/ActionPending/ActionPending.tsx
leemchale-okta: Its not mandatory, but it is the main heading for the pages it is used in, so technically there is a missing h1 on those pages, and our tests look for it to be the form title, so I think it makes sense to update it to h1 along with the other titles to keep it all consistent.

src/v3/src/components/AuthenticatorButton/AuthenticatorButton.tsx
leemchale-okta: @Luxu2-Okta yes this is to maintain header level order on the page so we don't violate accessiblity rules.  The `Title` component is shared on many pages so we cannot only make the change to the "Verify its you" page.  Well technically we could but want all pages to be consistent not just change one.  
Since we changed the `Title` component to h1, and these AuthenticatorButtons were h3, there was a gap (no h2 on the page), so these buttons needed to be bumped up to h2.

PR #3865, opened 2025-07-31
fix(oie): Empty custom i18n value fix

leemchale-okta:
## Description:

This PR fixes an issue where an L10N_ERROR translation error is shown when overriding an i18n key with an empty string.  

**Before:**

Screenshot 2025-07-31 at 3 35 59 PM

**After:**

Screenshot 2025-07-31 at 3 34 55 PM

## PR Checklist

### Issue:

- [OKTA-986305](https://oktainc.atlassian.net/browse/OKTA-986305)

### DIFF
--- a/src/v2/ion/i18nUtils.js
+++ b/src/v2/ion/i18nUtils.js
@@ -393,7 +393,7 @@ const isCustomizedI18nKey = (i18nKey, settings) => {
   
   // Lower case the language codes when searching for i18n key so it is case insensitive
   return Object.entries(i18n).some(
-    ([customizedLangCode, customizedI18nKeys]) => customizedLangCode.toLowerCase() === widgetLanguageLowerCase && customizedI18nKeys[i18nKey] !== undefined
+    ([customizedLangCode, customizedI18nKeys]) => customizedLangCode.toLowerCase() === widgetLanguageLowerCase && !!customizedI18nKeys[i18nKey]
   );
 };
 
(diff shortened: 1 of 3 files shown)


alexchen-okta (dismissed):

src/v2/ion/i18nUtils.js
```diff
@@ -393,7 +393,7 @@ const isCustomizedI18nKey = (i18nKey, settings) => {
   
   // Lower case the language codes when searching for i18n key so it is case insensitive
   return Object.entries(i18n).some(
-    ([customizedLangCode, customizedI18nKeys]) => customizedLangCode.toLowerCase() === widgetLanguageLowerCase && customizedI18nKeys[i18nKey] !== undefined
+    ([customizedLangCode, customizedI18nKeys]) => customizedLangCode.toLowerCase() === widgetLanguageLowerCase && !!customizedI18nKeys[i18nKey]
```
Luxu2-Okta, line 396: can we confirm Gen3 doesn't have this issue? I did some quick search [here](https://github.com/okta/okta-signin-widget/blob/4a014535b359b958ba8c957259f2f807025a2d84/src/util/loc.ts#L88-L89), it seems empty `value` will cause the issue in Gen3

leemchale-okta: yeah actually my mistake, I tested locally before and didn't see the issue in gen 3 for some reason, but I just double checked and actually do see the issue now in gen 3 too.  Will make the same change in gen 3.

Luxu2-Okta: Let's also add some test for empty string , unit test in `src/v3/src/transformer/i18n/isCustomizedI18nKey.test.ts`?

Luxu2-Okta (approved):
great fix! let's target 7.34.1

PR #3874, opened 2025-08-28
fix(gen3): security question verification i18n error

Luxu2-Okta:
fix the i18n error when configured security question as verification authenticator

Resolves: OKTA-998463

## Description:
The error happens when the idx response doesn't have the `profile` property in the `/idp/idx/challenge` response, e.g [this one](https://github.com/okta/okta-signin-widget/blob/928d0ba9b178339037947f2aba285233871ac7d0/src/v3/src/mocks/response/idp/idx/identify/securityquestion-verify.json#L106-L109), also can refer to the `har` file in the ticket

It works in Gen2 but not in Gen3 because Gen2 and Gen3 are implemented differently.
 
#### Gen2

This is the high-level workflow of security question as verification authenticator in Gen2
* call [uiSchemaTransformer.js](https://github.com/okta/okta-signin-widget/blob/928d0ba9b178339037947f2aba285233871ac7d0/src/v2/ion/uiSchemaTransformer.js#L38) first, it will get uiSchema from the [remediation form values here](https://github.com/okta/okta-signin-widget/blob/928d0ba9b178339037947f2aba285233871ac7d0/src/v3/src/mocks/response/idp/idx/identify/securityquestion-verify.json#L24-L41)
Screenshot 2025-08-27 at 4 25 34 PM

* call [i18nTransformer.js](https://github.com/okta/okta-signin-widget/blob/928d0ba9b178339037947f2aba285233871ac7d0/src/v2/ion/i18nTransformer.js#L72), but now the data will be in uiSchema, so the translation is correct, see in the screenshot below, the `disliked_food` key is obtained
Screenshot 2025-08-27 at 3 41 38 PM

#### Gen3
Gen3 the [securityQuestionVerify.ts](https://github.com/okta/okta-signin-widget/blob/928d0ba9b178339037947f2aba285233871ac7d0/src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts#L41-L47) transform happens on the layout layer, which is before [uiSchema and i18n](https://github.com/okta/okta-signin-widget/blob/928d0ba9b178339037947f2aba285233871ac7d0/src/v3/src/transformer/main.ts#L42-L51), to be secure, we should not change the transform order, so the fix would be still in the layout layer

When profile is missing, the key would be `security.undefined`
Screenshot 2025-08-28 at 3 42 20 PM

To fix it, we need to access the remediation form data, and we had similar implementation in [securityQuestionEnroll.ts](https://github.com/okta/okta-signin-widget/blob/928d0ba9b178339037947f2aba285233871ac7d0/src/v3/src/transformer/layout/securityQuestion/securityQuestionEnroll.ts#L50-L51)

## PR Checklist

### Issue:

- [OKTA-998463](https://oktainc.atlassian.net/browse/OKTA-998463)

#### Before fix
Gen2:

https://github.com/user-attachments/assets/be01b5c9-c197-4f09-baf2-c32fcaf40a75

Gen3:

https://github.com/user-attachments/assets/7358fffb-849a-4b7b-9fe5-f3b200af01f1

#### After fix
Gen2: works same
Gen3:

https://github.com/user-attachments/assets/97059f41-4387-4c16-8237-6a93d8a5b9d0

### DIFF
--- a/src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
+++ b/src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
@@ -50,8 +50,11 @@ export const transformSecurityQuestionVerify: IdxStepTransformer = ({ transactio
   } else {
     // get the security question from inputs and i18n
     Logger.warn('Security question key missing from profile, getting from inputs');
-    const securityQuestionKey = (inputs?.[0]?.value as Input[])?.find(({ name }) => name === 'questionKey')?.value;
-    securityQuestion = loc(`security.${securityQuestionKey}`, 'login');
+    const sequrityQuestionInput = (inputs?.[0]?.value as Input[])?.find(({ name }) => name === 'questionKey');
+    const securityQuestionKey = sequrityQuestionInput?.value;
+    securityQuestion = securityQuestionKey === 'custom' ?
+      sequrityQuestionInput?.label as string :
+      loc(`security.${securityQuestionKey}`, 'login');
   }
   answerElement.translations = [{
     name: 'label',
--- a/src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
+++ b/src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
@@ -52,9 +52,9 @@ export const transformSecurityQuestionVerify: IdxStepTransformer = ({ transactio
     Logger.warn('Security question key missing from profile, getting from inputs');
     const sequrityQuestionInput = (inputs?.[0]?.value as Input[])?.find(({ name }) => name === 'questionKey');
     const securityQuestionKey = sequrityQuestionInput?.value;
-    securityQuestion = securityQuestionKey === 'custom' ?
-      sequrityQuestionInput?.label as string :
-      loc(`security.${securityQuestionKey}`, 'login');
+    securityQuestion = securityQuestionKey === 'custom'
+      ? sequrityQuestionInput?.label as string
+      : loc(`security.${securityQuestionKey}`, 'login');
   }
   answerElement.translations = [{
     name: 'label',
--- a/src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
+++ b/src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { NextStep } from '@okta/okta-auth-js';
+import { Input, NextStep } from '@okta/okta-auth-js';
 
 import {
   ButtonElement,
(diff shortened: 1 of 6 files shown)


src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
```diff
@@ -38,12 +39,24 @@ export const transformSecurityQuestionVerify: IdxStepTransformer = ({ transactio
...
+    securityQuestion = relatesTo?.value?.profile?.question as string;
+  } else if (securityQuestionProfileKey) {
+    // get the security question from i18n with valid key
+    securityQuestion = loc(`security.${securityQuestionProfileKey}`, 'login');
+  } else {
+    // get the security question from inputs and i18n
+    Logger.warn('Security question key missing from profile, getting from inputs');
+    const securityQuestionKey = (inputs?.[0]?.value as Input[])?.find(({ name }) => name === 'questionKey')?.value;
+    securityQuestion = loc(`security.${securityQuestionKey}`, 'login');
+  }
```
alexchen-okta, line 58: if i'm understanding the gen2 code correctly, it gets the security question key from the remediation form data. is there any reason why we don't match the implementation in gen2 exactly rely only on the response from `inputs` in all cases regardless of `profile`?

Luxu2-Okta: @alexchen-okta Yes! Gen3 was using profile [in the first commit](https://github.com/okta/okta-signin-widget/blame/af10e508eeaa0ab1113f69a18d4ecbe6c290ddf0/src/v3/src/transformer/securityQuestion/securityQuestionVerify.ts#L30), and localization was added later [here](https://github.com/okta/okta-signin-widget/blame/928d0ba9b178339037947f2aba285233871ac7d0/src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts#L44-L46), but still using `profile` @lesterchoi-okta @glenfannin-okta do you have any ideas?

Luxu2-Okta: I just realized the `securityQuestionKey` from remediation form data can also be `'custom'`, I will improve the PR

src/v3/src/transformer/layout/securityQuestion/securityQuestionVerify.ts
```diff
@@ -38,12 +39,24 @@ export const transformSecurityQuestionVerify: IdxStepTransformer = ({ transactio
...
   ) as FieldElement;
+  const securityQuestionProfileKey = relatesTo?.value?.profile?.questionKey;
+  let securityQuestion = '';
+  if (securityQuestionProfileKey === 'custom') {
+    // get the custom security question from profile
+    securityQuestion = relatesTo?.value?.profile?.question as string;
+  } else if (securityQuestionProfileKey) {
+    // get the security question from i18n with valid key
+    securityQuestion = loc(`security.${securityQuestionProfileKey}`, 'login');
+  } else {
```
alexchen-okta, line 50: do we know in which cases the `profile` property in the idx response won't be present? is there something specific to OAMP that triggers this behavior?

Luxu2-Okta: @alexchen-okta , we are not certain the case `profile` missing, from Arsalan's reproduce, it is indeed with OAMP, but I can also try ASoP

PR #3884, opened 2025-09-17
OKTA-954339 - Add ALTCHA Integration into signin widget

ahmedalkoasmh-okta:
## Description:
Putting out this PR on behalf of Lijo Daniel and retrieving it from [his forked repo](https://github.com/lijodaniel-okta/okta-signin-widget-altcha/tree/cherry-from-local) since he is still working on getting access.

This PR adds integration for the new ALTCHA library into the sign-in widget repo.

The ALTCHA widget is treated as a CAPTCHA and follows the same flow as other CAPTCHAs to load into the page.

It is meant to work on both v2 and v3 versions of the widget.

## PR Checklist

### Issue:

- [OKTA-954339](https://oktainc.atlassian.net/browse/OKTA-954339)

### Screenshot/Video:
https://github.com/user-attachments/assets/6e2926d4-f336-444b-9837-e18169f91cf6

RECAPTCHA working with changes:

https://github.com/user-attachments/assets/40d0c397-b536-4d18-a8d1-d4d6b816646a

HCAPTCHA working with changes:

https://github.com/user-attachments/assets/3cfa19be-4ab9-439c-92b3-c4f8fc17e129

Build and CAPTCHA tests passing: https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&author=ahmed.alkoasmh%40okta.com&branch=d16t-okta-signin-widget-1a75565-68d30816&page=1&pageSize=6&tab=main

### DIFF
--- a/assets/sass/widgets/_altcha.scss
+++ b/assets/sass/widgets/_altcha.scss
@@ -1,15 +1,13 @@
-div {
-    .altcha {
-        background: #fff;
-        border: 1px solid #a0a0a0;
-    }
+.altcha {
+    background: #fff;
+    border: 1px solid #a0a0a0;
+}
 
-    .altcha-main {
-        padding: 12px;
-    }
+.altcha-main {
+    padding: 12px;
+}
 
-    .altcha-anchor-arrow {
-        border: 6px solid transparent;
-        border-bottom-color: #a0a0a0;
-    }
+.altcha-anchor-arrow {
+    border: 6px solid transparent;
+    border-bottom-color: #a0a0a0;
 }
--- a/assets/sass/widgets/_altcha.scss
+++ b/assets/sass/widgets/_altcha.scss
@@ -1,0 +1,15 @@
+div {
+    .altcha {
+        background: #fff;
+        border: 1px solid #a0a0a0;
+    }
+
+    .altcha-main {
+        padding: 12px;
+    }
+
+    .altcha-anchor-arrow {
+        border: 6px solid transparent;
+        border-bottom-color: #a0a0a0;
+    }
+}
--- a/src/v2/view-builder/internals/BaseForm.ts
+++ b/src/v2/view-builder/internals/BaseForm.ts
@@ -70,7 +70,7 @@ export default Form.extend({
     this.$el.find('.o-form-error-container').empty();
 
     // Execute Captcha if enabled for this form.
-    if (this.captchaObject) {
+    if (this.captchaObject && (this.captchaObject.tagName || '').toUpperCase() !== 'ALTCHA-WIDGET') {
       this.captchaObject.execute();
     } else {
       this.options.appState.trigger('saveForm', model);
--- a/src/v2/view-builder/views/captcha/CaptchaView.js
+++ b/src/v2/view-builder/views/captcha/CaptchaView.js
@@ -42,7 +42,7 @@ export default View.extend({
       let className;
 
       if (this.captchaConfig.type === 'ALTCHA') {
-        className = 'altcha';
+        className = 'altcha-captcha';
       } else if (this.captchaConfig.type === 'HCAPTCHA') {
         className = 'h-captcha';
       } else {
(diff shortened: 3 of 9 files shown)


src/v2/view-builder/views/captcha/CaptchaView.js
```diff
@@ -147,6 +189,29 @@ export default View.extend({
...
+          console.log('captcha lib loaded but no global onload callback found');
+        }
+      } catch (e) {
+        console.error('Error invoking captcha onload callback', e);
+      }
+    };
+
+    scriptTag.onerror = (evt) => {
+      console.error('Failed to load captcha script:', url, evt);
+    };
```
leemchale-okta: are `scriptTag.crossOrigin`, `scriptTag.onload`, and `scriptTag.onerror` only used for Altcha?  Or also used for HCaptcha and ReCaptcha?

leemchale-okta:
For gen 3 we may need to add some logic [here](https://github.com/okta/okta-signin-widget/blob/ce64ca7c41cccbdcc93a141b83d416e5a0213368/src/v3/src/components/Form/Form.tsx#L82) to prevent trying to execute on the `captchaRef` since it looks like the altcha implementation does not use the `captchaRef`

leemchale-okta:
Has the backend part of this feature already been implemented?  We may need to document that this feature can only be used on certain SIW version and higher, since if a customer tries to use this new Altcha on an older SIW it might break the widget.

src/v2/view-builder/views/captcha/CaptchaView.js
ahmedalkoasmh-okta: I'v been reviewing and refactoring these changes.

- `crossOrigin` is `anonymous` by default for any script added ([see here](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/crossorigin)). I believe Lijo added it so we explicitly set it while loading the script. We can remove it if you prefer, but I think it's a good idea to explicity reference the property.
- `onload` is required for `ALTCHA` - for reCAPTCHA and hCAPTCHA, the `onload` function is passed as a query param and the function is called by the loaded third party widget. For ALTCHA, we must control this logic. I refactored the code to remove setting `onload` here and instead set it from the `_addCaptcha` function to add a better separation of concern. [See here](https://github.com/okta/okta-signin-widget/pull/3884/files#diff-2e6d35d864abad8cfc56789e3e956f46e0ed90d2cc3719683a7c6cde355b5170R171).
- `onerror` is not needed. I think it was added for testing. Removed.

ahmedalkoasmh-okta:
> Has the backend part of this feature already been implemented? We may need to document that this feature can only be used on certain SIW version and higher, since if a customer tries to use this new Altcha on an older SIW it might break the widget.

Not yet. This feature will not be rolled out right away, and will be behind a flag.

Yes, we definitely need to document information about the version. Will message with details.

ahmedalkoasmh-okta:
> For gen 3 we may need to add some logic [here](https://github.com/okta/okta-signin-widget/blob/ce64ca7c41cccbdcc93a141b83d416e5a0213368/src/v3/src/components/Form/Form.tsx#L82) to prevent trying to execute on the `captchaRef` when submitting since it looks like the altcha implementation does not use the `captchaRef`

Took a look and I do not think we need to add logic because `captchaRef` will be `null`, so the code inside the condition will not be executed.

src/v2/view-builder/internals/BaseForm.ts
```diff
@@ -70,7 +70,7 @@ export default Form.extend({
     this.$el.find('.o-form-error-container').empty();
 
     // Execute Captcha if enabled for this form.
-    if (this.captchaObject) {
+    if (this.captchaObject && (this.captchaObject.tagName || '').toUpperCase() !== 'ALTCHA-WIDGET') {
```
ahmedalkoasmh-okta, line 73: Question for Lijo - why do we need to do this? Did a debug and ALTCHA never gets here.

assets/sass/widgets/_altcha.scss
```diff
@@ -0,0 +1,15 @@
+div {
```
shuowu-okta: not need to wrap it within a div block

src/v2/view-builder/views/captcha/CaptchaView.js
```diff
@@ -20,6 +20,7 @@ const OktaSignInWidgetOnCaptchaLoadedCallback = 'OktaSignInWidgetOnCaptchaLoaded
 const OktaSignInWidgetOnCaptchaSolvedCallback = 'OktaSignInWidgetOnCaptchaSolved';
 
 const HCAPTCHA_URL = 'https://hcaptcha.com/1/api.js';
+const ALTCHA_URL = 'https://cdn.jsdelivr.net/gh/altcha-org/altcha/dist/altcha.min.js';
```
shuowu-okta, line 23: Does backend handle CSP for this asset?

ahmedalkoasmh-okta: Yes correct, it will be handled accordingly.

src/v2/view-builder/views/captcha/CaptchaView.js
leemchale-okta: got it, lgtm

leemchale-okta:
> Not yet. This feature will not be rolled out right away, and will be behind a flag.
> 
> Yes, we definitely need to document information about the version before EA. Will message with details.

Okay got it sounds good thank you 👍

leemchale-okta:
> > For gen 3 we may need to add some logic [here](https://github.com/okta/okta-signin-widget/blob/ce64ca7c41cccbdcc93a141b83d416e5a0213368/src/v3/src/components/Form/Form.tsx#L82) to prevent trying to execute on the `captchaRef` when submitting since it looks like the altcha implementation does not use the `captchaRef`
> 
> Took a look and I do not think we need to add logic because `captchaRef` will be `null`, so the code inside the condition will not be executed.

I see yes that looks to be correct, thanks for checking.

assets/sass/widgets/_altcha.scss
ahmedalkoasmh-okta: Updated. Thank you!

package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "name": "@okta/okta-signin-widget",
   "description": "The Okta Sign-In Widget",
-  "version": "7.35.1",
+  "version": "7.36.0",
```
alexchen-okta: can you check out this change?

ahmedalkoasmh-okta: Done

PR #3888, opened 2025-09-22
Chrome Local Network Access Remediation

trevoring-okta:
## PR Checklist

### Issue:

- [OKTA-998031](https://oktainc.atlassian.net/browse/OKTA-998031)


src/v3/src/transformer/layout/oktaVerify/transformOktaVerifyFPLoopbackPoll.ts
```diff
@@ -10,42 +10,109 @@
...
+} from '../../../util';
 
 export const transformOktaVerifyFPLoopbackPoll: IdxStepTransformer = ({
+  prevTransaction,
   transaction,
   formBag,
 }) => {
+  const { nextStep: { name: prevStepName } = {} as NextStep } = prevTransaction ?? {};
+  // If there is no previous transaction, it is a silent probe triggered by the registered condition
+  const isRegisteredConditionSilentProbe = prevStepName === undefined;
```
trevoring-okta, line 42: [note] Will explain this more at RR meeting, but this handles one edge case where the permission state could be `'denied'`, but we don't actually want to immediately show the remediation view

src/v2/view-builder/utils/ChallengeViewUtil.js
```diff
@@ -30,19 +32,98 @@ export function appendLoginHint(deviceChallengeUrl, loginHint) {
...
+  view.removeChildren();
+  switch(currPermissionState) {
+  case 'prompt':
+  case 'granted':
+    updateChromeLNATitle(view, loc('deviceTrust.sso.redirectText', 'login'));
+    addLoopbackView(view, deviceChallenge);
+    break;
+  case 'denied': {
+    // If there is no previous form name, it is a silent probe triggered by the registered condition
+    const isRegisteredConditionSilentProbe = view.options?.appState?.get('previousFormName') === undefined;
```
trevoring-okta, line 88: [note] Same as comment for corresponding change in g3 code: will explain this more at RR meeting, but this handles one edge case where the permission state could be 'denied', but we don't actually want to immediately show the remediation view

src/util/BrowserFeatures.ts
```diff
@@ -84,6 +86,26 @@ fn.isIOS = function() {
   return /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream;
 };
 
+fn.getChromeLNAPermissionState = async function(
+  handlePermissionState: (currPermissionState?: PermissionState) => void
+) {
+  try {
+    if (navigator.permissions && typeof navigator.permissions.query === 'function') {
+      const result = await navigator.permissions.query({ name: 'local-network-access' as any });
```
alexchen-okta, line 94: for users on older versions of Chrome (i.e. that don't support the new permission prompt for LNA), what would be the output here for `result`? will this be handled gracefully in the subsequent line?

src/v2/view-builder/utils/ChallengeViewUtil.js
```diff
@@ -30,19 +32,98 @@ export function appendLoginHint(deviceChallengeUrl, loginHint) {
...
+    if (isRegisteredConditionSilentProbe) {
+      updateChromeLNATitle(view, loc('deviceTrust.sso.redirectText', 'login'));
+      addLoopbackView(view, deviceChallenge);
+    } else {
+      updateChromeLNATitle(view, loc('chrome.lna.fastpass.requires.permission.title', 'login'));
+      addLNAErrorView(view, deviceChallenge.chromeLocalNetworkAccessDetails?.chromeLNAHelpLink);
+      // eslint-disable-next-line max-depth
+      if (process.env.NODE_ENV === 'production') {
+        // Log error for Sentry monitoring in production
+        throw new ChromeLNADeniedError('Chrome Local Network Access permission was denied for FastPass.');
```
alexchen-okta: @shuowu-okta  to check for Sentry monitoring related changes

playground/mocks/data/idp/idx/identify-with-device-probing-loopback-chrome-lna.json
```diff
@@ -0,0 +1,91 @@
+{
```
alexchen-okta, line 1: just want to confirm -- are the changes both forward and backwards compatible? i.e.
- do these changes work without the additive IDX changes?
- with the new IDX changes, can the widget be pinned to an older version and still be functioning as expected?

src/v2/view-builder/utils/ChallengeViewUtil.js
```diff
@@ -30,19 +32,98 @@ export function appendLoginHint(deviceChallengeUrl, loginHint) {
...
+  view.title = newTitle;
+  if (view.$ && view.$('.okta-form-title').length) {
+    view.$('.okta-form-title').text(newTitle);
+  }
+}
+
+function handlePermissionState(view, currPermissionState, deviceChallenge) {
+  view.removeChildren();
+  switch(currPermissionState) {
+  case 'prompt':
```
alexchen-okta, line 81: if we proceed with the loopback here in the `prompt` case, isn't the user going to fail since they did not grant permission?

trevoring-okta: Good question: in practice, we have to proceed with loopback when the state is `prompt` to actually initiate a local network access request that will trigger the prompt to appear, so it's not guaranteed that the user will "fail" loopback in the `prompt` case because that depends on what action they take when the prompt appears.

That being said, your question actually makes me realize I should technically change some of my TestCafe cases to be more semantically correct. I've been looping `prompt` and `granted` test cases together as situations where loopback "succeeds", but the `prompt` case actually depends on what action the user takes when the prompt appears. I'm going to look into creating a specific `prompt` case right now 👍

src/util/BrowserFeatures.ts
trevoring-okta: Yeah, this is handled gracefully. Older versions of Chromium browsers AND non-Chromium browsers have no notion of the `local-network-access` permission and throw a `TypeError`. This gets caught by the `catch` block and I chose to pass `undefined` to `handlePermissionState`, which eventually flows into the `default` case in the switch statement that proceeds w/ loopback as normal.

playground/mocks/data/idp/idx/identify-with-device-probing-loopback-chrome-lna.json
trevoring-okta: > do these changes work without the additive IDX changes?

Yeah, all changes assert on the presence of `chromeLocalNetworkAccessDetails` in the IDX response before doing anything, so if the IDX changes aren't there we will just do loopback without our extra logic

> with the new IDX changes, can the widget be pinned to an older version and still be functioning as expected?

Yes for the same reason as above. If an old SIW version is sent an IDX payload with `chromeLocalNetworkAccessDetails` but it doesn't contain the code to handle it, it won't know to process that part of the payload and will proceed normally

PR #3893, opened 2025-09-25
fix(gen3): multiple steps polling

shuowu-okta:
## Description:

The change is verified with unit test

Before fix

Final mocks call snapshot
```
Array [
  Array [
    Object {
      "stateHandle": undefined,
      "step": "challenge-poll",
    },
  ],
  Array [
    Object {
      "stateHandle": undefined,
      "step": "challenge-poll",
    },
  ],
  Array [
    Object {
      "stateHandle": undefined,
      "step": "device-challenge-poll",
    },
  ],
  Array [
    Object {
      "stateHandle": undefined,
      "step": "challenge-poll",
    },
  ],
]
```

After fix:

Final mocks call snapshot
```
Array [
  Array [
    Object {
      "stateHandle": undefined,
      "step": "challenge-poll",
    },
  ],
  Array [
    Object {
      "stateHandle": undefined,
      "step": "challenge-poll",
    },
  ],
  Array [
    Object {
      "stateHandle": undefined,
      "step": "device-challenge-poll",
    },
  ],
]
```

## PR Checklist

### Issue:

- [OKTA-938817](https://oktainc.atlassian.net/browse/OKTA-938817)

### DIFF
--- a/src/v3/src/hooks/usePolling.test.tsx
+++ b/src/v3/src/hooks/usePolling.test.tsx
@@ -361,6 +361,5 @@ describe('usePolling', () => {
       // Validate proceed call step sequence
       expect(mockProceedFn.mock.calls).toMatchSnapshot();
     });
-
   });
 });


src/v3/src/hooks/usePolling.ts
```diff
@@ -64,13 +64,12 @@ export const usePolling = (
   const timerRef = useRef<NodeJS.Timeout>();
 
   const pollingStep = useMemo(() => {
-    const idxTransactionPollingStep = getPollingStep(idxTransaction);
-    if (!idxTransactionPollingStep) {
-      return undefined;
+    // If an internal transaction exists, rely solely on it for polling step detection.
+    // This prevents resuming polling after reaching a non-polling (e.g. terminal) state.
+    if (transaction) {
```
leemchale-okta, line 69: was the problem happening because this `transaction` state was more "up to date" than `idxTransaction`?  So `idxTransaction` was sort of one step behind `transaction`, causing the old logic to think the widget was still on a polling step but actually wasn't?  Just want to make sure im understanding this right.

shuowu-okta: In general, the usePolling hook should take the initial transaction from input (the idxTransaction) and start polling process when the input has a polling step as the next step, then keep following new transaction to keep polling as needed. And it should stop making polling request when a non-polling step is meet.

As you can see from the mock calls before fix, the last `challenge-poll` in the call array is unexpected, as it should stop at device-challenge-poll. This PR is targeting to this issue.

leemchale-okta: got it thanks for explaining 🙏

leemchale-okta (dismissed):
lgtm!

src/v3/src/hooks/usePolling.ts
Luxu2-Okta: @shuowu-okta , I guess this fix will solve the this `Unknown remediation choice [challenge-poll]` error?
image
And I guess that error is from this [idx.proceed](https://github.com/okta/okta-signin-widget/blob/68f756c6b72402e9f433e9bfae0a1100a4928b7b/src/v3/src/hooks/usePolling.ts#L101) method

I am curious if there are still `challenge-poll` after `device-challenge-poll`, does SIW still actually make the `challenge-poll: /idp/idx/authenticators/poll` polling?
for example, the remediation order is
1. `challenge-poll`: SIW request `/idp/idx/authenticators/poll`
2. `device-challenge-poll`: SIW request `/idp/device/poll`
3. ... (optional)
4. `challenge-poll`: SIW request `/idp/idx/authenticators/poll`

Does step 4 still happens? I just tried workday that successfully login, we will have `challenge-poll` after `device-challenge-poll`, just want to double check this to prevent regression

shuowu-okta: In the mock sequence, the last `challege-poll` is considered as unexpected, as there is no guidance from serve side to take it as the next polling. 

As long as `device-challenge-poll` is followed with an expected `challenge-poll`, it would be executed, I will also update a unit test with that scenario.

leemchale-okta (approved):

src/v3/src/hooks/usePolling.ts
Luxu2-Okta: Thanks @shuowu-okta !

PR #3945, opened 2026-01-26
feat: support new loopback-network permission for chrome 145+

trevoring-okta:
## Description:
- In Chromium 145, the existing `local-network-access` permission that we have been relying on to query whether a client has granted permission for "[Local Network Access](https://developer.chrome.com/blog/local-network-access)" requests (in our case, we care about the SIW making a loopback probing request to OV) will become an aggregate of two more granular permissions: `local-network` and `loopback-network` (Chrome adoption guide [here](https://docs.google.com/document/d/1QQkqehw8umtAgz5z0um7THx-aoU251p705FbIQjDuGs/edit?tab=t.0) for more info)
- For OV FP loopback to work, we only need 145+ clients to grant us access to the `loopback-network` permission. Chrome has said that all clients that upgrade to 145 will continue to work if they have already granted `local-network-access` permission, but if we do not update our permission querying logic to use the more granular `loopback-network` permission, we could unnecessarily deny auth to users who may have denied access to the other — irrelevant — `local-network` permission

- This PR introduces: 
    - Nested `try/catch` to first query for the new `loopback-network` permission. If that returns an error (e.g. browser is < 145 and doesn't know about this new perm), we will catch and fall back to the older `local-network-access` permission
    - We also updated our access denied messaging to reflect that the new Chrome permission tied to the `loopback-network` permission is called "Apps on device" in Chrome site settings

## PR Checklist

### Issue:

- [OKTA-1101159](https://oktainc.atlassian.net/browse/OKTA-1101159)

#### Testing on Chrome 145 browser — confirming that `Apps on devices` (`loopback-network) permission is the only one that gates access in 145+
https://github.com/user-attachments/assets/a78d76cb-0b3f-419c-96c0-b9aefb3349ae

#### Testing on Chrome < 145 browser — confirming that `Local network access` permission is the only one that gates access in < 145 (i.e. shows existing behavior)
https://github.com/user-attachments/assets/29877cd0-6c46-441c-be8a-d3c75538d3c1

### DIFF
--- a/src/util/BrowserFeatures.ts
+++ b/src/util/BrowserFeatures.ts
@@ -86,32 +86,36 @@ fn.isIOS = function() {
   return /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream;
 };
 
+/* eslint-disable compat/compat */
 fn.getChromeLNAPermissionState = async function(
   handlePermissionState: (currPermissionState?: PermissionState) => void
 ) {
+  if (!navigator.permissions || typeof navigator.permissions.query !== 'function') {
+    handlePermissionState(undefined);
+    return;
+  }
+
   try {
-    if (navigator.permissions && typeof navigator.permissions.query === 'function') {
-      let result: PermissionStatus;
-      try {
-        // Query for the new granular `loopback-network` permission only available on Chromium 145+
-        result = await navigator.permissions.query({ name: 'loopback-network' as any });
-      } catch (e) {
-        // Fallback to the legacy `local-network-access` aggregate permission on < Chromium 145
-        result = await navigator.permissions.query({ name: 'local-network-access' as any });
-      }
-      
-      handlePermissionState(result.state);
-    } else {
-      // Fallback for browsers that do not support Permissions API
-      handlePermissionState(undefined);
+    let result: PermissionStatus;
+    try {
+      // Query for the new granular `loopback-network` permission only available on Chromium 145+
+      result = await navigator.permissions.query({ name: 'loopback-network' as any });
+    } catch (e) {
+      // Fallback to the legacy `local-network-access` aggregate permission on < Chromium 145
+      result = await navigator.permissions.query({ name: 'local-network-access' as any });
     }
+    
+    handlePermissionState(result.state);
   } catch (error) {
     if (error instanceof ChromeLNADeniedError) {
       throw error; // Rethrow custom Chrome LNA denied error so Sentry can capture it for monitoring
     }
+
+    // Fallback if queries fail or browser blocks the API entirely
     handlePermissionState(undefined);
   }
 };
+/* eslint-enable compat/compat */
 
 // Returns a list of languages the user has configured for their browser, in
 // order of preference.


src/v3/src/util/browserUtils.ts
```diff
@@ -28,19 +28,27 @@ export const isIOS = (): boolean => (
 export const getChromeLNAPermissionState = async (
   handlePermissionState: (currPermissionState?: PermissionState) => void,
 ) => {
-  try {
-    if (navigator.permissions && typeof navigator.permissions.query === 'function') {
-      const result = await navigator.permissions.query({ name: 'local-network-access' as any });
+  if (!navigator.permissions || typeof navigator.permissions.query !== 'function') {
+    handlePermissionState(undefined);
+    return;
+  }
```
alexchen-okta, line 34: could we keep the logic for `getChromeLNAPermissionState` consistent with the function in `src/util/BrowserFeatures.ts`? I think these are functionally equivalent but there's a minor difference in logic

test/testcafe/spec/DeviceChallengePollView_spec.js
```diff
@@ -402,8 +402,9 @@ const mockChromeLNAPermissionsQuery = ClientFunction((permissionState) => {
   // Replace the original query method with a mock
   const originalQuery = navigator.permissions.query;
   navigator.permissions.query = (permissionDesc) => {
-    // Return a promise that resolves to the specified state
-    if (permissionDesc.name === 'local-network-access') {
+    // Return a promise that resolves to the specified state for either the legacy 'local-network-access'
+    // permission name (< Chromium 145) or the new 'loopback-network' name (Chromium 145+)
+    if (permissionDesc.name === 'local-network-access' || permissionDesc.name === 'loopback-network') {
```
alexchen-okta, line 407: nit: wondering if you see any value in storing these chrome permission names as constants?

trevoring-okta: Sure 👍  Don't think the Chrome team would ever change existing permission names, but I think it could improve readability + style here

trevoring-okta: Actually, I'll explore refactoring this if one additional permission name gets added. I think as it stands right now, there's few enough references in code and across different parts of the codebase (gen2, gen3, tests) and we don't have a super logical singular location to add these constants, imo

src/v3/src/util/browserUtils.ts
Luxu2-Okta, line 34: Just curious, is `getChromeLNAPermissionState` function will only be called on Chrome? Seems it is determined by the idx response. I am asking because the naviagor.permissions is not available on IE11, this sanity checking is great, maybe it would be better to run some e2e test on IE11 (if we never tried on IE11 before)

trevoring-okta: You are right that the function is called by the IDX response. The `!navigator.permissions` check should work to protect us from runtime errors on IE11 since `navigator.permissions` will return `undefined` on IE11, but I can see if I can do a test on SauceLabs to verify

trevoring-okta: Confirmed that my demo cell org loads fine in IE11 even with the `navigator.permissions` code

PR #3954, opened 2026-01-30
Okta 1105231 update altcha submission

ahmedalkoasmh-okta:
## Description:
When ALTCHA captcha was solved, it was directly triggering the form submission, bypassing the parent form's save method which is responsible for generating the device fingerprint. Unlike HCAPTCHA/reCAPTCHA where the captcha is executed after the form save flow starts, ALTCHA is a floating overlay widget that users interact with directly before form submission.

This fix introduces a new event that ALTCHA triggers when verification completes, which the base form listens for and properly routes through the form's save flow. This ensures the device fingerprint is generated before the form is submitted to the server. The existing HCAPTCHA and reCAPTCHA flows remain unchanged. 

## PR Checklist

### Issue:

- [OKTA-1105231](https://oktainc.atlassian.net/browse/OKTA-1105231)

### Reviewers:
@alexchen-okta

### Screenshot/Video:
https://github.com/user-attachments/assets/c4b5ea1a-2a1d-4638-b283-f00a0411caec

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&page=1&pageSize=6&sha=98439d5bdb6ebb5fa3f080dfa121018e4565b119&tab=main

### DIFF
--- a/src/v2/view-builder/internals/BaseForm.ts
+++ b/src/v2/view-builder/internals/BaseForm.ts
@@ -41,7 +41,7 @@ export default Form.extend({
 
     // When ALTCHA captcha is solved, trigger the form's save flow
     // This ensures IdentifierView.saveForm() runs (for deviceFingerprint, etc.)
-    this.listenTo(this.options.appState, 'onAltchaSolved', () => {
+    this.listenTo(this.options.appState, 'altchaSolved', () => {
       this.trigger('save', this.model);
     });
 
--- a/src/v2/view-builder/views/captcha/CaptchaView.js
+++ b/src/v2/view-builder/views/captcha/CaptchaView.js
@@ -117,7 +117,7 @@ export default View.extend({
       // Calls saveForm on the parent view (ex. IdentifierView) to trigger the form submission flow.
       // This ensures ParentView.saveForm() runs (for deviceFingerprint, etc. if applicable)
       // and the parent view triggers BaseForm.saveForm().
-      this.options.appState.trigger('onAltchaSolved');
+      this.options.appState.trigger('altchaSolved');
     };
 
     const setUpTempToken = () => {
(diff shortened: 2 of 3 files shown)


alexchen-okta:
is there any gen3 coverage needed for this change? or does the bug only apply to v2?

src/v2/view-builder/internals/BaseForm.ts
```diff
@@ -39,6 +39,12 @@ export default Form.extend({
       this.captchaObject = captchaObject;
     });
 
+    // When ALTCHA captcha is solved, trigger the form's save flow
+    // This ensures IdentifierView.saveForm() runs (for deviceFingerprint, etc.)
+    this.listenTo(this.options.appState, 'onAltchaSolved', () => {
```
alexchen-okta: nit: `onAltchaSolved` kind of reads like a callback, any thoughts on changing the app state property to something like `altchaSolved` instead?

ahmedalkoasmh-okta: Great point! Updated.

src/v2/view-builder/internals/BaseForm.ts
```diff
@@ -39,6 +39,12 @@ export default Form.extend({
       this.captchaObject = captchaObject;
     });
 
+    // When ALTCHA captcha is solved, trigger the form's save flow
+    // This ensures IdentifierView.saveForm() runs (for deviceFingerprint, etc.)
+    this.listenTo(this.options.appState, 'altchaSolved', () => {
+      this.trigger('save', this.model);
```
alexchen-okta, line 45: this seems to trigger `save` instead of `saveForm`, was this distinction intentional?

ahmedalkoasmh-okta: Triggering `save` instead of calling `saveForm` directly mimics what happens when a user clicks the Submit button which ensures we follow the normal form submission flow.

In `BaseForm'`s `initialize()`, there's a listener `this.listenTo(this, 'save', this.saveForm)` that catches the `save` event and routes it to the instance's `saveForm` method. If the form is an `IdentifierView` (which overrides saveForm), it will call `IdentifierView`'s version first to generate the `deviceFingerprint` before calling the base implementation.

PR #3962, opened 2026-02-12
OKTA-1083742: recover polling from failed fetch

shuowu-okta:
### Problem

When the polling encounters a transient network failure (e.g., server returning 500 non-JSON, or `fetch` throwing Error), the widget displays an error to the user and blocks UI interaction for some views (e.g. email verification). Once the network recovers, the user is stuck — they must refresh the page to resume verification.

### Root Cause

In `FormController.invokeAction`, all errors during polling were funneled through `showFormErrors`, displaying an error banner. There was no distinction between:
- **Server/IDX errors** (429 rate limit, session expired) — which have `rawIdxState` or `errorCode` and should be shown
- **Transient network errors** (fetch failures, non-JSON 500 responses) — which have neither and should be silently recovered from

### Solution

**Silent recovery for transient network failures during authenticator polling** — when `invokeAction` catches an error during a `-poll` action that is neither an IDX response nor a rate limit error, it silently restarts polling instead of showing an error.

### Scope

Gen2 only. The test is guarded with `if (userVariables.gen3) return`.

## PR Checklist

### Issue:

- [OKTA-1083742](https://oktainc.atlassian.net/browse/OKTA-1083742)

### DIFF
--- a/src/v2/controllers/FormController.ts
+++ b/src/v2/controllers/FormController.ts
@@ -249,7 +249,7 @@ export default Controller.extend({
         !IonResponseHelper.isIonErrorResponse(errorResp) 
         && !errorUtils.isRateLimitError(errorObj) 
         && isPollingAction 
-        && this.formView?.form?.startPolling
+        && typeof this.formView?.form?.startPolling === 'function'
       ) {
         // Silently restart polling for transient network failures
         this.formView.form.startPolling();


src/v2/controllers/FormController.ts
```diff
@@ -237,6 +238,24 @@ export default Controller.extend({
...
+      const errorObj = errorUtils.buildErrorObject(errorResp);
+      const isPollingAction = invokeOptions.actions?.[0]?.name?.endsWith('-poll');
+
+      // Here we only want to restart polling for network errors (non Ion errors) during polling actions
+      // As rate limit error is handled in view layer, we skip it here and leave to showFormErrors to decide whether to restart polling or not based on the view logic.
+      if (
+        !IonResponseHelper.isIonErrorResponse(errorResp) 
+        && !errorUtils.isRateLimitError(errorObj) 
+        && isPollingAction 
+        && this.formView?.form?.startPolling
```
Luxu2-Okta: should we check `typeof this.formView?.form?.startPolling === 'function'`

Luxu2-Okta: Just curious are these conditions will only filter 500 non json response? like if there is some other error like 503, should we continue polling or show the error message, the current implementation will also continue polling right?

shuowu-okta: showFormErrors function already get error presentation and polling restart (at view level), here we just targeting to a very special scenario to tell SIW to skip `showFormErrors` flow.

In the current implementation, it continues polling by default, but pause polliing for `rate limit` and `session expiration`, etc. The conditional logic in this PR make sure those cases will still be handled in the default flow (showFormErrors).

PR #3994, opened 2026-04-01
fix: add null guard for navigator.credentials.create result in WebAut…

kumarabhinav-okta:
## Description:

Webauthn Enrollment

When third-party credential managers (e.g. LastPass on Windows) intercept navigator.credentials.create(), they may resolve with null instead of throwing an InvalidStateError when a matching credential already exists in excludeCredentials. This causes a TypeError: Cannot read properties of null (reading 'id').

Add null checks on the credential result across all three engines (v1, v2, v3) to surface a user-friendly error instead of crashing.

## PR Checklist

### Issue:

- [OKTA-1147258](https://oktainc.atlassian.net/browse/OKTA-1147258)

### DIFF
--- a/packages/@okta/i18n/src/properties/login.properties
+++ b/packages/@okta/i18n/src/properties/login.properties
@@ -1046,6 +1046,7 @@ oie.enroll.webauthn.instructions.edge = Note: If you are enrolling a security ke
 oie.enroll.webauthn.save = Set up
 oie.webauthn.error.not.supported = Security key or biometric authenticator is not supported on this browser. Contact your admin for assistance.
 oie.webauthn.passkeysRebrand.error.not.supported = Passkeys are not supported on this browser. Contact your admin for assistance.
+oie.webauthn.error.nullcredential = Authenticator enrollment could not be completed. If a password manager intercepted this request, dismiss it and try again.
 oie.webauthn.error.invalidPasskey = Invalid passkey.
 oie.verify.webauth.title = Verify with Security Key or Biometric Authenticator
 oie.verify.webauth.passkeysRebrand.passkeys.title = Verify with a passkey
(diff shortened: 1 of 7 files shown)


src/v1/controllers/EnrollWebauthnController.js
```diff
@@ -84,6 +84,11 @@ export default FormController.extend({
             },
             excludeCredentials: getExcludeCredentials(activation.excludeCredentials),
           });
+          // Default rp.id per WebAuthn spec §5.4.2 — some credential managers
+          // (e.g. LastPass) require rp.id to be explicitly set
+          if (options.rp && !options.rp.id) {
+            options.rp.id = window.location.hostname;
+          }
```
stephanieyoshimoto-okta, line 91: this is a bug because of LastPass's implementation of the WebAuthn spec, right? if they fix the bug on their end, would we keep this check? essentially, is this a forever change for us or a band aid while we wait for a fix?

packages/@okta/i18n/src/properties/login.properties
```diff
@@ -1046,6 +1046,7 @@ oie.enroll.webauthn.instructions.edge = Note: If you are enrolling a security ke
 oie.enroll.webauthn.save = Set up
 oie.webauthn.error.not.supported = Security key or biometric authenticator is not supported on this browser. Contact your admin for assistance.
 oie.webauthn.passkeysRebrand.error.not.supported = Passkeys are not supported on this browser. Contact your admin for assistance.
+oie.webauthn.error.nullcredential = Authenticator enrollment could not be completed. If a password manager intercepted this request, dismiss it and try again.
```
stephanieyoshimoto-okta, line 1049: reminder that i18n keys need to be merged in a separate PR before this one so we prevent english leaks

alexchen-okta:
reached out to @kumarabhinav-okta, closing the PR as we're not intending on getting this change in

PR #4013, opened 2026-04-27
Okta-1156653: svg accessibility alt text

shuowu-okta:
## Description:                                     
  
- Add a visually-hidden role="status" live region that announces password requirement status changes to screen readers as the user types (e.g., "Not met: A lowercase letter, A number" or "All password requirements met")
- Move the live region outside the <figure> element (via Fragment) so aria-describedby on the password input doesn't include dynamic announcements  
- Remove aria-hidden from the PasswordMatches confirm-password list so it's accessible to screen readers
- Add 2 i18n keys: password.complexity.status.allMet, password.complexity.status.notMetPrefix

## PR Checklist

### Issue:

- [OKTA-1156653](https://oktainc.atlassian.net/browse/OKTA-1156653)

### Screenshot/Video:

https://github.com/user-attachments/assets/5b6c24ab-76d7-48af-a07a-5cf09f8212dd

### DIFF
--- a/src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
+++ b/src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
@@ -110,7 +110,7 @@ const PasswordRequirements: UISchemaElementComponent<{
     if (unmetRequirements.length === 0) {
       return loc('password.complexity.status.allMet', 'login');
     }
-    return loc('password.complexity.status.notMetPrefix', 'login', [unmetRequirements.join(', ')]);
+    return loc('password.complexity.status.notMetPrefix', 'login', [unmetRequirements.length, unmetRequirements.join(', ')]);
   }, [passwordValidations, requirements]);
 
   return requirements?.length > 0 ? (
--- a/src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
+++ b/src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
@@ -17,6 +17,7 @@ import { Fragment, h } from 'preact';
 import {
   useCallback,
   useEffect,
+  useMemo,
   useState,
 } from 'preact/hooks';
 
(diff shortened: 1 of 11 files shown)


packages/@okta/i18n/src/properties/login.properties
```diff
@@ -701,6 +701,11 @@ password.complexity.minAgeHours.description = At least {0} hour(s) must have ela
 # {0} is a number
 password.complexity.minAgeDays.description = At least {0} day(s) must have elapsed since you last changed your password
 
+## Password Requirement Accessibility
+password.complexity.status.allMet = All password requirements met
+# {0} is a comma-separated list of unmet password requirements
+password.complexity.status.notMetPrefix = Not met: {0}
```
alexchen-okta: We should have a separate PR for translations if we're going to consume these and hand it off before the handoff date on 5/13

src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
```diff
@@ -98,33 +98,67 @@ const PasswordRequirements: UISchemaElementComponent<{
...
+      .map(({ label }) => label)
+    : [];
+
+  let liveText = '';
+  if (hasValidations && unmetRequirements.length === 0) {
+    liveText = loc('password.complexity.status.allMet', 'login');
+  } else if (hasValidations) {
+    liveText = loc('password.complexity.status.notMetPrefix', 'login', [unmetRequirements.join(', ')]);
+  }
+
```
alexchen-okta: Is this something that should be solved at the Odyssey level?

packages/@okta/i18n/src/properties/login.properties
shuowu-okta: Done

src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
shuowu-okta: Odyssey's ScreenReaderText doesn't accept role/aria-live today, and there's no LiveRegion primitive — so solving this at the Odyssey level would mean proposing a new component to serve SIW only. The same pattern is already used in IdpList.tsx, so I'd like to keep it consistent here for easier update.

src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
```diff
@@ -98,33 +98,67 @@ const PasswordRequirements: UISchemaElementComponent<{
...
   ]);
 
+  const hasValidations = Object.keys(passwordValidations).length > 0;
+  const unmetRequirements = hasValidations
+    ? requirements
+      .filter(({ ruleKey }) => getPasswordStatus(ruleKey, passwordValidations) === 'incomplete')
+      .map(({ label }) => label)
+    : [];
+
+  let liveText = '';
```
leemchale-okta: Do you think its worth using `useMemo` (eg. `useMemo(() => { ... }, [passwordValidations, requirements]);)`,  for `unmetRequirements` and `liveText`?

src/v3/src/components/PasswordRequirements/PasswordRequirements.tsx
```diff
@@ -98,33 +98,67 @@ const PasswordRequirements: UISchemaElementComponent<{
...
+              label={label}
+            />
+          ))}
+        </Box>
       </Box>
       <Box
-        component="ul"
-        id={listId}
-        sx={{ listStyle: 'none', padding: '0', marginBlockStart: tokens.Spacing3 }}
+        role="status"
```
leemchale-okta, line 147: if the user types fast will the screen reader queue multiple announcements?

shuowu-okta: No queueing in practice. `role="status"` is `aria-live="polite"`, and screen readers replace (rather than queue) pending polite announcements — when they get a chance to speak, they read the current content, not every intermediate value.

Updates are also gated: `passwordValidations` is set from a debounced (`validationDelayMs`, default 50 ms) validator, so the live region's content only changes once the user pauses typing — not per keystroke. Verified with VoiceOver + fast typing: only the most-recent "Not met: …" string gets announced.

leemchale-okta (dismissed):
LGTM

PR #4024, opened 2026-05-06
Fix granular consent payload corruption when scope names share a prefix

TCezarRod:
## Description:

**Problem**

When an app requests both a plain scope and a dotted scope that shares its name as a prefix (e.g. `custom1` and `custom1.custom2`), the widget either throws a runtime error or silently drops the plain scope from the consent payload:

- If the plain scope is `true`: `unflatten` tries to set a property on a boolean, throwing `Cannot create property 'custom2' on boolean 'true'`
- If the plain scope is `false`: `unflatten` treats the falsy value as "not yet set" and overwrites it with a nested object, losing the value entirely

**Root cause**

`unflatten` in `Model.js` uses `if (!ref[part])` to decide whether to create an intermediate object node. This check is too coarse: it treats `false` (a valid unchecked scope value) the same as `undefined`, causing it to overwrite the existing boolean.

**Fix**

Two complementary changes:

1. **`Model.js`** — tightens the collision check in `unflatten` to `ref[part] != null && typeof ref[part] !== 'object'`. When a primitive already occupies an intermediate path node, the dotted key is left unexpanded rather than clobbering the existing value.

2. **`payloadTransformer.js`** — updates `transformOptedScopes` to collect any `optedScopes.*` top-level keys that `unflatten` intentionally left behind and merges them into the final flat payload.

## PR Checklist

- [x] Did you add tests, as appropriate, ~following our [Automated Test guidelines](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2676497890/Automated+Testing+in+the+Signin+Widget)~?
- [ ] ~Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?~
- [ ] ~Did you verify the change by running [downstream monolith artifact](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/102897979/Sign-in+Widget+Development#Sign-inWidgetDevelopment-Instructionstocreateandrunthedownstreamartifact(d16t))? (Provide link to build below)~
   -  No. Just a bug fix. 

### Issue:

- N/A

### Screenshot/Video:

Running the tests below after expanding the test cases with an additional scope `custom1.custom2` complementary to `custom1`. (First two commits in this PR)

```sh
yarn test -t testcafe --fixture "GranularConsent"
```

| Before | After |
|--------|--------|
| Screenshot 2026-05-06 at 16 37 43 | Screenshot 2026-05-06 at 16 51 32 | 

```sh
yarn test -t jest test/unit/spec/v2/ion/payloadTransformer_spec.js
```

| Before | After |
|--------|--------|
| Screenshot 2026-05-06 at 16 44 50 | Screenshot 2026-05-06 at 16 45 52 |

### DIFF
--- a/packages/@okta/courage-dist/esm/src/courage/framework/Model.js
+++ b/packages/@okta/courage-dist/esm/src/courage/framework/Model.js
@@ -90,8 +90,15 @@ function unflatten(data) {
     var parts = key.split('.');
 
     while ((part = parts.shift()) !== undefined) {
-      if (!ref[part]) {
-        ref[part] = parts.length ? {} : value;
+      if (parts.length) {
+        // Intermediate node — must be an object for traversal.
+        if (ref[part] != null && typeof ref[part] !== 'object') {
+          return;
+        } else if (!ref[part]) {
+          ref[part] = {};
+        }
+      } else if (!ref[part]) {
+        ref[part] = value;
       }
 
       ref = ref[part];
--- a/playground/mocks/data/idp/idx/consent-granular.json
+++ b/playground/mocks/data/idp/idx/consent-granular.json
@@ -44,6 +44,16 @@
                 "mutable": true
               },
               {
+                "name": "custom1.custom2",
+                "label": "View your mobile phone data plan sensitive details.",
+                "desc": "This allows the app to view your mobile phone data plan sensitive details.",
+                "type": "boolean",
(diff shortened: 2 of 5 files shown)


shuowu-okta:
Closing this PR, addressed in #4056

PR #4029, opened 2026-05-11
OKTA-1152243: Persist flow reset to meta for KMSI after MFA step-up

shuowu-okta:
## Description:

The original fix only mutated authClient.options.flow via setFlow() and passed flow='default' on the reset-authenticator submit. Subsequent proceed calls from later forms (MFA selection, OTP verify) omit the flow option, so auth-js falls back to meta.flow — which was still 'resetPassword' from interact() — and re-loads PasswordRecoveryFlow. keep-me-signed-in is not in that flow, so auth-js throws "No remediation can match current flow".

Also persist flow='default' to the saved transaction meta so the fallback in proceed() reads 'default' on every hop after reset-authenticator.

Adds a testcafe scenario that inserts an MFA step-up between reset-password and POST KMSI (Case 02866134).

## PR Checklist

### Issue:

- [OKTA-1152243](https://oktainc.atlassian.net/browse/OKTA-1152243)

### DIFF
--- a/src/v2/controllers/FormController.ts
+++ b/src/v2/controllers/FormController.ts
@@ -324,10 +324,18 @@ export default Controller.extend({
     // OKTA-1152243: After password reset, the server may return post-auth remediations
     // (e.g. keep-me-signed-in) that are not in PasswordRecoveryFlow. Reset to default flow
     // so auth-js does not throw when it encounters these in the recursive remediate() call.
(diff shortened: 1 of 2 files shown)


alexchen-okta:
do you have a local repro? if so, can we get videos for before/after? does this case only happen when there is a remediation following reset password and before KMSI?

alexchen-okta (approved):

leemchale-okta (approved):

PR #4049, opened 2026-06-05
OKTA-1142182: Render password history requirement in gen3 SIW

shuowu-okta:
## Description:

When an org's password policy enforces history (e.g. "Your password cannot be any of your last 4 passwords") or a minimum age between changes, gen2 SIW renders those rules in the password requirements list. Gen3 dropped both — the transformer ignored `data.age` — leaving customers like Vanguard with a generic "Password requirements were not met" error and no UI hint about which rule they violated.

This PR wires `age.historyCount` and `age.minAgeMinutes` through gen3's password requirements list, reusing the existing `getAgeFromMinutes` helper and existing i18n keys (`password.complexity.history.*`, `password.complexity.minAge*`). It also mirrors gen2's `ReEnrollAuthenticatorPasswordView.getPasswordPolicySettings` so the transformer prefers `recoveryAuthenticator` / `enrollmentAuthenticator` settings when present — on expired flows IDX exposes `currentAuthenticator` (the user's existing-password rules) via `relatesTo`. Gen2 behavior is unchanged.

The gen3 skip guards in three TestCafe specs that previously commented "decision was made by design to exclude history and age password requirements in gen3" are removed — the parity assertions now run on both engines.

## PR Checklist

### Issue:

- [OKTA-1142182](https://oktainc.atlassian.net/browse/OKTA-1142182)

### Screenshot/Video:

**Your password has expired**

| Before | After |
|--------|-------|
| before-expired-password | after-expired-password |

**Reset your password**

| Before | After |
|--------|-------|
| before-reset-password | after-reset-password |

### DIFF
--- a/src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
+++ b/src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
@@ -36,8 +36,19 @@ export const transformEnrollPasswordAuthenticator: IdxStepTransformer = ({
   transaction,
   formBag,
 }) => {
-  const { nextStep: { relatesTo } = {} } = transaction;
-  const passwordSettings = (relatesTo?.value?.settings || {}) as PasswordSettings;
+  const { context, nextStep: { relatesTo } = {} } = transaction;
+  // Mirror gen2 `ReEnrollAuthenticatorPasswordView.getPasswordPolicySettings`: for
+  // password-reset/expired flows IDX exposes the *current* password's policy via
+  // `nextStep.relatesTo` ($.currentAuthenticator), while the new-password policy lives
+  // on `recoveryAuthenticator` or `enrollmentAuthenticator`. Prefer those so requirements
+  // like "Password can't be the same as your last 4 passwords" surface in the UI.
+  const passwordSettings = (
+    // @ts-expect-error OKTA-1142182 - recoveryAuthenticator missing from IdxContext type
+    context?.recoveryAuthenticator?.value?.settings
+    || context?.enrollmentAuthenticator?.value?.settings
+    || relatesTo?.value?.settings
+    || {}
+  ) as PasswordSettings;
 
   const { uischema, dataSchema } = formBag;
   const userInfo = getUserInfo(transaction);
--- a/src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
+++ b/src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
@@ -12,6 +12,7 @@
 
 import { PASSWORD_REQUIREMENT_VALIDATION_DELAY_MS } from '../../constants';
 import {
+  ExtendedIdxContext,
   FieldElement,
   FormBag,
   HiddenInputElement,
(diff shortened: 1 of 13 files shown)


src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
```diff
@@ -36,8 +36,19 @@ export const transformEnrollPasswordAuthenticator: IdxStepTransformer = ({
...
+  // password-reset/expired flows IDX exposes the *current* password's policy via
+  // `nextStep.relatesTo` ($.currentAuthenticator), while the new-password policy lives
+  // on `recoveryAuthenticator` or `enrollmentAuthenticator`. Prefer those so requirements
+  // like "Password can't be the same as your last 4 passwords" surface in the UI.
+  const passwordSettings = (
+    // @ts-expect-error OKTA-1142182 - recoveryAuthenticator missing from IdxContext type
+    context?.recoveryAuthenticator?.value?.settings
+    || context?.enrollmentAuthenticator?.value?.settings
+    || relatesTo?.value?.settings
+    || {}
```
Luxu2-Okta, line 51: curious, in Gen2 [ReEnrollAuthenticatorPasswordView](https://github.com/okta/okta-signin-widget/blob/19243354702eaba9403c13fef0ed766c089605ba/src/v2/view-builder/views/password/ReEnrollAuthenticatorPasswordView.js#L19) doesn't have the `relatesTo?.value?.settings`, so we still keep `relatesTo?.value?.settings` in Gen3, but actually it should almost never be used since the first two will cover it right?

Luxu2-Okta (approved):
LGTM, have a question though

src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
```diff
@@ -36,8 +36,19 @@ export const transformEnrollPasswordAuthenticator: IdxStepTransformer = ({
...
-  const { nextStep: { relatesTo } = {} } = transaction;
-  const passwordSettings = (relatesTo?.value?.settings || {}) as PasswordSettings;
+  const { context, nextStep: { relatesTo } = {} } = transaction;
+  // Mirror gen2 `ReEnrollAuthenticatorPasswordView.getPasswordPolicySettings`: for
+  // password-reset/expired flows IDX exposes the *current* password's policy via
+  // `nextStep.relatesTo` ($.currentAuthenticator), while the new-password policy lives
+  // on `recoveryAuthenticator` or `enrollmentAuthenticator`. Prefer those so requirements
+  // like "Password can't be the same as your last 4 passwords" surface in the UI.
+  const passwordSettings = (
+    // @ts-expect-error OKTA-1142182 - recoveryAuthenticator missing from IdxContext type
```
alexchen-okta: is a ts-expect-error needed here? should we add it to the `IdxContent` type?

src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
shuowu-okta: Good catch. Pre-fix the only path here was `relatesTo?.value?.settings`, so I kept it as the third fallback to avoid regressing any flow that already relied on it — the first two cover the reset/expired bug from this ticket but I didn't want to expand scope by removing the original fallback. In practice for reset/expired flows `relatesTo` resolves to `currentAuthenticator` (i.e. the *old* password's policy) which is exactly what we want to skip, hence the ordering. Happy to drop it in a follow-up if we can confirm no enrollment flow depends on it.

src/v3/src/transformer/password/transformEnrollPasswordAuthenticator.ts
shuowu-okta: Good question. `recoveryAuthenticator` is genuinely missing from the upstream `IdxContext` type in `@okta/okta-auth-js` (`node_modules/@okta/okta-auth-js/types/lib/idx/types/idx-js.d.ts:147` — has `currentAuthenticator`, `currentAuthenticatorEnrollment`, `enrollmentAuthenticator`, but not `recoveryAuthenticator`). Bumping auth-js just for this is a high-risk change we'd rather defer to a dedicated release where it can be properly tested. I'll push a follow-up that adds a local extended type in `src/v3/src/types/` so we can drop the `@ts-expect-error` cleanly without an upstream upgrade.

leemchale-okta:
@shuowu-okta  is this behavior gated to only reset and expired password flows and not brand new password enrollement?

shuowu-okta:
> @shuowu-okta is this behavior gated to only reset and expired password flows and not brand new password enrollement?

The change should cover all flows, see the testcafe changes in PR for what gen2 parity is covered in detail.

leemchale-okta (approved):

PR #4056, opened 2026-06-15
OKTA-1174752: Fix granular consent payload when scopes share a prefix

shuowu-okta:
## Description:

When a granular-consent form contains both a plain scope and a dotted scope sharing its prefix (e.g. `custom1` and `custom1.custom2`), Courage's `Model.toJSON()` `unflatten` pass either threw `Cannot create property 'custom2' on boolean 'true'` (blocking the form from rendering) or silently dropped the user's `false` selection from the payload — leaving the user unable to grant consent. Customer-reported by Brex (#4024).

The crash fires twice in gen2:
- **At render time** — `BaseView.renderForm` calls `model.toJSON({verbose: true})`, which invokes `unflatten`. With prefix-colliding scopes, `unflatten` either throws (truthy plain scope) or silently overwrites it with an empty object (falsy plain scope, since `if (!ref[part])` treats `false` as unset).
- **At submit time** — the same `toJSON()` round-trip used to be the only way the payload was built, so the same hazard reappeared on the wire.

This PR fixes both with two contained changes scoped to granular consent only:

1. **`GranularConsentView.createModelClass`** returns a `Model.extend({ flat: false })`. `flat` is an existing Courage feature that disables the `flatten`/`unflatten` round-trip on this one model class. With `flat: false`, `model.toJSON()` returns a plain shallow clone of `model.attributes` (flat dotted keys), so the render-time call no longer crashes. One-line change plus a comment explaining why.

2. **`payloadTransformer.transformOptedScopes`** is rewritten to operate on flat input. Instead of re-flattening a nested object that `unflatten` produced (which crashed/lost data on prefix collisions), it now groups the flat `optedScopes.<scope>` keys from `model.toJSON()` into the nested `optedScopes` object the IDX backend expects, preserving dotted scope names verbatim. The function signature, `transformPayload` dispatch, and `FormNameToTransformerHandler` registry are unchanged.

No Courage changes; no other forms affected. Gen3 was never broken (it goes through `toNestedObject` with `keysWithoutNesting=['optedScopes']`, which already handles prefix collisions correctly), but a defensive regression test was added there to lock in the behavior for both customer scenarios.

## PR Checklist

### Issue:

- [OKTA-1174752](https://oktainc.atlassian.net/browse/OKTA-1174752)
- Customer-reported via #4024 (Brex)

### Test coverage:

- `test/unit/spec/v2/ion/payloadTransformer_spec.js` — both customer scenarios (true case + false case) at the submit boundary, plus dotted scope preservation
- `test/unit/spec/v2/view-builder/views/consent/GranularConsentView_spec.js` — verifies `flat: false` model: render-time `toJSON({verbose: true})` does not crash; flat shape preserved
- `test/testcafe/spec/GranularConsentView_spec.js` — end-to-end regression with the new prefix-collision mock (`playground/mocks/data/idp/idx/consent-granular-prefix-collision.json`)
- `src/v3/src/util/toNestedObject.test.ts` — defensive gen3 tests for both customer scenarios

### DIFF
--- a/src/v3/src/util/toNestedObject.test.ts
+++ b/src/v3/src/util/toNestedObject.test.ts
@@ -50,4 +50,44 @@ describe('toNestedObject', () => {
       },
     });
   });
+
+  // OKTA-1174752 / customer PR #4024 (Brex): gen2 Courage's `unflatten` mishandled
+  // a plain scope sharing its prefix with a dotted scope. Gen3 doesn't go through
+  // Courage at all (React form state -> useOnSubmit -> toNestedObject with
+  // keysWithoutNesting=['optedScopes']), so there is no render-time analog of
+  // the gen2 crash. The submit-time grouping below pins gen3's correct behavior
+  // for both customer scenarios.
+
+  // Customer scenario 1: plain scope is `true`. Gen2's unflatten threw
+  // "Cannot create property 'custom2' on boolean 'true'".
+  it('keeps a true plain key and a dotted key sharing its prefix as distinct siblings', () => {
+    expect(toNestedObject({
+      consent: true,
+      'optedScopes.custom1': true,
+      'optedScopes.custom1.custom2': false,
+    }, ['optedScopes'])).toEqual({
+      consent: true,
+      optedScopes: {
+        custom1: true,
+        'custom1.custom2': false,
+      },
+    });
+  });
+
+  // Customer scenario 2: plain scope is `false`. Gen2's unflatten check
+  // `if (!ref[part])` treated `false` as unset and silently overwrote it
+  // with an empty object, dropping the user's choice from the payload.
+  it('preserves a false plain key when a dotted key shares its prefix', () => {
+    expect(toNestedObject({
+      consent: true,
+      'optedScopes.custom1': false,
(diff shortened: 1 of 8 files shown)


src/v3/src/util/toNestedObject.test.ts
```diff
@@ -50,4 +50,44 @@ describe('toNestedObject', () => {
...
+      'optedScopes.custom1': false,
+      'optedScopes.custom1.custom2': true,
+    }, ['optedScopes'])).toEqual({
+      consent: true,
+      optedScopes: {
+        custom1: false,
+        'custom1.custom2': true,
+      },
+    });
+  });
```
Luxu2-Okta, line 92: I guess Gen3 doesn't have this issue, but we still add tests

Luxu2-Okta (approved):
LGTM

src/v3/src/util/toNestedObject.test.ts
shuowu-okta: Right, just use these tests to make sure gen3 is free from this bug.

leemchale-okta (approved):

PR #4063, opened 2026-06-17
OKTA-1182955: gate post-AppLink redirect on focus, not just visibility

yannongli-okta:
## Description:
- Prior backend change https://github.com/atko-eng/okta-core/pull/133512
- Demo https://drive.google.com/file/d/1l5ipGvWUGd8VglvqyVjGo-X2ltMqrztl/view?usp=drive_link

**Root Cause**
 When Android FastPass authentication is invoked via AppLink (Okta Verify launched as a separate app), the user returns to the SIW tab with the page reported as visible briefly before the window has actually regained focus. The post-auth redirect today is gated only on `document.visibilityState` via `Util.executeOnVisiblePage` (added under OKTA-702402), which fires inside that "visible-but-not-focused" window and ands the user on login.okta.com instead of completing the auth flow.                                                                          
                                                                                                               
  Non-AppLink flows don't leave the browser, so they don't have this gap — universally tightening executeOnVisiblePage would risk regressing every other redirect, so the fix is scoped to the AppLink case.

**Fix**
   Rely on a backend-attached signal on the success-redirect remediation and require an explicit user click to trigger the final redirect — a click event inherently guarantees window focus, so it sidesteps the visible-but-not-focused race entirely.                                      
  1. Backend signal (from okta-core#133512): the success link on the IDX response carries an optional `priorVerification: { method: 'APP_LINK',  success: true}` descriptor when the prior step was a successful Android AppLink FastPass verification.       
  2. Shared helper `Util.isPostAppLinkVerification(viewStateOrLink)` in `src/util/Util.js` — returns true only for the strict `method === 'APP_LINK' && success === true` shape. Used from the gen2 sites; gen3 uses an equivalent inline check for the same shape.                                  
  3. Forward the signal onto the transformed remediation:
    - gen2: `src/v2/ion/responseTransformer.js` copies priorVerification onto the success remediation value.                                       
    - gen3: `src/v3/src/transformer/redirect/redirectTransformer.ts` copies it onto the RedirectElement (guarded by success.href === url so it only attaches on the actual success-redirect path).                                                                                                
  4. Suppress the auto-redirect and render a Continue button when the signal is present:                                                         
    - `gen2: FormController.ts` short-circuits handleSaveForm so no auto-redirect fires, and `AutoRedirectView.js` renders a createButton  (#applink-continue-redirect, label "Continue") in place of the spinner. Click calls `Util.redirectWithFormGet(currentViewState.href)`.           
    - gen3: `Redirect.tsx` skips the executeOnVisiblePage call and renders the shared <Button> component (variant primary, wide). Click calls Util.changeLocation(options.url).                                                                                                              
  5. Type update: RedirectElement.options.priorVerification added as optional in `src/v3/src/types/schema.ts`.   
                                                                                                                                                 
  All non-Android-AppLink paths are unchanged — without the backend field the behavior is identical to today. 

## PR Checklist

### Issue:

- [OKTA-1182955](https://oktainc.atlassian.net/browse/OKTA-1182955)

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -247,6 +247,16 @@ Util.executeOnVisiblePage = function(cb) {
   }
 };
 
+// OKTA-1182955: backend may attach a priorVerification descriptor to the success-redirect remediation
+// indicating the prior step was a successful Android AppLink (FastPass) verification. In that case the
+// SIW must not auto-redirect — the page can be `visible` before the window has regained focus and the
+// redirect lands on login.okta.com. The view renders a button instead; the user click guarantees focus
+// before the redirect fires.
+Util.isPostAppLinkVerification = function(viewStateOrLink) {
+  const pv = viewStateOrLink && viewStateOrLink.priorVerification;
+  return !!pv && pv.method === 'APP_LINK' && pv.success === true;
+};
+
 /**
  * Why redirect via Form get rather using `window.location.href`?
  * At the time of writing, Chrome (<72) in Android would block window location change
--- a/src/v2/controllers/FormController.ts
+++ b/src/v2/controllers/FormController.ts
@@ -290,6 +290,13 @@ export default Controller.extend({
       sessionStorageHelper.removeStateHandle();
 
       const currentViewState = this.options.appState.getCurrentViewState();
+      // OKTA-1182955: when the backend signals a successful prior Android AppLink (FastPass)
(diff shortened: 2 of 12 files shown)


glenfannin-okta (dismissed):

Luxu2-Okta (approved):
LGTM

PR #4068, opened 2026-06-22
OKTA-1194394: Run before/after hooks on switchForm navigation

shuowu-okta:
## Description:

Customer-registered `oktaSignIn.before(formName, fn)` / `oktaSignIn.after(formName, fn)` hooks fired correctly on the first render of `select-authenticator-authenticate` (driven by an IDX server response), but were silently skipped on every subsequent render triggered by clicking "Verify with something else". The root cause was that `FormController.handleSwitchForm` set `currentFormName` directly, bypassing the hook-execution path that lives in `AppState.setIonResponse`.

The fix mirrors the `setIonResponse` ordering in `handleSwitchForm`: run `executeHooksBefore`, set `currentFormName` (which triggers re-render), then run `executeHooksAfter`. This applies to any client-side `switchForm` navigation, including repeated use of "Verify with something else" in the Forgot Password and Unlock Account flows. v2 (OIE) only; no UI changes.

## PR Checklist

### Issue:

- [OKTA-1194394](https://oktainc.atlassian.net/browse/OKTA-1194394)

### DIFF
--- a/src/v2/controllers/FormController.ts
+++ b/src/v2/controllers/FormController.ts
@@ -19,6 +19,7 @@ import {
 } from '../ion/RemediationConstants';
 import transformPayload from '../ion/payloadTransformer';
 import Util from 'util/Util';
+import { executeHooksBefore, executeHooksAfter } from 'util/Hooks';
 import sessionStorageHelper from '../client/sessionStorageHelper';
 import { HttpResponse, IdxStatus, ProceedOptions } from '@okta/okta-auth-js';
 import { EventErrorContext } from 'types/events';
@@ -141,7 +142,7 @@ export default Controller.extend({
     return eventData;
   },
 
-  handleSwitchForm(formName) {
+  async handleSwitchForm(formName) {
     // trigger formName change to change view
     if (this.options.appState.get('messages')) {
       // Clear messages before calling switch form.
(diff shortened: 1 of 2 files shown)


src/v2/controllers/FormController.ts
```diff
@@ -150,7 +151,17 @@ export default Controller.extend({
...
+    // hooks fire only on renders driven by an IDX response (see AppState.setIonResponse),
+    // and any client-side switchForm navigation (e.g. "Verify with something else")
+    // would silently skip them.
+    const hook = this.options.appState.hooks?.getHook(formName);
+    await executeHooksBefore(hook);
+    // Backbone fires `change:currentFormName` and re-renders synchronously inside
+    // set(), so the DOM is fully updated by the time set() returns — which means
+    // executeHooksAfter below runs against the new form, not the old one.
     this.options.appState.set('currentFormName', formName);
+    await executeHooksAfter(hook);
```
Luxu2-Okta, line 164: Do we also want to check Gen3?

Luxu2-Okta (approved):
Have a question about Gen3, but non blocker, LGTM

leemchale-okta (approved):

src/v2/controllers/FormController.ts
shuowu-okta: As some gen2 hooks does not work well in gen3, I would propose just focus on gen2 scope for now.

PR #4115, opened 2026-08-07
fix(a11y): ReminderPrompt uses aria-live polite to avoid interrupting screen readers

drashtitrivedi-okta:
## Description:

- The resend reminder prompt (`ReminderPrompt`) previously rendered inside Odyssey's `<Callout severity="warning">`, which defaults to `role="alert"` (`aria-live="assertive"`). This caused screen readers to **immediately interrupt** any ongoing announcement (e.g., a validation error) to read the reminder which is the wrong priority for a non-critical resend prompt. 
- **Fix:** Wrap the callout in a `<Box aria-live="polite" aria-atomic="true">` container and set `role="presentation"` on the inner `<Callout>` to neutralize its built-in live region. The outer `aria-live="polite"` box announces the reminder only after the screen reader finishes its current announcement.
-  The `role="alert"` was removed from the Callout, so the existing e2e selectors that relied on                queryAllByRole('alert') could no longer find the reminder. Added `data-se="reminder-prompt"` as a stable test selector instead, following the existing gen3 convention.
- No visual change. only the ARIA announcement priority changes.                                 
  
**Tested with:**  Verified the fix works correctly on VoiceOver (macOS).

## PR Checklist

- [X] Did you add tests, as appropriate, following our [Automated Test guidelines](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2676497890/Automated+Testing+in+the+Signin+Widget)?

### Issue:

- [OKTA-1227828](https://oktainc.atlassian.net/browse/OKTA-1227828)

### Screenshot/Video:

**Fix a11y before/after**

**Before**:

https://github.com/user-attachments/assets/3f922a0b-58dc-4f76-b3e4-4cf242675086

**After**:

https://github.com/user-attachments/assets/00ee301b-38a7-4aff-b20b-320ed0d6b4d5

**No UI change before/after**

**Before** 

https://github.com/user-attachments/assets/275fbfcf-4754-4134-86d1-ab8a6eebd8ce 

**After**

https://github.com/user-attachments/assets/75e6b43f-bc8f-4e88-aeec-7b9bf87c21e6

https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&page=1&pageSize=6&sha=bdac8b1c6a60f4dac0e79acb7e43d53f57913f53&tab=main

### DIFF
--- a/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
+++ b/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
@@ -131,18 +131,26 @@ const ReminderPrompt: UISchemaElementComponent<{
     );
   };
 
-  return show ? (
-    <Box marginBlockEnd={tokens.Spacing4}>
-      <Callout
-        severity="warning"
-        // visually-hidden severity text is not translated
-        translate="no"
-      >
-        {renderAlertContent()}
-        {renderActionLink()}
-      </Callout>
+  return (
+    <Box
+      aria-live="polite"
+      aria-atomic="true"
+    >
+      {show && (
+        <Box marginBlockEnd={tokens.Spacing4}>
+          <Callout
+            severity="warning"
+            role={"presentation" as unknown as "status"}
+            // visually-hidden severity text is not translated
+            translate="no"
+          >
+            {renderAlertContent()}
+            {renderActionLink()}
+          </Callout>
+        </Box>
+      )}
     </Box>
-  ) : null;
+  );
 };
 
 export default ReminderPrompt;
--- a/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
+++ b/src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
@@ -137,10 +137,13 @@ const ReminderPrompt: UISchemaElementComponent<{
       aria-atomic="true"
     >
       {show && (
-        <Box marginBlockEnd={tokens.Spacing4}>
+        <Box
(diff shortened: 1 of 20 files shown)


src/v3/src/components/ReminderPrompt/ReminderPrompt.tsx
```diff
@@ -131,18 +131,29 @@ const ReminderPrompt: UISchemaElementComponent<{
...
-        translate="no"
-      >
-        {renderAlertContent()}
-        {renderActionLink()}
-      </Callout>
+  return (
+    <Box
+      aria-live="polite"
+      aria-atomic="true"
+    >
```
leemchale-okta, line 138: is it okay to always have this box present in the DOM?  Should it be inside the conditional rendering below?

drashtitrivedi-okta: yes @leemchale-okta this is intentional. the `aria-live` wrapper needs to already be in the DOM before any content appears inside it. that is how screen readers know to watch it for changes. if we move it inside the conditional rendering block, the wrapper and the content would appear at the same time, and screen readers would miss the announcement. 

reference: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-live

leemchale-okta: got it thank you

leemchale-okta (approved):

PR #4131, opened 2026-09-04
fix(Gen3): associate lone password field with form-level error

Luxu2-Okta:
## Description:

On the identifier-first password challenge, a failed credential check renders a form-level error alert above the form. The alert is announced once on insertion, but the password input itself conveyed nothing, so a screen reader user returning to the field received no indication of the outstanding error.

Give the message an id and reference it from the password input via aria-describedby, so focusing the field announces the error. No visual change.

Scoped to views rendering exactly one field which is a password. On the combined username + password form the server returns a generic message (errors.E0000004, identical for a wrong password and an unknown user), so the failure is not attributable to either field; associating one would imply the other was accepted, which is a user-enumeration signal. An integration test locks that in.

The id is carried on a new InfoboxElement options.messageId, set only when something references it, so no other view's markup changes. It is namespaced as okta-sign-in-form-error rather than a bare "form-error", which could collide with the markup of the page embedding the widget.

aria-invalid is deliberately left as "false". Per WAI-ARIA it asserts that "the value entered by the user has failed validation", which the server does not assert. Odyssey also derives aria-invalid and aria-errormessage solely from its errorMessage prop, so setting them would require rendering duplicate visible error text. aria-describedby is the supported mechanism and is what the W3C WAI forms guidance names for associating a field with an error message.

And Odyssey doesn't expose `aria-invalid` and `aria-errormessage` to us, ideally we need to open a feature request to Odyssey team and update Odyssey, which is high effort and high risk, so only update `ariaDescribedBy` is a quick mitigation

## PR Checklist

### Issue:

- [OKTA-1257415](https://oktainc.atlassian.net/browse/OKTA-1257415)

#### without fix
After submitting wrong password, the form level "Unable to Sign in" element shows, but when tabbing to the password field, no error message on VoiceOver
Screenshot:
Screenshot 2026-09-04 at 4 04 44 PM

Video:

https://github.com/user-attachments/assets/e4fefe94-e910-45e5-bfb5-e2ec7766d8ce

#### after fix
After submitting wrong password, the form level "Unable to Sign in" element shows, but we use `aria-describedby` to point to the form-level error, so when tabbing to the password field, error message is read on VoiceOver
Screenshot:
Screenshot 2026-09-04 at 4 38 33 PM

Video:

https://github.com/user-attachments/assets/06bbecb0-2f60-41d2-9e53-630e27179b25

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved accessibility for generic password-verification errors by linking the form-level message to the password field.
  - Preserved existing accessibility descriptions while ensuring error messages are announced correctly.
  - Corrected error presentation so generic sign-in failures are not incorrectly shown as field-specific validation errors.

- **Tests**
  - Added coverage for incorrect-password flows, error announcements, form rendering, and accessibility relationships.

### DIFF
--- a/src/v3/src/components/InfoBox/InfoBox.tsx
+++ b/src/v3/src/components/InfoBox/InfoBox.tsx
@@ -36,10 +36,32 @@ const InfoBox: UISchemaElementComponent<{
       message,
       class: messageClass,
       dataSe,
+      messageId,
     },
   } = uischema;
 
   const tokens = useOdysseyDesignTokens();
+
+  const messageContent = Array.isArray(message)
+    ? message.map((msg) => (
+      <Box
+        marginBlockEnd={tokens.Spacing2}
+        key={msg.message}
+      >
+        <WidgetMessageContainer
+          key={msg.message}
+          message={msg}
+          parserOptions={{ replace: getLinkReplacerFn({}, 'monochrome') }}
+          linkVariant="monochrome"
+        />
+      </Box>
+    ))
+    : (
+      <WidgetMessageContainer
+        message={message}
+        linkVariant="monochrome"
+      />
+    );
 
   return loading ? null : (
     <Box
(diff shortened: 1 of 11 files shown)


src/v3/src/components/InfoBox/InfoBox.tsx
```diff
@@ -36,11 +36,33 @@ const InfoBox: UISchemaElementComponent<{
...
   const tokens = useOdysseyDesignTokens();
 
+  const messageContent = Array.isArray(message)
+    ? message.map((msg) => (
+      <Box
+        marginBlockEnd={tokens.Spacing2}
+        key={msg.message}
+      >
+        <WidgetMessageContainer
+          key={msg.message}
```
drashtitrivedi-okta, line 52: nit: React only needs the key attribute on the outer top-level element inside an array map function, so the inner key attribute can be removed to clean up unnecessary props.

PR #4077, opened 2026-07-06
fix (gen3): use "Set up" label for security question enroll button

leemchale-okta:
## Description:

The submit button on the **Enroll Security Question** screen (Gen3/v3) used the `mfa.challenge.verify` i18n key, which resolves to **"Verify"**. "Verify" is a challenge/verification label and does not convey the enrollment purpose of the control, so screen reader users do not get the intended purpose of the button (WCAG 2.4.6, Deque audit).

This change switches both the predefined-question and custom-question submit buttons to `oie.enroll.authenticator.button.text`, which resolves to **"Set up"** — matching the label already used by other authenticators (Phone, Okta Verify, Symantec, Yubikey) on their enrollment screens, per the design proposal on the ticket.

No new i18n string is needed; the `oie.enroll.authenticator.button.text` key already exists.

Downstream test: https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&page=1&pageSize=6&sha=80e149f266d219fd1322f3909d149562e0f1b15b&tab=main

Downstream v3 nightly: https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&page=1&pageSize=6&sha=eb822d8153e8dd9cd602bdb13a6f1524daac4d62&tab=main

## PR Checklist

- [x] Have you verified the basic functionality for this change?
- [x] Did you add tests, as appropriate, following our [Automated Test guidelines](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2676497890/Automated+Testing+in+the+Signin+Widget)?
- [x] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?
- [ ] Did you verify the change by running [downstream monolith artifact](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/102897979/Sign-in+Widget+Development#Sign-inWidgetDevelopment-Instructionstocreateandrunthedownstreamartifact(d16t))? (Provide link to build below)
- [x] Does this PR include noticeable changes to the UI? (If yes, attach screenshots/video below)

### Issue:

- [OKTA-1164547](https://oktainc.atlassian.net/browse/OKTA-1164547)

### Screenshot/Video:

The security question enrollment submit button now reads **"Set up"** instead of **"Verify"**.

**Before:**
Screenshot 2026-07-06 at 2 29 37 PM

**After:**
Screenshot 2026-07-06 at 2 29 24 PM

### Downstream Monolith Build:

[...]

PR #3550, opened 2024-02-13
Merge Odyssey 1.x Migration

trevoring-okta:
## Description:
- Merge Odyssey 1.x migration work into master

## PR Checklist

- [ ] Have you verified the basic functionality for this change?
- [ ] Did you add tests, as appropriate, following our [Automated Test guidelines](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2676497890/Automated+Testing+in+the+Signin+Widget)?
- [ ] Did you follow our [Security Best Practices](https://oktawiki.atlassian.net/wiki/display/eng/Security+Best+practices)?
- [ ] Did you verify the change by running [downstream monolith artifact](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/102897979/Sign-in+Widget+Development#Sign-inWidgetDevelopment-Instructionstocreateandrunthedownstreamartifact(d16t))? (Provide link to build below)
- [ ] Does this PR include noticeable changes to the UI? (If yes, attach screenshots/video below)

### Issue:

- [OKTA-695838](https://oktainc.atlassian.net/browse/OKTA-695838)

PR #4138, opened 2026-09-11
fix (gen3): always use "Set up" label on select-authenticator-enroll

shuowu-okta:
Gen3 rendered "Set up another" for an authenticator whose key already appeared in authenticatorEnrollments. In a self-registration flow with mandatory email verification, email (and password) get enrolled during registration, so the subsequent select-authenticator-enroll screen showed "Set up another" for email even though the user is enrolling it for the first time.

Gen2 OIE (src/v2 AuthenticatorEnrollOptions) has no "Set up another" concept at all — it always renders "Set up". The "Set up another" logic was gen3-only, introduced by the Odyssey list restyle (OKTA-506882 / #3550), and produced this parity gap. The classic engine's smart 0-vs-1-enrollment wording is cardinality-driven, and cardinality is not exposed anywhere in the IDX response, so gen3 cannot replicate it.

This aligns gen3 with gen2 by removing the isAdditionalEnroll / isAuthenticatorAlreadyEnrolled path so the enroll CTA is always "Set up". Follows the same gen2-parity direction as the security-question enroll button fix (OKTA-1164547 / #4077).

## PR Checklist

### Issue:

- [OKTA-1245836](https://oktainc.atlassian.net/browse/OKTA-1245836)

## Summary by CodeRabbit

- **Changes**
  - Additional-enrollment actions are no longer offered in the authenticator selection flow.
  - Authenticator setup buttons now use the standard setup label and arrow styling.
  - The enumerate-authenticators demo now displays five setup options instead of a separate additional-enrollment option.

- **Bug Fixes**
  - Enrollment selection behavior is now consistent across supported runtime versions, including optional enrollment scenarios.


src/v3/src/transformer/selectAuthenticator/utils.ts
```diff
@@ -405,8 +391,6 @@ const formatAuthenticatorOptions = (
       const id = getOptionValue(option.value as Input[], 'id')?.value;
       const methodType = getOptionValue(option.value as Input[], 'methodType')?.value;
       const enrollmentId = getOptionValue(option.value as Input[], 'enrollmentId')?.value;
-      const isAdditionalEnroll = isEnroll && authenticator
-        && isAuthenticatorAlreadyEnrolled(authenticator, authenticatorEnrollments);
```
Luxu2-Okta, line 409: @shuowu-okta , this line should be the culprit for this customer, but the 
```
const isAdditionalEnroll = isEnroll && ovRemediation?.relatesTo
    && isAuthenticatorAlreadyEnrolled(ovRemediation.relatesTo, authenticatorEnrollments);
```
in function `buildOktaVerifyOptions` seems for OV case, I understand we can remove them all since they are all introduced in the same PR before, but just want to callout

Luxu2-Okta (approved):
LGTM if we want to remove all `isAdditionalEnroll`, calling out for the OV case

leemchale-okta (approved):

src/v3/screenshots/base/UI_demo/UI_demo_RTL_VRT.png
alexchen-okta, line 1: can we double check this VRT? seems like we are inadvertently removing the "Authenticator Button 6 for Additional Enroll" component here

src/v3/src/transformer/selectAuthenticator/utils.ts
shuowu-okta: Thanks Lu, I have scoped this PR to only touch ctaLabel related code.

src/v3/screenshots/base/UI_demo/UI_demo_RTL_VRT.png
shuowu-okta: Thanks Alex, PR has been updated to only ctaLabel scope, which avoids the diff in VRT.

alexchen-okta (approved):

